PR #2, opened 2016-10-13
Cookiecutter initialization

e-kolpakov:
**Description:** App initialization using https://github.com/edx/cookiecutter-django-app


mattdrayer:
Getting started on the right foot with the cookiecutter initialization -- awesome!  👍

PR #4, opened 2016-10-24
Updated pypi deployment secret

kylecrawshaw:
@e-kolpakov I've updated the secrets just to make sure I didn't do it incorrectly the first time. As soon as this is merged to master it should attempt to deploy again.


clintonb:
I don't believe the credentials are the problem. The creation/first release of the package needs to be done manually. Travis won't create new packages. https://pypi.python.org/pypi/edx-enterprise

kylecrawshaw:
Thanks @clintonb. I didn't realize that was the case!

clintonb:
No worries. I see now that this information is not on the wiki page at https://openedx.atlassian.net/wiki/display/OpenOPS/Publishing+a+Package+to+PyPI+using+Travis.

kylecrawshaw:
I created the PKG_INFO and uploaded that to PyPi. It should be good to go. Closing out this PR now. Thanks @clintonb!

@e-kolpakov — On next commit to master, let's see what happens.

PR #5, opened 2016-10-26
ENT-20 Enterprise Customer Setup: Configure Cobranding

asadiqbal08:
[ENT-20](https://openedx.atlassian.net/browse/ENT-20)

@mattdrayer @e-kolpakov your initial thoughts ? 

screen shot 2016-10-26 at 3 46 18 pm
##### NOTE: It is in WIP State. missing test coverage

**Reviewers:**
- [x] @e-kolpakov
- [ ] @mattdrayer  (???)

**Checklist:**
- [ ] Reviewed and approved
- [ ] Add yourself to AUTHORS
- [ ] Bump version number in `enterprise/__init__.py`: 0.1.1
- [ ] Document new in this version CHANGELOG.rst
- [ ] Add relevant information to README and/or docs
- [ ] Make sure CI build pass
- [ ] (optional) achieve 100% diff coverage

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,15 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[0.1.1] - 2016-11-03
+~~~~~~~~~~~~~~~~~~~~
+
+Added
+_____
+
+* Enterprise Customer Branding Model and Django admin integration
+
+
 [0.1.0] - 2016-10-13
 ~~~~~~~~~~~~~~~~~~~~
 
--- a/enterprise/admin.py
+++ b/enterprise/admin.py
@@ -8,7 +8,7 @@ from simple_history.admin import SimpleHistoryAdmin  # likely a bug in import or
 from django.contrib import admin
 
 from enterprise.actions import export_as_csv_action
-from enterprise.models import EnterpriseCustomer, EnterpriseCustomerUser
+from enterprise.models import EnterpriseCustomer, EnterpriseCustomerBrandingConfiguration, EnterpriseCustomerUser
 
 
 def get_all_field_names(model):
@@ -24,15 +24,29 @@ def get_all_field_names(model):
     return [f.name for f in model._meta.get_fields()]
 
 
+class EnterpriseCustomerBrandingConfigurationInline(admin.StackedInline):
+    """
+    Django admin model for EnterpriseCustomerBrandingConfiguration.
+
+    The admin interface has the ability to edit models on the same page as a parent model. These are called inlines.
+    https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#django.contrib.admin.StackedInline
+    """
+
+    model = EnterpriseCustomerBrandingConfiguration
+    can_delete = False
+
+
 @admin.register(EnterpriseCustomer)
 class EnterpriseCustomerAdmin(SimpleHistoryAdmin):
     """
     Django admin model for EnterpriseCustomer.
     """
 
-    list_display = ("name", "uuid", "active",)
+    list_display = ("name", "uuid", "active", "logo")
+
     list_filter = ("active",)
     search_fields = ("name", "uuid",)
+    inlines = [EnterpriseCustomerBrandingConfigurationInline, ]
 
     EXPORT_AS_CSV_FIELDS = ["name", "active", "uuid"]
 
@@ -42,6 +56,15 @@ class EnterpriseCustomerAdmin(SimpleHistoryAdmin):
 
     class Meta(object):
         model = EnterpriseCustomer
+
+    @staticmethod
+    def logo(instance):
+        """
+        Instance is EnterpriseCustomer.
+        """
+        if instance.branding_configuration:
+            return instance.branding_configuration.logo
+        return None
 
 
 @admin.register(EnterpriseCustomerUser)
--- a/enterprise/apps.py
+++ b/enterprise/apps.py
@@ -14,3 +14,5 @@ class EnterpriseConfig(AppConfig):
     """
 
     name = 'enterprise'
+    valid_extensions = ['.png', ]
+    image_size = 250 * 1024
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -4,13 +4,19 @@ Database models for enterprise.
 """
 from __future__ import absolute_import, unicode_literals
 
+import os
 from uuid import uuid4
 
 from simple_history.models import HistoricalRecords  # likely a bug in import order checker
+
+from django.core.files.storage import default_storage
 from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.translation import ugettext_lazy as _
+
 from model_utils.models import TimeStampedModel
+
+from .validators import validate_image_extension, validate_image_size
 
 
 class EnterpriseCustomerManager(models.Manager):
@@ -107,3 +113,58 @@ class EnterpriseCustomerUser(TimeStampedModel):
         Return uniquely identifying string representation.
         """
         return self.__str__()
+
+
+def logo_path(instance, filename):
+    """
+    Delete the file if it already exist and returns the enterprise customer logo image path.
+
+    :param instance: EnterpriseCustomerBrandingConfiguration object
+    :param filename: file to upload
+    :return path: path of image file e.g. enterprise/branding/<model.id>/<model_id>_logo.<ext>.lower()
+    """
+    extension = os.path.splitext(filename)[1].lower()
+    instance_id = str(instance.id)
+    fullname = os.path.join('enterprise/branding/', instance_id, instance_id + '_logo' + extension)
+    if default_storage.exists(fullname):
+        default_storage.delete(fullname)
+    return fullname
+
+
+class EnterpriseCustomerBrandingConfiguration(TimeStampedModel):
+    """
+    Model that keeps track of enterprise branding configurations e.g. enterprise customer logo.
+
+    Fields:
+        enterprise_customer (ForeignKey[EnterpriseCustomer]): enterprise customer
+        logo (ImageField): enterprise customer image
+    """
+
+    enterprise_customer = models.OneToOneField(
+        EnterpriseCustomer,
+        blank=False,
+        null=False,
+        related_name='branding_configuration'
+    )
+    logo = models.ImageField(
+        upload_to=logo_path,
+        help_text=_(u'Please add only .PNG files for logo images.'),
+        null=True, blank=True, max_length=255,
+        validators=[validate_image_extension, validate_image_size]
+    )
+
+    class Meta:
+        """Meta class for this Django model."""
+
+        verbose_name = _('Branding Configuration')
+        verbose_name_plural = _('Branding Configurations')
+
+    def save(self, *args, **kwargs):
+        """Save the enterprise customer branding config."""
+        if self.pk is None:
+            logo_image = self.logo
+            self.logo = None
+            super(EnterpriseCustomerBrandingConfiguration, self).save(*args, **kwargs)
+            self.logo = logo_image
+
+        super(EnterpriseCustomerBrandingConfiguration, self).save(*args, **kwargs)
--- a/enterprise/validators.py
+++ b/enterprise/validators.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+"""
+Database models field validators.
+"""
+from __future__ import absolute_import, unicode_literals
+
+import os
+
+from django.apps import apps
+from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext_lazy as _
+
+
+def get_app_config():
+    """
+    :return Application configuration.
+    """
+    return apps.get_app_config('enterprise')
+
+
+def validate_image_extension(value):
+    """
+    Validate that a particular image extension.
+    """
+    config = get_app_config()
+    ext = os.path.splitext(value.name)[1]
+    if config and not ext.lower() in getattr(config, 'valid_extensions', []):
+        raise ValidationError(_('Unsupported file extension.'))
+
+
+def validate_image_size(image):
+    """
+    Validate that a particular image size.
+    """
+    config = get_app_config()
+    if config and not image.size < getattr(config, 'image_size', 0):
+        raise ValidationError(_("The logo image file size must be less than 250KB."))
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -3,15 +3,20 @@
 Tests for the `edx-enterprise` models module.
 """
 
-from __future__ import absolute_import, unicode_literals
+from __future__ import absolute_import, unicode_literals, with_statement
 
 import unittest
 
 import ddt
+import mock
 from faker import Factory as FakerFactory
 from pytest import mark
 
-from enterprise.models import EnterpriseCustomer
+from django.core.exceptions import ValidationError
+from django.core.files import File
+from django.core.files.storage import Storage
+
+from enterprise.models import EnterpriseCustomer, EnterpriseCustomerBrandingConfiguration, logo_path
 from test_utils.factories import EnterpriseCustomerFactory, EnterpriseCustomerUserFactory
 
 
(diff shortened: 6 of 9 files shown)


enterprise/admin.py
```diff
@@ -43,6 +53,11 @@ class EnterpriseCustomerAdmin(SimpleHistoryAdmin):
     class Meta(object):
         model = EnterpriseCustomer
 
+    @staticmethod
+    def logo(instance):
+        # instance is Enterprise customer
+        return instance.branding_configuration.logo
```
e-kolpakov: @asadiqbal08 Unless I'm missing something this should crash the view when no branding configuration is created for the EC. Should it include `instance.branding_configuration is not None` check?

enterprise/models.py
```diff
@@ -107,3 +110,56 @@ def __repr__(self):
...
+
+    def save(self, *args, **kwargs):
+        """save the enterprise customer branding config """
+        if self.pk is None:
+            logo_image = self.logo
+            self.logo = None
+            super(EnterpriseCustomerBrandingConfiguration, self).save(*args, **kwargs)
+            self.logo = logo_image
+
+        super(EnterpriseCustomerBrandingConfiguration, self).save(*args, **kwargs)
```
e-kolpakov: @asadiqbal08 nit: now newline at the end of file

enterprise/models.py
```diff
@@ -107,3 +110,56 @@ def __repr__(self):
...
+        verbose_name = _('Branding Configuration')
+        verbose_name_plural = _('Branding Configurations')
+
+    def save(self, *args, **kwargs):
+        """save the enterprise customer branding config """
+        if self.pk is None:
+            logo_image = self.logo
+            self.logo = None
+            super(EnterpriseCustomerBrandingConfiguration, self).save(*args, **kwargs)
+            self.logo = logo_image
```
e-kolpakov: @asadiqbal08 Why do you need to unset logo before calling base save (genuine question)?

asadiqbal08: Including `id` when making the directory structure e.g. `enterprise_logo/<model.id>/filename` The `instance.id` in `logo_path` method returns 'None' .

enterprise/models.py
```diff
@@ -107,3 +110,56 @@ def __repr__(self):
...
+    """
+
+    enterprise_customer = models.OneToOneField(
+        EnterpriseCustomer,
+        blank=False,
+        null=False,
+        related_name='branding_configuration'
+    )
+    logo = models.ImageField(
+        upload_to=logo_path,
```
bradenmacdonald: Aren't we planning to sync these `EnterpriseCustomer*` models in the `enterprise` app among different IDAs, such as Otto and the LMS? If so, syncing image files may be a pain. It may make more sense to upload the logos to something like S3, then just store a URL on the model. Or, if only the "source of truth" holds the EnterpriseCustomerBrandingConfiguration data, and others just access the logo via a URL to the source of truth app's image URL, that's fine, but should be explicitly modelled/explained in the code.

Also genuine, possibly related question: What's the advantage of making this a one-to-one model rather than a field on `EnterpriseCustomer`?

mattdrayer: @bradenmacdonald -- trying to understand the concern you've raised a bit more -- are you suggesting that we include a `logo_url` field along with the actual `logo` ImageField, or eliminating the ImageField altogether and only having logo_url -- handling the image upload in the implementation only?

bradenmacdonald: I was suggesting the latter, but I didn't really have a specific solution in mind, since I don't really yet understand how we're thinking this `enterprise` app will work when used in multiple IDAs.

My concern is this: if we use this model in, say, the LMS, and store a logo for some `EnterpriseCustomer`, then we also use the `enterprise` app in some other IDA and sync the data from the LMS to that IDA, then that IDA will read the `logo` field as containing a path value like `/uploads/foo/bar.jpg`, but that path won't resolve to anything on the IDA, and attempting to embed the image in HTML will cause a broken image / 404.

asadiqbal08: Sorry, I could not get the point. Storing an image depends upon the value of `DEFAULT_FILE_STORAGE` in system [LN#503-505](https://github.com/edx/edx-platform/blob/master/lms/envs/aws.py#L503). As per my understanding, prod uploading the media / static assets on **AWS S3** and locally uses the `FileSystemStorage`. However if you are talking about the bucket_name for media contents then let me know is it ? then what will be the credentials to create a new bucket on S3 ?

mattdrayer: @asadiqbal08, I think @bradenmacdonald is suggesting that the `enterprise` app should own its images and host them at a location -- such as S3 -- that can be easily-accessed by multiple services/IDAs.  This way both LMS and Otto could install the `enterprise` app and refer to the images.  I'm actually wondering if part of the appconfig should include the bucket root, which would be included as part of the information sync'ed from the source-of-truth to whichever IDA is using the package?

asadiqbal08: Okay @mattdrayer. As, currently Edx-Enterprise is utilizing the storage mechanism of edx-platform such as `FileSystemStorage` as implemented in common.py and on production, edx-platform uploading the assets on AWS S3 as commented above.

If you are saying that enterprise application should have its own storage settings under [edx-enterprise repo](https://github.com/edx/edx-enterprise) then I want to know how can we **restrict/configure** the enterprise application to store its media data on its own S3 bucket instead of using `FileSystemStorage` of edx-platform ?

asadiqbal08: @bradenmacdonald @mattdrayer what you suggest. **IF** 
The S3 owner create a bucket for enterprise media assets with public read access (access control list), It seems me that any files uploaded through FileField or ImageField attributes on your models should end up in `http://your-bucket-name.s3.amazonaws.com/media/`

Note: AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY is required to access the AWS s3 service.

enterprise/admin.py
```diff
@@ -43,6 +54,13 @@ class EnterpriseCustomerAdmin(SimpleHistoryAdmin):
     class Meta(object):
         model = EnterpriseCustomer
 
+    @staticmethod
+    def logo(instance):
+        """
+        Instance is Enterprise customer.
```
mattdrayer: "EnterpriseCustomer"

enterprise/models.py
```diff
@@ -6,11 +6,14 @@
...
 from uuid import uuid4
 
-from simple_history.models import HistoricalRecords  # likely a bug in import order checker
+import os
+from django.conf import settings
 from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.translation import ugettext_lazy as _
 from model_utils.models import TimeStampedModel
+from simple_history.models import HistoricalRecords  # likely a bug in import order checker
```
mattdrayer: We'll need to keep an eye on this -- for some reason @e-kolpakov had to keep this import statement at the top of the list.

e-kolpakov: @mattdrayer @asadiqbal08 I just let it be where `isort` placed it. In case you didn't know that, `isort` without `--check-only` modifies files it checks to put imports in the order it sees as correct.

enterprise/admin.py
```diff
@@ -43,6 +54,13 @@ class EnterpriseCustomerAdmin(SimpleHistoryAdmin):
...
+        """
+        return instance.branding_configuration.logo
```
e-kolpakov: @asadiqbal08 This line is a bit fragile - it will only work if some external code ensures branding configuration is always created for enterprise customer. Even if such mechanism exists now, new ways of creating EC's might arise in future.

I would suggest something like 

    if instance.branding_configuration:
    return None

enterprise/models.py
```diff
@@ -107,3 +110,57 @@ def __repr__(self):
...
+    """
+    Delete the file if it already exist and returns the enterprise customer logo image path.
+
+    :param instance: EnterpriseCustomerBrandingConfiguration object
+    :param filename: file to upload
+    :return path: path of image file e.g. enterprise_logo/1/filename
+    """
+    name = os.path.join('enterprise_logo', str(instance.id), filename)
+    fullname = os.path.join(settings.MEDIA_ROOT, name)
+    if os.path.exists(fullname):
```
e-kolpakov: @asadiqbal08 I wonder how would it work with S3 - I don't think `os` module is smart enough to handle that. Probably you should [storage backend object](https://docs.djangoproject.com/en/1.10/topics/files/#storage-objects) instead

enterprise/models.py
```diff
@@ -107,3 +110,57 @@ def __repr__(self):
...
+    enterprise_customer = models.OneToOneField(
+        EnterpriseCustomer,
+        blank=False,
+        null=False,
+        related_name='branding_configuration'
+    )
+    logo = models.ImageField(
+        upload_to=logo_path,
+        help_text=_(u'Please add only .PNG files for logo images.'),
+        null=True, blank=True, max_length=255,
```
e-kolpakov: @asadiqbal08 to me 255 chars seems insufficient :) But I've just had an argument with Clinton about this, so I don't want to push my vision.

enterprise/validators.py
```diff
@@ -0,0 +1,24 @@
...
+from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext_lazy as _
+
+
+def validate_image_extension(value):
+    """
+    Validate that a particular image extension.
+    """
+    ext = os.path.splitext(value.name)[1]  # [0] returns path+filename
+    valid_extensions = ['.png', ]
```
e-kolpakov: @asadiqbal08 nit: extract allowed extensions into constant?

Bonus-level 1: make constant an overridable setting (and put it into app config in app.py)
Bonus-level 2: using ninja skills, make `logo` field help text pick a list of allowed extensions from that setting.

Bonus levels are **really** optional.

enterprise/validators.py
```diff
@@ -0,0 +1,24 @@
...
+    if not ext.lower() in valid_extensions:
+        raise ValidationError(u'Unsupported file extension.')
+
+
+def validate_image_size(image):
+    """
+    Validate that a particular image size.
+    """
+    if not image.size < (250 * 1024):
```
e-kolpakov: @asadiqbal08 nit: extract size into setting in `app.py`?

Bonus-level 1: make constant an overridable setting (and put it into app config in app.py)
Bonus-level 2: using ninja skills, make `logo` field help text use that setting.

Bonus levels are **really** optional.

enterprise/validators.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def validate_image_extension(value):
+    """
+    Validate that a particular image extension.
+    """
+    ext = os.path.splitext(value.name)[1]  # [0] returns path+filename
+    valid_extensions = ['.png', ]
+    if not ext.lower() in valid_extensions:
+        raise ValidationError(u'Unsupported file extension.')
```
e-kolpakov: @asadiqbal08 nit: `u'Unsupported file extension.'` should be marked for translation.

Also: `u` should not be necessary when `unicode_literals` import is added.

enterprise/validators.py
```diff
@@ -0,0 +1,24 @@
+"""
+Database models field validators.
+"""
+import os
```
e-kolpakov: @asadiqbal08 I believe quality checker should yell at you to include `from __future__ import absolute_import, unicode_literals`.

enterprise/validators.py
```diff
@@ -0,0 +1,24 @@
...
+    if not ext.lower() in valid_extensions:
+        raise ValidationError(u'Unsupported file extension.')
+
+
+def validate_image_size(image):
+    """
+    Validate that a particular image size.
+    """
+    if not image.size < (250 * 1024):
+        raise ValidationError(_(u"The logo image file size must be less than 250KB."))
```
e-kolpakov: @asadiqbal08 nit: `u` should not be necessary when `unicode_literals` are imported.

enterprise/validators.py
```diff
@@ -0,0 +1,24 @@
+"""
```
e-kolpakov: @asadiqbal08 please add encoding declaration

    # -*- coding: utf-8 -*-

enterprise/validators.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    Validate that a particular image extension.
+    """
+    ext = os.path.splitext(value.name)[1]  # [0] returns path+filename
+    valid_extensions = ['.png', ]
+    if not ext.lower() in valid_extensions:
+        raise ValidationError(u'Unsupported file extension.')
+
+
+def validate_image_size(image):
```
e-kolpakov: @asadiqbal08 Please add unittests for this method (at least three: lower than limit, higher than limit, exactly at limit). Hint: `ddt` might be helpful.

enterprise/validators.py
```diff
@@ -0,0 +1,24 @@
+"""
+Database models field validators.
+"""
+import os
+from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext_lazy as _
+
+
+def validate_image_extension(value):
```
e-kolpakov: @asadiqbal08 please add unittests for this method (at least three: `png`, any other extension, no extension).

enterprise/admin.py
asadiqbal08: @e-kolpakov Thanks. I had noted it . In my next change I will update it.

enterprise/models.py
e-kolpakov: @asadiqbal08 yes, provided it works :)

asadiqbal08:
@e-kolpakov Thanks You! I have updated the PR as per your suggestions. 
One thing is in queue that is -- **`Add relevant information to README and/or docs`** 
Can you tell me what kind of information(s) is required in this file as per aspect to the PR change set ?

CHANGELOG.rst
```diff
@@ -22,3 +22,11 @@ _____
 
 * First release on PyPI.
 * Models and Django admin integration
+
+[0.1.1] - 2016-11-02
```
e-kolpakov: @asadiqbal08 I believe the most latest version should be on top

enterprise/actions.py
```diff
@@ -7,6 +7,7 @@
 
 import unicodecsv
 from six import string_types
+
```
e-kolpakov: @asadiqbal08 nit: might be good to remove this newline - the file is not otherwise affected

enterprise/admin.py
```diff
@@ -2,13 +2,15 @@
 """
 Django admin integration for enterprise app.
 """
+
```
e-kolpakov: @asadiqbal08 nit: redundant newline

enterprise/admin.py
```diff
@@ -2,13 +2,15 @@
...
 from __future__ import absolute_import, unicode_literals
 
 from simple_history.admin import SimpleHistoryAdmin  # likely a bug in import order checker
+
```
e-kolpakov: @asadiqbal08 nit: redundant newline

tests/test_models.py
```diff
@@ -89,3 +94,110 @@ def test_string_conversion(self, method):
...
+@mark.django_db
+@ddt.ddt
+class TestEnterpriseCustomerBrandingConfiguration(unittest.TestCase):
+    """
+    Tests of the EnterpriseCustomerBrandingConfiguration model.
+    """
+
+    def get_enterprise_customer(self):
+        """
+        @:return EnterpriseCustomer object.
```
e-kolpakov: @asadiqbal08 why `@:`? I'm not an expert, but I can't remember any restructured text directive like that.

tests/test_models.py
```diff
@@ -89,3 +94,110 @@ def test_string_conversion(self, method):
...
+        """
+        faker = FakerFactory.create()
+        customer_uuid = faker.uuid4()
+        return EnterpriseCustomerFactory(uuid=customer_uuid, name="QWERTY")
```
e-kolpakov: @asadiqbal08 seems unnecessary, uuid will be created in exact same way if not passed, and I don't think "QWERTY" is used anywhere else.

tests/test_models.py
```diff
@@ -89,3 +94,110 @@ def test_string_conversion(self, method):
...
+        file_mock.name = 'test1.png'
+        file_mock.size = image_size
+        branding_configuration = EnterpriseCustomerBrandingConfiguration(
+            enterprise_customer=self.get_enterprise_customer(),
+            logo=file_mock
+        )
+        try:
+            branding_configuration.full_clean()
+        except ValidationError:
+            raised = True
```
e-kolpakov: @asadiqbal08 nit: this block might be better written as something like

    if not valid_image:
        with self.assertRaises(ValidationError):
            branding_configuration.full_clean()
    else:
            branding_configuration.full_clean()  # exception here will fail the test

But this is a question of personal taste.

tests/test_models.py
```diff
@@ -89,3 +94,110 @@ def test_string_conversion(self, method):
...
+        file_mock.size = 240 * 1024
+        raised = False
+        branding_configuration = EnterpriseCustomerBrandingConfiguration(
+            enterprise_customer=self.get_enterprise_customer(),
+            logo=file_mock
+        )
+        try:
+            branding_configuration.full_clean()
+        except ValidationError:
+            raised = True
```
e-kolpakov: @asadiqbal08 same suggestion as above.

tests/test_models.py
```diff
@@ -89,3 +94,110 @@ def test_string_conversion(self, method):
...
+        if valid_image:
+            self.assertEqual(raised, False)
+        else:
+            self.assertEqual(raised, True)
```
e-kolpakov: @asadiqbal08 if you decided against the suggestion above, this `if..else` could be simplified to `self.assertEqual(raised, not valid_image)`

e-kolpakov:
@asadiqbal08 as of README and docs - use common sense :) It might be ok to skip updating the docs this early in the development, but the more we delay it the harder it will be to add the docs all at once.

e-kolpakov:
@asadiqbal08 Also, make sure docs are at least building (make docs) and check that module documentation generated from docstrings are adequate (i.e. there are no broken links, messy formatting etc.)

asadiqbal08:
@e-kolpakov Yes, I was thinking the same about document changes.

asadiqbal08:
@mattdrayer  @e-kolpakov Done. any thumbs up ?

requirements/test.txt
```diff
@@ -25,3 +25,4 @@ pytest==3.0.3             # via pytest-catchlog, pytest-cov, pytest-django
 python-dateutil==2.5.3    # via fake-factory, freezegun
 six==1.10.0               # via django-extensions, fake-factory, freezegun, mock, python-dateutil
 unicodecsv==0.14.1
+mock==2.0.0
```
e-kolpakov: @asadiqbal08 sorry for not noticing it earlier, but there's already `mock==2.0.0` in this file on line 17 - I believe you should remove this line.

e-kolpakov:
@asadiqbal08 note I've added another suggestion about improving unit tests.

mattdrayer:
:+1: -- let's go ahead and merge -- we can address any issues in a follow-on pull request

e-kolpakov:
@asadiqbal08 minor suggestion: it would be better to check the checklist items in the description - (1) to keep track of what's done and what's not and (2) github displays the checklist counts when someone references the PR somewhere else (other PR, issue etc.).

enterprise/apps.py
```diff
@@ -14,3 +14,5 @@ class EnterpriseConfig(AppConfig):
     """
 
     name = 'enterprise'
+    valid_extensions = ['.png', ]
+    image_size = 250 * 1024
```
asadiqbal08: @mattdrayer We need to finalize the size of enterprise customer logo image size. I just used it for the time being but it is not the ideal size. anyone else to tag ?

mattdrayer: we can ask @edx/edx-ux-team for help on this

asadiqbal08: @marcotuts ^^

bradenmacdonald: Plug: Why don't we use SVG? Better retina support, etc...

PR #6, opened 2016-11-02
ENT-47: Link the EnterpriseCustomer model to the Site model

saleem-latif:
Hi @mattdrayer , @e-kolpakov , @asadiqbal08 

Kindly take a look at this PR, This PR adds an association between `Site` and `EnterpriseCustomer` model.

__Testing instructions:__

1.  This PR adds only a `ForeignKey` relationship to `EnterpriseCustomer` with updates to django admin interface - smoke testing Django admin using Enterprise Customer should be enough.
    - Open "Enterprise Customer" page in Django admin
    - Try adding a new `Enterprise Customer`, try adding an "Enterprise Customer" with and without selecting a site. ("Enterprise Customer" addition without a site will error out)

2. Django admin features to export list of Enterprise Customers to CSV is updated and CSV should contain site info.
    - Open "Enterprise Customer" page in Django admin
    - Select all/some Enterprise Customers.
    - In "Actions" dropdown select "Export to CSV", than press "Go" button. Expected result: selected Enterprise Customers are exported to CSV
    - Make sure that exported CSV now has "site" column with site domain as value.


e-kolpakov:
@saleem-latif Could you please provide testing instructions similar to #1? 
@asadiqbal08 ^^ Applies to #5 as well.

saleem-latif:
@e-kolpakov I have updated the description, take a look now.

enterprise/models.py
```diff
@@ -61,6 +62,8 @@ class Meta(object):
     active = models.BooleanField(default=True)
     history = HistoricalRecords()
 
+    site = models.ForeignKey(Site, related_name="enterprise_customers")
```
mattdrayer: Every EnterpriseCustomer should be linked to a Site

mattdrayer:
I'm 👍  on this PR assuming we've confirmed that every EC will be linked to a Site

enterprise/models.py
asadiqbal08: @e-kolpakov `NULL, in the database, is default to False. Currently I cannot create a customer without specifying the `site`. 

`

e-kolpakov: @asadiqbal08 than probably the [migration is wrong](https://github.com/edx/edx-enterprise/pull/6/files#diff-6695276e9b640f365e80b4987fcd7810R26)

tests/test_models.py
```diff
@@ -66,6 +68,20 @@ def test_string_conversion(self, method):
...
+        """
+        faker = FakerFactory.create()
+        customer_uuid = faker.uuid4()
+        EnterpriseCustomerFactory(uuid=customer_uuid)
+
+        customer = EnterpriseCustomer.objects.get(uuid=customer_uuid)  # pylint: disable=no-member
+        site = Site.objects.get(domain="example.com")
+
+        self.assertEqual(customer.site, site)
+        self.assertListEqual([customer], list(site.enterprise_customers.all()))
```
asadiqbal08: @e-kolpakov jumping here, did you mean to remove the test `test_site_association` here ?

e-kolpakov: @asadiqbal08 yes.

e-kolpakov:
@saleem-latif @asadiqbal08 Looks like you're working on this together, so I'll tag you both on this. Please update version in `enterprise/__init__.py`, add relevant changes to changelog (and README and docs, if applicable).

asadiqbal08:
@e-kolpakov I have updated the code. 

#### NOTE: 
I was getting the migration conflicts after rebasing with master as [PR#5](https://github.com/edx/edx-enterprise/pull/5) also contained a migration file with same name  **---** `0002_*` and both are pointing to same model. Now I am keeping the one migration file (take a look). If you have any local migrations then you need to run it again. ( might be, firstly you need to move your local migration back to `0001_*` then re run the migrations )

enterprise/migrations/0002_auto_20161104_0417.py
```diff
@@ -24,11 +25,26 @@ class Migration(migrations.Migration):
...
             ],
             options={
-                'verbose_name': 'Enterprise Customer Branding',
-                'verbose_name_plural': 'Enterprise Customer Brandings',
+                'verbose_name': 'Branding Configuration',
+                'verbose_name_plural': 'Branding Configurations',
             },
         ),
+        migrations.AddField(
+            model_name='enterprisecustomer',
```
e-kolpakov: @asadiqbal08 That's not the right thing to do - you shouldn't merge the two migrations into one. Instead, the changes here should be added as migration 0003. Please reset this file into it's original state and create new migration.

asadiqbal08:
@e-kolpakov take a look

asadiqbal08:
@mattdrayer fyi here

e-kolpakov:
@asadiqbal08 I thought squashing commits is something obvious. Or at least giving commits sensible messages. Now we've got one commit that says "ENT-47 changes", which is not very informative :)

asadiqbal08:
ah.. I forgot to squash but usually we do it and give proper messages to commits :)

PR #7, opened 2016-11-11
[ENT-49] Add catalog field that searches the API for catalog names

haikuginger:
This pull request adds a catalog field to the EnterpriseCustomer, along with a form used in the admin view to dynamically grab the names of the available catalogs from a linked Course Discovery instance.

**JIRA tickets**: Implements ENT-49.

**Discussions**: See ENT-49

**Dependencies**: None

**Screenshots**:
screen shot 2016-11-10 at 7 49 15 pm

**Sandbox URL**: TBD

**Testing instructions**:

1. Clone edx/course-discovery.
2. Configure EdX OpenID Connect [per the instructions on this page](https://open-edx-course-catalog.readthedocs.io/en/latest/getting_started.html).
5. Set `COURSE_CATALOG_API_URL` in your devstack LMS env to use your host's local IP address.
2. Run `make open-devstack`
4. Inside the course discovery devstack:

        ./manage.py create_or_update_partner --code='abc' --name='ABC Partner' --oidc-url-root=http://HOST_IP:8000/oauth2 --oidc-key='OIDC_KEY' --oidc-secret='OIDC_SECRET' --courses-api-url=http://HOST_IP:8000/api/courses/v1/

        ./manage.py refresh_course_metadata

        ./manage.py createsuperuser #Create the user course_discovery_worker

5. Once you've run the above commands, close out of the docker guest, and run `make start-devstack`.
6. Navigate to http://localhost:18381/admin, log in with `course_discovery_worker` and set up a couple course catalogs. You don't need to test the queries.
2. Install this fork of edx-enterprise into your devstack, adding it as an enabled application.
1. In your devstack, run `./manage.py lms migrate --settings=devstack`.
1. In your devstack run `./manage.py lms createsuperuser`. Add `user course_discovery_worker`.
3. Go to the EnterpriseCustomer admin panel, using the `course_discovery_worker` user, create a new EnterpriseCustomer, and observe that the list of available catalogs matches the catalogs listed in your course discovery stack.

1. Remove catalog option needs to be added
2. Tests need to be created 

**Reviewers**
- [x] @e-kolpakov 
- [x] @asadiqbal08 
- [ ] @saleem-latif 

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag


enterprise/forms.py
```diff
@@ -0,0 +1,41 @@
...
+            'catalog',
+        )
+
+    def __init__(self, *args, **kwargs):
+        """
+        Save the user that the form was constructed for and initalize
+        using the parent class methods
+        """
+        super(EnterpriseCustomerAdminForm, self).__init__(*args, **kwargs)
+        self.fields['catalog'] = forms.ChoiceField(choices=self.get_catalog_options())
```
e-kolpakov: @haikuginger nit/optional suggestion: Just to make sure you haven't missed that approach - you could override field widget using `Meta.widgets` attribute. But probably you can't avoid using `self` here or down the stream, so if it's the case I'm fine with this way of doing it.

enterprise/models.py
```diff
@@ -65,6 +65,7 @@ class Meta(object):
 
     uuid = models.UUIDField(primary_key=True, default=uuid4, editable=False)
     name = models.CharField(max_length=255, blank=False, null=False, help_text=_("Enterprise Customer name."))
+    catalog = models.PositiveIntegerField(null=True, help_text=_("Course catalog for the Enterprise Customer."))
```
e-kolpakov: @haikuginger nit/optional suggestion: it would be a bit more human-friendly if `catalog` field stored some kind of catalogue name, not ID. Not sure if it is possible with catalogue service though (i.e. if catalogue service has string keys).

enterprise/utils.py
```diff
@@ -0,0 +1,60 @@
+# -*- coding: utf-8 -*-
+"""
+Utilities to be used across the enterprise djangoapp
```
e-kolpakov: @haikuginger suggestion - this `utils` file only contains parts relevant to course catalogue so far. I would suggest renaming it to `course_catalogue_api` and keeping it dedicated to course catalogue.

Also, I'm placing this comment into a docstring, because it'll change in the process, so outdated comment will be properly hidden by github (thank you for attending my 20seconds of self-praising :))

e-kolpakov:
@haikuginger One more thing

> Set up a course discovery instance using Docker

Isn't that helpful - could you please at least add a link on how to do that?

enterprise/admin.py
```diff
@@ -9,6 +9,7 @@
 
 from enterprise.actions import export_as_csv_action
 from enterprise.models import EnterpriseCustomer, EnterpriseCustomerBrandingConfiguration, EnterpriseCustomerUser
+from enterprise.forms import EnterpriseCustomerAdminForm
```
zubair-arbi: Please arrange imports in alphabetical order.

enterprise/admin.py
```diff
@@ -54,9 +55,17 @@ class EnterpriseCustomerAdmin(SimpleHistoryAdmin):
...
+
     class Meta(object):
         model = EnterpriseCustomer
 
+    def get_form(self, request, obj=None, **kwargs):
+        form = super(EnterpriseCustomerAdmin, self).get_form(request, obj, **kwargs)
+        form.user = request.user
+        return form
+
+
```
zubair-arbi: remove this extra newline

enterprise/admin.py
```diff
@@ -54,9 +55,17 @@ class EnterpriseCustomerAdmin(SimpleHistoryAdmin):
         export_as_csv_action("CSV Export", fields=EXPORT_AS_CSV_FIELDS)
     ]
 
+    form = EnterpriseCustomerAdminForm
+
     class Meta(object):
         model = EnterpriseCustomer
 
+    def get_form(self, request, obj=None, **kwargs):
+        form = super(EnterpriseCustomerAdmin, self).get_form(request, obj, **kwargs)
```
asadiqbal08: any doc string ?

enterprise/admin.py
```diff
@@ -54,9 +55,17 @@ class EnterpriseCustomerAdmin(SimpleHistoryAdmin):
         export_as_csv_action("CSV Export", fields=EXPORT_AS_CSV_FIELDS)
     ]
 
+    form = EnterpriseCustomerAdminForm
+
     class Meta(object):
         model = EnterpriseCustomer
 
+    def get_form(self, request, obj=None, **kwargs):
```
asadiqbal08: any doc string ?

enterprise/forms.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
+"""
+Forms to be used in the enterprise djangoapp
+"""
+from django import forms
+from enterprise.models import EnterpriseCustomer
+from enterprise.utils import get_all_catalogs
+
```
zubair-arbi: for consistency use 2 newlines after imports

enterprise/migrations/0004_auto_20161109_1511.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.2 on 2016-11-09 15:11
```
zubair-arbi: this is redundant, please remove it.

zubair-arbi: Please ignore above comment as this is auto generated migration by django.

enterprise/utils.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+try:
+    from openedx.core.djangoapps.api_admin.utils import course_discovery_api_client
+except ImportError:
+    course_discovery_api_client = None
+
```
zubair-arbi: Please add 2 newlines after imports section

enterprise/utils.py
```diff
@@ -0,0 +1,60 @@
...
+
+def get_all_catalogs(user):
+    """
+    Return a list of all course catalogs, including name and ID
+    """
+    # Initialize the permanent list to hold all results, as well as
+    # the ephemeral list to store each API call's results.
+    results = []
+    this_result = []
```
asadiqbal08: what is the difference b/w `results` and `this_result` ? , we should use some explanatory names.

enterprise/utils.py
```diff
@@ -0,0 +1,60 @@
...
+    # Set the initial offset and limit used in API calls
+    offset = 0
+    limit = 50
```
asadiqbal08: what is limit indicating here ? 50 ??

asadiqbal08: why `50 ` is  hard-codes here ?

enterprise/utils.py
```diff
@@ -0,0 +1,60 @@
...
+    # Initialize an API client
```
zubair-arbi: this is not providing any extra information so I suggest removing this comment.

enterprise/forms.py
```diff
@@ -0,0 +1,41 @@
...
+        fields = (
+            'name',
+            'site',
+            'active',
+            'catalog',
+        )
+
+    def __init__(self, *args, **kwargs):
+        """
+        Save the user that the form was constructed for and initalize
```
asadiqbal08: nit: ... and initialize `it` using the parent class methods.

enterprise/utils.py
```diff
@@ -0,0 +1,60 @@
...
+        return course_discovery_api_client(user)
+
+
+def get_all_catalogs(user):
+    """
+    Return a list of all course catalogs, including name and ID
+    """
+    # Initialize the permanent list to hold all results, as well as
+    # the ephemeral list to store each API call's results.
+    results = []
```
zubair-arbi: Can we change this variable to something like `user_course_catalogs` ?

enterprise/utils.py
```diff
@@ -0,0 +1,60 @@
...
+
+def get_catalog_api_client(user):
+    """
+    Retrieve an authenticated API client that can be used to access the course
+    discovery API; raise an exception to be caught at a higher level if the
+    package doesn't have OpenEdX resources available.
+    """
+    if course_discovery_api_client is None:
+        raise NotConnectedToOpenEdX(
+            'To get a catalog API client, this package must be installed in an OpenEdX environment.'
```
asadiqbal08: Will this message be displayed on UI ? Is it not required to mark for translation ?

enterprise/utils.py
```diff
@@ -0,0 +1,60 @@
+# -*- coding: utf-8 -*-
+"""
+Utilities to be used across the enterprise djangoapp
+"""
+try:
+    from openedx.core.djangoapps.api_admin.utils import course_discovery_api_client
```
zubair-arbi: What benefit we would get by catching this exception? Shouldn't we just let this import error bubble up so that we know early that we depend on the catalog service for EC's?
Also if we want to disable the EC app then there should be a proper way of doing this like some feature flag or switch.

enterprise/utils.py
```diff
@@ -0,0 +1,60 @@
...
+    # the ephemeral list to store each API call's results.
+    results = []
+    this_result = []
+    # Set the initial offset and limit used in API calls
+    offset = 0
+    limit = 50
+    # Initialize an API client
+    client = get_catalog_api_client(user)
+    # Loop while either we're at the beginning or we're still getting results
+    while offset == 0 or len(this_result):
```
zubair-arbi: If we are ultimately using client `EdxRestApiClient` and our goal is to get all paginated results then this is already handled here: https://github.com/edx/edx-platform/blob/master/openedx/core/lib/edx_api_utils.py#L92
So you can use method `get_edx_api_data` to get complete data from all pages. Example using can be found here: https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/catalog/utils.py#L50-L58

enterprise/admin.py
e-kolpakov: @zubair-arbi I would expect code quality check to fail because of this, so no need to explicitly mention this in the review.

enterprise/admin.py
e-kolpakov: @zubair-arbi same here - pylint should check for excess newlines

enterprise/admin.py
e-kolpakov: @asadiqbal08 and this should be covered by quality check too.

enterprise/forms.py
e-kolpakov: @zubair-arbi same code quality comment. Also, I believe it's dictated by `pep8`, not consistency, :)

enterprise/utils.py
e-kolpakov: @zubair-arbi same as above

enterprise/utils.py
e-kolpakov: @haikuginger +1 - please mark for translation

enterprise/utils.py
haikuginger: There may be cases (for example, during testing) where we won't be in an EdX environment, but that's OK. Raising the exception when we go to create the API client, rather than when the import fails, allows us to load the package properly, and then mock out the appropriate calls for testing.

enterprise/forms.py
haikuginger: Will fix.

enterprise/utils.py
haikuginger: This and the comment below should be handled with @zubair-arbi's comment about the ability to automatically handle paging.

enterprise/utils.py
haikuginger: I'll make that change.

enterprise/utils.py
haikuginger: Yes, sounds good.

enterprise/utils.py
haikuginger: 👍

enterprise/utils.py
haikuginger: All right - certainly doable. I was looking for existing patterns to follow, and a utils module seems typical.

enterprise/models.py
haikuginger: Hmm. I'll take a look and see if we can query the course discovery API with some sort of alternate string. I don't _think_ it's supported though; if not, we'll just have to leave it as-is, especially since the administrator of the course discovery API can change the name of a catalog whenever they feel like it.

enterprise/forms.py
haikuginger: Thanks for the suggestion - I'll take a second look at that!

haikuginger: Ah, we can't do it this way because we need to execute the catalog options API call in the context of a particular instance.

enterprise/models.py
haikuginger: Yeah, the course discovery API doesn't support querying catalogs by name, so we have to save the ID. I could see if I could find a way to save the name as well, but that seems a bit unnecessary given the admin UI will handle getting it for us.

haikuginger:
@e-kolpakov, I've added additional testing instructions. Working on getting unit tests going.

enterprise/course_catalog_api.py
```diff
@@ -0,0 +1,47 @@
...
+        )
+    else:
+        return course_discovery_api_client(user)
+
+
+def get_all_catalogs(user):
+    """
+    Return a list of all course catalogs, including name and ID
+    """
+    client = get_catalog_api_client(user)
```
e-kolpakov: @haikuginger suggestion: refactor these two methods into class - might be simpler to add more catalog api-related methods in future.

enterprise/actions.py
```diff
@@ -58,3 +58,18 @@ def export_as_csv(modeladmin, request, queryset):  # pylint: disable=unused-argu
 
     export_as_csv.short_description = description
     return export_as_csv
+
+
+def get_clear_catalog_id_action(description=None):
```
zubair-arbi: from filename `actions.py` its obvious that the methods in this file would perform some action so I would suggest removing `_action` from the name

zubair-arbi: In fact I would suggest changing the file name altogether to something like `custom_django_admin_actions.py`.

enterprise/admin/forms.py
```diff
@@ -0,0 +1,49 @@
...
+from django import forms
+from django.utils.translation import ugettext_lazy as _
+
+from enterprise.course_catalog_api import get_all_catalogs
+from enterprise.models import EnterpriseCustomer
+
+
+class EnterpriseCustomerAdminForm(forms.ModelForm):
+    """
+    Alternate form for the EnterpriseCustomer admin page; fetches catalog names and IDs from the Course Discovery API.
```
zubair-arbi: Please use max column length to 79 for readability in doctstrings.

enterprise/admin/forms.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+        Initialize the form, then substitute a ChoiceField in for the catalog field.
+        """
+        super(EnterpriseCustomerAdminForm, self).__init__(*args, **kwargs)
+        self.fields['catalog'] = forms.ChoiceField(choices=self.get_catalog_options())
+
+    def get_catalog_options(self):
+        """
+        Retrieve a list of catalog ID and catalog name pairs for use in the form.
+        catalog drop-down list.
```
zubair-arbi: remove this line or merge appropriately with above line

enterprise/course_catalog_api.py
```diff
@@ -0,0 +1,47 @@
...
+        )
+    else:
+        return course_discovery_api_client(user)
+
+
+def get_all_catalogs(user):
+    """
+    Return a list of all course catalogs, including name and ID.
+    """
+    client = get_catalog_api_client(user)
```
zubair-arbi: do we need to add check `if course_discovery_api_client is None:` here?

enterprise/admin/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""
+Module imports admin panels needed for Enterprise app.
+"""
+from enterprise.admin.admin import EnterpriseCustomerAdmin, EnterpriseCustomerUserAdmin
```
zubair-arbi: Are we taking advantage of declaring imports here? if not which seems to be the case here then I would suggest reverting/removing the changes for this file.

zubair-arbi: @e-kolpakov sounds good. Thanks for the explanation.

enterprise/course_catalog_api.py
haikuginger: No; this uses the local `get_catalog_api_client` function, which does that check and raises an appropriate exception.

haikuginger: On further review, though, we should do something about being able to import `CatalogIntegration` and `get_edx_api_data`.

enterprise/actions.py
haikuginger: Given that the file already existed with a given name, I'd prefer to leave it as-is in this PR unless there's a compelling reason for the rename. In addition, `actions.py` is an existing Django trope.

Finally, for the function name, I think it's important to differentiate between the outer function, whose job is to *get* the *action* function, and the inner function, whose job is to actually clear the catalog ID.

mattdrayer: Instead of saying "Return a method..." in the docstring, we could add a little more detail to clear up any confusion -- something as simple as "Return the action method..." would probably be fine.

enterprise/course_catalog_api.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+"""
+Utilities to get details from the course discovery API.
```
mattdrayer: nit: replace "discovery" with "catalog" here.

enterprise/course_catalog_api.py
```diff
@@ -0,0 +1,47 @@
...
+"""
+from django.utils.translation import ugettext_lazy as _
+
+try:
```
mattdrayer: We should try/except only the code we're concerned about, rather than all imports, since the except only deals with one use case.

enterprise/course_catalog_api.py
```diff
@@ -0,0 +1,47 @@
...
+    from openedx.core.lib.edx_api_utils import get_edx_api_data
+except ImportError:
+    course_discovery_api_client = None
+
+
+class NotConnectedToOpenEdX(Exception):
+    """
+    Exception that is raised when we try to import from the core openedx repository, but it isn't available.
+    """
+
```
mattdrayer: Remove extra line

haikuginger: Actually, pylint mandated the space.

enterprise/utils.py
bradenmacdonald: I thought that we only mark error messages for i18n if there's a good chance they'll be seen by users (otherwise it wastes translator time). In this case, is this error likely to be seen by a user using the django admin, or just a one-time error that would be seen by devops when installing this for the first time?

enterprise/course_catalog_api.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    if course_discovery_api_client is None:
+        raise NotConnectedToOpenEdX(
+            _('To get a catalog API client, this package must be installed in an OpenEdX environment.')
+        )
+    else:
+        return course_discovery_api_client(user)
+
+
+def get_all_catalogs(user):
```
asadiqbal08: Are we not fetching the all catalogs that are available ? Please explain the purpose of user with catalogs here ?

enterprise/course_catalog_api.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+    Return a list of all course catalogs, including name and ID.
```
asadiqbal08: It is returning the all course catalogs, I could not understand the role of `user` here ...

enterprise/actions.py
```diff
@@ -58,3 +58,18 @@ def export_as_csv(modeladmin, request, queryset):  # pylint: disable=unused-argu
...
+    """
+    Return the action method to clear the catalog ID for a EnterpriseCustomer.
+    """
+    description = description or "Unlink selected objects from existing course catalogs"
```
e-kolpakov: @haikuginger mark "Unlink selected objects from existing course catalogs" for translation + don't forget to run `make extract_translations` and commit django.po / djangojs.po

enterprise/admin/__init__.py
e-kolpakov: @haikuginger warning: you're going to have a significant conflict here, as I've kept most of `admin.py` in `admin/__init__.py`. Not sure how best to address it, but I'm ok with both variants (i.e. keeping admins in __init__.py or moving them into a dedicated module and importing from __init__.py)

enterprise/admin/admin.py
```diff
@@ -51,12 +52,25 @@ class EnterpriseCustomerAdmin(SimpleHistoryAdmin):
...
         model = EnterpriseCustomer
 
+    def get_form(self, request, obj=None, **kwargs):
+        """
+        Retrieve the appropriate form to use for this admin panel.
+
+        Save the request user into the form for use in loading catalog details.
+        """
+        form = super(EnterpriseCustomerAdmin, self).get_form(request, obj, **kwargs)
+        form.user = request.user
```
e-kolpakov: @haikuginger maybe `user` should become form constructor parameter, as it won't be of much use without it anyway.

enterprise/course_catalog_api.py
e-kolpakov: @haikuginger I've got similar concern about this method fetching catalogs for one particular user. Will catalogs created by other user appear here? Can we use LMS Admin with any other user having staff access (i.e. with someone else except `course_discovery_worker`)? Not sure what's the right option here - probably you should discuss that with @mattdrayer and/or @griffresch

test_utils/factories.py
```diff
@@ -46,6 +47,7 @@ class Meta(object):
     name = factory.LazyAttribute(lambda x: FAKER.company())
     active = True
     site = factory.SubFactory(SiteFactory)
+    catalog = factory.LazyAttribute(lambda x: random.uniform(1, 1000000))
```
haikuginger: FAKER.pyint() could give us a negative int. We could do math.abs on that, but either way, we're moving away from "pure" test methods.

enterprise/course_catalog_api.py
haikuginger: The course discovery API provides granular permissions for the catalogs; additional users or groups can be added, either on a per-catalog basis, or a by-model basis. The catalog API cannot be accessed anonymously, so the best we can do is generate an API client using a token generated with the current user and the secret OAuth token.

tests/test_admin.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def patch_enterprise_customer_user():
+    """
+    Set the class-level form user to None.
+    """
+    EnterpriseCustomerAdminForm.user = None
+
+
+def unpatch_enterprise_customer_user():
```
haikuginger: Okay! Couldn't hurt.

enterprise/admin/admin.py
haikuginger: Unfortunately, the ModelAdmin `get_form` method doesn't call `get_form_kwargs` or something similar where we can supply it in the form construction - as far as I can tell, we have to patch it in.

haikuginger:
@e-kolpakov, I can't replicate the UI you're seeing. Can you verify the HTTP response that the LMS sees when it makes the request?

test_utils/factories.py
e-kolpakov: fake.random_int(min=0, max=9999) than. 
https://faker.readthedocs.io/en/latest/providers/faker.providers.address.html#faker-providers-address

e-kolpakov:
@haikuginger unfortunately not today - and I haven't dug deep enough to see if LMS sends the request to course discovery and where it sends it (if it does).

test_utils/factories.py
haikuginger: Thanks for the link, @e-kolpakov; I had looked but didn't find the method spec.

asadiqbal08:
@e-kolpakov regarding your concern about UI, Are you accessing the course_discovery and Lms with same superuser e.g. course_discovery on both side ? I know there is some permissions on course discovery api. 
you can check the response by debugging at this [POINT](https://github.com/edx/edx-platform/blob/master/openedx/core/lib/edx_api_utils.py#L76)

haikuginger:
Rebased from b224f27.

haikuginger:
@e-kolpakov, I've rebased on master and updated everything to make all the tests happy. I'll be happy to assist with your Docker/course discovery API issues whenever you get me some more details. 🙂

haikuginger:
@asadiqbal08, @mattdrayer, this is no longer WIP.

tests/admin/test_forms.py
```diff
@@ -123,29 +129,113 @@ def test_clean_existing_pending_link(self, existing_email):
...
+    @mock.patch('enterprise.admin.forms.get_all_catalogs')
+    def test_interface_displays_selected_option(self, mock_method):
+        mock_method.return_value = [
+            {
+                "id": self.catalog_id,
+                "name": "My Catalog"
+            }
+        ]
+        form = EnterpriseCustomerAdminForm()
+        self.assertTrue(isinstance(form.fields['catalog'], forms.ChoiceField))
```
e-kolpakov: @haikuginger please use [pytest assertions](http://doc.pytest.org/en/latest/assert.html):

    assert isinstance(form.fields['catalog'], forms.ChoiceField)

Here and everywhere else.

tests/admin/test_forms.py
```diff
@@ -123,29 +129,113 @@ def test_clean_existing_pending_link(self, existing_email):
...
+        self.assertEqual(
+            form.fields['catalog']._get_choices(),  # pylint: disable=protected-access
+            forms.ChoiceField(  # pylint: disable=protected-access
+                choices=((None, 'None'), (self.catalog_id, 'My Catalog'),)
+            )._get_choices()
```
e-kolpakov: @haikuginger two problems with this assertion:

1. You don't have to call private `_get_choices` - there's public `choices` property available, with `_get_choices` being a getter.
2. I would assert on the expected choices result directly, i.e. `assert form.fields['catalog'].choices, [(None, None), (something, something)]`.

tests/admin/test_forms.py
```diff
@@ -123,29 +129,113 @@ def test_clean_existing_pending_link(self, existing_email):
...
         """
         Test set up.
         """
-        super(TestEnterpriseCustomerForm, self).setUp()
+        super(TestEnterpriseCustomerAdminForm, self).setUp()
         self.idp_choices_mock = mock.Mock()
         patcher = mock.patch("enterprise.admin.forms.utils.get_idp_choices", self.idp_choices_mock)
         patcher.start()
         self.addCleanup(patcher.stop)
+        self.catalog_id = random.uniform(1, 1000000)
```
e-kolpakov: @haikuginger nit: `catalog_id` becomes float in this case. You might want to use `random.randint`, or better yet, FAKER.random_int.

tests/admin/test_forms.py
```diff
@@ -123,29 +129,113 @@ def test_clean_existing_pending_link(self, existing_email):
...
+            form.fields['catalog']._get_choices(),  # pylint: disable=protected-access
+            forms.ChoiceField(  # pylint: disable=protected-access
+                choices=((None, 'None'), (self.catalog_id, 'My Catalog'),)
+            )._get_choices()
+        )
+
+    @mock.patch('enterprise.course_catalog_api.get_catalog_api_client')
+    @mock.patch('enterprise.course_catalog_api.CatalogIntegration')
+    @mock.patch('enterprise.course_catalog_api.get_edx_api_data')
+    def test_with_mocked_get_edx_data(self, mocked_method, *args):  # pylint: disable=unused-argument
```
e-kolpakov: @haikuginger nit: it's hard t say which method are you mocking - I can't recall off top of my head if the first argument would come from first decorator or the last (probably the last). Might make sense to give the argument a better name.

tests/admin/test_forms.py
```diff
@@ -123,29 +129,113 @@ def test_clean_existing_pending_link(self, existing_email):
...
+        mocked_method.return_value = [
+            {
+                "id": self.catalog_id,
+                "name": "My Catalog"
+            }
```
e-kolpakov: @haikuginger nit: would be nice to test how will it behave if more than one item is returned

tests/admin/test_forms.py
```diff
@@ -123,29 +129,113 @@ def test_clean_existing_pending_link(self, existing_email):
...
+            )._get_choices()
+        )
+
+    @mock.patch('enterprise.course_catalog_api.CatalogIntegration')
+    @mock.patch('enterprise.course_catalog_api.get_edx_api_data')
+    def test_raise_error_missing_course_discovery_api(self, *args):  # pylint: disable=unused-argument
+        message = re.escape(
+            'To get a catalog API client, this package must be installed in an OpenEdX environment.'
+        )
+        with assertRaisesRegex(self, NotConnectedToOpenEdX, message):
```
e-kolpakov: @haikuginger suggestion: use [pytest assertions](http://doc.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions). The benefit is that you can get hold on actual exception instance and assert on the message directly.

tests/test_util_functions.py
```diff
@@ -127,7 +127,7 @@ def test_get_available_idps(self):
     @ddt.data(
         (EnterpriseCustomer, [
             "enterprisecustomeruser", "pendingenterprisecustomeruser", "branding_configuration", "created", "modified",
-            "uuid", "name", "active", "identity_provider", "site"
+            "uuid", "name", "catalog", "active", "identity_provider", "site"
```
e-kolpakov: @haikuginger suggestion: using actual model doesn't occur to me as a good idea anymore - maybe we should switch to synthetic test-only models here, instead of actual ones? Feel free to skip/ignore this comment.

e-kolpakov:
@haikuginger note I have updated the description with two checklists.

enterprise/admin/__init__.py
```diff
@@ -37,7 +37,6 @@ class EnterpriseCustomerAdmin(DjangoObjectActions, SimpleHistoryAdmin):
     Django admin model for EnterpriseCustomer.
     """
 
-    form = EnterpriseCustomerForm
     list_display = ("name", "uuid", "site", "active", "logo", "identity_provider")
```
e-kolpakov: @haikuginger Should probably add `catalog` field to `list_display`.

enterprise/admin/__init__.py
```diff
@@ -47,11 +46,26 @@ class EnterpriseCustomerAdmin(DjangoObjectActions, SimpleHistoryAdmin):
     EXPORT_AS_CSV_FIELDS = ["name", "active", "site", "uuid", "identity_provider"]
```
e-kolpakov: @haikuginger should probably add `catalog` field to `EXPORT_AS_CSV_FIELDS`.

bit: also, it might make sense to rename this const to `DISPLAY_FIELDS` and use it both for CSV export action and `list_display`

e-kolpakov:
@haikuginger @asadiqbal08 using `course_discovery_worker` user was part of solution. The other part was adding `Catalog integration` record to enable course discovery integration (with URL http://<host_ip>:18381 and any significantly large timeout - default is 3600). Maybe Catalog Integration record should have already existed if my devstack was fresh baked, but it's about half year old.

enterprise/admin/__init__.py
haikuginger: Can't - list_display has `logo`; CSV export doesn't.

haikuginger:
>I believe this is a significant problem that must be resolved somehow.

I disagree, and I think you're missing the reason why only that user is working for you. The course catalog API has a rich permissions system all its own; by default, only that user who creates a catalog may access or view it. However, permission may be granted on a per-model or per-object basis to additional users or groups. If you created catalogs only with `course_discovery_worker`, then by default those catalogs would only be visible to `course_discovery_worker`. If you created catalogs with `admin`, those catalogs would only be available to `admin`. In either case, you could add the other user to the permission set, either directly or by adding them to a group with permissions, and they would gain access to the catalogs of the other.

When deployed, new administrative users will log into the course catalog API admin via OAuth2 with the LMS, and users with existing access will be able to add them to the relevant groups.

haikuginger:
Rebased from 4b6e2b4 to resolve migration conflict.

haikuginger:
@e-kolpakov, your requests have been addressed, and we're at a green build.

e-kolpakov:
@haikuginger Ok, than I recall my objections/comments about being able to use only `course_discovery_worker`ю

tests/admin/test_actions.py
```diff
@@ -47,6 +48,60 @@ class Meta(object):
...
+
+    def test_existing_value_removed(self):
+        """
+        Test that catalog IDs are cleared appropriately.
+        """
+        for _ in range(5):
+            EnterpriseCustomerFactory()
+        query_set = EnterpriseCustomer.objects.all()  # pylint: disable=no-member
+        for customer in query_set:
+            self.assertNotEqual(customer.catalog, None)
```
e-kolpakov: @haikuginger pytest assertions please - I'm trying to standardize on using them in this repo.

If this file uses unittest assertions - it would be very nice if you could replace them with pytest assertions, but that's not a requirement to land this PR.

tests/admin/test_actions.py
```diff
@@ -47,6 +48,60 @@ class Meta(object):
...
+        self.action(self.model_admin_mock, mock.Mock(), query_set)
+        for customer in query_set:
+            self.assertEqual(customer.catalog, None)
```
e-kolpakov: @haikuginger and here too

tests/admin/test_actions.py
```diff
@@ -47,6 +48,60 @@ class Meta(object):
...
+
+    def test_leave_existing_missing_value(self):
+        """
+        Test that already-clear catalog IDs are left alone.
+        """
+        for _ in range(5):
+            EnterpriseCustomerFactory()
+        query_set = EnterpriseCustomer.objects.all()  # pylint: disable=no-member
+        for customer in query_set:
+            self.assertNotEqual(customer.catalog, None)
```
e-kolpakov: @haikuginger and here

tests/admin/test_actions.py
```diff
@@ -47,6 +48,60 @@ class Meta(object):
...
+            customer.catalog = None
+            customer.save()
+            self.assertEqual(customer.catalog, None)
```
e-kolpakov: @haikuginger and here

tests/admin/test_actions.py
```diff
@@ -47,6 +48,60 @@ class Meta(object):
...
+        self.action(self.model_admin_mock, mock.Mock(), query_set)
+        for customer in query_set:
+            self.assertEqual(customer.catalog, None)
```
e-kolpakov: @haikuginger and here

haikuginger:
@e-kolpakov, set to go.

haikuginger:
@asadiqbal08, this is ready for upstream approval.

e-kolpakov:
@haikuginger good to go! :+1: 

One thing: looks like Travis experiences problems with deploying to PyPi, so if new version does not appear on PyPi after tag-initiated build it is ok.

enterprise/conf/locale/en/LC_MESSAGES/django.po
```diff
@@ -17,28 +17,51 @@ msgstr ""
...
+#: enterprise/admin/forms.py:86
+msgid "None"
+msgstr ""
+
 #: enterprise/admin/views.py:129
-#, python-brace-format
 msgid "Email {email} is not linked to Enterprise Customer {ec_name}"
 msgstr ""
 
+#: enterprise/course_catalog_api.py:45
```
asadiqbal08: Is it good to keep the line number here ? I meant, any upcoming change in code will not change the line number for django translation ?

asadiqbal08:
one question to ask beside 👍

enterprise/conf/locale/en/LC_MESSAGES/django.po
haikuginger: This is automatically generated; any further changes will also be automatically generated. We'll just need to remember to run translations that as part of our merge process.

@e-kolpakov, thoughts on that?

mattdrayer:
WOW this PR was epic!  Way to stick with it @haikuginger 😄 Changes LGTM -- I say we merge and make any further changes as follow-on PRs.  👍

haikuginger:
Rebased and squashed from 5d378e1.

e-kolpakov:
@haikuginger @asadiqbal08 @saleem-latif Looks like we've got some inconsistency between how quality is run locally and on CI. I've just rebased on top of this PR and run `make quality`. CI build is green here, but here's what I've got:

```
************* Module enterprise.course_catalog_api
W: 30, 0: native string literal (native-string)
W: 42, 0: native string literal (native-string)
W: 45, 0: native string literal (native-string)
W: 54, 0: native string literal (native-string)
W: 58, 0: native string literal (native-string)
W: 62, 0: native string literal (native-string)
W: 67, 0: native string literal (native-string)
```

It is happened because course_catalog_api does not contain `from __future__ import absolute_import, unicode_literals` mantra. But CI runs quality on py3.5, so all those strings already unicode and does not trigger pylint alarm.

So, please run `make quality` before pushing (or at least merging) until quality is run both for py2.7 and py3.5

bradenmacdonald:
@e-kolpakov Can you change that CI setting in your next PR, so that won't happen?

enterprise/admin/forms.py
```diff
@@ -52,23 +53,37 @@ def clean_email(self):
...
         """
-        Convert SlugField to TypedChoiceField if choices can be accessed.
+        Initialize the form.
+
+        Substitute a ChoiceField in for the catalog field that would
+        normally be set up as a plain number entry field.
         """
-        super(EnterpriseCustomerForm, self).__init__(*args, **kwargs)
+        super(EnterpriseCustomerAdminForm, self).__init__(*args, **kwargs)
+        self.fields['catalog'] = forms.ChoiceField(choices=self.get_catalog_options())
```
bradenmacdonald: We allow creating an EnterpriseCustomer before a catalog exists for that customer to use, right? (The field has `null=True` so that seems to be the goal of this PR as well). If so, there was a bug in this PR - the line above should also include `required=False`, or else it's impossible to create an EC without selecting a catalog:
screen shot 2016-11-24 at 10 53 19 pm

The above error is very inconvenient on a devstack for someone who doesn't have the catalog service running.

PR #8, opened 2016-11-11
Manage Learners

e-kolpakov:
**Description:** This PR adds learner management view to enterprise Django admin.

**JIRA:** [ENT-48](https://openedx.atlassian.net/browse/ENT-48)

**Dependencies:** cross-dependent on https://github.com/edx/edx-platform/pull/14003

1. Switch to this branch.
2. Make sure edx-enterprise is installed into edx-platform virtualenv locally (i.e. `pip install -e /edx/src/edx-enterprise` or `pip install -e git+https://...`)
3. Add `'django_object_actions'` to  `OPTIONAL_APPS` in common.py.

**Testing instructions:**
1. Open Enterprise Customer record in Django admin
2. In the top-right corner, next to "History" button, "Manage Learners" button should appear. Click on it.
3. You will arrive to "Manage Learners" view for that particular Enterprise Customer.
4. Check that breadcrumbs strip is correct (should contain link to Enterprise Customer record and link to lit of ECs)
5. Try linking a learner via form on the right side
5.1. By username - if user exists, his email should be added to a list. If not, validation error should be shown
5.2. By email - if invalid email is put, validation message should be shown. If email of an existing user is provided, username is shown in the list
5.3. In both cases, if user is already linked to EC (including current), validation error should be shown.
6. Make sure added learners appear in the list.
6.1. Make sure they survive page refresh.
6.2. Make sure they appear in Enterprise Customer User admin with correct data (i.e. linked to correct EC)
7. Try unlinking users
7.1. Make sure confirmation is shown.
7.2. Make sure survives page refresh.
7.3. Make sure they disappear from Enterprise Customer User admin.

**Reviewers:**
- [x] @mtyaka 
- [x] @saleem-latif and/or @asadiqbal08 

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed
- [x] Submit updates to cookiecutter template (docs, consistent `'` vs. `"` in setup.py etc.)

**Post merge:**
- [ ] Create a tag

**Author concerns:**
1. Lack of tests for JS part
2. Might be a good time to add integration tests - but this would be time-consuming.


e-kolpakov:
@mtyaka @asadiqbal08 @saleem-latif FYI - PR adding learners management. Not 100% ready as it lacks tests, but I would appreciate any feedback. There are a couple of unelegant solutions, so any suggestions to achieve the same in a better way are very welcome.

test_utils/factories.py
```diff
@@ -64,4 +66,28 @@ class Meta(object):
...
+
+    class Meta(object):
+        """
+        Meta for UserFactory.
+        """
+
+        model = User
+
+    email = factory.LazyAttribute(lambda x: FAKER.email())
+    username = factory.LazyAttribute(lambda x: FAKER.email())
```
mtyaka: Is `email` here intentional? Doesn't faker have `.username()` or similar?

enterprise/models.py
```diff
@@ -94,21 +109,33 @@ class EnterpriseCustomerUser(TimeStampedModel):
     """
 
     enterprise_customer = models.ForeignKey(EnterpriseCustomer, blank=False, null=False)
-    user_id = models.PositiveIntegerField(null=False, blank=False)
```
mtyaka: Maybe we should keep the `user_id`, but allow it to be `NULL`? Email can be changed by the user after registration.

e-kolpakov: @mtyaka That's a good point - I'll have to think about it a bit.

The problem here is that there's a use case when previously not seen email is sent. Such email should be recorded and when user registers with the email it will be automatically linked to EC. Current approach allows supporting this case (and other cases) without any hooks into registration workflow.

So, looks like we've got two options: 
1. Revert to use user_id and install some hook into registration workflow (so that when use uses an email already registered with EnterpriseCustomer EnterpriseCustomerUser record is updated to contain user_id)
2. Continue useing `user_email` - in this case we'll have to track email changes.

As I've said, I'll have to think about it, but suggestions welcome.

e-kolpakov:
@saleem-latif @asadiqbal08 This PR is ready for review.

enterprise/templates/enterprise/admin/manage_learners.html
```diff
@@ -0,0 +1,100 @@
+{% extends "admin/base_site.html" %}
+{% load i18n static admin_urls %}
+
+{% block extrastyle %}
+<!-- TODO: find a way to extract it into a file and make edx-platform find that file -->
```
saleem-latif: @e-kolpakov you  can create a directory named 'static\enterprise' (use 'enterprise' prefix so that your files do not clash with existing files in edx-platform) inside `enterprise` app and put your static files in that directory, those static files can be accessed via django's static template tag 
`<script type="text/javascript" src="{% static 'enterprise/js/test.js' %}"></script>`
or
`<link type="text/css" rel="stylesheet" href="{% static 'enterprise/css/test.css' %}">`.
edx-platform will be able to find those file because of `django.contrib.staticfiles.finders.AppDirectoriesFinder` [here](https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L1203).

enterprise/templates/enterprise/admin/manage_learners.html
```diff
@@ -0,0 +1,100 @@
...
+        <td>{% if learner.user %}{{ learner.user.username }}{% else %}{% trans "N/A" %}{% endif %}</td>
+        <td>{{ learner.created }}</td>
+        <td class="delete-cell"><input type="button" value="Unlink"></td>
+      </tr>
+    {% endfor %}
+    </tbody>
+  </table>
+</div>
+
+<!-- TODO: find a way to extract it into a file and make edx-platform find that file -->
```
saleem-latif: Same as above

enterprise/admin/forms.py
```diff
@@ -0,0 +1,51 @@
...
+    Form to manage learner additions.
+    """
+    email = forms.CharField(label=_("Email or Username"))
+
+    def clean_email(self):
+        """
+        Clean email form field
+        :return: str email
+        """
+        email_or_username = self.cleaned_data['email']
```
mtyaka: nit: It might be good to `.strip()` whitespace here.

enterprise/models.py
mtyaka: +1 for @bradenmacdonald's suggestion. It's more robust and consistent with how things are handled elsewhere in edx-platform.

mtyaka:
I tested this locally. The code works well, but we should decide how we want to model pending users.

enterprise/admin/forms.py
```diff
@@ -0,0 +1,51 @@
...
+        try:
+            validate_email(email)
+        except ValidationError:
+            message = _("{email_or_username} does not appear to be a valid email or known username").format(
+                email_or_username=email
+            )
+            raise ValidationError(message)
+
+        try:
+            existing_record = EnterpriseCustomerUser.objects.get(user_email=email)
```
e-kolpakov: @mtyaka As far as I can tell it is not supported. @mattdrayer or am I missing something?

enterprise/admin/__init__.py
```diff
@@ -66,6 +61,30 @@ def logo(instance):
             return instance.branding_configuration.logo
         return None
 
+    def manage_learners(self, request, obj):  # pylint: disable=unused-argument
+        """
+        Object tool handler method - redirects to "Manage Learners" view
+        """
+        # url names coming from get_urls are prefixed with 'admin' namespace
+        manage_learners_url = reverse('admin:'+UrlNames.MANAGE_LEARNERS, args=(obj.uuid,))
```
zubair-arbi: please add spaces around operator `'admin:' + UrlNames.MANAGE_LEARNERS`

enterprise/admin/__init__.py
```diff
@@ -66,6 +61,30 @@ def logo(instance):
...
+        return HttpResponseRedirect(manage_learners_url)
+
+    manage_learners.label = "Manage Learners"
```
zubair-arbi: Please be consistent with single quotes everywhere

enterprise/admin/views.py
```diff
@@ -0,0 +1,119 @@
...
+
+from enterprise.admin.forms import ManageLearnersForm
+from enterprise.models import EnterpriseCustomer, EnterpriseCustomerUser
+
+
+class EnterpriseCustomerManageLearnersView(View):
+    """
+    Manage Learners view.
+
+    Lists learners linked to chosen Enterprise Customer and allows adding and deleting them.
```
zubair-arbi: Please use max column length to 79 for readability in doctstrings.

enterprise/admin/views.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    template = "enterprise/admin/manage_learners.html"
```
zubair-arbi: use single quotes

test_settings.py
```diff
@@ -31,13 +31,42 @@ def root(*args):
...
     'django.contrib.contenttypes',
     'django.contrib.sites',
+    'django.contrib.sessions',
+    'django.contrib.admin',  # only used in DEBUG mode
     'enterprise',
 )
 
+MIDDLEWARE_CLASSES = [
+    "django.contrib.sessions.middleware.SessionMiddleware",
+    "django.contrib.auth.middleware.AuthenticationMiddleware"
```
zubair-arbi: single quotes

tests/admin/test_view.py
```diff
@@ -0,0 +1,208 @@
...
+    def _test_common_context(self, actual_context, context_overrides=None):
+        """
+        Test common context parts.
+        """
+        expected_context = {}
+        expected_context.update(self.default_context)
+        expected_context.update(context_overrides or {})
+
+        for context_key, expected_value in six.iteritems(expected_context):
+            assert actual_context[context_key] == expected_value
```
zubair-arbi: can we please use [unittest assert methods, e.g. assertEqual, assertNotEqual etc](https://docs.python.org/2/library/unittest.html#assert-methods) everywhere?

tests/test_models.py
```diff
@@ -74,6 +76,33 @@ def test_string_conversion(self, method):
 
 @mark.django_db
 @ddt.ddt
+class TestEnterpriseCustomerUserManager(unittest.TestCase):
+    """
+    Tests EnterpriseCustomerUserManager.
+    """
+
+    @ddt.data(
+        'some_email@example.com',
```
zubair-arbi: I would suggest using `first_email`, `second_email` and `third_email` as names

tests/admin/test_view.py
e-kolpakov: @zubair-arbi I find it really strange using `py.test` for test runner and not using anything else, especially smart assertions - the [top feature](http://doc.pytest.org/en/latest/#id1) as listed by pytest developers/documentation team.

tests/test_models.py
e-kolpakov: @zubair-arbi this is totally a matter of taste.

tests/test_models.py
```diff
@@ -86,15 +115,30 @@ def test_string_conversion(self, method):
...
+    @ddt.data(
+        'albert.einstein@princeton.edu', 'richard.feynman@caltech.edu', 'leo.susskind@stanford.edu'
+    )
+    def test_user_property_user_exists(self, email):
+        user_instance = UserFactory(email=email)
+        enterprise_customer_user = EnterpriseCustomerUserFactory(user_email=email)
+        assert enterprise_customer_user.user == user_instance  # pylint: disable=no-member
+
+    @ddt.data(
+        'yoda@jeditemple.net', 'luke_skywalker@resistance.org', 'darth_vader@empire.com'
```
zubair-arbi: Nice 👍

enterprise/templates/enterprise/admin/manage_learners.html
```diff
@@ -0,0 +1,100 @@
...
+
+<div class="forms-panel">
+  <h1>{% trans "Link learner" %}</h1>
+  <form action="" method="post">
+    {% csrf_token %}
+    {{ manage_learners_form }}
+    <input type="submit" value="Submit" />
+</form>
+</div>
+{% endblock %}
```
zubair-arbi: please add newline. There is a setting in pycharm `Editor > General > Other > 'Ensure line feed at file end on Save'` to automate this.

tests/admin/test_view.py
zubair-arbi: @e-kolpakov awesome, I am fine with current usage. thanks for the context.

enterprise/templates/enterprise/admin/manage_learners.html
e-kolpakov: @saleem-latif I've tried doing something like this, but was unsuccessful. Not exactly what you're suggesting, so maybe it'll help. I'll give it a try (but tomorrow, I'm sure I can't fit it today).

enterprise/templates/enterprise/admin/manage_learners.html
e-kolpakov: @saleem-latif same as above :)

enterprise/models.py
mattdrayer: If we must depend on `django.contrib.auth.User`, and we do not want to have references everywhere, then we could introduce a "user data service" (eg, `services/user_data.py`) and hide references to auth.User behind this abstraction.  This would be helpful for when `enterprise` is moved to its own IDA and the user data becomes truly remote.  This pattern could be expanded to cover many remote services for future workflow orchestration -- registration, enrollment, grading, data sharing, etc...

enterprise/admin/forms.py
```diff
@@ -0,0 +1,51 @@
...
+class ManageLearnersForm(forms.Form):
+    """
+    Form to manage learner additions.
+    """
+    email = forms.CharField(label=_("Email or Username"))
+
+    def clean_email(self):
+        """
+        Clean email form field
+        :return: str email
```
zubair-arbi: @e-kolpakov nice suggestion. Its upto us which pattern we follow going forward, I am fine with whatever pattern we choose. Only thing I would say is that we have a very wide community of open source users so we have to be consistent what ever we do.

enterprise/admin/views.py
e-kolpakov: @zubair-arbi There's no way to enforce it (pydocstyle does not support that yet), so I would suggest we drop that suggestion for now.

enterprise/admin/views.py
e-kolpakov: @zubair-arbi I prefer double quotes :)

test_settings.py
e-kolpakov: @zubair-arbi https://github.com/edx/edx-enterprise/pull/8#discussion_r88033108

enterprise/admin/views.py
cpennington: A recommendation I've seen (not sure how I feel about it) is to use single-quotes for strings that machines will use, and double quotes for strings that humans will read.

e-kolpakov: @cpennington too late! :)

enterprise/admin/forms.py
e-kolpakov: @cpennington Thanks - that works. However, it should have been added to cookiecutter template - I'll send a PR with improvements as soon as I land this one.

cpennington: Great, thanks!

enterprise/admin/views.py
bradenmacdonald: @cpennington I like that approach (machine vs human strings); it's something I've been trying to do lately, but I didn't know it was "a thing". It should also help with i18n.

In any case, if there's no edx style prescription, @e-kolpakov can decide here.

enterprise/models.py
bradenmacdonald: @e-kolpakov You can also move a lot more of the new `django.contrib.auth.models.User` references to the `admin.py` code rather than `models.py` if you want to keep the models code more "pure". I think the only reference you need in `models.py` is within `@property def user(self):` and even that could technically be removed if other parts of the code can live without that convenience method. The `post_save` hook might need to be moved to another file, because if a non-LMS app starts using this, we wouldn't want it registering that signal handler (assuming user_id is always assumed to be an LMS user and not "user in the current IDA").

@mattdrayer Has any work around that sort of user data service been done, for the edX IDA architecture in general?

mattdrayer: The only real-world implementation that comes to mind is the XBlock API/runtime, although the concept of a "user IDA" has been coming up frequently in recent architectural discussions.

enterprise/admin/views.py
e-kolpakov: @bradenmacdonald @cpennington I believe this is a python version of tabs vs. spaces holy war - any decision is as good as others, provided that it is consistently followed throughout the project.

e-kolpakov:
@mtyaka @saleem-latif @asadiqbal08 please review and approve this PR.

codecov.yml
```diff
@@ -0,0 +1,12 @@
+coverage:
+  status:
+    project:
+      default:
+        enabled: yes
+        target: auto
+    patch:
+      default:
+        enabled: yes
+        target: 100%
```
mattdrayer: @e-kolpakov is this value supposed to be an integer versus a percentage?

codecov.yml
```diff
@@ -0,0 +1,12 @@
+coverage:
+  status:
+    project:
+      default:
+        enabled: yes
+        target: auto
```
mattdrayer: @e-kolpakov I'm not sure what the "auto" value of `target` means here -- in the example provided by @rlucioni there was an integer value for this field.

mattdrayer: Ah, I am reading this now: https://github.com/codecov/support/wiki/Patch-Status -- perhaps it's all good

codecov.yml
rlucioni: @e-kolpakov try getting rid of the percentage sign. You want something like https://github.com/edx/course-discovery/blob/master/codecov.yml. You should probably disable comments, too; they're noisy.

e-kolpakov: @rlucioni Still green build. Either I've been tired and missed something stupid and obvious or it just ignores the codecov.yml for some reason.

I've tried both yaml and yml, if that makes any difference (but again I'm not 100% sure I haven't made a mistake somewhere).

saleem-latif:
@e-kolpakov I tried testing it locally but I get this error, any ideas what it is
```
2016-11-17 01:34:08,174 ERROR 8682 [django.request] base.py:256 - Internal Server Error: /admin/enterprise/enterprisecustomer/
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 164, in get_response
    response = response.render()
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/template/response.py", line 158, in render
    self.content = self.rendered_content
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/template/response.py", line 133, in rendered_content
    template = self._resolve_template(self.template_name)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/template/response.py", line 88, in _resolve_template
    new_template = self.resolve_template(template)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/template/response.py", line 80, in resolve_template
    return loader.get_template(template, using=self.using)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/template/loader.py", line 46, in get_template
    raise TemplateDoesNotExist(template_name)
TemplateDoesNotExist: django_object_actions/change_list.html
[17/Nov/2016 01:34:08] "GET /admin/enterprise/enterprisecustomer/ HTTP/1.1" 500 261457
```

e-kolpakov:
@saleem-latif You haven't added `django_object_actions` to LMS common.py. It is in the description **installation instructions**, item 3.

codecov.yml
e-kolpakov: @rlucioni There was an error in the config, so now it works properly - thank you for you help!

saleem-latif:
@e-kolpakov Thanks, it is working now, and changes LGTM 👍

mtyaka:
@e-kolpakov The PR description currently says this is cross-dependant on itself:

> Dependencies: cross-dependent on #8

Does it depend on something else?

enterprise/admin/forms.py
```diff
@@ -0,0 +1,75 @@
...
+    email = forms.CharField(label=_("Email or Username"))
+
+    def clean_email(self):
+        """
+        Clean email form field
+
+        Returns:
+            str: email to link
+        """
+        message_template = _(
```
mtyaka: Any reason to define this here instead of where it's used below (it's only used once)?

enterprise/admin/views.py
```diff
@@ -0,0 +1,136 @@
...
+
+        context.update({self.ContextParameters.MANAGE_LEARNERS_FORM: manage_learners_form})
+
+        return render(request, self.template, context)
+
+    def delete(self, request, customer_uuid):
+        """
+        Handle DELETE request - handle unlinking learner.
+
+
```
mtyaka: nit: extra newline

enterprise/models.py
```diff
@@ -83,29 +94,160 @@ def __repr__(self):
...
+        pending_linked_users_kwargs = self._get_supported_filter_options(kwargs, PendingEnterpriseCustomerUser)
+        if with_pending:
+            pending_linked_users = PendingEnterpriseCustomerUser.objects.filter(**pending_linked_users_kwargs)
+        else:
+            pending_linked_users = []
+
+        # list materializes the queryset, but we need it materialized for testing - chain creates a generator that
+        # is consumed at least once (when rendering a template), so by the time test gets to examine the context
+        # it contains exhausted generator. So, we're sacrificing laziness for repeatability
+        # TODO: find a way to retain queryset properties (laziness, repeatability)
```
mtyaka: I think it would be fine to just display the two types of links in separate sections("Linked Users", "Pending Users"), so that we don't have to worry about combining the two lists, but that's out of scope here.

e-kolpakov:
@mtyaka Right, it's copy-paste error. It's dependent on https://github.com/edx/edx-platform/pull/14003

mtyaka:
It works great :+1:

- [x] I tested this: Added and removed users (both existing and "pending") using the new "Manage Learners" view
- [x] I read through the code
- [ ] ~~I checked for accessibility issues~~
- [x] Includes documentation

enterprise/models.py
e-kolpakov: @mtyaka That's a good suggestion - I think it's worth implementing it now rather than delay it.

e-kolpakov:
@mtyaka Addressed the notes, you might want to take another look. If not, I'll merge it in about half an hour.

mtyaka:
@e-kolpakov It looks great, thanks!

haikuginger:
@e-kolpakov, you merged this with an incomplete migration - you have `unique=True` flagged in the code, but not enforced in the DB. This is causing tests with new migrations which include that (as automatically generated) to fail. Could you look into implementing a PR that adds a migration for that and fixes to ensure that tests with that change pass?

e-kolpakov:
@haikuginger Sure. Sorry if it caused any trouble.

PR #9, opened 2016-11-11
ENT-50: Link EnterpriseCustomer to IDP

saleem-latif:
Hi @mattdrayer, @asadiqbal08 , @zubair-arbi , @e-kolpakov , @haikuginger 

Kindly take a look at this PR, it adds an association between `EnterpriseCustomer` and IdP.

__Testing instructions:__

1. This PR adds only a `PositiveIntegerField` in `EnterpriseCustomer` model with updates to django admin interface
   - Open "Enterprise Customer" page in Django admin
   - Try adding a new Enterprise Customer, try adding an "Enterprise Customer" with and without selecting an "identity provider". If no "identity provider" is selected "Enterprise Customer" will be saved with `None` as identity provider.


mattdrayer:
@asadiqbal08 @zubair-arbi @haikuginger @e-kolpakov FYI

asadiqbal08:
@saleem-latif kindly update the `CHANGELOG.rst` with your version changes.

enterprise/admin.py
```diff
@@ -36,13 +37,54 @@ class EnterpriseCustomerBrandingConfigurationInline(admin.StackedInline):
...
+
+    A model form that converts an integer field to a foreign key if a particular
+    model is present in the system.
+    """
+
+    try:
+        # We give a choice field to user only if SAMLProviderConfig is present, otherwise we show
+        # an integer field, Since, we will be adding a custom enterprise admin so we will also be removing
+        # this dependency.
+        from third_party_auth.models import SAMLProviderConfig
```
asadiqbal08: I wonder, what will happened with the `list_display --> identity_provider UI behavior` if ImportError raised before its declaration ?

saleem-latif: @asadiqbal08 for `list_display` django admin displays the value or `identity_provider` model field (an integer). and in that case we do not have to worry about the import.

enterprise/admin.py
```diff
@@ -36,13 +37,54 @@ class EnterpriseCustomerBrandingConfigurationInline(admin.StackedInline):
     can_delete = False
 
 
+class ModelChoiceIntegerField(forms.ModelChoiceField):
+    """
+    A custom integer field for admin form, this field treats an integer field as a foreign key.
+    """
+
+    def to_python(self, value):
```
asadiqbal08: It seems me, It will show the `ID` in EnterpriseCustomer `DetailView`, right ? but the EnterpriseCustomer `ListView` will display the `objects` instead of ID in dropdown. If YES then can we display the text in dropdown with some explanatory text or ID instead of `SAMLProviderConfig object` ?

enterprise/admin.py
```diff
@@ -36,13 +37,54 @@ class EnterpriseCustomerBrandingConfigurationInline(admin.StackedInline):
...
+    A model form that converts an integer field to a foreign key if a particular
+    model is present in the system.
+    """
+
+    try:
+        # We give a choice field to user only if SAMLProviderConfig is present, otherwise we show
+        # an integer field, Since, we will be adding a custom enterprise admin so we will also be removing
+        # this dependency.
+        from third_party_auth.models import SAMLProviderConfig
+        identity_provider = ModelChoiceIntegerField(SAMLProviderConfig.objects.all(), required=False)
```
asadiqbal08: Should not we ONLY fetch those `SAMLProviderConfig` that are `enabled` in system ?

enterprise/models.py
```diff
@@ -134,7 +136,7 @@ def logo_path(instance, filename):
     return fullname
 
 
-class EnterpriseCustomerBrandingConfiguration(TimeStampedModel):
+class EnterpriseCustomerBrandingConfiguration(TimeStampedModel):    # pylint: disable=model-missing-unicode
```
e-kolpakov: @saleem-latif Don't suppress this - please add python_2_unicode_compatible decorator, `__unicode__`, `__str__` and tests for string conversion. You might be able to cherry-pick it from my PR where I've fixed that: https://github.com/edx/edx-enterprise/pull/8/files#diff-4bcae564bc272f5c2e6753dc6021eb1cR164.

If not, feel free to copy the methods and [tests](https://github.com/edx/edx-enterprise/pull/8/files#diff-327b130c82fa4e83e97f6c1a50a0da1cR163) - I'll resolve conflicts later

enterprise/models.py
```diff
@@ -68,6 +68,8 @@ class Meta(object):
     active = models.BooleanField(default=True)
     history = HistoricalRecords()
 
+    identity_provider = models.PositiveIntegerField(null=True, blank=True, default=None)
```
e-kolpakov: @saleem-latif suggestion: it might make sense to link to IdP via its slug - it should be more human-friendly (especially if `import SAMLProviderConfig` fails).

enterprise/admin.py
e-kolpakov: @saleem-latif At a glance, there might be a simpler (and probably less confusing to user) way to achieve the same. Here's the idea:

1. This form is still used
2. `identity_provider` becomes just a IntegerField (or maybe [TypedChoiceField](https://docs.djangoproject.com/en/dev/ref/forms/fields/#typedchoicefield) - depends on if you keep the field integer or switch to using IdP slugs).
3. In the form `__init__`, do `self.fields['identity_provider'].choices = get_available_idps()`
4. `get_available_idps` is a method that fetches `SAMLProviderConfig`s. It should live in `utils.py` or maybe `api.py`  -there's another PR doing `try: import something except: pass`, so it might make sense to collect such dependencies into one file.

enterprise/admin.py
```diff
@@ -36,13 +37,54 @@ class EnterpriseCustomerBrandingConfigurationInline(admin.StackedInline):
     can_delete = False
 
 
+class ModelChoiceIntegerField(forms.ModelChoiceField):
+    """
+    A custom integer field for admin form, this field treats an integer field as a foreign key.
```
zubair-arbi: Please use max column length to 79 for readability in doctstrings.

enterprise/admin.py
```diff
@@ -36,13 +37,54 @@ class EnterpriseCustomerBrandingConfigurationInline(admin.StackedInline):
...
+    """
+
+    def to_python(self, value):
+        """
+        Return an integer instead of a model instance.
+
+        Since we are emulating Foreign key from an Integer Field we need to return integer instead of fetching the
```
zubair-arbi: same comment about docstrings length

enterprise/__init__.py
```diff
@@ -4,6 +4,6 @@
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = '0.1.2'
+__version__ = '0.1.3'
```
zubair-arbi: @mattdrayer @e-kolpakov what is the procedure for versioning and releasing? I think we should update this version only when we do a release (and might [update edx-platform](https://github.com/edx/edx-platform/pull/13917/files) correspondingly)

e-kolpakov: @zubair-arbi @mattdrayer we haven't decided on the versioning yet, but so far we more or less adhered to "continuous integration" approach - releasing new version as soon as it is merged. This app is not used in production though, so we might change the approach later, but so far I would suggest keeping the "aggressive" releasing strategy as long as we can. In that case, every PR is a new minor version, and it is released to edx-platrofm right after it is tagged.

zubair-arbi: @e-kolpakov Ok. Just curious that what pattern we are following for versioning scheme? http://semver.org

e-kolpakov: @zubair-arbi an ad-hoc one - increase PATCH with every PR and increase MINOR whenever someone feels it is about time. And increase MAJOR to 1 when it is officially released (and I'm not aware if we ever increased anything beyond 1.x). Might be a good occasion to revise that.

@mattdrayer @haikuginger @saleem-latif @asadiqbal08  Thoughts?

saleem-latif:
@e-kolpakov , @asadiqbal08 , @zubair-arbi I have incoroporated your feedback, kindly take a look now

enterprise/admin.py
```diff
@@ -36,27 +38,47 @@ class EnterpriseCustomerBrandingConfigurationInline(admin.StackedInline):
     can_delete = False
 
 
+class EnterpriseCustomerForm(forms.ModelForm):
+    """
+    A custom model form to emulate a foreign key on an integer field.
+
+    A model form that converts an integer field to a foreign key if a particular
```
e-kolpakov: @saleem-latif nit: docstring is now inconsistent with actual code

asadiqbal08:
👍

enterprise/__init__.py
mattdrayer: I would prefer to follow http://semver.org since that is the generally-accepted approach for other edX packages.  We need to at least increase the patch with every change because this is how the package is referenced in edx-platform.

mattdrayer: MINOR updates are normally used for enhancements and fixes which do not break contracts (syntax) or alter existing behavior (semantics) in a backwards-incompatible way.  Breaking changes are represented by MAJOR version updates.

enterprise/utils.py
```diff
@@ -0,0 +1,38 @@
...
+        # this dependency.
+        from third_party_auth.models import SAMLProviderConfig
+        return SAMLProviderConfig.objects.current_set().filter(enabled=True).all()
+    except ImportError:
+        raise ValueError("SAMLProviderConfig is not available.")
+
+
+def get_idp_choices():
+    """
+    Get a list of ipd choices for enterprise customer.
```
mattdrayer: "ipd" -> "IdP"

test_utils/factories.py
```diff
@@ -46,6 +46,7 @@ class Meta(object):
     name = factory.LazyAttribute(lambda x: FAKER.company())
     active = True
     site = factory.SubFactory(SiteFactory)
+    identity_provider = factory.LazyAttribute(lambda x: FAKER.slug())
```
mattdrayer: I know we are using the `fuzzy` library in other code projects -- does it make sense to use it for `enterprise` as well?

enterprise/utils.py
mattdrayer: actually I would prefer that we use the full "identity provider" term in our docstrings, versus abbreviations.

test_utils/factories.py
e-kolpakov: @mattdrayer fuzzy is considered deprecated: http://factoryboy.readthedocs.io/en/latest/fuzzy.html and https://github.com/FactoryBoy/factory_boy/issues/271/

mattdrayer: AH, good to know, thanks!

saleem-latif:
@e-kolpakov [this](https://codecov.io/gh/edx/edx-enterprise/compare/38014ed533eb7834f37e56a82e92920fa70bbbe8...734579730a376e9b3dae56d0f1496117bbf583c3#656E74657270726973652F7574696C732E7079-22) is the line missing coverage. I tried adding tests for this but this line is executed only if the import statement
`from third_party_auth.models import SAMLProviderConfig `
is successful, which I do not think is possible in tests environements.

saleem-latif:
@mattdrayer , @e-kolpakov I have incorporated your feedback, kindly take a look.

e-kolpakov:
@saleem-latif 

> Since, third_party_auth is n external app that is only available when enterprise app is installed inside edx-platform.

You can mock it. Something like 

    with mock.patch('enterprise.utils.third_party_auth.models.SAMLProviderConfig'):
        import enterprise.utils   # you've got to mock before ANY import for mock to be effective
        # assertions

While this might seem minor dent in coverage, in fact it means that `get_available_idps` is not tested for the "primary" case. Actually, taking a closer look it appears that there are no tests for it at all (and I have no idea why codecov reports it as covered).

I understand that this is really frustrating to go into that much of trouble to cover just one branch in one method, but this is an important branch.

tests/test_models.py
```diff
@@ -147,6 +147,23 @@ def test_branding_configuration_saving_successfully(self):
...
+    )
+    def test_string_conversion(self, method):
+        """
+        Test ``EnterpriseCustomer`` conversion to string.
+        """
+        branding_config = EnterpriseCustomerBrandingConfiguration(
+            enterprise_customer=EnterpriseCustomerFactory(),
+            logo='test1.png'
+        )
+        expected_to_str = "<EnterpriseCustomerBrandingConfiguration {ID}>: {enterprise_name}".format(
```
zubair-arbi: nit: be consistent with single quotes

saleem-latif:
@zubair-arbi @e-kolpakov made the changes you suggested.

zubair-arbi:
LGTM 👍

enterprise/__init__.py
e-kolpakov: @mattdrayer In this case we'd have to increase MINOR (as we're adding new functionality with each PR), while we're increasing PATCH.

tests/test_util_functions.py
```diff
@@ -0,0 +1,119 @@
...
+        Test get_available_idps returns correct value or raises a ValueError if
+        thirdParty_auth is not installed.
+        """
+        with self.assertRaises(ValueError):
+            utils.get_available_idps()
+
+        expected_list = ['test1', 'test2']
+
+        with MockThirdPartyAuth(expected_list):
+            self.assertListEqual(expected_list, utils.get_available_idps())
```
e-kolpakov: @saleem-latif nit: swap `expected_list` and `utils.get_available_idps()`. I believe first parameter should be actual value and second parameter - expected value.

Also, it might make sense to start using [pytest assertions](http://doc.pytest.org/en/latest/assert.html) (`assert somerthing`) instead of unittest assertions (`self.assertSomething`)

enterprise/__init__.py
e-kolpakov: @mattdrayer In this case we'd have to increase MINOR (as we're adding new functionality with each PR), while we're increasing PATCH.

mattdrayer: So this should be `0.1.2` -> `0.2.0`

e-kolpakov:
@saleem-latif Congrats! Please delete the branch if you don't need it anymore

saleem-latif:
@e-kolpakov Thanks, I have deleted the branch, will add myself to AUTHORS in the next PR :)

enterprise/utils.py
```diff
@@ -0,0 +1,38 @@
...
+def get_idp_choices():
+    """
+    Get a list of identity providers choices for enterprise customer.
+
+    Return:
+        A list of choices of all identity providers, None if it can not get any available identity provider.
+    """
+    first = [("", "-"*7)]
+    try:
+        return first + [(idp.idp_slug, idp.name) for idp in get_available_idps()]
```
asadiqbal08: @saleem-latif  Should not we save the `idp.provider_id` instead of `idp_slug` here ? `provider_id` is also unique.

enterprise/models.py
```diff
@@ -67,7 +67,7 @@ class Meta(object):
     name = models.CharField(max_length=255, blank=False, null=False, help_text=_("Enterprise Customer name."))
     active = models.BooleanField(default=True)
     history = HistoricalRecords()
-
+    identity_provider = models.SlugField(null=True, blank=True, default=None)
```
asadiqbal08: I know that, One EC can be associated with One SAML-IDP. 
@mattdrayer @saleem-latif but one thing in confusing me, that is  **You can add multiple EC(s) against a single SAML-IDP.** If it is the required behavior then how can we fetch the particular EC against the tpa-hint (identity-provider) ? 

screen shot 2016-11-22 at 12 24 55 pm


`redirected_url?tpa_hint=saml-ubc`

bradenmacdonald: @asadiqbal08 We discussed this on another PR: https://github.com/edx/edx-enterprise/pull/11#discussion_r88712167 I think that this should have `unique=True`, though it requires a bit of work to make that possible.

I think we don't want to allow multiple IdPs per EC; we want to force that to be 1:1. (If we ever needed to allow multiple ECs for one IdP we could just create two IdPs with identical values but different IDs. I don't think we'll need that though.)

enterprise/utils.py
bradenmacdonald: @saleem-latif @asadiqbal08 Yes, you should definitely use the `provider_id` and not `idp_slug`. We may want to support other types of providers in the future (OAuth2, LDAP, ...), and it will be much easier to support that in the future if we use the generic ID `provider_id` and not the SAML-specific one.

Likewise, the `get_available_idps()` function is not the best way to approach this. You can just use:
```
try:
    from third_party_auth.provider import Registry
except ImportError:
    Registry = None

def get_idp_choices():
    choices = [("", "-"*7)]
    if Registry:
        choices += [(idp.provider_id, idp.name) for idp in Registry.enabled()]
    return choices
```

If we really need to only show SAML providers in this list, you can use `Registry.get_enabled_by_backend_name("tpa-saml")` instead of `Registry.enabled()`. I don't think there's any harm in also supporting other provider types now, but I could be missing something.

`Registry` is the public API to the third party auth backends and it should be used rather than accessing database tables directly, as in `SAMLProviderConfig.objects.current_set().filter(enabled=True).all()`. The Registry API also ensures that the resulting backends are sorted by name, which makes the admin nicer.



Three other notes:

1. Are you sure it's necessary to use `first = [("", "-"*7)]`? I thought that if `blank=True` on a field, the django forms library would generate the blank entry automatically?
1. When you declare the field `identity_provider = models.SlugField(null=True, blank=True, default=None)` in `models.py`, it should say what the allowed values are (`provider_id` values). Right now it doesn't indicate what sort of slugs should be stored.
1. Did this PR add a check to make sure that the IdP's `site` matches the EC's `site`? That's important and should raise an error if it's different.

saleem-latif: @asadiqbal08 , @bradenmacdonald thanks for the comment, I have created a PR https://github.com/edx/edx-enterprise/pull/16 that incorporates this along with other suggestions from ENT-76.

PR #11, opened 2016-11-18
[ENT-61] Move data sharing consent to Enterprise app

haikuginger:
This PR takes the previously-created SSO data sharing consent logic and moves it to live in the Enterprise app.

**JIRA tickets**: Implements ENT-61

**Dependencies**: Requires a specific branch of edx-platform; [available here](https://github.com/open-craft/edx-platform/tree/haikuginger/final-consent-hooks)

**Sandbox URL**: http://149.202.178.151 (use staff/edx username); sandbox currently at 5cec668.

**Testing instructions**:

1. In the sandbox admin, verify that the TestShib SAML IdP is linked to an EnterpriseCustomer, and that that EC does not require data sharing consent.
2. Register a new user, using the TestShib SAML IdP. Verify that the process is identical to the standard SSO process.
3. Disconnect that user from the IdP and sign out.
4. Modify the EC setting to request data sharing consent.
5. Register a new user, using the TestShib SAML IdP, noting that a checkbox for consent is present, checking it.
7. Disconnect this user from the IdP and sign out.
8. Register a new user, using the TestShib SAML IdP, noting that a checkbox for consent is present, but leaving it blank.
9. Sign out of this user, but do not disconnect it from the IdP.
11. In the admin panel, set the EC to require data sharing consent.
12. Attempt to sign in, using the TestShib SAML user account you registered previously. Note that it redirects you to a page stating that additional permissions are required. Do not grant permission, but click Submit, noting that you are logged out.
13. Again attempt to sign in, using the TestShib SAML user account you registered previously. Again note that you are redirected to a page stating that additional permissions are required. Check the box granting permissions and click Submit, noting that you are logged in successfully.
14. Disconnect the user from the IdP and sign out.
15. Attempt to register a new user using the TestShib SAML IdP. Note that the checkbox for data sharing consent is present; note that if you attempt to register a user without checking it, you encounter an error message.
16. Note that once registered, the OpenEdX platform behaves normally.

1. In functionality, this is very similar to what we already have. It does not currently include logic to create an EnterpriseCustomerUser object; we'll likely want to make the relevant changes as a follow-up.

2. The sandbox must be accessed directly via IP. We're using a load balancer to obtain HTTPS, and the domain name is pointed at that LB; however, Django is not aware of that and believes that everything is happening over HTTP. This has resulted in a failure where the redirect URL Django creates for us has an HTTP scheme, and the LB redirects the SAML postback to an HTTPS GET, dropping essential information. Long story short, it's fixable, and not an issue with the codebase, but for right now, just use the sandbox by IP address.

**Reviewers**
- [ ] @bradenmacdonald 
- [ ] @mattdrayer 
- [ ] @saleem-latif 

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag
      finished.


enterprise/middleware.py
```diff
@@ -0,0 +1,28 @@
+from enterprise.models import EnterpriseCustomer, UserDataSharingConsentAudit
+from enterprise.tpa_pipeline import (
+    active_provider_enforces_data_sharing,
+    get_consent_status_for_request,
+    get_enterprise_customer_for_request
+)
+class ResetSessionIfPipelineBrokenMiddleware(object):
+    """
+    Middleware signs the user out if they need to provide data sharing consent,
```
bradenmacdonald: Is there some way this can be implemented as a step in the pipeline, rather than as site-wide middleware? Middleware seems like overkill, and the pipeline is where this functionality seems to belong architecturally.

Or, if middleware is required, it would be nicer to move this code to a pipeline step and implement generic middleware in `third_party_auth` which simply ensures that the complete tpa pipeline always runs, and users cannot leave the pipeline prematurely.

enterprise/models.py
```diff
@@ -77,6 +77,35 @@ class Meta(object):
     history = HistoricalRecords()
     identity_provider = models.SlugField(null=True, blank=True, default=None)
     site = models.ForeignKey(Site, related_name="enterprise_customers")
+    
+    DATA_CONSENT_OPTIONAL = 'optional'
+    AT_LOGIN = 'at_login'
+    AT_ENROLLMENT = 'at_enrollment'
+    DATA_SHARING_CONSENT_CHOICES = (
+        (DATA_CONSENT_OPTIONAL, 'Optional'),
```
bradenmacdonald: Is there a "disabled/neither" option? Also, if I choose "optional", will it prompt users during SSO login, course enrollment, or both?

enterprise/models.py
```diff
@@ -77,6 +77,35 @@ class Meta(object):
...
+        (AT_LOGIN, 'At Login'),
+        (AT_ENROLLMENT, 'At Enrollment'),
+    )
+    
+    enable_data_sharing_consent = models.BooleanField(
+        default=False,
+        help_text=_(
+            "This field is used to determine whether data sharing consent is enabled or "
+            "disabled of users signing in using this SSO provider. If disabled, consent "
```
bradenmacdonald: "for" not "of".

Also "this" is not an SSO provider anymore, and this setting has implications for SSO login but also for non-SSO course enrollment. So I think this should really just say "This field is used to determine when to prompt users to consent to sharing their data with this enterprise customer. Consent can be required at login, when enrolling into eligible courses, or (...explain when/how 'optional' works...)."

enterprise/models.py
bradenmacdonald: Whoops, never mind my comments, I didn't notice that this was split into two fields. I'm going to delete these comments and add new ones.

enterprise/models.py
bradenmacdonald: "disabled **for** this enterprise customer's users."

("this" is not an SSO provider now)

enterprise/models.py
```diff
@@ -77,6 +77,35 @@ class Meta(object):
...
+    )
+    
+    enforce_data_sharing_consent = models.CharField(
+        max_length=25,
+        blank=False,
+        choices=DATA_SHARING_CONSENT_CHOICES,
+        default=DATA_CONSENT_OPTIONAL,
+        help_text=_(
+            "This field determines if data sharing consent is optional, if it's required at login, "
+            "or if it's required when registering for courses."
```
bradenmacdonald: I think this should be "when registering for **eligible** courses" because only courses associated with the EnterpriseCustomer (via the catalog selection) will show a data consent prompt. If the users enrolls in some random other course on edx.org, there must not be data sharing consent prompts.

Also, this doesn't explain whether the "Optional" mode will prompt users during login or during course enrollment. (Personally I think it's better to encourage at-enrollment as a default, since it provides a clearer audit trail and makes users more aware of for which courses their data will be shared.)

enterprise/models.py
```diff
@@ -90,6 +119,24 @@ def __repr__(self):
         """
         return self.__str__()
 
+    def enforces_data_sharing_consent(self, enforcement_location):
+        """
+        Does the provider enforce data sharing consent at the given point ?
```
bradenmacdonald: Replace "provider" with "enterprise customer". Same in next method below.

enterprise/models.py
```diff
@@ -321,6 +368,147 @@ def __repr__(self):
...
+
+    user = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+    )
+
+    enterprise_customer = models.ForeignKey(
+        EnterpriseCustomer,
+        on_delete=models.CASCADE,
+    )
```
bradenmacdonald: I think you should replace `user` and `enterprise_customer` with a ForeignKey to `EnterpriseCustomerUser` instead.

If there's some technical reason you want to keep them separate, `user` should not be a `ForeignKey` for consistency with the `user` field of `EnterpriseCustomerUser`. But I think an FK to `EnterpriseCustomerUser` makes way more sense here.

enterprise/models.py
```diff
@@ -321,6 +368,147 @@ def __repr__(self):
...
+    state = models.CharField(
+        max_length=8,
+        blank=False,
+        choices=STATE_CHOICES,
+        default=NOT_SET,
+        help_text=_(
+            "Stores whether the user linked to this model has consented to have "
+            "their information shared with the linked EnterpriseCustomer."
+        )
+    )
```
bradenmacdonald: @mattdrayer Are we planning to add a `course_id` field to this consent audit model eventually, so that users can consent on a per-course basis (at enrollment), as well as on the current per-enterprise-for-all-courses basis? I don't see a story for that in the backlog, but I seem to remember talking about that.

enterprise/tpa_pipeline.py
```diff
@@ -0,0 +1,153 @@
...
+        argument can either be "optional", 'at_login' or 'at_enrollment'
+    """
+    running_pipeline = request.session.get('partial_pipeline')
+    if running_pipeline:
+        customer = get_enterprise_customer_for_request(request)
+        return customer and customer.enforces_data_sharing_consent(enforcement_location)
+    return False
+
+
+def active_provider_requests_data_sharing(request):
```
bradenmacdonald: Is this method used anywhere?

enterprise/tpa_pipeline.py
```diff
@@ -0,0 +1,153 @@
...
+    customer = get_enterprise_customer_for_running_pipeline({'backend': backend.name, 'kwargs': kwargs})
+    if customer is None:
+        return
+
+    consent, _ = UserDataSharingConsentAudit.objects.get_or_create(user=user, enterprise_customer=customer)
+
+    if data_sharing_consent:
+        consent.enable()
+    else:
+        consent.disable()
```
bradenmacdonald: I think that if you change `UserDataSharingConsentAudit` to use `HistoricalRecords`, you can get rid of this need to call enable()/disable()/save() methods, and just use the much nicer single-statement

```python
UserDataSharingConsentAudit.objects.update_or_create(
    user=user,
    enterprise_customer=customer,
    defaults={'state': ENABLED if data_sharing_consent else DISABLED},
)
```

Same thing below in `GrantDataSharingPermissions`.`post()`

enterprise/tpa_pipeline.py
```diff
@@ -0,0 +1,153 @@
...
+    return get_enterprise_customer_for_sso(provider)
+
+
+def get_enterprise_customer_for_running_pipeline(pipeline):
+    """
+    Get the EnterpriseCustomer associated with a running pipeline
+    """
+    if Registry is None:
+        raise NotConnectedToEdX(
+            _("This package must be installed in an EdX environment to look up third-party auth providers.")
```
bradenmacdonald: DRY: Move this repeated string to within the `NotConnectedToEdX` definition itself.

enterprise/tpa_pipeline.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    Get the EnterpriseCustomer object tied to an identity provider.
+    """
+    if provider is None:
+        return None
+    try:
+        return EnterpriseCustomer.objects.get(identity_provider=provider.idp_slug)
+    except EnterpriseCustomer.MultipleObjectsReturned:
+        raise MultipleCustomersFound(
+            _("Multiple EnterpriseCustomer objects are tied to the same identity provider; expected one.")
```
bradenmacdonald: Why are we not enforcing this at the database level instead, by making the `identity_provider` field `unique=True` ?

enterprise/middleware.py
haikuginger: Based on my understanding of the pipeline, this cannot be done without middleware. The pipeline is entirely "opt-in"; it consists of state stored in the session. That state doesn't do anything unless it's interpreted by a view or form that understands it. If the user browses to a page outside the pipeline (meaning a page that doesn't eventually resolve to a redirect to re-enter the pipeline), then we can't "grab" the user to bring them back.

enterprise/models.py
haikuginger: 👍

enterprise/middleware.py
bradenmacdonald: @haikuginger I understand, hence my second suggestion of doing this in a more generic way.

enterprise/models.py
haikuginger: 👍

enterprise/models.py
haikuginger: I agree that an FK to EnterpriseCustomerUser makes the most sense, and is where we'll go _eventually_. However, that essentially also expands out this story to include decision-making about when to create an EnterpriseCustomerUser. @griffresch, @mattdrayer, thoughts on that?

bradenmacdonald: @haikuginger For now can we just throw an error if the EnterpriseCustomerUser does not yet exist (if the user is not yet linked to an EC)? I believe that at the moment, we are only supporting enterprise users that get manually linked to their EC by edX staff in the Django admin. I don't like an approach where we can model the data sharing consent for a user without also requiring that the user is linked to an EC. It invites data inconsistencies.

enterprise/models.py
```diff
@@ -321,6 +368,147 @@ def __repr__(self):
...
+
+    def __unicode__(self):
+        return 'UserDataSharingConsent: {} - {} - {}'.format(
+            self.user,
+            self.enterprise_customer.name,
+            self.state,
+        )
+
+
+class UserDataSharingConsentAuditHistory(TimeStampedModel):
```
haikuginger: We weren't using it on the feature branch for TPA, of which this is basically a clone, but I can look into whether that's an option here.

enterprise/tpa_pipeline.py
haikuginger: Not locally, but it is used in edx-platform - see the branch linked in the description.

enterprise/middleware.py
```diff
@@ -0,0 +1,28 @@
...
+        Conditionally sign out users who don't provide data sharing consent
+        """
+        view_module = view_func.__module__
+
+        if not active_provider_enforces_data_sharing(request, EnterpriseCustomer.AT_LOGIN):
+            return
+
+        running_pipeline = request.session.get('partial_pipeline')
+        if running_pipeline:
+            consent = get_consent_status_for_request(request)
```
haikuginger: 👍

enterprise/tpa_pipeline.py
haikuginger: Good to know!

enterprise/tpa_pipeline.py
haikuginger: So, we're already using a version of NotConnectedToEdX in #7; that one does have a variety of different strings, so while it's repeated here, the exception is used in a number of cases.

enterprise/tpa_pipeline.py
haikuginger: That'd work better; for whatever reason I didn't consider modifying the existing code instead. @mattdrayer, sound good to you?

enterprise/middleware.py
haikuginger: @bradenmacdonald, sorry, see the edited comment above.

enterprise/models.py
haikuginger: Sure.

enterprise/tpa_pipeline.py
haikuginger: Ah, we actually can't do unique=true. The default value stores a null in the DB, and Django allows duplicate nulls when unique=true, but emptying the field after that results in an empty Unicode string, which is not allowed to be duplicated when unique=true. This includes when I tried setting the value to None in the form - it saved as `u''`.

bradenmacdonald: Ah. They've [added a solution for that](https://github.com/django/django/commit/267dc4adddd2882182f71a7f285a06b1d4b15af0) to django 1.11 but we're obviously not using that yet... Maybe we can implement a clean() method that changes `u''` to `None`? If not, we can leave it as is.

enterprise/models.py
haikuginger: Actually, for right now, I'm going to go ahead and create the EC User for ease of testing; we won't be able to test the registration flow otherwise. @mattdrayer can have the final word on what the "finished product" for this PR should look like.

enterprise/tpa_pipeline.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def get_enterprise_customer_for_sso(provider):
+    """
+    Get the EnterpriseCustomer object tied to an identity provider.
+    """
+    if provider is None:
+        return None
+    try:
+        return EnterpriseCustomer.objects.get(identity_provider=provider.idp_slug)
```
bradenmacdonald: I assume this isn't your code, but we have an opportunity here to fix this to make it more general: The `identity_provider` field of Enterprise Customer should reference a complete [`provider_id`](https://github.com/edx/edx-platform/blob/f806f3a/common/djangoapps/third_party_auth/models.py#L158-L162) such as "`saml-mycorp`" and not just a SAML `provider_slug` like "`mycorp`". That way, we can support non-SAML providers in the future if we need to (OAuth2, Active Directory, or other types). We should not tie ourselves unnecessarily to SAML if the code can be fully generic.

This line would become
```
EnterpriseCustomer.objects.get(identity_provider=provider.provider_id)
```

enterprise/models.py
```diff
@@ -77,6 +77,35 @@ class Meta(object):
     history = HistoricalRecords()
     identity_provider = models.SlugField(null=True, blank=True, default=None)
```
bradenmacdonald: Same comment here, while I happen to be thinking of this: The `identity_provider` field of Enterprise Customer should reference a complete [`provider_id`](https://github.com/edx/edx-platform/blob/f806f3a/common/djangoapps/third_party_auth/models.py#L158-L162) such as "`saml-mycorp`" and not just a SAML `provider_slug` like "`mycorp`". That way, we can support non-SAML providers in the future if we need to (OAuth2, Active Directory, or other types). We should not tie ourselves unnecessarily to SAML if the code can be fully generic.

Right now, this field actually doesn't say what it references - it should have at least a comment explaining what sort of value is meant to be stored in this field, and that should be a `provider_id`.

If you want to consider this out of scope, that's fine, but it would be good to fix soon.

enterprise/tpa_pipeline.py
haikuginger: We could, but that'd still be application-layer rather than DB layer. Up to you.

enterprise/models.py
bradenmacdonald: Yes, you should be able to test the registration flow, thanks to #8: You can now link yet-unregistered users to an EC, which will store their email in the `PendingEnterpriseCustomerUser` table. When they register, the link-to-EC logic from #8 should take place before the verify-consent logic you're dealing with here.

If you use any other method of creating the EC, it won't properly test the integration between this and #8, and we could get bugs. So I think you should not add any new way of creating an EC User.

enterprise/tpa_pipeline.py
bradenmacdonald: Well the idea is that that `clean()` method would let you use `unique=True`, so it would still be the DB layer, no?

enterprise/models.py
haikuginger: @bradenmacdonald, I'd rather get the existing work up to spec first, and then we can go from there, if that's all right.

enterprise/tpa_pipeline.py
haikuginger: Yes, the unique check would be moved into the DB layer, but we'd be adding other application-layer logic to make the DB-layer uniqueness check feasible.

haikuginger:
Note that I've removed isort on this branch - this is because it was giving sorting instructions contradictory to pylint. Open to discussion on how to best address that. Will be jumping on later this morning to fill in tests and put together a sandbox.

enterprise/models.py
```diff
@@ -78,6 +78,35 @@ class Meta(object):
...
+        (AT_LOGIN, 'At Login'),
+        (AT_ENROLLMENT, 'At Enrollment'),
+    )
+
+    enable_data_sharing_consent = models.BooleanField(
+        default=False,
+        help_text=_(
+            "This field is used to determine whether data sharing consent is enabled or "
+            "disabled for users signing in using this enterprise customer. If disabled, consent "
+            "will not be requested, and eligible course data will not be shared."
```
mattdrayer: nit: I would just say "eligible data will not be shared", versus "course data"

enterprise/models.py
mattdrayer: @bradenmacdonald @haikuginger -- course-level consent is still being discussed, but I hope to have an answer this sprint -- @griffresch FYI

enterprise/tpa_pipeline.py
mattdrayer: How do we see the platform consuming this particular workflow in a future state where the local enterprise app is instead a distributed service?

enterprise/utils.py
```diff
@@ -5,6 +5,17 @@
 from __future__ import absolute_import, unicode_literals
 
 
+class NotConnectedToEdX(Exception):
+    """
+    Exception to raise when not connected to OpenEdX.
+
+    In general, this exception shouldn't be raised, because this package is
+    designed to be installed directly inside an existing OpenEdX platform.
```
mattdrayer: Since we're not expecting this exception to be triggered, if it DOES get triggered for some reason we'll need to keep an eye out for it -- should we write a special log entry?

haikuginger:
Rebased from 859a217.

enterprise/tpa_pipeline.py
bradenmacdonald: @mattdrayer In that case, perhaps the SSO code in the platform would use an API to ask the enterprise service if the provider requests data sharing, and then the enterprise service could requests that specific users  be redirected to the enterprise service (at a separate URL?) to consent during the SSO flow.

I'm still a bit fuzzy on how this thing will work when it becomes a service or even just gets used as a library by multiple IDAs. I think we should put more thought into that now.

enterprise/tpa_pipeline.py
bradenmacdonald: ^ That sounds fine to me, better than allowing invalid data and catching the exception later during user logins.

haikuginger:
Rebased from 5a5901d.

enterprise/middleware.py
bradenmacdonald: @haikuginger How did you end up handling this? (Discussion about the middleware you had in this PR and ensuring the pipeline gets completed)

enterprise/models.py
bradenmacdonald: @haikuginger My last comment here was that we should probably not be creating `EnterpriseCustomerUser` objects in this PR, in order to assure good integration with #8, but I see `EnterpriseCustomerUser.objects.get_or_create(...)` in a few places in the code now. Can you explain the rationale please?

enterprise/models.py
bradenmacdonald: It's ok, we are planning to fix this in ENT-76.

enterprise/models.py
haikuginger: @bradenmacdonald, I think this is a point where we need some clarification from @mattdrayer. My understanding was that the end goal would be that logging in with an identity provider associated to an EnterpriseCustomer would associate the user in question with the EnterpriseCustomer as well, without requiring the manual process of adding the user first.

To get this working without the creation of an EnterpriseCustomer is complicated - for example, we don't have a user email until after the registration form on which data sharing consent is requested is rendered. And, we could end up with a situation where the user has actually given consent (on that registration form), but we have no way of recording it.

If the goal is actually to require manual entry of each EnterpriseCustomerUser, then that's one thing, but I don't think it is.

enterprise/tpa_pipeline.py
haikuginger: Okay; I'll make that change.

enterprise/templates/enterprise/grant_data_sharing_permissions.html
```diff
@@ -0,0 +1,56 @@
...
+                            data-consent="${data_sharing_consent}"
+                            data-warning-message="${messages['warning'].get(data_sharing_consent, messages['warning']['optional'])}"
+                        />
+                        <label for="register-data_sharing_consent">
+                            ${_("I agree to allow {platform_name} to share data about my enrollment, completion and performance in all {platform_name} courses and programs where my enrollment is sponsored by {provider}.*").format(provider=sso_provider, platform_name=platform_name)}
+                        </label>
+                    </div>
+                    <input type="hidden" name="csrfmiddlewaretoken" value="${ csrf_token }" />
+                    <button type="submit" class="action-primary action">Submit</button>
+                    <p class="note">${messages['note'].get(data_sharing_consent, messages['note']['optional'])}
```
haikuginger: Done!

enterprise/models.py
haikuginger: Will do.

enterprise/middleware.py
haikuginger: So, I've made the middleware more generic, and moved it back into third_party_auth. However, based on my findings, a general solution for restricting the user to the TPA pipeline at large is not forthcoming.

The reason for this is that individual pipeline elements don't actually receive any request context to work with; they receive, essentially, an expansion of the pipeline's `kwargs` variable.

This is stored on the context, and is manipulable, but the problem is that it's only manipulable by actually _returning_ a dictionary with the variables to update. The problem being, of course, that the cases where we would need something like this most would be in cases where, rather than a dictionary that manipulates the pipeline kwargs, we're already returning a `redirect` object that sends the user to a difference page, and don't have the ability to return something else that would mutate the pipeline state.

Long story short, to have a fully-protected pipeline, we'd have to have something in the actual views that adds that protection, which is somewhat outside the scope here.

enterprise/models.py
bradenmacdonald: @haikuginger I think you are right; I see this in one of the docs: "edX Learners who come to edX through SSO from IDP are automatically added to the Enterprise Customer"

Carry on! Sorry for the confusion on my part.

haikuginger:
Rebased from 607b984.

enterprise/tpa_pipeline.py
haikuginger: Done.

haikuginger:
@bradenmacdonald, this is ready for a final review. Spawning a sandbox and will post a link here once it's ready.

enterprise/views.py
```diff
@@ -0,0 +1,184 @@
...
+        Update the EdX Mako paths to point to our consent template.
+        """
+        if add_lookup is None:
+            raise NotConnectedToEdX(
+                _('Must be installed in an OpenEdX environment to render this response.')
+            )
+        full_location = os.path.realpath(__file__)
+        directory = os.path.dirname(full_location)
+        template_location = os.path.join(directory, 'templates', 'enterprise')
+        add_lookup('main', template_location)
```
bradenmacdonald: This code should be moved to the `enterprise` app's `AppConfig.ready()` section. Doing it on every request is redundant and will leak memory. Or am I missing something that makes this only have effect for one request-response cycle?

enterprise/views.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def get(self, request):
+        """
+        Render a form to collect user input about data sharing consent.
+        """
+        self.patch_mako_lookup()
+
+        # Get the OpenEdX platform name
+        if configuration_helpers is None:
+            raise NotConnectedToEdX(_('Must be installed in an OpenEdX environment to use configuration helpers.'))
```
bradenmacdonald: It shouldn't be necessary to check this here. This whole view should only ever be enabled if the code is running inside the LMS.

enterprise/views.py
```diff
@@ -0,0 +1,184 @@
...
+            'platform_name': platform_name,
+            'sso_provider': customer.name,
+            'data_sharing_consent': 'required' if required else 'optional',
+            "messages": {
+                "warning": self.get_warning(customer.name, platform_name, required),
+                "note": self.get_note(customer.name, required),
+            },
+        }
+        if render_to_response is None:
+            raise NotConnectedToEdX(_('Must be installed in an OpenEdX environment to render this response.'))
```
bradenmacdonald: Same here - unnecessary, or should be. Please DRY this code by removing this repeated `if blah is None: raise NotConnectedToEdX` code throughout this file and just make sure that this whole module is only loaded/enabled (added to the urlconf) within the LMS.

tests/test_views.py
```diff
@@ -0,0 +1,367 @@
...
+                    os.path.realpath(__file__)
+                ),
+                os.pardir,
+                'enterprise',
+                'templates',
+                'enterprise'
+            )
+        )
+        GrantDataSharingPermissions.patch_mako_lookup()
+        assert add_lookup_mock.call_args == (('main', expected_path,), {})
```
bradenmacdonald: Not sure if this test is valuable as written. A better test might be to simply check if the new template file exists on the filesystem in the absolute path that was passed to `add_lookup_mock`

tox.ini
```diff
@@ -67,5 +67,4 @@ commands =
     rm tests/__init__.py
     pycodestyle enterprise tests test_utils
     pydocstyle enterprise tests test_utils
-    isort --check-only --recursive tests enterprise test_utils manage.py setup.py test_settings.py
```
bradenmacdonald: What specifically was the problem here? I think it would be better to disable or change the parameters of the pylint check if it conflicts with isort, rather than removing isort completely.

tests/test_models.py
```diff
@@ -51,6 +53,28 @@ def test_active_customers_get_queryset_returns_only_active(self):
...
+    def test_string_conversion(self, method):
+        """
+        Test ``UserDataSharingConsentAudit`` conversion to string
+        """
+        audit = UserDataSharingConsentAuditFactory()
+        expected_to_str = "<{class_name} for {user}: {state}>".format(
+            class_name=UserDataSharingConsentAudit.__name__,
+            user=audit.user,
+            state=audit.state
+        )
```
bradenmacdonald: Nit: This test code is almost identical to the implementation code. IMHO it would be better to construct an `UserDataSharingConsentAudit` instance with a known user name and then just compere the result of `method(audit)` to a hard-coded string like `"<UserDataSharingConsent for Bob: Not set>"`. Then it's also clear, for example, if `state` is the verbose version or the internal value.

tests/admin/test_forms.py
```diff
@@ -240,3 +255,25 @@ def test_raise_error_missing_get_edx_api_data(self, *args):  # pylint: disable=u
         with raises(NotConnectedToOpenEdX) as excinfo:
             get_all_catalogs(None)
         assert message == str(excinfo.value)
+
+    @mock.patch('enterprise.course_catalog_api.get_catalog_api_client')
+    @mock.patch('enterprise.course_catalog_api.CatalogIntegration')
+    @mock.patch('enterprise.course_catalog_api.get_edx_api_data')
+    def test_clean_valid_email(self, mocked_get_edx_data, *args):  # pylint: disable=unused-argument
+        """
+        Tests that valid emails are kept while cleaning email field
```
bradenmacdonald: The method name and docstring don't match the code.

test_utils/factories.py
```diff
@@ -71,6 +78,24 @@ class Meta(object):
...
+class UserDataSharingConsentAuditFactory(factory.django.DjangoModelFactory):
+    """
+    UserDataSharingConsentAuditFactory.
+
+    Creates an instance of UserDataSharingConsentAudit with minimal boilerplate.
+    """
+
+    class Meta(object):
+        """
+        Meta for UserDataSharingConsentAuditFactory.
```
bradenmacdonald: Totally optional side note: It is possible to disable the docstring requirement for classes named `Meta`. I find makework docstrings like this a tad annoying; they add no value and slow down refactoring / copy-paste coding.

enterprise/views.py
```diff
@@ -0,0 +1,184 @@
...
+        user = get_real_social_auth_object(request).user
+        ec_user, __ = EnterpriseCustomerUser.objects.get_or_create(
+            user_id=user.id,
+            enterprise_customer=customer,
+        )
+
+        UserDataSharingConsentAudit.objects.update_or_create(
+            user=ec_user,
+            defaults={
+                'state': 'enabled' if consent_provided else 'disabled'
```
bradenmacdonald: Use the constants like `UserDataSharingConsentAudit.ENABLED`, not the string values.

enterprise/views.py
```diff
@@ -0,0 +1,184 @@
...
+        request.session.pop('sso_quarantined_modules', None)
+
+    @staticmethod
+    def get_warning(provider, platform, required):
+        """
+        Get the appropriate warning for the form.
+        """
+        if required:
+            return _(
+                "Are you sure?\nIf you do not agree to share your data, you will have to use "
```
bradenmacdonald: Generally you shouldn't have explicit newlines in translated strings, especially if they're going into HTML anyways. I would just put the "Are you sure?" part into the template/js as data-warning-header or something.

enterprise/views.py
```diff
@@ -0,0 +1,184 @@
...
+            ).format(provider=provider)
+
+    @staticmethod
+    def get_note(provider, required):
+        """
+        Get the appropriate note for the form.
+        """
+        if required:
+            return _(
+                "*{provider} requires data sharing consent; if consent is not provided, you will"
```
bradenmacdonald: Why is the asterisk inside this string? It will be confusing to translators, and it is confusing to me. If it's supposed to be a bullet or something, add it later at the template level.

bradenmacdonald:
@haikuginger Please specify the git commit hash of the sandbox in your PR description and update that if/when you rebuild the sandbox.

tests/admin/test_forms.py
haikuginger: My bad; I used the previous form cleaning method as a template and forgot to change some things.

enterprise/views.py
haikuginger: Done.

enterprise/views.py
haikuginger: Given that the newline adds very little, I've removed it altogether. For translation purposes, the two phrases should be part of the same string.

enterprise/views.py
```diff
@@ -0,0 +1,184 @@
...
+    add_lookup = None
+
+try:
+    from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
+except ImportError:
+    configuration_helpers = None
+
+try:
+    from third_party_auth.pipeline import get_complete_url, get_real_social_auth_object
+except ImportError:
```
haikuginger: This package has extremely strict coverage requirements; putting them all in one try/except will fail those coverage requirements during testing, because it'll jump to the except clause after the first failure. We can get around it by patching the actual imports one at a time and only leaving the last one to fail, but that's not "real" coverage.

tests/test_views.py
haikuginger: Thanks for the idea! I agree that's a better way to go.

test_utils/factories.py
haikuginger: I would certainly be a fan of that. Can you point me in the right direction? I found a way to restrict linting based on block line length, but not on class name.

tests/test_models.py
haikuginger: Change made.

tests/test_views.py
haikuginger: Change made.

enterprise/views.py
haikuginger: Change made.

enterprise/views.py
haikuginger: In general, yes, but we need it to be accessible for testing purposes. I've revamped this set of things, though, so it's a single check at the beginning of the request.

haikuginger:
@bradenmacdonald, most of those changes have been made.

tox.ini
haikuginger: isort wants imports in try-except blocks in the last position, whereas pylint (more correctly, I think) keeps them in the standard stdlib-external-internal position. I may have found a way to flag this to get isort to ignore it, though.

test_utils/factories.py
bradenmacdonald: https://github.com/open-craft/opencraft/blob/abbe41f/pylintrc#L261-L263 is an example.

haikuginger:
Rebased from 8687bac.

tests/test_tpa_pipeline.py
```diff
@@ -6,29 +6,17 @@
...
-    verify_data_sharing_consent,
-    set_data_sharing_consent_record,
-)
+from django.http import HttpResponseRedirect
 
+from enterprise.models import EnterpriseCustomer, EnterpriseCustomerUser, UserDataSharingConsentAudit
+from enterprise.tpa_pipeline import (active_provider_enforces_data_sharing, active_provider_requests_data_sharing,
+                                     get_consent_status_for_pipeline, get_ec_for_running_pipeline,
+                                     get_enterprise_customer_for_request, get_enterprise_customer_for_sso,
+                                     set_data_sharing_consent_record, verify_data_sharing_consent)
```
bradenmacdonald: Nit: personally I think the earlier syntax (one imported symbol per line, comma after each one) is easier to read and produces nicer diffs. I see we are using this "cram everything over to the right and put up with ugly diffs" approach in various places in this repo though :)

enterprise/tpa_pipeline.py
```diff
@@ -0,0 +1,164 @@
...
+except ImportError:
+    Registry = None
+
+
+def get_enterprise_customer_for_request(request):
+    """
+    Get the EnterpriseCustomer associated with a particular request.
+    """
+    pipeline = request.session.get('partial_pipeline')
+    return get_ec_for_running_pipeline(pipeline)
```
bradenmacdonald: If `pipeline` is `None` (as in a blank session), this is causing a 500 error for me:

```python
  File "/edx/app/edxapp/edx-platform/common/djangoapps/util/enterprise_helpers.py", line 63, in insert_enterprise_fields
    add_data_sharing_consent_field(request, form_desc)
  File "/edx/app/edxapp/edx-platform/common/djangoapps/util/enterprise_helpers.py", line 71, in add_data_sharing_consent_field
    enterprise_customer = get_enterprise_customer_for_request(request)
  File "/edx/src/edx-enterprise/enterprise/tpa_pipeline.py", line 29, in get_enterprise_customer_for_request
    return get_ec_for_running_pipeline(pipeline)
  File "/edx/src/edx-enterprise/enterprise/tpa_pipeline.py", line 41, in get_ec_for_running_pipeline
    provider = Registry.get_from_pipeline(pipeline)
  File "/edx/app/edxapp/edx-platform/common/djangoapps/third_party_auth/provider.py", line 72, in get_from_pipeline
    if enabled.is_active_for_pipeline(running_pipeline):
  File "/edx/app/edxapp/edx-platform/common/djangoapps/third_party_auth/models.py", line 177, in is_active_for_pipeline
    return self.backend_name == pipeline['backend']
TypeError: 'NoneType' object has no attribute '__getitem__'
```

To reproduce: Open http://localhost:8000/login in an incognito window. Boom, 500 error.

bradenmacdonald:
@haikuginger Code is looking great - thanks for those change. I'm hitting a few 500 errors when testing this. Please let me know when it's not in progress, and I'll give it (and the other PR) a final spin.

enterprise/tpa_pipeline.py
haikuginger: Latest commit handles this. :)

tests/test_tpa_pipeline.py
haikuginger: I would agree; these imports are generated by isort. When it complains about import sorting in within a particular line, it does so with pretty much no additional information, so the easiest way to make it happy is to tell it to fix things itself.

Of course, now I actually look at it, it seems like it's just sorting the items alphabetically, so I can make that change manually from here on out, resulting in nice single-line imported items.

haikuginger:
@bradenmacdonald, I believe that the code is complete and ready for review. I've manually tested it out, and haven't come across any pain points. Red builds above are for coverage (forgot to look at the missing lines before pushing), which I am fixing right now.

haikuginger:
Also, FYI to everyone; the sandbox has been updated.

bradenmacdonald:
:+1:

- [x] I tested this: Tested on the sandbox and everything worked as described.
- [x] I read through the code
- [x] I checked for accessibility issues: I could complete the extended registration form including the consent field using a keyboard, and it follows the existing form patterns used for logistration.
- [x] Includes documentation: Admin fields have inline documentation, but the included rst documentation in the `docs` folder is yet not updated to explain data sharing consent. (@mattdrayer shouldn't we be doing that?)

I also notice that even when data sharing consent is granted, there is no way to tell that from the django admin. We should open a new ticket for "Display data sharing consent audit in the enterprise django admin" (strictly read only because for legal reasons we can't allow anyone other than the user to modify those).

Ready for second review :)

haikuginger:
@saleem-latif, take it away!

saleem-latif:
@haikuginger ChangeLog entry is missing for this change set.

haikuginger:
Squashed from f49e304.

haikuginger:
Rebased from caf13ab.

haikuginger:
@saleem-latif, that changelog entry has been added; translations have been added; commits have been squashed, and the build is green! Merging.

PR #14, opened 2016-11-23
ENT-32 added a helper method for EC branding details

asadiqbal08:
This PR is dependent on [ENT-76](https://openedx.atlassian.net/browse/ENT-76) and the changes suggested at [PR#9](https://github.com/edx/edx-enterprise/pull/9#discussion-diff-89055433R36)

@saleem-latif @mattdrayer

### DIFF
--- a/enterprise/api.py
+++ b/enterprise/api.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+"""
+Helper functions for enterprise app.
+"""
+
+from enterprise.models import EnterpriseCustomerBrandingConfiguration
+
+
+def get_enterprise_branding_info(provider_id=None):
+    """
+    Return the EnterpriseCustomer branding information.
+    """
+    return EnterpriseCustomerBrandingConfiguration.objects.filter(
+        enterprise_customer__identity_provider=provider_id
+    ).first()
--- a/test_utils/factories.py
+++ b/test_utils/factories.py
@@ -11,7 +11,8 @@ from faker import Factory as FakerFactory
 from django.contrib.auth.models import User
 from django.contrib.sites.models import Site
 
-from enterprise.models import EnterpriseCustomer, EnterpriseCustomerUser, PendingEnterpriseCustomerUser
+from enterprise.models import (EnterpriseCustomer, EnterpriseCustomerBrandingConfiguration, EnterpriseCustomerUser,
+                               PendingEnterpriseCustomerUser)
 
 FAKER = FakerFactory.create()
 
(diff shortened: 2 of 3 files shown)


mattdrayer:
Is this `helpers.py` module intended to be the Python programming interface for Django projects which import the edx-enterprise app?  If so we should rename the file to `api.py`.  If it intended for internal use only we might instead move the method into `utils.py`.

enterprise/helpers.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+"""
+Helper functions for enterprise app.
+"""
+
+from .models import EnterpriseCustomerBrandingConfiguration
```
zubair-arbi: can we please do absolute imports instead of relative imports?

enterprise/helpers.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def get_branding_info(provider_id=None):
+    """
+    Return the EnterpriseCustomer branding information.
+    """
+    info = None
+    try:
+        info = EnterpriseCustomerBrandingConfiguration.objects.filter(
```
zubair-arbi: You can achieve this functionality by using the django's `first` method.
https://docs.djangoproject.com/en/1.10/ref/models/querysets/#first

enterprise/helpers.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+"""
+Helper functions for enterprise app.
+"""
+
+from .models import EnterpriseCustomerBrandingConfiguration
+
+
+def get_branding_info(provider_id=None):
```
zubair-arbi: Since this util method is for enterprise only so can we name it to something like `get_enterprise_branding_config`

enterprise/helpers.py
bradenmacdonald: Why? (Relative imports are IMHO easier to read and require fewer changes when packages are renamed/moved. I believe the edX style guide allows either absolute or relative imports.)

zubair-arbi: @bradenmacdonald First of all, You are right about the advantage of relative imports when we rename/move the package but I think that this doesn't happen that often. The only thing I would like is that we have a common pattern of import that would contribute in our imports order and readability. If you see in other places in enterprise app then you will see absolute imports being used already, e.g. `edx-enterprise/enterprise/admin/forms.py` also in [django docs it is usually the absolute imports shown in examples](https://docs.djangoproject.com/en/1.10/intro/tutorial02/#playing-with-the-api). As long as we have consensus on whether we use [relative imports or absolute imports](http://stackoverflow.com/questions/4209641/absolute-vs-explicit-relative-import-of-python-module) I don't have any issue :)

enterprise/helpers.py
asadiqbal08: I will prefer `get` instead of fetch--first(). As due to dependency this PR is in WIP state.

enterprise/helpers.py
bradenmacdonald: @zubair-arbi Ok, well if people want to go for absolute imports we should turn on a pylint rule to enforce it; otherwise some people won't know or will forget.

asadiqbal08:
closing this in favor of [PR#18](https://github.com/edx/edx-enterprise/pull/18)

PR #15, opened 2016-11-23
Bulk upload users

e-kolpakov:
**Description:** This PR adds ability to bulk link users to enterprise customer.

**JIRA:** [ENT-62](https://openedx.atlassian.net/browse/ENT-62)

**Dependencies:** cross dependent on https://github.com/edx/edx-platform/pull/14003

**Testing instructions:**

1. Linking users by username or email was significantly affected - rerun #8 tests.
2. Prepare CSV file (*Yes, so far there are just one column, so it is more like "one record per line", but CSV is a superset of it anyway*).
2.1. First line should contain column header `email_or_username`
2.2. Each subsequent line should contain existing user username or an email.
2.3. Don't do any intentional mistakes to test how it will behave yet - we'll get to it later. Make sure the list does not contain users already linked to ECs (even the current one) or invalid emails.
3. Go to "Manage Learners" view in Django admin (admin -> Enterprise Customers -> choose one -> "Manage Learners" in top-right corner)
4. Test "Manage Learners" form validation:
4.1. Submit empty form - should complain that either username/email or bulk upload CSV must be present.
4.2. Submit both username/email and bulk upload CSV - should complain about both being present.
4.3. Submit invalid email only - should complain that email is invalid.
5. Try uploading the file created in step 2. Existing users should be linked to EC, unknown emails should end up in Pending EC Users
6. Try uploading the same file again (yes, don't modify the file, don't unlink anything). *Expected result:* Form explodes with a ton of errors. Errors should make sense so that non-technical person could understand what needs to be fixed.
7. Unlink (some) records, change the CSV (introduce invalid emails, already linked users, already existing pending links, etc.). Upload the file again. *Expected result:* form errors are shown, errors still make sense. Note that "valid" records are still supposed to be added to corresponding lists: valid emails with not used in pending link - in Pending Linked Users and usernames of existing users not linked to any EC - in Linked users.

**Reviewers:**
- [x] @bradenmacdonald 
- [x] @asadiqbal08 and/or @saleem-latif 

**Merge checklist:**
- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Translations updated

**Post merge:**
- [ ] Create a tag


e-kolpakov:
@bradenmacdonald @saleem-latif @asadiqbal08 heads up - this is more or less production-ready bulk upload. It lacks tests for new functionality (~86% diff coverage), so I believe it is safe to expect this PR to be ready-ready by tomorrow EOD. Meanwhile, it would be very helpful if you could review the general approach (as in past your valuable input influenced the overall approach and required some extra time to implement).

CC @mattdrayer

CHANGELOG.rst
```diff
@@ -14,6 +14,19 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[0.4.0] - 2016-11-23
```
mattdrayer: @haikuginger beat you to this one 😄 -- it will have to be v0.5.0 or higher by the time this branch merges

enterprise/__init__.py
```diff
@@ -4,6 +4,6 @@
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.3.1"
+__version__ = "0.4.0"
```
mattdrayer: Just a note: Will need to increment to unused version prior to merging

enterprise/admin/forms.py
```diff
@@ -5,52 +5,82 @@
...
+        USERNAME_OR_EMAIL = "username_or_email"
+        BULK_UPLOAD = "bulk_upload_csv"
+        MODE = "mode"
+
+    def clean_username_or_email(self):
         """
         Clean email form field
 
         Returns:
             str: email to link
```
mattdrayer: ...or username?

enterprise/admin/forms.py
```diff
@@ -5,52 +5,82 @@
...
         return email
 
+    def clean(self):
+        """
+        Clean fields that depend on each other.
+
+        In this case, the form can be used to link single user or bulk link multiple users. These are mutually
+        exclusive modes, so this method checks that only one field is passed.
+        """
+        cleaned_data = super(ManageLearnersForm, self).clean()
```
mattdrayer: Does it make sense to move `cleaned_data` closer to where it is being used -- ie near line 80?

enterprise/admin/utils.py
```diff
@@ -4,10 +4,102 @@
...
+def username_or_email__to__email(email_or_username):
+    """
+    Convert email_or_username to email.
+
+    Returns:
+        str: If `email_or_username` was a username returns user's email, otherwise assumes it was an email and returns
+             as is.
+    """
+    try:
+        user = User.objects.get(username=email_or_username)
```
mattdrayer: I suppose there's a weird case here where the username is also an email address, but it's not the same email address as the value actually stored in user.email.  I'm not sure if it matters but thought I would note it.

enterprise/templates/enterprise/admin/manage_learners.html
```diff
@@ -101,9 +101,9 @@
 
 <div class="forms-panel">
   <h1>{% trans "Link learner" %}</h1>
-  <form action="" method="post">
+  <form action="" method="post" enctype="multipart/form-data">
     {% csrf_token %}
-    {{ manage_learners_form }}
+    {{ manage_learners_form.as_p }}
```
mattdrayer: sorry, what is "p" here?

enterprise/utils.py
```diff
@@ -4,6 +4,11 @@
 """
 from __future__ import absolute_import, unicode_literals
 
+from functools import wraps
+
+
+USER_POST_SAVE_DISPATCH_UID = "user_post_save_upgrade_pending_enterprise_customer_user"
```
mattdrayer: Should we have a `constants.py` module to contain things like this?

enterprise/admin/views.py
```diff
@@ -94,19 +96,90 @@ def post(self, request, customer_uuid):
         Returns:
             django.http.response.HttpResponse: HttpResponse
         """
-        # TODO: bulk upload
         enterprise_customer = EnterpriseCustomer.objects.get(uuid=customer_uuid)  # pylint: disable=no-member
-        manage_learners_form = ManageLearnersForm(request.POST)
+        manage_learners_form = ManageLearnersForm(request.POST, request.FILES)
+
+        # initial form validation - check that form dat is well-formed
```
asadiqbal08: nit: `dat` ~ `data`

enterprise/admin/utils.py
```diff
@@ -4,10 +4,102 @@
...
+    for idx, row in enumerate(reader):
+        if expected_columns:
+            if not set(row.keys()) >= expected_columns:
+                raise ValidationError(ValidationMessages.MISSING_EXPECRED_COLUMNS.format(
+                    row_number=idx, expected_columns=", ".join(expected_columns), actual_columns=", ".join(row.keys())
+                ))
+        yield row
+
+
+def username_or_email__to__email(email_or_username):
```
asadiqbal08: can we keep the name of passing argument consistent with method ? e.g. `username_or_email`

enterprise/admin/views.py
```diff
@@ -94,19 +96,90 @@ def post(self, request, customer_uuid):
...
+        # if something went wrong - display bound form on the page
+        context = self._build_context(request, customer_uuid)
         context.update({self.ContextParameters.MANAGE_LEARNERS_FORM: manage_learners_form})
-
         return render(request, self.template, context)
 
+    @classmethod
+    def _handle_form(cls, enterprise_customer, manage_learners_form):
+        """
+        Handle ManageLearnersForm submission.
```
asadiqbal08: Can we add somewhat more detail what is going on inside `_handle_form` ?

enterprise/admin/utils.py
e-kolpakov: @mattdrayer At this moment we don't know if it was an email or username. So we just query user by username in hopes it was a username - if it was found than either (1) admin passed username into form or (2) someone used valid email as username (which is not supported I believe, but just in case).

If it was not found than it is either an email, or unknown username.

enterprise/templates/enterprise/admin/manage_learners.html
e-kolpakov: @mattdrayer "as paragraphs". Without this, it just dumped all the inputs without any structure, resulting in "whatever layout browser thinks correct". In this particular case, "Bulk CSV upload" label started on the same line as "Username or Email" input. With `as_p` it wraps each field into a paragraph (both input field and label).

bradenmacdonald:
> 2.1. First line should contain column header email_or_username

I think we only need to support email addresses in the CSV, per the ENT-62 ticket. It's unlikely an enterprise would know their user's edX usernames anyways; I think the main use case here will be users who aren't yet registered. So (unless someone else objects) I think we should drop support for usernames in the CSV. (It's good to support usernames for the individual linking but IMHO supporting it for the CSV is more trouble than it's worth, makes the code complex, and adds little value.) Or did I miss some requirement somewhere?

Even if we have to keep that functionality for some reason, naming the column "email_or_username" within the CSV itself is going to be annoying for admin users. If we need to keep both, we should require that the heading row be either "email" or "username", which are column headers that are much more likely to already be present in any real-world CSV file that someone might be sent for use.

bradenmacdonald:
A couple comments about the UI:

screen shot 2016-11-24 at 11 03 27 pm

1. It's not obvious that users should enter either a-username-or-an-email *or* upload a CSV. I think the word "or" should be included, e.g. "Or, upload a CSV:"
1. If there are any constraints on how the CSV should be formatted (which there obviously must be), it should be indicated in the UI: something like "Or, upload a CSV to enroll multiple users at once: [Choose file] (The CSV must have a column of email addresses, indicated by the heading 'email' in the first row.)"

e-kolpakov:
@bradenmacdonald  removed bulk linking by username and improved the form. Now it looks like this:

image

Please take another look (and since I'm away next week feel free to address any remaining notes).

enterprise/admin/forms.py
bradenmacdonald: I think it's first since it calls the superclass method, and generally calls to `super()` go first. However, in this case, the superclass method is just `def clean(self): return self.cleaned_data` so it doesn't really matter.

bradenmacdonald:
Very nice, @e-kolpakov! It works like a charm :)

Note that @e-kolpakov is going to be off next week so I will take over this PR.

I noticed that it was possible to trigger a 500 error if the CSV contains duplicate emails. I also think that we should ignore existing emails when uploading a CSV rather than failing completely, so we can support a workflow where the admin user gets a "revised" CSV from the EC with a few more emails added throughout, and can just upload that again rather than having to diff it and isolate the new emails. So I have pushed those changes, and also added a success message, which was missing.

screen shot 2016-11-25 at 7 18 23 pm

enterprise/admin/views.py
```diff
@@ -94,19 +97,113 @@ def post(self, request, customer_uuid):
...
+            manage_learners_form (ManageLearnersForm): bound ManageLearners form instance
+            request (django.http.request.HttpRequest): HTTP Request instance
+        """
+        errors = []
+        emails = set()
+        already_linked_emails = []
+        duplicate_emails = []
+        csv_file = manage_learners_form.cleaned_data[ManageLearnersForm.Fields.BULK_UPLOAD]
+        try:
+            parsed_csv = parse_csv(csv_file, expected_columns={ManageLearnersForm.CsvColumns.EMAIL})
```
e-kolpakov: @bradenmacdonald nit: I've got a slight concern that this method is now ~60 LOC - might be good to extract this part into a dedicated method.

bradenmacdonald:
@asadiqbal08 @saleem-latif @mattdrayer Can we please get a second review here?

I am 👍 for all the code except the small part I wrote myself :)

bradenmacdonald:
@asadiqbal08 @saleem-latif @mattdrayer Friendly reminder to review this please. I'll squash + rebase once it's had a second look.

enterprise/__init__.py
```diff
@@ -4,6 +4,6 @@
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.4.1"
+__version__ = "0.5.0"
```
asadiqbal08: `0.5.0` is the current version in [edx-enterprise repo](https://github.com/edx/edx-enterprise/blob/master/enterprise/__init__.py#L7), any change expected here ?

bradenmacdonald: Yes, I will rebase+squash this and bump the version number once it's approved.

enterprise/admin/forms.py
```diff
@@ -5,53 +5,94 @@
...
-
-        existing_record = EnterpriseCustomerUser.objects.get_link_by_email(email)
-        if existing_record:
-            message = _("User with email {email} is already registered with Enterprise Customer {ec_name}").format(
-                email=email, ec_name=existing_record.enterprise_customer.name
-            )
-            raise ValidationError(message)
+        email_or_username = self.cleaned_data[self.Fields.EMAIL_OR_USERNAME].strip()
+
+        if not email_or_username:
```
asadiqbal08: optional: It would be nice, If we can add a single line comment to describe why we are returning `email_or_username`. It would be helpful one sight understanding.

enterprise/templates/enterprise/admin/manage_learners.html
```diff
@@ -100,10 +100,10 @@
 </div>
 
 <div class="forms-panel">
-  <h1>{% trans "Link learner" %}</h1>
-  <form action="" method="post">
+  <h1>{% trans "Link learners" %}</h1>
+  <form action="" method="post" enctype="multipart/form-data">
     {% csrf_token %}
-    {{ manage_learners_form }}
+    {{ manage_learners_form.as_p }}
```
asadiqbal08: optional: we can add a line comment here: `as_p will render them wrapped in <p> tags`

asadiqbal08:
Overall LGTM

enterprise/__init__.py
smarnach: @bradenmacdonald I've rebased it to be able to work on #19 and changed the version number to 0.6.0.

enterprise/admin/utils.py
smarnach: @mattdrayer Email addresses are [not valid usernames on edx-platform](https://github.com/edx/edx-platform/blob/master/common/djangoapps/student/forms.py#L123-L124).  Since this requirement is enforced by the registration form rather than the model, it's certainly possible to circumvent it using managment commands, but I agree with your assessment that this corner case does not matter. :)

bradenmacdonald:
@smarnach  Thanks.

@asadiqbal08 "LGTM" - does that mean you give your +1 for this PR? Or still want to wait for some last piece?

enterprise/templates/enterprise/admin/manage_learners.html
bradenmacdonald: Sure, I can do that, though it is a standard django function.

asadiqbal08:
@bradenmacdonald LGTM 👍

bradenmacdonald:
Thanks @asadiqbal08. I added the two comments you requested, updated translations, squashed this, and will now merge.

PR #16, opened 2016-11-24
ENT-76: Make sure IdP and EC have OneToOne Relation

saleem-latif:
@mattdrayer , @e-kolpakov , @bradenmacdonald , @asadiqbal08 -- FYI

Kindly review this PR, it adds a `EnterpriseCustomerIdentityProvider` model and corresponding admin interface. 

JIRA: [ENT-76](https://openedx.atlassian.net/browse/ENT-76)

__Description:__
Currently, identity provider is a Slug field which can be duplicated across enterprise customers, which presents a problem for LMS apps that need to access enterprise customer for a given identity provider.
We need to make sure that enterprise customer and identity provider has a one to one relation.

__Testing Instructions:__
1. Go to SAML Provider Config admin page at "admin/third_party_auth/samlproviderconfig/" and add a couple of new identity providers. (make sure that the site selected for identity provider is same as the one associated with your current domain)
1. Go to enterprise customer admin portal at "/admin/enterprise/enterprisecustomer/"
    - Add a new enterprise customer without selecting any identity provider from the drop down. An enterprise customer should be added in the system, make sure that new enterprise customer does not have any associated identity provider.
    - Try adding another enterprise customer, this time select an identity provider from the dropdown  An enterprise customer should be added in the system, make sure that new enterprise customer have correct associated identity provider.
    -  Try adding another enterprise customer, select a site that is not the same as identity provider's site (this is the site that you selected in step 1). There should be an error message explaining site mismatch between enterprise customer and identity provider.
    - Try adding another Enterprise Customer with an identity provider that is already associated with another enterprise customer. There should be an error message saying that the selected identity provider is already associated with another enterprise customer.
    - Try removing an associated identity provider from an existing enterprise customer, you can do that by opening enterprise customer details view and selecting `delete` checkbox on the right corner of "ENTERPRISE CUSTOMER IDENTITY PROVIDERS" inline form and then hitting save.


enterprise/models.py
```diff
@@ -76,9 +77,20 @@ class Meta(object):
     catalog = models.PositiveIntegerField(null=True, help_text=_("Course catalog for the Enterprise Customer."))
     active = models.BooleanField(default=True)
     history = HistoricalRecords()
-    identity_provider = models.SlugField(null=True, blank=True, default=None)
     site = models.ForeignKey(Site, related_name="enterprise_customers")
 
+    @property
+    def identity_provider(self):
+        """
+        Positive integer id for the identity provider associated with this enterprise customer.
```
bradenmacdonald: The provider ID is a string, not a positive integer.

enterprise/models.py
```diff
@@ -322,6 +334,56 @@ def __repr__(self):
...
+
+    There should be a link between an enterprise customer and its Identity Provider. This relationship has
+    following constraints
+        1. An enterprise customer may or may not have an identity provider.
+        2. An enterprise customer can not have more than one identity providers.
+        3. Enterprise customer site should match with identity provider's site. (i.e. same domain names)
+
+    Fields:
+        enterprise_customer (ForeignKey[EnterpriseCustomer]): enterprise customer
+        provider_id (:class:`django.db.models.PositiveIntegerField`): positive integer id of identity provider.
```
bradenmacdonald: Same here - it's a `SlugField`, not a `PositiveIntegerField`. And I think the description should be "The provider_id string of the identity provider".

enterprise/models.py
```diff
@@ -322,6 +334,56 @@ def __repr__(self):
...
+        null=False,
+        related_name="enterprise_customer_identity_provider"
+    )
+    provider_id = models.SlugField(
+        null=False,
+        blank=False,
+        unique=True,
+        help_text="Slug field containing a unique identifier for the identity provider.",
+    )
+    provider_name = models.CharField(
```
bradenmacdonald: Since `EnterpriseCustomer` already has a `name` field, and the providers each have a `name` stored in their `ProviderConfig` subclass, why do we need this additional `provider_name` field? Can you just use

```
@property
def provider_name(self):
    return get_identity_provider(self.provider_id).name
```
instead?

bradenmacdonald:
Looking good, @saleem-latif! Thanks for this :)

saleem-latif:
@mattdrayer , @e-kolpakov , @bradenmacdonald , @asadiqbal08 I have finalized the changes, kindly take a look

enterprise/admin/__init__.py
```diff
@@ -75,6 +90,14 @@ def logo(instance):
             return instance.branding_configuration.logo
         return None
 
+    @staticmethod
+    def identity_provider(instance):
+        """
+        Instance is EnterpriseCustomer.
+        """
+        ecip = instance.enterprise_customer_identity_provider
```
asadiqbal08: nit: can we keep it like `ec_identity_provider / ec_idp` or just `identity_provider` ?

enterprise/admin/__init__.py
```diff
@@ -75,6 +90,14 @@ def logo(instance):
...
+        return ecip and ecip.provider_name or ecip.provider_id
```
asadiqbal08: nit: can we add some more detail in doc-string about that condition ?

enterprise/admin/forms.py
```diff
@@ -87,3 +88,66 @@ def get_catalog_options(self):
...
+    """
+    class Meta:
+        model = EnterpriseCustomerIdentityProvider
+        fields = "__all__"
+
+    def __init__(self, *args, **kwargs):
+        """
+        Initialize the form.
+
+        Substitute a ChoiceField in for the catalog field that would
```
asadiqbal08: > ... a ChoiceField for the catalog field

enterprise/models.py
```diff
@@ -73,9 +75,20 @@ class Meta(object):
...
 
+    @property
+    def identity_provider(self):
+        """
+        Unique slug for the identity provider associated with this enterprise customer.
+
+        Returns `None` if enterprise customer does not have any identity provider.
+        """
+        try:
+            return self.enterprise_customer_identity_provider and self.enterprise_customer_identity_provider.provider_id
```
asadiqbal08: Can you explain why are you checking for `provider_name` [THERE](https://github.com/edx/edx-enterprise/pull/16/files#diff-61ac02e9b82173a27f3bbf8a350dd5e2R99) along with `provider_id`

asadiqbal08:
overall LGTM

enterprise/models.py
saleem-latif: @asadiqbal08 function `identity_provider` [here](https://github.com/edx/edx-enterprise/pull/16/files#diff-61ac02e9b82173a27f3bbf8a350dd5e2R99) is responsible for returning a string to be displayed on admin portal. We fallback to provider_id to cover the cases when `third_party_auth` app will not be present.

e-kolpakov:
@saleem-latif I'm on vacation, so don't wait for me to approve this :)

bradenmacdonald:
Looks good! The only thing that is confusing is that once you set a provider, selecting the None/"-----" option will cause an error, and you have to use the tiny "Delete" checkbox on the right side of the screen to clear a provider setting.
screen shot 2016-11-28 at 4 44 40 pm


Also for some reason, the SAML providers were not showing up for me until I made a fix in the `edx-platform` code; I had to change the `Registry.enabled()` code in `third_party_auth/provider.py` from
```python
        if SAMLConfiguration.is_enabled(Site.objects.get_current(get_current_request())):
```
to
```python
        if SAMLConfiguration.is_enabled(Site.objects.get_current(get_current_request()).pk):
```
Is something wrong with my devstack, or is that a fix we need to make?

saleem-latif:
@bradenmacdonald 

> Looks good! The only thing that is confusing is that once you set a provider, selecting the None/"-----" option will cause an error, and you have to use the tiny "Delete" checkbox on the right side of the screen to clear a provider setting.

This is the expected behavior, This is because of the reason that you can not set `provider_id` to null so you must select the delete checkbox if you want to remove the provider id.

> Also for some reason, the SAML providers were not showing up for me until I made a fix in the edx-platform code; I had to change the Registry.enabled() code in third_party_auth/provider.py from

Could this have been caused by django cache [[ref](https://github.com/edx/django-config-models/blob/master/config_models/models.py#L117-L149)] ? Can you try again after clearing the cache and see if the problem is resolved ?

mattdrayer:
I'm good with this change -- 👍 -- I think we should merge and fix any issues going forward.

asadiqbal08:
@saleem-latif  kindly squash the commits before merge

PR #20, opened 2016-12-08
ENR-32/83 Return branding information on the basis of EC uuid.

asadiqbal08:
@mattdrayer @zubair-arbi Can you guys take a look ? 
[ENT-32](https://openedx.atlassian.net/browse/ENT-32) & [ENT-83](https://openedx.atlassian.net/browse/ENT-83)

### DIFF
--- a/enterprise/api.py
+++ b/enterprise/api.py
@@ -7,10 +7,21 @@ from __future__ import absolute_import, unicode_literals
 from enterprise.models import EnterpriseCustomerBrandingConfiguration
 
 
-def get_enterprise_branding_info(provider_id=None):
+def get_enterprise_branding_info_by_provider_id(provider_id=None):  # pylint: disable=invalid-name
     """
-    Return the EnterpriseCustomer branding information.
+    Return the EnterpriseCustomer branding information based on provider_id.
+
+    Provider_id: There is 1:1 relation b/w EnterpriseCustomer and Identity provider.
     """
     return EnterpriseCustomerBrandingConfiguration.objects.filter(
         enterprise_customer__enterprise_customer_identity_provider__provider_id=provider_id
+    ).first()
+
+
+def get_enterprise_branding_info_by_ec_uuid(ec_uuid=None):  # pylint: disable=invalid-name
+    """
+    Return the EnterpriseCustomer branding information based on enterprise customer uuid.
+    """
+    return EnterpriseCustomerBrandingConfiguration.objects.filter(
+        enterprise_customer__uuid=ec_uuid
     ).first()
--- a/enterprise/apps.py
+++ b/enterprise/apps.py
@@ -6,6 +6,7 @@ Enterprise Django application initialization.
 from __future__ import absolute_import, unicode_literals
 
 from django.apps import AppConfig, apps
+from django.conf import settings
 
 from enterprise.constants import USER_POST_SAVE_DISPATCH_UID
 
@@ -17,7 +18,7 @@ class EnterpriseConfig(AppConfig):
 
     name = "enterprise"
     valid_extensions = [".png", ]
-    image_size = 250 * 1024
+    image_size = getattr(settings, 'ENTERPRISE_CUSTOMER_LOGO_IMAGE_SIZE', 4 * 1024)
 
     @property
     def auth_user_model(self):
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,14 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[0.8.0] - 2016-12-08
+~~~~~~~~~~~~~~~~~~~~
+
+Added
+-----
+
+* added the branding information api methods to return the enterprise customer logo on the basis of provider_id or uuid.
+* Updated the logo image validator to take an image of size maximum of 4kb.
 
 [0.7.0] - 2016-12-07
 ~~~~~~~~~~~~~~~~~~~~
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.7.0"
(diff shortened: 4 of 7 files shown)


enterprise/api.py
```diff
@@ -7,10 +7,21 @@
 from enterprise.models import EnterpriseCustomerBrandingConfiguration
 
 
-def get_enterprise_branding_info(provider_id=None):
+def get_enterprise_branding_info(**kwargs):
```
mattdrayer: Do we really need to open this signature up to such a broad set of use cases?  It seems like below we are only concerned with two parameters -- we should be explicit about the arguments.

enterprise/api.py
```diff
@@ -7,10 +7,21 @@
...
     """
-    return EnterpriseCustomerBrandingConfiguration.objects.filter(
-        enterprise_customer__enterprise_customer_identity_provider__provider_id=provider_id
-    ).first()
+    provider_id = kwargs.get('provider_id', None)
+    enterprise_customer_uuid = kwargs.get('ec_src', None)
+    if provider_id:
+        return EnterpriseCustomerBrandingConfiguration.objects.filter(
+            enterprise_customer__enterprise_customer_identity_provider__provider_id=provider_id
+        ).first()
```
mattdrayer: will `.first()` fail if no matches are returned?

enterprise/apps.py
```diff
@@ -17,7 +17,7 @@ class EnterpriseConfig(AppConfig):
 
     name = "enterprise"
     valid_extensions = [".png", ]
-    image_size = 250 * 1024
+    image_size = 4 * 1024
```
mattdrayer: We might want to make this a configurable value for other Open edX operators to manage.

enterprise/api.py
asadiqbal08: I was thinking the same. Ok I will update it to take two input parameter.

enterprise/apps.py
asadiqbal08: It seems me, logo should be enough to have 4kb in size ( not need to change frequently). However we need to scale that logo on client side that will be happened on my other [PR](https://github.com/edx/edx-platform/pull/14064)

mattdrayer:
👍

asadiqbal08:
@haikuginger can you review the change set ? Wana to merge in upcoming day.

haikuginger:
@asadiqbal08, will review today.

enterprise/api.py
```diff
@@ -7,10 +7,19 @@
 from enterprise.models import EnterpriseCustomerBrandingConfiguration
 
 
-def get_enterprise_branding_info(provider_id=None):
+def get_enterprise_branding_info(provider_id=None, ec_uuid=None):
```
asadiqbal08: We can also split it into two separate methods in api.
1 - `get_enterprise_branding_info_by_provider_id(provider_id=None)`
2 - `get_enterprise_branding_info_by_ec_uuid(ec_uuid=None)`

mattdrayer: I like this idea -- it's good to have very specific functions representing these pieces of system behavior with their own test coverage.

asadiqbal08:
Green Build. 
any yellow thumbs up :)) ? @mattdrayer @haikuginger ^^

enterprise/apps.py
haikuginger: I kind of agree with @mattdrayer's initial assessment - 4kB is a bit small of a limit to not be configurable. I'd rather have this pull from the Django settings but have it default to 4kB.

enterprise/api.py
```diff
@@ -7,10 +7,21 @@
 from enterprise.models import EnterpriseCustomerBrandingConfiguration
 
 
-def get_enterprise_branding_info(provider_id=None):
+def get_enterprise_branding_info_by_provider_id(provider_id=None):  # pylint: disable=invalid-name
     """
-    Return the EnterpriseCustomer branding information.
+    Return the EnterpriseCustomer branding information based on enterprise customer provider_id.
+
+    Provider_id: There is 1:1 relation b/w EnterpriseCustomer and SAML Identity provider.
```
haikuginger: We can attach to any SSO provider, not just SAML.

enterprise/apps.py
asadiqbal08: what is actual required here ? can you give an example ?

enterprise/api.py
saleem-latif: @asadiqbal08 behavior was updated [here](https://github.com/edx/edx-enterprise/pull/16/files#diff-ce3a6d48e7f42961c07cd219f505d179R27), now we can attach any SSO provider.

asadiqbal08:
@haikuginger kindly take a look.

enterprise/api.py
```diff
@@ -7,10 +7,21 @@
 from enterprise.models import EnterpriseCustomerBrandingConfiguration
 
 
-def get_enterprise_branding_info(provider_id=None):
+def get_enterprise_branding_info_by_provider_id(provider_id=None):  # pylint: disable=invalid-name
     """
-    Return the EnterpriseCustomer branding information.
+    Return the EnterpriseCustomer branding information based on enterprise customer provider_id.
```
haikuginger: The provider ID is for the SSO provider, not for the enterprise customer.

enterprise/apps.py
```diff
@@ -17,7 +18,7 @@ class EnterpriseConfig(AppConfig):
 
     name = "enterprise"
     valid_extensions = [".png", ]
-    image_size = 250 * 1024
+    image_size = getattr(settings, 'EC_LOGO_IMAGE_SIZE', 4 * 1024)
```
haikuginger: Nit: `ENTERPRISE_CUSTOMER_LOGO_IMAGE_SIZE`: we're not short on bytes or line space; there's no need to abbreviate. This is going to be a variable that's going to be mixed in with variables for every other  app; we need it to be distinctive.

asadiqbal08:
Done

PR #21, opened 2016-12-08
Discovery work for enterprise entitlements

zubair-arbi:
This is a **sample code** for the prototype of enterprise entitlements.
Please ignore the print statements or commented code.
ENT-84
@asadiqbal08 @saleem-latif @mattdrayer @kashifch 

* add `enterprise customer entitlement` model
* add django admin for `enterprise customer entitlement` model
* add inline django admin for `enterprise customer entitlement` model inside `enterprise customer` model
* add dynamic `Coupon URL` for viewing the coupon details on `Ecom`

**Related Ecom PR for enterprise entitlements checkout flow:** https://github.com/edx/ecommerce/pull/1044

### Not Added (pending)
Part of the actual implementation would a public Enterprise API `/api/enterprise/v1/`. This API will take the `username` and `site domain` through a POST request and would generate a json response like:
```js
{
    'username': 'zubair',
    'enterprise': {
        'name': 'Arbisoft',
        'catalog': 1,
        'site_domain': 'example.com',
        'enable_data_sharing_consent': true,
        'enforce_data_sharing_consent': 'At Enrollment',
        'logo': 'https://www.edx.org/sites/default/files/homepage/banner/promo/bkg/onethird-539x290-csedweek.jpg',
        'user_data_sharing_consent': false
    }
}
```

### Django admin screen shots for enterprise entitlements
screen shot 2016-12-08 at 5 17 59 pm
screen shot 2016-12-08 at 5 18 11 pm
screen shot 2016-12-08 at 5 19 04 pm


mattdrayer:
Thanks, @zubair-arbi -- what do you see as the next step for work on this prototype?  Also, regarding the API you mentioned -- this is for internal client-service or service-service communication, and not communication with an external EnterpriseCustomer system, correct?

saleem-latif:
@zubair-arbi entitlements would eventually expire right ? if coupons for a particular entitlements are all used. 
There should be a way to check if an entitlement is expired in `EnterpriseCustomerEntitlement` model. May be a boolean field `expired` or `active`, or a date-time field `expires` (containing expiration date time stamp)

zubair-arbi:
@mattdrayer I think for next step we can have a consensus on the API of the enterprise app and have a sense of how the enterprise checkout UI will look like.
For the enterprise API you are right about it being internal service(enterprise)-service(Ecom) communication.

mattdrayer:
Okay, that sounds like a good plan -- we are working with UX on the checkout UI so hopefully we'll make some progress on this over the next couple of days.  @brittneyexline has also joined in the fun and will be getting involved on the Cambridge-side for Seat Entitlements starting with the next Sprint (it is Z!!!)

Regarding the API schema -- I'm not sure we're going to be able to preemptively determine all of the various endpoints that will be needed.  However we can begin with some basic GET URLs for the models which will allow LMS + Otto to request information from the Enterprise app, and grow the API as-needed.

mattdrayer:
@zubair-arbi if this proto-PR is no longer necessary we can close it

zubair-arbi:
@mattdrayer thanks for the reminder, I am going to close it.

PR #22, opened 2016-12-15
[ENT-96] Add details to EnterpriseCustomerUser; link from Manage Learners

haikuginger:
This pull request enhances the EnterpriseCustomerUser admin panel to contain the following additional information:
* Username
* email address
* Added date
* Data Sharing Consent status/history

In addition, the Manage Learners template has been enhanced to link directly to each EnterpriseCustomerUser.

**JIRA tickets**: Implements ENT-96

**Dependencies**: None

**Screenshots**:

screen shot 2016-12-15 at 4 57 02 pm

screen shot 2016-12-15 at 4 57 29 pm

**Sandbox URL**: https://ent96.opencraft.hosting/ (direct IP: http://213.32.75.54; use `staff` account for superuser access)

**Merge deadline**: Jan. 9 2016

**Testing instructions**:

1. In the sandbox, go to the Manage Learner view for the existing EnterpriseCustomer
2. Observe that the username for each linked user is now a link, rather than plain text
3. Click the link, and observe that you're taken to the EnterpriseCustomerUser admin
4. Observe that user details are visible, but not editable, on the admin.
5. Observe that historical UserDataSharingConsentAudit items are listed below the user details.

1. Note that at present, a "status" item is not shown; this is because all EnterpriseCustomerUsers are linked to EnterpriseCustomers.

**Reviewers**
- [ ] @itsjeyd 
- [ ] edX reviewer[s] TBD

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=22"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

bradenmacdonald:
@haikuginger I think pending users should be clickable too, which is why the ticket is framed in terms of a profile view that includes "Status (Linked, Pending)". Obviously username and data sharing consent status would be empty for pending users.

smarnach:
@bradenmacdonald That was my understanding of the ticket as well.  However, it would require to introduce a new view that can show either `EnterpriseCustomerUser` instances or `PendingEnterpriseCustomerUser` instances.  The additional value provided by that view seems small in comparision to the amount of code necessary to implement this.  We should check with edX before implementing this.

enterprise/admin/views.py
```diff
@@ -62,7 +62,7 @@ def _build_context(self, request, customer_uuid):
         """
         # TODO: pylint acts stupid - find a way around it without suppressing
         enterprise_customer = EnterpriseCustomer.objects.get(uuid=customer_uuid)  # pylint: disable=no-member
-        linked_learners = EnterpriseCustomerUser.objects.filter(enterprise_customer__uuid=enterprise_customer.uuid)
+        linked_learners = EnterpriseCustomerUser.objects.filter(enterprise_customer=enterprise_customer)
```
smarnach: Just out of curiosity – does this change make any difference?  The new code is more concise, and might even result in more efficient SQL, but it should do exactly the same thing, right?

enterprise/admin/__init__.py
```diff
@@ -15,9 +15,9 @@
 from enterprise.admin.utils import UrlNames
 from enterprise.admin.views import EnterpriseCustomerManageLearnersView
 from enterprise.django_compatibility import reverse
-from enterprise.models import (
+from enterprise.models import (  # pylint:disable=no-name-in-module
     EnterpriseCustomer, EnterpriseCustomerUser, EnterpriseCustomerBrandingConfiguration,
-    EnterpriseCustomerIdentityProvider,
+    EnterpriseCustomerIdentityProvider, HistoricalUserDataSharingConsentAudit
```
smarnach: The magic that automatically inserts `HistoricalUserDataSharingConsentAudit` into `models` is rather ugly.  I wasn't aware it was doing that, so this import looked a bit surprising to me.  Why do you need `HistoricalUserDataSharingConsentAudit` rather than `UserDataSharingConsentAudit` here?

enterprise/admin/__init__.py
```diff
@@ -135,12 +165,44 @@ class EnterpriseCustomerUserAdmin(admin.ModelAdmin):
...
+            return super(
+                EnterpriseCustomerUserAdmin,
+                self
+            ).get_readonly_fields(
+                request,
+                obj=obj
+            ) + tuple(
+                get_all_field_names(self.model)
+            )
+        return super(EnterpriseCustomerUserAdmin, self).get_readonly_fields(request, obj=obj)
```
smarnach: Nit: You have the code

    super(EnterpriseCustomerUserAdmin, self).get_readonly_fields(request, obj=obj)

twice, in both branches.  The code would become more readable when written like this:

    fields = super(EnterpriseCustomerUserAdmin, self).get_readonly_fields(request, obj=obj)
    if obj:
        fields += tuple(get_all_field_names(self.model))
    return fields

enterprise/admin/views.py
haikuginger: External behavior should be identical. Should being the key word.

It's just a matter of "let's trust the Django ORM to do its job".

enterprise/admin/__init__.py
haikuginger: Only one `UserDataSharingConsentAudit` is attached to a given `EnterpriseCustomerUser` at a given time. We want to show a list of the changes that have been made to a given `UserDataSharingConsentAudit`, rather than just the current state, which means we need to use the historical model.

haikuginger:
Sandbox has been updated.

e-kolpakov:
@haikuginger @bradenmacdonald @smarnach I would disagree with @smarnach about implementation cost - in theory it would be as simple as adding a model admin for `PendingEnterpriseCustomerUser` and resolving the link to it in "ManageLearners" view the same way it is done for linked users. I would  say it is ok to ask edX about it, but maybe it would be simpler to just implement it.

e-kolpakov:
:+1:

- [x] I tested this: Data sharing consent history is displayedon EnterpriseCustomerUser admin and there's a link to EnterpriseCustomerUser details page from "Manage Learners" view.
- [x] I read through the code
- [x] ~~Includes documentation~~ Too early to document.

haikuginger:
@e-kolpakov, you're probably right. I just added a model admin for `PendingEnterpriseCustomerUser`. I also added a list of enrolled courses at upstream request (we had been communicating back and forth about that), and also changed the link in the Manage Learners page to be on the user email, so that the links for `EnterpriseCustomerUser` and `PendingEnterpriseCustomerUser` match up in style.

You should be able to test most of that right away; I'm working on a sandbox update to fix CMS static assets so I can add a course to be listed in the admin panel. Right now, with no courses present, you should see "None" listed next to "Enrolled courses" for all of the users.

haikuginger:
@e-kolpakov, on the sandbox, you should now see a course ID listed for the user with username `alterego`.

enterprise/enrollment_api.py
```diff
@@ -37,3 +37,11 @@ def enroll_user_in_course(user, course_details, mode):
     """
     client = get_enrollment_api_client()
     return client.enrollment.post(dict(user=user, course_details=course_details, mode=mode))
+
+
+def get_enrolled_courses(user):
+    """
+    Query the enrollment API to get a list of the courses a user is enrolled in.
+    """
+    client = get_enrollment_api_client()
```
e-kolpakov: @haikuginger it might be a good time to refactor this api into a class - repeating `get_enrollment_api_client` in every method is a certain code smell.

haikuginger:
@saleem-latif, @asadiqbal08, @mattdrayer, @bradenmacdonald, @kashifch for review.

enterprise/enrollment_api.py
haikuginger: I don't disagree.

haikuginger:
Sandbox has been updated with the latest; enrolled courses are listed by name with a link to the course info page in the LMS.

enterprise/admin/__init__.py
```diff
@@ -135,12 +167,89 @@ class EnterpriseCustomerUserAdmin(admin.ModelAdmin):
...
+        'enrolled_courses',
+    )
+
+    inlines = (
+        HistoricalUserDataSharingConsentAuditInlineAdmin,
+    )
+
+    def username(self, ecu):
+        """
+        Return the username for the attached user.
```
mattdrayer: Please expand the `ecu` argument name to `enterprise_customer_user` and include a description in the docstring.

enterprise/admin/__init__.py
```diff
@@ -135,12 +167,89 @@ class EnterpriseCustomerUserAdmin(admin.ModelAdmin):
...
+        """
+        return ecu.user.username
+
+    def enrolled_courses(self, ecu):
```
mattdrayer: Please expand the `ecu` argument name to `enterprise_customer_user` and include a description in the docstring.

enterprise/admin/__init__.py
```diff
@@ -135,12 +167,89 @@ class EnterpriseCustomerUserAdmin(admin.ModelAdmin):
...
+        """
+        Return a string represented the courses enrolled in
+        """
+        courses_string = mark_safe(self.get_enrolled_course_string(ecu))
```
mattdrayer: I always get concerned about exceptions when I see nested function calls -- do we need to protect against any errors here?

enterprise/admin/__init__.py
```diff
@@ -135,12 +167,89 @@ class EnterpriseCustomerUserAdmin(admin.ModelAdmin):
...
+            return readonly_fields + tuple(get_all_field_names(self.model))
+        return readonly_fields
+
+    def get_enrolled_course_string(self, ecu):
+        """
+        Get an HTML string representing the courses the user is enrolled in.
+        """
+        enrolled_courses = get_enrolled_courses(user=self.username(ecu))
+        course_details = []
+        for course in enrolled_courses:
```
mattdrayer: What if the value of enrolled_courses is `'None'`?

enterprise/admin/__init__.py
```diff
@@ -135,12 +167,89 @@ class EnterpriseCustomerUserAdmin(admin.ModelAdmin):
...
+            course_id = course['course_details']['course_id']
```
mattdrayer: Do we need to protect against missing fields/values here?

enterprise/enrollment_api.py
mattdrayer: +1

enterprise/enrollment_api.py
```diff
@@ -37,3 +37,25 @@ def enroll_user_in_course(user, course_details, mode):
...
+    return client.enrollment.get(user=user)
+
+
+def get_full_course_details(course_id):
+    """
+    Retrieve all available details about a course.
```
mattdrayer: Please add a description of the argument and return value to the docstring.

enterprise/enrollment_api.py
```diff
@@ -37,3 +37,25 @@ def enroll_user_in_course(user, course_details, mode):
...
+    """
+    session = requests.Session()
```
mattdrayer: This also seems like something that could be bundled up into its own class.

enterprise/admin/__init__.py
haikuginger: The values here are pretty much fundamental to the data we're receiving; while we could suppress potential errors by doing a `.get()` call, this is one case where I'd never _expect_ to get a `KeyError`, so if there is one, as a platform administrator, I'd like to know right away, and loudly. One example of such a case would be if the API client returned a value when it should have raised its own exception instead.

I'm certainly open to making whatever change you request, though.

enterprise/admin/__init__.py
haikuginger: I don't believe so. `mark_safe` just adds a `__html__()` method to the string; it doesn't perform any operations on the data.

haikuginger:
@mattdrayer, I've made most of those changes, and also refactored the `enrollment_api` module out to be a bit more Pythonic.

enterprise/edx_api.py
```diff
@@ -0,0 +1,80 @@
+# -*- coding: utf-8 -*-
+"""
+Utilities to get details from the course catalog API.
```
mattdrayer: This is more than just course catalog access, right?

enterprise/edx_api.py
```diff
@@ -0,0 +1,80 @@
+# -*- coding: utf-8 -*-
```
mattdrayer: So I'm slightly confused by this new module -- I do like the general idea, but I'm trying to understand if this module manages access to clients across *all* Open edX services, or if it's just for the LMS.  Currently it seems we are only concerned with LMS services, in which case we might want to rename the module to reflect the domain.

However, what might be better here is to introduce a `services` package/module which could contain all of the various adapters to communicate with remote services.

@zubair-arbi @saleem-latif @asadiqbal08 FYI

enterprise/edx_api.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def get_course_details(self, course_id):
+        """
+        Query the Enrollment API for the course details of the given course_id.
+        """
+        return self.client.course(course_id).get()
+
+    def enroll_user_in_course(self, username, course_id, mode):
+        """
+        Query the enrollment API to enroll the user in the course specified by course_details.
```
mattdrayer: We're not technically "querying" in this case (queries == reports/reads), we are "commanding" (mods/writes)

mattdrayer: Also I'd like to see the proper docstyle applied to these methods for arguments+returns.

enterprise/admin/__init__.py
mattdrayer: I'm okay keeping it as-is -- just wanted to point it out in case there was something more we wanted to do.

mattdrayer:
Getting close!  Also, looks like a rebase+reconciliation is needed for this branch.

haikuginger:
@griffresch, updated sandbox here.

enterprise/edx_api.py
haikuginger: Ah, yes.

enterprise/edx_api.py
haikuginger: Good call; I'll rename this to `lms_api`.

enterprise/edx_api.py
haikuginger: I'll make those changes; these were largely moved directly from the current position without modification.

haikuginger:
Rebased from 041c27c.

haikuginger:
@mattdrayer, ready for another pass.

mattdrayer:
Looks like we need a rebase on this branch -- otherwise 👍

haikuginger:
Rebased from daab8c1.

PR #25, opened 2016-12-22
[ENT-105] Enroll PendingEnterpriseCustomerUsers in courses when converted

haikuginger:
This pull request adds a feature such that when learners are enrolled in a course via the Manage Learners pane, but they don't exist, and as such are created as `PendingEnterpriseCustomerUser`s, a `PendingEnrollment` object is created that enrolls the user in a course when the learner creates an actual platform account.

**JIRA tickets**: Implements ENT-105

**Dependencies**: None

**Sandbox URL**: https://ent105.opencraft.hosting/ (if DNS hasn't shown up, use direct IP address: http://213.32.76.186; use `staff` account for superuser access)

**Merge deadline**: 2017-01-03

**Testing instructions**:

1. Create an `EnterpriseCustomer`
2. Create a course, noting its course ID (including the `course-v1` prefix)
3. In the EnterpriseCustomer's Manage Learners page, attempt to add a learner with an email address that does not yet exist as a user in the platform. Be sure to also attempt to enroll it in the course.
4. Note that a PendingEnterpriseCustomerUser object is created.
5. Using the standard registration methods, create a user with the aforesaid email address.
6. Note that the user is immediately enrolled in the course with the appropriate mode without further intervention.

1. At present, it was necessary to do the registration with local database methods; the entire user post-save method is wrapped in a single transaction; if we do an HTTP API call, then the user doesn't exist yet (from the perspective of the API) in order for the enrollment to be processed. In Django 1.9 and later, a `post_commit` hook is available, which will allow us to bypass that by making the enrollment HTTP call after the database commit is finalized.

**Reviewers**
- [ ] @e-kolpakov
- [ ] edX reviewer[s] TBD

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag
      finished.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=25"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

enterprise/enrollment_api.py
```diff
@@ -37,3 +49,21 @@ def enroll_user_in_course(user, course_details, mode):
...
+    This is only used in one place; in models.py. It is necessary
+    because the enrollment API can't enroll a user before it exists, and
+    the post_save signal for users results in a user not actually being
+    saved until after the API call is made.
+
+    In Django 1.9 and later, there's a transaction.on_commit hook that we can
+    use to create a callback. Once we're able to depend on having Django 1.9, we
+    can shift over to that, but for right now, we have to do it this way.
+    """
+    if CourseKey is None and CourseEnrollment is None:
```
e-kolpakov: @haikuginger nit: the more methods in this module are created the harder it is to mock enrollment_api. Please find tim to collect these methods into a class.

tests/test_enrollment_api.py
```diff
@@ -55,3 +57,8 @@ def test_enroll_user_in_course():
     assert actual_response == expected_response
     request = responses.calls[0][0]  # pylint: disable=no-member
     assert json.loads(request.body) == expected_response
+
+
+def test_enroll_locally_raises():
+    with raises(NotConnectedToEdX):
```
e-kolpakov: @haikuginger and what about happy path, when it should actually work? Genuine question - is there a test that tests that?

tests/test_models.py
```diff
@@ -19,8 +19,31 @@
...
+@mark.django_db
+@ddt.ddt
+class TestPendingEnrollment(unittest.TestCase):
+    """
+    Test for pending enrollment
+    """
+    @ddt.data(
+        str, repr
+    )
+    def test_string_conversion(self, method):
```
e-kolpakov: @haikuginger looks like this test does not actually test the only "sensible" method `coplete_enrollment`. Please find some time to add it.

enterprise/admin/views.py
```diff
@@ -199,8 +200,19 @@ def _enroll_users(cls, emails, course_details, mode, request):
             ).format(enrolled_count=enrolled_count, course_id=course_id))
         if non_existing:
             messages.warning(request, _(
-                "The following users do not have accounts yet and were not enrolled: {}"
+                "The following users do not have accounts yet and were not yet enrolled: {}"
```
e-kolpakov: @haikuginger Please update translations with this change.

tests/test_enrollment_api.py
haikuginger: Yes, this is tested [here](https://github.com/edx/edx-enterprise/pull/25/files#diff-41cb95e2211fd63cab0a14339e33dc9eR107).

enterprise/enrollment_api.py
haikuginger: @e-kolpakov, I don't think a class, solely for the purpose of namespacing, is the right way to go - it's unPythonic, as there's no shared state. Rather, tests can simply mock the shared API call, as is currently done.

tests/test_models.py
haikuginger: Done.

enterprise/admin/views.py
haikuginger: Done.

haikuginger:
@saleem-latif @asadiqbal08 @kashifch @mattdrayer @bradenmacdonald for review.

enterprise/admin/views.py
```diff
@@ -165,7 +166,7 @@ def _handle_bulk_upload(cls, enterprise_customer, manage_learners_form, request)
         return list(emails) + already_linked_emails
 
     @classmethod
-    def _enroll_users(cls, emails, course_details, mode, request):
+    def _enroll_users(cls, enterprise_customer, emails, course_details, mode, request):
         """
         Enroll the users with the given email addresses to the course specified by course_details.
         """
```
mattdrayer: Since we're here, we should add descriptions for all of the arguments to the docstring for this function.

enterprise/admin/views.py
mattdrayer: @srpearce -- is there a way around the proposed double-yet?   😄

enterprise/admin/views.py
```diff
@@ -199,8 +200,19 @@ def _enroll_users(cls, emails, course_details, mode, request):
             ).format(enrolled_count=enrolled_count, course_id=course_id))
         if non_existing:
             messages.warning(request, _(
-                "The following users do not have accounts yet and were not enrolled: {}"
+                "The following users do not have accounts yet and were not yet enrolled: {}"
             ).format(", ".join(non_existing)))
+            for email in non_existing:
+                pending_user = PendingEnterpriseCustomerUser.objects.get(
```
mattdrayer: Should this be `get_or_create`?

enterprise/enrollment_api.py
mattdrayer: I agree with @haikuginger with the use of a class for namespacing.  However, if this module is becoming unwieldy due to the amount of methods, then we could consider splitting the functions into two modules, which I would be okay with.

enterprise/models.py
```diff
@@ -304,6 +305,54 @@ def __repr__(self):
...
+        max_length=25,
+        blank=False
+    )
+
+    class Meta(object):
+        unique_together = (("user", "course_id"),)
+
+    def complete_enrollment(self):
+        """
+        Enroll a newly-create user in the relevant course.
```
mattdrayer: They may not be newly-created at this point, all we know is that they are a user.

enterprise/models.py
```diff
@@ -304,6 +305,54 @@ def __repr__(self):
...
+        """
+        user = User.objects.get(email=self.user.user_email)
```
mattdrayer: Do we need to protect against a DoesNotExist exception here?

enterprise/signals.py
```diff
@@ -47,4 +47,6 @@ def handle_user_post_save(sender, **kwargs):  # pylint: disable=unused-argument
         enterprise_customer=pending_link_record.enterprise_customer,
         user_id=user_instance.id
     )
+    for enrollment in pending_link_record.pendingenrollment_set.all():
```
mattdrayer: Please add a code comment describing what is happening in this new addition to the workflow.  Also please update the docstring to mention this new work as well.

mattdrayer: Would it be possible to rename `pending_link_record` to `pending_enterprise_customer_user`?  I'm finding the change in terminology a little confusing.

bradenmacdonald:
@haikuginger @mattdrayer I thought that we were going to use the updated `CourseEnrollmentAllowed` model from https://github.com/edx/edx-platform/pull/13815 for this, since it already has the functionality we need. Did I miss some rationale for creating the new `PendingEnrollment` model, which seems to duplicate that functionality?

enterprise/admin/views.py
haikuginger: No; all pending ECUs should have already been created at this point. An email address in the non_existing list that hasn't been created as a pECU yet is a sign of an upstream logic failure.

enterprise/models.py
haikuginger: No; I don't think there's anything we could do here that'd be more helpful, diagnostically-speaking, than the DoesNotExist that would happen if `complete_enrollment` was called on a PendingEnrollment that didn't have a real user yet.

haikuginger:
@bradenmacdonald, I decided to add the logic locally, rather than rely on the edx-platform PR, for a couple reasons:

1. The length of time before it merges is, as of yet, unknown. Relying on it would push this feature back until it's ready.
2. The long-term goal of edx-enterprise is for it to be deployed as an IDA; designing this feature to rely on CourseEnrollmentAllowed would couple it more closely to edx-platform.

@mattdrayer, comments have been addressed.

bradenmacdonald:
@haikuginger 

> The length of time before it merges is, as of yet, unknown. Relying on it would push this feature back until it's ready.

I thought we could cherry-pick that part of the PR and merge it in ourselves.

> The long-term goal of edx-enterprise is for it to be deployed as an IDA; designing this feature to rely on CourseEnrollmentAllowed would couple it more closely to edx-platform.

That argument sounds reasonable... though it's probably better still to add an API to the platform for CourseEnrollmentAllowed, rather than duplicating the functionality in this app.

mattdrayer:
Along the lines of what @bradenmacdonald said -- I'm okay with utilizing edx-platform resources as long as we insert an API adapter versus binding directly to a class/function.  This will allow us to pull things apart down the road (and hopefully not too far along Ye Roade).

haikuginger:
@griffresch, updated sandbox here.

enterprise/admin/views.py
srpearce: I'm not sure why we're using the past tense in the second part of the sentence. Also, are both conditions true (they don't have accounts AND they're not enrolled), or does just one have to be true?

Assuming only one has to be true, how about this:

The following users do not have <platform name> accounts or are not enrolled in this course.

Assuming both have to be true, I'd suggest this:

The following users do not have <platform name> accounts and are not enrolled in this course.

enterprise/signals.py
haikuginger: Switched to `pending_ecu`; pylint didn't like `pending_enterprise_customer_user` as a variable name.

haikuginger:
> I thought we could cherry-pick that part of the PR and merge it in ourselves.

That whole PR is, essentially, "that part". The only part we'd get to leave out would be the UI, which I don't think is the controversial part.

enterprise/admin/views.py
haikuginger: @srpearce, this message is used when an action is taken which would result in enrollment. In other words, the second clause isn't a conditional - it's the result of the operation.

Similar to the phrase: "The following users are lactose-intolerant, and were not given ice cream cones:"

haikuginger:
@mattdrayer, ready for a second pass.

enterprise/admin/views.py
srpearce: OK, makes sense. 

Is this better?

The following users do not have an account on `<platform name>`. They have not been enrolled in the course.

mattdrayer: NO ICE CREAM CONES???

haikuginger: @srpearce, close! We want to convey that (as of this PR), those users _will_ be enrolled in the course, though, once they register.

srpearce: @mattdrayer They can have the cone, just not the ice cream. Unless they have to eat gluten-free, in which case they're outta luck.

The following users do not have an account on `<platform name>`. They have not been enrolled in the course. When these users create an account, they will be enrolled in the course automatically.

mattdrayer:
Rebase needed, otherwise LGTM -- 👍 -- @zubair-arbi was not tagged initially so adding him now for context.

haikuginger:
Rebased, squashed, etc., from 31b0435.

PR #27, opened 2017-01-12
ENT-97: Email learners on enrollment

haikuginger:
**Description:**

This pull request adds a feature that enables platform owners to email new learners upon enrollment in a course, using either a built-in template with approved text, or by setting up a custom, site-specific, set of templates that can be used to render messages.

**JIRA:** [ENT-97](https://openedx.atlassian.net/browse/ENT-97)

**Dependencies:** Depends on an incoming PR from @e-kolpakov; the ticket acceptance criteria specify that learners can be notified when being enrolled in a program. However, program enrollment does not exist as of yet.

**Merge deadline:** 30 January 2017

**Sandbox Details:** TBD (being provisioned)

**Testing instructions:**

1. In the Manage Learners page in a devstack (where emails are logged to the console), add a learner to a course using an arbitrary email address.
    - Observe in the console that an email has been "sent" using both the standard plaintext template and the standard HTML template - email clients will choose the appropriate one to render.
2. Set up an `EnrollmentNotificationEmailTemplate` object in the Enterprise admin; be sure to fill in templates for both the plaintext and HTML templates, as well as for the subject line.
3. Again, add a learner to a course using an arbitrary email address.
    - Observe in the console that the email was "sent", and that your custom templates were used to render the message, rather than the standard templates.

**Reviewers:**
- [x] @bradenmacdonald 
- [ ] edX TBD

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:**

1. More tests are still incoming; this is not entirely complete yet.
2. A flag needs to be added to the Enterprise admin to make notification optional
3. Because program enrollment doesn't exist yet, notification on program enrollment isn't implemented yet.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=27"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

haikuginger:
Rebased from caeb0ee.

enterprise/models.py
```diff
@@ -541,3 +543,71 @@ def __repr__(self):
...
+    BODY_HELP_TEXT = _(
+        'Fill in a standard Django template; available variables include `user_name`, '
+        '`course_name`, `program_type`, `course_type`, `organization_name`, `start_date`, '
+        'and `course_url`. All items are strings, except for `start_date`, which is a '
+        '`datetime.datetime` object.'
+    )
+
+    SUBJECT_HELP_TEXT = _(
+        'Fill in a Python format string, using the same variables as are available for '
+        'the HTML and plaintext body template.'
```
bradenmacdonald: Can you provide an example?

enterprise/templates/enterprise/emails/user_notification.html
```diff
@@ -0,0 +1,10 @@
+<p>{% if user_name %}Dear {{ user_name }},{% else %}Hi!{% endif %}</p>
+<p>
+You have been enrolled in <a href="{{ course_url}}">{{ course_name }}</a>, a {% if program_type %}{{ program_type }} {% endif %}{{ course_type }} offered by {{ organization_name }}. This {{ course_type }} begins {{ start_date | date:"F j, Y" }}. For more information, see <a href="{{ course_url}}">{{ course_name }}</a>.
+</p>
+<p>
+Thanks,
+</p>
+<p>
+The {{course_name}} team
+</p>
```
bradenmacdonald: An HTML email template should have a full `<html>` document - see http://localhost:8000/admin/bulk_email/courseemailtemplate/ on your devstack for an example. It would even be good to use that as a template here, since it's more detailed and pretty. (Arguably too detailed).

Also, admin users should have some way of seeing a default template - I think that's done somehow for http://localhost:8000/admin/bulk_email/courseemailtemplate/ but for this new template, I have no examples to work from in the django admin.

enterprise/templates/enterprise/emails/user_notification.txt
```diff
@@ -0,0 +1,9 @@
+{% if user_name %}Dear {{ user_name }},{% else %}Hi!{% endif %}
+
+You have been enrolled in {{ course_name }}, a {% if program_type %}{{ program_type }} {% endif %}{{ course_type }} offered by {{ organization_name }}. This {{ course_type }} begins {{ start_date | date:"F j, Y" }}. For more information, see the following link:
```
bradenmacdonald: Can you mark this up for i18n?

Also, this way of combining integrating `course_type` and other vars into a sentence does not work for translation, even if the word `"course"` is translated, because it may need to be inflected differently depending on the context of the translated sentence. So you'll need to refactor this a bit.

```
{%if is_a_program%}
{% blocktrans %}
You've been enrolled in the {{name}} {{program_type}} program, ...
{% blocktrans %}
{%else%}
...
{%endif%}
```

(`program_type` can be inlined since it's a trademarked name and would not be translated nor inflected.)

enterprise/utils.py
```diff
@@ -166,3 +171,97 @@ def get_catalog_admin_url_template():
...
+        organization: The name of the organization sponsoring the enrollment
+        start_date: The date the course first becomes available
+        program_type: An additional classifier for the type of the course - for example,
+            a program could be a "MicroMaster" program
+        email_connection: An existing Django email connection that can be used without
+            creating a new connection for each individual message
+    """
+    if hasattr(user, 'first_name') and hasattr(user, 'username'):
+        # PendingEnterpriseCustomerUsers don't have usernames or real names. We should
+        # template slightly differently to make sure weird stuff doesn't happen.
```
bradenmacdonald: Needs to be mentioned in the admin help text or explained clearly in a comment in the example template. (Without referencing any python class names :-p )

enterprise/utils.py
```diff
@@ -166,3 +171,97 @@ def get_catalog_admin_url_template():
...
+        'user_name': user_name,
+        'course_name': course,
+        'program_type': program_type,
+        'course_type': course_type,
+        'organization_name': organization,
+        'start_date': start_date,
+        'course_url': course_url
+    }
+    try:
+        current_site = get_current_site(request)
```
bradenmacdonald: What happens if an edx staff person is logged in to the django admin on edx.org, but enrolls a learner in a WL site course? Wouldn't this send the edx email template, when we actually want the template of that particular WL site?

enterprise/utils.py
```diff
@@ -166,3 +171,97 @@ def get_catalog_admin_url_template():
...
+        html_msg = render_to_string(
+            'enterprise/emails/user_notification.html',
+            msg_context
+        )
+        subject_line = 'You\'ve been enrolled in {}!'.format(course)
+    else:
+        if site_template.subject_line:
+            subject_line = site_template.subject_line.format(**msg_context)
+        else:
+            subject_line = 'You\'ve been enrolled in {}!'.format(course)
```
bradenmacdonald: Optional: Can you just make `'You\'ve been enrolled in {course_name}!' the `default` on the model, so you don't have to have so many if/else conditions here?

enterprise/utils.py
```diff
@@ -166,3 +171,97 @@ def get_catalog_admin_url_template():
...
+            plain_msg, html_msg = site_template.render_templates(msg_context)
+        else:
+            plain_msg = render_to_string(
+                'enterprise/emails/user_notification.txt',
+                msg_context
+            )
+            html_msg = render_to_string(
+                'enterprise/emails/user_notification.html',
+                msg_context
+            )
```
bradenmacdonald: Can you refactor this so that the

```python
+            plain_msg = render_to_string(
 +                'enterprise/emails/user_notification.txt',
 +                msg_context
 +            )
 +            html_msg = render_to_string(
 +                'enterprise/emails/user_notification.html',
 +                msg_context
 +            )
```
code doesn't appear twice?

enterprise/admin/views.py
```diff
@@ -273,6 +292,17 @@ def _enroll_users(cls, enterprise_customer, emails, course_id, mode, request):
                     course_id=course_id,
                     course_mode=mode
                 )
+                if notify:
+                    send_notification_message(
+                        request,
+                        pending_user,
+                        course_details['name'],
+                        course_url,
+                        _('course'),
```
haikuginger: Makes sense.

enterprise/templates/enterprise/emails/user_notification.html
haikuginger: Re: the content: that's prescribed by the ticket, so I'm not sure if we want a different template, especially as we're using a different set of data here.

But I agree that there should be a preview view.

enterprise/templates/enterprise/emails/user_notification.txt
haikuginger: Makes sense.

enterprise/utils.py
```diff
@@ -166,3 +171,97 @@ def get_catalog_admin_url_template():
...
+):
+    """
+    Send an email notifying a user about their enrollment in a course.
+
+    Arguments:
+        user: Either a User object or a PendingEnterpriseCustomerUser that we can use
+            to get details for the email
+        course: The name of the course or program that the user's been enrolled in
+        course_url: The URL to the LMS details page for the course or program.
+        course_type: The classification of the learning unit ("course" or "program")
```
haikuginger: I'll think on it. I agree that the current naming is confusing.

enterprise/utils.py
haikuginger: Agreed.

enterprise/utils.py
haikuginger: Right now, it'd send from edx.org - good catch. Possibly grab the correct site off the EnterpriseCustomer?

enterprise/utils.py
haikuginger: We won't always have the model available, but I'll take a look at what we can do.

enterprise/utils.py
haikuginger: Yes.

enterprise/admin/views.py
```diff
@@ -328,7 +358,15 @@ def post(self, request, customer_uuid):
             if course_details:
                 course_mode = manage_learners_form.cleaned_data[ManageLearnersForm.Fields.COURSE_MODE]
                 course_id = course_details['course_id']
-                self._enroll_users(enterprise_customer, linked_learners, course_id, course_mode, request)
+                notify = manage_learners_form.cleaned_data.get(ManageLearnersForm.Fields.NOTIFY) == 'email'
```
haikuginger: I'll check into it.

enterprise/admin/__init__.py
```diff
@@ -18,8 +18,9 @@
 from enterprise.django_compatibility import reverse
 from enterprise.lms_api import CourseApiClient, EnrollmentApiClient
 from enterprise.models import (  # pylint:disable=no-name-in-module
-    EnterpriseCustomer, EnterpriseCustomerUser, EnterpriseCustomerBrandingConfiguration,
-    EnterpriseCustomerIdentityProvider, HistoricalUserDataSharingConsentAudit, PendingEnterpriseCustomerUser,
+    EnrollmentNotificationEmailTemplate, EnterpriseCustomer, EnterpriseCustomerUser,
+    EnterpriseCustomerBrandingConfiguration, EnterpriseCustomerIdentityProvider,
+    HistoricalUserDataSharingConsentAudit, PendingEnterpriseCustomerUser,
```
haikuginger: I absolutely agree, but it'd be impossible to make the `isort` linter accept that without making that change on all existing code - which might cause some issues to PRs in flight. I'd love to handle that as a separate PR though.

enterprise/utils.py
bradenmacdonald: Yep, grab the site off the EnterpriseCustomer - that sounds simple and would provide the correct result.

enterprise/templates/enterprise/emails/user_notification.html
bradenmacdonald: At the very least, the template should have `<html>`, `<body>` etc. Check out https://github.com/leemunroe/responsive-html-email-template for an example that's very nice and also where the code is easy to read, unlike the edx example.

enterprise/admin/__init__.py
mattdrayer: Please feel free to log ENT project tickets for any tech debt you identify -- these can be used as filler-jobs for when there are small gaps at the ends of sprints.

enterprise/admin/forms.py
```diff
@@ -46,6 +46,15 @@ class ManageLearnersForm(forms.Form):
             ("honor", _("Honor")),
         ],
     )
+    notify_on_enrollment = forms.ChoiceField(
+        label=_("Notify learners of enrollment"),
+        choices=[
+            ("email", _("Email")),
```
mattdrayer: Since it appears that we are describing the notification mechanism with these choices, we might instead say "by_email", instead of just "email".

enterprise/admin/forms.py
```diff
@@ -103,6 +113,12 @@ def clean_course(self):
         except (HttpClientError, HttpServerError):
             raise ValidationError(ValidationMessages.INVALID_COURSE_ID.format(course_id=course_id))
 
+    def clean_notify(self):
+        """
+        Clean the notify_on_enrollment field.
+        """
+        return self.cleaned_data[self.Fields.NOTIFY] or 'email'
```
mattdrayer: should `'email'` here be instead something like `settings.DEFAULT_NOTIFICATION_MECHANISM`?

enterprise/admin/views.py
```diff
@@ -218,7 +222,7 @@ def _handle_bulk_upload(cls, enterprise_customer, manage_learners_form, request)
         return list(emails) + already_linked_emails
 
     @classmethod
-    def _enroll_users(cls, enterprise_customer, emails, course_id, mode, request):
+    def _enroll_users(cls, enterprise_customer, emails, course_id, mode, request, notify=True):
```
mattdrayer: Please add `notify` to the arguments listed in the docstring

enterprise/admin/views.py
```diff
@@ -234,23 +238,38 @@ def _enroll_users(cls, enterprise_customer, emails, course_id, mode, request):
...
+                enrollment_client.enroll_user_in_course(user.username, course_id, mode)
             except HttpClientError as exc:
                 failed.append(email)
                 error_message = json.loads(exc.content.decode()).get("message", "No error message provided.")
                 logging.error(
                     "Error while enrolling user %(user)s: %(message)s",
-                    dict(user=username, message=error_message),
+                    dict(user=user.username, message=error_message),
                 )
             else:
```
mattdrayer: For my own knowledge, why use `else` here versus using a `continue` in the preceding `except` block?

enterprise/admin/views.py
mattdrayer: Along the lines of readability, this private/internal function is orchestrating a fair amount of work and it might be better to break it up into additional private/internal functions.  I understand that it's possible to go too far with that approach, but it would help to separate things a bit more, imho.

enterprise/admin/views.py
mattdrayer: Ah, I made a similar comment above, suggesting something like `settings.DEFAULT_NOTIFICATION_MECHANISM`

enterprise/models.py
```diff
@@ -541,3 +543,71 @@ def __repr__(self):
...
+    """
+    Store optional templates to use when emailing users about course enrollment events.
+    """
+
+    class Meta(object):
+        app_label = 'enterprise'
+
+    BODY_HELP_TEXT = _(
+        'Fill in a standard Django template; available variables include `user_name`, '
+        '`course_name`, `program_type`, `course_type`, `organization_name`, `start_date`, '
```
mattdrayer: `program_type`, for that matter

enterprise/models.py
```diff
@@ -541,3 +543,71 @@ def __repr__(self):
...
+        """
+        return self.render_template(mark_safe(self.html_template), kwargs)
+
+    def render_plaintext_template(self, kwargs):
+        """
+        Render just the plaintext template and return it as a string.
+        """
+        return self.render_template(self.plaintext_template, kwargs)
+
+    def render_templates(self, kwargs):
```
mattdrayer: If the goal of this function is to render ALL templates, then let's be explicit about that -> `render_all_templates()` -- otherwise I'd expect to be able to send in some sort of switch to control which template(s) are rendered.

enterprise/admin/views.py
haikuginger: To my knowledge, they'd work the same at the moment, but a `continue` might cause issues later on if we want to add something that happens in the loop, even after an exception, later on. I chose not to modify the existing code.

haikuginger:
I've just pushed a big commit with a refactor of the code addressing the majority of the comments: Still remaining: set up a "preview" for the email template object, which I'll be doing either tonight or tomorrow.

enterprise/admin/views.py
```diff
@@ -229,28 +232,47 @@ def _enroll_users(cls, enterprise_customer, emails, course_id, mode, request):
...
             request: The HTTP request the enrollment is being created by
+            notify: Whether to notify (by email) the users that have been enrolled
         """
         enrolled = []
         non_existing = []
         failed = []
         enrollment_client = EnrollmentApiClient()
+        course_details = CourseApiClient().get_course_details(course_id)
+        reverse_method = get_reverse_by_site(request, enterprise_customer.site)
+        course_url = reverse_method('about_course', args=(course_id,))
```
bradenmacdonald: This `get_reverse_by_site` API seems needlessly complicated to me. Can't you just create a `reverse_site_url(request, site, *args, **kwargs)` method?

```python
course_url = reverse_site_url(request, site=enterprise_customer.site, view_name='about_course', args=(course_id, ))
```

You can also implement it simply as `return 'https://' + site.domain + reverse(*args, **kwargs)`.

Actually, on second thought, this assumes that the enterprise app and the LMS are deployed into the same django project. Does the Course catalog API not provide the URL of the course?

enterprise/admin/views.py
```diff
@@ -229,28 +232,47 @@ def _enroll_users(cls, enterprise_customer, emails, course_id, mode, request):
...
+        email_conn = mail.get_connection()
+        email_conn.open()
```
bradenmacdonald: You are not closing the connection - better to use it as a context manager so you don't have to explicitly close it:

```python
with mail.get_connection() as connection:
```

Also, can we avoid opening the mail connection if we are not going to be notifying the users?

enterprise/admin/views.py
```diff
@@ -229,28 +232,47 @@ def _enroll_users(cls, enterprise_customer, emails, course_id, mode, request):
...
+                    send_notification_message(
+                        user=user,
+                        enrolled_in={
+                            'name': course_details['name'],
+                            'url': course_url,
+                            'type': 'course',
+                            'start': datetime.datetime.strptime(course_details['start'], '%Y-%m-%dT%H:%M:%SZ'),
+                        },
+                        enterprise_customer=enterprise_customer,
+                        email_connection=email_conn,
```
bradenmacdonald: This way of structuring the signature of this method is much nicer - thanks :)

enterprise/admin/views.py
```diff
@@ -229,28 +232,47 @@ def _enroll_users(cls, enterprise_customer, emails, course_id, mode, request):
...
             else:
                 enrolled.append(email)
+                if notify:
+                    send_notification_message(
+                        user=user,
+                        enrolled_in={
+                            'name': course_details['name'],
+                            'url': course_url,
+                            'type': 'course',
+                            'start': datetime.datetime.strptime(course_details['start'], '%Y-%m-%dT%H:%M:%SZ'),
```
bradenmacdonald: Is there some reason why so many edX API clients do not parse dates into python `datetime` objects? I don't like having date parsing logic scattered around all over the place.

enterprise/admin/views.py
```diff
@@ -328,7 +362,16 @@ def post(self, request, customer_uuid):
             if course_details:
                 course_mode = manage_learners_form.cleaned_data[ManageLearnersForm.Fields.COURSE_MODE]
                 course_id = course_details['course_id']
-                self._enroll_users(enterprise_customer, linked_learners, course_id, course_mode, request)
+                notification_type = manage_learners_form.cleaned_data.get(ManageLearnersForm.Fields.NOTIFY)
+                notify = notification_type == ManageLearnersForm.NotificationTypes.BY_EMAIL
```
bradenmacdonald: Are we expecting other types of notification in the future? Because in the form, you've defined the notification options as an enumerated choice set even though it's essentially a boolean choice, and in this part of the code you are explicitly treating it as a boolean.

enterprise/templates/enterprise/emails/user_notification.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n %}{% load enterprise_string_utils %}<html>
+<body>
+<p>{% if user_name %}{% blocktrans %}Dear {{ user_name }},{% endblocktrans %}{% else %}{% blocktrans %}Hi!{% endblocktrans %}{% endif %}</p>
+<p>{% if enrolled_in.type == "program" %}
+{% blocktrans with program_url=enrolled_in.url program_name=enrolled_in.name program_branding=enrolled_in.branding|pad_right start_date=enrolled_in.start|date:"F j, Y" %}You have been enrolled in <a href="{{ program_url }}">{{ program_name }}</a>, a {{ program_branding }}program offered by {{ organization_name }}. This program begins {{ start_date }}. For more information, see <a href="{{ program_url }}">{{ program_name }}</a>.{% endblocktrans %}{% else %}
```
bradenmacdonald: For the HTML template, you don't need `pad_right` - just put a space in `{{ program_branding }}program`. Browsers automatically combine adjacent whitespace, so the end result is the same. I'd probably do the same for the plaintext too - it's simpler and worst case scenario is that some program with no branding name shows up as "a  program" and mode people won't even notice the extra space. Many modern mail clients render plain text emails using an HTML interface anyways, which means that adjacent spaces get combined there too.

haikuginger:
@bradenmacdonald, I've just added a preview option for HTML email templates.

enterprise/templates/enterprise/emails/user_notification.html
bradenmacdonald: For dates, it's better not to use any specific format like `F j, Y` since that is specific to English speakers. Instead, use  "DATE_FORMAT", "DATETIME_FORMAT", "SHORT_DATE_FORMAT" or "SHORT_DATETIME_FORMAT", which will pick an appropriate date format based on the active locale. ([ref](https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#date))

e.g. `{{ value|date:"SHORT_DATE_FORMAT" }}`

enterprise/utils.py
```diff
@@ -166,3 +173,156 @@ def get_catalog_admin_url_template():
...
+
+
+def get_notification_subject_line(course_name, template_configuration=None):
+    """
+    Get a subject line for a notification email.
+
+    The method is designed to fail in a "smart" way; if we can't render a
+    database-backed subject line template, then we'll fall back to a template
+    saved in the Django settings; if we can't render _that_ one, then we'll
+    fall through to a friendly string written into the code.
```
bradenmacdonald: Can you mention why this would fall through?

bradenmacdonald:
Thanks @haikuginger  - looking good. Are you going to have a sandbox or should I plan to test locally?

enterprise/admin/views.py
```diff
@@ -12,23 +13,71 @@
...
+    }
+
+    def get(self, request, template_id, view_type):
+        """
+        Render the given template with the stock data.
+        """
+        template = get_object_or_404(EnrollmentNotificationEmailTemplate, pk=template_id)
+        if view_type not in self.view_type_contexts:
+            return HttpResponse(status=404)
+        logger.warning(template)
```
mattdrayer: Just checking -- is this `logger` needed?

enterprise/admin/views.py
mattdrayer: I'm not sure why this is, but we could at least move the date-parsing logic into a utility method to contain the sprawl.

enterprise/admin/views.py
mattdrayer: I'd prefer to keep things open-ended and stick with the enum pattern -- I'm sure at some point someone will want Instagram support...

enterprise/utils.py
```diff
@@ -166,3 +173,156 @@ def get_catalog_admin_url_template():
...
+    except KeyError:
+        pass
+
+    try:
+        return default_subject_template.format(course_name=course_name)
+    except KeyError:
+        return stock_subject_template.format(course_name=course_name)
+
+
+def send_notification_message(user, enrolled_in, enterprise_customer, email_connection=None):
```
mattdrayer: If this method is solely-intended to handle email communications we should probably be more specific with the name -- eg `send_email_notification_message`

enterprise/admin/views.py
haikuginger: > This get_reverse_by_site API seems needlessly complicated to me. Can't you just create a reverse_site_url(request, site, *args, **kwargs) method?

Certainly! I wanted to produce a method that had the same signature as the stock `reverse()` function, but I'll change it out.

enterprise/admin/views.py
haikuginger: Good catch! From what I can tell, though, those two items are mutually exclusive by default, so I'm putting together a context manager that conditionally opens and closes a connection.

enterprise/admin/views.py
haikuginger: In re: your code suggestion. I'd rather not do it that way; it assumes that the site is accessible over https. My code here makes an assumption too; that the other site can be accessed using the same scheme as the admin site, but I think that's just enough safer that I'd rather keep it, even though it's more complicated.

> Actually, on second thought, this assumes that the enterprise app and the LMS are deployed into the same django project. Does the Course catalog API not provide the URL of the course?

As far as I can tell from available information, it does not. I've done extensive testing, and have been unable to locate anything that suggests it does.

enterprise/admin/views.py
haikuginger: The EdX API client is pretty sparse; it doesn't actually know much about the API structure. And, since JSON doesn't actually have a standard datetime format or way to specify that a field is, in fact, a datetime, there's not a reliable way to automatically parse a datetime from JSON.

enterprise/admin/views.py
haikuginger: I agree with Matt on this one.

enterprise/templates/enterprise/emails/user_notification.html
haikuginger: Ah, good to know! I actually didn't know about the automatic whitespace combination. I'll change to use the built-in date format string.

enterprise/utils.py
haikuginger: Will add - for future readers of this PR, one case might be if a site admin entered a subject line template with template tags other than `{course_name}`; that'd result in a `KeyError`.

enterprise/admin/views.py
haikuginger: Nope!

enterprise/utils.py
haikuginger: Works for me!

haikuginger:
@bradenmacdonald, ready for another pass. I'm spawning a sandbox for testing, etc., but if you want to review in the meantime, the local setup should be extremely simple.

enterprise/admin/views.py
bradenmacdonald: > In re: your code suggestion. I'd rather not do it that way; it assumes that the site is accessible over https.

Yeah, but any production site should always be using HTTPS (and e.g. all OpenCraft and edX sandboxes do, and certainly anyone selling "enterprise" offerings would be using HTTPS). The only time HTTPs is not used AFAIK would be on devstacks, and then no emails are getting sent anyways - just logs to the console, so I assume it's not a big deal if the URLs shown in the console logged emails use the wrong protocol. But if that offends your sensibilities, it's fine to leave as is :)

enterprise/admin/views.py
bradenmacdonald: Fine by me! Can we treat it as an enum everywhere though? Or maybe even a set :-p (e.g. why should "email" and "SMS", or whatever other options you envision be mutually exclusive). This change is not super important though imho.

enterprise/admin/views.py
bradenmacdonald: @haikuginger I was talking to @mattdrayer just now and he said that the course discovery API should return a URL for each course, but only if the course is on the drupal marketing site. (So it would not currently work on devstack or on white label sites).

enterprise/utils.py
```diff
@@ -49,6 +49,39 @@ def __init__(self, *args, **kwargs):
...
+        if self.open_conn:
+            self.conn = mail.get_connection()
+            self.conn.open()
+        return self.conn
+
+    def __exit__(self, *excinfo):
+        """
+        Close the email connection if it was opened.
+        """
+        if self.conn and self.open_conn:
```
bradenmacdonald: This is a nit, but it's better just to check `if self.conn` here - regardless of whether a connection was supposed to be open or not, if the connection was opened, it should be closed.

bradenmacdonald:
@haikuginger Thanks! Looking good to me. I appreciate the ability to preview templates from the admin - very handy.

:+1:

- [x] I tested this: tested on my devstack, saw emails coming through to the console.
- [x] I read through the code
- [x] Includes documentation: somewhat - inline in the admin, yes.

enterprise/utils.py
haikuginger: Yeah, sounds good.

haikuginger:
Rebased from 6666ae2.

@bradenmacdonald, I've added notification handling for Programs, now that @e-kolpakov's work is merged, and after doing some research, I've also moved some data gathering over to the course discovery API.

haikuginger:
@mattdrayer, I've squashed from b33b501; this is ready for merge whenever you declare it so.

mattdrayer:
👍 🚀

bradenmacdonald:
@haikuginger, FYI when verifying this on a sandbox without the catalog service, I got this stacktrace - might be worth providing better error handling if we can add that to a future PR?

```
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/utils/decorators.py", line 145, in inner
    return func(*args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/newrelic-2.74.0.54/newrelic/hooks/framework_django.py", line 503, in wrapper
    return wrapped(*args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/utils/decorators.py", line 110, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/views/decorators/cache.py", line 57, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/contrib/admin/sites.py", line 233, in inner
    return view(request, *args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/newrelic-2.74.0.54/newrelic/hooks/framework_django.py", line 885, in wrapper
    return wrapped(*args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/views/generic/base.py", line 89, in dispatch
    return handler(request, *args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/enterprise/admin/views.py", line 471, in post
    notify=notify,
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/enterprise/admin/views.py", line 289, in _enroll_users
    course_url = course_details.get('marketing_url')
AttributeError: 'list' object has no attribute 'get'
```

mattdrayer:
@bradenmacdonald @haikuginger can we file a bug on this so the issue does not get lost in the shuffle?

haikuginger:
@mattdrayer, creating a story now.

PR #29, opened 2017-01-13
ENT-94: Enroll students to a program

e-kolpakov:
**Description:** This PR adds ability to enroll students into a program

**JIRA:** [ENT-94](https://openedx.atlassian.net/browse/ENT-94)

**Dependencies:** no dependencies so far

**Testing instructions:**

First, prepare the environment:
1. Create Enterprise Customer (or reuse existing)
2. Create a couple of courses (or reuse existing)
3. Launch LMS and course-discovery devstack, make sure they are working together nicely.
4. In course catalog create course runs for courses selected for this test session. Add seats to those courses (+ make sure there are courses with multiple seat types and there's at least one combination of courses that have at least one common seat type).
5. Create program(s) and include some combinations of courses into each.

Testing instructions:
1. Attempt adding single user and enrolling it into a program by UUID:
1.a. Go to course discovery admin and note program UUID (or send a GET to `course-discory/api/v1/programs` and pick one from the response)
1.b. Go to "Manage Learners" view
1.c. Fill in user email
1.d. Paste program UUID into a field
1.e. Choose one of the common enrollment modes
1.f. Submit form
1.g. Observe: user is added to corresponding list of users and (will be) enrolled into all courses in the program with the chosen enrollment mode
2. Attempt adding single user and enrolling it into a program by title - same as (1), but in step d use program title instead of ID.
3. Same as 1, but use bulk students upload - all students must be enrolled in all courses in chosen program.
4. Same as 2, but use bulk students upload.
5. Check validation:
5.a. If program was not found neither by UUID nor by name - form should display sensible error and not enroll/link student(s)
5.b. If chosen mode is not available across all selected course runs - form should display sensible error and not enroll/link student(s) and also list what common modes are available for courses in the program.
5.c. If LMS API or course discovery API is not available - it is going to crash when submitting the form (behavior not affected by this PR).
6. Make sure JS interactions work fine:
6.1. When non empty string is typed/pasted into course ID, program ID input becomes disabled and after 300 milliseconds after course modes are loaded (if user continues to type it waits) and put into course mode select.
6.2. When non empty string is typed/pasted into program ID, course ID input becomes disabled and after 300 milliseconds after default course modes are put into course mode select.
6.3. When course ID or program ID is emptied, other input becomes available immediately.

**Reviewers:**
- [x] @haikuginger 
- [x] @asadiqbal08 
- [x] @saleem-latif

CC @mattdrayer 

**Merge checklist:**
- [x] All reviewers approved
- [ ] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:**

`EnterpriseCustomerManageLearnersView` has become too big and performs too much actions: manages form, links users to EnterpriseCustomer, enrolls them into courses, etc. It makes it harder and harder to test. Might be good to start pulling it apart, or at least to split the logic of parsing user input and actually performing actions on that input (enrolling/linking students) into a separate objects.

Also, `ManageLearnersForm` starts to become too big too - might be a good place to stop and re-think the enrollment+linking approach.


edx-webhook:
Thanks for the pull request, @e-kolpakov! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=29"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

enterprise/admin/utils.py
```diff
@@ -112,3 +130,18 @@ def validate_email_to_link(email, raw_email=None, message_template=None, ignore_
             email=email, ec_name=existing_record.enterprise_customer.name
         ))
     return bool(existing_record)
+
+
+def get_course_runs_from_program(program):
+    """
+    Return course runs from program data.
+
```
haikuginger: EdX has been asking for arguments listed in the docstrings of new and updated methods - could you add that here and wherever else is relevant?

enterprise/admin/utils.py
```diff
@@ -112,3 +130,18 @@ def validate_email_to_link(email, raw_email=None, message_template=None, ignore_
...
+    Returns:
+        set: course runs in given program
+    """
+    return set(
```
haikuginger: The logic here is a little hard to follow; I think it might be better to break this out into a set of nested clauses rather than a comprehension.

enterprise/course_catalog_api.py
```diff
@@ -22,50 +24,93 @@
...
-    This is raised when we try to use the resources, rather than on
-    import, so that during testing we can load the module
-    and mock out the resources we need.
+    Object builds an API client to make calls to the Catalog API.
     """
 
-    pass
+    def __init__(self, user):
+        """
+        Create an Enrollment API client, authenticated with the API token from Django settings.
```
haikuginger: s/Enrollment/Course Catalog/

enterprise/course_catalog_api.py
```diff
@@ -22,50 +24,93 @@
...
-    if get_edx_api_data is None:
-        raise NotConnectedToOpenEdX(
-            _('To parse a catalog API response, this package must be installed in an OpenEdX environment.')
+    def get_program_by_title(self, program_title):
+        """
+        Return single program by name, or None if not found.
+        """
+        all_programs = self._load_data('programs')
+        try:
+            return next(program for program in all_programs if program.get('title') == program_title)
```
haikuginger: Should we return a single program in a scenario where multiple programs match the same title? Or should we raise an exception to complain about the ambiguity?

Better yet, we might want to yield each matching program from here, and then raise later on if either none of the programs are eligible for enrollment, or if multiple programs are eligible for enrollment.

test_utils/fake_catalog_api.py
```diff
@@ -0,0 +1,263 @@
...
+    """
+    return FAKE_PROGRAM_RESPONSES.get(program_uuid)
+
+
+def get_program_by_title(program_title):
+    """
+    Fake implementation returning program by title.
+    """
+    try:
+        return next(response for response in FAKE_PROGRAM_RESPONSES.values() if response["title"] == program_title)
```
haikuginger: Note that we'll need to change the implementation here to match the implementation in the actual code if that modification is made.

tests/admin/test_forms.py
```diff
@@ -19,29 +20,51 @@
...
+from enterprise.course_catalog_api import CourseCatalogApiClient
+from test_utils import fake_catalog_api, fake_enrollment_api
 from test_utils.factories import (EnterpriseCustomerFactory, EnterpriseCustomerUserFactory,
                                   PendingEnterpriseCustomerUserFactory, SiteFactory, UserFactory)
+from test_utils.fake_catalog_api import FAKE_PROGRAM_RESPONSE1, FAKE_PROGRAM_RESPONSE2
 
 FAKER = FakerFactory.create()
 
 
+class TestWithCourseCatalogApiMixin(object):
```
haikuginger: To make this a proper mixin, let's do `self.catalog_api = self._make_catalog_api()` in this class's `setUp` method. Or, even inline that code into this class's `setUp`.

tests/admin/test_forms.py
```diff
@@ -19,29 +20,51 @@
...
+        patcher = mock.patch('enterprise.admin.forms.CourseCatalogApiClient', mock.Mock(return_value=patched_api))
+        patcher.start()
+        self.addCleanup(patcher.stop)
+        return patched_api
+
+
 @mark.django_db
 @ddt.ddt
-class TestManageLearnersForm(unittest.TestCase):
+class TestManageLearnersForm(unittest.TestCase, TestWithCourseCatalogApiMixin):
```
haikuginger: Let's place `TestWithCourseCatalogApiMixin` before `unittest.TestCase` here. That way, its `setUp` method will be called.

tests/admin/test_forms.py
```diff
@@ -222,9 +245,114 @@ def test_clean_valid_course_invalid_mode(self, enrollment_client):
...
+        assert form.errors == {
+            "__all__": [ValidationMessages.COURSE_MODE_NOT_AVAILABLE.format(
+                mode=course_mode, program_title=FAKE_PROGRAM_RESPONSE2.get("title"), modes=", ".join({"professional"})
+            )]
+        }
+
 
 @mark.django_db
-class TestEnterpriseCustomerAdminForm(unittest.TestCase):
+class TestEnterpriseCustomerAdminForm(unittest.TestCase, TestWithCourseCatalogApiMixin):
```
haikuginger: Mixin changes as above.

tests/test_course_catalog_api.py
```diff
@@ -0,0 +1,200 @@
...
+        resource, resource_id = self._get_important_parameters(self.get_data_mock)
+
+        assert resource == "catalogs"
+        assert resource_id is None
+        assert actual_result == response_dict
+
+    @ddt.data(
+        "course-v1:JediAcademy+AppliedTelekinesis+T1",
+        "course-v1:TrantorAcademy+Psychohistory101+T1",
+        "course-v1:StarfleetAcademy+HyperspaceNavigation+T2337",
```
haikuginger: s/Hyperspace/Warpspeed/

😜

haikuginger:
@e-kolpakov, behavior is looking pretty good in general! One thing that would be great - you'll note that on the current master, we have javascript that disables the Mode dropdown when a valid course mode hasn't been entered. As a result, it's currently impossible to select a mode via the UI. It would be great if we could dynamically load the valid program modes into the list when a program ID or name is entered, but failing that, we should at least set the dropdown to have the default values and be enabled.

tests/test_course_catalog_api.py
e-kolpakov: @haikuginger Aw crap, right :)

test_utils/fake_catalog_api.py
e-kolpakov: @haikuginger it is a good point, but it might make more sense to keep it as is, since it is a fake method - it is supposed to be simpler than actual implementation. Also, it won't have such problem (at least for now), since the only two programs avaiable to it have different titles.

enterprise/course_catalog_api.py
e-kolpakov: @haikuginger I've added some protection against it (haven't pushed it yet), but title is a primary/unique key for course catalog program:

```
IntegrityError at /admin/course_metadata/program/add/
(1062, "Duplicate entry 'Program1' for key 'title'")
```

enterprise/admin/utils.py
e-kolpakov: @haikuginger That's just a two-level for comprehension with filter, nothin' serious :)

e-kolpakov:
@haikuginger I've addressed most of your comments - please take another look.

I plan to add tests for the new frontend functionality tomorrow, but please take a moment to look at it and comment on usability.

CHANGELOG.rst
```diff
@@ -14,6 +14,15 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[0.14.0] - 2017-01-15
```
mattdrayer: We've got another `0.14.0` in the works [here](https://github.com/edx/edx-enterprise/pull/30/files#diff-de0a9b67ffe2874a076b5fdb15bad484R17), just FYI to you and @saleem-latif

enterprise/admin/forms.py
```diff
@@ -14,10 +14,12 @@
 from django.utils.translation import ugettext as _
 
 from enterprise import utils
-from enterprise.admin.utils import ValidationMessages, email_or_username__to__email, validate_email_to_link
-from enterprise.course_catalog_api import get_all_catalogs
+from enterprise.admin.utils import (ProgramStatuses, ValidationMessages, email_or_username__to__email,
```
mattdrayer: At what point do we just import the entire `enterprise.admin.utils` package?

enterprise/admin/forms.py
```diff
@@ -35,6 +37,10 @@ class ManageLearnersForm(forms.Form):
         label=_("Also enroll these learners in this course"), required=False,
         help_text=_("Provide a course ID if enrollment is desired."),
     )
+    program = forms.CharField(
+        label=_("Or enroll these learners in this program"), required=False,
```
mattdrayer: Is this either/or selection (I guess it's actually Also/Or) enforced in the code?

enterprise/admin/forms.py
```diff
@@ -65,13 +71,25 @@ class Fields(object):
...
 
     class CsvColumns(object):
         """
         Namespace class for CSV column names.
         """
         EMAIL = "email"
 
+    def __init__(self, *args, **kwargs):
+        """
+        Initializes
```
mattdrayer: Would be great to have a more descriptive docstring here

enterprise/admin/forms.py
```diff
@@ -65,13 +71,25 @@ class Fields(object):
...
+
+        Arguments:
+            user(django.contrib.auth.models.User): current user
```
mattdrayer: Nit: Insert a space between user and parenthesis -> `user (django...`

enterprise/admin/forms.py
```diff
@@ -103,6 +121,37 @@ def clean_course(self):
...
+        Try obtaining program treating form value as program UUID or title.
+
+        Returns:
+            dict: Program information if program found
+        """
+        program_id = self.cleaned_data[self.Fields.PROGRAM].strip()
+        if not program_id:
+            return None
+
+        try:
```
mattdrayer: I'd prefer to see specific try/excepts versus wrapping all of the work we want to protect in one big try with multiple excepts at the end.

enterprise/admin/forms.py
```diff
@@ -144,7 +201,28 @@ def clean(self):
...
+            available_modes = client.get_common_course_modes(course_runs)
+            course_mode = self.cleaned_data.get(self.Fields.COURSE_MODE)
+
+            if not course_mode:
+                raise ValidationError(ValidationMessages.COURSE_WITHOUT_COURSE_MODE)
+            if course_mode not in available_modes:
+                raise ValidationError(ValidationMessages.COURSE_MODE_NOT_AVAILABLE.format(
+                    mode=course_mode, program_title=program.get("title"), modes=", ".join(available_modes)
+                ))
+        except (HttpClientError, HttpServerError):
```
mattdrayer: This except can be moved closer to the location being protected.

enterprise/admin/forms.py
```diff
@@ -185,7 +263,8 @@ def get_catalog_options(self):
         Once retrieved, these name pairs can be used directly as a value
         for the `choices` argument to a ChoiceField.
         """
-        catalogs = get_all_catalogs(self.user)
+        catalog_api = CourseCatalogApiClient(self.user)
```
mattdrayer: Do we need to protect against network/communication errors here?

enterprise/admin/utils.py
```diff
@@ -28,8 +38,19 @@ class ValidationMessages(object):
...
+        "Could not retrieve details for the program {program_id}. Please specify a valid ID or program name."
+    )
+    PROGRAM_IS_INACTIVE = _("Enrollment in program {program_id} is closed as it is in {status} status.")
+    MULTIPLE_PROGRAM_MATCH = _(
+        "Searching programs by title returned {program_count} programs. Please try using program UUID"
+    )
+    COURSE_MODE_NOT_AVAILABLE = _(
+        "Course mode {mode} is not available for all courses in program {program_title}. Available modes are {modes}."
+    )
+    FAILED_TO_OBTAIN_MODES = _("Failed to obtain available course modes for program {program_title}")
```
mattdrayer: `FAILED_TO_OBTAIN_PROGRAM_COURSE_MODES` ?

enterprise/admin/utils.py
```diff
@@ -28,8 +38,19 @@ class ValidationMessages(object):
     BOTH_FIELDS_SPECIFIED = _("Either \"Email or Username\" or \"CSV bulk upload\" must be specified, but both were.")
     BULK_LINK_FAILED = _("Bulk operation failed - no users were linked. Please correct the errors listed below.")
     COURSE_MODE_INVALID_FOR_COURSE = _("Enrollment mode {course_mode} not available for course {course_id}.")
-    COURSE_WITHOUT_COURSE_MODE = _("Please select a course enrollment mode for the given course.")
+    COURSE_WITHOUT_COURSE_MODE = _("Please select a course enrollment mode for the given course/program.")
     INVALID_COURSE_ID = _("Could not retrieve details for the course ID {course_id}. Please specify a valid ID.")
+    INVALID_PROGRAM_ID = _(
```
mattdrayer: I go back and forth on whether it's better to alphabetize or group by concept -- we need to do it one way or the other, however, and alphabetical is the current convention, so maybe we should stick with that.

enterprise/admin/views.py
```diff
@@ -322,13 +323,19 @@ def post(self, request, customer_uuid):
...
         # _handle_form might add form errors, so we check if it is still valid
         if manage_learners_form.is_valid():
-
-            # Enroll linked users in a course if requested
+            course_ids = []
             course_details = manage_learners_form.cleaned_data.get(ManageLearnersForm.Fields.COURSE)
             if course_details:
+                course_ids.append(course_details['course_id'])
+
+            program_detais = manage_learners_form.cleaned_data.get(ManageLearnersForm.Fields.PROGRAM)
```
mattdrayer: program_detais -> program_details

mattdrayer: Should we still do this work if there are course_details available from above?

mattdrayer: ah, I think I see now, nevermind

enterprise/course_catalog_api.py
```diff
@@ -22,50 +24,126 @@
...
+            course_run = self.get_course_run(course_run_id) or {}
+            course_run_modes = {seat.get('type') for seat in course_run.get('seats', [])}
+
+            if available_course_modes is None:
+                available_course_modes = course_run_modes
+            else:
+                available_course_modes &= course_run_modes
+
+            if not available_course_modes:
+                return available_course_modes
```
mattdrayer: If there are three course_runs and the first two do not have course_run_modes, but the third one does, will that information be missed due to this return statement?

enterprise/course_catalog_api.py
```diff
@@ -22,50 +24,126 @@
...
+
+        return available_course_modes
```
mattdrayer: I'm not really seeing any sort of commonality/reduction process here -- it seems like we actually mean "all defined modes", versus what is actually common across all course runs?

enterprise/utils.py
```diff
@@ -42,6 +42,33 @@ def __init__(self, *args, **kwargs):
...
+    pass
+
+
+class CourseCatalogApiError(Exception):
+    """
+    Exception to raise when we we received data from Course Catalog but it contained an error.
+    """
+
+
+class MultipleProgramMatch(CourseCatalogApiError):
```
mattdrayer: Seems like this class name should end in either Error or Exception

enterprise/course_catalog_api.py
e-kolpakov: @mattdrayer `available_course_modes` is a set. `&=` on line 126 is apply `&` to lhs and rhs and store in lhs. `&` on sets is defined as intersection of the operands, so it does the filtering.

There should be tests to check that this returns only modes available across all course runs.

CHANGELOG.rst
e-kolpakov: @mattdrayer thanks. No one knows for sure which PR will be merged first, so it makes little sense to try and predict what version should one's PR have

enterprise/admin/forms.py
e-kolpakov: @mattdrayer if you mean `from admin.utils import *` - probably never as pylint will complain. If you mean just `import admin.utils` and prefixing utils members - probably someday. But the better approach might be to split utils into more granular modules or (to the opposite) impose some classes to contain logically related utils.

enterprise/admin/forms.py
e-kolpakov: @mattdrayer only on frontend so far, but it's just because I didn't have time to implement it yesterday.

enterprise/admin/forms.py
e-kolpakov: @mattdrayer sure. Looks like I've started typing a docstring, got distracted and forgot about it :)

enterprise/course_catalog_api.py
e-kolpakov: @mattdrayer in this scenario, since the first two does not have run modes the right result would be an empty set. I'll add a test for this case.

e-kolpakov:
@haikuginger @mattdrayer Please take another look

enterprise/static/enterprise/admin/manage_learners.js
```diff
@@ -70,24 +83,43 @@ function loadPage() {
...
+    var programEnrollment = {
+        $control: $("#id_program"),
+        oldValue: $("#id_program").val(),
+        timeout: null,
+        // TODO: pull data from edx-enterprise endpoint (to be created) or course catalog API
+        apply: function() { applyModes(defaultModes); }
+    };
+    
+    var defaultModes = [
+        {slug: "", name: "---------"},
```
haikuginger: Nit: Should these be marked for i18n?

haikuginger:
:+1:

- [x] I tested this: I was able to enroll users in programs
- [x] I read through the code

e-kolpakov:
@asadiqbal08 @saleem-latif @mattdrayer Ready for you reviev.

enterprise/admin/forms.py
mattdrayer: @e-kolpakov -- just checking -- are you planning on addressing this suggestion or leaving as-is?

enterprise/admin/forms.py
mattdrayer: Checking on this one as well -- `try/except`?

enterprise/admin/views.py
```diff
@@ -322,13 +323,19 @@ def post(self, request, customer_uuid):
...
         if manage_learners_form.is_valid():
-
-            # Enroll linked users in a course if requested
+            course_ids = []
             course_details = manage_learners_form.cleaned_data.get(ManageLearnersForm.Fields.COURSE)
+            program_details = manage_learners_form.cleaned_data.get(ManageLearnersForm.Fields.PROGRAM)
+
             if course_details:
+                course_ids.append(course_details['course_id'])
+            elif program_details:
```
mattdrayer: Is it ever possible for both course_details and program_details to exist simultaneously?

enterprise/course_catalog_api.py
mattdrayer: Okay -- it might be helpful for future developers if we explain what is going on here with some code comments.

enterprise/course_catalog_api.py
mattdrayer: Cool, thanks for the explanation, I had a hard time finding a solid description of `&=` on my own.  I think I get it now.  So ultimately the end result will contain only those modes which appear across all course runs.

mattdrayer:
Just a couple things remaining for me -- @asadiqbal08 and/or @saleem-latif please take a look as well.

e-kolpakov:
@mattdrayer @saleem-latif @asadiqbal08 rebasing from f01d687704ad041b622655955fffe20809dab896 onto master

enterprise/admin/forms.py
e-kolpakov: @mattdrayer Question to @saleem-latif as author of this code - I only switched it to use `CourseCatalogApiClient` as part of refactoring.

enterprise/admin/views.py
e-kolpakov: @mattdrayer No, now it is prevented by frontend and form validation

enterprise/__init__.py
```diff
@@ -4,6 +4,6 @@
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.13.0"
+__version__ = "0.14.0"
```
e-kolpakov: @asadiqbal08 I'm still rebasing :)

e-kolpakov:
@asadiqbal08 Rebased

asadiqbal08:
@e-kolpakov sorry I could not see the version update in `edx-enterprise/enterprise/__init__.py`.

e-kolpakov:
@asadiqbal08 Good catch - I've missed that one.

e-kolpakov:
@asadiqbal08 bumped

asadiqbal08:
👍

e-kolpakov:
@mattdrayer Final check before merging - good to go>

mattdrayer:
:+1: :ship:

e-kolpakov:
@mattdrayer :rocket:  --> :artificial_satellite:  --> :ship:  --> :boom:

PR #30, opened 2017-01-17
ENT-112: Add read-only enterprise API so that IDAs can get enterprise related data

saleem-latif:
@mattdrayer , @asadiqbal08 , @zubair-arbi 

This PR adds enterprise API so that IDAs such as `ecommerce` may be able to fetch enterprise customer and enterprise learner data including data sharing consent for coupon application and similar tasks.


bradenmacdonald:
Cool :)

enterprise/api/v1/urls.py
```diff
@@ -0,0 +1,20 @@
+"""
+URl definitions for enterprise api version 1 endpoint.
```
mattdrayer: URl -> URL

enterprise/api/v1/urls.py
```diff
@@ -0,0 +1,20 @@
...
+"""
+from __future__ import absolute_import, unicode_literals
+
+from rest_framework.routers import DefaultRouter
+
+from enterprise.api.v1 import views
+
+router = DefaultRouter()  # pylint: disable=invalid-name
```
mattdrayer: Why does pylint dislike `router`?

enterprise/api/throttles.py
```diff
@@ -0,0 +1,29 @@
...
+
+        Returns True if the request is coming from the service user, and
+        defaults to UserRateThrottle's configured setting otherwise.
+        """
+        service_users = {
+            getattr(settings, 'ECOMMERCE_SERVICE_WORKER_USERNAME', None),
+            getattr(settings, 'ENTERPRISE_SERVICE_WORKER_USERNAME', None),
+        }
+        if request.user.username in service_users:
+            return True
```
mattdrayer: I'm going back and forth on whether or not we should eliminate the throttle entirely, or have a second throttle specific to the service user that allows for a greater number of requests/second but is not unbounded.  Thoughts?

enterprise/utils.py
```diff
@@ -166,3 +168,23 @@ def get_catalog_admin_url_template():
 
     # Return matched FQDN from catalog api url appended with catalog admin path
     return match.group("fqdn").rstrip("/") + "/admin/catalogs/catalog/{catalog_id}/change/"
+
+
+def get_enterprise_branding_info_by_provider_id(provider_id=None):  # pylint: disable=invalid-name
+    """
+    Return the EnterpriseCustomer branding information based on provider_id.
+
+    Provider_id: There is 1:1 relation b/w EnterpriseCustomer and Identity provider.
```
mattdrayer: Provider_id -> provider_id

mattdrayer: actually, Provider_id -> identity_provider_id

mattdrayer:
Looking good @saleem-latif -- just a couple questions and nits from me -- let's have @zubair-arbi and @asadiqbal08 take a pass through, then we can ask OpenCraft to take a closer look.

saleem-latif:
@mattdrayer, @bradenmacdonald, @e-kolpakov Seems like `djangorestframework<3.4` is not compatible with `django>=1.10` and we need `djangorestframework==3.2.3` because of [edx-platform's DRF version](https://github.com/edx/edx-platform/blob/master/requirements/edx/base.txt#L40) Any thoughts on how to fix this ?

enterprise/api/v1/urls.py
saleem-latif: pylint considers module level variables `constants` and expects them to be uppercased as specified [here](https://www.python.org/dev/peps/pep-0008/#constants).

enterprise/api/throttles.py
saleem-latif: @mattdrayer I have modified `ServiceUserThrottle` class so that instead of eliminating throttle we may specify some throttle rate for service users in settings file, let me me know if this is what you had in mind ?

enterprise/api/v1/urls.py
mattdrayer: Ah, I see -- can we add a code comment explaining this?

enterprise/api/throttles.py
```diff
@@ -15,15 +15,40 @@ class ServiceUserThrottle(UserRateThrottle):
...
-            return True
+            self.update_throttle_scope()
+
         return super(ServiceUserThrottle, self).allow_request(request, view)
+
+    def update_throttle_scope(self):
+        """
+        Update throttle scope so that service user throttle rates are applied.
+        """
+        self.scope = 'service_user'
```
mattdrayer: Should this scope value be moved to a setting or constant?

mattdrayer:
@haikuginger FYI -- here is the API  pull request

haikuginger:
This API doesn't currently support OAuth authentication; could you enable that, please? As [here](https://github.com/edx/edx-platform/blob/master/lms/djangoapps/course_structure_api/v0/views.py#L32).

saleem-latif:
@haikuginger I have added OAuth authentication, can you verify it is now working on your PR ?

CHANGELOG.rst
```diff
@@ -14,6 +14,15 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[0.14.0] - 2017-01-18
```
asadiqbal08: what ever goes first, need to update it too . 

ref: https://github.com/edx/edx-enterprise/pull/33/files#diff-de0a9b67ffe2874a076b5fdb15bad484R17

enterprise/api.py
```diff
@@ -1,27 +0,0 @@
-# -*- coding: utf-8 -*-
-"""
-Helper functions for enterprise app.
-"""
-from __future__ import absolute_import, unicode_literals
-
-from enterprise.models import EnterpriseCustomerBrandingConfiguration
-
-
-def get_enterprise_branding_info_by_provider_id(provider_id=None):  # pylint: disable=invalid-name
```
asadiqbal08: @saleem-latif we need to log a ticket or a new PR to update edx-platform code according to these changes. 
ref: https://github.com/edx/edx-platform/blob/master/common/djangoapps/util/enterprise_helpers.py#L10

enterprise/api.py
```diff
@@ -1,27 +0,0 @@
...
-    Return the EnterpriseCustomer branding information based on provider_id.
-
-    Provider_id: There is 1:1 relation b/w EnterpriseCustomer and Identity provider.
-    """
-    return EnterpriseCustomerBrandingConfiguration.objects.filter(
-        enterprise_customer__enterprise_customer_identity_provider__provider_id=provider_id
-    ).first()
-
-
-def get_enterprise_branding_info_by_ec_uuid(ec_uuid=None):  # pylint: disable=invalid-name
```
asadiqbal08: same here too. 

ref: https://github.com/edx/edx-platform/blob/master/common/djangoapps/util/enterprise_helpers.py#L10

CHANGELOG.rst
saleem-latif: I have updated PR https://github.com/edx/edx-platform/pull/14340 according to your suggestion.

enterprise/api/v1/views.py
```diff
@@ -0,0 +1,123 @@
...
+class EnterpriseCustomerViewSet(viewsets.ReadOnlyModelViewSet):
+    """
+    API views for `enterprise customer` api endpoint.
+    """
+    queryset = models.EnterpriseCustomer.active_customers.all()
+    serializer_class = serializers.EnterpriseCustomerSerializer
+    filter_backends = (filters.OrderingFilter, filters.DjangoFilterBackend)
+    permission_classes = (permissions.IsAuthenticated,)
+    authentication_classes = (OAuth2Authentication, SessionAuthentication,)
+    throttle_classes = (ServiceUserThrottle,)
```
asadiqbal08: @saleem-latif I am seeing the same things in all other views. 
Is it possible to move the common things such as `filter_backends, permission_classes`, `authentication_classes`, `throttle_classes` in separate class that is inherited with `viewsets.ReadOnlyModelViewSet`

haikuginger:
This is working with OAuth now, thanks.

enterprise/utils.py
```diff
@@ -166,3 +168,23 @@ def get_catalog_admin_url_template():
 
     # Return matched FQDN from catalog api url appended with catalog admin path
     return match.group("fqdn").rstrip("/") + "/admin/catalogs/catalog/{catalog_id}/change/"
+
+
+def get_enterprise_branding_info_by_provider_id(identity_provider_id=None):  # pylint: disable=invalid-name
+    """
+    Return the EnterpriseCustomer branding information based on provider_id.
+
+    identity_provider_id: There is 1:1 relation b/w EnterpriseCustomer and Identity provider.
```
mattdrayer: Please use the proper Arguments/Returns convention for docstrings

requirements/base.txt
```diff
@@ -16,3 +16,7 @@ requests==2.12.3          # via slumber
 six==1.10.0               # via django-extensions
 slumber==0.7.1            # via edx-rest-api-client
 unicodecsv==0.14.1
+djangorestframework==3.2.3
```
mattdrayer: Maybe I'm not really understanding how dependency resolution works here, but in edx-platform the version of Django is pinned to [1.8.17](https://github.com/edx/edx-platform/blame/master/requirements/edx/base.txt#L41), so it seems odd that we would be able to specify 1.10.3 for this package and expect it to work.

I'm wondering if it's possible to downgrade the Django dependency here to 1.8.17 in order to address the DRF 3.2.3 conflict, or if doing that would cause issues with some of the other Django-based features we are utilizing, such as some of the advanced Admin tooling.  Or perhaps downgrading Django is not the right solution at all?  I don't see us bumping edx-platform to 1.10+ any time soon.

@bradenmacdonald @haikuginger @e-kolpakov would be great to get your take on this.

tox.ini
```diff
@@ -26,6 +26,10 @@ deps =
     django18: Django>=1.8,<1.9
     django19: Django>=1.9,<1.10
     django110: Django>=1.10,<1.11
+    djangorestframework32: djangorestframework>=3.2.3,<3.3.0
```
mattdrayer: I have to admit I am able to understand what is going on with the version conflict solution, but I'm not clear on the specifics.  I will try to find someone who can verify these specific changes -- @cpennington would you be able to help us here?  If not, can you recommend someone?

saleem-latif:
@e-kolpakov,  @bradenmacdonald , @mattdrayer, @asadiqbal08   kindly take a look at the PR now, it is stable now.

mattdrayer:
👍

enterprise/api/urls.py
```diff
@@ -0,0 +1,11 @@
+"""
+URl definitions for enterprise API endpoint.
```
asadiqbal08: nit: `URL`

enterprise/api/v1/views.py
```diff
@@ -0,0 +1,125 @@
...
+from django.contrib.auth.models import User
+from django.contrib.sites.models import Site
+
+from enterprise import models
+from enterprise.api.filters import EnterpriseCustomerUserFilterBackend
+from enterprise.api.throttles import ServiceUserThrottle
+from enterprise.api.v1 import serializers
+
+
+class EnterpriseReadOnlyModelViewSet(viewsets.ReadOnlyModelViewSet):
```
asadiqbal08: good

asadiqbal08:
@saleem-latif Good Work 👍

cpennington:
Is `edx-enterprise` a standalone django project? Or is it a django app that gets installed into other projects? 

If it's the latter, it shouldn't have pinned requirements. Instead, it should have version ranges that for the libraries that it has known specific version dependencies.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = {py27,py35}-django{18,19,110}
+envlist = {py27,py35}-django{18,19}-djangorestframework{32}-djangofilter{011}, {py27,py35}-django{110}-djangorestframework{35}-djangofilter{10}
```
cpennington: You don't need the `{}` around single values. You only need them if you're doing something comma-separated.

Also, this seems like a lot of different variations. If particular versions of DRF and django-filter are needed to be compatible with django 1.10, you can just specify them using the same `django110` factor, rather than giving them their own factors.

bradenmacdonald:
> @bradenmacdonald @haikuginger @e-kolpakov would be great to get your take on this.

What @cpennington said makes a lot of sense to me.

tox.ini
saleem-latif: @cpennington I have updated it, can you take a look now ?

saleem-latif:
@cpennington `edx-enterprise` is a django app that can only be installed in `edx-platform` (It will raise [`NotConnectedToEdX`](https://github.com/edx/edx-enterprise/blob/master/enterprise/views.py#L51) anywhere else).

It will soon be converted to an IDA but until it is done we have to provide support for `djangorestframwork > 3.2.0, < 3.3.0`,  and since `djangorestframwork < 3.4.0` does not work on `django>1.10.0`, we have separate DRF dependencies for `django 1.8, 1.9` and separate ones for `django 1.10`.

mattdrayer:
Thanks everyone for your feedback on this PR -- @saleem-latif please address what you can and then squash+merge when ready.

tox.ini
```diff
@@ -24,8 +24,14 @@ norecursedirs = .* docs requirements
 [testenv]
 deps =
     django18: Django>=1.8,<1.9
+    django18: djangorestframework>=3.2.3,<3.3.0
+    django18: django-filter>=0.11.0,<1.0.0
     django19: Django>=1.9,<1.10
+    django19: djangorestframework>=3.2.3,<3.3.0
+    django19: django-filter>=0.11.0,<1.0.0
```
cpennington: I know I'm late to the party, but you can consolidate this: `django1{8,9}: djangorestframework>=3.2.3, <3.3.0`

requirements/base.txt
```diff
@@ -16,3 +16,8 @@ requests==2.12.3          # via slumber
 six==1.10.0               # via django-extensions
 slumber==0.7.1            # via edx-rest-api-client
 unicodecsv==0.14.1
+djangorestframework==3.5.3
+django-filter==1.0.1
+edx-django-oauth2-provider==1.1.4
+djangorestframework-oauth==1.1.0
+edx-drf-extensions==1.2.2
```
e-kolpakov: @saleem-latif @asadiqbal08 @zubair-arbi @haikuginger @bradenmacdonald Please don't modify requirements files directly and don't approve/merge PRs that contain such changes. Spent ~ 20 minutes trying to figure out why my new tests does not run properly only to find out that `make upgrade && make requirements` removes a couple of required packages and find out which ones.

In particular, `edx-django-oauth2-provider`, `djangorestframework-oauth` and `edx-drf-extensions` was misisng, since they are not listed in the `base.in`

mattdrayer: Sorry about that, @e-kolpakov -- I guess it wasn't clear to us until now the relationship between `base.txt` and `base.in` -- we certainly meant well 😉 live and learn!

PR #34, opened 2017-01-26
ENT-101: Modify view and templates to allow for course-specific consent questioning

haikuginger:
**Description:**

This pull request updates the data sharing consent views and templates to enable it to meet the requirements of collecting data sharing consent for a particular course.

**JIRA:** [ENT-101](https://openedx.atlassian.net/browse/ENT-101)

**Dependencies:** None

**Merge deadline:** 30 January 2017

**Testing instructions:**

See [testing instructions for edx-platform PR #14400](https://github.com/edx/edx-platform/pull/14400).

**Reviewers:**
- [ ] @smarnach 
- [ ] edX TBD

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:** Additional tests are in progress, and I need to figure out the best strategy to prevent the form from redirecting to an arbitrary destination.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=34"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

haikuginger:
Actually, given the tight timeline for this PR, it'd be great if we could get concurrent review from @mattdrayer, @asadiqbal08, and @saleem-latif. See also: the linked edx-platform PR.

enterprise/templates/enterprise/grant_data_sharing_permissions.html
```diff
@@ -31,7 +31,11 @@
                         </h2>
                     </div>
                     <span class="text">
-                        ${_("In order to link your account, we need consent to share your course data with {provider}.").format(provider=sso_provider)}
+                        % if not course_id:
+                            ${_("In order to link your account, we need consent to share your course data with {provider}.").format(provider=ec_name)}
+                        % else:
+                            ${_("In order to access {course_name}, we need consent to share your course data with {provider}.").format(provider=ec_name, course_name=course_name)}
```
mattdrayer: It depends on the configuration for a given Enterprise / Identity Provider -- it's possible to force people to provide consent when attempting to access the system through the SSO door.

enterprise/templates/enterprise/grant_data_sharing_permissions.html
```diff
@@ -31,7 +31,11 @@
                         </h2>
                     </div>
                     <span class="text">
-                        ${_("In order to link your account, we need consent to share your course data with {provider}.").format(provider=sso_provider)}
+                        % if not course_id:
```
mattdrayer: I think it would be better to have a specific parameter to control this behavior instead of using `course_id` as the switch

enterprise/templates/enterprise/grant_data_sharing_permissions.html
```diff
@@ -43,10 +47,20 @@
                             data-warning-message="${ messages['warning'] }"
                         />
                         <label for="register-data_sharing_consent">
-                            ${_("I agree to allow {platform_name} to share data about my enrollment, completion and performance in all {platform_name} courses and programs where my enrollment is sponsored by {provider}.*").format(provider=sso_provider, platform_name=platform_name)}
+                            % if not course_id:
```
mattdrayer: Same here -- we should have a dedicated switch instead of hanging on `course_id`

enterprise/templates/enterprise/grant_data_sharing_permissions.html
```diff
@@ -43,10 +47,20 @@
...
+                                ${_("I agree to allow {platform_name} to share data about my enrollment, completion and performance in all {platform_name} courses and programs where my enrollment is sponsored by {provider}.*").format(provider=ec_name, platform_name=platform_name)}
+                            % else:
+                                ${_("I agree to allow {platform_name} to share data about my enrollment, completion and performance in {course_name} with {provider}.*").format(provider=ec_name, course_name=course_name, platform_name=platform_name)}
+                            % endif
                         </label>
                     </div>
                     <input type="hidden" name="csrfmiddlewaretoken" value="${ csrf_token }" />
+                    % if course_id:
```
mattdrayer: We should use the dedicated switch here as well, although we may also want to keep the check for `course_id` unless it's safe to assume that it will always be present in that context.

enterprise/templates/enterprise/grant_data_sharing_permissions.html
```diff
@@ -43,10 +47,20 @@
...
+                        <input type="hidden" name="course_id" value="${ course_id }" />
+                    % endif
+                    % if redirect_url:
```
mattdrayer: I'm assuming that `redirect_url` is orthogonal to the status of the switch being employed above?

enterprise/views.py
```diff
@@ -106,12 +122,54 @@ def get_note(provider, required):
...
+        try:
+            client = CourseApiClient()
+            course_details = client.get_course_details(course_id)
+        except Exception:
+            raise Http404
+        next_url = request.GET.get('next')
+
+        customer = get_object_or_404(EnterpriseCustomerUser, user_id=request.user.id).enterprise_customer
+        if not consent_necessary_for_course(request.user, course_id):
+            raise Exception
```
mattdrayer: also we should consider logging, if that makes sense in this context.

enterprise/views.py
```diff
@@ -106,12 +122,54 @@ def get_note(provider, required):
...
+        platform_name = configuration_helpers.get_value("PLATFORM_NAME", settings.PLATFORM_NAME)
+        course_name = course_details['name']
+        data = {
+            'platform_name': platform_name,
+            'data_sharing_consent': 'required',
```
mattdrayer: is it correct to specify a static value of `required` here, versus a constant, setting, or customer-specfic value?

enterprise/views.py
```diff
@@ -106,12 +122,54 @@ def get_note(provider, required):
...
+            'ec_name': customer.name,
+        }
+        return render_to_response('grant_data_sharing_permissions.html', data, request=request)
+
+    def get_account_consent(self, request):
+        """
+        Render a form to collect consent for account-wide data sharing.
         """
-        # Verify that all necessary resources are present
-        verify_edx_resources()
```
mattdrayer: So we're dropping this verification work from the `post`, but it is still present below in the `get` -- just wondering why we need it there but not here.

enterprise/views.py
```diff
@@ -127,27 +185,52 @@ def get(self, request):
...
             },
+            "course_id": None
         }
         return render_to_response('grant_data_sharing_permissions.html', data, request=request)
 
-    def post(self, request):
+    def get(self, request):
         """
-        Process the above form.
+        Render a form to collect user input about data sharing consent.
```
mattdrayer: I'm not sure if it should be in the docstring or in a code comment, but I think it's important to explain the choice being made here based on the presence of a course parameter and what the difference in behavior for the two branches.

enterprise/views.py
```diff
@@ -127,27 +185,52 @@ def get(self, request):
...
+        course = request.GET.get('course_id')
+        if course:
+            return self.get_course_specific_consent(request, course)
+        return self.get_account_consent(request)
+
+    def post_course_specific_consent(self, request, course_id, consent_provided):
+        """
+        Save user-provided consent directives for a particular course enrollment.
+        """
+        if not request.user.is_authenticated():
```
mattdrayer: It's not entirely clear to me when authentication is required and when it is not -- I can see we are performing a check here but not in the other `post` operations.

enterprise/templates/enterprise/grant_data_sharing_permissions.html
```diff
@@ -31,7 +31,11 @@
                         </h2>
                     </div>
                     <span class="text">
-                        ${_("In order to link your account, we need consent to share your course data with {provider}.").format(provider=sso_provider)}
```
mattdrayer: Was this the only dependency on `sso_provider` that needed to be modified to work with `ec_name`?  Also I'd prefer the more verbose `enterprise_customer_name` vs. `ec_name`.

mattdrayer:
@srpearce if you can please take a spin through this PR as well to review the learner-facing messages that would be fantastic.

enterprise/templates/enterprise/grant_data_sharing_permissions.html
haikuginger: Yup, it was. Residue from when this was inside edx-platform.

enterprise/views.py
haikuginger: Yeah, I should add this check to the other `post` operation.

enterprise/views.py
haikuginger: Sounds good!

enterprise/views.py
haikuginger: This is still in `post`; the changes just aren't very conducive to an easily-readable diff.

enterprise/views.py
haikuginger: I believe we'll only be rendering this view when consent is required, but it wouldn't hurt to change.

enterprise/views.py
```diff
@@ -106,12 +122,54 @@ def get_note(provider, required):
...
+        ).format(provider=provider, course_name=course_name)
+
+    def get_course_specific_consent(self, request, course_id):
+        """
+        Render a form with course-specific information about data sharing consent.
+        """
+        try:
+            client = CourseApiClient()
+            course_details = client.get_course_details(course_id)
+        except Exception:
```
haikuginger: Yeah, I'm fixing this to cover `HttpClientError`.

enterprise/templates/enterprise/grant_data_sharing_permissions.html
haikuginger: Yes. We don't use it in the at-login mode right now, but I suspect we'll need to once enterprise spins out into being a full IDA.

enterprise/templates/enterprise/grant_data_sharing_permissions.html
haikuginger: Sounds good- same for below.

enterprise/views.py
haikuginger: Yeah, this is being swapped out to an Http404.

enterprise/templates/enterprise/grant_data_sharing_permissions.html
haikuginger: @smarnach: In general, that'll be the case. However, in this particular case, the user has _already been enrolled_ in the course by the Enterprise admin. In cases where the user is manually enrolling in a course and has a coupon applied by the EnterpriseCustomer, there should be an option to opt-out.

enterprise/views.py
```diff
@@ -106,12 +122,54 @@ def get_note(provider, required):
...
+            raise Http404
+        next_url = request.GET.get('next')
+
+        customer = get_object_or_404(EnterpriseCustomerUser, user_id=request.user.id).enterprise_customer
```
haikuginger: Not in a database-guaranteed way, but the application logic prevents linking a user to multiple EnterpriseCustomers.

Right now, I'm not sure there's a good way to handle the case where there are multiple EnterpriseCustomerUsers anyway; the form can only handle providing consent to a single one of them.

I think I can narrow down the _potential_ for issues by doing this; let me know what you think:

```python
customer = get_object_or_404(
    EnterpriseCourseEnrollment,
    enterprise_customer_user__user_id=request.user.id,
    course_id=course_id
).enterprise_customer_user.enterprise_customer
```

srpearce:
@mattdrayer Will do, hopefully tomorrow morning.

bradenmacdonald:
@haikuginger  Since this PR changes UX, can you please add screenshots?

enterprise/templates/enterprise/grant_data_sharing_permissions.html
```diff
@@ -31,7 +31,11 @@
...
+                        % if not course_id:
+                            ${_("In order to link your account, we need consent to share your course data with {provider}.").format(provider=ec_name)}
```
srpearce: Before your account can be linked, you must agree to share your course data with {provider}.

enterprise/templates/enterprise/grant_data_sharing_permissions.html
srpearce: Before you can access {course_name}, you must agree to share your course data with {provider}.

enterprise/templates/enterprise/grant_data_sharing_permissions.html
```diff
@@ -43,10 +47,20 @@
                             data-warning-message="${ messages['warning'] }"
                         />
                         <label for="register-data_sharing_consent">
-                            ${_("I agree to allow {platform_name} to share data about my enrollment, completion and performance in all {platform_name} courses and programs where my enrollment is sponsored by {provider}.*").format(provider=sso_provider, platform_name=platform_name)}
+                            % if not course_id:
+                                ${_("I agree to allow {platform_name} to share data about my enrollment, completion and performance in all {platform_name} courses and programs where my enrollment is sponsored by {provider}.*").format(provider=ec_name, platform_name=platform_name)}
+                            % else:
+                                ${_("I agree to allow {platform_name} to share data about my enrollment, completion and performance in {course_name} with {provider}.*").format(provider=ec_name, course_name=course_name, platform_name=platform_name)}
```
srpearce: I'm assuming that Legal has provided this text. It looks fine to me.

enterprise/views.py
```diff
@@ -91,6 +93,20 @@ def get_warning(provider, platform, required):
...
+        """
+        Get a course-specific warning for the form.
+
+        Arguments:
+            provider: The name of the linked EnterpriseCustomer
+            course_name: The name of the course in question
+        """
+        return _(
+            "Are you sure? If you do not agree to share your data with {provider}, you will "
+            "not be able to access {course_name}."
```
srpearce: you will not be able to access --> you cannot access

enterprise/views.py
```diff
@@ -106,12 +122,54 @@ def get_note(provider, required):
...
-        Render a form to collect user input about data sharing consent.
+        Get a course-specific note for the form.
+
+        Arguments:
+            provider: The name of the linked EnterpriseCustomer
+            course_name: The name of the course in question
+        """
+        return _(
+            "Data sharing consent is required for courses provided by {provider}, including "
+            "{course_name}; if consent is not provided, you will be returned to the Dashboard."
```
srpearce: Courses from {provider} require data sharing consent. If you do not agree to share your data, you will be redirected to your dashboard.

enterprise/templates/enterprise/grant_data_sharing_permissions.html
mattdrayer: @griffresch see this comment regarding the term "manual enrollment" ^^  (no issue/worries @haikuginger, just a point I'm trying to make)

enterprise/templates/enterprise/grant_data_sharing_permissions.html
srpearce: Since I can't comment directly on lines 1-33, I'm adding comments here.

**l. 14**
Data sharing consent required

**l. 29**
OK

enterprise/views.py
haikuginger: Yeah, a static requirement is fine here - higher up in the block, we raise a 404 if consent isn't required.

enterprise/views.py
```diff
@@ -30,9 +30,11 @@
...
-from enterprise.models import EnterpriseCustomer, EnterpriseCustomerUser, UserDataSharingConsentAudit
+from enterprise.lms_api import CourseApiClient
+from enterprise.models import (EnterpriseCourseEnrollment, EnterpriseCustomer, EnterpriseCustomerUser,
+                               UserDataSharingConsentAudit)
 from enterprise.tpa_pipeline import active_provider_enforces_data_sharing, get_enterprise_customer_for_request
-from enterprise.utils import NotConnectedToEdX
+from enterprise.utils import NotConnectedToEdX, consent_necessary_for_course
 
 
 def verify_edx_resources():
```
srpearce: The doc strings in lines 53, 86-87, and 91-92 are the same as some of the doc strings in enterprise/conf/locale/en/LC_MESSAGES/django.po. I'll make the same comments here:

**l. 53**
OpenEdX --> Open edX

**ll. 86-87**
OK

**ll. 91-92**
get --> receive, strike "any"

enterprise/views.py
haikuginger: Actually, we can't do this for the other POST operation, based on the way `.is_authenticated()` gets patched in. However, the TPA pipeline shouldn't let us get to that point unless there is, in fact, a real user.

enterprise/templates/enterprise/grant_data_sharing_permissions.html
haikuginger: Legal provided the existing platform-wide consent text, which I modified for the course-specific prompt.

haikuginger:
Rebased from 88652e7. @mattdrayer, ready for one last review. Still working a bit on the edx-platform PR.

mattdrayer:
LGTM -- 👍 after rebase, squash, green checks, etc.

haikuginger:
Squashed, etc, from 8849481.

PR #39, opened 2017-02-01
[WIP] ENT-163: SuccessFactors 

brittneyexline:
This is my first stab at defining the views and models I think we'll need to implement the integration with SuccessFactors. There is no logic yet in the views but I've outlined what I imagine should happen.

This is my first time contributing here, and I'm far from a python expert, so I'm happy to get any feedback.


enterprise/models.py
```diff
@@ -764,3 +774,37 @@ def __repr__(self):
         Return uniquely identifying string representation.
         """
         return self.__str__()
+
+
+class EnterpriseIntegratedChannel(TimeStampedModel):
+    """
+    EnterpriseIntegratedChannel is an external system that the Enterprise Customer uses to manage enterprise
+    related data. Each of these requires an edx plugin in order to send data to these systems, this model stores
+    information related to what and where to send data from the enterprise app.
```
bradenmacdonald: Cool, I like this approach.

enterprise/models.py
```diff
@@ -764,3 +774,37 @@ def __repr__(self):
...
+    """
+    id = models.PositiveIntegerField(primary_key=True)
+    name = models.CharField(max_length=255, blank=False, null=False, help_text=_("Third Party name."))
+    data_type = models.CharField(max_length=100, blank=False, null=False, help_text=_("Data Type"))
+    plugin_endpoint = models.CharField(max_length=255, blank=False, null=False, help_text=_("Internal Plugin endpoint."))
```
bradenmacdonald: From the code so far, I can't really tell what `data_type` and `plugin_endpoint` are used for or what sort of values they accept.

enterprise/models.py
```diff
@@ -300,6 +302,9 @@ class PendingEnterpriseCustomerUser(TimeStampedModel):
     enterprise_customer = models.ForeignKey(EnterpriseCustomer, blank=False, null=False)
     user_email = models.EmailField(null=False, blank=False, unique=True)
 
+    class Meta(object):
+        app_label = 'enterprise'
```
bradenmacdonald: I'm curious why you had to add these `app_label` entries?

integrated_channel_plugin_sapsf/models.py
```diff
@@ -0,0 +1,67 @@
+from __future__ import unicode_literals
+
+from django.db import models
+from django.utils.encoding import python_2_unicode_compatible
+from model_utils.models import TimeStampedModel, TimeFramedModel
+
+
+# Create your models here.
```
bradenmacdonald: You can remove this comment ^

enterprise/views.py
```diff
@@ -294,3 +297,59 @@ def post(self, request):
...
+
+        For a given learner who has completed a course, find out the following:
+            if the learner is associated with an enterprise customer.
+            if the course is part of an enterprise enrollment.
+            if the enterprise customer is configured for SuccessFactors.
+
+        Continue if all the conditions above are met, otherwise abort processing.
+
+        Based on the EnterpriseCustomerPluginConfiguration for the Enterprise, determine which endpoints are
+        configured to receive learner data, and send a formatted request to each of those endpoints.
```
bradenmacdonald: Is this all done synchronously or will there be celery workers that are calling these endpoints?

integrated_channel_plugin_sapsf/models.py
```diff
@@ -0,0 +1,67 @@
...
+
+class OCNWebServicesConfiguration(TimeStampedModel):
```
bradenmacdonald: What does OCN stand for?

integrated_channel_plugin_sapsf/models.py
```diff
@@ -0,0 +1,67 @@
...
+    class Meta:
+        app_label = 'integrated_channel_plugin_sapsf'
+
+
+class OCNWebServicesEnterpriseCustomerConfiguration(TimeStampedModel):
+    """
+    The Enterprise specific configuration we need for integrating with SuccessFactors.
+    """
+
+    enterprise_customer_uuid = models.UUIDField(unique=True)
```
bradenmacdonald: Since this is part of the same codebase as `enterprise`, can we just make this a `models.ForeignKey(EnterpriseCustomer)`? That's much more robust and then cascades deletes, etc.  Or is there a possibility of this part being split off to become yet another IDA?

integrated_channel_plugin_sapsf/models.py
```diff
@@ -0,0 +1,67 @@
...
+    sapsf_base_url = models.CharField(max_length=255)
+    key = models.TextField(blank=True, verbose_name="Client ID")
+    secret = models.TextField(blank=True, verbose_name="Client Secret")
```
bradenmacdonald: How long are the keys and secrets? I would think that `CharField` is a better fit for that type of data. (Do you want a textarea or a single-line input in the django admin?)

integrated_channel_plugin_sapsf/models.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class CompletionStatusEventAudit(models.Model):
+    """
+    The payload we sent to SuccessFactors at a given point in time for an enterprise course enrollment.
+    """
+
+    enterprise_course_enrollment_id = models.PositiveIntegerField(blank=False, null=False)
+    sapsf_user_id = models.CharField(max_length=255, blank=False, null=False)
+    course_id = models.CharField(max_length=255, blank=False, null=False)
```
bradenmacdonald: I wonder if we should be using `openedx.core.djangoapps.xmodule_django.models.CourseKeyField` instead of `CharField` - I'm guessing not since that's part of edx-platform, but it could be split out into a separate codebase. (Just thinking aloud here - no need to change this if that would be a lot of work)

integrated_channel_plugin_sapsf/models.py
```diff
@@ -0,0 +1,67 @@
...
+    course_completed = models.BooleanField(default=True)
+    completed_timestamp = models.BigIntegerField()  # we send a UNIX timestamp to SuccessFactors
```
bradenmacdonald: Totally optional suggestion: if you want, you can remove the `course_completed` field and consider the course to be completed when `completed_timestamp` is non-null:

```python
@property
def course_completed(self):
    return self.completed_timestamp is not None
```

enterprise/models.py
brittneyexline: I have been getting warnings when starting up the LMS in devstack:
RemovedInDjango19Warning: Model class enterprise.models.EnterpriseCustomerIdentityProvider doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.

enterprise/models.py
brittneyexline: I wanted to make data_type an enum, it's supposed to be something like 'course' or 'learner', which are the two types of data so far that we are concerned with sending to an external system. I feel like there is probably a better way to express plugin_endpoint, which is meant to represent our internal endpoint for forwarding data to a particular channel's plugin. It's not clear yet if these will just be function calls or be made through api calls, so I'm not sure what this should look like yet.

The idea is you could configure multiple channels to receive data for a given enterprise, and all of the channels that are of the same data type would have plugins that follow a common interface for that data type and deal with the details of sending to a specific channel.

integrated_channel_plugin_sapsf/models.py
brittneyexline: Open Content Network, which is the part of SuccessFactors that we're integrating with. I'm open to changing the name since it's really information about how to talk with SAP SuccessFactors.

integrated_channel_plugin_sapsf/models.py
brittneyexline: There is a possibility that this could be split off. For simplicity and ease of development I've put it in the enterprise repo for now.

integrated_channel_plugin_sapsf/models.py
brittneyexline: I don't know yet, but you are probably right that CharField will do. I copied this from another Oauth configuration model from third_party_auth.

integrated_channel_plugin_sapsf/models.py
brittneyexline: Yes, the possibility of splitting off is one reason. Another is that conceptually, this is what SuccessFactor's course identifier is for our course. But we are the ones who tell them what the course identifiers are so they will likely be the same unless there is some reason we need to modify them for sending to SuccessFactors.

enterprise/views.py
brittneyexline: This will probably be done asynchronously. I'm not that familiar with celery so I'd like to learn more about how we use them for asynchronous processing.

enterprise/views.py
```diff
@@ -294,3 +297,59 @@ def post(self, request):
...
+    def post(self, request):
+        """
+        Hit when certificates are generated or when audit learners in self paced courses reach a passing grade.
+
+        When this happens, this method will route the learners data appropriately.
+
+        For a given learner who has completed a course, find out the following:
+            if the learner is associated with an enterprise customer.
+            if the course is part of an enterprise enrollment.
+            if the enterprise customer is configured for SuccessFactors.
```
e-kolpakov: @brittneyexline ... is configured for *integration channel*? I believe this part should not reference any particular integration channel.

enterprise/views.py
e-kolpakov: @brittneyexline I believe the following docs should provide you with a good starting point:

* http://docs.celeryproject.org/en/latest/userguide/tasks.html
* http://docs.celeryproject.org/en/latest/userguide/calling.html

In short, the idea is to have edx-enterprise provide a celery task that would notify integration endpoints. The task would be called asynchronously (with `task.apply_async`). But please make sure task result (and, especially, any errors happened) are not just discarded, but at least logged. Better yet to have it exposed to admin - as a DB model + admin integration, but that's slightly more involving, so it would be nice to have, but log record would be sufficient.

enterprise/views.py
```diff
@@ -294,3 +297,59 @@ def post(self, request):
...
+
+        Continue if all the conditions above are met, otherwise abort processing.
+
+        Based on the EnterpriseCustomerPluginConfiguration for the Enterprise, determine which endpoints are
+        configured to receive learner data, and send a formatted request to each of those endpoints.
+        If nothing is configured, abort processing.
+        Note that the request format will be uniform across the different plugin endpoints.
+        """
+        return Response(request.data)
```
e-kolpakov: @brittneyexline It doesn't seem to do anything - just echo the request. Is it supposed to be so? If that's just a dummy implementation supposed to be overridden by nested classes, I'd suggest replacing it with `raise NotImplementedError("Must override in descendent class")` or something like that.

enterprise/views.py
```diff
@@ -294,3 +297,59 @@ def post(self, request):
...
+        """
+        Trigger sending an enterprise's course catalog data to any configured endpoints.
+
+        For a given enterprise customer, based on the EnterpriseCustomerPluginConfiguration for the Enterprise,
+        determine which plugin endpoints are configured to receive course data, and send a formatted request to
+        each of those endpoints.
+        If nothing is configured, abort processing.
+        Note that the request format will be uniform across the different plugin endpoints.
+        """
+        return Response(request.data)
```
e-kolpakov: @brittneyexline another echo endpoint?

integrated_channel_plugin_sapsf/models.py
```diff
@@ -0,0 +1,135 @@
...
+    created = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        app_label = 'integrated_channel_plugin_sapsf'
+
+    def __str__(self):
+        """
+        Return human-readable string representation.
+        """
+        return '<CompletionStatusEventAudit {} for enterprise enrollment {}, SAP user {}, and course {}>'.format(
```
e-kolpakov: @brittneyexline `__str__` exhibits the same behavior everywhere, should we  just replace it with some generic mixin (just thinking out load as well, no need to rush into implementing this).

integrated_channel_plugin_sapsf/views.py
```diff
@@ -0,0 +1,89 @@
...
+
+        Transform the data into the format expected by SuccessFactors.
+
+        Send the transformed learner data to SuccessFactors using the configured endpoint
+        (taken from OCNWebServicesConfiguration and OCNWebServicesEnterpriseCustomerConfiguration) and the oauth token.
+
+        Record information about the success/failure of posting the learner data to SuccessFactors in
+         CompletionStatusEventAudit
+        """
+        return Response(request.data)
```
e-kolpakov: @brittneyexline This method does not actually do any of the actions listed in the docstring. If that's a stub and intended to be extended further in scope of later PR, please replace it (and all other similar methods) with
   
     raise NotImplementedError

or return 501 NOT IMPLEMENTED

e-kolpakov:
@brittneyexline Also, please make sure CI build is green. Something test-related is dramatically broken - almost all tests error - which indicates some low-level problem (and most likely relatively simple solution)

enterprise/views.py
brittneyexline: It is meant to be a stub, so I'll switch it to raise the NotImplementedError

integrated_channel_plugin_sapsf/models.py
brittneyexline: I was copying what I have seen in the enterprise models, so it might be a broader discussion to introduce a mixin, I don't have a strong opinion about this either way.

brittneyexline:
Thanks @e-kolpakov for the review. I will address the test issues - I think part of it is missing migrations but I'm also having trouble getting things to pass locally with master, so I have some environmental issues to sort out.

This probably won't be the final iteration with respect to the views (since none of them are actually implemented), but I would like to focus on getting the models in a good enough state to merge.

enterprise/models.py
mattdrayer: It does seem odd that the system is complaining about this particular set of models.  We don't seem to have to do this with other apps/packages -- I wonder if we're missing something?

mattdrayer: It does seem odd that the system is complaining about this particular set of models.  We don't seem to have to do this with other apps/packages -- I wonder if we're missing something?

enterprise/models.py
```diff
@@ -764,3 +775,82 @@ def __repr__(self):
...
+
+    def __repr__(self):
+        """
+        Return uniquely identifying string representation.
+        """
+        return self.__str__()
+
+
+@python_2_unicode_compatible
+class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
```
mattdrayer: I'm wondering if this should be set up as a ConfigurationModel?

enterprise/views.py
```diff
@@ -294,3 +296,47 @@ def post(self, request):
...
+
+        Based on the EnterpriseCustomerPluginConfiguration for the Enterprise, determine which endpoints are
+        configured to receive learner data, and send a formatted request to each of those endpoints.
+        If nothing is configured, abort processing.
+        Note that the request format will be uniform across the different plugin endpoints.
+        """
+        raise NotImplementedError
+
+
+class ExportCourseCatalog(APIView):
```
mattdrayer: What's the difference between "routing" and "exporting"?  In both cases are these methods actually doing the transmitting, or are they selecting the proper integration channel / plugin and handing off?

integrated_channel_plugin_sapsf/models.py
```diff
@@ -0,0 +1,135 @@
...
+
+from django.db import models
+from django.utils.encoding import python_2_unicode_compatible
+from model_utils.models import TimeStampedModel, TimeFramedModel
+
+
+# Create your models here.
+
+@python_2_unicode_compatible
+class OCNWebServicesConfiguration(TimeStampedModel):
```
mattdrayer: As above, should this be a ConfigurationModel?

integrated_channel_plugin_sapsf/models.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def __repr__(self):
+        """
+        Return uniquely identifying string representation.
+        """
+        return self.__str__()
+
+
+@python_2_unicode_compatible
+class OCNWebServicesEnterpriseCustomerConfiguration(TimeStampedModel):
```
mattdrayer: "Configuration" --> ConfigurationModel?

integrated_channel_plugin_sapsf/models.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def __repr__(self):
+        """
+        Return uniquely identifying string representation.
+        """
+        return self.__str__()
+
+
+@python_2_unicode_compatible
+class CompletionStatusEventAudit(models.Model):
```
mattdrayer: If we need audit-trail-like functionality you could review @e-kolpakov's work with django-simple-history in this repo

integrated_channel_plugin_sapsf/views.py
```diff
@@ -0,0 +1,88 @@
...
+        Fetch the remaining learner data we need to send to SuccessFactors. This includes:
+            The SuccessFactors user id which should be available from when the user was first authenticated via SSO.
+
+        Transform the data into the format expected by SuccessFactors.
+
+        Send the transformed learner data to SuccessFactors using the configured endpoint
+        (taken from OCNWebServicesConfiguration and OCNWebServicesEnterpriseCustomerConfiguration) and the oauth token.
+
+        Record information about the success/failure of posting the learner data to SuccessFactors in
+         CompletionStatusEventAudit
```
mattdrayer: If we are recording the status of the transmission of data, we might instead name this model `CompletionStatusEventTransmissionAudit`

integrated_channel_plugin_sapsf/models.py
mattdrayer: I don't know that it will be spun out as a separate IDA -- although I can certainly envision this as a standalone microservice.  The idea for now is to contain all of the SAP-SF communication logic in its own app and move it to a standalone code project/repo as soon as it makes sense.

mattdrayer: Along these lines, I'm not sure if it makes sense to include the concept of an Enterprise Customer in this particular app/package.  Really all we need is a unique reference identifier to be able to select the correct configuration based on the context.  Whether or not this identifier is for an enterprise, or some other concept, isn't really a concern of the package.  Thoughts?

enterprise/models.py
brittneyexline: Can you explain the benefit of doing that, and why we haven't used that for other configuration type models in enterprise (such as the branding config) so far?
Sounds reasonable I just want to understand the differences.

enterprise/views.py
brittneyexline: They are selecting the proper integration channel and handing off. 
I suppose there isn't really a difference in what they're doing, any thoughts on a different name for this action?

integrated_channel_plugin_sapsf/models.py
brittneyexline: That depends on whether or not we would ever want to use this package without being tied to the concept of an enterprise. In my mind that doesn't seem likely, especially because our definition of an enterprise is pretty broad.
I guess we could put this new identifier in the plugin configuration model in enterprise that references the id for this model, is that what you would envision?

enterprise/models.py
douglashall: http://stackoverflow.com/questions/29635765/django-1-9-deprecation-warnings-app-label

enterprise/models.py
```diff
@@ -764,3 +775,82 @@ def __repr__(self):
...
+    """
+
+    name = models.CharField(max_length=255, blank=False, null=False, help_text=_("Third Party name."))
+    data_type = models.CharField(max_length=100, blank=False, null=False, help_text=_("Data Type"))
+    plugin_endpoint = models.CharField(
+        max_length=255,
+        blank=False,
+        null=False,
+        help_text=_("Internal Plugin endpoint.")
+    )
```
douglashall: Could you provide an example of the data that would get stored using this model?

enterprise/models.py
```diff
@@ -764,3 +775,82 @@ def __repr__(self):
...
+
+@python_2_unicode_compatible
+class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
+    """
+    Store information related to integrating with external systems for a particular enterprise customer.
+
+    EnterpriseCustomerPluginConfiguration stores the relationship between EnterpriseCustomer
+    and EnterpriseIntegratedChannel, representing which IntegratedChannels are configured for an Enterprise and
+    what data to send where for a given Enterprise.
+    """
```
douglashall: Do you need a separate model for this? It looks like this is just storing the relationship. You could accomplish the same thing by adding a foreign key to either `EnterpriseCustomer` or `EnterpriseIntegratedChannel`.

enterprise/views.py
```diff
@@ -294,3 +296,47 @@ def post(self, request):
...
+
+    def post(self, request):
+        """
+        Hit when certificates are generated or when audit learners in self paced courses reach a passing grade.
+
+        When this happens, this method will route the learners data appropriately.
+
+        For a given learner who has completed a course, find out the following:
+            if the learner is associated with an enterprise customer.
+            if the course is part of an enterprise enrollment.
```
douglashall: What does this mean? How can a course be a part of an enterprise enrollment?

enterprise/views.py
douglashall: We should make it clear which part of the processing chain here is happening asynchronously. Since we are creating a web API endpoint here, I was assuming that LMS would add a task to the celery queue which posts to the RouteLearnerData endpoint and all of the work documented here would happen in the same process. It might be good to throw together a quick diagram which illustrates the process flow.

enterprise/views.py
douglashall: Maybe `PushCatalogDataToIntegratedChannel` and `PushLearnerDataToIntegratedChannel`? You might be able to create a super class that factors out some of the common stuff.

enterprise/models.py
```diff
@@ -764,3 +775,82 @@ def __repr__(self):
         Return uniquely identifying string representation.
         """
         return self.__str__()
+
+
+@python_2_unicode_compatible
+class EnterpriseIntegratedChannel(TimeStampedModel):
```
douglashall: Should this live in a separate `integrated_channels` app?

integrated_channel_plugin_sapsf/apps.py
```diff
@@ -0,0 +1,13 @@
+"""
+Enterprise Integrated Channel SAP SuccessFactors Django application initialization.
+"""
+from __future__ import unicode_literals
+
+from django.apps import AppConfig
+
+
+class IntegratedChannelPluginSapsfConfig(AppConfig):
```
douglashall: I wonder if there should just be an `integrated_channels` app in the edx-enterprise Django project. You could then have separate directories for each `IntegratedChannel` implementation.

integrated_channel_plugin_sapsf/models.py
douglashall: What is `OCNWebServices`?

integrated_channel_plugin_sapsf/models.py
douglashall: I would keep all code related to the implementation of the Success Factors API client as "extractable" as possible.

integrated_channel_plugin_sapsf/models.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+
+    enterprise_course_enrollment_id = models.PositiveIntegerField(blank=False, null=False)
+    sapsf_user_id = models.CharField(max_length=255, blank=False, null=False)
+    course_id = models.CharField(max_length=255, blank=False, null=False)
+    course_completed = models.BooleanField(default=True)
+    completed_timestamp = models.BigIntegerField()  # we send a UNIX timestamp to SuccessFactors
+    instructor_name = models.CharField(max_length=255, blank=True)
+    grade = models.CharField(max_length=100, blank=False, null=False)
+    error_message = models.TextField(blank=True)
```
douglashall: You might want to persist the full response from the SF API regardless of success or failure.

integrated_channel_plugin_sapsf/models.py
```diff
@@ -0,0 +1,135 @@
...
+    created = models.DateTimeField(auto_now_add=True)
```
douglashall: Is this model meant to enable replay of failed pushes to the SF API? If so, you might want a `status` field.

integrated_channel_plugin_sapsf/views.py
```diff
@@ -0,0 +1,88 @@
...
+Views for Enterprise Integrated Channel SAP SuccessFactors.
+"""
+
+# Create your views here.
+from __future__ import absolute_import, unicode_literals
+
+from rest_framework.views import APIView
+
+
+class SuccessFactorsLearnerData(APIView):
```
douglashall: Why is this a web API?

enterprise/models.py
e-kolpakov: @haikuginger @bdero You might have answers to the question above

integrated_channels/integrated_channel/views.py
```diff
@@ -0,0 +1,50 @@
...
+    def post(self, request):
+        """
+        Hit when certificates are generated or when audit learners in self paced courses reach a passing grade.
+
+        When this happens, this method will route the learners data appropriately.
+
+        For a given learner who has completed a course, find out the following:
+            if the learner is associated with an enterprise customer.
+            if the course is part of an enterprise enrollment.
+            if the enterprise customer is configured for an Integrated Channel.
```
mattdrayer: We should think about which check makes the most sense to do first -- for example, if the Enterprise Customer is not configured for an Integrated Channel it doesn't really make sense to do anything else.

mattdrayer: ...but if the learner isn't associated with an Enterprise Customer then I guess that trumps the EC check

integrated_channels/sap_success_factors/models.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def __repr__(self):
+        """
+        Return uniquely identifying string representation.
+        """
+        return self.__str__()
+
+
+@python_2_unicode_compatible
+class CourseTransmissionAudit(TimeFramedModel):
```
mattdrayer: Maybe call this CatalogTransmissionAudit?

integrated_channels/sap_success_factors/transmitters/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""
+Package for transmitting data to SuccessFactors.
+"""
+from __future__ import unicode_literals
+
+
+class BaseSuccessFactorsTransmitter(object):
```
mattdrayer: It depends on the language, but typically I would expect to see "Base" at the end of the name.

integrated_channels/sap_success_factors/transmitters/courses.py
```diff
@@ -0,0 +1,30 @@
+"""
+Class for transmitting course data to SuccessFactors.
+"""
+from __future__ import unicode_literals
+from integrated_channels.sap_success_factors.transmitters import BaseSuccessFactorsTransmitter
+
+
+class SuccessFactorsCourseTransmitter(BaseSuccessFactorsTransmitter):
+    """
+    This endpoint is intended to carry out an export of course data to SuccessFactors for a given Enterprise.
```
mattdrayer: We should probably be a little more specific about whether we are transmitting Course data or Course Run data

integrated_channels/sap_success_factors/views.py
```diff
@@ -0,0 +1,5 @@
+"""
+Views for Enterprise Integrated Channel SAP SuccessFactors.
+"""
+
+# Create your views here.
```
e-kolpakov: @brittneyexline nit: please remove this comment

integrated_channels/integrated_channel/views.py
brittneyexline: I'll be sure to keep this in mind when this gets implemented.

integrated_channels/sap_success_factors/transmitters/courses.py
brittneyexline: It will be course data - the api accepts a list of schedules, so we can send multiple course runs for a single course object on their side.

e-kolpakov:
@brittneyexline @mattdrayer @douglashall Part of mine fault as well as I've approved that, but that's again exactly what I asked to stop doing - another conflict in dependencies.

Now, a more significant one: django-config-models [does not support Django 1.10](https://github.com/edx/django-config-models/blob/master/setup.py#L49), so by adding it we're basically ruling out Django 1.10 for edx-enterprise, until django-config-models change.

I'm cmmenting Django 1.10-related code in edx-enterprise and logging a task to do something about it (i.e. either quit supporting 1.10 in edx-enteprise as well and cleanup the comments, or make django-config-models support Django 1.10)

mattdrayer:
@e-kolpakov is the issue that we have introduced django-config-models, or that we specified the requirement incorrectly?  If the latter, then perhaps the team needs some training on the proper way to handle dependencies.

bradenmacdonald:
@mattdrayer  Seems like in addition to the django 1.10 issue, there's actually a problem with django-config-models requirements - see https://github.com/edx/edx-enterprise/pull/44#issuecomment-279822832

I think the best way forward is to fix the DRF requirement in django-config-models and change it to support django 1.10

e-kolpakov:
@mattdrayer Adding DCM as dependency is not a problem itself, but DCM's dependencies is.

>  If the latter, then perhaps the team needs some training on the proper way to handle dependencies.

Might be a good thing to do. Actually, I've already added bullet point to regenerate requirements to the pull request template, but it's in #44, which is now blocked, so some time might pass until it is merged.

I think I'll just ping everyone I know working on edx-enterprise here so that they are aware of this problem and what to do.

So, @asadiqbal08 @bradenmacdonald @bdero @brittneyexline @haikuginger @mattdrayer @saleem-latif FYI this comment thread (starting [there](https://github.com/edx/edx-enterprise/pull/39#issuecomment-279694462). *tl;dr* Before submitting PR, do `make upgrade` to upgrade dependencies and resolve conflicts.

One other thing I've just realized is that no one actually updates dependencies in setup.py. That might not seem very important (since tests pass and so on), but I believe `setup.py` dependencies are actually *more* important that requirements/*.txt, as those are used to set up edx-enterprise with edx-platfrom (and any other app using `pip install edx-enterprise`).

I'll try to add a test (in #44) to check that `setup.py` dependencies list is roughly equal to what base.in (and conflict resolution files) generate, but not 100% sure how feasible the approach I have in mind will be (and if I have time to do that).

PR #36, opened 2017-01-30
[WIP] ENT-134: Display consent to share data for Coupon use

mtyaka:
**Description:** 

This PR is not complete yet. More info to follow.

**JIRA:** [ENT-134](https://openedx.atlassian.net/browse/ENT-134)

**Testing instructions:**

TBD

**Reviewers:**
- [ ] OpenCraft: @haikuginger 
- [ ] edX: TBD 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are (reasonably) squashed
- [ ] Translations are updated
- [ ] PR author is listed in AUTHORS

**Post merge:**
- [ ] Create a tag
- [ ] Check new version is pushed to PyPi after tag-triggered build is 
      finished.
- [ ] Delete working branch (if not needed anymore)

PR #40, opened 2017-02-03
ENT-134/ENT-162: Add option to defer enrollments in the course-specific consent gate

bdero:
This allows for ecommerce to collect consent by redirecting users
to the usual enterprise data consent page before the user has an
EnterpriseCourseEnrollment.

**JIRA tickets**: ENT-134.

**Discussions**: Architecture discussed extensively on the wiki and in meetings, then the revised proposal was presented to the Arch Council on Oct. 21 and given thumbs up.

**Dependencies**: None

**Testing instructions**:

Test this PR by following the testing instructions on https://github.com/edx/ecommerce/pull/1127.

**Reviewers**
- [x] @mtyaka 
- [ ] edX reviewer[s] TBD

### DIFF
--- a/enterprise/api/permissions.py
+++ b/enterprise/api/permissions.py
@@ -1,0 +1,26 @@
+"""
+Permission classes for the Enterprise API.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from rest_framework import permissions
+
+from enterprise.api.utils import get_service_usernames
+
+
+class IsServiceUserOrReadOnly(permissions.IsAuthenticated):
+    """
+    This request is authenticated as a service user, or is a read-only request.
+    """
+
+    def has_permission(self, request, view):
+        """
+        Enforce that the user is authenticated. If the request isn't read-only (unsafe), then deny permission unless
+        the user is a service user.
+        """
+        service_users = get_service_usernames()
+
+        return (
+            request.method in permissions.SAFE_METHODS or
+            request.user.username in service_users
+        )
--- a/enterprise/api/utils.py
+++ b/enterprise/api/utils.py
@@ -1,0 +1,19 @@
+"""
+Utility functions for the Enterprise API.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from django.conf import settings
+
+
+SERVICE_USERNAMES = (
+    'ECOMMERCE_SERVICE_WORKER_USERNAME',
+    'ENTERPRISE_SERVICE_WORKER_USERNAME'
+)
+
+
+def get_service_usernames():
+    """
+    Return the set of service usernames that are given extended permissions in the API.
+    """
+    return {getattr(settings, username, None) for username in SERVICE_USERNAMES}
--- a/enterprise/api/v1/urls.py
+++ b/enterprise/api/v1/urls.py
@@ -12,6 +12,11 @@ from enterprise.api.v1 import views
 router = DefaultRouter()  # pylint: disable=invalid-name
 router.register("site", views.SiteViewSet, 'site')
 router.register("auth-user", views.UserViewSet, 'auth-user')
+router.register(
+    "enterprise-course-enrollment",
+    views.EnterpriseCourseEnrollmentViewSet,
+    'enterprise-course-enrollment'
+)
 router.register("enterprise-customer", views.EnterpriseCustomerViewSet, 'enterprise-customer')
 router.register("enterprise-learner", views.EnterpriseCustomerUserViewSet, 'enterprise-learner')
 router.register("user-data-sharing-consent", views.UserDataSharingConsentAuditViewSet, 'user-data-sharing-consent')
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -15,11 +15,12 @@ from django.contrib.sites.models import Site
 
 from enterprise import models
 from enterprise.api.filters import EnterpriseCustomerUserFilterBackend
+from enterprise.api.permissions import IsServiceUserOrReadOnly
 from enterprise.api.throttles import ServiceUserThrottle
 from enterprise.api.v1 import serializers
 
 
-class EnterpriseReadOnlyModelViewSet(viewsets.ReadOnlyModelViewSet):
+class EnterpriseModelViewSet(object):
     """
     Base class for attribute and method definitions common to all view sets.
     """
@@ -27,6 +28,20 @@ class EnterpriseReadOnlyModelViewSet(viewsets.ReadOnlyModelViewSet):
     permission_classes = (permissions.IsAuthenticated,)
     authentication_classes = (OAuth2Authentication, SessionAuthentication, BearerAuthentication, JwtAuthentication)
     throttle_classes = (ServiceUserThrottle,)
+
+
+class EnterpriseReadOnlyModelViewSet(EnterpriseModelViewSet, viewsets.ReadOnlyModelViewSet):
+    """
+    Base class for all read only Enterprise model view sets.
+    """
+    pass
+
+
+class EnterpriseReadWriteModelViewSet(EnterpriseModelViewSet, viewsets.ModelViewSet):
+    """
+    Base class for all read/write Enterprise model view sets.
+    """
+    permission_classes = (permissions.IsAuthenticated, IsServiceUserOrReadOnly,)
 
 
 class EnterpriseCustomerViewSet(EnterpriseReadOnlyModelViewSet):
@@ -42,6 +57,27 @@ class EnterpriseCustomerViewSet(EnterpriseReadOnlyModelViewSet):
     )
     filter_fields = FIELDS
     ordering_fields = FIELDS
+
+
+class EnterpriseCourseEnrollmentViewSet(EnterpriseReadWriteModelViewSet):
+    """
+    API views for `enterprise course enrollment` api endpoint.
+    """
+    queryset = models.EnterpriseCourseEnrollment.objects.all()
+
+    FIELDS = (
+        'enterprise_customer_user', 'consent_granted', 'course_id'
+    )
+    filter_fields = FIELDS
+    ordering_fields = FIELDS
+
+    def get_serializer_class(self):
+        """
+        Use a special serializer for any requests that aren't read-only.
+        """
+        if self.request.method in ('GET', ):
+            return serializers.EnterpriseCourseEnrollmentReadOnlySerializer
+        return serializers.EnterpriseCourseEnrollmentWriteSerializer
 
 
 class SiteViewSet(EnterpriseReadOnlyModelViewSet):
@@ -70,12 +106,11 @@ class UserViewSet(EnterpriseReadOnlyModelViewSet):
     ordering_fields = FIELDS
 
 
-class EnterpriseCustomerUserViewSet(EnterpriseReadOnlyModelViewSet):
+class EnterpriseCustomerUserViewSet(EnterpriseReadWriteModelViewSet):
     """
     API views for `enterprise customer user` api endpoint.
     """
     queryset = models.EnterpriseCustomerUser.objects.all()
-    serializer_class = serializers.EnterpriseCustomerUserSerializer
     filter_backends = (filters.OrderingFilter, filters.DjangoFilterBackend, EnterpriseCustomerUserFilterBackend)
 
     FIELDS = (
@@ -83,6 +118,14 @@ class EnterpriseCustomerUserViewSet(EnterpriseReadOnlyModelViewSet):
     )
     filter_fields = FIELDS
     ordering_fields = FIELDS
+
+    def get_serializer_class(self):
+        """
+        Use a flat serializer for any requests that aren't read-only.
+        """
+        if self.request.method in ('GET', ):
+            return serializers.EnterpriseCustomerUserReadOnlySerializer
+        return serializers.EnterpriseCustomerUserWriteSerializer
 
     @detail_route()
     def entitlements(self, request, pk=None):  # pylint: disable=invalid-name,unused-argument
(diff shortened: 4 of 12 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=40&repo=edx%2Fedx-enterprise"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

enterprise/views.py
```diff
@@ -231,13 +244,15 @@ def post_course_specific_consent(self, request, course_id, consent_provided):
...
-        )
+        enrollment_deferred = request.POST.get('enrollment_deferred')
+        if enrollment_deferred is None:
+            EnterpriseCourseEnrollment.objects.update_or_create(
+                enterprise_customer_user__user_id=request.user.id,
+                course_id=course_id,
+                defaults={
+                    'consent_granted': consent_provided,
+                }
+            )
```
mtyaka: Nit: instead of handling both standard and deferred enrollment in `post_course_specific_consent` it might be cleaner to implement a separate `post_deferred_course_specific_consent` handler and invoke it based on the presence of the `enrollment_defferred` parameter from the generic `post` handler below.

bdero: Hmm, I can't think of a way to separate this out that would make it easier to understand. It's one condition gating one statement that needs to stay in the same execution order relative to the surrounding checks. So if I factor it out, most of the code would be repeated, and then if I upstreamed the common code to `post()`, the functions would vanish and everything would just be inlined.
Do you think it would suffice to add a comment explaining `enrollment_deferred` instead?

bdero:
@mtyaka This PR now has tests.

enterprise/views.py
```diff
@@ -152,14 +152,25 @@ def get_course_specific_consent(self, request, course_id):
         except HttpClientError:
             raise Http404
         next_url = request.GET.get('next')
+        enrollment_deferred = request.GET.get('enrollment_deferred')
```
mattdrayer: Nit: Move this line to below the empty line that follows

enterprise/views.py
```diff
@@ -152,14 +152,25 @@ def get_course_specific_consent(self, request, course_id):
...
+
+        if enrollment_deferred is None:
+            customer = get_object_or_404(
```
mattdrayer: We don't really have much in the docstring about the error states -- it would be good to include something about the possible 404(s) that can be thrown.

mattdrayer:
Looks pretty good to me so far @bdero -- just a couple small things so far

enterprise/views.py
bdero: Done.

enterprise/views.py
bdero: Alrighty, I added this information to the docstring.

enterprise/views.py
bdero: Alright, thanks!

tests/test_views.py
```diff
@@ -402,30 +413,34 @@ def test_get_course_specific_consent(
...
+                        "warning": (
+                            "Are you sure? If you do not agree to share your data "
+                            "with Starfleet Academy, you cannot access edX Demo Course."
+                        ),
+                    },
+                    "course_id": "course-v1:edX+DemoX+Demo_Course",
+                    "course_name": "edX Demo Course",
+                    "redirect_url": "https://google.com",
+                    "enterprise_customer_name": ecu.enterprise_customer.name,
+                    "course_specific": True
```
bdero: Yes we should, thanks for pointing it out. I added it to the dict.

tests/test_views.py
```diff
@@ -575,18 +596,21 @@ def test_post_course_specific_consent(self, reverse_mock, *args):  # pylint: dis
...
         )
         reverse_mock.return_value = '/dashboard'
+        post_data = {
+            'course_id': course_id,
+            'data_sharing_consent': data_sharing_consent,
+            'redirect_url': '/successful_enrollment',
+        }
+        if enrollment_deferred:
+            post_data['enrollment_deferred'] = True
         resp = self.client.post(
```
bdero: Fixed, thanks.

bdero:
@mattdrayer I've addressed all of your comments and this is ready for another pass!

tests/test_views.py
```diff
@@ -402,30 +413,37 @@ def test_get_course_specific_consent(
...
-                "course_name": "edX Demo Course",
-                "redirect_url": "https://google.com",
-                "enterprise_customer_name": ecu.enterprise_customer.name,
-                "course_specific": True
-        }.items():
-            assert response.context[key] == value  # pylint:disable=no-member
+        if enrollment_deferred:
+            url += '&enrollment_deferred=true'
+        if supply_customer_uuid:
+            url += '&enterprise_id={uuid}'.format(uuid=enterprise_customer.uuid)
```
bdero: Ah, good idea. Yes, it's a request factory so this should work. I changed this up to use the data param.

tests/test_views.py
```diff
@@ -402,30 +413,37 @@ def test_get_course_specific_consent(
...
+        response = self.client.get(url)
+        assert response.status_code == expected_status_code
+        if expected_status_code == 200:
```
bdero: Done.

enterprise/api/utils.py
```diff
@@ -0,0 +1,16 @@
...
+
+from django.conf import settings
+
+
+def get_service_usernames():
+    """
+    Return the set of service usernames that are given extended permissions in the API.
+    """
+    return {
+        getattr(settings, 'ECOMMERCE_SERVICE_WORKER_USERNAME', None),
```
bdero: Seems reasonable, so I changed it. Thanks!

enterprise/api/permissions.py
```diff
@@ -0,0 +1,30 @@
...
+    def has_permission(self, request, view):
+        """
+        Enforce that the user is authenticated. If the request isn't read-only (unsafe), then deny permission unless
+        the user is a service user.
+        """
+        authenticated = super(IsServiceUserOrReadOnly, self).has_permission(request, view)
+
+        service_users = get_service_usernames()
+
+        return (
```
mattdrayer: I know that this is a perfectly acceptable way to write code, but imho it's much less readable than doing the work prior to the return and then returning a simple answer.  In addition to readability I find it harder to write tests for implementations like this as well.  It's just MHO, but wanted to toss it out there.

enterprise/api/utils.py
```diff
@@ -0,0 +1,19 @@
+"""
+Utility functions for the Enterprise API.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from django.conf import settings
+
+
+WORKER_USERNAMES = (
```
mattdrayer: are these `WORKER` usernames, `SERVICE` usernames, or `SERVICE_WORKER` usernames?  I know it's semantics, but might as well be clear on what we're dealing with here.

enterprise/api/v1/urls.py
```diff
@@ -13,6 +13,11 @@
 router.register("site", views.SiteViewSet, 'site')
 router.register("auth-user", views.UserViewSet, 'auth-user')
 router.register("enterprise-customer", views.EnterpriseCustomerViewSet, 'enterprise-customer')
+router.register(
```
mattdrayer: You could bump this up one line to satisfy the alphabetization spirits 😄

enterprise/api/v1/views.py
```diff
@@ -27,6 +28,20 @@ class EnterpriseReadOnlyModelViewSet(viewsets.ReadOnlyModelViewSet):
...
 
 
+class EnterpriseReadOnlyModelViewSet(EnterpriseModelViewSet, viewsets.ReadOnlyModelViewSet):
+    """
+    Base class for all read only Enterprise model view sets.
+    """
+    pass
+
+
+class EnterpriseReadWriteModelViewSet(EnterpriseModelViewSet, viewsets.ModelViewSet):
```
mattdrayer: Deja vu!  Does this PR cover https://github.com/edx/edx-enterprise/pull/36#discussion-diff-98638742R30?

enterprise/api/v1/views.py
```diff
@@ -82,6 +110,14 @@ class EnterpriseCustomerUserViewSet(EnterpriseReadOnlyModelViewSet):
     filter_fields = FIELDS
     ordering_fields = FIELDS
 
+    def get_serializer_class(self):
+        """
+        Use a flat serializer for any requests that aren't read-only.
+        """
+        if self.request.method in ('GET', ):
+            return serializers.EnterpriseCustomerUserSerializer
```
mattdrayer: We've been specific about Read vs. Write in other areas -- do we need to follow-suit with this serializer as well and name it `EnterpriseCustomerUserReadSerializer` (or ReadOnlySerializer)?

mattdrayer:
Last pass for me, @bdero -- I left some questions about a couple things but in general it is looking good to me -- if you can respond to my comments with either changes or justifications we can probably get this over the line soon.

enterprise/api/v1/views.py
bdero: @mattdrayer It looks like this PR overlaps heavily with that one. I didn't base this PR on it, so I accidentally rewrote some of the work in https://github.com/edx/ecommerce/pull/1127 as I completed this. :/

@mtyaka Do you think this PR fulfills the stuff in #36? If not, I will try to rebase this PR on top of that one and add in the new permission class.
Really sorry about the overlap I caused here @mtyaka - I should have checked more carefully that #36 had one of the API views needed for ENT-162, and then I should have rebased this PR on top of it before starting work on ENT-162.

enterprise/api/permissions.py
bdero: @mattdrayer Makes sense, I changed it to this:

```python
    def has_permission(self, request, view):
        """
        Enforce that the user is authenticated. If the request isn't read-only (unsafe), then deny permission unless
        the user is a service user.
        """
        authenticated = super(IsServiceUserOrReadOnly, self).has_permission(request, view)
        if not authenticated:
            return False

        service_users = get_service_usernames()

        return (
            request.method in permissions.SAFE_METHODS or
            request.user.username in service_users
        )
```

enterprise/api/v1/urls.py
bdero: Done.

enterprise/api/utils.py
bdero: I think it would make sense to call them "service usernames" as it seems to be the clearest of the three, especially since the users are not exclusively in the context of async workers. I changed the variable name.

enterprise/api/v1/views.py
bdero: Done, I renamed it to `EnterpriseCustomerUserReadOnlySerializer`.

mattdrayer:
Alright, this is in good shape -- good to merge after we figure out what is going on between this PR and #36, and Jenkins turns happy -- 👍

enterprise/api/v1/views.py
mtyaka: @bdero No worries, I closed #36. It looks like this PR contains everything required and is far more complete than #36.

enterprise/views.py
```diff
@@ -232,13 +255,15 @@ def post_course_specific_consent(self, request, course_id, consent_provided):
...
-            course_id=course_id,
-            defaults={
-                'consent_granted': consent_provided,
-            }
-        )
+        enrollment_deferred = request.POST.get('enrollment_deferred')
+        if enrollment_deferred is None:
+            EnterpriseCourseEnrollment.objects.update_or_create(
+                enterprise_customer_user__user_id=request.user.id,
+                course_id=course_id,
```
asadiqbal08: Are we not going to sure either the course exist with the given course_id here ?

bdero: Good catch, I added a check that should result in a 404 if the course doesn't exist using the CourseApiClient (just like the corresponding GET method does).

bdero:
@mattdrayer This is ready for your continued review.

The EnterpriseCourseEnrollment POST now directs to a [new serializer](https://github.com/edx/edx-enterprise/pull/40/files#diff-9b920c17601896b079cd031907580cb8R87) that replaces the checks during validation.

With the current design, the LMS enrollment view will be making a single POST to it in order to create an EnterpriseCourseEnrollment. The logistics of finding the necessary EnterpriseCustomerUser is taken care of in the [username validator](https://github.com/edx/edx-enterprise/pull/40/files#diff-9b920c17601896b079cd031907580cb8R99)

However, I have another concern about the design: Can we assume that when the LMS enrollment post is called and there's an EnterpriseCustomerUser that already exists for the user, that consent was already granted?
In the flow for coupons, consent is [collected and verified in Ecommerce during coupon redemption](https://github.com/edx/ecommerce/pull/1127/files#diff-c797ef6707ae3dfb1ee8828592244a9eR169), but the LMS doesn't have that information currently when making the post to the enrollment view.

If we can't assume course-specific consent was granted just because an EnterpriseCustomerUser exists (which I suspect we can't), one option would be to also pass the consent parameter when we call the [LMS enrollment view](https://github.com/edx/ecommerce/pull/1127/files#diff-c9f38bb98fd490cc1127bb1cbdd272a3R231), and simply have the LMS enrollment view do the EnterpriseCourseEnrollment post whenever [`has_api_key_permissions`](https://github.com/edx/edx-platform/blob/master/common/djangoapps/enrollment/views.py#L531) is true, and forward the given consent value as the [`content_granted`](https://github.com/edx/edx-enterprise/pull/40/files#diff-9b920c17601896b079cd031907580cb8R121) param.

enterprise/views.py
mtyaka: Should we perform `course_id` validation in `EnterpriseCourseEnrollmentWriteSerializer` as well?

bdero:
@mattdrayer Could you let me know if you have any more comments for this?

mattdrayer:
Made another pass thru just now -- it LRGTM -- let's go ahead and 🚢 -- can always fix issues in a follow-on change set.   Travis appears to be complaining about something and we should squash commits before merging -- nice work 👍

bdero:
@mattdrayer Thanks!

e-kolpakov:
@bdero Missing CHANGELOG record - please add as a separate PR without version bump.

PR #43, opened 2017-02-07
ENT-181: Add API endpoint to fetch enterprise learner entitlements

saleem-latif:
Hi @mattdrayer , @asadiqbal08 ,

Kindly take a look at the PR, it adds an API endpoint to fetch enterprise learner's entitlements.

__Instructions for Testing:__

1. Add an Enterprise Customer with the following configurations.
    1.  Enable data sharing consent: [x] (Checked)
    2.  Enforce data sharing consent: 'at_enrollment'
    3. Add an enterprise customer entitlement using section "Enterprise Customer Entitlements" at the bottom
2. Create three user accounts and associate them with the enterprise customer using enterprise learner management console at "http://localhost:8000/admin/enterprise/enterprisecustomer/{enterprise_customer_uuid}/manage_learners"
3. Make sure that user-1 has consented to data sharing, user-2 has not consented to data sharing and user-3 has not set any value. (If user is created and then associated with an enterprise customer its data sharing consent will be unset). Other User's data sharing consent can be set by logging in via enterprise SSO.
4. Open learner entitlements page at "http://localhost:8000/enterprise/api/v1/enterprise-learner/{enterprise_customer_id}/entitlements/". (Note: Replace {enterprise_customer_id} with enterprise customer user's id)
5. Verify that 
     1. for user-1 all entitlements are displayed
     2. for user-2  and user-3 entitlements returned are an empty list.
6. Now, open enterprise customer details page and update the following settings
    1.  Enable data sharing consent: [x] (Checked)
    2.  Enforce data sharing consent: 'optional'
7. Verify that entitlements are now displayed for all user-1, user-2, user-3. URL for enteprise learner entitlements is http://localhost:8000/enterprise/api/v1/enterprise-learner/{enterprise_customer_id}/entitlements/. (Note: Replace {enterprise_customer_id} with enterprise customer user's id)

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -97,3 +97,27 @@ class EnterpriseCustomerUserSerializer(serializers.ModelSerializer):
     user = UserSerializer()
     enterprise_customer = EnterpriseCustomerSerializer()
     data_sharing_consent = UserDataSharingConsentAuditSerializer(many=True)
+
+
+class EnterpriseCustomerUserEntitlementSerializer(serializers.Serializer):
+    """
+    Serializer for the entitlements of EnterpriseCustomerUser.
+
+    This Serializer is for read only endpoint of enterprise learner's entitlements
+    It will ignore any state changing requests like POST, PUT and PATCH.
+    """
+    entitlements = serializers.ListField(
+        child=serializers.DictField()
+    )
+
+    def create(self, validated_data):
+        """
+        Do not perform any operations for state changing requests.
+        """
+        pass
+
+    def update(self, instance, validated_data):
+        """
+        Do not perform any operations for state changing requests.
+        """
+        pass
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -6,6 +6,8 @@ from __future__ import absolute_import, unicode_literals
 from edx_rest_framework_extensions.authentication import BearerAuthentication, JwtAuthentication
 from rest_framework import filters, permissions, viewsets
 from rest_framework.authentication import SessionAuthentication
+from rest_framework.decorators import detail_route
+from rest_framework.response import Response
 from rest_framework_oauth.authentication import OAuth2Authentication
 
 from django.contrib.auth.models import User
@@ -81,6 +83,26 @@ class EnterpriseCustomerUserViewSet(EnterpriseReadOnlyModelViewSet):
     )
     filter_fields = FIELDS
     ordering_fields = FIELDS
+
+    @detail_route()
+    def entitlements(self, request, pk=None):  # pylint: disable=invalid-name,unused-argument
+        """
+        Retrieve the list of entitlements available to this learner.
+
+        Only those entitlements are returned that satisfy enterprise customer's data sharing setting.
+
+        Arguments:
+            request (HttpRequest): Reference to in progress request instance.
+            pk (Int): Primary key value of the selected enterprise learner.
+
+        Returns:
+            (HttpResponse): Response object containing a list of learner's entitlements.
+        """
+        enterprise_customer_user = self.get_object()
+
+        instance = {"entitlements": enterprise_customer_user.entitlements}
+        serializer = serializers.EnterpriseCustomerUserEntitlementSerializer(instance, context={'request': request})
+        return Response(serializer.data)
 
 
 class EnterpriseCustomerBrandingConfigurationViewSet(EnterpriseReadOnlyModelViewSet):
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -272,6 +272,58 @@ class EnterpriseCustomerUser(TimeStampedModel):
             return self.user.email
         return None
 
+    @property
+    def entitlements(self):
+        """
+        Return entitlement ids available to the learner along-with consent data.
+
+        Returns an empty list if enterprise customer requires data sharing consent and learner does not agree.
+
+        Returns:
+            (list): A list of entitlements that learner can avail. Each item in the list is a dict with two
+                key-value pairs,
+                {
+                    "requires_consent": True ,
+                    "entitlement_id": 1
+                }
+                "requires_consent": True if learner must consent to data
+                    sharing in order to get benefits of entitlement.
+                "entitlement_id: id of the entitlements available to the learner.
+        """
+        # Check if Enterprise Learner consents to data sharing and store the boolean result
+        learner_consent_state = self.data_sharing_consent.first()
+        learner_consent_enabled = learner_consent_state and learner_consent_state.enabled
+
+        entitlements = self.enterprise_customer.enterprise_customer_entitlements
+
+        # If Enterprise Customer requires account wide consent then we return either all or no entitlement
+        # depending on learner's consent state
+        if self.enterprise_customer.enforces_data_sharing_consent(EnterpriseCustomer.AT_LOGIN):
+            return [
+                {
+                    "entitlement_id": entitlement.entitlement_id,
+                    "requires_consent": not learner_consent_enabled,
+                } for entitlement in entitlements.all() if learner_consent_enabled
+            ]
+
+        # If Enterprise Customer requires account course specific consent then we return all entitlements
+        # including whether or not to acquire learner's consent.
+        if self.enterprise_customer.enforces_data_sharing_consent(EnterpriseCustomer.AT_ENROLLMENT):
+            return [
+                {
+                    "entitlement_id": entitlement.entitlement_id,
+                    "requires_consent": not learner_consent_enabled,
+                } for entitlement in entitlements.all()
+            ]
+
+        # for all other cases learner is eligible to all entitlements.
+        return [
+            {
+                "entitlement_id": entitlement.entitlement_id,
+                "requires_consent": False,
+            } for entitlement in entitlements.all()
+            ]
+
     def __str__(self):
         """
         Return human-readable string representation.
--- a/tests/api/test_views.py
+++ b/tests/api/test_views.py
@@ -12,6 +12,7 @@ from rest_framework.reverse import reverse
 
 from django.conf import settings
 
+from enterprise.models import EnterpriseCustomer, UserDataSharingConsentAudit
 from test_utils import APITest, factories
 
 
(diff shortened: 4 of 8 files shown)


enterprise/api/v1/views.py
```diff
@@ -82,6 +84,26 @@ class EnterpriseCustomerUserViewSet(EnterpriseReadOnlyModelViewSet):
...
+        Only those entitlements are returned that satisfy enterprise customer's data sharing setting.
+
+        Arguments:
+            request (HttpRequest): Reference to in progress request instance.
+            pk (Int): value if the primary key for the selected enterprise learner
+
+        Returns:
+            (HttpResponse): Response object containing a list of learner's entitlements.
+        """
+        enterprise_customer_user = self.get_object()
```
asadiqbal08: what if no record found against the `pk` by `self.get_object()` ?

enterprise/api/v1/views.py
```diff
@@ -82,6 +84,26 @@ class EnterpriseCustomerUserViewSet(EnterpriseReadOnlyModelViewSet):
...
+    @detail_route()
+    def entitlements(self, request, pk=None):  # pylint: disable=invalid-name,unused-argument
+        """
+        Retrieve the list of entitlements available to this learner.
+
+        Only those entitlements are returned that satisfy enterprise customer's data sharing setting.
+
+        Arguments:
+            request (HttpRequest): Reference to in progress request instance.
+            pk (Int): value if the primary key for the selected enterprise learner
```
asadiqbal08: can you re-write that statement ?

enterprise/models.py
```diff
@@ -270,6 +270,37 @@ def user_email(self):
             return self.user.email
         return None
 
+    @property
+    def entitlements(self):
+        """
+        Return entitlement ids available to the learner.
+
+        Returns an empty list if enterprise customer requires data sharing consent and learner has not agreed.
```
asadiqbal08: nit: `learner is not agreed` ?

enterprise/models.py
```diff
@@ -270,6 +270,37 @@ def user_email(self):
...
+            (list): A list of entitlements that learner can avail.
+        """
+        # Check if Enterprise Customer requires data sharing consent and store the boolean result
+        # pylint: disable=invalid-name
+        enterprise_customer_requires_consent = \
+            self.enterprise_customer.enforces_data_sharing_consent(EnterpriseCustomer.AT_LOGIN) or \
+            self.enterprise_customer.enforces_data_sharing_consent(EnterpriseCustomer.AT_ENROLLMENT)
+
+        # Check if Enterprise Learner consents to data sharing and store the boolean result
+        learner_consent_state = self.data_sharing_consent.first()
```
asadiqbal08: just for the information. what are the possible `learner_consent_state` here ?

enterprise/models.py
```diff
@@ -270,6 +270,37 @@ def user_email(self):
...
+        learner_consent_enabled = learner_consent_state and learner_consent_state.enabled
+
+        # return empty list of enterprise customer requires data sharing consent
```
asadiqbal08: nit: return empty list `if` enterprise customer ....

tests/api/test_views.py
```diff
@@ -199,8 +200,106 @@ def test_api_views(self, factory, url, sorting_key, model_items, expected_json):
...
+            False, EnterpriseCustomer.DATA_CONSENT_OPTIONAL, UserDataSharingConsentAudit.NOT_SET,
+            [1, 2, 3], {"entitlements": [1, 2, 3]},
+        ),
+    )
+    @ddt.unpack
+    def test_enterprise_learner_entitlements(
+            self, enable_data_sharing_consent, enforce_data_sharing_consent,
+            learner_consent_state, entitlements, expected_json
+    ):
+        user_id = 1
```
asadiqbal08: any `docstring` for the test behavior ?

tests/test_models.py
```diff
@@ -379,6 +381,50 @@ def test_user_email_property_user_missing(self):
...
+        (False, EnterpriseCustomer.AT_ENROLLMENT, UserDataSharingConsentAudit.NOT_SET, [1, 2, 3], [1, 2, 3]),
+        (False, EnterpriseCustomer.DATA_CONSENT_OPTIONAL, UserDataSharingConsentAudit.ENABLED, [1, 2, 3], [1, 2, 3]),
+        (False, EnterpriseCustomer.DATA_CONSENT_OPTIONAL, UserDataSharingConsentAudit.DISABLED, [1, 2, 3], [1, 2, 3]),
+        (False, EnterpriseCustomer.DATA_CONSENT_OPTIONAL, UserDataSharingConsentAudit.NOT_SET, [1, 2, 3], [1, 2, 3]),
+    )
+    @ddt.unpack
+    def test_entitlements(
+            self, enable_data_sharing_consent, enforce_data_sharing_consent,
+            learner_consent_state, entitlements, expected_entitlements,
+    ):
```
asadiqbal08: any `docstring` to describe the test behavior ?

asadiqbal08:
Overall LGTM 👍  small nits and questions. Be sure to resolve the conflicts as well.

tests/api/test_views.py
```diff
@@ -199,8 +200,106 @@ def test_api_views(self, factory, url, sorting_key, model_items, expected_json):
...
-        # import pdb; pdb.set_trace()
         response = self.client.get(settings.TEST_SERVER + url)
         response = self.load_json(response.content)
 
         assert sorted(expected_json, key=sorting_key) == sorted(response['results'], key=sorting_key)
+
+    @ddt.data(
+        (
+            True, EnterpriseCustomer.AT_LOGIN, UserDataSharingConsentAudit.ENABLED,
+            [1, 2, 3], {"entitlements": [1, 2, 3]},
```
asadiqbal08: any test if not entitlements are in the system ?

enterprise/api/v1/serializers.py
```diff
@@ -97,3 +97,25 @@ class Meta:
...
+    """
+    entitlements = serializers.ListField(
+        child=serializers.IntegerField()
+    )
+
+    def create(self, validated_data):
+        """
+        Do not perform any operations for state changing requests.
+        """
+
```
mattdrayer: Do we need to include a `pass` here or anything along those lines?

enterprise/api/v1/views.py
```diff
@@ -82,6 +84,26 @@ class EnterpriseCustomerUserViewSet(EnterpriseReadOnlyModelViewSet):
...
+
+        Returns:
+            (HttpResponse): Response object containing a list of learner's entitlements.
+        """
+        enterprise_customer_user = self.get_object()
+
+        instance = {"entitlements": enterprise_customer_user.entitlements}
```
mattdrayer: should we use `getattr()` here?

enterprise/models.py
mattdrayer: Can we count on at least one item always being returned here?  I've run into issues in the past with unprotected `.first()` references.

enterprise/api/v1/views.py
saleem-latif: details view creates a url something like `/enterprise/api/v1/enterprise-learner/{pk}/entitlements/`, so, if a record does not exist drf will return 404 before even executing this code.

enterprise/models.py
saleem-latif: `data_sharing_consent` comes via a reverse relationship on `UserDataSharingConsentAudit ` [here](https://github.com/edx/edx-enterprise/blob/master/enterprise/models.py#L519). Technically, this is a OneToMany relationship but our code and current business requirements make sure that it is OneToOne relationship.

to answer your original question, value of `learner_consent_state ` is an instance of `UserDataSharingConsentAudit` model.

enterprise/api/v1/serializers.py
saleem-latif: syntactically it is okay to skip `pass` if you have a doc-string in a no-op function.

enterprise/api/v1/views.py
saleem-latif: `self.get_object` returns an object or [raises a `404` error](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/generics.py#L97), so we are guaranteed to have an object at line 103. And documentation of DRF also follows this pattern. (Second example on this link http://www.django-rest-framework.org/api-guide/routers/#example)
```
    @detail_route()
    def group_names(self, request):
        """
        Returns a list of all the group names that the given
        user belongs to.
        """
        user = self.get_object()
        groups = user.groups.all()
        return Response([group.name for group in groups])
```
So, I think it would be redundant to use getattr. let me know if you had something else in mind

enterprise/models.py
saleem-latif: `first` on queryset either returns an object or `None` ref: https://github.com/django/django/blob/master/django/db/models/query.py#L553
 I am handling the `None` case in the next [line](https://github.com/edx/edx-enterprise/pull/43/files/2cfad789922757276533e180450d17fa9a6646d0#diff-4bcae564bc272f5c2e6753dc6021eb1cR291).

saleem-latif:
@mattdrayer , @asadiqbal08 I have incorporated your feedback, kindly take another look.

mattdrayer:
Great, changes LGTM -- 👍

asadiqbal08:
nice job. LGTM 👍

enterprise/api/v1/serializers.py
saleem-latif: @mattdrayer I have added pass statement to be on the safe side.

saleem-latif:
@e-kolpakov , @haikuginger , @bradenmacdonald can you guys take a look at the PR ?

bradenmacdonald:
@saleem-latif Yep, sure thing - likely not until Mon or Tues though.

mattdrayer:
@saleem-latif if you can rebase this branch and resolve the conflict(s) then hopefully the OpenCraft guys can take a look in this next cycle.

enterprise/models.py
```diff
@@ -272,6 +272,37 @@ def user_email(self):
             return self.user.email
         return None
 
+    @property
+    def entitlements(self):
```
saleem-latif: @haikuginger yes, I think we should also have course specific entitlements in the response. Here is the sample response for entitlements after the modifications, let me know your thoughts on this
```
GET /enterprise/api/v1/enterprise-learner/1/entitlements/

{
    "global": [],
    course_specific: {
        "course-id-1": [5],
        "course-id-2": [2]
    }
}
```
However, I am a bit confused about using `EnterpriseCourseEnrollment` for consent as  `EnterpriseCourseEnrollment` model contains entries  that correspond to the consent for course enrollment, meaning the learner has already enrolled in the course, so, should the learner still be eligible for the entitlements ? (if learner somehow wants to enroll in the same course twice). I think it would make more sense if we had course specific consent rather than enrollment specific consent.
I think we should not be returning course specific entitlements using `EnterpriseCourseEnrollment` model as it does not entail course specific consent but an enrollment specific consent. let me know your thoughts on this.
@mattdrayer -- FYI

haikuginger: Okay - so is this API designed to be used by the ecommerce stack to determine which entitlements a particular user is eligible to receive?

If so, then I think gating on course-specific consent is a mistake. Course-specific consent is obtained _at the time of enrollment_, not before. In fact, we're making modifications to the ecommerce stack to ensure that users are prompted for course-specific consent as part of the redemption process.

So, eligibility in this context isn't a question of "has the user fulfilled every requirement necessary to redeem this offer?", but a question of "has the user done anything to specifically exclude their ability to redeem?"

If that's the goal, I would suggest gating based on an account-level consent requirement, if extant, since account-level requirements are enforced at login, prior to the redemption of any offers.

In contrast, we should _not_ gate on course-specific consent, because such a consent cannot be provided until after the user attempts to take advantage of the offer.

Does that clarify things?

bradenmacdonald: @haikuginger Hmm, I thought the opposite - whether it's course level consent or account-level consent, if the learner must pay for the course, I thought we always want to prompt them for consent before they pay.

If GriffCo is using course-level consent only, and users go to use an entitlement that gives them 50% off, they must be prompted to agree to data sharing before they pay (hence before they enroll). Right? We don't want users to pay for a course and then disagree with data sharing, since then they've paid for something they cannot use.

> A user who is logged in, but has not provided consent, is completely ineligible to redeem offers from an EC who requires account-level consent.

Yes, that makes sense.

> In contrast, we should not gate on course-specific consent, because such a consent cannot be provided until after the user attempts to take advantage of the offer.

When you say "cannot" here, you mean "cannot currently due to how we implemented course-level consent", rather than "cannot in principle", right?

I seem to recall that this is a mistake due to us technically tying consent to the enrollment, and is not the desired behavior. I think we talked about changing this and using cookies or some other way to record per-course consent before enrollment takes place.

haikuginger: > When you say "cannot" here, you mean "cannot currently due to how we implemented course-level consent", rather than "cannot in principle", right?

Not what I mean at all. I mean that even _asking_ for consent for a specific course isn't meaningful unless the user has tried to interact with that course. So if the question this API is trying to answer is "which coupons is this user allowed to interact with/attempt to use?", then we ought not to gate on the fact that course-specific consent is required.

On the other hand, if the intended question is "which coupons can be immediately applied without further user intervention?", then yes, we ought to gate on consent.

All of this is the entire reason I asked about the specific use-case for this API. Obviously, the two main ones have very different requirements.

@saleem-latif, perhaps we can fit a greater number of use cases with this API by returning something like the below?

```json
{
    "entitlements": [
        {
            "requires_consent": true,
            "entitlement_id": 87
        },
        {
            "requires_consent": false,
            "entitlement_id": 32
        }
    ]
}
```

bradenmacdonald: @haikuginger Ah, gotcha. Makes sense - sorry I misunderstood.

enterprise/api/v1/serializers.py
```diff
@@ -97,3 +97,27 @@ class Meta:
     user = UserSerializer()
     enterprise_customer = EnterpriseCustomerSerializer()
     data_sharing_consent = UserDataSharingConsentAuditSerializer(many=True)
+
+
+class EnterpriseCustomerUserEntitlementSerializer(serializers.Serializer):
```
mattdrayer: We've been including the Read/Write capabilities of these classes in their names -- at least we've done this for some others -- so for example this might be `EnterpriseCustomerUserEntitlementReadOnlySerializer`

saleem-latif: @mattdrayer read/write capabilities are part of views like `EnterpriseReadOnlyModelViewSet` but for this serializer we do not have a view as this serializer is only being used for a `detail_route`.

saleem-latif:
@haikuginger I have updated the code as per your suggestions, kindly take a look now. API response is now like this
```
GET /enterprise/api/v1/enterprise-learner/1/entitlements/

{
    "entitlements": [
        {
            "entitlement_id": 5,
            "requires_consent": false
        }
    ]
}
```

haikuginger:
:+1:

- [x] I tested this: I tested all the relevant user-consent/consent-requirement permutations to verify that data was presented as expected.
- [x] I read through the code
- [x] ~~I checked for accessibility issues~~ N/A - API only

PR #44, opened 2017-02-08
[ENT-172] Catalog Service API graceful degradation

e-kolpakov:
**Description:** Makes Course Catalog API tolerant to empty lists returned from `get_edx_api_data`

**JIRA:** [ENT-172](https://openedx.atlassian.net/browse/ENT-172)

**Testing instructions:**

1. **Do not start Course Catalog**
2. Open "Manage Leaarners View" for any customer.
3. Observe that it didn't crashed with 500 error

**Merge checklist:**
- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:** Only last commit is specific to the issue - rest are problems slipped through code review and/or CI build from previous PRs.


edx-webhook:
Thanks for the pull request, @e-kolpakov! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=44&repo=edx%2Fedx-enterprise"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

CHANGELOG.rst
```diff
@@ -14,6 +14,14 @@ Change Log
 Unreleased
 ~~~~~~~~~~
 
+[0.21.3] - 2017-02-07
+~~~~~~~~~~~~~~~~~~~~~
+
+Added
+-----
+
+* Course Catalog API degrades gracefully in absence of Course Catalog  service.
```
mtyaka: Nit: double space between "Catalog" and "service".

enterprise/course_catalog_api.py
```diff
@@ -158,20 +160,23 @@ def get_common_course_modes(self, course_run_ids):
...
 
-    def _load_data(self, resource, **kwargs):
+    def _load_data(self, resource, default=DEFAULT_VALUE_SAFEGUARD, **kwargs):
         """
         Load data from API client.
 
         Arguments:
-            resource(str): type of resource to load
+            resource(string): type of resource to load
+            default(any): value to return if API query returned empty result. Sensible values: [], {}, None etc.
```
mtyaka: While this works fine, I wonder whether we shouldn't change `get_edx_api_data` from edx/edx-platform instead. It currently [always returns an empty list when it cannot contact the API](https://github.com/edx/edx-platform/blob/424b4356a583fccfc9f5dd64b387e39911aecc5c/openedx/core/lib/edx_api_utils.py#L78-L84), which makes sense when requesting collections, but makes no sense when requesting a particular resource (by providing the `resource_id` parameter).

mtyaka:
Thanks for fixing the dependency issues!

enterprise/course_catalog_api.py
e-kolpakov: @mtyaka Created https://tasks.opencraft.com/browse/OC-2344

e-kolpakov:
@mtyaka Rebased from https://github.com/edx/edx-enterprise/pull/44/commits/921ae5da0a640610df50cf51b8f801fa066d884f

e-kolpakov:
@mtyaka Please take another look

enterprise/admin/views.py
```diff
@@ -291,6 +291,13 @@ def _handle_bulk_upload(cls, enterprise_customer, manage_learners_form, request,
 
         return all_processable_emails
 
+    # TODO: this method is HUGE (> 100 lines) and complicated. Refactor it.
```
mattdrayer: When will we address this TODO?  Will it be part of this PR?  Will we log a ticket to revisit after a merge?

enterprise/admin/views.py
```diff
@@ -291,6 +291,13 @@ def _handle_bulk_upload(cls, enterprise_customer, manage_learners_form, request,
...
+    # Potential improvements:
+    # 1. Restructure flow so that notifications are sent after users and pending users are enrolled
+    #    FIXME: `send_email_notification_message` for pending users are executed OUTSIDE of ConditionalEmailConnection
```
mattdrayer: As with the TODO, when do we plan on addressing this FIXME?  If it will not be a part of this PR we should log a follow-up ticket so we don't forget about it.

enterprise/admin/views.py
```diff
@@ -311,6 +318,15 @@ def _enroll_users(cls, enterprise_customer, emails, course_id, mode, request, no
         enrollment_client = EnrollmentApiClient()
         course_details = CourseCatalogApiClient(request.user).get_course_run(course_id)
 
+        if not course_details:
+            logging.warning(
+                _(
+                    "Course details were not found for course key {}. "
```
mattdrayer: I think we should be more specific here -- can we say that course details for the course key were not returned by the Catalog Service

enterprise/course_catalog_api.py
```diff
@@ -158,20 +160,23 @@ def get_common_course_modes(self, course_run_ids):
...
 
         Returns:
             dict: Deserialized response from Course Catalog API
         """
-        return get_edx_api_data(
+        default_val = default if default != self.DEFAULT_VALUE_SAFEGUARD else {}
```
mattdrayer: Why have `DEFAULT_VALUE_SAFEGUARD` at all if we will return `{}` when it is detected?

mattdrayer: I guess I mean, why not make the value of `DEFAULT_VALUE_SAFEGUARD` `{}`?

enterprise/admin/views.py
e-kolpakov: @mattdrayer No, it won't be part of this PR. I've logged an internal Opencraft ticket for that, so please discuss with @bradenmacdonald if we need edX ticket or Opencraft ticket should suffice.

enterprise/course_catalog_api.py
e-kolpakov: @mattdrayer For two reasons:

1. Mutable default keyword argument values [behave differently from what intuition tells](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments), so generally it is a bad idea, so pylint and PyCharm inspector complain about that.
2. Can't use standard `default = None` and later `if not default: default = {}` as `None` might be a valid default value (see `get_program_by_title`).

enterprise/admin/views.py
e-kolpakov: @mattdrayer I've logged a follow-up ticket in Opencraft board - please discuss with Braden if it makes sense to have it exposed as edX ticket as well.

Also, just to clarify, if I need something refactored/fixed or tech debt repaid I usually just do that and only resort to TODOs and FIXMEs when I'm *not* planning to address it in scope of current work, because the improvement suggested does not fit in scope ("does not fit in scope" basically means it would increase the amount of work to be done at least 75-100%).

This task as a whole was supposed to be a small one - just wrap Course Catalog API return values so that they are always consistent with what the rest of  edx-enterprise expects. But there was that problem with dependencies management I had to fix. In short, I've already exhausted "drive-by fixes and improvements" budget for this task :)

enterprise/course_catalog_api.py
```diff
@@ -29,6 +29,8 @@ class CourseCatalogApiClient(object):
     Object builds an API client to make calls to the Catalog API.
     """
 
+    DEFAULT_VALUE_SAFEGUARD = "DEFAULT_VALUE_SAFEGUARD"
```
mtyaka: Nit: In edx-platform, I've seen `object()` used for sentinel values ([example1](https://github.com/edx/edx-platform/blob/62d129fdab1e047299441b78031a030308460464/common/lib/capa/capa/inputtypes.py#L130-L132), [example2](https://github.com/edx/edx-platform/blob/6d2468d0dcf782a7561ecc61946146d6fae2762f/lms/djangoapps/courseware/masquerade.py#L186)). It doesn't really make any practical difference, but is a bit cleaner form the theoretical perspective.

e-kolpakov:
@mattdrayer @asadiqbal08 @saleem-latif Please review

enterprise/admin/views.py
bradenmacdonald: Created ENT-205

enterprise/course_catalog_api.py
bradenmacdonald: +1

mattdrayer:
No further concerns from me -- 👍 after conflicts are resolved -- @saleem-latif please take a look as well since this PR addresses our prior attempt re: dependency management.

e-kolpakov:
@mattdrayer @saleem-latif Thank you for approval, but I can't resolve the rebase conflict - I don't (yet) know how to resolve that, as edx-enterprise want's to run on [djangorestframework < 3.3.0](https://github.com/edx/edx-enterprise/blob/master/tox.ini#L27) and django-config-models - on [djangorestframework >= 3.4.7](https://github.com/edx/django-config-models/blob/master/setup.py#L50).

@saleem-latif Why exactly do we need to limit edx-enterprise to DRF < 3.3.0? it looks like tests fail in Django 1.8 and Django 1.9 env if DRF >= 3.3.0, buut what's the actual reason?

CC @bradenmacdonald

bradenmacdonald:
Maybe @doctoryes knows the answer to that... Looks like the problem could be django-config-models since it [specifies](https://github.com/edx/django-config-models/blob/0.1.3/setup.py#L50) a version of DRF that's incompatible with [the version of DRF used in edx-platform](https://github.com/edx/edx-platform/blob/7597bcb/requirements/edx/base.txt#L20-L40). 

CC @haikuginger

doctoryes:
@bradenmacdonald @e-kolpakov When I moved django-config-models out of edx-platform, I used pip-tools to generate those requirements' versions - that's where the 3.4.7 version originated. We've since decided that external modules should pin requirements only when known incompatibilities exist - and instead use the versions required by the encompassing service. In this case, I don't think there's an incompatibility with previous versions of DRF, as the module was fairly recently moved out of the edx-platform codebase. Hopefully, there will be an OEP generated about requirement pinning sometime soon. The fix would be to either unpin the DRF requirement altogether -or- determine the last incompatible version and pin it to above that version.

saleem-latif:
@e-kolpakov I do not remember the exact point of failure but `rest_framework.reverse.reverse` which internally uses 'django.core.urlresolvers.reverse' method was causing the problem.

bdero:
@doctoryes It looks like @cpennington already opened a PR to relax the dependencies here: https://github.com/edx/django-config-models/pull/3. Can we merge/release that to unblock this PR?

bradenmacdonald:
@e-kolpakov Please open a PR to remove the version pinning from django-config-models and let's get this merged.

If we run into other delays (e.g. lack of Django 1.10 support on django-config-models), we can split out the drive-by fixes into a separate PR so that the main ENT-172 story here doesn't get blocked. That will make review easier, and we could ping all the enterprise developers on that PR to make sure everyone is on the same page about dependency management.

bradenmacdonald:
Update: Oh, great news @bdero  :)

doctoryes:
@bdero The django-config-models PR:
https://github.com/edx/django-config-models/pull/3
has now been merged. I tagged the release and the new version should be on PyPI shortly.

mattdrayer:
That's great!  Thanks @doctoryes!!!

bdero:
Thanks @doctoryes :)

mattdrayer:
@e-kolpakov how are we looking on this PR now given the recent update to django-config-models?

e-kolpakov:
@mattdrayer Haven't had a chance to check it out today - was quite busy with other stuff. ButI've tried to do without DCM yesterday and it didn't go lightly anyway. So we won't land it this week for sure (unless @bdero wants to be a hero :)), but I'd say it has good chances to land early next week (Mon-Tue-Wed).

bradenmacdonald:
@e-kolpakov Could we take up my suggestion of just splitting off the requirements untangling into a separate story, so that the actual goal of this story can be merged now?

And is it that hard to make django-config-models support Django 1.10?

e-kolpakov:
@bradenmacdonald Unfortunately, no, unless we're ok with failing build on master. As soon as I rebase there will be a dependency on DCM and it prevents CI build from running.

bradenmacdonald:
@e-kolpakov  Well the good news is that @doctoryes said we can go ahead to try enabling Django 1.10 in django-config-models and there's no specific reason why it's not enabled on there - so it might not be too hard to fix.

e-kolpakov:
@bradenmacdonald The bad news is that DRF needs 3.5+ to run on Django 1.10 and most of the things recently added (DRF extensions, DCM, etc.) *depend* on Django or DRF, so if those are listed in "*.in" file Django and DRF appear in "*.txt" file, with different versions and stuff and here we go again :) That's why this particular PR takes so long - even before DCM was introduced (with totally non-matching versions) there was a problem of making it work with different Django and DRF.

All in all, I strongly recommend dropping Django 1.10 - it's the easiest way to proceed now. If need be, we can add Django 1.10 support back in a follow up task.

haikuginger:
We could drop Python 2 instead. 😃

bradenmacdonald:
@e-kolpakov 

> All in all, I strongly recommend dropping Django 1.10 - it's the easiest way to proceed now. If need be, we can add Django 1.10 support back in a follow up task.

Ok, we gotta do what we gotta do. That does seem like the most expedient way forward here. Or you could rewrite python's entire package loading ecosystem so that multiple versions of each package can co-exist in a nice dependency tree :-p

e-kolpakov:
@mtyaka @mattdrayer @saleem-latif Conflicts seems to be resolved, build passes - please review one last time.

mattdrayer:
I've got nothing else to add at this point -- looks like there's one CI failure to resolve (maybe just rerun?) -- otherwise I'm good -- 👍

e-kolpakov:
@mattdrayer Yep - it never get to run that build - restarting and will merge as soon as it go green.

saleem-latif:
@e-kolpakov version update in ` enterprise/__init__.py` of `0.22.1` is not present in the change set, seems like it have been removed during rebase maybe.  As a result I was getting this error trying to update edx-enterprise version on sandbox for testing.
```
Collecting edx-enterprise==0.22.1 (from -r requirements/edx/base.txt (line 54))
  Could not find a version that satisfies the requirement edx-enterprise==0.22.1 (from -r requirements/edx/base.txt (line 54)) (from versions: 0.1.0, 0.1.2, 0.3.1, 0.4.0, 0.4.1, 0.6.0, 0.7.0, 0.8.0, 0.9.0, 0.10.0, 0.11.0, 0.12.0, 0.13.0, 0.15.0, 0.16.0, 0.17.0, 0.18.0, 0.19.0, 0.19.1, 0.20.0, 0.21.0, 0.21.1, 0.21.2, 0.22.0, 0.23.0)
No matching distribution found for edx-enterprise==0.22.1 (from -r requirements/edx/base.txt (line 54))
```

May be we should just update release tag [0.22.1](https://github.com/edx/edx-enterprise/releases/tag/0.22.1) to `0.23.0` ?

@mattdrayer -- FYI

PR #46, opened 2017-02-15
[WIP] ENT-195 Building django admin control for integrated channels m…

brittneyexline:
…odels

**Description:** 
*adjusted setup.py to include the new modules under integrated_channels
*added basic admin functionality for manipulating the models for configuring integrated_channel and sap_success_factors

**JIRA:** https://openedx.atlassian.net/browse/ENT-195

**Dependencies:** There will be changes in edx-platform, I will link the PR when I create it.

instructions.

**Testing instructions:**

will update

**Reviewers:**
- [ ] tag reviewer 
- [ ] tag reviewer 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag
      finished.

migrations, etc.


integrated_channels/integrated_channel/admin/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def enterprise_customer_name(self, ent_customer_plugin_config):
+        """
+        Return the name for the attached EnterpriseCustomer.
+
+        Args:
+            ent_customer_plugin_config: The instance of EnterpriseCustomerPluginConfiguration
+                being rendered with this admin form.
+        """
+        return ent_customer_plugin_config.enterprise_customer.name \
```
mattdrayer: It might be better for the EnterpriseCustomer to own the rule for determining what to use for the name, instead of doing it here -- thoughts?

integrated_channels/integrated_channel/models.py
```diff
@@ -65,13 +67,27 @@ class Meta:
...
+
     def __str__(self):
         """
         Return human-readable string representation.
         """
-        return "<EnterpriseCustomerPluginConfiguration for enterprise {enterprise} using channel {channel}>".format(
-            enterprise=self.enterprise_customer_uuid,
-            channel=self.enterprise_integrated_channel.name
+        return "<EnterpriseCustomerPluginConfiguration for enterprise {} using channel {} for data type {}>".format(
+            (self.enterprise_customer.name if self.enterprise_customer else self.enterprise_customer_uuid),
```
mattdrayer: Yeah, seeing this here makes me think that we should introduce a "name" property for EnterpriseCustomer -- it's a weird case that we're describing here, however -- basically we have a UUID but not an actual record to go along with it -- if this is the case, where did the UUID even come from?

mattdrayer:
Some questions around determination of the EnterpriseCustomer name (take it or leave it, tbh), otherwise the changes LGTM -- 👍 -- @bradenmacdonald if you (or a team member) can take a look as well, that would be great.

integrated_channels/integrated_channel/models.py
brittneyexline: Yea, I debated whether or not to introduce a ForeignKey to EnterpriseCustomer or not. Having uuid without the key makes it so that it doesn't have to be tied to EnterpriseCustomer per se, but when I looked at it in the admin interface it seemed like having a name would be more user friendly if we don't introduce the foreign key.

bradenmacdonald:
@mattdrayer Yep, one of us will review - is Monday/Tuesday ok? My plate is full today.

integrated_channels/integrated_channel/admin/__init__.py
haikuginger: We seem to be using logic like this in multiple places. Let's move it into a method on the PluginConfiguration. I'd also prefer an `if-return-else-return` pattern as opposed to trying to fit everything onto one line equivalent.

tests/test_models.py
```diff
@@ -678,15 +678,16 @@ def test_string_conversion(self, method):
...
         faker = FakerFactory.create()
         customer_uuid = faker.uuid4()
         plugin_config = EnterpriseCustomerPluginConfiguration(
             enterprise_customer_uuid=customer_uuid, enterprise_integrated_channel=channel, active=False
         )
-        expected_to_str = "<EnterpriseCustomerPluginConfiguration for enterprise {uuid} using channel CorporateLMS>"\
-            .format(uuid=customer_uuid)
+        expected_to_str = "<EnterpriseCustomerPluginConfiguration for enterprise {} using channel CorporateLMS" \
+                          " for data type catalog>"\
+            .format(customer_uuid)
```
haikuginger: I'm not thrilled about the line joining format here. I'd prefer to use Python's native multiline string joining like so:

```python
expected_to_str = (
    "<EnterpriseCustomerPluginConfiguration for enterprise {} using "
    "channel CorporateLMS for data type catalog>"
).format(customer_uuid)

File: integrated_channels/sap_success_factors/models.py [hunk 4]
```diff
@@ -45,7 +45,8 @@ class SAPSuccessFactorsEnterpriseCustomerConfiguration(ConfigurationModel):
     The Enterprise specific configuration we need for integrating with SuccessFactors.
```
haikuginger: I want to make sure I'm properly understanding why we're using a ConfigurationModel here - my understanding of that model is that it's designed to produce a single canonical configuration for the whole platform, with rollback features built in so that prior versions can be restored. Wouldn't we want to have any number of canonical models that are tied to different EnterpriseCustomers? We can use HistoricalRecords() to make sure the history of each model is saved.

brittneyexline:
Thanks @haikuginger for the feedback, I will put up another round of changes to address your comments.

I have one quality error that I can't figure out how to address and I'm wondering if you have any ideas?

haikuginger:
@brittneyexline, I think that's one of the ones you just have to cover with a pragma.

haikuginger:
Okay, I've verified the behavior, and I'm happy. One thing I noticed is that CodeCov.io doesn't seem to be tracking changes against the integrated_channels app; can you set that up and make sure we're at a good level of coverage?

integrated_channels/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""
+Integrated channel package for Enterprises.
```
mattdrayer: Can you provide a little more context here around what "integrated channels" are?  Also I'd say "[...] for the Open edX Enterprise Service."

integrated_channels/integrated_channel/admin/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+        "enterprise_customer_name",
+    )
+
+    list_filter = ("enterprise_integrated_channel", "active",)
+    search_fields = ("enterprise_customer_name", "integrated_channel_name",)
+
+    class Meta(object):
+        model = EnterpriseCustomerPluginConfiguration
+
+    def integrated_channel_name(self, ent_customer_plugin_config):
```
mattdrayer: Is this the right place to consider using a `property`?

mattdrayer: It sounds like a property name...if it's not going to be a property we usually prefix the function name with "get", as in `get_integrated_channel_name`

integrated_channels/integrated_channel/admin/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        Return the name for the attached integrated_channel.
```
mattdrayer: I'd prefer that we follow convention and add a `Returns:` section to the docstring instead of stating it this way.  This goes for the rest of the docstrings as well.

integrated_channels/integrated_channel/admin/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+
+        Args:
+            ent_customer_plugin_config: The instance of EnterpriseCustomerPluginConfiguration
```
mattdrayer: here you can do something like `enterprise_customer_plugin_config (EnterpriseCustomerPluginConfiguration): [description]`

mattdrayer: Same for the rest of the docstrings that specify Args

integrated_channels/integrated_channel/admin/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+                being rendered with this admin form.
+        """
+        return ent_customer_plugin_config.enterprise_integrated_channel.data_type
+
+    def enterprise_customer_name(self, ent_customer_plugin_config):
+        """
+        Return the name for the attached EnterpriseCustomer.
```
mattdrayer: We might want to indicate here that the caller will receive either the name or the UUID value for the customer, since that is what the code does

integrated_channels/integrated_channel/admin/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+
+        Args:
+            ent_customer_plugin_config: The instance of EnterpriseCustomerPluginConfiguration
+                being rendered with this admin form.
+        """
+        if ent_customer_plugin_config.enterprise_customer:
+            return ent_customer_plugin_config.enterprise_customer.name
+        else:
+            return ent_customer_plugin_config.enterprise_customer_uuid
```
mattdrayer: Is this an actual case?  Could we have a UUID but no matching EC record?

mattdrayer: Also I think you can just do:

```
if ent_customer_plugin_config.enterprise_customer:
    return ent_customer_plugin_config.enterprise_customer.name
return ent_customer_plugin_config.enterprise_customer_uuid
```

integrated_channels/integrated_channel/apps.py
```diff
@@ -1,7 +1,7 @@
 """
 Enterprise Integrated Channel Django application initialization.
 """
-from __future__ import unicode_literals
+from __future__ import absolute_import, unicode_literals
```
mattdrayer: Would be helpful to include a comment describing why absolute_import is necessary -- maybe also include link to the Django documentation if pertinent

integrated_channels/integrated_channel/migrations/0001_initial.py
```diff
@@ -2,9 +2,10 @@
 # Generated by Django 1.9.12 on 2017-02-10 16:00
 from __future__ import unicode_literals
 
-from django.db import migrations, models
 import django.db.models.deletion
 import django.utils.timezone
+from django.db import migrations, models
```
mattdrayer: I assume this change was added automatically by the Django utility, right?

integrated_channels/integrated_channel/models.py
```diff
@@ -1,13 +1,15 @@
...
 """
-from __future__ import unicode_literals
+from __future__ import absolute_import, unicode_literals
 
 from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.translation import ugettext_lazy as _
 from model_utils.models import TimeStampedModel
 
+from enterprise.models import EnterpriseCustomer
```
mattdrayer: I get anxious when I see cross-app model binding like this -- maybe it's not a big deal in this case but we might consider calling an accessor function instead of going directly to the model.

integrated_channels/integrated_channel/models.py
```diff
@@ -65,13 +67,27 @@ class Meta:
...
+        """
+        Return Enterprise Customer associated with this instance.
+
+        Return :class:`enterprise.models.EnterpriseCustomer` instance associated with this
+        :class:`EnterpriseCustomerPluginConfiguration` instance via uuid.
+        """
+        try:
+            return EnterpriseCustomer.objects.get(uuid=self.enterprise_customer_uuid)  # pylint: disable=no-member
+        except EnterpriseCustomer.DoesNotExist:
+            return None
```
mattdrayer: This seems like an odd scenario to me -- we have a UUID but the record does not exist in the system -- I would consider logging this situation instead of simply sweeping it under the rug

integrated_channels/integrated_channel/models.py
mattdrayer: I'm okay not having an explicit FK here to limit the extent of the coupling -- it just seems odd to me that we'd have a UUID at all if the record does not exist.  I guess it's a future-proofing thing since the FK is absent...

integrated_channels/sap_success_factors/admin/__init__.py
```diff
@@ -0,0 +1,42 @@
+"""
+Django admin integration for sap_success_factors app.
```
mattdrayer: If we can provide some more context here describing the actual purpose for this app, that would be great -- please include a link to pertinent documentation @ SAP, such as the OCN integration guide.

integrated_channels/sap_success_factors/migrations/0002_auto_20170222_1102.py
```diff
@@ -0,0 +1,64 @@
+# -*- coding: utf-8 -*-
```
mattdrayer: One thing we can do before the initial merge of a new app is merge the migration files into a single 0001_initial.py.  This basically involves deleting all of the existing migration files and running makemigrations one last time to pick up all of the changes.

mattdrayer:
@brittneyexline first pass complete!

integrated_channels/integrated_channel/admin/__init__.py
brittneyexline: @mattdrayer what does it mean to make it a property?

haikuginger: Decorate it with `@property`, and Python will call the function when you just use it - example:

```python
from uuid import uuid4

class Thing:
    @property
    def real_static_value(self):
        return uuid4().hex

>>> x = Thing()
>>> x.real_static_value
    '34eb598ca3434f5fb949397528eb0580'
```

But it only works for functions where the only argument is `self`.

integrated_channels/integrated_channel/admin/__init__.py
brittneyexline: i think it's possible, but not likely. seems simple enough to safeguard against.

integrated_channels/integrated_channel/apps.py
brittneyexline: I added this because of one of the linters - after I enabled more of them that I had forgotten before this came up for a bunch of files. I can look into why it is that way.

integrated_channels/integrated_channel/migrations/0001_initial.py
brittneyexline: the isort check was complaining about it, so i ran isort on it :-)

integrated_channels/integrated_channel/models.py
brittneyexline: @mattdrayer is there an example you can point me to for an alternative to doing this?

integrated_channels/integrated_channel/admin/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+    search_fields = ("name", "data_type",)
+
+    class Meta(object):
+        model = EnterpriseIntegratedChannel
+
+
+@admin.register(EnterpriseCustomerPluginConfiguration)
+class EnterpriseCustomerPluginConfigurationAdmin(admin.ModelAdmin):
+    """
+    Django admin model for EnterpriseCustomerPluginConfiguration.
```
bradenmacdonald: @brittneyexline After seeing your demo today, it made me question if we even need this model?

We already have `EnterpriseIntegratedChannel` to determine what channels are enabled or not.

Then we also have models like `SAPSuccessFactorsEnterpriseCustomerConfiguration` which can link each particular enabled channel to the `EnterpriseCustomer`.

So if you add an `active` flag to `SAPSuccessFactorsEnterpriseCustomerConfiguration`, then `EnterpriseCustomerPluginConfiguration` seems completely redundant. I think it might even be unhelpful to have - after all, what would it mean if GriffCo is enabled for SAP according to `EnterpriseCustomerPluginConfiguration` but there is no corresponding `SAPSuccessFactorsEnterpriseCustomerConfiguration` record for GriffCo? If you remove the intermediate model then you can't have that kind of inconsistent state in the database. (Which is especially important when some of these links between models are not enforced by foreign keys.)


This is very analogous to [the models in `third_party_auth`](https://github.com/edx/edx-platform/blob/f806f3a/common/djangoapps/third_party_auth/models.py#L79) - there, I put the common functionality (similar to `EnterpriseCustomerPluginConfiguration`) into an [abstract base class](https://docs.djangoproject.com/en/1.10/topics/db/models/#abstract-base-classes) (which includes, for example, the `enabled` flag, analogous to the `active` flag), and I made each type of provider inherit from that abstract base class.

brittneyexline:
Thanks @bradenmacdonald, the comments were helpful. I think you hit on something I've been struggling with, which is how independent should these apps in edx-enterprise be from each other? I think they all have distinct purposes, but they are also highly interrelated, and it's difficult for me to imagine them being used independently of each other. The enterprise package is really the only one that can exist by itself at this point. @mattdrayer and @haikuginger, what do you think?

mattdrayer:
If you're suggesting that we merge the integrated channels stuff into the enterprise app vs. having independent apps I am okay with it.  One exercise I find helpful in situations like this is to consider whether an app could (and if yes, should) be hosted as a separate service at some point in the future -- if either could or should do not seem to make sense then the app might be better suited as a set of functionality located within another app.

bradenmacdonald:
My suggestion is to keep them as separate (django) apps but within the same overall package and repository for now (which is what we currently have), and to use foreign keys between them wherever possible for an overall better user experience and data consistency.

haikuginger:
I agree with Braden - separate apps, but designed to work together.

brittneyexline:
I think that makes sense. I will update this accordingly because I'd rather handle the migrations now as opposed to later.

integrated_channels/integrated_channel/models.py
mattdrayer: As long as we're not making it harder for ourselves in the future I'm okay with it -- I think the scope here is fairly contained, especially with the recent decision to combine everything into the same package.

integrated_channels/sap_success_factors/models.py
mattdrayer: Looks like we've changed to use HistoricalRecords, which is aok with me.

mattdrayer:
Looking pretty good at this point, @brittneyexline -- unless @bradenmacdonald or @haikuginger have more they would like to see changed, I'd say it's time to wrap things up.  👍 for me.

integrated_channels/sap_success_factors/admin/__init__.py
```diff
@@ -33,10 +33,28 @@ class SAPSuccessFactorsEnterpriseCustomerConfigurationAdmin(admin.ModelAdmin):
...
+
+    list_filter = ("active",)
+    search_fields = ("enterprise_customer_name",)
+
     class Meta(object):
         model = SAPSuccessFactorsEnterpriseCustomerConfiguration
+
+    def enterprise_customer_name(self, obj):
+        """
+        Returns: the name for the attached EnterpriseCustomer, or the uuid if the EnterpriseCustomer does not exist.
```
bradenmacdonald: I don't see any logic here to return "the uuid if the EnterpriseCustomer does not exist."

PR #47, opened 2017-02-16
ENT-182: Defer EnterpriseCustomerUser creation until after consent is provided

haikuginger:
**Description:** This pull request refactors the `post_account_consent` method to ensure that new users, entering the system via SSO, are not linked to an EnterpriseCustomer unless they provide the relevant level of data sharing consent.

**JIRA:** [ENT-182](https://openedx.atlassian.net/browse/ENT-182)

**Dependencies:** None

**Merge deadline:** 27 February 2017

**Installation instructions:** Trivial installation; standard features. SSO provider setup required.

**Sandbox link**: https://ent182.sandbox.edx.org/; use staff/edx user for admin access

**Testing instructions:**

1. Set up a SAML integration with TestShib
2. Set up an EnterpriseCustomer linked to that SAML integration
3. Register a user account using standard methods, without using the SSO integration.
4. In a new incognito tab, browse to the instance main page and click Login.
5. Click on the TestShib SSO button; proceed through the login flow.
6. When redirected back to the instance, you'll be presented with an option to log in with your username and password. Log in with the username and password of the account you created earlier.
7. Because the SSO provider is linked to your EnterpriseCustomer, you'll be prompted for data sharing consent. Do not check the consent box, and click Submit.
8. You will be redirected back to the login page, with your browsing session flushed.
9. In a separate window, viewing the Manage Learners admin page of your EnterpriseCustomer, observe that the user you just logged in with is _not_ linked to the EnterpriseCustomer.
10. Observe that if you attempt to log in with the SSO provider, you are prompted for consent. Provide it.
11. Observe that only after providing consent in the previous step, the user you created is added to the list of linked learners in the Manage Learners admin page.

**Reviewers:**
- [x] @mtyaka  
- [x] EdX TBD

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag
      finished.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=47&repo=edx%2Fedx-enterprise"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

mtyaka:
:+1: 

The ENT ticket says the expected results is:

> For data sharing consent required at login option, user should not be able to login to application as well as their account should not link with enterprise until user check data sharing consent checkbox.

The "user should not be able to login to application" part is probably not correct since if the user already has an existing account (not linked to and `EnterpriseCustomer`), we can only prevent them from getting linked to the `EnterpriseCustomer` without giving data sharing consent, but we cannot prevent them from logging in using their existing credentials.

- [x] I tested this: I followed the steps provided in "Testing instructions" on the sandbox, using the myself/myself TestShib user.
- [x] I read through the code
- [ ] ~~Includes documentation~~ bugfix; not required

enterprise/views.py
```diff
@@ -248,10 +248,21 @@ def post_account_consent(self, request, consent_provided):
         Interpret the account-wide form above, and save it to a UserDataSharingConsentAudit object for later retrieval.
         """
         self.lift_quarantine(request)
+
         customer = get_enterprise_customer_for_request(request)
-        if customer is None:
+        social_auth = get_real_social_auth_object(request)
+        user = getattr(social_auth, 'user', None)
+        if customer is None or user is None:
```
mattdrayer: I'd prefer to see this broken up into separate guards because although the outcomes are similar the error cases are a bit different:

```
# Load the EnterpriseCustomer object tied to this request.  Trigger a 404 response if not available.
enteprise_customer = get_enterprise_customer_for_request(request)
if enterprise_customer is None:
    raise Http404

# Load the User object tied to the authentication object for this request.  Trigger a 404 response if not available.
social_auth = get_real_social_auth_object(request)
user = getattr(social_auth, 'user', None)
if user is None:
    raise Http404
```

enterprise/views.py
```diff
@@ -248,10 +248,21 @@ def post_account_consent(self, request, consent_provided):
...
+            # Check that we can get both a user and an EnterpriseCustomer from the running
+            # TPA pipeline; if not, this view isn't relevant, and shouldn't be used.
             raise Http404
-        user = get_real_social_auth_object(request).user
+
+        if not consent_provided and active_provider_enforces_data_sharing(request, EnterpriseCustomer.AT_LOGIN):
```
mattdrayer: Does it make sense to perform this check prior to loading the customer+user information above?

mattdrayer: ie, it would be nice to locate the customer+user loaders closer to where the information is being used on L266

haikuginger: I don't think so in this case. The 404s are "you shouldn't be seeing this page at all" blocks; we don't want to do a session flush if, for example, an EnterpriseCustomer simply doesn't exist.

enterprise/views.py
haikuginger: Done!

mattdrayer:
Looks great @haikuginger -- thanks for addressing my items -- time to shipppppit  👍

haikuginger:
Added version bump stuff; squashed from 0f4df4f. Will merge at CoB today if there are no objections.

PR #59, opened 2017-03-02
ENT-220/ENT-223 Collect learner completion data for SAP SuccessFactors

pomegranited:
**Description:** Adds a management command and updates supporting classes to collect learner completion data records, and prepare them for submission to the SAP SuccessFactors integration channel.

**JIRA:** [ENT-220](https://openedx.atlassian.net/browse/ENT-220), [ENT-223](https://openedx.atlassian.net/browse/ENT-223)

**Dependencies:** 

* https://github.com/edx/edx-platform/pull/14563
* https://github.com/edx/edx-enterprise/pull/61 (merged here)

**Sandbox:**

* LMS: https://ent-pr59.sandbox.opencraft.hosting/
* Studio: https://studio-ent-pr59.sandbox.opencraft.hosting/

Also created an edX Sandbox: https://ent-220-ent-223-pomegranited.sandbox.edx.org
I've added github keys for @haikuginger , @brittneyexline , @mattdrayer , so you can shell in as `pomegranited` and test the management command.  

**Merge deadline:** 13 March 2017

1. Ensure edx-platform is running with changes from https://github.com/edx/edx-platform/pull/14563.
1. Install this branch's latest commit on your edxapp devstack:

    ```
    pip uninstall -y edx-enterprise && pip install "git+https://github.com/open-craft/edx-enterprise.git@jill/sap_learner_completion#egg=edx-enterprise"
    ```
1. You may have to run data migrations:

   ```
    ./manage.py lms migrate --settings=devstack
   ```
1. Restart the LMS without clobbering the installed `edx-enterprise`:

    ```
    NO_PREREQ_INSTALL=1 paver run_all_servers --fast
    ```
1. Set up a [TestShib SAML integration](http://testshib.org) with [these instructions](http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/configuration/tpa/tpa_SAML_SP.html).

**Testing instructions:**

1. In Studio, create a new course.
1. From the Admin, create a course mode for the course under `Course_Modes › Course modes`.  Choose a mode other than Audit.
1. In Studio,  navigate to the Settings > Certificates page, and create a certificate.  Don't worry about adding signatories.  Activate the certificate.
1. Run Testing instruction steps 1-4 from https://github.com/edx/edx-platform/pull/14400.  Use a new learner, so that it can be registered via TestShib.  Enrol the new learner in your course.
1. Link the `EnterpriseCustomer` to the TestShib SAML integration as the EC's `Identity Provider`.
1. From the Admin, add an [Enterprise SAP SuccessFactors Integration › Sap success factors enterprise customer configurations](http://localhost:8000/admin/sap_success_factors/sapsuccessfactorsenterprisecustomerconfiguration/) for the `EnterpriseCustomer`.  The only crucial field is `enabled=True`.  The other fields do't matter yet, because no actually communication is being done.
1. Run the new management command.  No output expected, unless any of your enrolled learners have granted consent.

    ```
    # For all enterprise customers
    ./manage.py lms transmit_learner_data  --settings=devstack

    # Or just one.
    ./manage.py lms transmit_learner_data --enterprise_customer=<enterprise_customer.uuid>  --settings=devstack
    ```
1. Register your enrolled learner via TestShib, and grant data sharing consent.
1. Re-run the `transmit_learner_data` management command to see the learner data record logged. Note that it reports `courseCompleted: "false"`.
1. Generate a certificate for your course.  

    ```
    ./manage.py lms ungenerated_certs --course="<course_id>" --settings=devstack
    ```
    *Note*: I got a lot of `VersionConflict` errors when I ran this, which didn't seem to matter.  And the log line that indicates the certificate has been generated may say something like "Student 5 has certificate status 'unavailable'".  But that seems to be ok too?
1. Re-run the `transmit_learner_data` management command, and check that the output now says `courseCompleted: "true"`, and the `grade` field has a value.

**Author notes and concerns:**

* `grade` field value may change, depending on discussion.  Currently using the `certificate.grade`, but may provide `Pass/Fail` instead.
* What to do about audit modes, in self-paced courses with no certificates to indicate completion and/or final grade?  Currently, we can only send learner completion data when a certificate has been issued.

**Reviewers:**

- [x] @haikuginger 
- [x] @brittneyexline 

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped - to `0.27.0`
- [x] Changelog record added
- [ ] Translations are updated - [see comment](https://github.com/edx/edx-enterprise/pull/59#issuecomment-285582381)

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.27.2] - 2017-03-10
+---------------------
+
+* Added integrated_channels management command to transmit learner completion data to SAP SuccessFactors.
+
 [0.27.1] - 2017-03-13
 ---------------------
 
@@ -24,7 +29,6 @@ Unreleased
 ---------------------
 
 * Added API endpoint for fetching catalogs and catalog courses.
-
 
 [0.26.3] - 2017-03-02
 ---------------------
--- a/integrated_channels/integrated_channel/management/commands/__init__.py
+++ b/integrated_channels/integrated_channel/management/commands/__init__.py
@@ -1,0 +1,101 @@
+"""
+Enterprise Integrated Channel management commands.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from django.core.management.base import CommandError
+from django.utils.translation import ugettext as _
+
+from enterprise.models import EnterpriseCustomer
+from integrated_channels.sap_success_factors.models import SAPSuccessFactorsEnterpriseCustomerConfiguration
+
+
+# Import djcelery, or stub it if not available.
+try:
+    from djcelery.celery import task as celery_task
+except ImportError:
+    def celery_task(func):
+        """Use a no-op decorator if djcelery is not available."""
+        def no_delay(*args, **kwargs):
+            """
+            Provides the celery.task.delay function, which can be used to spawn an asynchronous celery task.
+            Here, it just calls the function.
+            """
+            func(*args, **kwargs)
+        func.delay = no_delay
+        return func
+
+# Mapping between the channel code and the channel configuration class
+INTEGRATED_CHANNEL_CHOICES = {
+    channel_class.channel_code(): channel_class
+    for channel_class in (SAPSuccessFactorsEnterpriseCustomerConfiguration, )
+}
+
+
+class IntegratedChannelCommandMixin(object):
+    """
+    Contains common functionality for the IntegratedChannel management commands.
+    """
+    def add_arguments(self, parser):
+        """
+        Adds the optional arguments: ``--enterprise_customer``, ``--channel``
+        """
+        parser.add_argument(
+            '--enterprise_customer',
+            dest='enterprise_customer',
+            default=None,
+            metavar=_('ENTERPRISE_CUSTOMER_UUID'),
+            help=_('Transmit data for only this EnterpriseCustomer. '
+                   'Omit this option to transmit to all EnterpriseCustomers with active integrated channels.'),
+        )
+        parser.add_argument(
+            '--channel',
+            dest='channel',
+            default='',
+            metavar=_('INTEGRATED_CHANNEL'),
+            help=_('Transmit data to this IntegrateChannel. '
+                   'Omit this option to transmit to all configured, active integrated channels.'),
+            choices=INTEGRATED_CHANNEL_CHOICES.keys(),
+        )
+
+    def get_integrated_channels(self, options):
+        """
+        Generates a list of active integrated channels, filtered from the given options.
+
+        Raises errors when invalid options are encountered.
+
+        See ``add_arguments`` for the accepted options.
+        """
+        # If a valid enterprise customer uuid was provided, transmit for that customer.
+        # Otherwise, transmit for all customers.
+        enterprise_customer = None
+        enterprise_customer_uuid = options['enterprise_customer']
+        if enterprise_customer_uuid:
+            try:
+                enterprise_customer = EnterpriseCustomer.active_customers.get(uuid=enterprise_customer_uuid)
+            except EnterpriseCustomer.DoesNotExist:
+                raise CommandError(
+                    _('Enterprise customer {uuid} not found, or not active').format(uuid=enterprise_customer_uuid))
+
+        # Assemble a list of integrated channel classes to transmit to.
+        # If a valid channel type was provided, use it.
+        # Otherwise, use all the available channel types.
+        channel_code = options['channel'].upper()
+        if channel_code:
+            if channel_code not in INTEGRATED_CHANNEL_CHOICES:
+                raise CommandError(_('Invalid integrated channel: {channel}').format(channel=channel_code))
+
+            channel_classes = [INTEGRATED_CHANNEL_CHOICES[channel_code]]
+        else:
+            channel_classes = INTEGRATED_CHANNEL_CHOICES.values()
+
+        # Loop through each channel class (optionally for a specific enterprise customer)
+        for channel_class in channel_classes:
+            # Use Active channels only
+            integrated_channels = channel_class.objects.filter(active=True)
+            if enterprise_customer:
+                integrated_channels = integrated_channels.filter(enterprise_customer=enterprise_customer)
+
+            # Gen the learner data to each integrated channel
+            for integrated_channel in integrated_channels:
+                yield integrated_channel
--- a/integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
+++ b/integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
@@ -1,0 +1,38 @@
+"""
+Transmits consenting enterprise learner data to the integrated channels.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from django.core.management.base import BaseCommand
+from django.utils.translation import ugettext as _
+from . import IntegratedChannelCommandMixin, celery_task, INTEGRATED_CHANNEL_CHOICES
+
+
+class Command(IntegratedChannelCommandMixin, BaseCommand):
+    """
+    Management command which transmits learner course completion data to the IntegratedChannel(s) configured for the
+    given EnterpriseCustomer.
+
+    Collect the enterprise learner data for enrollments with data sharing consent, and transmit each to the
+    EnterpriseCustomer's configured IntegratedChannel(s).
+    """
+    help = _('''
+    Transmit Enterprise learner course completion data for the given EnterpriseCustomer.
+    ''')
+
+    def handle(self, *args, **options):
+        """
+        Transmit the learner data for the EnterpriseCustomer(s) to the active integration channels.
+        """
+        # Transmit the learner data to each integrated channel
+        for integrated_channel in self.get_integrated_channels(options):
+            self.transmit_learner_data.delay(integrated_channel.channel_code(), integrated_channel.pk)
+
+    @staticmethod
+    @celery_task
+    def transmit_learner_data(channel_code, channel_pk):
+        """
+        Allows each enterprise customer's integrated channel to collect and transmit data within its own celery task.
+        """
+        integrated_channel = INTEGRATED_CHANNEL_CHOICES[channel_code].objects.get(pk=channel_pk)
+        integrated_channel.transmit_learner_data()
--- a/integrated_channels/integrated_channel/models.py
+++ b/integrated_channels/integrated_channel/models.py
@@ -8,7 +8,18 @@ from django.utils.encoding import python_2_unicode_compatible
 from django.utils.translation import ugettext_lazy as _
 from model_utils.models import TimeStampedModel
 
-from enterprise.models import EnterpriseCustomer
+from enterprise.models import EnterpriseCustomer, EnterpriseCourseEnrollment
+from enterprise.utils import NotConnectedToOpenEdX
+
+try:
+    from certificates.api import GeneratedCertificate
+except ImportError:
+    GeneratedCertificate = None
+
+try:
+    from opaque_keys.edx.keys import CourseKey
+except ImportError:
+    CourseKey = None
 
 
 @python_2_unicode_compatible
@@ -62,3 +73,70 @@ class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
 
     class Meta:
         abstract = True
+
+    @staticmethod
+    def channel_code():
+        """
+        Returns an capitalized identifier for this channel class, unique among subclasses.
+        """
+        raise NotImplementedError(_('Implemented in concrete subclass.'))
+
+    def transmit_learner_data(self):
+        """
+        Collect and transmit learner data for the ``EnterpriseCustomer``.
+        """
+        transmitter = self.get_learner_data_transmitter()
+        for learner_data in self.collect_learner_data():
+            transmitter.transmit(learner_data)
+
+    def collect_learner_data(self):
+        """
+        Collect learner data for the ``EnterpriseCustomer`` where data sharing consent is granted.
+
+        Yields a learner data object for each enrollment, containing:
+
+        * ``enterprise_enrollment``: ``EnterpriseCourseEnrollment`` object.
+        * ``certificate``: ``GeneratedCertificate`` for the user+course; None if not found.
+          "Course completion" occurs when course certificates are issued.
+        """
+
+        if CourseKey is None or GeneratedCertificate is None:
+            raise NotConnectedToOpenEdX(_('This package must be installed in an OpenEdX environment.'))
+
+        # Fetch the consenting enrollment data, including the enterprise_customer_user
+        enrollment_queryset = EnterpriseCourseEnrollment.objects.select_related(
+            'enterprise_customer_user'
+        ).filter(
+            enterprise_customer_user__enterprise_customer=self.enterprise_customer,
+        )
+        for enterprise_enrollment in enrollment_queryset:
+
+            # Omit any enrollments where consent has not been granted
+            if not enterprise_enrollment.consent_available():
+                continue
+
+            # Fetch the user+course certificate, if found
+            try:
+                certificate = GeneratedCertificate.eligible_certificates.get(
+                    user__id=enterprise_enrollment.enterprise_customer_user.user_id,
+                    course_id=CourseKey.from_string(enterprise_enrollment.course_id),
+                )
+            except GeneratedCertificate.DoesNotExist:
+                certificate = None
+
+            yield self.get_learner_data(
+                enterprise_enrollment=enterprise_enrollment,
+                certificate=certificate,
+            )
+
+    def get_learner_data(self, enterprise_enrollment, certificate):
+        """
+        Returns the class that can serialize the learner completion data to the integrated channel.
+        """
+        raise NotImplementedError(_('Implemented in concrete subclass.'))
+
+    def get_learner_data_transmitter(self):
+        """
+        Returns the class that can transmit the learner completion data to the integrated channel.
+        """
+        raise NotImplementedError(_('Implemented in concrete subclass.'))
--- a/integrated_channels/sap_success_factors/models.py
+++ b/integrated_channels/sap_success_factors/models.py
@@ -4,15 +4,19 @@ Database models for Enterprise Integrated Channel SAP SuccessFactors.
 
 from __future__ import absolute_import, unicode_literals
 
+import json
+from datetime import datetime
 from config_models.models import ConfigurationModel
 from simple_history.models import HistoricalRecords
 
 from django.db import models
+from django.utils import timezone
 from django.utils.encoding import python_2_unicode_compatible
 
 from model_utils.models import TimeFramedModel
 
 from integrated_channels.integrated_channel.models import EnterpriseCustomerPluginConfiguration
+from .transmitters.learner_data import SuccessFactorsLearnerDataTransmitter
 
 
 @python_2_unicode_compatible
@@ -75,6 +79,30 @@ class SAPSuccessFactorsEnterpriseCustomerConfiguration(EnterpriseCustomerPluginC
         """
         return self.__str__()
 
+    @staticmethod
+    def channel_code():
+        """
+        Returns an capitalized identifier for this channel class, unique among subclasses.
+        """
+        return 'SAP'
+
+    def get_learner_data(self, enterprise_enrollment, certificate):
+        """
+        Returns a LearnerDataTransmissionAudit initialized from the given enrollment and certificate data.
+        """
+        return LearnerDataTransmissionAudit(
+            enterprise_enrollment=enterprise_enrollment,
+            certificate=certificate,
+            # Used only if certificate is None
+            course_completed=False,
+        )
+
+    def get_learner_data_transmitter(self):
+        """
+        Returns a SuccessFactorsLearnerDataTransmitter instance.
+        """
+        return SuccessFactorsLearnerDataTransmitter()
+
 
 @python_2_unicode_compatible
 class LearnerDataTransmissionAudit(models.Model):
@@ -93,8 +121,36 @@ class LearnerDataTransmissionAudit(models.Model):
     error_message = models.TextField(blank=True)
     created = models.DateTimeField(auto_now_add=True)
 
+    epoch = datetime(1970, 1, 1, tzinfo=timezone.utc)
+
     class Meta:
         app_label = 'sap_success_factors'
+
+    def __init__(self, *args, **kwargs):
+        """
+        Return a new ``LearnerDataTransmissionAudit`` object.
+
+        Optional arguments:
+        * ``enterprise_enrollment``: If provided, then ``enterprise_course_enrollment_id``, ``user_id``, * ``course_id``
+          are pulled from this EnterpriseCourseEnrollment object.
+        * ``certificate``: If provided, then ``course_completed=True``; ``completed_timestamp`, ``grade``
+          are pulled from this GeneratedCertificate object.
+
+        Otherwise, the field values are pulled directly from the kwargs, as usual.
+        """
+        enterprise_enrollment = kwargs.pop('enterprise_enrollment', None)
+        if enterprise_enrollment is not None:
+            kwargs['enterprise_course_enrollment_id'] = enterprise_enrollment.id
+            kwargs['sapsf_user_id'] = enterprise_enrollment.enterprise_customer_user.get_remote_id()
+            kwargs['course_id'] = enterprise_enrollment.course_id
+
+        certificate = kwargs.pop('certificate', None)
+        if certificate is not None:
+            kwargs['course_completed'] = True
+            kwargs['completed_timestamp'] = int((certificate.created_date - self.epoch).total_seconds())
+            kwargs['grade'] = certificate.grade
+
+        super(LearnerDataTransmissionAudit, self).__init__(*args, **kwargs)
 
     def __str__(self):
         """
@@ -112,6 +168,45 @@ class LearnerDataTransmissionAudit(models.Model):
         Return uniquely identifying string representation.
         """
         return self.__str__()
+
+    @property
+    def provider_id(self):
+        '''
+        Fetch ``provider_id`` from global configuration settings
+        '''
+        return SAPSuccessFactorsGlobalConfiguration.current().provider_id
+
+    def _payload_data(self, empty_value):
+        """
+        Convert the audit record's fields into SAP SuccessFactors key/value pairs.
+        """
+        return dict(
+            userID=self.sapsf_user_id,
+            courseID=self.course_id,
+            providerID=self.provider_id,
+            # SAP SuccessFactors requires strings, not boolean values.
+            courseCompleted="true" if self.course_completed else "false",
+            completedTimestamp=self.completed_timestamp,
+            grade=self.grade,
+            # TODO: determine these values from the enrollment seat?
+            price=empty_value,
+            currency=empty_value,
+            creditHours=empty_value,
+            # These fields currently have no edX source, so remain empty.
+            totalHours=empty_value,
+            contactHours=empty_value,
+            cpeHours=empty_value,
+            instructorName=empty_value,
+            comments=empty_value,
+        )
+
+    def serialize(self, empty_value=''):
+        """
+        Return a JSON-serialized representation.
+
+        Sort the keys so the result is consistent and testable.
+        """
+        return json.dumps(self._payload_data(empty_value=empty_value), sort_keys=True)
 
 
 @python_2_unicode_compatible
--- a/integrated_channels/sap_success_factors/transmitters/learner_data.py
+++ b/integrated_channels/sap_success_factors/transmitters/learner_data.py
@@ -2,7 +2,11 @@
 Class for transmitting learner data to SuccessFactors.
 """
 from __future__ import absolute_import, unicode_literals
+import logging
 from integrated_channels.sap_success_factors.transmitters import SuccessFactorsTransmitterBase
+
+
+LOGGER = logging.getLogger(__name__)
 
 
 class SuccessFactorsLearnerDataTransmitter(SuccessFactorsTransmitterBase):
@@ -28,5 +32,13 @@ class SuccessFactorsLearnerDataTransmitter(SuccessFactorsTransmitterBase):
     (taken from OCNWebServicesConfiguration and OCNWebServicesEnterpriseCustomerConfiguration) and the oauth token.
 
     Record information about the success/failure of posting the learner data to SuccessFactors in
-     CompletionStatusEventAudit
+    the LearnerDataTransmissionAudit object provided, and save it to the database.
+
     """
+    def transmit(self, learner_data):
+        """
+        Transmit the learner data payload to the SAP SuccessFactors channel.
+        """
+        # TODO ENT-221 will handle transmitting the learner data payload to the channel
+        # We only want to transmit if learner_data.course_completed == True.
+        LOGGER.info(learner_data.serialize())
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,3 +4,4 @@ Jesse Shapiro <jesse@opencraft.com>
 Saleem Latif <saleem_ee@hotmail.com>
 Sven Marnach <sven@opencraft.com>
 Brittney Exline <bexline@edx.org>
+Jillian Vogel <jill@opencraft.com>
--- a/docs/configuration_and_usage.rst
+++ b/docs/configuration_and_usage.rst
@@ -99,6 +99,77 @@ to the help strings provided by the form.
 You can preview emails in the template edit view using the "Preview (program)" and "Preview (course)" buttons in
 top-right corner.
 
+Integrated Channels
+-------------------
+
+SAP SuccessFactors Configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+One of the integrated channels available for an ``EnterpriseCustomer`` is `SAP SuccessFactors`_.
+
+To integrate an ``EnterpriseCustomer`` with SAP SuccessFactors:
+
+* Set up a `TPA SAML Integration`_ for the SAP SuccessFactors endpoint.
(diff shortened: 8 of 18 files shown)


edx-webhook:
Thanks for the pull request, @pomegranited! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=59&repo=edx%2Fedx-enterprise"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
```diff
@@ -0,0 +1,84 @@
...
+        )
+
+    def handle(self, *args, **options):
+        """
+        Collect the learner data for the given EnterpriseCustomer, and transmit it to its active integration channels.
+        """
+        # Ensure there's a valid enterprise customer
+        enterprise_customer_uuid = options['enterprise_customer']
+        if not enterprise_customer_uuid:
+            raise CommandError('Enterprise customer uuid required.')
```
haikuginger: @pomegranited, I was actually talking to @brittneyexline, and it looks like for Jenkins, we need to run one single command to do everything - we can't run multiple instances with EC UUIDs as arguments. I'll get a more thorough review in later today, but we'd like to:

1. Accept EC UUIDs as an _optional_ parameter, with behavior as seen here
2. When an EC UUID is _not_ provided, retrieve a list of EnterpriseCustomers, and perform the task for each EC UUID in separate Celery tasks

Sorry for the late-breaking change; we've been simultaneously talking with Ops about how the EdX Jenkins tasks run.

pomegranited: No worries, fixed with 0228c50.

pomegranited:
@haikuginger I've got one test failure that looks unrelated to my code.  Any idea what's going on there?

CHANGELOG.rst
```diff
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.26.3] - 2017-03-02
```
haikuginger: This would be a new version - 0.27.0 likely. However, I generally try to avoid bumping this until right before merge to avoid merge conflicts from other PRs.

We only do revision version bumps for bugfixes; everything else is a minor version.

integrated_channels/integrated_channel/models.py
```diff
@@ -62,3 +77,55 @@ class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
...
+
+        Yields a dict for each enrollment, containing:
+
+        * enterprise_enrollment: EnterpriseCourseEnrollment
+        * certificate: GeneratedCertificate for the user+course, None if not found.
+          "Course completion" occurs when course certificates are issued.
+        """
+
+        # Fetch the consenting enrollment data, including the enterprise_customer_user
+        enrollment_queryset = EnterpriseCourseEnrollment.objects.select_related(
```
haikuginger: We also need to handle the account-level UserDataSharingConsentAudit object.

pomegranited: @haikuginger Could you explain the interaction between these two objects?  If they conflict, which takes precedence?

CHANGELOG.rst
pomegranited: @haikuginger  This change came from merging ENT-191, since it was a dependency.  The final diff shouldn't include this.

But thanks for letting me know whether this change is a major or minor version bump.

integrated_channels/integrated_channel/models.py
```diff
@@ -62,3 +77,55 @@ class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
...
         abstract = True
+
+    def transmit_learner_data(self):
+        """
+        Collect and transmit learner data for the enterprise_customer.
+        """
+        for learner_data in self.collect_learner_data():
+            LOGGER.info(learner_data)
+
+            # TODO ENT-221 will handle transmitting the learner data to the channel
```
brittneyexline: FYI, I had intended the logic for sending data to a channel to live in the transmitters folder of each integrated_channel, so if you want to create a stub method to call that could be a place to put it.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
```diff
@@ -0,0 +1,96 @@
...
+            default=None,
+            metavar='ENTERPRISE_CUSTOMER_UUID',
+            help='Transmit data for this EnterpriseCustomer to its configured integration channel(s).',
+        )
+        parser.add_argument(
+            '--channel',
+            dest='channel',
+            default=None,
+            metavar='INTEGRATED_CHANNEL',
+            help='Transmit data to this INTEGRATED_CHANNEL.  '
```
mattdrayer: INTEGRATED_CHANNEL can probably be IntegratedChannel for this help string

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
```diff
@@ -0,0 +1,96 @@
...
+        else:
+            enterprise_customer = None
+
+        channel_code = options['channel']
+        if channel_code is not None:
+            if channel_code not in self.integrated_channel_choices:
+                raise CommandError('Invalid integrated channel: {}'.format(channel_code))
+            plugin_classes = [self.integrated_channel_choices.get(channel_code)]
+        else:
+            # Otherwise, use all the configured channel types
```
mattdrayer: imho, this comment would read better if we instead explain what is going on overall, just prior to the `if` statement.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
```diff
@@ -0,0 +1,96 @@
...
+        active_integrated_channel_found = False
+        for plugin_class in plugin_classes:
+            integrated_channels = plugin_class.objects.filter(active=True)
+            if enterprise_customer:
+                integrated_channels.filter(enterprise_customer=enterprise_customer)
+
+            # Transmit the learner data to each integrated channel
+            for integrated_channel in integrated_channels:
+                self.transmit_learner_data(integrated_channel)
+                active_integrated_channel_found = True
```
mattdrayer: Do we only care if an active channel was found?  Does it matter (in this context) if data was actually transmitted?

integrated_channels/integrated_channel/models.py
```diff
@@ -3,12 +3,27 @@
...
 
+import logging
 from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.translation import ugettext_lazy as _
 from model_utils.models import TimeStampedModel
 
-from enterprise.models import EnterpriseCustomer
+from enterprise.models import EnterpriseCustomer, EnterpriseCourseEnrollment
+from enterprise.utils import NotConnectedToEdX
```
mattdrayer: Just FYI, I believe this exception is being renamed to `NotConnectedToOpenEdx`

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # Ensure there's at least one active, integrated channel for this enterprise customer
+        if not active_integrated_channel_found:
+            raise CommandError('No active integrated channels found{}'.format(
+                ' for enterprise customer {}.'.format(enterprise_customer_uuid) if enterprise_customer else ''
+            ))
+
+    @staticmethod
+    @celery.task
```
mattdrayer: Is this going to be a problem if `celery = None` from above?

integrated_channels/integrated_channel/models.py
```diff
@@ -62,3 +77,55 @@ class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
...
+        # Fetch the consenting enrollment data, including the enterprise_customer_user
+        enrollment_queryset = EnterpriseCourseEnrollment.objects.select_related(
+            'enterprise_customer_user'
+        ).filter(
+            consent_granted=True,
+            enterprise_customer_user__enterprise_customer=self.enterprise_customer,
+        )
+        for enterprise_enrollment in enrollment_queryset:
+
+            if CourseKey is None:
```
mattdrayer: Looks like you might also need to protect against `GeneratedCertificate` being `None` from the import

integrated_channels/integrated_channel/models.py
```diff
@@ -62,3 +77,55 @@ class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
...
+                )
+            except GeneratedCertificate.DoesNotExist:
+                certificate = None
+
+            yield self.learner_data_record(
+                enterprise_enrollment=enterprise_enrollment,
+                certificate=certificate,
+            )
+
+    def learner_data_record(self, enterprise_enrollment, certificate=None):
```
mattdrayer: this function is named like a property, but it's not actually a property...might make more sense to name it `get_learner_data_record` or just `get_learner_data`

mattdrayer: Actually if it returns a class/implementation we should probably name it along those lines -- eg `get_learner_data_class`

integrated_channels/sap_success_factors/models.py
```diff
@@ -143,3 +155,74 @@ def __repr__(self):
         Return uniquely identifying string representation.
         """
         return self.__str__()
+
+
+class SAPSuccessFactorsLearnerCompletionPayload(object):
```
mattdrayer: This is a pattern that I've seen in various locations but unfortunately we're not very consistent with it in Open edX -- sometimes there is an FK and sometimes the data is replicated into the audit record with no actual FK defined.  Does anyone feel strongly one way or the other?

integrated_channels/sap_success_factors/models.py
```diff
@@ -143,3 +155,74 @@ def __repr__(self):
...
+            completed_timestamp = None
+            course_completed = "false"
+            grade = cls.empty_value
+        else:
+            completed_timestamp = int((certificate.created_date - cls.epoch).total_seconds())
+            course_completed = "true"
+            grade = certificate.grade
+
+        # Get the price and currency from the enrollment seat
+        price = cls.empty_value
```
mattdrayer: @brittneyexline are we planning on actually populating these values?  If not we could just set them in the `return` below along with the other empty values.

integrated_channels/sap_success_factors/models.py
pomegranited: I agree with @brittneyexline  - I think the `LearnerDataTransmissionAudit` is sufficient.  Fewer classes is cleaner.

37299d0 moves the payload generation functionality I had in `SAPSuccessFactorsLearnerCompletionPayload` into a new `payload()` method.

It also overrides `__init__` to allow some fields to be pulled from optional `EnterpriseCourseEnrollment` and `GeneratedCertificate` objects.

integrated_channels/integrated_channel/models.py
pomegranited: 37299d0 changes this to [`get_learner_data()`](https://github.com/edx/edx-enterprise/pull/59/commits/37299d0df0e34dd0e9f00da8bfacee8900f2561e#diff-4386ccd909983c900b04b5a992e66fddR127), which returns a `LearnerDataTransmissionAudit`

`LearnerDataTransmissionAudit` now has a `payload()` method, which does the serialization.

integrated_channels/sap_success_factors/models.py
```diff
@@ -143,3 +155,74 @@ def __repr__(self):
...
+        currency = cls.empty_value
+        credit_hours = cls.empty_value
+
+        return SAPSuccessFactorsLearnerCompletionPayload(
+            edxUserID=user_id,
```
pomegranited: Yep, sorry about that.  Fixed in 37299d0 - see [`_payload_data()`](https://github.com/edx/edx-enterprise/pull/59/files#diff-b2d1ad7443d2344126a60d0213e2d192R174)

integrated_channels/integrated_channel/models.py
pomegranited: Cool, thanks for letting me know...  I'll need to rebase to get that change in, but will wait for the next round of reviews to be done first.

integrated_channels/sap_success_factors/models.py
pomegranited: FYI: moved to [`_payload_data()`](https://github.com/edx/edx-enterprise/pull/59/files#diff-b2d1ad7443d2344126a60d0213e2d192R181)

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
pomegranited: Leaving that up to you guys..  do you need the jenkins job to fail if no data gets sent?

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
pomegranited: Addressed by ddb2b7c.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
pomegranited: Addressed by ddb2b7c.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
pomegranited: Good point.. Originally I was only worried about patching it in the tests, but it turned out to be easy to patch in the management command directly.

cf d670e4b

integrated_channels/integrated_channel/models.py
pomegranited: Yep!  Thanks for catching that.  Fixed in 01d3678.

integrated_channels/integrated_channel/models.py
pomegranited: No worries, I've addressed this with 1cbb639.

pomegranited:
@brittneyexline @haikuginger @mattdrayer Thank you for your comments! I think the code is looking much better now.

I think I've addressed all of the review issues so far, except for:

* [How do `EnterpriseCourseEnrollment.consent_granted` and `UserDataSharingConsentAudit` interact?](https://github.com/edx/edx-enterprise/pull/59#discussion_r104481471)
* [Should the management command die if no data is sent?](https://github.com/edx/edx-enterprise/pull/59#discussion_r104577953) 
* [Separate ENT ticket to use Certificates API instead of `GeneratedCertificate` model?](https://github.com/edx/edx-enterprise/pull/59#discussion_r104571455)
* [`NotConnectedToEdX` is now `NotConnectedToOpenEdx`](https://github.com/edx/edx-enterprise/pull/59#discussion_r104578041) - rebase required, will do once everyone's ok with this first pass.
* merge checklist items - version bumping and changelog still todo.

And @haikuginger , could I get an amen (if it's amen-worthy :smile:) ?

haikuginger:
> 1. How do EnterpriseCourseEnrollment.consent_granted and UserDataSharingConsentAudit interact?

UserDataSharingConsentAudit provides a base level of consent, which EnterpriseCourseEnrollment.consent_granted overrides. You can take a look at the logic inside `EnterpriseCourseEnrollment.consent_available` to see what the overall value is.

> 2. Should the management command die if no data is sent?

I don't think so - as long as we're doing plenty of good logging, I think we're fine letting it succeed.

> 3. Separate ENT ticket to use Certificates API instead of GeneratedCertificate model?

I'll leave that one to @mattdrayer to determine whether it's in-scope for the Enterprise project.

And the behavior looks good to me, so amen! 👍

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
```diff
@@ -0,0 +1,103 @@
+"""
+Transmits consenting enterprise learner data to the Enterprise Integrated Channel SAP SuccessFactors.
```
haikuginger: Nit: this is actually generic for all integrated channels.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
```diff
@@ -0,0 +1,103 @@
...
+"""
+from __future__ import absolute_import, unicode_literals
+
+from django.core.management.base import BaseCommand, CommandError
+
+from enterprise.models import EnterpriseCustomer
+from integrated_channels.sap_success_factors.models import SAPSuccessFactorsEnterpriseCustomerConfiguration
+
+try:
```
haikuginger: Any reason to do it this way, as compared to putting `django-celery` in the requirements and setting `CELERY_ALWAYS_EAGER` in the settings?

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
```diff
@@ -0,0 +1,103 @@
...
+        # Loop through each channel class (optionally for a specific enterprise customer)
+        active_integrated_channel_found = False
+        for channel_class in channel_classes:
+            integrated_channels = channel_class.objects.filter(active=True)
+            if enterprise_customer:
+                integrated_channels.filter(enterprise_customer=enterprise_customer)
+
+            # Transmit the learner data to each integrated channel
+            for integrated_channel in integrated_channels:
+                self.transmit_learner_data(integrated_channel)
```
haikuginger: Do we need to do `apply_async` to get Celery to actually pick it up? Wouldn't just calling the function run it inline, as would normally happen?

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
haikuginger: Because django-celery is broken in testing, that's why. :P

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
haikuginger: I'm doing it with `delay`, and catching the AttributeError to retry if `delay` isn't a valid attribute.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
pomegranited: Ah good catch!  I've moved that file around a lot..  Fixed with 9753039.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
pomegranited: I didn't even know that `@celery.task` wasn't asynchonous by default!  Thanks for noticing that.. could have ground jenkins to a halt 😟 

For code clarity, I chose to have the [no djcelery stub code add a`delay` attribute](https://github.com/open-craft/edx-enterprise/blob/479c463f00006c73500c8333b76ceb2a3d97d250/integrated_channels/integrated_channel/management/commands/__init__.py#L11-L17) to the wrapped function, so you don't have to catch the `AttributeError`.  And since it's useful for your PR too, I've moved both the celery import and the stub into [management/commands/__init__.py](https://github.com/edx/edx-enterprise/pull/59/commits/479c463f00006c73500c8333b76ceb2a3d97d250#diff-11c9715fd5316220a126561ca420c9dcR1).

Let me know if you see any issues with this, or if I should move it somewhere else?

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
pomegranited: 32b7cdd removes this error, and lets the task succeed even if there's no output.

integrated_channels/integrated_channel/models.py
pomegranited: @haikuginger Awesome, thanks for pointing me to the `EnterpriseCourseEnrollment.consent_available()` method.  Fixed with 5f98982.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        # If a valid enterprise customer uuid was provided, transmit for that customer.
+        # Otherwise, transmit for all customers.
+        enterprise_customer_uuid = options['enterprise_customer']
+        if enterprise_customer_uuid:
+            try:
+                enterprise_customer = EnterpriseCustomer.active_customers.get(uuid=enterprise_customer_uuid)
+            except EnterpriseCustomer.DoesNotExist:
+                raise CommandError('Enterprise customer {} not found, or not active'.format(enterprise_customer_uuid))
+        else:
```
mattdrayer: Does this need to be an "else"?  What if you just set `enterprise_customer = None` ahead of L54?  You're already checking to see if it's None down on L76, so maybe we can simplify a tiny bit here.

integrated_channels/sap_success_factors/models.py
```diff
@@ -93,9 +114,41 @@ class LearnerDataTransmissionAudit(models.Model):
...
+          are pulled from this EnterpriseCourseEnrollment object.
+        * ``certificate``: If provided, then ``course_completed=True``; ``completed_timestamp`, ``grade``, ``status``
+          are pulled from this GeneratedCertificate object.
+
+        Otherwise, the field values are pulled directly from the kwargs, as usual.
+        """
+        enterprise_enrollment = kwargs.pop('enterprise_enrollment', None)
+        if enterprise_enrollment is not None:
+            kwargs['enterprise_course_enrollment_id'] = enterprise_enrollment.id
+            # TODO: ENT-164 will provide a way to get the SAP User ID for this user from the SAP SSO backend.
```
mattdrayer: ENT-164 is the overall Epic for this project -- I took a look thru the stories/tasks but did not see anything specific to adding a workflow to retrieve the SAP User ID (which is actually the SAML IdP User ID).  @brittneyexline might have a better idea of where that work is being handled.  In the meantime, here is a link to the documentation on how we store that information in Open edX: http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/configuration/tpa/tpa_IDmap_api.html#third-party-auth-id-mapping-api

integrated_channels/sap_success_factors/transmitters/learner_data.py
```diff
@@ -28,5 +32,13 @@ class SuccessFactorsLearnerDataTransmitter(SuccessFactorsTransmitterBase):
...
-     CompletionStatusEventAudit
+    the LearnerDataTransmissionAudit object provided, and save it to the database.
+
     """
+    def transmit(self, learner_data):
+        """
+        Transmit the learner data payload to the SAP SuccessFactors channel.
+        """
+        # TODO ENT-221 will handle transmitting the learner data payload to the channel
+        # Question: do we only want to transmit if learner_data.course_completed == True?
```
mattdrayer: For the SAP integration it is recommended that we only send a single completion event for each user, which means we'll need to wait until we are confident that the user has completed the course -- this can be difficult to do in certain scenarios, such as a self-paced course.

mattdrayer:
I had one tiny code nit (take or leave) otherwise I just had some additional information/context to share.  👍, let's :shipit:

integrated_channels/sap_success_factors/models.py
pomegranited: I think [ENT-190](https://openedx.atlassian.net/browse/ENT-190) might be the closest story?  I didn't want to pin it to that, in case it required extra work beyond basic SSO.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
pomegranited: Sure thing, done with d3e47c2.

integrated_channels/sap_success_factors/models.py
brittneyexline: My understanding was that there was no new work for the ability to store a third party id for a user coming in through SSO, except that we need to make sure it happens when we configure SAP customers for SSO in our system, which implies a soft dependency on ENT-190. But now that Matt has found the readthedocs about this functionality, does this sufficiently address your todo @pomegranited ?

pomegranited: Ok yep..  I didn't know it was something I could do in advance of the SSO integration.  I'll add a `enterprise.lms_api.ThirdPartyAuthIdClient` to fetch the remote user ID, and mock its response in the tests.

If we fail to find a remote user ID, I'll log an error, and not generate a learner completion record for that user.   Sound ok?

integrated_channels/sap_success_factors/transmitters/learner_data.py
pomegranited: Yep, I don't know what to do about self-paced courses right now, certificates are all I've got.

For this line though, do you want me to skip the record if `not learner_data.course_completed`?  Currently, nothing is being saved to the database; I was waiting for ENT-221 to do that.

brittneyexline: Now that I have a first pass at the transmit function (https://github.com/edx/edx-enterprise/pull/65), which is doing the most basic thing it can do, I think it might be easier to figure out how to handle when to send the payload.
SAP is expecting us to send this payload only once, when the user has completed a course. Using certificate generation as it stands in this PR, we are covering 3 out of 4 cases. The case we are not handling is audit/self-paced learners, but this might end up being a follow up item at this point.
It is unclear to me from reading through this PR if the LearnerDataTransmissionAudit object is getting persisted. I feel that the transmit function would ideally: 
1. Determine if an event was already sent for this learner/course, referencing the audit table.
2. Perform the call to SuccessFactors if this is a new event, else exit.
3. Save a new audit object reflecting the status of the call.
What do you think? If you agree I will add this into my implementation, but I want to make sure it makes sense given what you're doing.

integrated_channels/sap_success_factors/models.py
brittneyexline: I think that makes sense :-)

integrated_channels/sap_success_factors/transmitters/learner_data.py
pomegranited: Sounds good to me.

> The case we are not handling is audit/self-paced learners, but this might end up being a follow up item at this point.

A follow-up item would be great.  We can use the Grades API to fetch the current grade, but we need to know what the cut-off point is, especially since we can only send once per enrolment.

> It is unclear to me from reading through this PR if the LearnerDataTransmissionAudit object is getting persisted.
The `LearnerDataTransmissionAudit` object is currently only being created, not persisted.

> 1. Determine if an event was already sent for this learner/course, referencing the audit table.
Yep, uniqueness can be easily done with the `enterprise_course_enrollment_id`.  It sounds like we also want to check that the `course_completed == True` before saving.

> 2. Perform the call to SuccessFactors if this is a new event, else exit.
Yes.

> 3. Save a new audit object reflecting the status of the call.
Yep.  Can use the `status` and `error_message` fields - I wasn't sure if [`status` was supposed to come from the `certificate` or not?](https://github.com/edx/edx-enterprise/pull/59/files#diff-b2d1ad7443d2344126a60d0213e2d192R148).  I can fix that if it's intended for the SAP response.

I assume this will be part of a separate PR, after #59 and #65 merge?

integrated_channels/sap_success_factors/models.py
pomegranited: Ok, I'll get on that today.  Should have something for you by the time you start your day tomorrow 😄

pomegranited: P.S. Is it ok for me to rebase on master to address the `NotConnectedToOpenEdx` change?

brittneyexline: I don't see why not

integrated_channels/sap_success_factors/transmitters/learner_data.py
brittneyexline: I will update 65 to reflect this functionality. I was intending to use status for the SAP response.

integrated_channels/sap_success_factors/models.py
pomegranited: Please see 8f6d386 for the addition of the `ThirdPartyAuthApiClient` class, which is used to fetch the `remote_id` for the `EnterpriseCustomerUser`'s username.

Also updated the test instructions above to reflect the additional IDP setup.

pomegranited:
@brittneyexline @mattdrayer @haikuginger  Ok, I think I've addressed all of your comments, and am now just waiting for the tests to pass.

Tomorrow, if all is ok, I'll go through the merge checklist.  What should I use for this PR's version number?

Thanks for all your help and advice!

integrated_channels/integrated_channel/management/commands/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+            channel_classes = [self.integrated_channel_choices.get(channel_code)]
+        else:
+            channel_classes = self.integrated_channel_choices.values()
+
+        # Loop through each channel class (optionally for a specific enterprise customer)
+        for channel_class in channel_classes:
+            # Use Active channels only
+            integrated_channels = channel_class.objects.filter(active=True)
+            if enterprise_customer:
+                integrated_channels.filter(enterprise_customer=enterprise_customer)
```
haikuginger: I think you need to assign this to a variable - if you just do `.filter()` on a queryset, it doesn't change the original item.

pomegranited: Oops, good catch!  Fixed with b27f67c.

pomegranited:
@haikuginger Have made the changes we discussed:

* dd13542 implements the celery serialization changes
* Ensures that my data object has a `serialize` method that the SAP client can use to generate a JSON payload.  If we end up needing different serialization formats, we can deal with that then.

Sound ok?

pomegranited:
@brittneyexline Could I get 👍  from you, so I know if it's ok to squash these commits and start working through the merge checklist?

pomegranited:
Changes squashed and merge checklist addressed.  @mattdrayer Ok to merge?

*Notes*:
* Version bumped to `0.27.0`.
* [Updated Configuration and Usage docs](https://github.com/edx/edx-enterprise/pull/59/commits/75da853aa300744136d7200fdeef68a299196661#diff-39096f6801e1aeb926636974b0ec9f1e) to add a section for SAP SuccessFactors, and the new management command. Can these be reviewed/updated as part of the other SAP SuccessFactors tasks?
* I've ensured that the user-facing strings are wrapped in `ugettext`.  However, when I ran `make extract_translations`, I got this error, which doesn't look related to this PR.

   ```
     File "/edx/app/edx-enterprise/edx-enterprise/enterprise/admin/utils.py", line 42, in ValidationMessages
    BOTH_FIELDS_SPECIFIED = _("Either \"Email or Username\" or \"CSV bulk upload\" must be specified, but both were.")

   ...
   ValueError: plural forms expression could be dangerous
   make: *** [extract_translations] Error 1
   ```
  Am hoping maybe https://github.com/edx/edx-enterprise/pull/62#issuecomment-285262523 addresses this error?

integrated_channels/sap_success_factors/models.py
```diff
@@ -93,9 +121,37 @@ class LearnerDataTransmissionAudit(models.Model):
...
+        enterprise_enrollment = kwargs.pop('enterprise_enrollment', None)
+        if enterprise_enrollment is not None:
+            kwargs['enterprise_course_enrollment_id'] = enterprise_enrollment.id
+            kwargs['sapsf_user_id'] = enterprise_enrollment.enterprise_customer_user.get_remote_id()
+            kwargs['course_id'] = enterprise_enrollment.course_id
+
+        certificate = kwargs.pop('certificate', None)
+        if certificate is not None:
+            kwargs['course_completed'] = True
+            kwargs['completed_timestamp'] = int((certificate.created_date - self.epoch).total_seconds())
```
pomegranited: Is this supposed to be in seconds or milliseconds?

mattdrayer: Hmm, good question -- @brittneyexline do you know what the requirement is here?

mattdrayer:
I'm not sure if getting an answer to the question about the time units is a blocker for this PR -- otherwise I think it looks great and fingers-crossed that the string extraction issue will be resolved by #62 -- assuming the time units are fine I think we're good to move forward -- 👍

integrated_channels/sap_success_factors/models.py
pomegranited: The learner completion documentation is inconclusive.  The example given doesn't provide a number:

```
{
"userID": "<USERID>",
"courseID":"<COURSEID>",
"providerID":<PROVIDERID>,
"courseCompleted":"true",
"completedTimestamp":<COMPLETEDTIMESTAMP>,
"instructorName":"<InstructorName>",
"grade":"<grade>",
"totalHours":"<TOTALHOURS>",
"creditHours":<CREDITHOURS>,
"contactHours":<CONTACTHOURS>,
"cpeHours":<CPE>,
"comments":<COMMENTS>,
"price":<PRICE>,
"currency":<CURRENCY>
}
```

And the field description doesn't help: 
> This is the Unix Epoch time that the user completed the course. SAP SuccessFac­tors Learning uses this completion date for the audit trail of users' completion.

pomegranited:
Rebased 9a09bd0 on master, and bumped version to `0.27.2`.

Ok to merge now, @mattdrayer, @brittneyexline , @haikuginger ?

integrated_channels/sap_success_factors/models.py
mattdrayer: @brittneyexline let's definitely add this to our list of questions for the Wednesday call

mattdrayer:
I've gone through the testing instructions on the sandbox and everything looks good to me -- I think it's time to merge, yep!

pomegranited:
Thanks @mattdrayer !  Merging and tagging now.

pomegranited:
@mattdrayer  The [tag-triggered build has succeeded](https://travis-ci.org/edx/edx-enterprise/builds/211198187)..  How do I push the tagged build to PyPi?

mattdrayer:
I believe PyPI should magically update via Travis -- I see version 27.2 is there now:  https://pypi.python.org/pypi/edx-enterprise/

pomegranited:
Nifty!  Thanks @mattdrayer .

PR #62, opened 2017-03-06
ENT-232: Add catalog and catalog courses endpoint to enterprise service

saleem-latif:
Hi @mattdrayer , @asadiqbal08 , @haikuginger ,

Kindly review this PR.

**Description:**
This PR adds api endpoints to fetch list of catalogs, catalog derails and catalog courses.

**JIRA:** https://openedx.atlassian.net/browse/ENT-232

**Testing instructions:**

1. Open enterprise api main page at '/enterprise/api/v1/' and make sure that a new api endpoint for catalogs is present (sample: "catalogs": "http://open.edx:8000/enterprise/api/v1/catalogs/").  
2. Open catalogs api endpoint at `enterprise/api/v1/catalogs/` and make sure it list all the catalogs accessible to the logged in user (All catalogs are accessible for staff users.).
3. Open catalog details page at `enterprise/api/v1/catalogs/1/`, and verify that catalog details are correct. Make sure `404` is returned if you try to access a catalog that does not exist.
4. Finally Open a catalog courses page at `enterprise/api/v1/catalogs/1/courses` and make sure all the courses are returned present in the selected catalog. Make sure pagination links are working.

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag


saleem-latif:
@e-kolpakov I am getting this error on `make upgrade`. thoughts on how to fix this ?

```
pip-compile --upgrade -o requirements/dev.txt requirements/base.in requirements/dev.in requirements/quality.in
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/bin/pip-compile", line 11, in <module>
    sys.exit(cli())
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/lib/python3.5/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/lib/python3.5/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/lib/python3.5/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/lib/python3.5/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/lib/python3.5/site-packages/piptools/scripts/compile.py", line 223, in cli
    reverse_dependencies = resolver.reverse_dependencies(results)
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/lib/python3.5/site-packages/piptools/resolver.py", line 273, in reverse_dependencies
    return self.dependency_cache.reverse_dependencies(non_editable)
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/lib/python3.5/site-packages/piptools/cache.py", line 139, in reverse_dependencies
    return self._reverse_dependencies(ireqs_as_cache_values)
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/lib/python3.5/site-packages/piptools/cache.py", line 163, in _reverse_dependencies
    for name, version_and_extras in cache_keys
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/lib/python3.5/site-packages/piptools/utils.py", line 192, in lookup_table
    for value in values:
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/lib/python3.5/site-packages/piptools/cache.py", line 164, in <genexpr>
    for dep_name in self.cache[name][version_and_extras])
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2986, in parse
    reqs = list(parse_requirements(s))
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2930, in parse_requirements
    "version spec")
  File "/Users/saleemlatif/venvs/edx-enterprise-p3/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2906, in scan_list
    raise RequirementParseError(msg, line, "at", line[p:])
pkg_resources.RequirementParseError: Expected ',' or end-of-list in virtualenv>=1.11.2; python_version != "3.2" at ; python_version != "3.2"

```

enterprise/api/v1/serializers.py
```diff
@@ -214,3 +215,88 @@ def update(self, instance, validated_data):
...
+        """
+        pass
+
+
+class EnterpriseCatalogCoursesSerializer(serializers.Serializer):
+    """
+    Serializer for enterprise customer catalog courses.
+    """
+    count = serializers.IntegerField(read_only=True, help_text=_('Total count of catalog courses.'))
+    next = serializers.CharField(read_only=True, help_text=_("URL to fetch next page of courses."))
```
asadiqbal08: is it python builtin iterator  ?

enterprise/api/pagination.py
```diff
@@ -0,0 +1,46 @@
+"""
+Pagination callses for enterprise api.
```
asadiqbal08: nit: `callses` ?

enterprise/api/v1/serializers.py
saleem-latif: yes, but we can not change it to something as this field is expected in catalog api response.

enterprise/api/v1/serializers.py
```diff
@@ -214,3 +215,88 @@ def update(self, instance, validated_data):
...
+            'tpa_hint': utils.get_identity_provider_id_for_user(request.user)
+        }
+
+        for course in self.data['results']:
+            courses.append(
+                self.update_course(course, request, global_context)
+            )
+        self.data['results'] = courses
+
+    def update_course(self, course, request, global_context):  # pylint: disable=unused-argument
```
asadiqbal08: It seems me, This function is not using anywhere else and contains a one line operation of `update`. Is it possible to remove this function  ?

enterprise/api/v1/views.py
```diff
@@ -188,3 +195,80 @@ class EnterpriseCustomerEntitlementViewSet(EnterpriseReadOnlyModelViewSet):
...
+        Returns:
+            (Response): DRF response object containing course catalogs.
+        """
+        # fetch course catalog for the given catalog id.
+        catalog_api = CourseCatalogApiClient(request.user)
+        catalog = catalog_api.get_catalog(pk)
+
+        if not catalog:
+            logger.error("Unable to fetch API response for given catalog from endpoint '/catalog/%s/'.", pk)
+            raise NotFound("Resource your are looking for does not exist.")
```
asadiqbal08: nit: `you`
nit: `The resource you are looking for does not exist`

enterprise/api/v1/views.py
```diff
@@ -188,3 +195,80 @@ class EnterpriseCustomerEntitlementViewSet(EnterpriseReadOnlyModelViewSet):
...
+        courses = catalog_api.get_paginated_catalog_courses(pk, page)
+
+        # if API returned an empty response, that means pagination has ended.
+        # An empty response can also means that there was a problem fetching data from catalog API.
+        if not courses:
+            logger.error(
+                "Unable to fetch API response for catalog courses from endpoint '/catalog/%s/course?page=%s'.",
+                pk, page,
+            )
+            raise NotFound("Resource your are looking for does not exist.")
```
asadiqbal08: nit: you
nit: The resource you are looking for does not exist.

enterprise/utils.py
```diff
@@ -419,3 +419,30 @@ def get_enterprise_branding_info_by_ec_uuid(ec_uuid=None):  # pylint: disable=in
...
+    for any other case return `None`.
+
+    Arguments:
+        auth_user (contrib.auth.User): Django User
+
+    Returns:
+        (slug): slug identifier for the identity provider that can be used for identity verification of
+            users associated the enterprise customer of the given user.
+    """
+    ecu = enterprise.models.EnterpriseCustomerUser.objects.filter(user_id=auth_user.id).first()
```
asadiqbal08: can we write `enterprise_customer_user` instead of `ecu` ?

enterprise/api/v1/serializers.py
saleem-latif: I added this function as it would be helpful for future modifications of the serializer.

asadiqbal08: at that time it is creating a confusion point. `update_enterprise_courses` and then `update_course` with only one line of action :)

enterprise/api/v1/serializers.py
```diff
@@ -214,3 +215,88 @@ def update(self, instance, validated_data):
...
+    next = serializers.CharField(read_only=True, help_text=_("URL to fetch next page of courses."))
+    previous = serializers.CharField(read_only=True, help_text=_("URL to fetch previous page of courses."))
+    results = serializers.ListField(read_only=True, help_text=_("list of courses."))
+
+    def update_enterprise_courses(self, request):
+        """
+        This method adds enterprise specific metadata for each course.
+
+        We are adding following fields in all the courses.
+            1. tpa_hint: a string for identifying Identity Provider.
```
asadiqbal08: currently, is it not adding only one field (tpa_hint) instead of fields ?

saleem-latif: yes, in future we might add more fields but currently we are adding only one.

asadiqbal08: So we will make bullets in future :)

e-kolpakov:
@saleem-latif Unfortunately, no, I haven't faced the error like that until now. My intuition tells that it looks much like dependency problem, but I have no hard proofs to that of course. Myabe try removing stuff one-by-one from `quality.in` / `dev.in` until the error is gone to see which package cause that?

enterprise/api/v1/serializers.py
```diff
@@ -214,3 +215,88 @@ def update(self, instance, validated_data):
     user = UserSerializer(read_only=True)
     enterprise_customer = EnterpriseCustomerSerializer(read_only=True)
     data_sharing_consent = UserDataSharingConsentAuditSerializer(many=True, read_only=True)
+
+
+class EnterpriseCatalogSerializer(serializers.Serializer):
```
mattdrayer: How about `EnterpriseCourseCatalogSerializer`

enterprise/api/v1/serializers.py
```diff
@@ -214,3 +215,88 @@ def update(self, instance, validated_data):
...
+    """
+    Serializer for enterprise customer catalog.
+    """
+    # pylint: disable=invalid-name
+    id = serializers.IntegerField(read_only=True, help_text=_('Course catalog primary key.'))
```
mattdrayer: 'Enteprise course catalog primary key'

enterprise/api/v1/serializers.py
```diff
@@ -214,3 +215,88 @@ def update(self, instance, validated_data):
...
+    name = serializers.CharField(help_text=_('Catalog name'))
+    query = serializers.CharField(help_text=_('Query to retrieve catalog contents'))
```
mattdrayer: should this be `catalog_query`?

enterprise/api/v1/serializers.py
```diff
@@ -214,3 +215,88 @@ def update(self, instance, validated_data):
...
+    courses_count = serializers.IntegerField(read_only=True, help_text=_('Number of courses contained in this catalog'))
+    viewers = serializers.ListField(
```
mattdrayer: Is there a reason why we are including this information?  Does the client require it?

enterprise/api/v1/serializers.py
mattdrayer: Also I see below that we are preventing state-changing requests -- perhaps this should be `EnterpriseCourseCatalogReadOnlySerializer`

enterprise/api/v1/serializers.py
```diff
@@ -214,3 +215,88 @@ def update(self, instance, validated_data):
...
+    next = serializers.CharField(read_only=True, help_text=_("URL to fetch next page of courses."))
+    previous = serializers.CharField(read_only=True, help_text=_("URL to fetch previous page of courses."))
+    results = serializers.ListField(read_only=True, help_text=_("list of courses."))
+
+    def update_enterprise_courses(self, request):
+        """
+        This method adds enterprise specific metadata for each course.
+
+        We are adding following field in all the courses.
+            tpa_hint: a string for identifying Identity Provider.
```
mattdrayer: I am starting to wonder if we should be using `tpa_hint` or if we should instead use the enterprise customer uuid...I think for now it is okay to use tpa_hint but in the future we may add the capability to instead provide the uuid

enterprise/api/v1/serializers.py
```diff
@@ -214,3 +215,88 @@ def update(self, instance, validated_data):
...
+        pass
+
+    def update(self, instance, validated_data):
+        """
+        Do not perform any operations for state changing requests.
+        """
+        pass
+
+
+class EnterpriseCatalogCoursesSerializer(serializers.Serializer):
```
mattdrayer: It seems this is another `ReadOnlySerializer` -- perhaps it should be a mixin?

enterprise/utils.py
```diff
@@ -419,3 +419,30 @@ def get_enterprise_branding_info_by_ec_uuid(ec_uuid=None):  # pylint: disable=in
     return enterprise.models.EnterpriseCustomerBrandingConfiguration.objects.filter(
         enterprise_customer__uuid=ec_uuid
     ).first()
+
+
+def get_identity_provider_id_for_user(auth_user):
```
mattdrayer: @pomegranited @brittneyexline FYI on this method

enterprise/api/v1/serializers.py
saleem-latif: This field should match with the course catalog field from course discovery, where it is `query`. Here is sample catalog from discovery service.
```
{
       "id": 12,
       "name": "Enterprise Dummy Catalog",
       "query": "*",
       "courses_count": 2258,
       "viewers": []
}
```

enterprise/api/v1/serializers.py
saleem-latif: I created a one to one mapping of the fields from discovery service, but I can remove this field if this does not seem useful ?

saleem-latif:
@asadiqbal08 , @mattdrayer I have incorporated your feedback, kindly take another look.

saleem-latif:
@e-kolpakov , @haikuginger It would be great if one of you could take a look at this PR today.

enterprise/api/pagination.py
```diff
@@ -0,0 +1,46 @@
...
+from collections import OrderedDict
+
+from rest_framework.response import Response
+
+from six.moves.urllib.parse import urlparse  # pylint: disable=import-error
+
+
+def get_paginated_response(data, request):
+    """
+    Fix pagination links in course catalog data.
```
mattdrayer: I'm not sure if "Fix" is the right term to use to describe the behavior of this function?

enterprise/api/v1/serializers.py
```diff
@@ -214,3 +215,88 @@ def update(self, instance, validated_data):
...
+        """
+        This method adds enterprise specific metadata for each course.
+
+        We are adding following field in all the courses.
+            tpa_hint: a string for identifying Identity Provider.
+        """
+        courses = []
+
+        global_context = {
+            'tpa_hint': utils.get_identity_provider_id_for_user(request.user)
```
mattdrayer: We need to include the Track Selection URL as well, and it needs to include the tpa_hint parameter -- here is an example: 

`https://courses.edx.org/course_modes/choose/course-v1:DelftX+NUCLEAR01x+1T2017/?tpa_hint={tpa_hint}`

mattdrayer: It probably makes sense to have a new function to `get_course_track_selection_url`

saleem-latif:
@e-kolpakov can you take a look at the last commit to see if plural forms expression  of `Plural-Forms` conforms with `ungettext` usages.

tests/api/test_serializers.py
```diff
@@ -51,3 +64,331 @@ def test_create(self):
...
+                    'sponsors': [],
+                    'modified': '2017-03-03T07:34:19.322916Z',
+                    'full_description': None,
+                    'subjects': [],
+                    'video': None,
+                    'key': 'edX+DemoX',
+                    'short_description': None,
+                    'marketing_url': None,
+                    'level_type': None,
+                    'course_runs': [
```
mattdrayer: Need to add a check for track_selection_url, yes?

mattdrayer:
Generally speaking I'm 👍 with this PR -- let's just make sure that we are including `track_selection_url` in all of the pertinent tests.

mattdrayer:
@haikuginger can you take a look at this today?  @e-kolpakov if you'd like to take a pass thru, too, please feel free although I'd say we only need one more +1 in total.

e-kolpakov:
@saleem-latif Frankly, I just learned about that feature, but judging by [GNU gettext docs|https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html] you've set it up for "2 forms, singular used for 1 only", which seems correct for English.

The only concern I have is - have you modified it manually? If so, what would happen if you run `make extract_translations`? My concern here is essentially that it is an autogenerated file, so any manual changes might be overwritten next time we regenerate it.

So, if you can confirm the follwoing - I'm totally happy with that change:

* `make extract_translations` keeps your changes. At least on your local machine, but if you happen to have access to another machine with different OS (or dual-boot, or VM, or whatever) and ideally with different regional settings - it would be great. But don't go into too much trouble for that - your local should be enough. In case it breaks it on my machine I'll sigh, blame you (kidding) and fix it :).
* run `make compile_translations` to see if it actually compiles.

Bonus track: it would be great if you (or someone else) could add `make compile_translations` step to CI build, so that failed compilation fails CI build as well.
Bonus track 2: it would be totally awesome if we could have a "verify all" make command to run everything that might break: dependencies, tests, quality, docs, translations (both extract and compile).

e-kolpakov:
@mattdrayer @saleem-latif I'd rather not get involved in formal review and approval here, but I've already answered the translations queestion.

haikuginger:
@mattdrayer, will review today.

enterprise/api/v1/views.py
```diff
@@ -188,3 +195,80 @@ class EnterpriseCustomerEntitlementViewSet(EnterpriseReadOnlyModelViewSet):
...
+class EnterpriseCatalogViewSet(viewsets.ViewSet):
+    """
+    API views for `enterprise customer catalogs` api endpoint.
+    """
+    serializer_class = serializers.EnterpriseCourseCatalogReadOnlySerializer
+    permission_classes = (permissions.IsAuthenticated,)
+    authentication_classes = (OAuth2Authentication, SessionAuthentication, BearerAuthentication, JwtAuthentication)
+    throttle_classes = (ServiceUserThrottle,)
+
+    def list(self, request):
```
haikuginger: Do we need this method? What's the use case for an Enterprise consumer being able to list all course catalogs?

enterprise/api/v1/views.py
```diff
@@ -188,3 +195,80 @@ class EnterpriseCustomerEntitlementViewSet(EnterpriseReadOnlyModelViewSet):
...
+        Returns:
+            (Response): DRF response object containing course catalogs.
+        """
+        # fetch all course catalogs.
+        catalog_api = CourseCatalogApiClient(request.user)
+        catalogs = catalog_api.get_all_catalogs()
+        serializer = self.serializer_class(catalogs, many=True)
+        return Response(serializer.data)
+
+    def retrieve(self, request, pk=None):  # pylint: disable=invalid-name
```
haikuginger: What's the use case for using a catalog PK as compared to an EnterpriseCustomer UUID to derive the relevant catalog? If this is EnterpriseCustomer-facing, then indexing by catalog ID means that if the EnterpriseCustomer switches to use a different catalog (for whatever reason), existing clients will keep pulling old data until they're reconfigured.

enterprise/api/v1/views.py
```diff
@@ -188,3 +195,80 @@ class EnterpriseCustomerEntitlementViewSet(EnterpriseReadOnlyModelViewSet):
...
+
+        if not catalog:
+            logger.error("Unable to fetch API response for given catalog from endpoint '/catalog/%s/'.", pk)
+            raise NotFound("The resource you are looking for does not exist.")
+
+        serializer = self.serializer_class(catalog)
+        return Response(serializer.data)
+
+    @detail_route()
+    def courses(self, request, pk=None):  # pylint: disable=invalid-name
```
haikuginger: As above; if this is enterprise-facing, why not use the EnterpriseCustomer UUID?

enterprise/api/v1/serializers.py
```diff
@@ -214,3 +215,106 @@ def update(self, instance, validated_data):
...
+
+    def update_enterprise_courses(self, request):
+        """
+        This method adds enterprise specific metadata for each course.
+
+        We are adding following field in all the courses.
+            tpa_hint: a string for identifying Identity Provider.
+        """
+        courses = []
+        tpa_hint = utils.get_identity_provider_id_for_user(request.user)
```
haikuginger: I'm not sure how I feel about deriving this based on "session" information that's basically hidden - it conflates the concepts of "the user who can retrieve a piece of information" and "the user who can use a piece of information". I'd rather split out "enterprise_customer_uuid" or something similar into the URL, and then derive the tpa_hint based on the URL, with appropriate access restrictions. If we switch to fetching based on EC UUID, as suggested below, rather than on catalog ID, then we get that for free - we can pull the relevant course catalog data for a given customer, and also apply customer-specific transforms to it.

mattdrayer: @haikuginger welcome to [my soapbox](https://openedx.atlassian.net/wiki/display/AC/edX+REST+API+Conventions#edXRESTAPIConventions-ExplicitFilters) 😉 See especially the comments at the bottom of the page.  This is a long, drawn-out argument, unfortunately...

bradenmacdonald: +1 though note that there's talk of multiple catalogs per enterprise, so we need to account for that in the URL.

saleem-latif:
@e-kolpakov Yes, I updated .po files manually, I looked around to see for better approaches but it seems that manually updating po files for `Plural Forms` is the right approach ([ref](https://code.djangoproject.com/ticket/11186)). 

To verify that `make extract_translations ` and `make compile_translations ` do not overwrite `Plural Forms` expression in po files. I ran both of these commands on my local machine and on `zubair-arbi.sandbox.edx.org` and on both machine commands work as expected without overriding `Plural Forms` expressions.

enterprise/api/v1/serializers.py
```diff
@@ -214,3 +215,106 @@ def update(self, instance, validated_data):
...
+
+        global_context = {
```
saleem-latif: @haikuginger `global_context` contains key-value pairs that will be added to all the courses and the `tpa_hint` argument is used by `update_course` method to construct track selection url. 
Currently, `global_context` happens to have `tpa_hint` as key-value pair but it is likely that we change it to something else or remove it in the future. And I do not think it would be a good idea to put any restrictions on `global_context` to have certain value. That is why I passed `tpa_hint` as separate argument instead of retrieving it from `global_context`.

enterprise/api/v1/views.py
saleem-latif: @haikuginger currently, we have only one catalog per customer so this may not seem useful, but in future when we plan to have multiple catalogs for an enterprise customer then it would be necessary to have an API endpoint that returns all course catalogs for an enterprise customer.
At this time this API endpoint will return only one catalog associated with the enterprise customer.

enterprise/api/v1/views.py
saleem-latif: same as above,

using catalog PK is useful because when there are multiple catalogs per enterprise customer there would not be a way to find a catalog for an EnterpriseCustomer UUID (as there will be many). In that scenario enterprise customer would want to fetch catalog by PK.

mattdrayer:
@haikuginger @saleem-latif we need to wrap up this PR -- please resolve any open issues either through modification or discussion and merge it up!

saleem-latif:
@haikuginger I have updated code to resolve most of your concerns and replied to others, please take a look and let me know what do you think.

asadiqbal08:
@saleem-latif  changes LGTM 👍

saleem-latif:
@mattdrayer , @haikuginger I have added fields suggested in [ENT-197](https://openedx.atlassian.net/browse/ENT-197), Coverage will fail for now, I am working on that

haikuginger:
I'm still not thrilled about all the session-based state in here, but it seems to work as expected when paired with edx/edx-platform#14660. I verified that all the different endpoints worked as expected, including the fact that courses were updated with the details from the EnterpriseCustomer linked to the user I was logged in with. When I unlinked the user from the EnterpriseCustomer, those details were null. Approved, once the edx-platform PR is merged and coverage is back up to snuff.

PR #63, opened 2017-03-06
ENT-215/ENT-217: Set up data transformation framework/do transformations for SAP

haikuginger:
**Description:** 

This PR adds a framework for course metadata transformation/transmission to integrated channels.

**JIRA:** [ENT-215](https://openedx.atlassian.net/browse/ENT-215)

**Dependencies:** None

**Merge deadline:** 13 March 2017

**Installation instructions:** Requires course discovery devstack populated with course metadata

**Testing instructions:**

1. Create an `EnterpriseCustomer` with a catalog attached, and with courses present in the catalog.
2. Create an `SAPSuccessFactorsEnterpriseCustomerConfiguration` linked to the EnterpriseCustomer
3. In a devstack, run the SAP management command, and verify that the logging output shows the JSON structure that are expected for SAP. Be sure to use a username with access to the catalog in use:
`./manage.py lms transmit_courseware_data --catalog_user staff`

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=63"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

haikuginger:
@brittneyexline @mattdrayer @pomegranited, this is ready for a first pass. Note that this is based on @pomegranited's current PR and currently is lugging all of her commits around.

enterprise/course_catalog_api.py
```diff
@@ -64,6 +64,30 @@ def get_all_catalogs(self):
         """
         return self._load_data('catalogs', default=[])
 
+    def get_single_catalog_courses(self, catalog_id):
```
mattdrayer: why `get_single_catalog_courses` instead of `get_catalog_courses`?  Is the latter already in-use?

enterprise/course_catalog_api.py
```diff
@@ -64,6 +64,30 @@ def get_all_catalogs(self):
...
+        """
+        Return the courses included in a single course catalog by ID.
+
+        Args:
+            catalog_id (int): The catalog ID we want to retrieve.
+
+        Returns:
+            dict: Courses of the catalog in question
+        """
+        return self._load_data('catalogs/{catalog_id}/courses'.format(catalog_id=catalog_id))
```
mattdrayer: Do we need to protect against an invalid `catalog_id` value?

enterprise/course_catalog_api.py
```diff
@@ -64,6 +64,30 @@ def get_all_catalogs(self):
...
+
+    def get_single_course_details(self, course_key):
```
mattdrayer: Here as well -- why `get_single_course_details` instead of `get_course_datails'?

enterprise/course_catalog_api.py
```diff
@@ -64,6 +64,30 @@ def get_all_catalogs(self):
...
+        """
+        Return the details of a single course by key - not a course run key.
+
+        Args:
+            course_key (str): The unique key for the course in question.
+
+        Returns:
+            dict: Details of the course in question.
+        """
+        return self._load_data('courses', resource_id=course_key, many=False)
```
mattdrayer: Are we concerned at all about invalid `course_key` values?

integrated_channels/integrated_channel/course_metadata.py
```diff
@@ -0,0 +1,127 @@
...
+
+EXCLUDED_COURSE_DETAIL_KEYS = [
+    'course_runs',
+]
+
+
+LOGGER = getLogger(__name__)
+
+
+def updated_course_run(course_details, course_run_details):
```
mattdrayer: This function is named like a property, but it's performing an action -- maybe better to name it `update_course_run_details`?

integrated_channels/integrated_channel/course_metadata.py
```diff
@@ -0,0 +1,127 @@
...
+    course_run_details = course_run_details.copy()
+    for key, base_course_value in course_details.items():
+        if course_run_details.get(key) is None and key not in EXCLUDED_COURSE_DETAIL_KEYS:
+            course_run_details[key] = base_course_value
+    return course_run_details
+
+
+def get_course_list(user, enterprise_customer):
+    """
+    Determine which catalog belongs to a given Enterprise Customer and list the course runs it contains.
```
mattdrayer: Context-wise it doesn't matter to the caller if we are using a catalog to obtain the enterprise customer's list of course runs.  I'd drop that bit from the docstring.  Also you might consider naming this function `get_course_run_list` or `get_course_runs`.  If knowing that the catalog _is_ important to the caller, then we might name the function `get_catalog_course_runs`.

integrated_channels/integrated_channel/course_metadata.py
```diff
@@ -0,0 +1,127 @@
...
+        user: A Django user requesting the course list
+        enterprise_customer: The given Enterprise Customer
+
+    Returns:
+        iterable: An iterable containing the details of each course run.
+    """
+    catalog_id = enterprise_customer.catalog
+
+    if catalog_id is None:
+        LOGGER.info('No catalog attached to %s', enterprise_customer)
```
mattdrayer: Is it okay to fail silently here and return None?

integrated_channels/integrated_channel/course_metadata.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __init__(self, user, plugin_configuration):
+        """
+        Save the appropriate details for use elsewhere in the object.
+        """
+        self.user = user
+        self.enterprise_customer = plugin_configuration.enterprise_customer
+        self.plugin_configuration = plugin_configuration
+
+    def get_parsed(self):
```
mattdrayer: `get_parsed"...course_runs?  What are we actually doing when we say we are "parsing"?

integrated_channels/integrated_channel/course_metadata.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        Retrieve the course runs for the linked catalog, and parse them appropriately.
+        """
+        results = (self.process_course_run(run) for run in get_course_list(self.user, self.enterprise_customer))
+        return results
+
+    def process_course_run(self, course_run_details):
```
mattdrayer: I think I'd prefer something along the lines of `transform_course_run_details`

integrated_channels/integrated_channel/course_metadata.py
```diff
@@ -0,0 +1,127 @@
...
+            'Parsing course for %s: %s',
+            self.enterprise_customer,
+            json.dumps(course_run_details, indent=4),
+        )
+        output = {}
+        for key, transform in self.data_transform.items():
+            output[key] = transform(course_run_details) if transform is not None else course_run_details.get(key)
+        return output
+
+    def send_all(self):
```
mattdrayer: `send_all` what?  courses?  Is this more like `transmit_course_metadata`?

integrated_channels/integrated_channel/course_metadata.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        Wrapper for sending all course metadata blocks.
+        """
+        for course in self.get_parsed():
+            self.send(course)
+
+    def send(self, course):
```
mattdrayer: As with the other function names it would be great if we could be more specific about what is going on -- eg `transmit_course_metadata`

integrated_channels/sap_success_factors/utils.py
```diff
@@ -0,0 +1,86 @@
...
+def parse_datetime_to_epoch(datestamp):
+    """
+    Convert an ISO-8601 datetime string to a Unix epoch timestamp in milliseconds.
+    """
+    parsed_datetime = parse_lms_api_datetime(datestamp)
+    time_since_epoch = parsed_datetime - UNIX_EPOCH
+    return int(time_since_epoch.total_seconds() * 1000)
+
+
+def in_interval(start, end):
```
mattdrayer: How about `is_in_interval`

test_utils/fake_catalog_api.py
```diff
@@ -235,6 +235,261 @@
...
+        },
+    ]
+}
+
+FAKE_CATALOG_COURSE_DETAILS_RESPONSES = {
+    'edX+DemoX': {
+        "key": "edX+DemoX",
+        "uuid": "cf8f5cce-1370-46aa-8162-31fdff55dc7e",
+        "title": "Fancy Course",
+        "course_runs": [
```
mattdrayer: Just a note that `track_selection_url` will be added to the course_runs dicts at some point

integrated_channels/integrated_channel/course_metadata.py
brittneyexline: I agree - especially to have some consistency with the learner data version. Is this where you would call the transmitter class I am creating?

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
```diff
@@ -0,0 +1,133 @@
...
+        # If a valid enterprise customer uuid was provided, transmit for that customer.
+        # Otherwise, transmit for all customers.
+        enterprise_customer_uuid = options['enterprise_customer']
+        if enterprise_customer_uuid:
+            try:
+                enterprise_customer = EnterpriseCustomer.active_customers.get(uuid=enterprise_customer_uuid)
+            except EnterpriseCustomer.DoesNotExist:
+                raise CommandError('Enterprise customer {} not found, or not active'.format(enterprise_customer_uuid))
+            else:
+                enterprise_customers.append(enterprise_customer)
```
brittneyexline: Is there a reason this can't be done in the try block? I'm not sure I understand what the if condition is for this else...

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
```diff
@@ -0,0 +1,133 @@
...
+        else:
+            enterprise_customer = None
```
brittneyexline: If we only depend on enterprise_customers below, do we need this variable?

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
```diff
@@ -0,0 +1,133 @@
...
+def send_data_task(username, enterprise_customer_uuid):
+    """
+    Task to send course data to each linked integrated channel
+
+    Args:
+        user: The user whose access token to use for course discovery API access
+        enterprise_customer: The EnterpriseCustomer whose integrated channels we need to send.
+    """
+    user = User.objects.get(username=username)
+    enterprise_customer = EnterpriseCustomer.objects.get(uuid=enterprise_customer_uuid)
```
brittneyexline: Could you pass the model in from the caller since you've already got the customers there, to avoid reading the object again?

integrated_channels/sap_success_factors/utils.py
```diff
@@ -0,0 +1,86 @@
...
+            },
+        ],
+        'thumbnailURI': lambda x: (x['image']['src'] or ''),
+        'content': lambda x: [
+            {
+                'providerID': SAPSuccessFactorsGlobalConfiguration.current().provider_id,
+                'launchURL': x['marketing_url'] or '',
+                'contentTitle': 'Course Description',
+                'contentID': x['key'],
+                'launchType': 2,
```
brittneyexline: We will need to change this to 3

integrated_channels/sap_success_factors/utils.py
```diff
@@ -0,0 +1,86 @@
...
+            {
+                'locale': 'English',
+                'value': x['full_description'] or '',
+            },
+        ],
+        'thumbnailURI': lambda x: (x['image']['src'] or ''),
+        'content': lambda x: [
+            {
+                'providerID': SAPSuccessFactorsGlobalConfiguration.current().provider_id,
+                'launchURL': x['marketing_url'] or '',
```
brittneyexline: I think @mattdrayer noted in some Jira ticket that we'll likely be sending the track selection url instead, but I don't think you need to change this just yet, it seems like it could be easily updated later. We also might have some other updates to this transform once we get some feedback from Griff.

integrated_channels/integrated_channel/course_metadata.py
pomegranited: I think it's best to filter out any `enterprise_customer` objects with no `catalog` way back in the management command, and just raise here.

Or better yet, take a `catalog_id` parameter, instead of `enterprise_customer`.

integrated_channels/integrated_channel/course_metadata.py
pomegranited: Or `get_complete_course_run_details` ?  Or `merge_course_run_details`?

integrated_channels/sap_success_factors/models.py
```diff
@@ -113,6 +166,45 @@ def __repr__(self):
...
+            # TODO: determine these values from the enrollment seat?
+            price=empty_value,
+            currency=empty_value,
+            creditHours=empty_value,
+            # These fields currently have no edX source, so remain empty.
+            totalHours=empty_value,
+            contactHours=empty_value,
+            cpeHours=empty_value,
+            instructorName=empty_value,
+            comments=empty_value,
```
mattdrayer: @brittneyexline we should confirm with @griffresch on the empty fields we are sending here -- for example we may want to include some additional information in the 'comments' section.

pomegranited:
Also note: `Enterprise SAP SuccessFactors` is available in the LMS Admin since https://github.com/edx/edx-platform/pull/14563 merged (yay!) e.g.,

http://localhost:8000/admin/sap_success_factors/sapsuccessfactorsenterprisecustomerconfiguration/

enterprise/course_catalog_api.py
haikuginger: Will change.

enterprise/course_catalog_api.py
haikuginger: We can add some protection, but it shouldn't be necessary - the upstream catalog utils will give us an empty list if the catalog ID isn't valid.

enterprise/course_catalog_api.py
haikuginger: Agreed.

enterprise/course_catalog_api.py
haikuginger: @mattdrayer: No. As above, the edX API utils will give us an empty result if the request fails.

@pomegranited: https://github.com/edx/edx-platform/blob/master/openedx/core/lib/edx_api_utils.py#L17-L18

integrated_channels/integrated_channel/course_metadata.py
haikuginger: I like `get_complete_course_run_details`; will go with that.

integrated_channels/integrated_channel/course_metadata.py
haikuginger: Good call.

integrated_channels/integrated_channel/course_metadata.py
haikuginger: @mattdrayer, @brittneyexline: Since this is a generator, not having an explicit return value means that the caller gets an empty iterable back. I could make it more explicit by raising `StopIteration` if you'd prefer.

@pomegranited, that's a good call; it's a database restriction that we can make pretty easily.

integrated_channels/integrated_channel/course_metadata.py
haikuginger: I like it!

integrated_channels/integrated_channel/course_metadata.py
haikuginger: Inline with below, `get_transformed_course_runs`?

integrated_channels/integrated_channel/course_metadata.py
haikuginger: `transmit_all_course_details`?

integrated_channels/integrated_channel/course_metadata.py
haikuginger: As above, will rename to be more specific.

@brittneyexline, this isn't exactly where you'll tap in - rather, you'll do that in the SAP-specific subclass of this class. But within this method, yes.

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
```diff
@@ -0,0 +1,133 @@
...
+from enterprise.models import EnterpriseCustomer
+from integrated_channels.sap_success_factors.models import SAPSuccessFactorsEnterpriseCustomerConfiguration
+from integrated_channels.sap_success_factors.utils import SapCourseExporter
+
+try:
+    from djcelery.celery import task as celery_task
+except ImportError:
+    def celery_task(func):
+        """Use a no-op decorator if djcelery is not available."""
+        return func
```
haikuginger: Nice, thanks @pomegranited!

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
haikuginger: For Python, you can do a `try-except-else` block; the try will catch errors raised inside that block, and the else will allow processing if no errors are raised. It lets us be specific about the particular lines of code we want to protect.

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
haikuginger: In fact we don't. Good catch; residue of an earlier version.

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
```diff
@@ -0,0 +1,133 @@
...
+            except EnterpriseCustomer.DoesNotExist:
+                raise CommandError('Enterprise customer {} not found, or not active'.format(enterprise_customer_uuid))
+            else:
+                enterprise_customers.append(enterprise_customer)
+        else:
+            enterprise_customer = None
+
+        # Find the correct user with which to access the catalog API
+        manual_catalog_username = options['catalog_user']
+        settings_catalog_username = getattr(settings, 'EDXAPP_ENTERPRISE_CATALOG_API_USERNAME', None)
```
haikuginger: This is a new setting, but it looks like (based on conversations with EdX people) we're going to want to switch to a configuration model.

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
```diff
@@ -0,0 +1,133 @@
...
+                send_data_task.delay(user.username, customer.uuid)
+            except AttributeError:
+                # Celery isn't available; run the task inline
+                send_data_task(user.username, customer.uuid)
+
+
+    @staticmethod
+    def validate_user(username):
+        if not username:
+            raise CommandError('A username must be specified manually or in settings to access the course catalog API.')
```
haikuginger: Can do!

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
haikuginger: Celery requires that we be able to serialize the arguments we transmit to the task; strings can be JSON-serialized, but models cannot.

integrated_channels/sap_success_factors/utils.py
haikuginger: Good to know! I tried to find the appropriate value in the doc, but couldn't find any details.

integrated_channels/sap_success_factors/utils.py
haikuginger: Sounds good. The whole point of this structure is to make it easy to write/change transforms for individual fields.

integrated_channels/sap_success_factors/utils.py
haikuginger: I'll do you one better and make it `current_time_is_in_interval`!

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
haikuginger: After some talk, we're removing this altogether; the command will just require a username passed as an argument.

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
haikuginger: So, with the above (removing the ability to set a global worker username), we can just flag this arg as required. Problem solved!

enterprise/course_catalog_api.py
pomegranited: Lol whoops..  I was on the wrong fork!  Sorry about that @haikuginger .

haikuginger:
Comments addressed; ready for another pass @pomegranited, @mattdrayer @brittneyexline.

integrated_channels/sap_success_factors/utils.py
```diff
@@ -0,0 +1,96 @@
...
+    data_transform = {
+        'courseID': lambda x: x['key'],
+        'providerID': lambda x: apps.get_model(
+            'sap_success_factors',
+            'SAPSuccessFactorsGlobalConfiguration'
+        ).current().provider_id,
+        'status': lambda x: 'ACTIVE' if x['availability'] == 'Current' else 'INACTIVE',
+        'title': lambda x: [
+            {
+                'locale': 'English',
```
pomegranited: Is it ok to hard-code this locale?  I would have thought we'd have to [pull the locale out of Django somehow](http://stackoverflow.com/a/28048237/4302112) and map it to SAP's locale string.

Though it's probably this is out of scope for this first pass..  Maybe just make sure there's a follow-up ticket somewhere?

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+        Add required arguments to the parser.
+        """
+        parser.add_argument(
+            '--catalog_user',
+            dest='catalog_user',
+            required=True,
+            metavar='ENTERPRISE_CATALOG_API_USERNAME',
+            help='Use this user to access the Course Catalog API. Omit this option to use the'
+            ' username specified in the EDXAPP_ENTERPRISE_CATALOG_API_USERNAME setting.'
```
pomegranited: nit: please update this comment to remove the "Omit this option.." part, since it's now required.

integrated_channels/sap_success_factors/utils.py
```diff
@@ -0,0 +1,96 @@
...
+            },
+        ],
+        'thumbnailURI': lambda x: (x['image']['src'] or ''),
+        'content': lambda x: [
+            {
+                'providerID': apps.get_model(
+                    'sap_success_factors',
+                    'SAPSuccessFactorsGlobalConfiguration'
+                ).current().provider_id,
+                'launchURL': x['marketing_url'] or '',
```
pomegranited: Since @mattdrayer [said there's going to be a track selection URL eventually](https://github.com/edx/edx-enterprise/pull/63/files#r105011522), is it worth having here for when it does happen?  e.g.

```
 'launchURL': x['track_selection_url'] or x['marketing_url'] or '',
```

integrated_channels/sap_success_factors/utils.py
```diff
@@ -0,0 +1,96 @@
...
+            },
+        ],
+        'description': lambda x: [
+            {
+                'locale': 'English',
+                'value': x['full_description'] or '',
+            },
+        ],
+        'thumbnailURI': lambda x: (x['image']['src'] or ''),
+        'content': lambda x: [
```
pomegranited: My understanding from the SAP docs was that this was supposed to be a list of sections in the course, with details about each section.

However, since this data isn't available from the Course Catalog, it's probably out of scope for this task.  It is available from the [Course Blocks API](http://edx.readthedocs.io/projects/edx-platform-api/en/latest/courses/blocks.html#example-response-showing-a-list-of-course-blocks-in-a-specified-course) though, if we need to have a follow-up task?

integrated_channels/sap_success_factors/utils.py
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+
+def parse_datetime_to_epoch(datestamp):
+    """
+    Convert an ISO-8601 datetime string to a Unix epoch timestamp in milliseconds.
+    """
+    parsed_datetime = parse_lms_api_datetime(datestamp)
+    time_since_epoch = parsed_datetime - UNIX_EPOCH
+    return int(time_since_epoch.total_seconds() * 1000)
```
pomegranited: I thought the Unix epoch timestamps are usually expressed in seconds, not milliseconds?

If not, then [my PR needs fixing too](https://github.com/edx/edx-enterprise/pull/59/files#diff-b2d1ad7443d2344126a60d0213e2d192R150), which means we need a common method for this.

tests/test_management.py
```diff
@@ -0,0 +1,324 @@
...
+        get_course_details=get_course_details,
+        get_catalog_courses=get_catalog_courses,
+    )
+
+    caplog.set_level(logging.INFO)
+
+    UserFactory(username='C-3PO')
+    EnterpriseCustomerFactory(
+        catalog=1,
+        name='Veridian Dynamics',
```
pomegranited: Serious geek creds there 😄

integrated_channels/sap_success_factors/utils.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+
+    def __init__(self, user, plugin_configuration):
+        super(SapCourseExporter, self).__init__(user, plugin_configuration)
+
+    def get_serialized_data(self):
+        final_structure = {
+            'ocnCourses': self.courses
+        }
+        return json.dumps(final_structure, sort_keys=True).encode('utf-8')
```
brittneyexline: I don't know... I will ask SAP. In terms of where that will live, I am leaning towards in the client. But this can stay for now and I can adjust when I stitch together these scripts with the transmitters/client.

integrated_channels/sap_success_factors/utils.py
brittneyexline: I think it makes sense for us to only send one content item representing the course as a whole from a logistical standpoint. It probably is worth exploring how sending multiple content items will behave and render in SAP's software and then decide if we want to take advantage of it.

integrated_channels/sap_success_factors/utils.py
brittneyexline: I am pretty sure it should be seconds.

haikuginger:
Rebased on top of master; switched to epoch seconds from milliseconds.

@brittneyexline @mattdrayer, let me know when you're good to go, and I'll go through the merge checklist.

mattdrayer:
MMMMMMILLLLLLLLLISSEEECCCONDDSSSS!!!!

haikuginger:
@brittneyexline, this is ready for merge on your final thumbs-up.

integrated_channels/sap_success_factors/utils.py
pomegranited: @brittneyexline Any update on this?  Should it be seconds or milliseconds?

I can fix this as part of https://github.com/edx/edx-enterprise/pull/69

haikuginger: @pomegranited, final verdict was milliseconds.

pomegranited: Sweet, thanks!

PR #64, opened 2017-03-08
ENT-250 Introduce catalog api endpoints in enterprise api.yaml

asadiqbal08:
**Description:**  Introduce catalogs API endpoints in enterprise api.yaml file for amazon swagger 

@mattdrayer @saleem-latif This is a WIP PR to get feedback. It is fully dependent on [PR#62](https://github.com/edx/edx-enterprise/pull/62/)

**JIRA:** [ENT-250](https://openedx.atlassian.net/browse/ENT-250)

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

### DIFF
--- a/api-compact.yaml
+++ b/api-compact.yaml
@@ -1,0 +1,133 @@
+# Enterprise IDA: upstream API definitions, including vendor extensions.
+# Note, while this document is not strictly to Swagger spec, each HTTP method"s
+#  definition _must_ be to spec or the downstream ref will fail.
+
+apigateway_responses: &apigateway_responses
+  default:
+    statusCode: "400"
+  200:
+    statusCode: "200"
+  401:
+    statusCode: "401"
+  403:
+    statusCode: "403"
+  404:
+    statusCode: "404"
+  429:
+      statusCode: "429"
+  500:
+      statusCode: "500"
+
+produces: &produces
+  - "application/json"
+  - "application/csv"
+
+responses: &responses
+  200:
+    description: "OK"
+  400:
+    description: "Bad Request"
+  401:
+    description: "Unauthorized"
+  403:
+    description: "Forbidden"
+  404:
+    description: "Not Found"
+  429:
+    description: "Too Many Requests"
+  500:
+    description: "Internal Server Error"
+
+id_parameter: &id_parameter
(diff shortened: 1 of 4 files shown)


asadiqbal08:
We need to create the corresponding swagger mapping(s) in [api-manager](https://github.com/edx/api-manager/blob/master/swagger/api.yaml)

mattdrayer:
This is a good start -- I am assuming we will also need to include an `api-compact.yaml` as well.  What other work do you feel is necessary for this PR before we can add it to the project?

asadiqbal08:
@mattdrayer I am curious to know the purpose of api-compact.yaml file. I can see [API-Manager](https://github.com/edx/api-manager/blob/master/swagger/api.yaml#L33) make a reference to api.yaml. 
I think, We are ok with that change sets at that point.

mattdrayer:
👍

api-compact.yaml
```diff
@@ -0,0 +1,125 @@
...
+            uri: "https://${stageVariables.enterprise_host}/v1/catalogs/{id}/"
+
+    # /v1/catalogs/{id}/courses
+    catalogCourses:
+        get:
+          produces: *produces
+          parameters:
+            - *auth_header
+            - *id_parameter
+            - *limit_qs_parameter
```
saleem-latif: for catalog corses, we are using paginated response as opposed to `limit-offset` response. So, we need to update also allow `page` parameter whose value will be int. I have not tested it properly although I am sure  `limit-offset` style will also work. Any way we still need to add `page` query string parameter.

api.yaml
```diff
@@ -0,0 +1,201 @@
...
+            name: Authorization
+            required: true
+            type: string
+          -
+            in: path
+            name: id
+            required: true
+            type: number
+          -
+            in: querystring
```
saleem-latif: same as above, we can leave `limit-offset` parameters as is too, but we need `page` qs parameter here.

asadiqbal08:
@mattdrayer Should I merge it ?

mattdrayer:
Yep, two thumbs, ready to go!

PR #65, opened 2017-03-08
ENT-218/ENT-221 Implementing clients for success factors endpoints

brittneyexline:
**Description:** Implemented the client functions needed to interact with the success factors api.
Also began implementation of the transmitters that will be called from the processes Jill and Jesse are building.

**JIRA:** Link to JIRA ticket

**Dependencies:** https://github.com/edx/edx-enterprise/pull/59/files
https://github.com/edx/edx-enterprise/pull/63/files

instructions.

**Testing instructions:**

See https://docs.google.com/document/d/13vol2OQT_vVzD164k7fh7TH2_xsGtj1rWb4xWC3ozAg/edit

**Merge checklist:**

- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


integrated_channels/sap_success_factors/client.py
```diff
@@ -31,28 +31,91 @@ def get_oauth_access_token(url_base, client_id, client_secret, company_id, user_
...
+        if not oauth_access_token:
+            raise ValueError('An API access token must be supplied!')
+
+        self.url_base = url_base
+
+        session = requests.Session()
+        session.timeout = timeout
+        session.headers['Authorization'] = 'Bearer {}'.format(oauth_access_token)
+        self.session = session
+        super(SAPSuccessFactorsAPIClient, self).__init__()
```
pomegranited: If you're going to create one client and use that many times over a potentially long period of time, you'll probably want some mechanism for refreshing the oauth token when it expires.

e.g. https://github.com/edx/edx-analytics-pipeline/blob/master/edx/analytics/tasks/util/edx_api_client.py#L94-L97

integrated_channels/sap_success_factors/client.py
```diff
@@ -31,28 +31,91 @@ def get_oauth_access_token(url_base, client_id, client_secret, company_id, user_
...
+            payload (str): JSON object containing course import data per SuccessFactors documentation.
+
+        Returns:
+            The body of the response from SAP SuccessFactors, if successful
+        Raises:
+            HTTPError: if we received a failure response code from SAP SuccessFactors
+        """
+        global_sap_config = SAPSuccessFactorsGlobalConfiguration.current()
+        url = self.url_base + global_sap_config.course_api_path
+        response = self.session.post(url, json=payload)
```
pomegranited: @brittneyexline :
@haikuginger and I are now thinking that `transmit` should take a payload object, not the already-prepared JSON string.

```python
    response = self.session.post(url, json=payload.serialize())
```

Ok with you?

brittneyexline: yes - i was moving towards this direction :-)
i'll be updating this PR momentarily

pomegranited: Lovely!

integrated_channels/sap_success_factors/client.py
```diff
@@ -4,20 +4,22 @@
...
-class SAPSuccessFactorsAPIClient(slumber.API):
+class SAPSuccessFactorsAPIClient(object):
     """
     Client for connecting to SAP SuccessFactors.
 
     Specifically, this class supports obtaining access tokens and posting to the courses and
      completion status endpoints.
     """
 
+    SESSION_TIMEOUT = 5
```
mattdrayer: Should this value be moved to settings?

integrated_channels/sap_success_factors/client.py
```diff
@@ -31,28 +33,119 @@ def get_oauth_access_token(url_base, client_id, client_secret, company_id, user_
...
         )
 
+        response.raise_for_status()
         data = response.json()
-
         try:
-            return data['access_token'], datetime.datetime.utcfromtimestamp(data['expiresIn'])
-        except KeyError:
+            return data['access_token'], datetime.datetime.utcfromtimestamp(data['expires_in'] + int(time.time()))
+        except Exception:
```
mattdrayer: Seems like we could be more specific about the Exception we want to catch here.

integrated_channels/sap_success_factors/client.py
```diff
@@ -31,28 +33,119 @@ def get_oauth_access_token(url_base, client_id, client_secret, company_id, user_
...
+        Args:
+            payload: JSON object (serialized from LearnerDataTransmissionAudit)
+                containing completion status fields per SuccessFactors documentation.
+
+        Returns:
+            The body of the response from SAP SuccessFactors, if successful
+        Raises:
+            HTTPError: if we received a failure response code from SAP SuccessFactors
+        """
+        global_sap_config = SAPSuccessFactorsGlobalConfiguration.current()
```
mattdrayer: It appears we are using this in both of the "public" functions for the class -- does it make sense to DRY it by relocating to `__init__() as a `self` field so we don't have to remember to include in new functions going forward?

integrated_channels/sap_success_factors/client.py
haikuginger: I agree.

integrated_channels/sap_success_factors/client.py
haikuginger: ^ This makes sense to me.

integrated_channels/sap_success_factors/client.py
```diff
@@ -31,28 +33,119 @@ def get_oauth_access_token(url_base, client_id, client_secret, company_id, user_
...
+        Args:
+            url (str): The url to post to.
+            payload: The payload to post.
+        """
+        now = datetime.datetime.utcnow()
+        if now >= self.expires_at:
+            # Create a new session with a valid token
+            self.session.close()
+            self._create_session()
+        response = self.session.post(url, json=payload.serialize())
```
haikuginger: If I remember correctly, then the `json` argument takes a Python dict and serializes it to a JSON stream. You want to be using `data`, since the payload is already serialized.

integrated_channels/sap_success_factors/client.py
```diff
@@ -31,28 +33,119 @@ def get_oauth_access_token(url_base, client_id, client_secret, company_id, user_
...
+        return response.status_code, response.text
```
haikuginger: I would personally just return the full response object.

integrated_channels/sap_success_factors/transmitters/courses.py
```diff
@@ -3,28 +3,47 @@
...
-    Transform course data into the format expected by SuccessFactors.
-
-    Send the transformed course data to SuccessFactors using the configured endpoint
-    (taken from OCNWebServicesConfiguration and OCNWebServicesEnterpriseCustomerConfiguration) and the oauth token.
-
-    Record information about the success/failure of posting the course data to SuccessFactors in
-     CourseContentExportAudit.
     """
+
+    def __init__(self, enterprise_configuration):
```
haikuginger: Note that if you're not doing anything with the data passed into `__init__`, you can just not implement it altogether and let it fall through to the base class.

integrated_channels/sap_success_factors/transmitters/courses.py
```diff
@@ -3,28 +3,47 @@
...
+        Args:
+            payload (list): The OCN course import data payload to send to SAP SuccessFactors
+        """
+        try:
+            code, body = self.client.send_course_import(payload)
+        except RequestException as request_exception:
+            code = 500
+            body = request_exception.message
+
+        error_message = body if code >= 400 else None
```
haikuginger: Do we want `else None` or `else ''`?

integrated_channels/sap_success_factors/transmitters/learner_data.py
```diff
@@ -3,30 +3,51 @@
...
+            return None
+
+        try:
+            code, body = self.client.send_completion_status(payload.payload())
+        except RequestException as request_exception:
+            code = 500
+            body = request_exception.message
+
+        payload.status = code
+        payload.error_message = body if code >= 400 else None
```
haikuginger: As above - do we want `None` or `''`?

integrated_channels/sap_success_factors/client.py
mattdrayer: Also please note the units in a comment

integrated_channels/sap_success_factors/transmitters/courses.py
```diff
@@ -14,28 +16,37 @@
...
         """
-        Get serialized data from the courseware exporter and transmit it.
+        Send a course data import call to SAP SuccessFactors using the client.
+
+        Args:
+            payload (SapCourseExporter): The OCN course exporter object to send to SAP SuccessFactors
         """
-        LOGGER.info(courseware_exporter.get_serialized_data())
+        serialized_payload = payload.get_serialized_data()
+        LOGGER.info(serialized_payload)
```
mattdrayer: After this lands in production, do we still want it to be treated as INFO, or should it be changed to DEBUG?

integrated_channels/sap_success_factors/transmitters/learner_data.py
```diff
@@ -3,42 +3,53 @@
...
+        Send a completion status call to SAP SuccessFactors using the client.
+
+        Args:
+            payload (LearnerDataTransmissionAudit): The learner completion data payload to send to SAP SuccessFactors
         """
-        # TODO ENT-221 will handle transmitting the learner data payload to the channel
-        # We only want to transmit if learner_data.course_completed == True.
-        LOGGER.info(learner_data.serialize())
+        serialized_payload = payload.serialize()
+        LOGGER.info(serialized_payload)
```
mattdrayer: Same question as above/earlier -- INFO or DEBUG after this is in the wild?

mattdrayer:
I think I'm basically down to nits at this point -- no big changes required from me -- 👍

PR #66, opened 2017-03-09
ENT-213 Add optional contact_email field to EnterpriseCustomer

bdero:
**Description:** 
This adds a simple optional contact_email field to `EnterpriseCustomer`, which is in turn displayed to the user in messages to the user. See https://github.com/edx/ecommerce/pull/1189.

**JIRA:** Prerequisite for [ENT-213](https://openedx.atlassian.net/browse/ENT-213).

**Testing instructions:**

1. Install this version of `edx-enterprise` into a devstack.
1. Go to the [Enterprise Customer](http://localhost:8000/admin/enterprise/enterprisecustomer/add/) admin panel.
1. Verifying that the new optional "Contact email" is configurable in the Enterprise Customer form:
    

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

**Reviewers**
- [ ] @pomegranited 
- [ ] edX reviewer[s] TBD

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -118,6 +118,13 @@ class EnterpriseCustomer(TimeStampedModel):
         )
     )
 
+    contact_email = models.EmailField(
+        blank=True,
+        help_text=_(
+            "Optional contact email that is displayed to learners."
+        )
+    )
+
     @property
     def identity_provider(self):
         """
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -63,7 +63,7 @@ class EnterpriseCustomerSerializer(serializers.ModelSerializer):
         model = models.EnterpriseCustomer
         fields = (
             'uuid', 'name', 'catalog', 'active', 'site', 'enable_data_sharing_consent', 'enforce_data_sharing_consent',
-            'enterprise_customer_users', 'branding_configuration', 'enterprise_customer_entitlements'
+            'enterprise_customer_users', 'branding_configuration', 'enterprise_customer_entitlements', 'contact_email'
         )
 
     site = SiteSerializer()
--- a/enterprise/migrations/0016_auto_20170308_1738.py
+++ b/enterprise/migrations/0016_auto_20170308_1738.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('enterprise', '0015_auto_20170130_0003'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='enterprisecustomer',
+            name='contact_email',
+            field=models.EmailField(help_text='Optional contact email that is displayed to learners.', max_length=254, blank=True),
+        ),
+        migrations.AddField(
+            model_name='historicalenterprisecustomer',
+            name='contact_email',
+            field=models.EmailField(help_text='Optional contact email that is displayed to learners.', max_length=254, blank=True),
+        ),
+    ]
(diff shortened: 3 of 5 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=66"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

enterprise/models.py
```diff
@@ -118,6 +118,13 @@ class Meta:
         )
     )
 
+    contact_email = models.EmailField(
```
mattdrayer: hmm...in general I don't really like the idea of adding fields like this directly to the "aggregate root" of a domain...it seems to me we'd be better off adding an EnterpriseCustomerContactInfo model which could contain this email field and an FK back to the EnterpriseCustomer...

mattdrayer: It's sort of like the email field in the Django auth_user table -- awful place to store that type of info, imnsho.

pomegranited: @mattdrayer Are you anticipating we'll need to store more customer contact information ?  If not, then I think a separate model is overkill.

But if so, then maybe this should be `learner_contact_email` instead, to differentiate it from our internal contact email addresses, which we might use for support or otherwise.

mattdrayer: I can see more contact info being needed in the future, yes -- for example the phone number of the HR/Learning office, and the name of the training manager.  But maybe this could also be a `contact_user_id` FK field that points to an actual learner record in the system -- I think I'd be alright with that.

pomegranited: @mattdrayer Seems like a reasonable compromise to me.

@bdero this would require another change to add a `EnterpriseContactUserSerializer` class, pulled from the user that matches this ID.

What do you think?

bdero: @pomegranited I'm having trouble figuring out where `EnterpriseContactUser` came from? It seems like @mattdrayer was talking about potentially adding a `contact_user_id` FK to `EnterpriseCustomerContactInfo` sometime in the future with his last comment. Let me know if I'm getting it wrong, though. :)

I can understand just creating an `EnterpriseCustomerContactInfo` model with a 1-1 relationship with `EnterpriseCustomer`, and giving it only one other field in this PR for now: `learner_contact_email`. Would this make sense @mattdrayer @pomegranited?

It seems like for this particular purpose, we need a single optional email field that represents a general point of contact for the `EnterpriseCustomer` that learners can use - based on the requirements of the message in ENT-216:

pomegranited: @bdero Ah ok, yep, I misunderstood.  Sounds like the `EnterpriseCustomerContactInfo` 1-1 relationship with `EnterpriseCustomer`, with just the single `contact_email` field is fine at this point.

bdero: @mattdrayer Would you mind if I called this contact model work a separate ticket? I probably don't have enough time to work it through this week since I budgeted my time based on the ENT-213 description, and I think we can get the Ecommerce PR for ENT-213 merged this week since it doesn't actually have a hard dependency on this change. :)

mattdrayer: I'm fine with that, yep -- sorry for the slow reply

mattdrayer: Also what I was thinking above is that we *could* add a `contact_enterprise_customer_user_id` FK field to the `EnterpriseCustomer` record.  This FK would point to an `EnterpriseCustomerUser` record, and we could determine the actual contact info based on the subsequent relationship-hop to the actual `User` and `UserProfile` records.  Hopefully that's not too crazy...

mattdrayer:
Closing this PR per the discussion on the related JIRA ticket.

PR #69, opened 2017-03-14
ENT-259 SAP Learner Completion for self-paced courses

pomegranited:
**Description:** Allows self-paced courses to determine course completion using the course end date and/or a passing grade for the learner.

**JIRA:** [ENT-259](https://openedx.atlassian.net/browse/ENT-259)

**Dependencies:** 
* https://github.com/edx/edx-platform/pull/14710
* https://github.com/edx/edx-platform/pull/14753

**Merge deadline:** 28 March 2017

**Sandbox URL:**

https://ent-259-pomegranited.sandbox.edx.org

Note: Staff user has superuser privileges on the sandbox, and the test setup instructions have been run.

**Test setup instructions:** 

1. In the Django Admin, [Add a Self Paced Configuration](https://ent-259-pomegranited.sandbox.edx.org/admin/self_paced/selfpacedconfiguration/),
   and set the `Enabled` and `Enable course home page improvements` flags 
   ([ref](http://edx.readthedocs.io/projects/open-edx-building-and-running-a-course/en/latest/set_up_course/setting_pacing.html#set-pacing-for-your-course)).
1. In Studio, create the following courses:
    * Instructor-paced course, titled e.g. [Instructor-Paced Course](https://studio-ent-259-pomegranited.sandbox.edx.org/course/course-v1:OpenCraft+InstrPace+2017).
    * Self-paced with no end date, titled e.g. [Self-Paced Ongoing](https://studio-ent-259-pomegranited.sandbox.edx.org/course/course-v1:OpenCraft+SelfPace+ongoing).
    * Self-paced with an course end date in the future titled e.g. [Self-Paced w End Date](https://studio-ent-259-pomegranited.sandbox.edx.org/course/course-v1:OpenCraft+SelfPace+Mar2017)

    For each of the above courses
    * Under `Settings > Schedule & Details`, set Course Pacing to "Self paced" or "Instructor paced" as appropriate, and Save Changes.
    * Under `Settings > Schedule & Details`, set the Start Date to sometime in the past.
    * Under `Settings > Grading`, delete all but one of the Assignment Types, and give the remaining one a weight of 100.
    * Under `Content > Outline`, add a Graded subsection, and some problems.
    * Publish the course content.
1. In the Django Admin, [add a Course Mode](https://ent-259-pomegranited.sandbox.edx.org/admin/course_modes/coursemode/) for each course above. `Audit` mode is fine for the self-paced courses, but the instructor-paced must use something else.
1. Set up TestShib SAML Integration ([ref](http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/configuration/tpa/tpa_SAML_SP.html))
    * Ensure `FEATURES['ENABLE_THIRD_PARTY_AUTH'] == true`
    * Run `openssl req -new -x509 -days 3652 -nodes -out saml.crt -keyout saml.key` to generate `saml.crt` (public key),
      and `saml.key` (private key)
    * In Django Admin, [Add SAML Configuration](https://ent-259-pomegranited.sandbox.edx.org/admin/third_party_auth/samlconfiguration/)
      using the above keys.  Ensure it's Enabled, and has a unique Entity ID.
    * Download the [SAML metadat.xml](https://ent-259-pomegranited.sandbox.edx.org/auth/saml/metadata.xml), and save to
      a unique file name.
    * Upload the metadata.xml to [testshib.org/register](https://www.testshib.org/register.html).
      This will eventually create a SAML backend called `tpa-saml`.
    * [Add a Provider Configuration (SAML IdPs)](https://ent-259-pomegranited.sandbox.edx.org/admin/third_party_auth/samlproviderconfig/).

      ```
      Enabled: Yes
      Visible: Yes
      Backend name: tpa-saml
      Idp Slug: testshib
      Entity ID: https://idp.testshib.org/idp/shibboleth
      Metadata source: https://www.testshib.org/metadata/testshib-providers.xml
      ```
      After submitting, the Metadata Available flag should go to Green (might have to refresh the page to see it.)
1. In the Django Admin, [Add an EnterpriseCustomer](https://ent-259-pomegranited.sandbox.edx.org/admin/enterprise/enterprisecustomer/) with:

   ```
   Enable data sharing consent: Yes
   Enforce data sharing consent: At Enrollment or At Login
   Provider ID: testshib
   ```
1. In the Django Admin, [Add an SAP SuccessFactors Enterprise Customer Configuration](https://ent-259-pomegranited.sandbox.edx.org/admin/sap_success_factors/sapsuccessfactorsenterprisecustomerconfiguration/) for the new Enterprise Customer.  Most of the fields don't matter, since we're not actually sending data to SAP yet, but ensure:

    ```
    Active: Yes
    Enterprise Customer: linked to EC created above.
    Remaining fields can be anything.
    ```
1. In Studio, navigate to the Instructor-paced course.  Under Settings > Certificates page, create a certificate. Don't worry about adding signatories.  Activate the certificate.

**Testing instructions:**

1. On the [Manage Learners](https://ent-259-pomegranited.sandbox.edx.org/admin/enterprise/enterprisecustomer/) page for your EnterpriseCustomer , link a new or existing learner to each of the courses you created above.
    *Note* Currently, you can't simply enrol a single user into multiple enterprise courses without linking them to a program. (Shows error "User with email {email} is already registered with Enterprise Customer {customer}").  You have enroll 3 separate users into each course.  So for the sandbox server, I've enrolled:
    * `audit` in [Self-Paced w End Date](https://ent-259-pomegranited.sandbox.edx.org/courses/course-v1:OpenCraft+SelfPace+Mar2017)
    * `honor` in [Self-Paced Ongoing](https://ent-259-pomegranited.sandbox.edx.org/courses/course-v1:OpenCraft+SelfPace+ongoing)
    * `verified` in [Instructor Paced Course](https://ent-259-pomegranited.sandbox.edx.org/courses/course-v1:OpenCraft+InstrPace+2017)
1. Login to the LMS as the Enterprise user(s) linked above.  Visit each enrolled course, and grant data sharing consent.

1. Shell into the server (ask @pomegranited or @mattdrayer  to add your credentials), and sudo to the `edxapp` user:
   ```
   ssh pomegranited@ent-259-pomegranited.sandbox.edx.org
   sudo -u edxapp -Hs
   cd ~/edx-platform
   source ../edxapp_env
   ```
1. Run the learner completion management command.  Note the output.
   There should be an entry for each learner+course enrollment, with `course_completed: False`, and `grade: In Progress` for all courses.
   ```
   ./manage.py lms transmit_learner_data --api_user=staff --settings=aws
   ```
   The `--api_user` setting is new, and must be set to a username with staff access to the courses above.
   Also, because testshib is not a real SAML provider, you'll also see an error for each user, e.g. "remote_id not found for third party provider=saml-testshib, username=audit".
1. Demonstrate that self-paced courses with end dates can result in a Failing or Passing grade once the end date is past:
    1. Update the End Date of the [Self-Paced w End Date](https://ent-259-pomegranited.sandbox.edx.org/courses/course-v1:OpenCraft+SelfPace+Mar2017) to sometime in the past, and re-run the `transmit_learner_data` command.
    1. Note the `transmit_learner_data` record for that enrollment should now have `course_completed: True`, and `grade: Fail`.
    1. Login as the learner for [Self-Paced w End Date](https://ent-259-pomegranited.sandbox.edx.org/courses/course-v1:OpenCraft+SelfPace+Mar2017), and complete the problem(s) correctly to earn a passing grade.
    1. Note the `transmit_learner_data` record for that enrollment should now have `course_completed: True`, and `grade: Pass`.
1. Demonstrate that self-paced courses with no end date show In Progress or Passing grades:
    1. Login as the learner for [Self-Paced Ongoing](https://ent-259-pomegranited.sandbox.edx.org/courses/course-v1:OpenCraft+SelfPace+ongoing)
    1. Complete the problem(s) incorrectly, to generate a non-passing grade.
    1. Note the `transmit_learner_data` record for that enrollment should now have `grade: In Progress`.
    1. Complete the problem(s) correctly to generate a passing grade.
    1. Note the `transmit_learner_data` record for that enrollment should now have `course_completed: True`, and `grade: Pass`.

1. Demonstrate that instructor-paced courses show In Progress until certificates are generated.
    1. Run this command to generate the certificates for your instructor-paced course, e.g.

    ```
   ./manage.py lms ungenerated_certs --course="course-v1:OpenCraft+InstrPace+2017" --settings=aws
   ```
   1. Note the `transmit_learner_data` record for that enrollment should now have `course_completed: True`.  Grade sent is currently a percentage.  See below for discussion on sending Pass/Fail instead.

**Reviewers:**

- [x] @e-kolpakov 
- [x] @mattdrayer 

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped - to `0.29.0`
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:**
- What `grade` values should be sent to SAP?  This PR sends `Pass/Fail/In Progress` for self-paced courses, but https://github.com/edx/edx-enterprise/pull/59 sends the `GeneratedCertificate.grade` value.


edx-webhook:
Thanks for the pull request, @pomegranited! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=69&repo=edx%2Fedx-enterprise"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

docs/configuration_and_usage.rst
```diff
@@ -99,6 +99,77 @@ to the help strings provided by the form.
...
+
+   If data sharing consent is *not* enabled for your ``EnterpriseCustomer``, then learner data may be sent to SAP
+   SuccessFactors without their explicit consent, so use these settings with care.
+
+
+Transmit Learner Data
+_____________________
+
+The ``transmit_learner_data`` command sends course completion data for each ``EnterpriseCustomerCourseEnrollment`` where
+evidence of course completion exists for an
```
mattdrayer: Looks like we're missing a piece of this sentence?

docs/configuration_and_usage.rst
```diff
@@ -99,6 +99,77 @@ to the help strings provided by the form.
...
+These Django management commands send data to ``EnterpriseCustomer``'s active, integrated channels.
+
+They must be run from the command line on a server with the Open edX environment installed.
+
+.. note::
+
+   If data sharing consent is enabled for your ``EnterpriseCustomer``, then any Enterprise learners enrolled in the
+   associated courses must consent to data sharing, thus permitting their data to be sent back to SAP SuccessFactors.
+
+   If data sharing consent is *not* enabled for your ``EnterpriseCustomer``, then learner data may be sent to SAP
```
mattdrayer: This makes me think that we should have some sort of confirmation prompt to provide a small guard against accidentally transmitting data in the case of consent not being enabled.

pomegranited: @mattdrayer If this is a concern, I think we should change the behaviour of the data transmitters to send data only if data sharing consent is required.

docs/configuration_and_usage.rst
pomegranited: Thanks for catching that, @mattdrayer .  This change came from https://github.com/edx/edx-enterprise/pull/59 , and I've fixed it [here](https://github.com/edx/edx-enterprise/pull/59/commits/b8b28e4e6e9e0181c43b0868d8d845dcf9c021aa#diff-39096f6801e1aeb926636974b0ec9f1eR146).

test_settings.py
```diff
@@ -103,7 +103,7 @@ def root(*args):
         'user': '50/minute',
         'service_user': '60/minute',
     },
-
+    'DATETIME_FORMAT': '%Y-%m-%dT%H:%M:%SZ',
```
pomegranited: Our dates are now all tz-aware, UTC dates, so it's ok to explicitly specify the terminating `Z`.  Note that the default format is `iso-8601`, which renders the timezone offset as `+00:00`.  The [terminating `Z` is also accepted under `iso-8601`](https://en.wikipedia.org/wiki/ISO_8601#UTC), but python's `isoformat()` function prefers the explicit numeric tz offset.

This will also guard against the changes in Django REST framework 3.4, which [uses millisecond precision by default](http://www.django-rest-framework.org/topics/3.4-announcement/#microsecond-precision-when-returning-time-or-datetime).

tests/api/test_views.py
```diff
@@ -66,7 +68,7 @@ def create_items(self, factory, items):
                     'email': 'test1@example.com',
                     'is_staff': True,
                     'is_active': False,
-                    'date_joined': (now - datetime.timedelta(days=10)).isoformat(),
+                    'date_joined': (now - datetime.timedelta(days=10)).strftime(LMS_API_DATETIME_FORMAT),
```
pomegranited: [See above](https://github.com/edx/edx-enterprise/pull/69/commits/835e06a373651c57c2fc0165f4e3287ac20efaae#r106059492) for why this change and similar below were required.

docs/configuration_and_usage.rst
mattdrayer: This sounds like a good idea to me, however today there were some discussions regarding exactly when we need to gather consent and when it can be assumed that consent has been granted due to the employee-employer relationship.  I believe we ended up agreeing that consent should be gathered on the edX side regardless of the status of E2E (@griffresch please confirm here or on ENT-259).  

So, I *think* I'm okay with limiting this process to transmit only those learners who have explicitly granted consent, although I think the overall on/off flag for this feature should also be one of the gates.  That is, if consent is not enabled, we should not be sending any data at all.  If we allow data to be shared with consent disabled, that seems like an end-around to the whole process.  I'll defer to @griffresch for the final say on this, however.

pomegranited: @mattdrayer CC @griffresch  I agree that requiring explicit learner consent is surely safer, so we should exclude from this transmitted data any EnterpriseCustomers that don't require consent, and remove this warning note.  It does make me wonder about the use case for EnterpriseCustomers who don't require data sharing consent -- but I don't know the full Enterprise project scope.

In the interests of keeping this PR on task with ENT-259, can we create a separate ticket for fixing this issue?  Note that https://github.com/edx/edx-enterprise/pull/63 has the same concern, but since it shares code with https://github.com/edx/edx-enterprise/pull/59 (merged), we should be able to fix it in [one place](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/management/commands/__init__.py#L92-L101).  CC @haikuginger

integrated_channels/integrated_channel/models.py
```diff
@@ -60,5 +78,133 @@ class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
...
+        """
+        Collect learner data for the ``EnterpriseCustomer`` where data sharing consent is granted.
+
+        Yields a learner data object for each enrollment, containing:
+
+        * ``enterprise_enrollment``: ``EnterpriseCourseEnrollment`` object.
+        * ``certificate``: ``GeneratedCertificate`` for the user+course; None if not found.
+          "Course completion" occurs when course certificates are issued.
+        """
+
```
mattdrayer: At this point have we already made the determination that the EnterpriseCustomer has enabled data sharing consent?  If not then this might be a good spot to do that check.

docs/configuration_and_usage.rst
mattdrayer: I'm okay tracking this as a separate ticket, yep.

integrated_channels/integrated_channel/models.py
pomegranited: I'm not sure what is simplest.. the consent logic is getting pretty tangled.

We could certainly check here.  We do say pretty clearly on [`EnterpriseCustomer.enable_data_sharing_consent`](https://github.com/edx/edx-enterprise/blob/864030005bb8c84b2b996399fbaca4c9adc02561/enterprise/models.py#L101-L107):
> This field is used to determine whether data sharing consent is enabled or disabled for users signing in using this enterprise customer. If disabled, consent will not be requested, and eligible data will not be shared.

And/or filter out the `EnterpriseCustomer`s without data sharing enabled [in the management command filter](https://github.com/edx/edx-enterprise/pull/69/files/835e06a373651c57c2fc0165f4e3287ac20efaae#diff-11c9715fd5316220a126561ca420c9dcR100).

And/or, we could [change this check below](https://github.com/edx/edx-enterprise/pull/69/files/835e06a373651c57c2fc0165f4e3287ac20efaae#diff-4386ccd909983c900b04b5a992e66fddR134) to something more specific for each enrollment.  I started off checking `enrollment.consent_granted`, but then changed it to `enrollment.consent_available()`, which took the [`UserDataSharingConsentAudit`](https://github.com/edx/edx-enterprise/blob/864030005bb8c84b2b996399fbaca4c9adc02561/enterprise/models.py#L574) records into account as well.

mattdrayer: The earlier we can filter out ineligible possibilities, the better -- if we can constrain the initial set of enterprise customers that all of this work is driven from based on whether or not consent is enabled, I think that would be a good thing.

CHANGELOG.rst
```diff
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.27.2] - 2017-03-10
```
e-kolpakov: @pomegranited Should be 0.28 - the schema is major.minor.patch where

* major - for backward-incompatible changes
* minor - for new backward-compatible features 
* patch - for bugfixes.

Checkout semver schema (in the comment above) - makes a lot of sense when you read it.

CHANGELOG.rst
```diff
@@ -25,7 +30,6 @@ Unreleased
 
 * Added API endpoint for fetching catalogs and catalog courses.
 
-
```
e-kolpakov: @pomegranited not a big deal, but please restore this.

docs/configuration_and_usage.rst
```diff
@@ -99,6 +99,77 @@ to the help strings provided by the form.
...
+Usage
+~~~~~
+
+.. code-block:: bash
+
+   # View command help
+   $ ./manage.py lms transmit_learner_data --help --settings=$EDX_PLATFORM_SETTINGS
+
+   # Transmit learner data for all EnterpriseCustomers, to all active integrated channels.
+   $ ./manage.py lms transmit_learner_data --settings=$EDX_PLATFORM_SETTINGS
```
e-kolpakov: @pomegranited suggestion: might be good to clarify what $EDX_PLATFORM_SETTINGS mean here. We all kind of know that these are either aws or devstack or whatnot, but for unsuspecting developer it might be a puzzle.

enterprise/__init__.py
```diff
@@ -4,6 +4,6 @@
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.27.1"
+__version__ = "0.27.2"
```
e-kolpakov: @pomegranited 0.28.0

enterprise/lms_api.py
```diff
@@ -24,24 +26,36 @@
     CourseEnrollment = None
 
 
+LOGGER = logging.getLogger(__name__)
```
e-kolpakov: @pomegranited here's the thing - on one hand, pylint demands capitalizing this as it thinks it is a constant; on the other hand it's actually not a constant, but an object. 

I believe other files in this repo [suppressed the error in this case](https://github.com/edx/edx-enterprise/blob/4d78623a3897d2eb8614bb30fc994ccda9d26ae0/enterprise/signals.py#L12), so would be quite nice if you could do the same.

enterprise/lms_api.py
```diff
@@ -24,24 +26,36 @@
...
 LMS_API_DATETIME_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
 
 
-class EnrollmentApiClient(object):
+class LmsApiClient(object):
```
e-kolpakov: @pomegranited suggestion: make it abstract?

I have relatively negative experienec with `abc`, but it might be just the right use case for it.

enterprise/lms_api.py
```diff
@@ -114,7 +119,79 @@ def get_course_details(self, course_id):
...
+        """
+        Retrieve the grade for the given username for the given course_id.
+
+        Args:
+        * ``course_id`` (str): The string value of the course's unique identifier
+        * ``username`` (str): The username ID identifying the user for which to retrieve the grade.
+
+        Returns: None if no grade found, or a dict containing:
+
+        * ``username``: A string representation of a user's username passed in the request.
```
e-kolpakov: @pomegranited do I understand right it affects documentation formatting? If that's so - "oooh, nice, can you do that everywhere (in a separate PR preferable, to keep this one focused)"; if not - "please make these args normal, without apostrophes" :)

enterprise/lms_api.py
```diff
@@ -114,7 +119,79 @@ def get_course_details(self, course_id):
...
+        * ``course_key``: A string representation of a Course ID.
+        * ``passed``: Boolean representing whether the course has been passed according the course's grading policy.
+        * ``percent``: A float representing the overall grade for the course
+        * ``letter_grade``: A letter grade as defined in grading_policy (e.g. 'A' 'B' 'C' for 6.002x) or None
+
+        """
+        try:
+            results = self.client.course_grade(course_id).users().get(username=username)
+        except HttpNotFoundError:
```
e-kolpakov: @pomegranited suggestion: this `try .. except: log.error(...) results = []` repeats a couple of times, how about extracting it into method? Might be not worth it, so please use common sense :)

integrated_channels/integrated_channel/management/commands/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+
+from enterprise.models import EnterpriseCustomer
+from integrated_channels.sap_success_factors.models import SAPSuccessFactorsEnterpriseCustomerConfiguration
+
+
+# Import djcelery, or stub it if not available.
+try:
+    from djcelery.celery import task as celery_task
+except ImportError:
+    def celery_task(func):
```
e-kolpakov: @pomegranited I wonder, should we make this `integrated channels` thing a plugin app with it's own dependencies. I mean, we will probably do that at some point, but maybe this point is now.

No need to address this, just thinking out aloud.

integrated_channels/integrated_channel/management/commands/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+            '--channel',
+            dest='channel',
+            default='',
+            metavar=_('INTEGRATED_CHANNEL'),
+            help=_('Transmit data to this IntegrateChannel. '
+                   'Omit this option to transmit to all configured, active integrated channels.'),
+            choices=INTEGRATED_CHANNEL_CHOICES.keys(),
+        )
+
+    def get_integrated_channels(self, options):
```
e-kolpakov: @pomegranited nit: this method is 40 lines lon - probably a good time to split it. How about extracting the for loop at the end to a dedicated method?

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
```diff
@@ -0,0 +1,61 @@
...
+class Command(IntegratedChannelCommandMixin, BaseCommand):
+    """
+    Management command which transmits learner course completion data to the IntegratedChannel(s) configured for the
+    given EnterpriseCustomer.
+
+    Collect the enterprise learner data for enrollments with data sharing consent, and transmit each to the
+    EnterpriseCustomer's configured IntegratedChannel(s).
+    """
+    help = _('''
+    Transmit Enterprise learner course completion data for the given EnterpriseCustomer.
```
e-kolpakov: @pomegranited nit: feels like a broken identation

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
```diff
@@ -0,0 +1,61 @@
...
+        super(Command, self).add_arguments(parser)
+
+    def handle(self, *args, **options):
+        """
+        Transmit the learner data for the EnterpriseCustomer(s) to the active integration channels.
+        """
+        # Ensure that we were given an api_user name, and that User exists.
+        api_user = options['api_user']
+        if not api_user:
+            raise CommandError('Error: argument --api_user is required')
```
e-kolpakov: @pomegranited I might be terribly wrong (too much languages and frameworks in my head), but aren't there a config option that make sthe parser fail when argument value is missing or empty?

integrated_channels/integrated_channel/models.py
```diff
@@ -60,5 +78,141 @@ class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
...
+                continue
+
+            course_id = enterprise_enrollment.course_id
+            completed_date = None
+            grade = None
+
+            # Fetch the user+course certificate, if found
+            try:
+                # If a certificate is found, then we use that to determine course completion.
+                certificate = GeneratedCertificate.eligible_certificates.get(
```
e-kolpakov: @pomegranited please make it an API - we should abstract all LMS dependencies to be able to make `edx-enterprise` a dedicated app at some time in future.

integrated_channels/integrated_channel/models.py
```diff
@@ -60,5 +78,141 @@ class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
...
+                    user__id=enterprise_enrollment.enterprise_customer_user.user_id,
+                    course_id=CourseKey.from_string(course_id),
+                )
+                completed_date = certificate.created_date
+                # TODO: change to Pass/Fail/In Progress?
+                grade = certificate.grade
+
+            except GeneratedCertificate.DoesNotExist:
```
e-kolpakov: @pomegranited 40 lines long _exception handling_ - probably needs a method (and probably more than one)

integrated_channels/integrated_channel/models.py
```diff
@@ -60,5 +78,141 @@ class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
...
+
+                # If this is a self-paced course, then check the Grades API
+                if course_details is None or course_details['course_id'] != course_id:
+                    course_details = course_api.get_course_details(course_id)
+
+                if course_details is not None and course_details.get('pacing') == 'self':
+                    user = enterprise_enrollment.enterprise_customer_user.user
+                    grades_data = grades_api.get_course_grade(course_id, user.username)
+
+                    # This should not happen, but if it does, we have nothing to report about this enrollment.
```
e-kolpakov: @pomegranited I'm not quite sure of the flow, but can't it be mistakenly invoked when the course is not completed? I'd suggest a `warn` level, as it's not an error in the code or data - just a bad timing.

But I might just misunderstand when this method is hit - could you please explain this a little bit?

integrated_channels/integrated_channel/models.py
```diff
@@ -60,5 +78,141 @@ class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
...
+                enterprise_enrollment=enterprise_enrollment,
+                completed_date=completed_date,
+                grade=grade,
+            )
+
+    def get_learner_data(self, enterprise_enrollment, completed_date=None, grade=None):
+        """
+        Returns the class that can serialize the learner course completion data to the integrated channel.
+        """
+        raise NotImplementedError(_('Implemented in concrete subclass.'))
```
e-kolpakov: @pomegranited two things here:

1. This should never be shown to the user, so no need to mark it for localization
2. There should be a constant for "Must be implemented in descendant class" you've likely meant here.

integrated_channels/sap_success_factors/models.py
```diff
@@ -93,9 +123,35 @@ class LearnerDataTransmissionAudit(models.Model):
     error_message = models.TextField(blank=True)
     created = models.DateTimeField(auto_now_add=True)
 
+    epoch = datetime(1970, 1, 1, tzinfo=timezone.utc)
```
e-kolpakov: @pomegranited nit: this is a constant, so might make sense to capitalize it.

test_settings.py
```diff
@@ -88,6 +88,8 @@ def root(*args):
 
 LMS_ROOT_URL = "http://localhost:8000"
 
+OAUTH_ID_TOKEN_EXPIRATION = 60*60
```
e-kolpakov: @pomegranited nit: might make sense to include measurement unit into the var name - to make it clear it's in seconds.

tests/test_management.py
```diff
@@ -0,0 +1,240 @@
...
+
+    # Self-paced course with future end date sends grade=Pass, or grade=In Progress, depending on current grade.
+    (dict(), None, True, FUTURE, False, dict(completed='false', timestamp='null', grade='"In Progress"')),
+    (dict(), None, True, FUTURE, True, dict(completed='true', timestamp=NOW_TIMESTAMP, grade='"Pass"')),
+
+    # Self-paced course with past end date sends grade=Pass, or grade=Fail, depending on current grade.
+    (dict(), None, True, PAST, False, dict(completed='true', timestamp=PAST_TIMESTAMP, grade='"Fail"')),
+    (dict(), None, True, PAST, True, dict(completed='true', timestamp=PAST_TIMESTAMP, grade='"Pass"')),
+])
+def test_transmit_learner_data(caplog, command_args, certificate, self_paced, end_date, passed, expected_completion):
```
e-kolpakov: @pomegranited I'd suggest to either split this test into smaller chunks, or at least extract setup methods (`responses.add(...)`) - it's like 100 lines long, so it is hard to comprehend it.

tests/test_models.py
```diff
@@ -857,6 +882,321 @@ def test_string_conversion(self, method):
...
+            course_id=self.course_id,
+            consent_granted=True,
+        )
+
+        # Return a mock certificate
+        mock_course_key.from_string.return_value = None
+        certificate = mock.MagicMock(
+            user=self.user,
+            course_id=self.course_id,
+            grade="A-",
```
e-kolpakov: @pomegranited note the line number - totally time to split this module into smaller onces (aren't pylint already complaining about this?)

pomegranited:
@mattdrayer @brittneyexline CC @haikuginger @e-kolpakov 

I've run into some difficulty with this task.. Currently, the [Grades API will only return the logged-in user's grade for a course](https://github.com/edx/edx-platform/blob/b2cd0359e5d11e0c0dae50c69c0f477ab7a40932/lms/djangoapps/grades/api/views.py#L134), and doesn't have any special allowances for staff users.

So I'm going to have to submit a PR to update the Grades API to allow staff members for a course to grab any of their learners' grades.  @nasthagiri Is this something you guys would accept?

nasthagiri:
@pomegranited Yes, for sure.  The current implementation of the Grades API was implemented by MIT as a stop-gap solution for their immediate requirements.  Enhancing the API to support staff access to other learners' grades would be very useful.  Please note that https://openedx.slack.com/team/kabirkhan requested a similar change in the openedx.slack.com general-channel on March 9th and may be working on a PR to support this.  Can you check with him to see if he has already started on this work?

pomegranited:
Oo, thanks @nasthagiri !  I've asked [Kabir on Slack](https://openedx.slack.com/archives/C02SNA1U4/p1489709653101129).

pomegranited:
Rebased from 694c6c6 to f0247c7 .

CHANGELOG.rst
pomegranited: Thanks, @e-kolpakov - I'll change the major version when I do that bump.

CHANGELOG.rst
pomegranited: I think that might have been an artefact of a rebase, because it's now not showing CHANGELOG.rst in my diff.

But I'll update this file when it comes time to merge.

enterprise/__init__.py
pomegranited: @e-kolpakov Yep, ref [comment above](https://github.com/edx/edx-enterprise/pull/69#discussion_r106581914)

docs/configuration_and_usage.rst
pomegranited: @e-kolpakov  Thanks for the suggestion.. I've gone into a bit more detail with e56fbf4, but don't know if that's clear enough?

enterprise/lms_api.py
pomegranited: @e-kolpakov [Other files in the repo also use the `LOGGER` name](https://github.com/edx/edx-enterprise/search?utf8=%E2%9C%93&q=logger), so I believe this is up for debate.

@brittneyexline @haikuginger Would you like to weigh in?  I'm not fussed either way, but would like to be consistent.

enterprise/lms_api.py
pomegranited: @e-kolpakov `LmsApiClient` actually functions on its own, since the Slumber API is so flexible, so I'm inclined to leave it as concrete.

enterprise/lms_api.py
pomegranited: @e-kolpakov It does repeat, but each time, the message is subtly different, and/or the return value.  So I'm going to leave it as-is.

integrated_channels/integrated_channel/management/commands/__init__.py
pomegranited: @e-kolpakov I would much rather use the real libraries, and not have to mock all this stuff!

@brittneyexline @mattdrayer @haikuginger Wanna weigh in?

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
pomegranited: pylint wouldn't let me do it any other way 😦

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
pomegranited: @e-kolpakov There totally is:  `required=True`.  But I couldn't get it to work on my local devstack!  And I'd worn a dent in the wall trying to figure out why, so I just let it go.

But I'll try again once the rest of the issues are resolved.

integrated_channels/integrated_channel/models.py
pomegranited: @e-kolpakov We need to update the Certificates API to return the created date, cf [comment on ENT-193](https://openedx.atlassian.net/browse/ENT-193?focusedCommentId=249118&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-249118).

integrated_channels/integrated_channel/models.py
pomegranited: @e-kolpakov The `transmit_learner_data` management command gets run by Jenkins on some regular schedule, for all enterprise customers and their courses.

So yes, there will be some courses that aren't completed yet.  However, there should still be a grade record for that user, even if it's incomplete.

integrated_channels/integrated_channel/models.py
pomegranited: @e-kolpakov You're totally right, this is getting out of hand.  I would rather refactor this into a separate Exporter class, like what @haikuginger did with https://github.com/edx/edx-enterprise/pull/63.

integrated_channels/sap_success_factors/models.py
pomegranited: Again, @haikuginger 's https://github.com/edx/edx-enterprise/pull/63 helps us here.. then I can use his [`EPOCH`](https://github.com/edx/edx-enterprise/pull/63/files#diff-3727947e8e9274e6b3ed166f76f07d7cR15) and [`parse_datetime_to_epoch`](https://github.com/edx/edx-enterprise/pull/63/files#diff-3727947e8e9274e6b3ed166f76f07d7cR81)

test_settings.py
pomegranited: @e-kolpakov This setting [comes from edx-platform](https://github.com/edx/edx-platform/search?utf8=%E2%9C%93&q=OAUTH_ID_TOKEN_EXPIRATION), so I have to keep the same name.

tests/test_management.py
pomegranited: Will do.

integrated_channels/integrated_channel/models.py
pomegranited: Refactored with 7357bba.

integrated_channels/sap_success_factors/models.py
pomegranited: Handled by https://github.com/edx/edx-enterprise/commit/7357bba85cbd12d507de2dfa278bec8e29ad4a73#diff-3727947e8e9274e6b3ed166f76f07d7cR102

tests/test_models.py
pomegranited: Refactor moved a lot of this to test_utilities.  Also less mocking was required after the refactor (yay!).

https://github.com/edx/edx-enterprise/commit/7357bba85cbd12d507de2dfa278bec8e29ad4a73#diff-b12dd73705f83d2c7a954a988f607d6fR979

integrated_channels/integrated_channel/models.py
pomegranited: Addressed with 3717446.

integrated_channels/integrated_channel/management/commands/__init__.py
pomegranited: Addressed with a0e279e.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
pomegranited: Addressed by 5ac3fbb.

enterprise/lms_api.py
pomegranited: @e-kolpakov Yep, the double ticks do affect documentation formatting. 

E.g., [enterprise/utils.py L121](https://github.com/edx/edx-enterprise/blob/master/enterprise/utils.py#L121) results in a nicely-formmated `get_all_field_names` string here in the [generated docs](http://open-edx-enterprise-service-documentation.readthedocs.io/en/latest/enterprise.html#enterprise.utils.get_all_field_names).

I'm using them throughout this PR, but would rather clean up the remaining code as part of a separate task.

pomegranited:
@e-kolpakov I believe I've addressed everything in your review, with these issues as outstanding questions for the ENT team:
        
* To [LOGGER or logger](https://github.com/edx/edx-enterprise/pull/69#discussion_r106414166)?
* [Should `integrated_channels` have its own dependencies](https://github.com/edx/edx-enterprise/pull/69#discussion_r106416099)?

The [quality tests are failing](https://travis-ci.org/edx/edx-enterprise/builds/212891950) due to @&#%#%* isort import ordering issues in tests_management.py, which I'm hoping to sort out soon.

I've deployed edx-enterprise 32d8417 to the sandbox, and finished updating the test instructions.

Could you please have another look when you get a chance?

e-kolpakov:
@pomegranited hint: `isort` can be run without `--check-only` to actually sort imports in a way that would satisfy the checker.

pomegranited:
@e-kolpakov Thanks for the tip!  I tried that, but since isort on my devstack throws errors about files that Travis seems ok with, it's not much use.  I don't know why my devstack differs from Travis; they're even running the same version of isort!

tests/test_lms_api.py
```diff
@@ -176,11 +193,114 @@ def test_get_remote_id():
...
+
+    client.something()
+    assert not client.token_expired(), "Token was refreshed"
+    assert client.something_called == 3, "Wrapped method was called"
+
+
+@responses.activate
+@mock.patch('enterprise.lms_api.JwtBuilder', mock.Mock())
+def test_get_course_grades_not_found():
+    username = "DarthVadar"
```
e-kolpakov: @pomegranited It was Vader, wasn't it? Or is it to avoid copyright infringement? :)

tests/test_utilities.py
```diff
@@ -968,3 +974,273 @@ def test_base_course_exporter_serialized_data_raises(self, mock_get_course_runs)
...
+        assert learner_data[1].course_id == self.course_id
+        assert not learner_data[1].course_completed
+        assert learner_data[1].completed_timestamp is None
+        assert learner_data[1].grade == 'In Progress'
+
+        assert learner_data[2].enterprise_course_enrollment_id == enrollment2.id
+        assert learner_data[2].course_id == course_id2
+        assert learner_data[2].course_completed
+        assert learner_data[2].completed_timestamp == self.NOW_TIMESTAMP
+        assert learner_data[2].grade == 'Pass'
```
e-kolpakov: @pomegranited Aren't pylint complaining about too long file? Probably it's a good point to split it up anyway.

pomegranited:
@e-kolpakov Sorry!  I've added your github public key to my login account, since I think edxapp doesn't have a shell.

```
ssh pomegranited@ent-259-pomegranited.sandbox.edx.org
```

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -0,0 +1,215 @@
...
+        Returns the string used for a failing grade.
+        """
+        return 'Fail'
+
+    @property
+    def grade_incomplete(self):
+        """
+        Returns the string used for an incomplete course grade.
+        """
+        return 'In Progress'
```
pomegranited: @e-kolpakov I figured there's a chance that we'd want to send different grade values to different integrated channels, so made them overridable properties instead.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
```diff
@@ -20,19 +21,40 @@ class Command(IntegratedChannelCommandMixin, BaseCommand):
...
 
+    def add_arguments(self, parser):
+        """
+        Add required --api_user argument to the parser.
+        """
+        parser.add_argument(
+            '--api_user',
+            dest='api_user',
+            required=True,
+            metavar=_('LMS_API_USERNAME'),
```
pomegranited: @e-kolpakov It's displayed to users of the management command, and isn't integral to the code, so I figured it was fair game for translating.

https://docs.python.org/3/library/argparse.html#metavar

tests/test_lms_api.py
pomegranited: @e-kolpakov Nope, just an honest misspelling..  removed with 4edb20e.

integrated_channels/integrated_channel/management/commands/__init__.py
```diff
@@ -58,48 +58,65 @@ def add_arguments(self, parser):
...
+        See ``add_arguments`` for the accepted options.
+
+        filter_kwargs is passed as an additional set of parameters that jobs can use to restrict
+        the database query used to retrieve relevant integrated channels.
+        """
+        enterprise_customer = self._get_enterprise_customer(options['enterprise_customer'])
+        if enterprise_customer:
+            filter_kwargs['enterprise_customer'] = enterprise_customer
+
+        channel_classes = self._get_channel_classes(options['channel'])
```
pomegranited: Fixed with 4edb20e.

tests/test_utilities.py
pomegranited: @e-kolpakov Addressed with 3622fae.

enterprise/lms_api.py
```diff
@@ -158,6 +221,48 @@ def get_remote_id(self, identity_provider, username):
...
+        * ``username``: A string representation of a user's username passed in the request.
+        * ``course_key``: A string representation of a Course ID.
+        * ``passed``: Boolean representing whether the course has been passed according the course's grading policy.
+        * ``percent``: A float representing the overall grade for the course
+        * ``letter_grade``: A letter grade as defined in grading_policy (e.g. 'A' 'B' 'C' for 6.002x) or None
+
+        """
+        results = self.client.course_grade(course_id).users().get(username=username)
+        for row in results:
+            if row.get('username') == username:
```
brittneyexline: is there a reason we have to check the username again if we're specifically querying by username?

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
```diff
@@ -20,19 +21,40 @@ class Command(IntegratedChannelCommandMixin, BaseCommand):
...
+            help=_('Username of a user authorized to fetch grades from the LMS API.'),
+        )
+        super(Command, self).add_arguments(parser)
+
     def handle(self, *args, **options):
         """
         Transmit the learner data for the EnterpriseCustomer(s) to the active integration channels.
         """
+        # Ensure that we were given an api_user name, and that User exists.
+        api_username = options['api_user']
```
brittneyexline: just to clarify, we're introducing a user argument so we can hit the grades api?

enterprise/lms_api.py
pomegranited: No, I'm just being overly thorough probably..  Since the Grades API returns a list of results, it looks like they might one day return more than one, e.g. 'fetch all grades for this course'.

But currently, the Grades API only returns one set of grades for the requested `username`, and throws a 404 if you're not allowed to see them.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
pomegranited: Yes.  But if we move to using the [Certificates API](https://github.com/edx/edx-platform/blob/master/lms/djangoapps/certificates/apis/v0/views.py#L70) instead of accessing the GeneratedCertificate model directly, we'll need a username for that query too.

pomegranited:
@mattdrayer @brittneyexline This is now ready for your full review.  I've updated the sandbox to run the latest commit.

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -0,0 +1,215 @@
...
+
+        If no certificate is found, then returns the completed_date = None, grade = In Progress, on the idea that a
+        certificate will eventually be generated.
+        """
+
+        if CourseKey is None or GeneratedCertificate is None:
+            raise NotConnectedToOpenEdX(_('This package must be installed in an OpenEdX environment.'))
+
+        try:
+            # TODO - use Certificates API instead?
```
mattdrayer: Have we made a decision on this TODO?

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -0,0 +1,215 @@
...
+            certificate = GeneratedCertificate.eligible_certificates.get(
+                user__id=enterprise_enrollment.enterprise_customer_user.user_id,
+                course_id=CourseKey.from_string(enterprise_enrollment.course_id),
+            )
+
+            completed_date = certificate.created_date
+            # TODO: change to Pass/Fail/In Progress?
```
mattdrayer: What information do we need in order to resolve this TODO?

integrated_channels/integrated_channel/management/commands/__init__.py
```diff
@@ -58,48 +58,65 @@ def add_arguments(self, parser):
...
-        filter_kwargs is passed as an additional set of parameters that jobs can use to restrict
-        the database query used to retrieve relevant integrated channels.
+        Raises CommandError if uuid is invalid.
         """
-        # If a valid enterprise customer uuid was provided, transmit for that customer.
-        # Otherwise, transmit for all customers.
-        enterprise_customer = None
-        enterprise_customer_uuid = options['enterprise_customer']
-        if enterprise_customer_uuid:
+        if uuid is not None:
```
mattdrayer: It looks like you might be able to short-circuit this: 

```
if not uuid:
    return None
```

...then move on to the try/except

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
mattdrayer: I'm not sure "LMS_API_USERNAME" will actually be translated by anyone?

test_settings.py
```diff
@@ -90,6 +90,8 @@ def root(*args):
 
 ENTERPRISE_PUBLIC_ENROLLMENT_API_URL = ENTERPRISE_ENROLLMENT_API_URL
 
+OAUTH_ID_TOKEN_EXPIRATION = 60*60
```
mattdrayer: Please note the units in a comment

integrated_channels/integrated_channel/learner_data.py
pomegranited: @mattdrayer It requires an update to the Certificates API: https://github.com/edx/edx-enterprise/pull/69#discussion_r106584884

integrated_channels/integrated_channel/learner_data.py
pomegranited: @mattdrayer  Just need someone to tell me what the grade should be :)  I've got all the information I need from the certificate, so can show pass/fail instead of the certificate grade.

integrated_channels/integrated_channel/management/commands/__init__.py
pomegranited: Addressed with 7dbdc35.

test_settings.py
pomegranited: Done with f999062.

integrated_channels/integrated_channel/management/commands/transmit_learner_data.py
pomegranited: No worries, 475bc23 removes the gettext wrappers.

pomegranited:
@mattdrayer  I believe I've addressed all of your comments, with one outstanding question:

* [Should grades for certificate courses be sent as "Pass/Fail/In Progress"](https://github.com/edx/edx-enterprise/pull/69#discussion_r107407669) like we're now doing for the audit courses?

integrated_channels/integrated_channel/learner_data.py
mattdrayer: Made a comment on [ENT-193](https://openedx.atlassian.net/browse/ENT-193)

integrated_channels/integrated_channel/learner_data.py
mattdrayer: Made a comment on [ENT-193](https://openedx.atlassian.net/browse/ENT-193)

mattdrayer:
Thanks for responding to my comments -- I've followed the discussion chain back to [ENT-193](https://openedx.atlassian.net/browse/ENT-193) and left a comment there.  Nutshell:  Yes we should add `created_date` to the Certificates API and use that, and I think we need some guidance from Product/UX on the value we are sending to SAP for the grade.

integrated_channels/integrated_channel/learner_data.py
pomegranited: Addressed with 169ca0e.

pomegranited:
@mattdrayer Could I get another review on this PR?  I've updated the sandbox to use the latest code, and have added your Github ssh keys to the `pomegranited` account on the sandbox, just in case you need to shell in and run the commands.

enterprise/lms_api.py
```diff
@@ -24,36 +28,97 @@
...
 
 class LmsApiClient(object):
     """
     Object builds an API client to make calls to the edxapp LMS API.
+
+    Authenticates using settings.EDX_API_KEY.
     """
 
-    def __init__(self, url, append_slash=False):
+    API_BASE_URL = settings.LMS_ROOT_URL + '/api/'
```
mattdrayer: I've been doing some work with the [settings for Enterprise](https://github.com/edx/edx-platform/pull/14762), and seeing all of these URL definitions in the client implementations with references to 'settings.LMS_ROOT_URL` makes me think we should be adding them to `common.py` with corresponding overrides in `aws.py`.

What are your thoughts on this?

enterprise/lms_api.py
```diff
@@ -158,6 +221,87 @@ def get_remote_id(self, identity_provider, username):
...
+        Returns:
+
+        a dict containing:
+
+        * ``username``: A string representation of an user's username passed in the request.
+        * ``course_id``: A string representation of a Course ID.
+        * ``certificate_type``: A string representation of the certificate type.
+        * ``created_date`: Datetime the certificate was created (tz-aware).
+        * ``status``: A string representation of the certificate status.
+        * ``passing``: True if the certificate has a passing status, False if not.
```
mattdrayer: this is now "is_passing", right?

pomegranited: Yep, missed that one.  Fixed with 2c773cf.

enterprise/lms_api.py
pomegranited: I do agree that some standard ways to reference the API URLs would be great, since I'm sure enterprise is not the only service that wants to talk to these endpoints.

And edx-platform is the logical, central location, since it manages all its service URLs.  But I'm not sure that just a straight move from our `API_BASE_URL` values into edx-platform env is the way to go.

I don't know much about the REST API approach and plans for improvements though.  Are there any plans to standardise the REST APIs?  Each individual sub-service seems to provide its own path, with no standard way to expose the path, and the currently supported version number.

So no suggestions offhand..  Would need to do some more digging into how the other services access the APIs, and to see what edX's plans are.

pomegranited:
@mattdrayer I've answered your comments, and fixed that missing `is_passing` - thanks for that!

Once I've got your 👍 , I'll work through the merge checklist and merge.

mattdrayer:
I'm good at this point -- no more major changes so let's scoop up what we have and we can always circle back to things on a follow-on effort.  Thanks!  👍

PR #72, opened 2017-03-21
Remove OAuth2 authentication class from API viewsets

mattdrayer:
**Description:** According to @clintonb, `OAuth2Authentication` is not a DRF authentication class that we use in Open edX, so I've removed it from this project.  I've also reordered the authentication classes such that `JwtAuthentication` is the first one in the list, followed by `BearerAuthentication`, and lastly `SessionAuthentication`.

**JIRA:** ENT-264, ENT-275

**Dependencies:** N/A

**Merge deadline:** Now

**Installation instructions:** No special installation instructions.

**Testing instructions:** Unsure of how to test at the moment.

**Merge checklist:**

- [ ] Regenerate requirements with `make upgrade && make requirements` (and make sure to fix any errors).
  **DO NOT** just add dependencies to `requirements/*.txt` files.
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Documentation updated (not only docstrings)
- [ ] Commits are (reasonably) squashed
- [ ] Translations are updated
- [ ] PR author is listed in AUTHORS

**Post merge:**
- [ ] Create a tag
- [ ] Check new version is pushed to PyPi after tag-triggered build is finished.
- [ ] Delete working branch (if not needed anymore)

**Author concerns:** List any concerns about this PR - inelegant 
solutions, hacks, quick-and-dirty implementations, concerns about 
migrations, etc.

PR #73, opened 2017-03-22
Remove unused OAuth2 requirements

mattdrayer:
**Description:** We removed OAuth2 authentication in PR #72, and because that was the only place we were referencing the packages we can now remove the requirements themselves.  @clintonb FYI.

**JIRA:** ENT-264, ENT-275

**Dependencies:** N/A

**Merge deadline:** 3/23

**Installation instructions:** Normal setup

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


mattdrayer:
Thanks for pointing that out -- I have removed the classifier because the package used to -- but no longer does -- support Django 1.10.  We have a tech debt ticket to restore 1.10 support, which will either happen when LMS is upgraded, or when Enterprise is hosted as a separate service. 

Will merge after checks pass.

clintonb:
Related: https://github.com/edx/edx-enterprise/pull/44

It would be ideal if Django 1.10 support was fixed sooner so that there is one less package for those handling the upgrade to worry about.

PR #77, opened 2017-03-28
ENT-286 Adjustment to grade and courseCompleted values

brittneyexline:
**Description:** We were informed that courseCompleted should be false if the learner received a failing grade. Additionally, I updated the instructor paced grades to be sent as Pass/Fail like the self paced grades were.

**JIRA:** https://openedx.atlassian.net/browse/ENT-286

**Dependencies:** n/a

**Merge deadline:** 3/31/2017

**Installation instructions:** n/a

**Testing instructions:**

I'm not sure if any additional verification is necessary beyond the unit tests.

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:**

### DIFF
--- a/integrated_channels/integrated_channel/learner_data.py
+++ b/integrated_channels/integrated_channel/learner_data.py
@@ -25,26 +25,30 @@ class BaseLearnerExporter(object):
     Base class for exporting learner completion data to integrated channels.
     """
 
+    GRADE_PASSING = 'Pass'
+    GRADE_FAILING = 'Fail'
+    GRADE_INCOMPLETE = 'In Progress'
+
     @property
     def grade_passing(self):
         """
         Returns the string used for a passing grade.
         """
-        return 'Pass'
+        return self.GRADE_PASSING
 
     @property
     def grade_failing(self):
         """
         Returns the string used for a failing grade.
         """
-        return 'Fail'
+        return self.GRADE_FAILING
 
     @property
     def grade_incomplete(self):
         """
         Returns the string used for an incomplete course grade.
         """
-        return 'In Progress'
+        return self.GRADE_INCOMPLETE
 
     def __init__(self, user, plugin_configuration):
         """
@@ -117,16 +121,17 @@ class BaseLearnerExporter(object):
 
             # For instructor-paced courses, let the certificate determine course completion
             if course_details.get('pacing') == 'instructor':
-                completed_date, grade = self._collect_certificate_data(enterprise_enrollment)
+                completed_date, grade, is_passing = self._collect_certificate_data(enterprise_enrollment)
 
             # For self-paced courses, check the Grades API
             else:
-                completed_date, grade = self._collect_grades_data(enterprise_enrollment, course_details)
+                completed_date, grade, is_passing = self._collect_grades_data(enterprise_enrollment, course_details)
 
             yield self.get_learner_data_record(
                 enterprise_enrollment=enterprise_enrollment,
                 completed_date=completed_date,
                 grade=grade,
+                is_passing=is_passing,
             )
 
     def _collect_certificate_data(self, enterprise_enrollment):
@@ -137,6 +142,15 @@ class BaseLearnerExporter(object):
 
         If no certificate is found, then returns the completed_date = None, grade = In Progress, on the idea that a
         certificate will eventually be generated.
+
+        Args:
+            enterprise_enrollment (EnterpriseCourseEnrollment): the enterprise enrollment record for which we need to
+            collect completion/grade data
+
+        Returns:
+            completed_date: Date the course was completed, this is None if course has not been completed.
+            grade: Current grade in the course.
+            is_passing: Boolean indicating if the grade is a passing grade or not.
         """
 
         if self.certificates_api is None:
@@ -152,20 +166,33 @@ class BaseLearnerExporter(object):
                 completed_date = parse_datetime(completed_date)
             else:
                 completed_date = timezone.now()
-            # TODO: use certificate['is_passing'] to send Pass/Fail instead of numeric grade?
-            grade = certificate.get('grade', self.grade_incomplete)
+
+            # For consistency with _collect_grades_data, we only care about Pass/Fail grades. This could change.
+            is_passing = certificate.get('is_passing')
+            grade = self.grade_passing if is_passing else self.grade_failing
 
         except HttpNotFoundError:
             completed_date = None
             grade = self.grade_incomplete
+            is_passing = False
 
-        return completed_date, grade
+        return completed_date, grade, is_passing
 
     def _collect_grades_data(self, enterprise_enrollment, course_details):
         """
         Collect the learner completion data from the Grades API.
 
         Used for self-paced courses.
+
+        Args:
+            enterprise_enrollment (EnterpriseCourseEnrollment): the enterprise enrollment record for which we need to
+            collect completion/grade data
+            course_details (dict): the course details for the course in the enterprise enrollment record.
+
+        Returns:
+            completed_date: Date the course was completed, this is None if course has not been completed.
+            grade: Current grade in the course.
+            is_passing: Boolean indicating if the grade is a passing grade or not.
         """
         if self.grades_api is None:
             self.grades_api = GradesApiClient(self.user)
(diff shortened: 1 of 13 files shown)


integrated_channels/integrated_channel/learner_data.py
```diff
@@ -179,7 +183,7 @@ def _collect_grades_data(self, enterprise_enrollment, course_details):
         except HttpNotFoundError:
             # Grade not found, so we have nothing to report.
             LOGGER.error("No grades data found for %d: %s, %s", enterprise_enrollment.pk, course_id, username)
-            return None, None
+            return None, None, None
```
mattdrayer: Please document what is being returned by this function in the docstring -- while you're there you can add the argument info as well.

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -204,4 +208,4 @@ def _collect_grades_data(self, enterprise_enrollment, course_details):
             completed_date = None
             grade = self.grade_incomplete
 
-        return completed_date, grade
+        return completed_date, grade, passed
```
mattdrayer: is `passed` different from `is_passing`?

tests/test_learner_data.py
```diff
@@ -138,7 +138,7 @@ def test_learner_data_instructor_paced_w_certificate(self, mock_certificate_api,
         assert learner_data[0].course_id == self.course_id
         assert learner_data[0].course_completed
         assert learner_data[0].completed_timestamp == self.NOW_TIMESTAMP
-        assert learner_data[0].grade == 'A-'
+        assert learner_data[0].grade == 'Pass'
```
mattdrayer: Is it possible to replace this string with the enum reference?  That way if "Pass" ever changes to "Passing" we won't have to update the tests.

tests/test_learner_data.py
```diff
@@ -202,13 +202,13 @@ def test_learner_data_self_paced_course(self, passing, end_date, expected_comple
...
         assert learner_data[0].course_id == self.course_id
-        assert learner_data[0].course_completed == (expected_completion is not None)
+        assert learner_data[0].course_completed == (passing and expected_completion is not None)
         assert learner_data[0].completed_timestamp == expected_completion
         assert learner_data[0].grade == expected_grade
 
     @ddt.data(
         ('self', 'Pass'),
-        ('instructor', 'A-'),
+        ('instructor', 'Pass'),
```
mattdrayer: See comment above re: reference vs. string

tests/test_management.py
```diff
@@ -228,7 +228,7 @@ def test_transmit_courseware_task_no_catalog(fake_catalog_client, caplog):
 COMPLETE_CERTIFICATE = dict(
     completed='true',
     timestamp=NOW_TIMESTAMP,
-    grade='A-',
+    grade='Pass',
```
mattdrayer: enum vs. string?

integrated_channels/sap_success_factors/models.py
```diff
@@ -114,7 +114,7 @@ def get_learner_data_record(self, enterprise_enrollment, completed_date=None, gr
         course_completed = False
         if completed_date is not None:
             completed_timestamp = parse_datetime_to_epoch(completed_date)
-            course_completed = True
+            course_completed = True if is_passing else False
```
pomegranited: Could simply be `course_completed = is_passing`

tests/test_management.py
```diff
@@ -441,7 +441,7 @@ def get_expected_output(**expected_completion):
     (dict(), None, True, FUTURE, True, dict(completed='true', timestamp=NOW_TIMESTAMP, grade='Pass')),
 
     # Self-paced course with past end date sends grade=Pass, or grade=Fail, depending on current grade.
-    (dict(), None, True, PAST, False, dict(completed='true', timestamp=PAST_TIMESTAMP, grade='Fail')),
+    (dict(), None, True, PAST, False, dict(completed='false', timestamp=PAST_TIMESTAMP, grade='Fail')),
```
pomegranited: Worth adding an additional test using a certificate with a failing grade.

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -25,26 +25,30 @@ class BaseLearnerExporter(object):
     Base class for exporting learner completion data to integrated channels.
     """
 
+    GRADE_PASSING = 'Pass'
```
haikuginger: Should these values be defined here, or in a subclass specific to SAP SuccessFactors? If we integrated another channel (please, not for a while), then `GRADE_PASSING` might be `127`, `GRADE_FAILING` might be `0`, and `GRADE_INCOMPLETE` might be `-1`. It's only in the case of SuccessFactors that we actually know that we can send off a free-form string, right?

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -25,26 +25,30 @@ class BaseLearnerExporter(object):
...
+    GRADE_FAILING = 'Fail'
+    GRADE_INCOMPLETE = 'In Progress'
+
     @property
     def grade_passing(self):
         """
         Returns the string used for a passing grade.
         """
-        return 'Pass'
+        return self.GRADE_PASSING
```
haikuginger: It's not entirely clear to me what we get out of making this a property referring statically to the value stored on the class, rather than just storing the value on the class (or, as mentioned above, the subclass) with this name in the first place. Can you clarify what the goal is?

In conjunction with the above, I might actually keep these methods and make them raise `NotImplementedError`, thus forcing subclasses to override with the relevant values for that particular `IntegratedChannel`.

integrated_channels/integrated_channel/learner_data.py
brittneyexline: I suppose that because SAP is free form, what we are sending them is effectively our default - which hearkens back to what we imagined we would send in a csv to enterprise customers as v1 of data sharing. I think this will be simple enough to change when the next channel that requires specific values comes along.

integrated_channels/integrated_channel/learner_data.py
brittneyexline: I can't recall @pomegranited 's original intent in introducing these property functions. The reason I didn't change this is as follows: If the grades are really default values, it makes sense for them to be constants owned by the base class that can be referenced elsewhere. These property functions are also performing the default behavior - by returning the constants. They can be overridden by channels that do have constraints on the grades we send back. But I agree that it feels clunky to have both - I think I would rather get rid of these functions and keep the constants if we insist on choosing one.

integrated_channels/integrated_channel/learner_data.py
haikuginger: Okay, fine with me!

integrated_channels/integrated_channel/learner_data.py
haikuginger: Okay, that makes sense! As with above, let's leave that alone for right now, then, and we can revisit as needed. I'm going to go ahead and do some testing now. :)

haikuginger:
:+1:

- [x] I tested this: I configured my devstack accordingly and verified that the new values were being set on the records that were printed to the console.
- [x] I read through the code
- [x] ~~I checked for accessibility issues~~ N/A

Now, as part of testing this, I did run into some issues that need to be fixed to get SuccessFactors working correctly.

The biggest issue is [this line](https://github.com/edx/edx-enterprise/blob/master/enterprise/lms_api.py#L212). It should look like this (username is a query parameter):

```python
returned = self.client.providers(identity_provider).users().get(username=username)
```

On top of that, the endpoint in question in the LMS only supports OAuth authentication; not JWT or Session. I believe that the API key we use fulfills the requirements of that authentication, but we might want to add additional auth classes to that endpoint anyway so we can actually use it in other cases.

Additionally, I don't see a hard fail point where we drop SAP learners who _don't_ have a linked ID; we just stick `None` in the field, and keep passing them along. Rather than having [this clause](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/sap_success_factors/models.py#L121) inline with the function call, we should check it first, raise an exception (something like `IneligibleLearner`?) if it's `None`, and skip records that raise that exception in the extractor.

pomegranited:
@haikuginger  Oo, good catch on the IDP API call!  I didn't have a real endpoint to work with, and accidentally used [these `UserView` docs](https://github.com/edx/edx-platform/blob/6d2468d0dcf782a7561ecc61946146d6fae2762f/common/djangoapps/third_party_auth/api/views.py#L32) instead of the correct [`UserMappingView` docs](https://github.com/edx/edx-platform/blob/6d2468d0dcf782a7561ecc61946146d6fae2762f/common/djangoapps/third_party_auth/api/views.py#L123).

integrated_channels/integrated_channel/learner_data.py
pomegranited: I'm not fussed about whether this should be a property or a constant.  It's possible that other integration channels will want something totally different, e.g. boolean pass/fail, or numeric grade, or who knows what, so it's hard to know how best to abstract it with just one example.

PR #79, opened 2017-03-31
ENT-284: Refine SAP SuccessFactors course export flow

haikuginger:
**Description:**

This pull request finalizes a couple features to do with the SAP course export feature. Specifically:

1. Retrieve the course language and use it in the submission
2. Create a course track selection URL for the main link, rather than the marketing URL
3. Send up to 1000 courses at once to SAP

**JIRA:** [ENT-284](https://openedx.atlassian.net/browse/ENT-284)

**Dependencies:** None

**Merge deadline:** 10 April 2017

**Installation instructions:**
Requires:
- Course Discovery
    - At least one course
    - At least one catalog containing at least one course
- Enterprise customer
    - Attached to catalog required above
- Configured SapSuccessFactorsGlobalConfiguration (see document provided offline)
- Configured SapSuccessFactorsEnterpriseCustomerConfiguration (see document provided offline)
    - Link to our existing EnterpriseCustomer

**Testing instructions:**
1. Run the following (where `staff` is a local superuser with permissions to the course catalog):
    ```sh
    ./manage.py lms transmit_courseware_data --catalog_user staff --settings=devstack
    ```
2. In the stdout logging of your devstack, observe the following facts:
    - The `locale` key under `description` and `title` should (roughly) match the language entered for the course in the Course Discovery stack, unless the language selected there isn't available. In that case, it should be `English`. Change this selection around a bit, and rerun the command to be sure it tracks expected behavior.
    - Under `content`, the single present item should have a `launch_url` that corresponds to the course track selection page, and which matches the domain of the site linked to your EnterpriseCustomer. Scheme will be `https` by default, unless you set the `SUCCESSFACTORS_COURSE_EXPORT_DEFAULT_URL_SCHEME` environment variable before running the command. Example:
         ```
        https://example.com/course_modes/choose/course-v1:edX+fancy+demo123/
        ```

**Merge checklist:**

- [x] Regenerate requirements with `make upgrade && make requirements` (and make sure to fix any errors).
  **DO NOT** just add dependencies to `requirements/*.txt` files.
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Documentation updated (not only docstrings)
- [x] Commits are (reasonably) squashed
- [x] Translations are updated
- [x] PR author is listed in AUTHORS

**Post merge:**
- [ ] Create a tag
- [ ] Check new version is pushed to PyPi after tag-triggered build is finished.
- [ ] Delete working branch (if not needed anymore)

[...]

PR #82, opened 2017-04-04
ENT-289 Improve accounting for inactive courses

brittneyexline:
**Description:** We were missing a corner case where courses stop being returned for a catalog,
in which case we will need to mark them as inactive.

This change intoduces a lightweight audit summary of the course updates which
will be used in subsequent pushes to fully account for course state changes.

**JIRA:** https://openedx.atlassian.net/browse/ENT-289

**Dependencies:** n/a

**Merge deadline:** 4/7/2017

Make sure this version of edx-enterprise is installed in the environment.
Run the migrations for sap_success_factors.

Course Discovery
- At least two active courses
- At least one catalog containing at least one course

Enterprise customer
- Attached to catalog required above

Configured SapSuccessFactorsGlobalConfiguration (see document provided offline)
Configured SapSuccessFactorsEnterpriseCustomerConfiguration (see document provided offline)
Link to our existing EnterpriseCustomer

**Testing instructions:**

1. Run the following with the catalog having just one course (where `staff` is a local superuser with permissions to the course catalog):
    ```sh
    ./manage.py lms transmit_courseware_data --catalog_user staff --settings=devstack
    ```

2. Observe that in the output, that the active course was pushed to SuccessFactors. Check the sap_success_factors_catalogtransmissionaudit table to see that the audit summary reflects the course that was pushed and that there are no errors.

3. Add another active course to the catalog, re run the script, and observe that both courses are pushed in the same way as above.

4. Remove a course from the catalog, and run the script again. The removed course should show that the status for it was changed to INACTIVE. This can be found in the script output that is logged as well as in the sap_success_factors_catalogtransmissionaudit audit summary.

5. Run the script again, without changing the catalog. Observe that the audit summary for the latest push does not include the course that was removed in step 4.

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:**


haikuginger:
@brittneyexline, you should be set to rebase on master now that #79 is merged.

integrated_channels/sap_success_factors/transmitters/courses.py
```diff
@@ -18,14 +20,35 @@ class SuccessFactorsCourseTransmitter(SuccessFactorsTransmitterBase):
...
+            last_catalog_transmission = CatalogTransmissionAudit.objects.filter(
+                error_message='',
+                enterprise_customer_uuid=self.enterprise_configuration.enterprise_customer.uuid
+            ).latest('created')
+        except ObjectDoesNotExist:
+            last_catalog_transmission = None
+
+        if last_catalog_transmission:
+            last_audit_summary = json.loads(last_catalog_transmission.audit_summary)
+        else:
```
haikuginger: This could be switched to a single try/except/else block, like so:

```python
try:
    last_catalog_transmission = CatalogTransmissionAudit.objects.filter(
        error_message='',
        enterprise_customer_uuid=self.enterprise_configuration.enterprise_customer.uuid,
    ).latest('created')
except ObjectDoesNotExist:
    last_audit_summary = {}
else:
    last_audit_summary = json.loads(last_catalog_transmission.audit_summary)
```

The long and the short of it is that if there's no caught exception, the else block is executed immediately after the try block. It's a little odd, but not quite as odd as the `for-else` construction.

As a side note, this logic will likely need to change, post-#79. The pagination and possibility of errors on one page but not others makes things all the more complicated.

integrated_channels/sap_success_factors/utils.py
```diff
@@ -32,45 +36,91 @@ def get_serialized_data(self):
...
+            new_audit_summary[key] = {
+                'in_catalog': True,
+                'status': status,
+            }
+            # Remove the key from previous audit summary so we can process courses that are no longer present.
+            if key in previous_audit_summary:
+                del previous_audit_summary[key]
+
+        for key, summary in previous_audit_summary.items():
+            new_audit_summary[key] = {
```
haikuginger: Do we want to add this unconditionally? Or only if the course has not previously been set as inactive? It seems like this would carry courses forward in the audit indefinitely. I would think that, when removing inactive courses, we'd only want them to show up in the audit log of the run we actually used to perform the removal.

integrated_channels/sap_success_factors/transmitters/courses.py
```diff
@@ -42,17 +44,34 @@ def transmit_block(self, serialized_payload):
...
         Args:
-            payload (SapCourseExporter): The OCN course exporter object to send to SAP SuccessFactors
+            course_exporter (SapCourseExporter): The OCN course exporter object to send to SAP SuccessFactors
         """
+        CatalogTransmissionAudit = apps.get_model(  # pylint: disable=invalid-name
+            app_label='sap_success_factors',
+            model_name='CatalogTransmissionAudit'
+        )
+
+        try:
```
mattdrayer: Can you add a comment describing the exception case, when/how often it is expected to occur, etc.

integrated_channels/sap_success_factors/utils.py
```diff
@@ -166,26 +169,74 @@ def get_serialized_data_blocks(self):
...
+            An audit summary of courses with information about their presence in the catalog and current status.
+        """
+        if self.removed_courses_resolved:
+            return {}
+
+        new_audit_summary = {}
+        new_courses = []
+
+        for course in self.courses:
+            key = course['courseID']
```
mattdrayer: key --> course_key (or course_id)

integrated_channels/sap_success_factors/utils.py
```diff
@@ -166,26 +169,74 @@ def get_serialized_data_blocks(self):
...
+            status = course['status']
```
mattdrayer: status --> course_status

integrated_channels/sap_success_factors/utils.py
```diff
@@ -166,26 +169,74 @@ def get_serialized_data_blocks(self):
...
+        return new_audit_summary
+
     data_transform = {
         'courseID': lambda x: x['key'],
         'providerID': lambda x: apps.get_model(
             'sap_success_factors',
             'SAPSuccessFactorsGlobalConfiguration'
         ).current().provider_id,
-        'status': lambda x: 'ACTIVE' if x['availability'] == 'Current' else 'INACTIVE',
+        'status': lambda x: (SapCourseExporter.STATUS_ACTIVE if x['availability'] == 'Current'
```
mattdrayer: Seems like we might be able to move these availability values into a similar enum/constant thing.

mattdrayer:
Took a pass thru -- seems like we're on the right track here.

integrated_channels/sap_success_factors/utils.py
```diff
@@ -266,3 +320,13 @@ def transform_language_code(code):
     language_family = SUCCESSFACTORS_OCN_LANGUAGE_CODES[language_code]
     language_name = language_family.get(country_code, language_family['_'])
     return language_name
+
+
+def safe_extract_key(data, key, default=''):
```
brittneyexline: @mattdrayer that makes sense, would you recommend another edx owned library or just a new file in this package? also I'm wondering if there is something like this that already exists.

mattdrayer: I'd start with https://github.com/edx/edx-enterprise/blob/master/enterprise/utils.py -- at some point we might review all of the generalizations located in that module and see about lifting them to another library, but for now utils.py makes sense to me.

mattdrayer:
👍

PR #90, opened 2017-04-18
ENT-247 Increase the max size limit for enterprise logo

zubair-arbi:
@saleem-latif @asadiqbal08 @mattdrayer 

**Description:** Increase the max size limit for enterprise logo

**JIRA:** [ENT-247](https://openedx.atlassian.net/browse/ENT-247)

**Dependencies:** https://github.com/edx/edx-platform/pull/14907

**Merge deadline:** 21 April, 2017

**Installation instructions:** Install `edx-enterprise` from this branch `zub/ENT-247-update-enterprise-logo-image-size` in `edx-platform`.

**Testing instructions:**

1. Go to enterprise customers django admin at `admin/enterprise/enterprisecustomer/`
2. Try to add a new enterprise with a `png` image of size upto `512 KB`
3. Verify that enterprise is saved successfully with an image of size upto `512 KB`
4. Try again to a new enterprise with a `png` image of size greater than `512 KB`
5. Verify that a validation error is raise with message: `The logo image file size must be less than or equal to 512KB.`

**Merge checklist:**

- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/apps.py
+++ b/enterprise/apps.py
@@ -17,8 +17,8 @@ class EnterpriseConfig(AppConfig):
     """
 
     name = "enterprise"
-    valid_extensions = [".png", ]
-    image_size = getattr(settings, 'ENTERPRISE_CUSTOMER_LOGO_IMAGE_SIZE', 4 * 1024)
+    valid_image_extensions = [".png", ]
+    valid_max_image_size = getattr(settings, 'ENTERPRISE_CUSTOMER_LOGO_IMAGE_SIZE', 512)  # Value in KB's
 
     @property
     def auth_user_model(self):
--- a/enterprise/validators.py
+++ b/enterprise/validators.py
@@ -24,7 +24,7 @@ def validate_image_extension(value):
     """
     config = get_app_config()
     ext = os.path.splitext(value.name)[1]
-    if config and not ext.lower() in getattr(config, "valid_extensions", []):
+    if config and not ext.lower() in config.valid_image_extensions:
         raise ValidationError(_("Unsupported file extension."))
 
 
@@ -33,5 +33,7 @@ def validate_image_size(image):
     Validate that a particular image size.
     """
     config = get_app_config()
-    if config and not image.size < getattr(config, "image_size", 0):
-        raise ValidationError(_("The logo image file size must be less than 4KB."))
+    valid_max_image_size_in_bytes = config.valid_max_image_size * 1024
+    if config and not image.size <= valid_max_image_size_in_bytes:
+        raise ValidationError(
+            _("The logo image file size must be less than or equal to %s KB.") % getattr(config, "image_size", 0))
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -730,27 +730,28 @@ class TestEnterpriseCustomerBrandingConfiguration(unittest.TestCase):
             self.assertEqual(EnterpriseCustomerBrandingConfiguration.objects.count(), 2)
 
     @ddt.data(
-        (False, 350 * 1024),
-        (False, 251 * 1024),
-        (False, 250 * 1024),
-        (False, 4 * 1024),
-        (True, 2 * 1024),
-        (True, 3 * 1024),
+        (False, 2048),
+        (False, 1024),
+        (True, 512),
+        (True, 256),
+        (True, 128),
     )
     @ddt.unpack
-    def test_image_size(self, valid_image, image_size):
+    def test_image_size(self, is_valid_image_size, image_size):
         """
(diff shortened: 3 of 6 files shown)


asadiqbal08:
Are we not going to introduce `ENTERPRISE_CUSTOMER_LOGO_IMAGE_SIZE` in edx settings ?

zubair-arbi:
@asadiqbal08 I have created a [PR in edx-platform](https://github.com/edx/edx-platform/pull/14907) to override setting `ENTERPRISE_CUSTOMER_LOGO_IMAGE_SIZE` from `LMS` env.

asadiqbal08:
👍

zubair-arbi:
@brittneyexline Please review

enterprise/apps.py
```diff
@@ -18,7 +18,7 @@ class EnterpriseConfig(AppConfig):
 
     name = "enterprise"
     valid_extensions = [".png", ]
```
haikuginger: While we're in here, can we make this variable more explicit?

tests/test_models.py
```diff
@@ -711,12 +711,11 @@ def test_branding_configuration_saving_successfully(self):
             self.assertEqual(EnterpriseCustomerBrandingConfiguration.objects.count(), 2)
 
     @ddt.data(
-        (False, 350 * 1024),
-        (False, 251 * 1024),
-        (False, 250 * 1024),
-        (False, 4 * 1024),
-        (True, 2 * 1024),
-        (True, 3 * 1024),
+        (False, 2048 * 1024),
```
haikuginger: Rather than needing to change this test, let's actually add the settings to `test_settings.py` and rely on that configuration remaining the same. Either that or use the `@override_settings` decorator.

enterprise/apps.py
```diff
@@ -18,7 +18,7 @@ class EnterpriseConfig(AppConfig):
...
-    image_size = getattr(settings, 'ENTERPRISE_CUSTOMER_LOGO_IMAGE_SIZE', 4 * 1024)
+    image_size = getattr(settings, 'ENTERPRISE_CUSTOMER_LOGO_IMAGE_SIZE', 512 * 1024)
```
haikuginger: Let's store the expected max image size in KB instead of in bytes.

enterprise/validators.py
```diff
@@ -33,5 +33,5 @@ def validate_image_size(image):
     Validate that a particular image size.
     """
     config = get_app_config()
-    if config and not image.size < getattr(config, "image_size", 0):
-        raise ValidationError(_("The logo image file size must be less than 4KB."))
+    if config and not image.size <= getattr(config, "image_size", 0):
+        raise ValidationError(_("The logo image file size must be less than or equal to 512KB."))
```
haikuginger: This string is wrong. In combination with the above change to store the max image size in KB, change this string to be formatable so that our configured value is what shows up as the error message, rather than just the platform default.

zubair-arbi:
@haikuginger I have addressed your feedback, please review.

enterprise/validators.py
```diff
@@ -24,7 +24,7 @@ def validate_image_extension(value):
     """
     config = get_app_config()
     ext = os.path.splitext(value.name)[1]
-    if config and not ext.lower() in getattr(config, "valid_extensions", []):
+    if config and not ext.lower() in getattr(config, "valid_image_extensions", []):
```
haikuginger: Nit: Given that the `valid_image_extensions` variable used here (as well as the `image_size` variable used below) is defined unconditionally on the config, do we really need the `getattr` calls here and below? I'd just go with `config.valid_image_extensions`; we don't need yet another layer of defaults.

tests/test_models.py
```diff
@@ -711,26 +711,26 @@ def test_branding_configuration_saving_successfully(self):
...
+        (True, 512),
+        (True, 256),
+        (True, 128),
     )
     @ddt.unpack
-    def test_image_size(self, valid_image, image_size):
+    def test_image_size(self, is_valid_image_size, image_size):
         """
-        Test image size, image_size < (4 * 1024) e.g. 4kb. See apps.py.
+        Test image size in KB's, image_size < (4 * 1024) e.g. 4KB. See apps.py.
```
haikuginger: The string here still refers to the old base limit. Please update that string to incorporate the new 512KB limit, as well as some language indicating that the 512KB limit is just the default.

enterprise/validators.py
zubair-arbi: Agreed. I will update it.

PR #91, opened 2017-04-18
ENT-303: Restrict entitlements to Enterprise-linked coupons

haikuginger:
**Description:**

This pull request updates the EnterpriseCustomer admin panel such that entitlement coupon IDs are entered via a dropdown selection, rather than via a blank field. This allows us to restrict the selection of coupons that are either linked to no EnterpriseCustomer, or coupons that are linked to a different EnterpriseCustomer, preventing an edge case where entitlements that weren't linked to an EnterpriseCustomer on the ecommerce side didn't prompt for consent.

**JIRA:** [ENT-303](https://openedx.atlassian.net/browse/ENT-303)

**Dependencies:** Depends on edx/ecommerce#1247

**Merge deadline:** 24 April 2017

1. Requires fully-configured ecommerce setup
2. Create multiple coupons; link some of them to your devstack's EnterpriseCustomer; leave some of them disconnected.

**Testing instructions:**

1. Go to the EnterpriseCustomer admin; click to add an entitlement.
2. In the drop-down menu, verify that only those coupons that are linked to your EnterpriseCustomer in the ecommerce database are available to select.
3. Shut down your ecommerce stack and refresh the page.
4. Observe that the widget for selecting an entitlement ID is blank and doesn't show any options.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/enterprise/admin/__init__.py
+++ b/enterprise/admin/__init__.py
@@ -85,12 +85,37 @@ class EnterpriseCustomerEntitlementInline(admin.StackedInline):
     ecommerce_coupon_url.short_description = 'Coupon URL'
 
     def get_formset(self, request, obj=None, **kwargs):
+        """
+        Retrieve the inline formset for the EnterpriseCustomerEntitlementInline admin.
+
+        Inlines don't get to have their get_form method called, like regular admins do.
+        The effect of this is that we're a bit more limited in what we can patch onto
+        that form. For a standard get_form method, we could make a call to retrieve the
+        list of allowed coupons and patch the options on directly, but we can't do that
+        here. What we _can_ do is set the formfield_callback kwarg.
+
+        What this does is provide a function that will, given a particular database field
+        that needs to be included in the inline, returns the appropriate form field widget
+        to render in the form. We provide the standard method that would be used anyway,
+        but patched with an extra kwarg containing the details of the EnterpriseCustomer
+        object that's provided to this method, but not provided to that method by default.
+        functools.partial lets us add that variable on, and then the standard logic is enough.
+        """
         kwargs['formfield_callback'] = partial(self.formfield_for_dbfield, request=request, obj=obj)
         return super(EnterpriseCustomerEntitlementInline, self).get_formset(request, obj, **kwargs)
 
     def formfield_for_dbfield(self, db_field, **kwargs):
+        """
+        For most fields, just fall through to the standard behavior. If the field is entitlement_id set up a dropdown.
+
+        This method gets wrapped with functools.partial, as described in the get_formset method. When
+        wrapped in that manner, it receives extra `request` and `obj` kwargs that can be used to get
+        additional information. For most fields, simply call super() and return the standard result;
+        for `entitlement_id`, call the EnterpriseCustomer to retrieve a list of the coupons that can
+        be associated as entitlements.
+        """
         enterprise_customer = kwargs.pop('obj', None)
-        request = kwargs.get('request', None)
+        request = kwargs.get('request')
         formfield = super(EnterpriseCustomerEntitlementInline, self).formfield_for_dbfield(db_field, **kwargs)
         if db_field.name == "entitlement_id" and enterprise_customer:
             return ChoiceField(choices=enterprise_customer.coupon_options(request.user), required=True)
--- a/enterprise/admin/__init__.py
+++ b/enterprise/admin/__init__.py
@@ -4,10 +4,13 @@ Django admin integration for enterprise app.
 """
 from __future__ import absolute_import, unicode_literals
 
+from functools import partial
+
 from django.conf.urls import url
 from django.contrib import admin
 from django.utils.html import format_html
 from django.contrib.auth import settings
+from django.forms import ChoiceField
 from django.http import HttpResponseRedirect
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext as _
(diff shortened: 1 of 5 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=91&repo=edx%2Fedx-enterprise"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

enterprise/admin/__init__.py
```diff
@@ -81,6 +84,18 @@ def ecommerce_coupon_url(self, instance):
     ecommerce_coupon_url.allow_tags = True
     ecommerce_coupon_url.short_description = 'Coupon URL'
 
+    def get_formset(self, request, obj=None, **kwargs):
+        kwargs['formfield_callback'] = partial(self.formfield_for_dbfield, request=request, obj=obj)
+        return super(EnterpriseCustomerEntitlementInline, self).get_formset(request, obj, **kwargs)
+
+    def formfield_for_dbfield(self, db_field, **kwargs):
+        enterprise_customer = kwargs.pop('obj', None)
+        request = kwargs.get('request', None)
```
itsjeyd: @haikuginger You can remove `None` here.

enterprise/ecommerce_api.py
```diff
@@ -0,0 +1,62 @@
...
+        When we call this name, return False, since we couldn't import.
+        """
+        return False
+
+    ecommerce_api_client = None
+
+
+class EcommerceApiClient(object):
+    """
+    Object builds an API client to make calls to the Ecommerce API.
```
itsjeyd: @haikuginger Nit: Standardize spelling: `ecommerce API` vs. `Ecommerce API`. I'd slightly prefer the former.

enterprise/ecommerce_api.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def get_single_coupon(self, coupon_id):
+        """
+        Get the details of a single coupon, identified by ID.
+        """
+        if self.client is None:
+            return {}
+        try:
+            response = self.client.coupons(coupon_id).get()
+        except Exception:  # pylint: disable=broad-except
```
itsjeyd: @haikuginger What types of exceptions is this supposed to catch? It would be worth mentioning in the docstring that this method will always return a result/silence any errors, and explaining why this is necessary.

enterprise/ecommerce_api.py
```diff
@@ -0,0 +1,62 @@
...
+        results = []
+        page = 1
+        next_page = True
+        try:
+            while next_page:
+                response = self.client.coupons.get(enterprise_customer=enterprise_customer_uuid, page=page)
+                results += response.get('results', [])
+                next_page = response.get('next')
+                page += 1
+        except Exception:  # pylint: disable=broad-except
```
itsjeyd: @haikuginger [Same comment]() here.

tests/test_ecommerce_api.py
```diff
@@ -0,0 +1,123 @@
...
+    def setUp(self):
+        self.subclient = mock.MagicMock()
+        with mock.patch('enterprise.ecommerce_api.ecommerce_api_client') as client_getter:
+            client_getter.return_value = self.subclient
+            with mock.patch('enterprise.ecommerce_api.is_commerce_service_configured') as is_configured:
+                is_configured.return_value = True
+                self.client = EcommerceApiClient(mock.MagicMock(spec=User))
+        self.id_getter = self.subclient.coupons
+        self.get_mock = self.subclient.coupons.return_value.get
+        self.get_multiple_mock = self.subclient.coupons.get
```
itsjeyd: @haikuginger Can we update the names of these fields as follows?

* `get_mock` ⟶ `get_coupons_mock`
* `get_multiple_mock` ⟶ `get_multiple_coupons_mock`
* And perhaps: `id_getter` ⟶ `coupon_getter`

tests/test_ecommerce_api.py
```diff
@@ -0,0 +1,123 @@
...
+        assert response == {}
+
+    def test_get_multiple_coupons(self):
+        self.get_multiple_mock.side_effect = [
+            {"results": [1, 2, 3], "next": "url"},
+            {"results": [4, 5, 6]},
+            {"results": [7, 8, 9]}
+        ]
+        response = self.client.get_coupons_by_enterprise_customer('123-456-789-ABC-DEF-0')
+        assert response == [1, 2, 3, 4, 5, 6]
```
itsjeyd: @haikuginger Why would the the last three coupons (`7`, `8`, `9`) be missing from the response?

tests/test_ecommerce_api.py
```diff
@@ -0,0 +1,123 @@
...
+        mock_is_configured.return_value = True
+        resulting_api_client_singleton = object()
+        mock_api_client.return_value = resulting_api_client_singleton
+        client = EcommerceApiClient(mock.MagicMock(spec=User))
+        assert client.client is resulting_api_client_singleton
+
+
+class TestEcommerceApiClientBehavior(unittest.TestCase):
+    """
+    Test the behavior of getting a single coupon by ID.
```
itsjeyd: @haikuginger Let's go with a more general docstring here, to take into account the fact that this class also tests behavior for getting multiple coupons.

enterprise/admin/__init__.py
haikuginger: True!

enterprise/admin/__init__.py
```diff
@@ -81,6 +84,18 @@ def ecommerce_coupon_url(self, instance):
...
+        formfield = super(EnterpriseCustomerEntitlementInline, self).formfield_for_dbfield(db_field, **kwargs)
+        if db_field.name == "entitlement_id" and enterprise_customer:
+            return ChoiceField(choices=enterprise_customer.coupon_options(request.user), required=True)
+        return formfield
+
```
haikuginger: Good point; will do.

enterprise/ecommerce_api.py
haikuginger: Agreed.

itsjeyd:
@haikuginger I verified that the changes work as expected. Will drop a thumbs-up when you're done addressing comments from the code review.

enterprise/ecommerce_api.py
haikuginger: It needs to catch any exception from edx-rest-api-client, slumber, requests, and urllib3, not all of which inherit from a common root. Good idea on the docstring.

enterprise/ecommerce_api.py
haikuginger: As above.

tests/test_ecommerce_api.py
haikuginger: Ah, yes. I was originally going to have separate classes before realizing there was a bunch of shared scope.

tests/test_ecommerce_api.py
haikuginger: Sounds good!

tests/test_ecommerce_api.py
haikuginger: Note that the second item in the `side_effect` list doesn't have a `next` value; in other words, from the perspective of the client, it's the last page. Therefore, it doesn't make another API call. Having the `{"results": [7, 8, 9]}` item ensures that we stop paging at the appropriate place.

haikuginger:
After having a conversation with Griff, we're going to close this PR and defer the overall story until later; it sounds like coupons are in a bit of a state of flux right now, so we don't need the extra complication.

PR #94, opened 2017-04-25
ENT-362 Fixing inactivation of courses that are removed from catalogs

brittneyexline:
**Description:** We weren't sending all of the required fields for a course update, which causes the entire call to fail even for properly formatted courses in the same payload.

**JIRA:** https://openedx.atlassian.net/browse/ENT-362

**Dependencies:** n/a

**Installation instructions:** Requires an environment with the catalog service running.

**Testing instructions:**

Edxapp sandbox: https://brittneyexline.sandbox.edx.org
Discovery sandbox: https://discovery-brittneyexline.sandbox.edx.org

1. Set up a configuration for an enterprise customer and success factors against our test SAP instance. (This has been done already on the sandbox)
2. Make a catalog with two or more courses. (On the sandbox it is "One Catalog to rule them all")
3. Run the catalog export script, ensure that it has run successfully. To do this on the sandbox, ssh to the edxapp sandbox. Then run the following commands:

`$ sudo -H -u edxapp bash`
`$ source /edx/app/edxapp/edxapp_env`
`$ cd /edx/app/edxapp/edx-platform/`
`$ /edx/bin/python.edxapp /edx/bin/manage.edxapp lms transmit_courseware_data --catalog_user staff --settings=aws`

To see the courses on SAP, follow the instructions here for Importing Courses to view which courses have been sent.
https://openedx.atlassian.net/wiki/pages/viewpage.action?spaceKey=SOL&title=SAP+Testing+and+Configuration+notes

4. Remove one of the courses from the catalog.
5. Run the script again, and observe that the call to SAP was successful, and that the removed course was sent as INACTIVE. Check SAP's interface, the course that was removed should not show up in the Imported Courses search view.

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


integrated_channels/sap_success_factors/utils.py
```diff
@@ -322,3 +319,35 @@ def transform_language_code(code):
...
+def get_course_metadata_for_inactivation(course_id, enterprise_customer):
+    """
+    Provide the minimal course metadata structure for updating a course to be inactive.
+    """
+    return {
+        'courseID': course_id,
+        'providerID': apps.get_model(
+            'sap_success_factors',
+            'SAPSuccessFactorsGlobalConfiguration'
+        ).current().provider_id,
```
bradenmacdonald:
```python
        apps.get_model(
            'sap_success_factors',
        ).current()
```
is repeated... can you pull that part out to happen before the `return {...}` ?

haikuginger:
@brittneyexline, can you update the PR to take advantage of edx/edx-platform#14941? In the meantime, I'll just use a version of platform slightly older than that. :)

haikuginger:
Can you also update the client to call `response.raise_for_status()` on the response? Otherwise, it just fails silently.

haikuginger:
Okay, I tested this, and I was able to successfully send a deactivated course, but SAP's rather slim API limits raise another issue - we're saving a list of courses in the audit field, even if the request fails. If it fails, we should probably either _not_ save the audit summary, or we should check for an error code prior to taking action on it. We're headed in the right direction, though.

brittneyexline:
The logic to get the previous audit summary filters to look for the last one that didn't have an error message, would that be sufficient? So if the call to deactivate the course fails, we will be using the last successful call to say that it still needs to be deactivated the next time around.

haikuginger:
Ah, I must have missed that - sounds okay to me.

integrated_channels/sap_success_factors/utils.py
```diff
@@ -322,3 +319,33 @@ def transform_language_code(code):
...
+
+
+def get_course_metadata_for_inactivation(course_id, enterprise_customer):
+    """
+    Provide the minimal course metadata structure for updating a course to be inactive.
+    """
+    provider_id = apps.get_model(
+        'sap_success_factors',
+        'SAPSuccessFactorsGlobalConfiguration'
+    ).current().provider_id
```
douglashall: Any chance we can pass the provider_id into this function and avoid making this call for every course?

PR #98, opened 2017-05-04
ENT-317 Add externally managed consent option for data sharing

brittneyexline:
**Description:** For integrations like SAP, consent to share data back has already been gathered by the partner, and to simplify the experience for users, we want to skip prompting for consent while still recording and collecting data to send back to the partner.

**JIRA:** https://openedx.atlassian.net/browse/ENT-317

**Dependencies:** n/a

**Merge deadline:** 5/5/2017

**Testing instructions:**

Sandbox: https://brittneyexline.sandbox.edx.org/dashboard

1. Login as staff to django admin, configure an enterprise customer (Bexline Inc recommended on the above sandbox) to have data sharing consent enabled and enforces data sharing consent to be externally managed. Save and logout.
2. Go to the registration page and sign in using the TestShib IDP, which should be linked to the enterprise customer. Finish the registration flow through to the dashboard.
3. SSH to the machine and using the mysql command line tool, verify that a new record has been added to enterprise_userdatasharingconsentaudit with state = 'externally_managed'.
4. Bonus (but not properly configured on my sandbox because discovery is down): Go to a tpa-hinted track selection page with a verified option for the course. Go through the verified course option all the way to checkout and the receipt page. You should not be prompted for consent at any point, and a new Enterprise Enrollment record should be written.

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:** I want to note that this option will only work, and only makes sense, if SSO is configured for the enterprise customer. This PR was made on that assumption.


enterprise/tpa_pipeline.py
```diff
@@ -142,10 +142,17 @@ def redirect_to_consent():
...
+        ec_user, _ = EnterpriseCustomerUser.objects.get_or_create(
             enterprise_customer=enterprise_customer,
             user_id=user.id
         )
+
+        if enterprise_customer.enforce_data_sharing_consent == EnterpriseCustomer.EXTERNALLY_MANAGED:
+            UserDataSharingConsentAudit.objects.update_or_create(
+                user=ec_user,
+                state=UserDataSharingConsentAudit.EXTERNALLY_MANAGED
+            )
```
douglashall: I think you want this to be:
```
UserDataSharingConsentAudit.objects.update_or_create(
    user=ec_user,
    defaults={
        state=UserDataSharingConsentAudit.EXTERNALLY_MANAGED
    }
)
```

enterprise/tpa_pipeline.py
```diff
@@ -142,10 +142,17 @@ def redirect_to_consent():
     if not enterprise_customer.requests_data_sharing_consent:
         # This enterprise customer attached to this pipeline element does not request data sharing consent;
         # proceed with the creation of a link between the user and the enterprise customer, then exit.
-        EnterpriseCustomerUser.objects.get_or_create(
+        ec_user, _ = EnterpriseCustomerUser.objects.get_or_create(
             enterprise_customer=enterprise_customer,
             user_id=user.id
         )
+
+        if enterprise_customer.enforce_data_sharing_consent == EnterpriseCustomer.EXTERNALLY_MANAGED:
```
douglashall: So if consent is disabled and is not externally managed, do we not create a UserDataSharingConsentAudit record for the EnterpriseCustomerUser? Sorry, that is a lot of NOTs.

brittneyexline: Yes, that is true - if consent is disabled we shouldn't write this record. I'm going to update this to make that clear, since I believe a model could have consent disabled but still have this value set to externally managed, and the consent enabled field takes precedence.

enterprise/tpa_pipeline.py
```diff
@@ -142,10 +142,21 @@ def redirect_to_consent():
...
-        EnterpriseCustomerUser.objects.get_or_create(
+        ec_user, _ = EnterpriseCustomerUser.objects.get_or_create(
             enterprise_customer=enterprise_customer,
             user_id=user.id
         )
+
+        if (enterprise_customer.enforce_data_sharing_consent == EnterpriseCustomer.EXTERNALLY_MANAGED
+                and enterprise_customer.enable_data_sharing_consent):
+
+            UserDataSharingConsentAudit.objects.update_or_create(
```
haikuginger: I could be wrong, but I think we don't need to go all the way out to the UserDataSharingConsentAudit object; we should be able to do something like `ec_user.data_sharing_consent.update_or_create()`. Not positive, though, but I think Django can do that.

tests/test_tpa_pipeline.py
```diff
@@ -155,6 +155,33 @@ def test_handle_enterprise_logistration_consent_not_required(self):
                 enterprise_customer=enterprise_customer,
                 user_id=self.user.id
             ).count() == 1
+            assert UserDataSharingConsentAudit.objects.filter(
+                user__user_id=self.user.id,
+                user__enterprise_customer=enterprise_customer,
+            ).count() == 0
```
haikuginger: 👍

tests/test_tpa_pipeline.py
```diff
@@ -155,6 +155,33 @@ def test_handle_enterprise_logistration_consent_not_required(self):
...
+
+    def test_handle_enterprise_logistration_consent_externally_managed(self):
+        """
+        Test that when consent is externally managed, we create an EnterpriseCustomerUser and
+        UserDataSharingConsentAudit object, then return.
+        """
+        backend = mock.MagicMock(name=None)
+        with mock.patch('enterprise.tpa_pipeline.get_ec_for_running_pipeline') as fake_get_ec:
```
haikuginger: You can make this as a decorator and pass it as an arg to help avoid getting too indented. Like so:

```python
@mock.patch('enterprise.tpa_pipeline.get_ec_for_running_pipeline')
def test_handle_enterprise_logistration_consent_externally_managed(self, fake_get_ec):
    pass
```

tests/test_tpa_pipeline.py
```diff
@@ -155,6 +155,33 @@ def test_handle_enterprise_logistration_consent_not_required(self):
...
+                user__user_id=self.user.id,
+                user__enterprise_customer=enterprise_customer,
+            ).count() == 0
+
+    def test_handle_enterprise_logistration_consent_externally_managed(self):
+        """
+        Test that when consent is externally managed, we create an EnterpriseCustomerUser and
+        UserDataSharingConsentAudit object, then return.
+        """
+        backend = mock.MagicMock(name=None)
```
haikuginger: This doesn't do what you would probably think it does. Passing `name` as an arg to the mock instantiator does _not_ cause it to mock out the `name` attribute on the produced mock. You have to create the mock and then do `mock.name = None`.

```python
>>> import mock
>>> x = mock.MagicMock(name=None)
>>> x
<MagicMock id='4548168976'>
>>> print(x.name)
<MagicMock name='mock.name' id='4548479440'>
>>> x.name = None
>>> print(x.name)
None
```

Are you using that value to test anything? If so, double-check your tests, because they're not working. If not, then we can probably skip putting a value there anyway - mocks will automatically return another `MagicMock` for any attributes that aren't explicitly defined.

haikuginger:
Behavior looks great! I'll test again once you've made the suggested changes.

enterprise/models.py
```diff
@@ -589,16 +591,18 @@ class Meta(object):
...
         (ENABLED, 'Enabled'),
         (DISABLED, 'Disabled'),
+        (EXTERNALLY_MANAGED, 'Managed Externally'),
     )
 
     user = models.ForeignKey(EnterpriseCustomerUser, related_name='data_sharing_consent')
 
     state = models.CharField(
-        max_length=8,
+        max_length=20,
```
douglashall: Would we have enough data in PROD at this point for this to be a problem? Though I suppose it might be a good thing to shorten the new state value string down.

brittneyexline: Last I checked we had about 10 enterprise customers - but I'm ok with shortening the string.

enterprise/tpa_pipeline.py
brittneyexline: What would be the benefit of doing that over this method?

douglashall: I think they are two different ways of accessing the same ModelManager. To me, referencing the model class as you have here is more readable.

PR #104, opened 2017-05-11
[ENT-290] Enable switch in Enterprise setup to hide Audit track

replaceafill:
**Description:** This PR adds a boolean flag "Hide audit track" to the EnterpriseCustomer model and adjusts the new landing page to filter course modes accodingly.

**JIRA:** [ENT-290](https://openedx.atlassian.net/browse/ENT-290)

**Dependencies:**  
[PR on edx/edx-platform for the new `ENTERPRISE_COURSE_ENROLLMENT_AUDIT_MODES` setting](https://github.com/edx/edx-platform/pull/15095)

** Screenshots:**

**Testing instructions:**

1. Add an Enterprise Customer in the Django admin. Check its ID
2. Add a course in the customer catalog. Check its ID.
3. Visit the new landing page for the enterprise customer and course by typing a URL like this:

http://localhost:8000/enterprise/7c3c286a-9834-46ae-809e-e01b114ba980/course/course-v1:edX+DemoX+Demo_Course/enroll/

where 7c3c286a-9834-46ae-809e-e01b114ba980 should be the ID of the customer and 'course-v1:edX+DemoX+Demo_Course' should be the ID of the course

4. Check that the landing page shows both course modes (Enroll and Audit).
5. Go to the Enterprise Customer and check the "Hide audit track" flag
6. Go back to the landing page and verify that the Audit track is now hidden

**Merge checklist:**

- [ ] ~Regenerate requirements with `make upgrade && make requirements` (and make sure to fix any errors).
  **DO NOT** just add dependencies to `requirements/*.txt` files.~
- [X] All reviewers approved
- [X] CI build is green
- [X] Version bumped
- [X] Changelog record added
- [X] Documentation updated (not only docstrings)
- [X] Commits are (reasonably) squashed
- [X] Translations are updated
- [X] PR author is listed in AUTHORS

**Post merge:**
- [ ] Create a tag

**Reviewers**
- [ ] @haikuginger


edx-webhook:
Thanks for the pull request, @replaceafill! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=104"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

mattdrayer:
Wow great timing!  I was just talking with @haikuginger about this :)

enterprise/utils.py
```diff
@@ -512,3 +512,20 @@ def safe_extract_key(data, key, default=''):
...
+    Arguments:
+        enterprise_customer: The EnterpriseCustomer that the enrollment was created using.
+        course_modes: iterable with dictionaries containing a required 'mode' key
+    """
+    if not enterprise_customer.hide_audit_track:
+        return course_modes
+    return [
+        course_mode
+        for course_mode in course_modes
+        if not course_mode['mode'] in settings.ENTERPRISE_COURSE_ENROLLMENT_AUDIT_MODES
```
bradenmacdonald: I'm curious why `ENTERPRISE_COURSE_ENROLLMENT_AUDIT_MODES` needs to be a setting? And why it's specific to enterprise?

haikuginger: I asked for that actually. We can set a sane default and probably never touch it, but this way it can be changed if it needs to - as it has in the past (honor->audit).

replaceafill:
I don't understand why the Travis build failed with a transaction error on TestEnterpriseAPIViews. `make test-all` doesn't complain locally. Travis hiccup? Could we try again?

bradenmacdonald:
Just asking in case this hasn't been discussed yet: Are we certain that the only use case for this is every going to be "hide audit mode"? Because if not, it could make sense to replace the new boolean field with a comma-separated list field on the EnterpriseCustomer where we can specify "excluded modes", which could also make the code simpler and remove the need for that setting.

mattdrayer:
Also we need to think about this from an API client perspective -- for example I am making a slight modification to the LMS track selection page to also be aware of the scenario where we need to hide the audit mode option -- or perhaps as @bradenmacdonald suggested it would be a list of modes -- one case I can think of right off the bat is the "honor" mode, which may or may not be in use but there is code in the system which looks for that.

mattdrayer:
See [here](https://github.com/edx/edx-platform/blob/master/common/djangoapps/course_modes/views.py#L157) for an example of where information is being pulled from the Enterprise service for use in the LMS -- my plan is to `pop()` any mode choices from the `modes` container which should be hidden from the user's view.

enterprise/templates/enterprise/enterprise_course_enrollment_page.html
```diff
@@ -55,26 +55,29 @@ <h3 class="course-confirmation-title">${ confirmation_text }</h3>
...
+                <strong class="title">${ course_mode['title'] }</strong>
+                <span class="price">
+                  ${ price_text }:
+                  % if course_mode['final_price'] and course_mode['original_price'] != course_mode['final_price']:
+                  <strike>${ course_mode['original_price'] }</strike> ${ course_mode['final_price'] }
+                  % else:
+                  ${ course_mode['original_price'] }
+                  % endif
+                </span>
+                ${ course_mode['description'] }
```
haikuginger: Should this be inside a `<span>` or a `<p>`?

enterprise/utils.py
haikuginger: On that topic, though, we should generally _not_ be using bare `settings.VARIABLE` calls. Instead, do something like this:

```python
audit_modes = getattr(settings, 'ENTERPRISE_COURSE_ENROLLMENT_AUDIT_MODES', ['audit'])
return [course_mode for course_mode in course_modes if not course_mode['mode'] in audit_modes]
```

enterprise/views.py
```diff
@@ -514,15 +532,9 @@ def get_enterprise_course_enrollment_page(self, request, enterprise_uuid, course
...
-            'enroll_text': self.context_data['enroll_text'],
-            'audit_text': self.context_data['audit_text'],
             'price_text': self.context_data['price_text'],
-            'free_price_text': self.context_data['free_price_text'],
-            'discount_text': self.context_data['discount_text'].format(
-                enterprise_customer_name=enterprise_customer.name,
-            ),
-            'not_eligible_for_certificate_text': self.context_data['not_eligible_for_cert_text'],
             'continue_link_text': self.context_data['continue_link_text'],
+            'course_modes': filter_audit_course_modes(enterprise_customer, course_modes),
```
haikuginger: 👍 Much better!

tests/test_utilities.py
```diff
@@ -1221,3 +1227,47 @@ def test_resolve_removed_courses(
...
         assert second_audit_summary == {}
+
+    @ddt.data(
+        (
+            {
+                'mode': 'enroll',
+                'number': 0,
+            },
+            {
+                'mode': '__audit__',
```
haikuginger: It doesn't come into play here, but because the double-underscore pattern has specific semantic meaning, I'd prefer that you avoid using it in testing strings.

replaceafill:
@haikuginger I just addressed your observations.

enterprise/utils.py
```diff
@@ -512,3 +512,17 @@ def safe_extract_key(data, key, default=''):
...
+    """
+    Filter audit course modes out if the enterprise customer has the 'Hide audit track' flag enabled.
+
+    Arguments:
+        enterprise_customer: The EnterpriseCustomer that the enrollment was created using.
+        course_modes: iterable with dictionaries containing a required 'mode' key
+    """
+    if not enterprise_customer.hide_audit_track:
+        return course_modes
+    audit_modes = getattr(settings, 'ENTERPRISE_COURSE_ENROLLMENT_AUDIT_MODES', ['audit'])
```
mattdrayer: We'll need to add this setting to [common.py](https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L3126) and [aws.py](https://github.com/edx/edx-platform/blob/master/lms/envs/aws.py#L910)

mattdrayer: As well as add it to https://github.com/edx/configuration and/or https://github.com/edx/edx-internal

enterprise/models.py
```diff
@@ -121,6 +121,13 @@ class Meta:
         )
     )
 
+    hide_audit_track = models.BooleanField(
```
mattdrayer: please add this new field to the [EnterpriseCustomer serializer](https://github.com/edx/edx-enterprise/blob/master/enterprise/api/v1/serializers.py#L65)

douglashall: Could we change this to `enable_audit_enrollment` with a default of `False`? To me, this is a better description of what this field means and is a more appropriate default based on the use cases I have heard.

enterprise/views.py
```diff
@@ -490,6 +490,24 @@ def get_enterprise_course_enrollment_page(self, request, enterprise_uuid, course
...
+                ),
+            },
+            {
+                'mode': 'audit',
+                'title': self.context_data['audit_text'],
+                'original_price': self.context_data['free_price_text'],
+                'final_price': '',
+                'description': self.context_data['not_eligible_for_cert_text'],
+            },
+        ]
```
douglashall: Do we have a followup story for pulling this data from an actual data source?

enterprise/templates/enterprise/enterprise_course_enrollment_page.html
```diff
@@ -55,26 +55,29 @@ <h3 class="course-confirmation-title">${ confirmation_text }</h3>
...
             <div class="caption">${ select_mode_text }</div>
+            % for course_mode in course_modes:
             <div class="radio-block">
               <div class="radio">
-                <input type="radio" name="data_sharing_consent" id="radio1" checked="checked"/>
+                <input type="radio" name="data_sharing_consent" id="radio${loop.index}"
+                       % if loop.first:
+                       checked="checked"
+                       % endif
+                       />
```
douglashall: Do we need a radio button if there is only one option?

enterprise/views.py
haikuginger: Yes, that's ENT-348, which I'm working on.

enterprise/templates/enterprise/enterprise_course_enrollment_page.html
haikuginger: Good point - possibly change this to a hidden field with a default value if there's just one item.

enterprise/utils.py
replaceafill: @mattdrayer I created this [PR](https://github.com/edx/edx-platform/pull/15095) to add the setting to the common and aws modules.

Could you please [review this commit](https://github.com/replaceafill/configuration/commit/98b70ea90a1dfd8feb5f235c8602f7e73d58d318) to add the setting to the configuration repository and recommend how could I test that?

replaceafill:
@mattdrayer @douglashall I've addressed your observations:

- Changed the flag to enable_audit_enrollment and added it to the api serializer
- Used a hidden field if there's only one mode displayed in the landing page

I don't have experience yet testing changes to the configuration repository. I'd appreciate any pointers on that.

haikuginger:
@replaceafill, please go through the merge checklist, except don't set a version number quite yet. We'll want to get edx/edx-platform#15095 and the related configuration change approved first.

mattdrayer:
:shipit:

replaceafill:
@haikuginger Merge checklist done. Let me know if I missed something.

PR #105, opened 2017-05-12
ENT-355 Skip Landing pages if learner already enrolled

bdero:
**Description:** This change adds a check to the enterprise course enrollment view that forwards the user directly to the course home page when the user is already enrolled in the course.

**JIRA:** [ENT-355](https://openedx.atlassian.net/browse/ENT-355)

**Dependencies:** No dependencies.

**Merge deadline:** May 19th

1. Setup a devstack on `master`
1. Clone ecommerce into `/edx/src/edx-ecommerce`
1. Checkout this branch: `git checkout bdero/ent-355`
1. Install the local ecommerce:
    ```
    sudo -Hu edxapp bash
    pip install -e file:///edx/src/edx-enterprise#egg=edx-enterprise==localdev
    ./manage.py lms migrate enterprise --settings=devstack
    ```

**Testing instructions:**

1. Create a new EnterpriseCustomer in the [admin panel](http://localhost:8000/admin/enterprise/enterprisecustomer/add/):
    1. **Name** it "Test Enterprise Customer"
    1. For the **Site**, select the default "example.com" site
    1. Upload an image for the new EnterpriseCustomer under the "Branding configuration" section
    1. Click "Save"
1. Login to edX with an account that's not currently enrolled in the course; you should see the landing page:
    screenshot of landing page
1. Navigate to the enterprise course enrollment view. For example:
    http://localhost:8000/enterprise/d986d249-5b8f-4992-9672-c53668ef5502/course/course-v1%3Aasdf%2Basdf%2Basdf/enroll/
1. Go to the EnterpriseCustomerLearner [admin panel](http://localhost:8000/admin/enterprise/enterprisecustomeruser/add/) and create a new EnterpriseCustomerUser:
    1. For the "User id", enter the LMS user ID number of a user to enroll
    1. For the "EnterpriseCustomer", select the enterprise created during **step 1**
    1. Click "Save"
1. Now, enroll in the course as the user used in **step 4**
1. Navigate back to the URL from **step 3**; it should redirect you straight to the **course's home page**

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -60,7 +60,7 @@ from six.moves.urllib.parse import urljoin  # pylint: disable=import-error
 
 logger = getLogger(__name__)  # pylint: disable=invalid-name
 
-LMS_COURSE_URL = urljoin(settings.LMS_ROOT_URL, '/courses/{course_id}')
+LMS_COURSE_URL = urljoin(settings.LMS_ROOT_URL, '/courses/{course_id}/courseware')
 
 
 def verify_edx_resources():
@@ -505,10 +505,9 @@ class CourseEnrollmentView(View):
         enterprise_customer = get_enterprise_customer_or_404(enterprise_uuid)
 
         enrollment_client = EnrollmentApiClient()
-        enrolled_courses = enrollment_client.get_enrolled_courses(request.user.username)
-        enrolled_course_ids = [course['course_details']['course_id'] for course in enrolled_courses]
+        enrolled_course = enrollment_client.get_course_enrollment(request.user.username, course_id)
 
-        if (course_id in enrolled_course_ids or
+        if (enrolled_course is not None or
                 EnterpriseCourseEnrollment.objects.filter(
                     enterprise_customer_user__enterprise_customer=enterprise_customer,
                     enterprise_customer_user__user_id=request.user.id,
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -40,7 +40,7 @@ except ImportError:
 
 # isort:imports-firstparty
 from enterprise.constants import CONFIRMATION_ALERT_PROMPT, CONFIRMATION_ALERT_PROMPT_WARNING, CONSENT_REQUEST_PROMPT
-from enterprise.lms_api import CourseApiClient
+from enterprise.lms_api import CourseApiClient, EnrollmentApiClient
 from enterprise.models import (
     EnterpriseCourseEnrollment,
     EnterpriseCustomer,
@@ -53,10 +53,14 @@ from enterprise.utils import (
     consent_necessary_for_course,
     enterprise_login_required,
     filter_audit_course_modes,
+    get_enterprise_customer_or_404,
 )
+from six.moves.urllib.parse import urljoin  # pylint: disable=import-error
 
 
 logger = getLogger(__name__)  # pylint: disable=invalid-name
+
+LMS_COURSE_URL = urljoin(settings.LMS_ROOT_URL, '/courses/{course_id}')
 
 
 def verify_edx_resources():
@@ -498,7 +502,23 @@ class CourseEnrollmentView(View):
             logger.error('Unable to find course details for course ID: %s', course_id)
             raise Http404
 
-        platform_name = configuration_helpers.get_value('PLATFORM_NAME', settings.PLATFORM_NAME)
+        enterprise_customer = get_enterprise_customer_or_404(enterprise_uuid)
+
+        enrollment_client = EnrollmentApiClient()
+        enrolled_courses = enrollment_client.get_enrolled_courses(request.user.username)
+        enrolled_course_ids = [course['course_details']['course_id'] for course in enrolled_courses]
+
+        if (course_id in enrolled_course_ids or
+                EnterpriseCourseEnrollment.objects.filter(
+                    enterprise_customer_user__enterprise_customer=enterprise_customer,
+                    enterprise_customer_user__user_id=request.user.id,
+                    course_id=course_id
+                ).exists()):
+            # The user is already enrolled in the course through the Enterprise Customer, so redirect to the course
+            # info page.
+            return redirect(LMS_COURSE_URL.format(course_id=course_id))
+
+        platform_name = configuration_helpers.get_value("PLATFORM_NAME", settings.PLATFORM_NAME)
         course_start_date = ''
         if course_details['start']:
             course_start_date = parse(course_details['start']).strftime('%B %d, %Y')
--- a/enterprise/lms_api.py
+++ b/enterprise/lms_api.py
@@ -152,6 +152,27 @@ class EnrollmentApiClient(LmsApiClient):
             }
         )
 
+    def get_course_enrollment(self, username, course_id):
+        """
+        Query the enrollment API to get information about a single course enrollment.
+
+        Args:
+            username (str): The username by which the user goes on the OpenEdX platform
+            course_id (str): The string value of the course's unique identifier
+
+        Returns:
+            dict: A dictionary containing details of the enrollment, including course details, mode, username, etc.
+        """
+        endpoint = getattr(
+            self.client.enrollment,
+            '{username},{course_id}'.format(username=username, course_id=course_id)
+        )
+        try:
+            return endpoint.get()
+        except HttpNotFoundError:
+            LOGGER.error('course enrollment details not found for username=%s, course=%s', username, course_id)
+            return None
+
     def get_enrolled_courses(self, username):
         """
         Query the enrollment API to get a list of the courses a user is enrolled in.
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -534,6 +534,26 @@ def filter_audit_course_modes(enterprise_customer, course_modes):
         return course_modes
 
 
+def get_enterprise_customer_or_404(enterprise_uuid):
+    """
+    Given an EnterpriseCustomer UUID, return the corresponding EnterpriseCustomer or raise a 404.
+
+    Arguments:
+        enterprise_uuid (str): The UUID (in string form) of the EnterpriseCustomer to fetch.
+
+    Returns:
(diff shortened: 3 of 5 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=105"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

enterprise/views.py
```diff
@@ -486,6 +487,24 @@ def get_enterprise_course_enrollment_page(self, request, enterprise_uuid, course
...
+        try:
+            # pylint: disable=no-member
+            enterprise_customer_user = EnterpriseCustomerUser.objects.get(
+                user_id=request.user.id,
+                enterprise_customer=enterprise_customer
+            )
+            enterprise_customer_user.enterprise_enrollments.get(course_id=course_id)
+            # The user is already enrolled in the course through the Enterprise Customer, so redirect to the course
+            # info page.
+            return redirect(urljoin(
```
bdero: @haikuginger If the goal for enterprise is to eventually have it hosted as a separate service, shouldn't we write this as if the LMS is a separate service and not use the reverse router for LMS URLs? There might be a more preferred way to do it than the way I did here, but I couldn't think of one or find an example.

enterprise/views.py
```diff
@@ -486,6 +487,24 @@ def get_enterprise_course_enrollment_page(self, request, enterprise_uuid, course
             logger.error('Unable to find enterprise customer for UUID: %s', enterprise_uuid)
             raise Http404
 
+        try:
+            # pylint: disable=no-member
+            enterprise_customer_user = EnterpriseCustomerUser.objects.get(
+                user_id=request.user.id,
+                enterprise_customer=enterprise_customer
+            )
+            enterprise_customer_user.enterprise_enrollments.get(course_id=course_id)
```
bdero: Yeah, this is much better, thanks!

enterprise/views.py
```diff
@@ -486,6 +487,18 @@ def get_enterprise_course_enrollment_page(self, request, enterprise_uuid, course
...
+                enterprise_customer_user__enterprise_customer=enterprise_customer,
+                enterprise_customer_user__user_id=request.user.id,
+                course_id=course_id
+        ).exists():
+            # The user is already enrolled in the course through the Enterprise Customer, so redirect to the course
+            # info page.
+            return redirect(urljoin(
+                settings.LMS_ROOT_URL,
+                '/courses/{course_id}'.format(course_id=course_id)
+            ))
```
douglashall: This URL pattern should be a setting, that way we only need to change references like this in one place should the LMS URL pattern change in the future.

bdero: Hey Douglas, I made this change, although I think this'll need an accompanying PR in `edx-platform` as a result.

bdero: Do you think there's a reasonable way I should write this setting into the enterprise app only (without adding the setting to edx-platform)? It feels really odd adding `LMS_COURSE_URL` to the LMS setting envs, whether using the reverse router to do so or not.

douglashall: Yeah, maybe I'm making this too complicated. Instead of a setting, maybe we just make it a module variable and use it here.

bdero: Alrighty, I changed it to a module var.

enterprise/views.py
```diff
@@ -54,10 +54,13 @@
 )
 from enterprise.tpa_pipeline import active_provider_enforces_data_sharing, get_enterprise_customer_for_request
 from enterprise.utils import NotConnectedToEdX, consent_necessary_for_course, filter_audit_course_modes
+from six.moves.urllib.parse import urljoin  # pylint: disable=import-error
 
 
 logger = getLogger(__name__)  # pylint: disable=invalid-name
 
+lms_course_url = urljoin(settings.LMS_ROOT_URL, '/courses/{course_id}')
```
douglashall: Sorry, one more thing. Can you capitalize `LMS_COURSE_URL`?

enterprise/views.py
```diff
@@ -486,6 +489,15 @@ def get_enterprise_course_enrollment_page(self, request, enterprise_uuid, course
             logger.error('Unable to find enterprise customer for UUID: %s', enterprise_uuid)
             raise Http404
 
+        if EnterpriseCourseEnrollment.objects.filter(
+                enterprise_customer_user__enterprise_customer=enterprise_customer,
+                enterprise_customer_user__user_id=request.user.id,
+                course_id=course_id
+        ).exists():
```
douglashall: What is the user is enrolled in this course, just not through the `EnterpriseCustomer`?

haikuginger: Then the user is arriving at the Enterprise enrollment page with a specific course in mind (i.e., they're going to it via an employer LMS), but their enrollment hasn't been tied to the Enterprise. By looking specifically for an Enterprise enrollment, we're able to make the enrollment/Enterprise connection to existing enrollments. I think that's the correct behavior in this case.

douglashall: But if they are already enrolled in the course, I don't think we would want to send them to the enterprise course enrollment landing page right?

bdero: @douglashall That's a good question - I'm not sure if this case was decided on, so I made a comment on the upstream ticket asking about what the behavior should be: https://openedx.atlassian.net/browse/ENT-355?focusedCommentId=263107&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-263107

douglashall: Thanks @bdero! I'll make sure we get an answer from Griff today.

enterprise/views.py
bdero: Yup, pylint caught this as well. I've addressed this, thanks!

enterprise/views.py
bdero: Thanks @douglashall :+1:

enterprise/views.py
```diff
@@ -58,11 +58,15 @@
...
     enterprise_login_required,
     filter_audit_course_modes,
+    get_enterprise_customer_or_404,
 )
+from six.moves.urllib.parse import urljoin  # pylint: disable=import-error
 
 
 logger = getLogger(__name__)  # pylint: disable=invalid-name
 
+LMS_COURSE_URL = urljoin(settings.LMS_ROOT_URL, '/courses/{course_id}')
```
haikuginger: Small nit; if there are no objections, can we change this out to the following?

```python
LMS_COURSE_URL = urljoin(settings.LMS_ROOT_URL, '/courses/{course_id}/courseware')
```

That'll take the learner to where they were in the course, rather than to the course homepage.

If anyone has any contrary thoughts, though, let's merge as-is, and have a discussion on the appropriate thing to do.

enterprise/views.py
douglashall: @bdero I think we are still waiting for an update here to account for learners that are enrolled in the course via a non-enterprise pathway. Is that right?

bdero: @douglashall Yup, I've added this functionality in and this should be good to go. It looks like there's a build delay issue on travis right now though, so I'm just waiting for the tests to go through now.

douglashall: Thanks @bdero!

enterprise/views.py
```diff
@@ -499,7 +503,23 @@ def get_enterprise_course_enrollment_page(self, request, enterprise_uuid, course
             logger.error('Unable to find course details for course ID: %s', course_id)
             raise Http404
 
-        platform_name = configuration_helpers.get_value('PLATFORM_NAME', settings.PLATFORM_NAME)
+        enterprise_customer = get_enterprise_customer_or_404(enterprise_uuid)
+
+        enrollment_client = EnrollmentApiClient()
+        enrolled_courses = enrollment_client.get_enrolled_courses(request.user.username)
```
douglashall: Can we use [this](https://github.com/edx/edx-platform/blob/master/common/djangoapps/enrollment/views.py#L86) enrollment API endpoint instead so that we don't have to retrieve all of the learner's enrollments?

bdero: Good idea - I've added this endpoint to the EnrollmentApiClient and we're now using that instead.

enterprise/views.py
bdero: I've made this change. :+1:

enterprise/views.py
```diff
@@ -499,7 +503,23 @@ def get_enterprise_course_enrollment_page(self, request, enterprise_uuid, course
             logger.error('Unable to find course details for course ID: %s', course_id)
             raise Http404
 
-        platform_name = configuration_helpers.get_value('PLATFORM_NAME', settings.PLATFORM_NAME)
+        enterprise_customer = get_enterprise_customer_or_404(enterprise_uuid)
+
+        enrollment_client = EnrollmentApiClient()
+        enrolled_course = enrollment_client.get_course_enrollment(request.user.username, course_id)
+        enrolled_course_id = enrolled_course['course_details']['course_id']
```
douglashall: I think this will blow up if the user is not enrolled in the given course right? Can we do `enrolled_course.get('course_details', {}).get('course_id')`? And then just `if (enrolled_course_id or ...` below?

douglashall: It looks like you also need to account for `enrolled_course` being `None`.

bdero: Actually, we only need to check if the value returned is `None` in this case to prove enrollment. I've adjusted this, thanks.
Most likely there will be a couple more iterations of test failure before this is mergable as well; my test suite is currently not working locally.

douglashall: No worries. Thanks for persisting.

bdero:
@douglashall Ok, this should be all set now! Could you take one more look and let me know if you think this is good to merge?

douglashall:
@bdero Ready for merge! Thanks.

PR #107, opened 2017-05-15
[WIP] ENT-356 Update enterprise api and SAP course export with landing page url

brittneyexline:
**Description:** There are several things going on in this PR:
1. Adding the landing page url to the course data responses from the enterprise api.
2. Adding a new endpoint to the enterprise api to get courses by enterprise uuid. I added this because the courses endpoint by catalog id inferred the enterprise by whether or not the request's user was linked to an enterprise, which wasn't sufficient if we were to switch over the course export job to use the enterprise api.
3. Switch over the course export job to use the new enterprise courses endpoint, and to use the landing page url returned in the response as the launchURL to SAP.

**JIRA:** https://openedx.atlassian.net/browse/ENT-356

**Dependencies:** n/a

**Merge deadline:** 5/17/2017 would be nice

**Installation instructions:** Standard enterprise installation/set up.

**Testing instructions:**

TBD - I haven't tested this on a sandbox yet.

**Merge checklist:**

- [ ] Regenerate requirements with `make upgrade && make requirements` (and make sure to fix any errors).
  **DO NOT** just add dependencies to `requirements/*.txt` files.
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Documentation updated (not only docstrings)
- [ ] Commits are (reasonably) squashed
- [ ] Translations are updated
- [ ] PR author is listed in AUTHORS

**Post merge:**
- [ ] Create a tag
- [ ] Check new version is pushed to PyPi after tag-triggered build is finished.
- [ ] Delete working branch (if not needed anymore)

**Author concerns:** List any concerns about this PR - inelegant 
solutions, hacks, quick-and-dirty implementations, concerns about 
migrations, etc.

PR #121, opened 2017-06-02
ENT-356 Simple version of using the new landing page url

brittneyexline:
**Description:** This updates both the enterprise api and SAP course export to use the new
landing page url, but without transitioning the SAP course export to use
the enterprise api endpoint.

**JIRA:** https://openedx.atlassian.net/browse/ENT-356

**Dependencies:** n/a, though it is related to https://github.com/edx/edx-enterprise/pull/107

**Merge deadline:** 

**Testing instructions:**

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -11,10 +11,12 @@ from uuid import uuid4
 import six
 from simple_history.models import HistoricalRecords
 
+from django.conf import settings
 from django.contrib.auth.models import User
 from django.contrib.sites.models import Site
 from django.core.exceptions import ObjectDoesNotExist
 from django.core.files.storage import default_storage
+from django.core.urlresolvers import reverse
 from django.db import models
 from django.template import Context, Template
 from django.utils.encoding import python_2_unicode_compatible
@@ -26,7 +28,14 @@ from model_utils.models import TimeStampedModel
 
 from enterprise import utils
 from enterprise.lms_api import ThirdPartyAuthApiClient, enroll_user_in_course_locally
+from enterprise.utils import NotConnectedToOpenEdX
 from enterprise.validators import validate_image_extension, validate_image_size
+from six.moves.urllib.parse import urljoin  # pylint: disable=import-error,ungrouped-imports
+
+try:
+    from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
+except ImportError:
+    configuration_helpers = None
 
 logger = getLogger(__name__)  # pylint: disable=invalid-name
 
@@ -168,6 +177,28 @@ class EnterpriseCustomer(TimeStampedModel):
         Determine whether the enterprise customer has enabled the data sharing consent request.
         """
         return self.enable_data_sharing_consent and self.enforce_data_sharing_consent != self.EXTERNALLY_MANAGED
+
+    def get_course_enrollment_url(self, course_run_key):
+        """
+        Return enterprise landing page url for the given course.
+
+        Arguments:
+            course_run_key (str): The course run id for the course to be displayed.
+        Returns:
+            (str): Enterprise landing page url.
+        """
+        if configuration_helpers is None:
+            raise NotConnectedToOpenEdX(
+                _("This package must be installed in an EdX environment to look up configuration.")
+            )
+
+        return urljoin(
+            configuration_helpers.get_value('LMS_ROOT_URL', settings.LMS_ROOT_URL),
+            reverse(
+                'enterprise_course_enrollment_page',
+                kwargs={'enterprise_uuid': self.uuid, 'course_id': course_run_key}
+            )
+        )
 
 
 class EnterpriseCustomerUserManager(models.Manager):
--- a/integrated_channels/sap_success_factors/utils.py
+++ b/integrated_channels/sap_success_factors/utils.py
@@ -17,6 +17,11 @@ from enterprise.lms_api import parse_lms_api_datetime
 from enterprise.utils import safe_extract_key
 from integrated_channels.integrated_channel.course_metadata import BaseCourseExporter
 
+try:
+    from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
+except ImportError:
+    configuration_helpers = None
+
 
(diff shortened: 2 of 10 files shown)


integrated_channels/sap_success_factors/utils.py
```diff
@@ -253,7 +253,8 @@ def resolve_removed_courses(self, previous_audit_summary):
                     'sap_success_factors',
                     'SAPSuccessFactorsGlobalConfiguration'
                 ).current().provider_id,
-                'launchURL': get_course_track_selection_url(x['enterprise_customer'], x['key']),
+                'launchURL': (get_enterprise_course_enrollment_url(x['enterprise_customer'], x['key']) or
+                              get_course_track_selection_url(x['enterprise_customer'], x['key'])),
```
douglashall: I'm wondering if we should waffle switch this. I don't think you would ever get to the second clause of the `or` statement here would you?

douglashall: Oh nevermind, I just looked below. Still wondering if we should waffle though.

brittneyexline: I've actually thought of making this a setting for the Enterprise Customer, if some Enterprises might want the track selection page instead... but I suppose an overarching flag would make more sense for transitioning in case the page isn't ready. We can talk more Monday.

enterprise/models.py
```diff
@@ -9,12 +9,15 @@
 from uuid import uuid4
 
 import six
+from requests.compat import urljoin
```
haikuginger: I'd prefer to import `urljoin` from `six.moves.urllib.parse`; `requests.compat` is an internal module for use specifically inside `requests`, whereas `six` is all about providing cross-Python-version compatibility.

enterprise/models.py
```diff
@@ -169,6 +178,28 @@ def requests_data_sharing_consent(self):
...
+    def get_course_enrollment_url(self, course_run_key):
+        """
+        Return enterprise landing page url for the given course.
+
+        Arguments:
+            course_run_key (str): The course run id for the course to be displayed.
+        Returns:
+            (str): Enterprise landing page url.
+        """
+        if configuration_helpers is None:
```
haikuginger: The more we use this pattern (which I created when moving consent stuff over initially), the less happy I am with it. We're going to have to deal with it eventually. Not here, but eventually.

enterprise/models.py
```diff
@@ -169,6 +178,28 @@ def requests_data_sharing_consent(self):
...
+            raise NotConnectedToOpenEdX(
+                _("This package must be installed in an EdX environment to look up configuration.")
+            )
+
+        return urljoin(
+            configuration_helpers.get_value('LMS_ROOT_URL', settings.LMS_ROOT_URL),
```
haikuginger: I'm thinking we might want some additional logic. We're assuming here that Enterprise will be installed inside the LMS, but that may not always be the case. We may want to create an additional variable `ENTERPRISE_PUBLIC_URL_ROOT` and default to `LMS_ROOT_URL`. I don't think we need to stick that variable into `edx-platform` or `configuration` yet, though; it's a long road to IDA-ness.

haikuginger: Actually, since we're using `configuration_helpers`, the above comment is somewhat irrelevant, since this codepath _can't_ operate without `edx-platform`, but I'm gonna leave it for posterity anyway.

integrated_channels/sap_success_factors/utils.py
```diff
@@ -275,6 +276,21 @@ def resolve_removed_courses(self, previous_audit_summary):
...
+    Given an EnterpriseCustomer and a course ID, craft a URL that links to the enterprise enrollment landing
+     page for that course.
+
+    Args:
+        enterprise_customer (EnterpriseCustomer): The EnterpriseCustomer that a URL needs to be built for
+        course_id (str): The string identifier of the course in question
+    """
+    try:
+        return enterprise_customer.get_course_enrollment_url(course_id)
+    except NotConnectedToOpenEdX:
```
haikuginger: The fact that we have to catch this for a configuration check is incredibly frustrating to me. I've filed [ENT-428](https://openedx.atlassian.net/browse/ENT-428) to handle this and make for a happier overall path.

integrated_channels/sap_success_factors/utils.py
```diff
@@ -253,7 +258,10 @@ def resolve_removed_courses(self, previous_audit_summary):
                     'sap_success_factors',
                     'SAPSuccessFactorsGlobalConfiguration'
                 ).current().provider_id,
-                'launchURL': get_course_track_selection_url(x['enterprise_customer'], x['key']),
+                'launchURL': (get_enterprise_course_enrollment_url(x['enterprise_customer'], x['key']) if
+                              configuration_helpers and
+                              configuration_helpers.get_value('SAP_USE_ENTERPRISE_ENROLLMENT_PAGE') else
+                              get_course_track_selection_url(x['enterprise_customer'], x['key'])),
```
douglashall: Could you move this to a wrapper function to improve readability?

brittneyexline:
@douglashall @haikuginger can i get a final ok on this?

douglashall:
@brittneyexline Ready for merge! Thanks.

haikuginger:
Hey, one sec here - it looks like the EnterpriseCustomer metadata tagged onto the catalog is selected by seeing which EnterpriseCustomer the _requesting user_ is attached to - is that what we want? When I request the endpoint with a non-attached user, I get a 500 error.

brittneyexline:
@haikuginger I discovered this from my first iteration, which is why that PR introduces a new version of the endpoint, because we need to be able to hit it with a worker user with admin privileges if we switch SAP course export to use it. I talked to Matt and he agreed that for enterprise customers using the endpoint, we would link their user to the enterprise.

haikuginger:
👍

PR #124, opened 2017-06-07
[ENT-365] Move from old python-social-auth to latest, split PSA

UmanShahzad:
**Description:** Old [python-social-auth](https://github.com/python-social-auth) was a monolithic repository and the version we utilize is almost a year old. We are looking to upgrade to the latest, split version of python-social-auth which is more modular, and provides several new features, including a configurable maximum session length, a DB-packed partial pipeline, utilizing the latest backend (provider) versions, and so on.

**JIRA:** Implements [ENT-365](https://openedx.atlassian.net/browse/ENT-365).

**Dependencies:** [Complementary PR at edx-platform](https://github.com/edx/edx-platform/pull/15135).

**Merge deadline:** Mostly handles technical debt, so not that urgent.

**Installation instructions:** N/A for now.

**Testing instructions:** See [edx-platform PR](https://github.com/edx/edx-platform/pull/15135) for now.

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:** This **must** be merged at the **same time** as the corresponding [edx-platform PR](https://github.com/edx/edx-platform/pull/15135), or else things will break.

### DIFF
--- a/enterprise/tpa_pipeline.py
+++ b/enterprise/tpa_pipeline.py
@@ -10,7 +10,7 @@ from django.shortcuts import redirect
 from django.utils.translation import ugettext as _
 
 from enterprise.models import EnterpriseCustomer, EnterpriseCustomerUser, UserDataSharingConsentAudit
-from enterprise.utils import NotConnectedToEdX
+from enterprise.utils import NotConnectedToOpenEdX
 
 try:
     from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
@@ -18,9 +18,14 @@ except ImportError:
     configuration_helpers = None
 
 try:
-    from social.pipeline.partial import partial
+    from social_core.pipeline.partial import partial
 except ImportError:
     from enterprise.utils import null_decorator as partial  # pylint:disable=ungrouped-imports
+
+try:
+    from third_party_auth.pipeline import get as get_partial_pipeline
+except ImportError:
+    get_partial_pipeline = None
 
 try:
     from third_party_auth.provider import Registry
@@ -28,23 +33,38 @@ except ImportError:
     Registry = None
 
 
+def verify_third_party_auth_dependencies():
+    """
+    Ensure that all necessary third_party_auth dependencies are present.
+    """
+    third_party_dependencies = {
+        'Registry': Registry,
+        'get_partial_pipeline': get_partial_pipeline,
+    }
+
+    for dependency in third_party_dependencies:
+        if third_party_dependencies[dependency] is None:
+            raise NotConnectedToOpenEdX(
+                _("The following dependency must be installed in an Open edX environment to look up third-party-auth "
+                  "dependencies.")
+                + "\nUnavailable: {dependency}".format(dependency=dependency)
+            )
+
+
 def get_enterprise_customer_for_request(request):
     """
     Get the EnterpriseCustomer associated with a particular request.
     """
-    pipeline = request.session.get('partial_pipeline')
-    return get_ec_for_running_pipeline(pipeline)
+    verify_third_party_auth_dependencies()
+    pipeline = get_partial_pipeline(request)
+    return get_enterprise_customer_for_running_pipeline(pipeline)
 
 
-def get_ec_for_running_pipeline(pipeline):
+def get_enterprise_customer_for_running_pipeline(pipeline):  # pylint: disable=invalid-name
     """
     Get the EnterpriseCustomer associated with a running pipeline.
     """
-    if Registry is None:
-        raise NotConnectedToEdX(
-            _("This package must be installed in an EdX environment to look up third-party auth providers.")
-        )
-
+    verify_third_party_auth_dependencies()
     if pipeline is None:
         return None
     provider = Registry.get_from_pipeline(pipeline)
@@ -78,7 +98,7 @@ def active_provider_enforces_data_sharing(request, enforcement_location):
         enforcement_location (str): the point where to see data sharing consent state.
         argument can either be "optional", 'at_login' or 'at_enrollment'
     """
-    running_pipeline = request.session.get('partial_pipeline')
+    running_pipeline = request.session.get('partial_pipeline_token')
     if running_pipeline:
         customer = get_enterprise_customer_for_request(request)
         return customer and customer.enforces_data_sharing_consent(enforcement_location)
@@ -89,7 +109,7 @@ def active_provider_requests_data_sharing(request):
     """
     Determine if the active EnterpriseCustomer requests data sharing consent.
     """
-    running_pipeline = request.session.get('partial_pipeline')
+    running_pipeline = request.session.get('partial_pipeline_token')
     if running_pipeline:
         customer = get_enterprise_customer_for_request(request)
         return customer and customer.requests_data_sharing_consent
@@ -100,7 +120,7 @@ def get_consent_status_for_pipeline(pipeline):
     """
     Get the consent object for the current pipeline.
     """
-    customer = get_ec_for_running_pipeline(pipeline)
+    customer = get_enterprise_customer_for_running_pipeline(pipeline)
     user = pipeline.kwargs['user']
     try:
         return UserDataSharingConsentAudit.objects.get(
@@ -141,7 +161,7 @@ def handle_enterprise_logistration(backend, user, **kwargs):
         """
         return redirect(reverse('grant_data_sharing_permissions'))
 
-    enterprise_customer = get_ec_for_running_pipeline({'backend': backend.name, 'kwargs': kwargs})
+    enterprise_customer = get_enterprise_customer_for_running_pipeline({'backend': backend.name, 'kwargs': kwargs})
     if enterprise_customer is None:
         # This pipeline element is not being activated as a part of an Enterprise logistration
         return
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -34,7 +34,7 @@ except ImportError:
 LOGGER = logging.getLogger(__name__)
 
 
-class NotConnectedToEdX(Exception):
+class NotConnectedToOpenEdX(Exception):
     """
     Exception to raise when not connected to OpenEdX.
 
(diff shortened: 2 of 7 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=124&repo=edx%2Fedx-enterprise"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

tests/test_views.py
```diff
@@ -202,26 +197,65 @@ def test_get_render_patched(
...
+        )
+        expected_context = {
+            'consent_request_prompt': expected_prompt,
+            'confirmation_alert_prompt': expected_alert,
+            'confirmation_alert_prompt_warning': expected_warning,
+            'platform_name': 'This Platform',
+            'enterprise_customer_name': 'Fake Customer Name',
+        }
+        for key, value in expected_context.items():
+            assert response.context[key] == value  # pylint: disable=no-member
```
UmanShahzad: The above function is the only significant change in this file -- the rest is just adding some necessary mock patches and refactoring to make the file a little shorter and more consistent.

UmanShahzad:
@douglashall and @brittneyexline -- this PR, along with https://github.com/edx/edx-platform/pull/15135, is ready for review. Note that we'll have to merge this one first, and at the same time update the edx-platform PR to use this latest edx-enterprise version, and then merge the latter.

UmanShahzad:
@brittneyexline @douglashall Just a reminder -- we've got the edx-platform PR ready to go (upstream PSA PR is merged!), so only this needs review and the former needs approval, then we're good to go. Some OpenCraft dev will also be looking at this shortly.

enterprise/tpa_pipeline.py
```diff
@@ -10,41 +10,58 @@
...
+    Ensure that all necessary third_party_auth dependencies are present.
+    """
+    third_party_dependencies = (
+        Registry,
+        get_partial_pipeline,
+    )
+
+    if any(third_party_dependency is None for third_party_dependency in third_party_dependencies):
+        raise NotConnectedToOpenEdX(
+            _("This package must be installed in an Open edX environment to look up third-party auth dependencies.")
```
brittneyexline: Maybe this message should be updated to say what package needs to be installed?

enterprise/utils.py
```diff
@@ -49,19 +49,7 @@ def __init__(self, *args, **kwargs):
...
-        super(NotConnectedToEdX, self).__init__(*args, **kwargs)
-
-
-class NotConnectedToOpenEdX(Exception):
-    """
-    Exception to raise when we weren't able to import needed resources.
-
-    This is raised when we try to use the resources, rather than on
-    import, so that during testing we can load the module
-    and mock out the resources we need.
```
brittneyexline: Just curious about why we decided to consolidate these exceptions now? Not that I'm not in favor - I've never fully understood the difference between them.

enterprise/views.py
```diff
@@ -74,10 +76,11 @@ def verify_edx_resources():
         get_enterprise_customer_for_request,
         get_real_social_auth_object,
         quarantine_session,
-        lift_quarantine
+        lift_quarantine,
+        get_partial_pipeline,
     )
     if any(method is None for method in required_methods):
-        raise NotConnectedToEdX(_('Methods in the Open edX platform necessary for this view are not available.'))
+        raise NotConnectedToOpenEdX(_('Methods in the Open edX platform necessary for this view are not available.'))
```
brittneyexline: also worth updating the message here to be more specific about the missing dependencies.

enterprise/tpa_pipeline.py
UmanShahzad: Sure! :+1:

enterprise/utils.py
UmanShahzad: To tell you the truth, this PR was picked up from @haikuginger, and this change was a part of it, so I decided to keep it going. It doesn't look like to me the changes are exactly that significant, it's just a squashing of two exception classes to maybe simplify what it is we work with.

I guess its original form could be considered a sort of minor tech debt, and it would have been better to eventually clean it up than to simply ignore it up until a dedicated PR. Since it's here and not really bothering anyone, I suppose we can continue with it, is that okay? If not, I'll open a dedicated PR to do it but will take more work. Either that or I can separate the commits for this change and for the rest in this PR.

enterprise/views.py
UmanShahzad: Yep, easy peasy. :+1:

enterprise/utils.py
brittneyexline: Yea, I'm in favor of keeping it in this PR, I just wanted to call it out since it wasn't essential to the purpose of the PR.

UmanShahzad:
@brittneyexline Ok, I had to jump some hoops to get it to display the missing method/dependency name properly, including dealing with tests looking explicitly for matching error messages ( :| ) and a significant rebase, but it's done now.

@haikuginger should, hopefully, be able to review this soon.

When approved I want to squash this.

enterprise/views.py
```diff
@@ -75,15 +77,21 @@ def verify_edx_resources():
...
-    if any(method is None for method in required_methods):
-        raise NotConnectedToEdX(_('Methods in the Open edX platform necessary for this view are not available.'))
+    required_methods = {
+        'configuration_helpers': configuration_helpers,
+        'get_enterprise_customer_for_request': get_enterprise_customer_for_request,
+        'get_real_social_auth_object': get_real_social_auth_object,
+        'quarantine_session': quarantine_session,
+        'lift_quarantine': lift_quarantine,
+        'get_partial_pipeline': get_partial_pipeline,
+    }
```
bradenmacdonald: Nit (which means this is optional and you can ignore): instead of duplicating all the imports here, you could just change the code at the top of this file to say

```python
import_error = None

try:
    from ... import ...
except ImportError as e:
    import_error = import_error or e.message
```

Then you can just display `import_error` in the error message below:
```python
def verify_edx_resources():
    if import_error:
        raise NotConnectedToOpenEdX("An Open edX dependency required for this method is missing: {}".format(import_error))
```

Or another option is to just log a warning message within the `except ImportError` block itself, and don't bother calling `verify_edx_resources()` elsewhere.

enterprise/tpa_pipeline.py
```diff
@@ -78,7 +98,10 @@ def active_provider_enforces_data_sharing(request, enforcement_location):
         enforcement_location (str): the point where to see data sharing consent state.
         argument can either be "optional", 'at_login' or 'at_enrollment'
     """
-    running_pipeline = request.session.get('partial_pipeline')
+    try:
+        running_pipeline = get_partial_pipeline(request)
+    except TypeError:
+        running_pipeline = request.session.get('partial_pipeline_token')
```
bradenmacdonald: What are the circumstances where `get_partial_pipeline()` does not work but `running_pipeline = request.session.get('partial_pipeline_token')` does? And aren't those different types? (`get_partial_pipeline` returns a `dict` but `request.session.get('partial_pipeline_token')` returns a `str`)

If this is for a test case, it should be mocked out or replaced with a fake (install a simple `third_party_auth` fake app in the test environment, full of `Mock`s), rather than being here in the regular code.

UmanShahzad: @bradenmacdonald I think this is to protect against the case when we're "NotConnectedToOpenEdX". I'm not sure if it's better to check for the dependency and remove the try-catch block, or to keep it going like this.

enterprise/views.py
UmanShahzad: @bradenmacdonald I actually agree with the above solutions, but I've been beating this exception bush quite a bit in this PR and I think I want to just dedicate a new, smaller PR with a cleaner approach to this; for now it'll do, if that's alright -- but definitely want to change it.

enterprise/tpa_pipeline.py
bradenmacdonald: Hmm, generally code specific to tests should be kept in the `_test` files as much as possible, so that it's clear what is normal/expected behavior and what are the mocks/hacks/workarounds put in just for tests, which can be removed/refactored anytime.

In this case, can we just change the test code to mock out `get_partial_pipeline` rather than having this `try`/`except`?

enterprise/views.py
bradenmacdonald: Makes sense - fine by me to keep that for another PR!

enterprise/tpa_pipeline.py
UmanShahzad: @bradenmacdonald The issue is actually that the test code where this fails _is_ mocking out `get_partial_pipeline`.

Do you think it's okay to just leave it as `request.session.get('partial_pipeline_token')`? In this function we're just checking for whether the pipeline is running and nothing more, so what type is returned doesn't really matter as long as it indicates to us that the pipeline is indeed running, which the token should be enough for?

UmanShahzad:
@bradenmacdonald I gave a reply on the try-catch block issue. Not sure if this'll be enough, but let me know.

haikuginger:
@UmanShahzad, I ran into issues while testing this - some in the edx-platform PR, some here. Specifically, I had issues with the quarantine stuff dropping my session after submitting data sharing consent. LMK if you need more details on that.

enterprise/tpa_pipeline.py
bradenmacdonald: Yeah that makes sense to me.

UmanShahzad:
@haikuginger Any idea of how to reproduce the error?

enterprise/tpa_pipeline.py
```diff
@@ -78,7 +98,7 @@ def active_provider_enforces_data_sharing(request, enforcement_location):
         enforcement_location (str): the point where to see data sharing consent state.
         argument can either be "optional", 'at_login' or 'at_enrollment'
     """
-    running_pipeline = request.session.get('partial_pipeline')
+    running_pipeline = request.session.get('partial_pipeline_token')
```
UmanShahzad: @haikuginger I would say it's sufficient: we use the token to get the pipeline anyway, and getting the pipeline is thus wholly dependent on that. If the token doesn't exist, we wouldn't have gotten the pipeline anyway.

UmanShahzad:
@haikuginger Thanks! Good to know the bug wasn't related. I answered your review question, hopefully it's sufficient. I'll test on Android, and if all's good (it should be since iOS/Android are related in this respect as @bradenmacdonald mentioned), I'll fix these branch conflicts and give you a ping.

haikuginger:
@UmanShahzad, can you rebase this and update the version to 0.36.0? This is definitely worthy of a bump, and it looks like 0.35.2 was recently merged.

UmanShahzad:
@haikuginger Done, and rebased.

PR #125, opened 2017-06-08
ENT-429: Add query string parameter mappings to API Gateway configuration

mattdrayer:
**Description:** The query string parameters have been defined in the API Gateway configuration, but the corresponding mappings were not including.  Because of this, the API Gateway was dropping the parameters from the request.

**JIRA:** https://openedx.atlassian.net/browse/ENT-429

**Dependencies:** None

**Merge deadline:** Soon, hopefully

**Installation instructions:** Set up an AWS API Gateway, import configuration -- it's a bit of a PITA but after merge we'll be able to test on our stage gateway by updating the reference in the api-manager repository.

**Testing instructions:**

After updating API Gateway, should be able to use Postman and request alternate size result sets as well as "next" pages (via offset).

**Merge checklist:**

~- [ ] Regenerate requirements with `make upgrade && make requirements` (and make sure to fix any errors).~
- [ ] All reviewers approved
- [ ] CI build is green
~- [ ] Version bumped~
- [ ] Changelog record added
~- [ ] Documentation updated (not only docstrings)~
~- [ ] Translations are updated~

**Post merge:**
~- [ ] Create a tag~
~- [ ] Check new version is pushed to PyPi after tag-triggered build is finished.~

**Author concerns:** None, this should be a useful update once we have it configured correctly.


mattdrayer:
Okay, I think I've finally got this worked out on a test gateway:

https://20enjrmi05.execute-api.us-west-2.amazonaws.com/edx_api/enterprise/v1/catalogs/124/courses?limit=100

There appears to be an issue now with the Enterprise Service actually handling the query parameters, but that's a separate thing vs. the API Gateway configuration to pass the params, and I've logged [a separate ticket](https://openedx.atlassian.net/browse/ENT-442) for that.

api-compact.yaml
```diff
@@ -23,9 +23,9 @@ apigateway_responses: &apigateway_responses
   404:
     statusCode: "404"
   429:
-      statusCode: "429"
+    statusCode: "429"
```
douglashall: Is it AWS that expects these to be strings?

api-compact.yaml
```diff
@@ -60,22 +60,22 @@ auth_header: &auth_header
   type: "string"
 
 limit_qs_parameter: &limit_qs_parameter
-  in: query
-  name: limit
+  in: "query"
+  name: "limit"
```
douglashall: Why do we need to quote these strings?

api.yaml
```diff
@@ -205,3 +216,109 @@ endpoints:
...
+      statusCode: "403"
+    404:
+      statusCode: "404"
+    429:
+      statusCode: "429"
+    500:
+      statusCode: "500"
+    default:
+      statusCode: "400"
+  type: http
```
douglashall: It would be good to use [aliases](https://github.com/cyklo/Bukkit-OtherBlocks/wiki/Aliases-(advanced-YAML-usage)) to reduce some of the duplication here. It totally just googled that link though so let me know if that is not a real thing.

api-compact.yaml
mattdrayer: To be like everyone else!  There's a good chance that these values do not need to be quoted, although the examples I've seen have the values quoted, including the prior art in our own code.  I'll try it with unquoted values to see what happens.

douglashall: Nevermind, I didn't realize there was already a convention which the existing code was straying from.

api.yaml
mattdrayer: all of the stuff in api.yaml is codegen from www.yamllint.com -- the only file we actually manage is api-compact.yaml

api-compact.yaml
mattdrayer: :man_shrugging: -- just following-suit with the rest of the block

api.yaml
douglashall: Got it. Makes sense.

api-compact.yaml
mattdrayer: actually all I did here was remove some extra tab spaces

PR #127, opened 2017-06-09
ENT-296: Enable account-level consent requirement

haikuginger:
**Description:**

At present, users are prompted to consent to the exact scope of consent that's relevant to the consent gate that they're attempting to pass - for example, if attempting to access a consent-gated courses, they'll be prompted for course consent; if attempting to access a consent-gated SSO provider, they'll be prompted for account consent.

This change adds a feature whereby, on a per-Enterprise Customer basis, a setting can be made that will force users to be prompted for account-level consent rather than course-specific consent. This change also patches the `EnterpriseCourseEnrollment` save method such that when a record is being saved for an EnterpriseCustomer with the above flag set, an account-wide consent audit record is written.

**JIRA:** [ENT-296](https://openedx.atlassian.net/browse/ENT-296)

**Dependencies:** None

**Sandbox:** https://ent296.opencraft.hosting/

**Merge deadline:** 18 June 2017

**Installation instructions:** No special installation steps; standard Enterprise installation

**Testing instructions:**

1. In [the Enterprise admin panel](https://ent296.opencraft.hosting/admin/enterprise/enterprisecustomer/fcd2b0e7-c687-4e19-af64-8889d835c15f/manage_learners), proxy-enroll a user in two courses
2. Ensure that the [EnterpriseCustomer is set to](https://ent296.opencraft.hosting/admin/enterprise/enterprisecustomer/fcd2b0e7-c687-4e19-af64-8889d835c15f/)
    1. require consent at enrollment
    2. *not* require account-level consent
3. As a user, browse to [one of](https://ent296.opencraft.hosting/courses/course-v1:OpenCraft+Secondary+Run1/courseware) [the courses](https://ent296.opencraft.hosting/courses/course-v1:edX+DemoX+Demo_Course/courseware) that you were enrolled in.
4. Observe the consent prompt; note that the messaging is specific to the single course you're attempting to access. Do not provide consent at this time.
5. Ensure that [EnterpriseCustomer is set to](https://ent296.opencraft.hosting/admin/enterprise/enterprisecustomer/fcd2b0e7-c687-4e19-af64-8889d835c15f/) require account-level consent.
6. Again, browse to one of the courses.
7. Observe that the messaging has changed, and is distinctly supportive of the premise that consent will be given for the entire account.
8. Provide consent.
9. Navigate to the dashboard and click on the _other_ course; note that you're taken directly to it without being prompted for consent.
[...]

PR #128, opened 2017-06-14
Only redirect to data sharing consent page if DSC is required at login.

douglashall:
**Description:** Currently, if a user logs in using SSO, we will always show the DSC page directly after account creation/login even if the EntepriseCustomer is configured with an "Enforce data sharing consent" value of "At Enrollment". This modifies the enterprise TPA pipeline to check the "Enforce data sharing consent" value before deciding whether or not to redirect to the DSC page.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/tpa_pipeline.py
+++ b/enterprise/tpa_pipeline.py
@@ -184,15 +184,16 @@ def handle_enterprise_logistration(backend, user, **kwargs):
 
         return
 
-    try:
-        # Find an existing account-level consent record for the user
-        consent = UserDataSharingConsentAudit.objects.get(
-            user__user_id=user.id,
-            user__enterprise_customer=enterprise_customer,
-        )
-    except UserDataSharingConsentAudit.DoesNotExist:
-        return redirect_to_consent()
+    if enterprise_customer.enforces_data_sharing_consent(EnterpriseCustomer.AT_LOGIN):
+        try:
+            # Find an existing account-level consent record for the user
+            consent = UserDataSharingConsentAudit.objects.get(
+                user__user_id=user.id,
+                user__enterprise_customer=enterprise_customer,
+            )
+        except UserDataSharingConsentAudit.DoesNotExist:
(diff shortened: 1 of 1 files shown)


douglashall:
@haikuginger @zubair-arbi What will this break? Is there a reason we didn't write the logic this way to begin with?

zubair-arbi:
@douglashall Since currently SSO flow is the only way where we take account level DSC, so only issue I can think of is that with these changes we will not be able to ask for account level DSC from leaners with enterprise customer's configuration of `"Enforce data sharing consent"` with value `"At Enrollment"`. 
@saleem-latif has sent email with the current behaviour for the DSC.

zrana:
One more thing is that the option for `"Enforce data sharing consent"` with value of `"Optional"` for enterprise customer's configuration is going to be [deprecated as mentioned in this comment](https://openedx.atlassian.net/browse/ENT-280?focusedCommentId=251554&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-251554) by @griffresch.

douglashall:
@haikuginger Can you take a look at this when you get a chance? What if we did this instead of ENT-452? Along with https://github.com/edx/edx-enterprise/pull/127, I think this would allow us to preserve the option to show DSC directly after login/registration while also allowing us to configure the other desired workflows. What do you think?

haikuginger:
@douglashall, I think @pomegranited's PR  at #131 is the right way to go here; it's affecting the same conditional, but it allows for us to link the EnterpriseCustomer to the User when we have course-level consent requested. In comparison, this would skip the linkage entirely for an EnterpriseCustomerUser arriving via SSO for whom course-level gating is enabled. I'm going to go ahead and close this.

PR #129, opened 2017-06-19
ENT-455 show consent decline notification on enterprise course enrollment page

zubair-arbi:
@saleem-latif @asadiqbal08 @douglashall 

**Description:** Redirect learner to enterprise course enrollment page after declining consent and display consent decline notification.

**JIRA:** [ENT-455](https://openedx.atlassian.net/browse/ENT-455)

**Dependencies:** N/A

**Merge deadline:** 30 June, 2017

**Installation instructions:** Install `edx-enterprise` from this branch `zub/ENT-455-ent-landing-page-dsc-decline` in `edx-platform`.

**Testing instructions:**

1. Go to enterprise customers django admin at `admin/enterprise/enterprisecustomer/`
2. Add enterprise customer with logo and with consent enabled with option `At Enrollment` and link it with an `idp`.
3. Create/publish a course in studio and create related modes in Ecom course admin tool.
4. As an anonymous user view the enterprise landing page for the enterprise course by accessing the enterprise course enrollment url with enterprise customer uuid from step no 2, e.g: http://zubair.localhost:8000/enterprise/72416e52-8c77-4860-9584-15e5b06220fb/course/course-v1:edX+TEST_01+2016_R1/enroll/
5. The learner will be redirected to `idp` login page related to the enterprise and on successful login learner will be able view enterprise landing page after linking the edX account with enterprise.
6. On clicking the button `"Continue"`, after selecting a course mode the learner will be redirected to a course specific consent page.
7. Clicking on link `"No, take me back."` will show a modal with consent declining message "Are you aware...".
8. Now verify that on clicking the button `"I decline"` the learner is redirected back to enterprise course enrollment page with consent decline warning notification (screenshot attached).
screen shot 2017-06-23 at 2 30 32 pm

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


tests/test_views.py
```diff
@@ -915,6 +914,73 @@ def test_post_course_specific_consent_not_provided(
     @mock.patch('enterprise.views.configuration_helpers')
     @mock.patch('enterprise.views.render', side_effect=fake_render)
     @mock.patch('enterprise.views.CourseApiClient')
+    def test_post_course_specific_consent_not_provided_with_notification(
+            self,
+            course_api_client_mock,
+            *args  # pylint: disable=unused-argument
+    ):
```
asadiqbal08: A docstring would be helpful to understand the test.

enterprise/views.py
```diff
@@ -423,6 +424,37 @@ def post_course_specific_consent(self, request, course_id, consent_provided):
...
+            scheme, netloc, path, query_string, fragment = urlsplit(failure_url)
+            url_params = parse_qs(query_string)
+            # pylint: disable=invalid-name
+            show_consent_decline_notification = url_params.pop('show_consent_decline_notification', None)
+
+            if show_consent_decline_notification:
+                messages.warning(
+                    request,
+                    _(
+                        '{span_start}We could not enroll you in {italic_start}{course_name}{italic_end}.{span_end} '
```
asadiqbal08: is it OK to have HTML tags in translation strings ?

asadiqbal08: we need to add the corresponding string in `.po`

zubair-arbi: these are placeholders for html tags and should not affect translation strings. I have added `.po` files.

tests/test_views.py
zubair-arbi: done

zubair-arbi:
@douglashall @brittneyexline can some one of you please review?

tests/test_views.py
```diff
@@ -915,6 +914,76 @@ def test_post_course_specific_consent_not_provided(
...
+        # Verify that enterprise learner is redirected back to enterprise
+        # course enrollment page with consent decline warning message, if
+        # the learner decline the DSC consent.
```
brittneyexline: nit: "the learner declines data sharing consent."

enterprise/views.py
```diff
@@ -71,10 +71,11 @@
     get_enterprise_customer_user,
     is_consent_required_for_user,
 )
-from six.moves.urllib.parse import urlencode, urljoin  # pylint: disable=import-error
+from six.moves.urllib.parse import parse_qs, urlencode, urljoin, urlsplit, urlunsplit  # pylint: disable=import-error
 
 
 logger = getLogger(__name__)  # pylint: disable=invalid-name
+EDX_ENTERPRISE_SUPPORT_URL = 'https://support.edx.org/hc/en-us/sections/115002578128-Enterprise-Learner'
```
zubair-arbi: I believe this is a static link for all enterprises so we are fine at this stage.

If we want to have separate support links for different enterprises or if we need to update this link frequently then we should definitely make it configurable (maybe add a field in enterprise customer model). I think we can decide the solution/implementation with a new story.

asadiqbal08:
👍

zubair-arbi:
@brittneyexline @zrana you can verify this change here: https://ent.sandbox.edx.org/enterprise/cb7ca9a3-52d2-4181-961c-99c35b5714f7/course/course-v1:edX+DemoX+Demo_Course/enroll/

brittneyexline:
This is working as expected, however I noticed that the support link opens in the same window, I'm wondering if it's better to have it open in a new tab or window? Otherwise we take them out of the flow, and hitting back does not keep the message.

chris-mike:
After a quick convo with @cptvitamin , we could (1) fix the back problem or (2) open it in a new tab, but we'd need to indicate this (e.g., with an icon:  fa-external-link [&#xf08e;])

tests/test_views.py
```diff
@@ -915,6 +914,76 @@ def test_post_course_specific_consent_not_provided(
...
+        expected_consent_decline_msg = '<span>We could not enroll you in <i>{course_name}</i>.</span> ' \
+                                       'If you have questions or concerns about sharing your data, please contact ' \
+                                       'your learning manager at {enterprise_customer_name}, or contact ' \
+                                       '<a href="{edx_enterprise_support_link}">edX support</a>.'.format(
+                                           course_name=course_name,
+                                           enterprise_customer_name=enterprise_customer.name,
+                                           edx_enterprise_support_link=EDX_ENTERPRISE_SUPPORT_URL,
+                                       )
+        # pylint: disable=protected-access, no-member
+        response_messages = messages.storage.cookie.CookieStorage(response)._decode(response.cookies['messages'].value)
```
asadiqbal08: @zubair-arbi kindly take a look on that https://github.com/edx/ecommerce/blob/8fde125db17565c28cdf182fd35416c5c05b9dee/ecommerce/extensions/dashboard/tests.py#L8

zubair-arbi:
@brittneyexline @chris-mike I have added external link fa icon and modified support link to open in new tab. I have also [updated the sandbox](https://ent.sandbox.edx.org/enterprise/cb7ca9a3-52d2-4181-961c-99c35b5714f7/course/course-v1:edX+DemoX+Demo_Course/enroll/) to reflect the new changes. Please have a look.

tests/test_views.py
```diff
@@ -58,6 +57,13 @@ def setUp(self):
 
     url = reverse('grant_data_sharing_permissions')
 
+    def _get_messages_from_response_cookies(self, response):
+        """
+        Get django messages set in response cookies.
+        """
+        # pylint: disable=protected-access
+        return messages.storage.cookie.CookieStorage(response)._decode(response.cookies['messages'].value)
```
asadiqbal08: Try to use response context instead of `._decode`

zubair-arbi: Currently the django messages are set in cookie for tests.

douglashall:
@zubair-arbi I have cherry-picked this commit onto a sprint branch (edx/edx-enterprise:mighty_mouse) and deployed that branch to https://business.sandbox.edx.org.

enterprise/views.py
```diff
@@ -774,6 +782,9 @@ def get_enterprise_course_enrollment_page(self, request, enterprise_customer, co
             organization_logo = None
             organization_name = None
 
+        if enterprise_course_enrollment and not enterprise_course_enrollment.consent_granted:
+            add_consent_declined_message(request, enterprise_customer, course_details)
```
zubair-arbi: @douglashall shouldn't we show consent decline message only when a learner declines the consent? instead of showing consent decline message always if the learner has not provided course level consent.
There could be the case that the learner has provided account level consent.

brittneyexline: I agree with Zubair's point - I know we are initially using this with course level consent but it seems like it wouldn't be too difficult to check the UserDataSharingConsentAudit record if it exists.

zubair-arbi: @douglashall we can use the util method [is_consent_required_for_user(enterprise_customer_user, course_id=None)](https://github.com/edx/edx-enterprise/blob/master/enterprise/utils.py#L389) here.

enterprise/views.py
```diff
@@ -806,22 +821,22 @@ def post(self, request, enterprise_uuid, course_id):
...
-            return self.get_enterprise_course_enrollment_page(request, enterprise_customer, course, course_modes)
+            return self.get_enterprise_course_enrollment_page(request, enterprise_customer, course, course_modes,
+                                                              enterprise_course_enrollment)
 
         user_consent_needed = is_consent_required_for_user(enterprise_customer_user, course_id)
         if not selected_course_mode.get('premium') and not user_consent_needed:
             # For the audit course modes (audit, honor), where DSC is not
             # required, enroll the learner directly through enrollment API
             # client and redirect the learner to LMS courseware page.
-            with transaction.atomic():
```
zubair-arbi: I believe previously we wanted to revert the db transaction for creating `EnterpriseCourseEnrollment` object in case the external HTTP call to `EnrollmentApiClient` fails. Since enterprise is being installed as a django app in LMS so this call isn't external but we can decide if we need to avoid committing the `EnterpriseCourseEnrollment` in db just in case the API call fails.  @haikuginger can provide some insight.

enterprise/views.py
```diff
@@ -438,18 +440,23 @@ def post_course_specific_consent(self, request, course_id, consent_provided):
         except HttpClientError:
             raise Http404
 
-        enrollment_deferred = request.POST.get('enrollment_deferred')
```
brittneyexline: when I was going through this to update the tests, i was confused about the purpose of this enrollment_deferred variable as well as the consequences of removing it.... can you explain?

douglashall: @brittneyexline I was also confused about the use of the `enrollment_deferred` POST parameter. I did not understand why we would ever NOT want to create an EnterpriseCourseEnrollment record here if one did not already exist. Perhaps @haikuginger has some background he could share?

enterprise/views.py
douglashall: Since the `UserDataSharingConsentAudit` model is going to be short-lived and we are not actually going to make use of account-level consent in the short term, I'm thinking we should only consider course-level consent here. Let me know if there is a strong argument to do otherwise.

enterprise/views.py
haikuginger: I can comment on `enrollment_deferred`; not on why it'd be in use/removed here. Essentially, `enrollment_deferred` is used in the Ecom flow when a user goes to redeem a coupon- it prevents a consent record from being recorded, but lets us use the consent UI anyway. It's assumed that if the user is then redirected back to `success_url`, any relevant consent requirement was met, and the overall process can proceed - in the Ecom flow, the consent record is created when the actual enrollment is.

PR #130, opened 2017-06-19
ENT-454 Apply session management logic to new Enterprise landing page

brittneyexline:
**Description:** We had to implement a decorator (https://github.com/edx/edx-platform/pull/14896/files) for the track selection page when we were testing SAP that terminated any existing session coming into the page if coming in through SSO. Since we are using this new landing page, we need the same logic here in order to stay in compliance with SAPs expectations.

**JIRA:** https://openedx.atlassian.net/browse/ENT-454

**Dependencies:** 

**Merge deadline:** ASAP

**Installation instructions:** Standard Installation with this version of edx-enterprise.

**Testing instructions:**
1. Set up a SAML Configuration and SAML IdP Provider Configuration against TestShib, and set up an Enterprise Customer linked to the Provider Configuration. Ensure that the provider configuration has the following enabled: skip hinted login dialog, skip registration form. Leave the following disabled: drop existing session.

2. Go to https://brittneyexline.sandbox.edx.org/enterprise/3d3fa892-58e0-435f-b7d2-37a45b3ce18c/course/course-v1:edX+DemoX+Demo_Course/enroll/ and register a new user, going through the normal flow until you get to the dashboard. Do not log out.

3. Sign out of the user on the IdP or unlinking the first user from the IdP account, and revisit the page. It should redirect to the first user's account.

4. Change the Provider Configuration to enable drop existing session.

5. Revisit the page (if you have not logged out of the first account, otherwise repeat step 2 first). It should direct you to the account creation page as if you were a new user coming in. You should then be able to go through the normal flow as expected.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:**
I'd like to refactor the tests for our enterprise view to avoid so much duplication - they all seem to have similar set up.
I still need to add test coverage around the specific logic I added.

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -601,15 +601,29 @@ def enterprise_login_required(view):
 
         enterprise_uuid = kwargs['enterprise_uuid']
         enterprise_customer = get_enterprise_customer_or_404(enterprise_uuid)
+        provider_id = enterprise_customer.identity_provider or ''
+        sso_provider = None
 
-        # Now verify if the user is logged in. If user is not logged in then
-        # send the user to the login screen to sign in with an
-        # Enterprise-linked SSO and the pipeline will get them back here.
+        try:
+            sso_provider = get_identity_provider(provider_id)
+        except ValueError:
+            pass
+
+        decorator_already_processed = request.GET.get('session_cleared')
+
         if not request.user.is_authenticated():
-            next_url = '{current_url}?{query_string}'.format(
+            # If the user isn't logged in, send them to the log in page and redirect them
+            # back to the original view, indicating that the decorator has been processed.
+            next_url = '{current_url}?{params}'.format(
                 current_url=quote(request.get_full_path()),
-                query_string=urlencode({'tpa_hint': enterprise_customer.identity_provider})
+                params=urlencode(
+                    [
+                        ('tpa_hint', provider_id),
+                        ('session_cleared', 'yes')
+                    ]
+                )
             )
+
             return redirect(
                 '{login_url}?{params}'.format(
                     login_url='/login',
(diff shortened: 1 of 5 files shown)


enterprise/utils.py
```diff
@@ -613,15 +613,30 @@ def wrapper(request, *args, **kwargs):
...
-        if not request.user.is_authenticated():
-            next_url = '{current_url}?{query_string}'.format(
-                current_url=quote(request.get_full_path()),
-                query_string=urlencode({'tpa_hint': enterprise_customer.identity_provider})
+        try:
+            sso_provider = get_identity_provider(provider_id)
+        except ValueError:
+            sso_provider = None
+
+        if not sso_provider:
```
brittneyexline: I don't know what we should do in this case, so it could change.

enterprise/utils.py
```diff
@@ -601,15 +601,29 @@ def wrapper(request, *args, **kwargs):
...
         enterprise_customer = get_enterprise_customer_or_404(enterprise_uuid)
+        provider_id = enterprise_customer.identity_provider or ''
+        sso_provider = None
+
+        try:
+            sso_provider = get_identity_provider(provider_id)
+        except ValueError:
+            pass
+
+        decorator_already_processed = request.GET.get('session_cleared') == 'yes'
```
douglashall: Do you need to check the value of param here or is the fact that it exists enough?

enterprise/utils.py
```diff
@@ -601,15 +601,29 @@ def wrapper(request, *args, **kwargs):
...
+            # If the user isn't logged in, send them to the log in page and redirect them
+            # back to the original view, indicating that the decorator has been processed.
+            next_url = '{current_url}?{params}'.format(
                 current_url=quote(request.get_full_path()),
-                query_string=urlencode({'tpa_hint': enterprise_customer.identity_provider})
+                params=urlencode(
+                    [
+                        ('tpa_hint', provider_id),
+                        ('session_cleared', 'yes')
+                    ]
```
douglashall: Why a list of tuples instead of a dict?

brittneyexline: actually after changing this, i think the dict makes it so that the order is non deterministic, which is a minor testing headache, so i'm changing it back :-)

douglashall: Cool. Maybe change it down below as well then on line 644 to keep things consistent?

PR #132, opened 2017-06-20
ENT-442: Fix query string parameters in next and previous pages of enterprise catalog courses endpoint.

saleem-latif:
Hi @douglashall , @zubair-arbi , @asadiqbal08 

Please take a look at the changes. 

**Description:** 
The "next" and "previous" fields appear to be returning invalid page references when requesting pages other than the first (ie, no page value specified). This also happens when other parameters are specified, such as "limit" and "offset".
This PR fixes this problem for next and previous urls in API response.

**JIRA:** [ENT-442](https://openedx.atlassian.net/browse/ENT-442)

**Testing instructions:**
1. In Postman, with a valid authentication code for a staff user, attempt to access: /enterprise/api/v1/catalogs/124/courses/?page=2

**Merge checklist:**

- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/enterprise/api/pagination.py
+++ b/enterprise/api/pagination.py
@@ -21,7 +21,8 @@ def get_paginated_response(data, request):
     Returns:
         (Response): DRF response object containing pagination links.
     """
-    url = request.build_absolute_uri().rstrip('/') + '/'
+    url = urlparse(request.build_absolute_uri())._replace(query=None).geturl()
+
     next_page = None
     previous_page = None
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 ----------
 
+[0.35.2] - 2017-06-20
+---------------------
+
+* Fix Next and Previous page urls for enterprise catalog courses.
+
+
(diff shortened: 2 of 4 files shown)


enterprise/api/pagination.py
```diff
@@ -21,7 +21,8 @@ def get_paginated_response(data, request):
     Returns:
         (Response): DRF response object containing pagination links.
     """
-    url = request.build_absolute_uri().rstrip('/') + '/'
+    url = urlparse(request.build_absolute_uri())._replace(query=None).geturl()
```
brittneyexline: I think in general we should persist the query parameters in the original request to the next and previous urls, instead of completely replacing them, is it possible to recombine the url with a consolidation of the query parameters here and below?

douglashall: @brittneyexline Turns out we aren't actually throwing anything away here. This line is meant to extract the protocol, host, and path from the request to the Enterprise service so that they can be used to update the next/previous URLs that are returned from the Catalog service. It is appropriate to use the query parameters from the next/previous URLs from the Catalog data below.

PR #140, opened 2017-06-26
ENT-454 Refactor session management logic.

douglashall:
**Description:** The first commit in this PR is a refactor which moves edx-enterprise decorators out of the enterprise.utils module and into an enterprise.decorators module. The second commit in this PR is a refactor of the work done in https://github.com/edx/edx-enterprise/pull/130. I have moved the session termination logic into its own decorator so that it can be applied to only the views that need it.

**JIRA:** https://openedx.atlassian.net/browse/ENT-454

**Dependencies:** None

**Merge deadline:** ASAP

**Installation instructions:** Standard Installation with this version of edx-enterprise.

**Testing instructions:**

1. Go to https://business.sandbox.edx.org/enterprise/2b5a2956-7746-4fc9-9587-bc887ba45973/course/course-v1:edX+DemoX+Demo_Course/enroll/ and register a new user, going through the normal flow until you get to the dashboard. You should be able to submit DSC consent after the course enrollment landing page without error. Do not log out.

3. Sign out of testshib by deleting the idp.testshib.org cookies stored in your browser and unlink the first user from the IdP account, and revisit the page. It should redirect to the first user's account.

4. Change the Provider Configuration to enable drop existing session.

5. Revisit the page (if you have not logged out of the first account, otherwise repeat step 2 first). It should direct you to the login page as if you were a new user coming in. You should then be able to go through the normal flow as expected.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/enterprise/decorators.py
+++ b/enterprise/decorators.py
@@ -12,7 +12,7 @@ from django.http import Http404
 from django.shortcuts import redirect
 
 from enterprise.utils import get_enterprise_customer_or_404, get_identity_provider
-from six.moves.urllib.parse import urlencode  # pylint: disable=import-error
+from six.moves.urllib.parse import parse_qs, urlencode, urlparse  # pylint: disable=import-error
 
 
 def disable_for_loaddata(signal_handler):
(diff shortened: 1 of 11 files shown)


enterprise/decorators.py
```diff
@@ -0,0 +1,175 @@
...
+            raise Http404
+
+        enterprise_uuid = kwargs['enterprise_uuid']
+        enterprise_customer = get_enterprise_customer_or_404(enterprise_uuid)
+
+        # Now verify if the user is logged in. If user is not logged in then
+        # send the user to the login screen to sign in with an
+        # Enterprise-linked IdP and the pipeline will get them back here.
+        if not request.user.is_authenticated():
+            next_url = '{current_url}?{query_string}'.format(
```
brittneyexline: will this be ok if the current_url has query string params already? it seems like this might generate something invalid in that case

enterprise/decorators.py
```diff
@@ -0,0 +1,175 @@
...
+            # Some functionality ...
+
+        OR
+
+        class MyView(View):
+            ...
+            @method_decorator(force_fresh_session)
+            def get(self, request, enterprise_uuid):
+                # Some functionality ...
+
```
brittneyexline: Is it true that this decorator wouldn't really make sense to use without the login required decorator? Would that be worth noting?

douglashall: Yes, I will add a comment.

enterprise/decorators.py
douglashall: Good point. We should be using `urlparse` here. I will adjust this code.

PR #146, opened 2017-07-10
ENT-472: Add page_size in querystring and Host in Header for aws gateway integration request.

saleem-latif:
**Description:** 
The Enterprise Service returns a "next" field in the case of a large response which requires pagination. This field contains a fully-qualified URL specifying the service host (eg, courses.edx.org), which makes sense for clients interacting directly with the service.
However, this breaks down for clients interacting with the Enterprise Service via the API Gateway. In this case, the "next" parameter must NOT specify the service host, it must instead specify the gateway host (api.gateway.com).

**JIRA:** [ENT-472](https://openedx.atlassian.net/browse/ENT-472)

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/api-compact.yaml
+++ b/api-compact.yaml
@@ -27,6 +27,45 @@ apigateway_responses: &apigateway_responses
   500:
     statusCode: "500"
 
+apigateway_responses_with_mapping_template: &apigateway_responses_with_mapping_template
+  default:
+    statusCode: "400"
+  200:
+    statusCode: "200"
+    responseTemplates:
+      # Response body mapping template, this template is used for updating
+      # "next" and "previous" urls (both host and url path) while preserving
+      # the protocol (http | https) and querystring parameters.
+      application/json: >
+        #set($inputRoot = $input.path('$'))
+        #set($host = $stageVariables.gateway_host)
+
+        #macro (updateURL $url)
+            #if($url != '') {
+                $url.replaceAll("(^https?://)[^/]*[^?]*(.*$)", "$1$host$context.resourcePath$2")
+            } #else {
+                none
+            }
+            #end
+        #end
+
+        {
+          "count": $inputRoot.count,
+          "next": "$updateURL$inputRoot.next",
+          "previous": "$updateURL$inputRoot.previous",
+          "results": $inputRoot.results,
+        }
+  401:
+    statusCode: "401"
+  403:
+    statusCode: "403"
+  404:
+    statusCode: "404"
+  429:
+    statusCode: "429"
+  500:
+    statusCode: "500"
+
 produces: &produces
   - "application/json"
   - "application/csv"
@@ -77,10 +116,16 @@ page_qs_parameter: &page_qs_parameter
   required: false
   type: "number"
 
+page_size_qs_parameter: &page_size_qs_parameter
+  in: "query"
+  name: "page_size"
+  required: false
+  type: "number"
+
 # AWS API Gateway vendor extension point. This information is used
 #  by https://github.com/awslabs/aws-apigateway-importer.
 x-amazon-apigateway-integration: &apigateway_integration
-  responses: *apigateway_responses
+  responses: *apigateway_responses_with_mapping_template
   httpMethod: "GET"
   type: "http"
   requestParameters:
@@ -88,6 +133,7 @@ x-amazon-apigateway-integration: &apigateway_integration
     integration.request.querystring.limit: "method.request.querystring.limit"
     integration.request.querystring.offset: "method.request.querystring.offset"
     integration.request.querystring.page: "method.request.querystring.page"
+    integration.request.querystring.page_size: "method.request.querystring.page_size"
 
 x-amazon-apigateway-integration-with-id: &apigateway_integration_with_id_parameter
   responses: *apigateway_responses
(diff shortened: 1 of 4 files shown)


api-compact.yaml
```diff
@@ -85,9 +91,11 @@ x-amazon-apigateway-integration: &apigateway_integration
   type: "http"
   requestParameters:
     integration.request.header.Authorization: "method.request.header.Authorization"
+    integration.request.header.Host: "${stageVariables.enterprise_host}"
```
saleem-latif: @mattdrayer , @efagin Does `stageVariables.enterprise_host` contains the host name for api manager ? 
or is it the host where enterprise API is deployed ?

saleem-latif: if it is the latter than is there a way to get the domain of API gateway here ? 
may be something like `method.request.domain` ?

mattdrayer: I believe `stageVariables.enterprise_host` contains the host of the Enterprise Service / API

mattdrayer: Is this helpful?  https://docs.aws.amazon.com/apigateway/api-reference/resource/integration/

saleem-latif: I think it would be best to add new stage variables (e.g enterprise_api_host, discovery_api_host) that contain api gateway's domain. We can then use these varaibles in the configuration. 

@efagin , @mattdrayer  Any thoughts on this approach ?

mattdrayer: My only concern here, and it is a very small concern, is that if we decide to change the api host name at some point in the future from `api.edx.org` to, say, `enterprise-api.edx.org`, we'll need to remember to change this value as well or else we'll have broken integrations.  Like I said, however, this is probably a fringe case -- I assume we will continue to utilize `api.edx.org` as our single logical integration host for at least another year, if not for several.

efagin: Seems like we should be using the `X-Forwarded-Host` header, set to the gateway's custom domain name, which might be more appropriate here and is something we can change trivially later - might be worth creating a new globally-available stage variable called `gateway_host` or something that we can configure as needed? (I imagine that could be a separate PR against api-manager that I can take on if it's valuable.)

Another question is around pathing. The path of the enterprise endpoint through the gateway is something like `api.gateway.host/enterprise/v1/endpoint`, while through the origin is `edxapp.host/enterprise/api/v1/endpoint`. (Note the `api` in the path in the second example.) So this isn't just host rewriting, it's also path rewriting right?

efagin: Also note the known issue that the gateway does not forward `Host` headers through today (near the bottom of http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html). So we'd have to hard-code the gateway's custom domain to a stage variable, we can't dynamically figure it out by forwarding the host header from the original request.

saleem-latif: > So this isn't just host rewriting, it's also path rewriting right?

the answer is YES, the api manager does both of host rewriting as well as path rewriting.

saleem-latif: I think best way forward would be to add new stage variable named `gateway_host` (with api gateway's domain) and use this to forward `Host` headers.

@mattdrayer , @douglashall , @efagin -- Thoughts ?

efagin: Yes, it may rewrite the path TO the origin, but does it rewrite the path FROM the origin (as in, the pagination "next" URI)? How does the enterprise service know what the correct pagination URIs are?

That said, I recommend filing an AWS support ticket to ask if there's a better way to pass API Gateway custom domain information to the origin, or if we really do have to hard-code a stage variable. It would be really nice to have this "just work" across any deployment of the gateway.

mattdrayer: According to the API Gateway [known issues document](http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html), the host mapping is unavailable only when the system is configured for HTTP proxy mode, so I think we're in the clear?  From what I've read in that doc, the name of the Host header may actually be `x-amzn-Remapped-Host`

efagin: We're running in HTTP proxy mode (as opposed to mock, or lambda mode). This is confusing as they also have an ability to proxy all request/header data through that we are NOT using... but this applies to us. You could check that x-amzn header, but I wouldn't recommend relying on it unless you're able to map it to something more generic like x-forwarded-for.

mattdrayer: AH, I see that now -- yes, it was the "proxy all" configuration that I thought the document was referring to.

saleem-latif:
@douglashall @brittneyexline If you are good with this I think we can merge it and update [enterprise hash](https://github.com/edx/api-manager/blob/master/swagger/api.yaml#L40-L45) as part of https://github.com/edx/api-manager/pull/56.
@efagin has already approved the approach (see comments [here](https://openedx.atlassian.net/browse/ENT-472?focusedCommentId=273751&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-273751)) .

api.yaml
```diff
@@ -133,8 +170,26 @@ endpoints:
             integration.request.querystring.limit: method.request.querystring.limit
             integration.request.querystring.offset: method.request.querystring.offset
             integration.request.querystring.page: method.request.querystring.page
+            integration.request.querystring.page_size: method.request.querystring.page_size
           responses:
             200:
+              responseTemplates:
+                application/json: |
+                    #set($inputRoot = $input.path('$')) #set($host = $stageVariables.gateway_host)
+                    #macro (updateURL $url)
```
brittneyexline: I wonder if there is a way to not repeat this logic in so many different places?

efagin: @brittneyexline unless I'm mistaken, this file is generated from `api-compact.yaml` so there's some repetition from the expansion. Any code-quality questions should focus on the compact variant. (The expanded version is required for gateway integration, the compact version is for developers.)

brittneyexline: ah, thanks for the clarification @efagin!

PR #147, opened 2017-07-10
ENT-476 New enterprise endpoint for courses by enterprise

brittneyexline:
**Description:** The previous iteration of this endpoint took a catalog id, inferred the enterprise customer based on that catalog, and if the request user was not an EnterpriseCustomerUser associated with that enterprise, then there would be a permissions error.

This new endpoint will pull the courses in the catalog linked to the provided enterprise, which is explicitly stated in the call. There will be appropriate responses if:
*there is no catalog linked to the given enterprise
*if the request user is neither staff nor associated with the enterprise
*if the catalog api returns an empty response (different than having no courses)

On success, it will return the catalog api courses response format with augmented enterprise specific data as the earlier endpoint did.

**JIRA:** https://openedx.atlassian.net/browse/ENT-476

**Dependencies:** https://github.com/edx/edx-enterprise/pull/107 was the previous iteration of this.

**Merge deadline:** 

**Testing instructions:**
Using the business sandbox (https://business.sandbox.edx.org), do the following.

_with staff user_
1. Obtain an access token for the staff user (ask me for the client id and secret)
`$ curl -X POST -d 'grant_type=client_credentials&client_id=***&client_secret=***&token_type=jwt' -H 'content-type: application/x-www-form-urlencoded' https://business.sandbox.edx.org/oauth2/access_token/`

2. Using the token obtained above, hit the endpoint for the configured Pied Piper enterprise customer. The response should be identical to hitting the courses endpoint for the catalog attached (should be catalog 1):
`$ curl -X GET \
  https://business.sandbox.edx.org/enterprise/api/v1/enterprise-customer/2b5a2956-7746-4fc9-9587-bc887ba45973/courses \
  -H 'authorization: JWT eyJhbGciOiJIUzI1NiJ9.eyJzY29wZXMiOiBbInJlYWQiLCAid3JpdGUiLCAicHJvZmlsZSIsICJlbWFpbCJdLCAiYWRtaW5pc3RyYXRvciI6IHRydWUsICJhdWQiOiAibG1zLWtleSIsICJmYW1pbHlfbmFtZSI6ICIiLCAiaXNzIjogImh0dHBzOi8vYnVzaW5lc3Muc2FuZGJveC5lZHgub3JnL29hdXRoMiIsICJwcmVmZXJyZWRfdXNlcm5hbWUiOiAic3RhZmYiLCAibmFtZSI6ICIiLCAiZ2l2ZW5fbmFtZSI6ICIiLCAiZXhwIjogMTUwMDY5NDM5OCwgImlhdCI6IDE1MDA2NTgzOTgsICJlbWFpbCI6ICJzdGFmZkBleGFtcGxlLmNvbSIsICJzdWIiOiAiMTA2ZWNkODc4ZjQxNDhhNWNhYmI2YmJiMDk3OWI3MzAifQ.NTgkSlUlqY3UxJAX17NPBac15qqJq5DoQ64JQ-3j2YE' \`

_with bexline_api_user_
I set this user up as a non staff user to test the permissions aspect of this endpoint.

1. Ensure that the user is not an enterprise customer user attached to Pied Piper. If they are, remove the association in django admin.
2. Execute the above steps with the user's client and secret (ask me). Observe that you get an error indicating not having access to this enterprise's data.
3. Go into django admin and add bexline_api_user as an enterprise customer user for Pied Piper.
4. Execute the above steps again, observe that you get the catalog courses response.

Note that there appears to be throttling on the catalog endpoint so sometimes there will be responses that indicate that the data is unavailable - if you encounter this wait a few minutes before trying again.

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -19,7 +19,7 @@ from django.utils.decorators import method_decorator
 from enterprise import models
 from enterprise.api.filters import EnterpriseCustomerUserFilterBackend
 from enterprise.api.pagination import get_paginated_response
-from enterprise.api.permissions import IsServiceUserOrReadOnly
+from enterprise.api.permissions import IsServiceUserOrReadOnly, IsStaffUserOrLinkedToEnterprise
 from enterprise.api.throttles import ServiceUserThrottle
 from enterprise.api.v1 import decorators, serializers
 from enterprise.course_catalog_api import CourseCatalogApiClient
@@ -64,6 +64,57 @@ class EnterpriseCustomerViewSet(EnterpriseReadOnlyModelViewSet):
     )
     filter_fields = FIELDS
     ordering_fields = FIELDS
+
+    @detail_route(
+        permission_classes=[permissions.IsAuthenticated, IsServiceUserOrReadOnly, IsStaffUserOrLinkedToEnterprise],
+        authentication_classes=[JwtAuthentication, BearerAuthentication, SessionAuthentication],
+        throttle_classes=[ServiceUserThrottle],
+    )  # pylint: disable=invalid-name,unused-argument
+    def courses(self, request, pk=None):
+        """
+        Retrieve the list of courses contained within the catalog linked to this enterprise.
+
+        Only courses with active course runs are returned. A course run is considered active if it is currently
+        open for enrollment, or will open in the future.
+        ---
+        serializer: serializers.CourseSerializerExcludingClosedRuns
+        """
+        enterprise_customer = self.get_object()
+        self.check_object_permissions(request, enterprise_customer)
+
+        # Make sure there is a catalog associated with the enterprise
+        if not enterprise_customer.catalog:
+            error_message = (
+                "No catalog is associated with Enterprise {enterprise_name} from endpoint '{path}'.".format(
+                    enterprise_name=enterprise_customer.name,
+                    path=request.get_full_path()
+                )
+            )
+            logger.error(error_message)
+            raise NotFound(error_message)
+
+        # We have handled potential error cases and are now ready to call out to the Catalog API.
+        catalog_api = CourseCatalogApiClient(request.user)
+        courses = catalog_api.get_paginated_catalog_courses(enterprise_customer.catalog, request.GET)
+
+        # An empty response means that there was a problem fetching data from Catalog API, since
+        # a Catalog with no courses has a non empty response indicating that there are no courses.
+        if not courses:
+            error_message = (
+                "Unable to fetch API response for catalog courses for Enterprise {enterprise_name} from endpoint "
+                "'{path}'.".format(
+                    enterprise_name=enterprise_customer.name,
+                    path=request.get_full_path()
+                )
+            )
+            logger.error(error_message)
+            raise NotFound(error_message)
+
+        serializer = serializers.EnterpriseCatalogCoursesReadOnlySerializer(courses)
+
+        # Add enterprise related context for the courses.
+        serializer.update_enterprise_courses(enterprise_customer, enterprise_customer.catalog)
+        return get_paginated_response(serializer.data, request)
 
 
 class EnterpriseCourseEnrollmentViewSet(EnterpriseReadWriteModelViewSet):
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -194,7 +194,6 @@ class EnterpriseCustomer(TimeStampedModel):
             course_run_key (str): The course run id for the course to be displayed.
         Returns:
             (str): Enterprise landing page url.
-
         """
         if configuration_helpers is None:
             raise NotConnectedToOpenEdX(
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.39.2] - 2017-07-27
+---------------------
+
+* Introduce new endpoint to the Enterprise API to query for courses by enterprise id.
+
 [0.39.1] - 2017-07-27
 ---------------------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.39.1"
+__version__ = "0.39.2"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/enterprise/api/permissions.py
+++ b/enterprise/api/permissions.py
@@ -3,9 +3,14 @@ Permission classes for the Enterprise API.
 """
 from __future__ import absolute_import, unicode_literals
 
+from logging import getLogger
+
 from rest_framework import permissions
 
 from enterprise.api.utils import get_service_usernames
+from enterprise.models import EnterpriseCustomerUser
+
+logger = getLogger(__name__)  # pylint: disable=invalid-name
 
 
 class IsServiceUserOrReadOnly(permissions.IsAuthenticated):
@@ -24,3 +29,35 @@ class IsServiceUserOrReadOnly(permissions.IsAuthenticated):
             request.method in permissions.SAFE_METHODS or
             request.user.username in service_users
         )
+
+
+class IsStaffUserOrLinkedToEnterprise(permissions.IsAuthenticated):
+    """
+    This request is authenticated as a staff user, or as a user who is linked to the Enterprise Customer.
+    """
+
+    message = 'User must be a staff user or associated with the specified Enterprise.'
+
+    def has_object_permission(self, request, view, obj):
+        """
+        Enforce that the user is authenticated. If the user isn't a staff user, or isn't linked to the
+        Enterprise Customer as an Enterprise Customer user, deny the request.
+        The obj argument is expected to be an Enterprise Customer.
+        """
+
+        if not request.user.is_staff and not EnterpriseCustomerUser.objects.filter(
+                enterprise_customer=obj,
+                user_id=request.user.id,
+        ).exists():
+            error_message = (
+                "User '{username}' is not a staff user, and is not associated with "
+                "Enterprise {enterprise_name} from endpoint '{path}'.".format(
+                    username=request.user.username,
+                    enterprise_name=obj.name,
+                    path=request.get_full_path()
+                )
+            )
+            logger.error(error_message)
+            return False
+
+        return True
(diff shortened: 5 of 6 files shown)


tests/api/test_views.py
```diff
@@ -582,6 +620,132 @@ def test_post_enterprise_customer_user_logged_out(self):
...
+            enterprise_uuid,
+            catalog,
+            url,
+            link_user,
+            mocked_catalog_courses,
+            expected,
+            mock_catalog_api_client
+    ):
+        """
+        Make sure if an enterprise does not exist for this call, we return a descriptive error.
```
itsjeyd: @brittneyexline I think we should update this to reflect the fact that this test checks multiple scenarios; right now it sounds like the test is only verifying a single scenario.

tests/api/test_views.py
```diff
@@ -582,6 +620,132 @@ def test_post_enterprise_customer_user_logged_out(self):
         response = self.client.post(settings.TEST_SERVER + reverse('enterprise-learner-list'), data=data)
         assert response.status_code == 401
 
+    @ddt.data(
+        (
+            'd2098bfb-2c78-44f1-9eb2-b94475356a3f',
```
itsjeyd: @brittneyexline Nit: Introduce a constant (such as `DEFAULT_ENTERPRISE_UUID`) for this, perhaps?

brittneyexline:
@itsjeyd @douglashall this is ready for re-review and testing on the business sandbox.

itsjeyd:
@brittneyexline Just a quick heads-up that I'll need to push testing this to tomorrow morning. And to confirm, the courses endpoint for the attached catalog is https://business.sandbox.edx.org/catalog/v1/catalogs/1/courses/, and it's possible to access that as `staff`, with the token that gets generated in step 1?

I had a look at the code updates though; they look good to me.

brittneyexline:
@itsjeyd, the url I would suggest for accessing the catalog's course for comparison is https://business.sandbox.edx.org/enterprise/api/v1/catalogs/1/courses. It should be accessible with the staff user.

itsjeyd:
@brittneyexline Thanks for the details! I tested this:

* [x] Generating a JWT for `staff`/`bexline_api_user` went fine.

* [ ] Hitting the new `https://business.sandbox.edx.org/enterprise/api/v1/enterprise-customer/2b5a2956-7746-4fc9-9587-bc887ba45973/courses/` API endpoint as `staff` worked as expected; semantically, it returned the same result as `https://business.sandbox.edx.org/enterprise/api/v1/catalogs/1/courses/` but one minor difference was that the new endpoint returned the value of the `catalog_id` of a course as an int, while the existing endpoint returned it as a string.

* [x] After unlinking `bexline_api_user` from Pied Piper via the Django admin, hitting the new endpoint produced a 403 with `{"detail":"The user does not have permission to access this resource: User 'bexline_api_user' is not associated with Enterprise Pied Piper from endpoint '/enterprise/api/v1/enterprise-customer/2b5a2956-7746-4fc9-9587-bc887ba45973/courses/'."}`, as expected.

* [ ] After linking `bexline_api_user` to Pied Piper via the Django admin, hitting the new endpoint produced a 404 with `{"detail":"The resource you are looking for does not exist: Unable to fetch API response for catalog courses for Enterprise Pied Piper from endpoint '/enterprise/api/v1/enterprise-customer/2b5a2956-7746-4fc9-9587-bc887ba45973/courses/'."}`, so this part doesn't seem to be working yet (although I'm not sure if the problem is related to your changes).

brittneyexline:
@itsjeyd it was a permissions issue on the catalog, and should be fixed for you to retest.
For future reference, I had to go into the django admin for the discovery service, go to the catalog in question, click on "Object Permissions", and add permissions for bexline_api_user to view the catalog.

itsjeyd:
@brittneyexline OK, thanks for looking into the issue. It's working now! But I have one more question before formally approving these changes:

On a conceptual level, what's the benefit of/use case behind letting a user that is associated with an Enterprise customer access the new endpoint if they don't have permission to view the corresponding catalog? Based on the PR description I was assuming that users associated with an Enterprise customer would always get a meaningful response (in the form of a list of courses). Is that not what we want?

brittneyexline:
@itsjeyd, it has to do with the fact that this endpoint is wrapping the catalog api, and this was a configuration required for accessing the catalog api. I had forgotten to configure that.
In production, we have affiliate customers that hit the catalog api directly, and that configuration is what is handling the permissions of who gets to see which catalogs.
For the enterprise api, we are using the linkage of the customer's api user to ensure that they are only able to access that enterprise's data. The catalog permissioning is then irrelevant in a sense, but because we are wrapping an existing api, it still has to be configured properly there too.

itsjeyd:
@brittneyexline I see, thanks -- it still means that regular learners that are associated with an Enterprise customer wouldn't be able to get a meaningful response from the Enterprise API (or the catalog API) by default, but that's fine. We can address that use case separately if it ever comes up.

mattdrayer:
@saleem-latif @douglashall the work in this PR seems like it might support #157

douglashall:
@brittneyexline We'll need to also make this endpoint available via the API gateway.

brittneyexline:
@douglashall agreed, could you point me in the right direction for how to do that?

douglashall:
Yes. This work kind of ties in to some other work that Saleem has in progress.

enterprise/models.py
```diff
@@ -37,6 +37,11 @@
 except ImportError:
     configuration_helpers = None
 
+try:
+    from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
+except ImportError:
+    configuration_helpers = None
+
```
douglashall: It looks like we already have this block directly above.

enterprise/api/v1/views.py
```diff
@@ -64,6 +64,71 @@ class EnterpriseCustomerViewSet(EnterpriseReadOnlyModelViewSet):
...
+        # Make sure there is a catalog associated with the enterprise
+        if not enterprise_customer.catalog:
+            error_message = (
+                "No catalog is associated with Enterprise {enterprise_name} from endpoint '{path}'.".format(
+                    enterprise_name=enterprise_customer.name,
+                    path=request.get_full_path()
+                )
+            )
+            logger.error(error_message)
+            raise NotFound("The resource you are looking for does not exist: " + error_message)
```
douglashall: Should the message here be something similar to the above "This enterprise does not have a catalog associated with it."?

enterprise/api/v1/views.py
```diff
@@ -64,6 +64,71 @@ class EnterpriseCustomerViewSet(EnterpriseReadOnlyModelViewSet):
...
+        ).exists():
+            error_message = (
+                "User '{username}' is not associated with Enterprise {enterprise_name} from endpoint '{path}'.".format(
+                    username=request.user.username,
+                    enterprise_name=enterprise_customer.name,
+                    path=request.get_full_path()
+                )
+            )
+            logger.error(error_message)
+            raise PermissionDenied("The user does not have permission to access this resource: " + error_message)
```
douglashall: This block would be better as a permission class, similar to `IsServiceUserOrReadOnly`.

enterprise/api/v1/views.py
```diff
@@ -64,6 +64,71 @@ class EnterpriseCustomerViewSet(EnterpriseReadOnlyModelViewSet):
...
+        if not courses:
+            error_message = (
+                "Unable to fetch API response for catalog courses for Enterprise {enterprise_name} from endpoint "
+                "'{path}'.".format(
+                    enterprise_name=enterprise_customer.name,
+                    path=request.get_full_path()
+                )
+            )
+            logger.error(error_message)
+            raise NotFound("The resource you are looking for does not exist: " + error_message)
```
douglashall: It might be better for the message to indicate that there was a problem with retrieving the catalog instead of saying that the resource doesn't exist.

enterprise/api/v1/serializers.py
```diff
@@ -278,15 +278,16 @@ class EnterpriseCatalogCoursesReadOnlySerializer(serializers.Serializer):
...
         """
         This method adds enterprise specific metadata for each course.
 
         We are adding following field in all the courses.
             tpa_hint: a string for identifying Identity Provider.
         """
         courses = []
-        enterprise_customer = utils.get_enterprise_customer_for_user(request.user)
+        if enterprise_customer is None:
+            enterprise_customer = utils.get_enterprise_customer_for_user(request.user)
```
douglashall: Should we just always require that an EnterpriseCustomer is passed to this function instead?

brittneyexline: The other courses endpoint by catalog uses the same serializer, and does not pass in enterprise_customer. I thought this would be simpler as to not change that other view. If we choose to deprecate that, then we can make this required.

douglashall:
As far as the gateway goes, you'll need to add the new endpoint to the block [here](https://github.com/edx/edx-enterprise/blob/master/api-compact.yaml#L158). It should look similar to the configuration for the `/v1/catalogs/{id}/courses/` endpoint.

enterprise/api/v1/views.py
brittneyexline: would it be better to just use the error message instead of just prefixing it with this message? the error message string does get passed and echoed back to the caller.

enterprise/api/v1/serializers.py
douglashall: Hmm, I think it would be better to keep things consistent with the two endpoints and how they call this function on the serializer, but not a blocker for me.

enterprise/api/v1/views.py
douglashall: Ah, I missed the fact that the error message above was getting concatenated. Yeah, I would drop the part about the resource not existing.

enterprise/api/v1/serializers.py
douglashall: Actually, I am going to go ahead and change how the catalog courses endpoint calls this function as a part of https://github.com/edx/edx-enterprise/pull/157.

enterprise/api/v1/views.py
brittneyexline: I've looked into this (http://www.django-rest-framework.org/api-guide/permissions/), and my only concern about doing it that way is that we would lose expressiveness in the message returned to the user, as one message is returned for the PermissionClass and I don't see a way to dynamically generate it as we do here.

douglashall: It looks like there is a way to customize the error message returned by the endpoint. Search for "To change the error message associated with the exception" on the page you referenced above.

brittneyexline: right, but can you format that message attribute to include the request username and enterprise customer and path?

douglashall: Ah, gotcha. I think that is okay as long as we log the username/enterprise customer name when denying access.

brittneyexline:
@douglashall, can you take a look at this for a final review?

PR #151, opened 2017-07-18
[ENT-492] Add initial implementation of a generic Consent API.

UmanShahzad:
**Description:** (to quote the upstream ticket): Right now, our consent architecture is fragmented across the `EnterpriseCustomerUser` API and the `EnterpriseCourseEnrollment` API. This PR creates an API that can handle multiple use cases, based on the existing set of data models, in a new consent Djangoapp.

**JIRA:** [ENT-492](https://openedx.atlassian.net/browse/ENT-492)

**Dependencies:** None, but future stories depend on this.

**Merge deadline:** End of this week.

**Testing instructions:**

1. Confirm that Consent API tests cover all important cases.
1. Install this version of `edx-enterprise` in `edx-platform` on a devstack.
1. Register an account on that devstack.
1. Make an `EnterpriseCustomer`.
1. Make an `EnterpriseCustomerUser` based off of your registered account and associate it with the customer from the previous step.
1. Configure a course for the `EnterpriseCustomer` with whatever consent requirements you plan on testing.
1. Enroll in the course, and either provide or do not provide consent for that course, depending on what you plan on testing.
1. Send queries to endpoint `/consent/api/v1/data_sharing_consent` in accordance with the `username` of the account you created, the UUID of the `enterprise_customer` you created, and the `course_id` of the course you enrolled in. Ensure that consent-related responses match with what you planned on testing.
1. If you did not provide consent, send a `POST` request using the `curl` command line tool to the same endpoint (`/consent/api/v1/data_sharing_consent`) and confirm that consent is granted for the course.
1. If you did provide consent, send a `DELETE` request using the `curl` command line tool to the same endpoint (`/consent/api/v1/data_sharing_consent`) and confirm that consent is revoked for that course.

**Merge checklist:**

- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:** 

1. This is a stepping stone to something more complete; currently we're using models from the `enterprise` app, and we'd like to eventually use models internal to the `consent` app.
1. **The first commit of this application reorganizes the tests to make it easier and more sensible to add tests for new applications.**

**Reviewers**
- @haikuginger 
- @brittneyexline @douglashall


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=151"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

tests/consent/api/test_views.py
```diff
@@ -0,0 +1,452 @@
...
+                'course_id': TEST_COURSE,
+                'exists': True,
+                'consent_provided': False,
+                'consent_required': True
+            },
+            200
+        ),
+    )
+    @ddt.unpack
+    def test_consent_api_get_endpoint(self, factory, items, request_body, expected_response_body, expected_status_code):
```
haikuginger: Let's add checks for the different cases of EnterpriseCustomer consent requirement as well.

tests/consent/api/test_views.py
```diff
@@ -0,0 +1,452 @@
...
+            }],
+            {
+                'username': TEST_USERNAME,
+                'enterprise_customer': TEST_UUID,
+                'course_id': TEST_COURSE
+            },
+            {
+                'username': TEST_USERNAME,
+                'enterprise_customer': TEST_UUID,
+                'course_id': TEST_COURSE,
```
haikuginger: Why aren't we including state, etc, in our expectations for this test?

tests/consent/api/test_views.py
```diff
@@ -0,0 +1,452 @@
...
```
haikuginger: As above. We need actual expectations of what the end status ought to be.

tests/consent/test_helpers.py
```diff
@@ -0,0 +1,41 @@
...
+from django.test import testcases
+
+
+class ConsentHelpersTest(testcases.TestCase):
+    """
+    Test cases for helper functions for the Consent application.
+    """
+
+    def test_consent_exists(self):
+        pass
```
haikuginger: I'm assuming these will be filled in. 😉

tox.ini
```diff
@@ -23,17 +23,22 @@ match-dir = (?!migrations)
...
+    touch
 deps =
     ficus: -r{toxinidir}/requirements/test-ficus.txt
     ginkgo: -r{toxinidir}/requirements/test-ginkgo.txt
     hawthorn: -r{toxinidir}/requirements/test-hawthorn.txt
     master: -r{toxinidir}/requirements/test-master.txt
 commands =
+    touch tests/__init__.py
     py.test -Wd {posargs}
+    rm tests/__init__.py
```
haikuginger: Why?

consent/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""
+edX Enterprise's Consent application.
```
bradenmacdonald: In this, and perhaps also in some other docstrings for this app, can we please say more explicitly what "Consent" is? It's a very generic word and if I weren't part of the enterprise team I would have a hard time figuring out what this "consent" app is about.

Something like:
"The consent application provides an API that enables the platform to prompt the user to consent to share data about their activity and performance in (a) particular course(s) with the enterprise customer that is (usually) subsidizing their enrollment in those courses. When users grant consent to share data, that agreement is recorded in the database in an easily auditable way."

UmanShahzad: Yeah the docstrings definitely need work, I just didn't want to spend too much time on them just yet, so thank you for this example; it looks good enough to be straight dropped right into it. I'll edit the rest to reflect something similar in a shorter fashion where appropriate.

tox.ini
UmanShahzad: With the changes to the testing organization, I need to temporarily be able to access the `tests` package in some imports. Note that the same thing is being done below for quality testing. I'm not 100% sure why we don't just keep the `__init__.py`, though.

tests/consent/test_helpers.py
UmanShahzad: Of course! Next on my agenda.

tests/consent/api/test_views.py
UmanShahzad: See above comment for discussion / question.

tests/consent/api/test_views.py
UmanShahzad: So this is basically because I haven't actually made `POST` return any useful information. This is something I probably should have asked/discussed earlier: should `POST` return the following 2 extra keys:

```json
{
    "created": boolean,
    "consent_provided": boolean,
}
```

In my opinion, if `POST` returns a 200 status code, then it's obvious that `'consent_provided'` is going to be `True`, because we were able to successfully perform the only function relevant to `POST` here (provide consent), or create the resource and then perform that function. And if it returns 404, then `False`, because the request format was incorrect and we couldn't identify nor create the resource and provide consent. So do we need to return it? Either way, what do you reckon we should be returning to indicate state besides these?

tests/consent/api/test_views.py
UmanShahzad: I will do that in a coming push.

consent/helpers.py
```diff
@@ -0,0 +1,162 @@
...
+    provided = enterprise_course_enrollment is not None and enterprise_course_enrollment.consent_available
+
+    if not provided or enterprise_course_enrollment is None:
+        user_data_consent_audit = get_user_dsc_audit(user_id, enterprise_customer)
+        provided = user_data_consent_audit is not None and user_data_consent_audit.enabled
+
+    return provided is not None and provided is not False
+
+
+def consent_required(user_id, course_id, enterprise_customer_uuid):
```
UmanShahzad: Yeah at first I had a pretty bad understanding of what the relation was between consent being required and consent being provided. Can you double check the changes I made to some of the existing test's `consent_required` expectations to see if they make sense?

consent/helpers.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+    enterprise_customer = get_enterprise_customer(enterprise_customer_uuid)
+    enterprise_course_enrollment = get_enterprise_course_enrollment(user_id, course_id, enterprise_customer)
+    provided = enterprise_course_enrollment is not None and enterprise_course_enrollment.consent_available
+
+    if not provided or enterprise_course_enrollment is None:
+        user_data_consent_audit = get_user_dsc_audit(user_id, enterprise_customer)
+        provided = user_data_consent_audit is not None and user_data_consent_audit.enabled
+
+    return provided is not None and provided is not False
```
UmanShahzad: Yep, I had changed this function up before and forgot to change this.

consent/helpers.py
```diff
@@ -0,0 +1,162 @@
...
+    :param user_id: The user that grants consent.
+    :param course_id: The course for which consent is granted.
+    :param enterprise_customer_uuid: The consent requester.
+    :return: Whether consent is provided to the Enterprise customer by the user for a course.
+    """
+    enterprise_customer = get_enterprise_customer(enterprise_customer_uuid)
+    enterprise_course_enrollment = get_enterprise_course_enrollment(user_id, course_id, enterprise_customer)
+    provided = enterprise_course_enrollment is not None and enterprise_course_enrollment.consent_available
+
+    if not provided or enterprise_course_enrollment is None:
```
UmanShahzad: Got it.

consent/helpers.py
```diff
@@ -0,0 +1,162 @@
...
+
+    :param user_id: The user that grants consent.
+    :param course_id: The course for which consent is granted.
+    :param enterprise_customer_uuid: The consent requester.
+    :return: Whether any consent (provided or unprovided) exists.
+    """
+    enterprise_customer = get_enterprise_customer(enterprise_customer_uuid)
+    enterprise_course_enrollment = get_enterprise_course_enrollment(user_id, course_id, enterprise_customer)
+
+    if enterprise_course_enrollment is None or not enterprise_course_enrollment.consent_available:
```
UmanShahzad: I'm quite happy with getting rid of the NoneType check :+1:

consent/api/v1/views.py
```diff
@@ -0,0 +1,214 @@
...
+        :param user_id:
+        :param course_id:
+        :param enterprise_customer_uuid:
+        :return:
+        """
+        if not (user_id and course_id and enterprise_customer_uuid):
+            raise ConsentAPIRequestError
+        enterprise_customer = get_enterprise_customer(enterprise_customer_uuid)
+        enterprise_customer_user = get_enterprise_customer_user(enterprise_customer, user_id)
+        enterprise_course_enrollment = get_enterprise_course_enrollment(user_id, course_id, enterprise_customer)
```
UmanShahzad: Did it, thanks!

consent/api/v1/views.py
```diff
@@ -0,0 +1,214 @@
...
+            and not a course run key (course-v1:edX+DemoX+Demo_Course).
+
+    """
+
+    # TODO: Use staff-only permission.
+    permission_classes = []
+
+    def get_required_query_params(self, request):
+        """
+
```
UmanShahzad: Yep, gotta fill this and some others.

consent/api/v1/views.py
```diff
@@ -0,0 +1,214 @@
...
+            POST or DELETE /consent/api/v1/data_sharing_consent
+            >>> {
+            >>>     "username": "bob",
+            >>>     "course_id": "course-v1:edX+DemoX+Demo_Course",
+            >>>     "enterprise_customer": "enterprise-uuid-goes-right-here"
+            >>> }
+
+            The API accepts JSON objects with these key-value pairs for
+            POST or DELETE.
```
UmanShahzad: I don't think we should be doing that, see https://stackoverflow.com/questions/2539394/rest-http-delete-and-parameters for some reasons. I also tried it, and it broke a lot of tests, probably because this isn't really how it's meant to be used. I don't think we should have `DELETE` depend on query parameters.

tests/consent/api/test_views.py
haikuginger: Let's return a value indicating the state as we would for a GET request with the given parameters.

UmanShahzad: Alrighty!

consent/api/v1/views.py
brittneyexline: I agree with Uman here - I don't typically see DELETE calls take query parameters.

consent/helpers.py
```diff
@@ -0,0 +1,149 @@
...
+    Get whether consent is provided by the user to the Enterprise customer.
+
+    :param user_id: The user that grants consent.
+    :param course_id: The course for which consent is granted.
+    :param enterprise_customer_uuid: The consent requester.
+    :return: Whether consent is provided to the Enterprise customer by the user for a course.
+    """
+    enterprise_customer = get_enterprise_customer(enterprise_customer_uuid)
+    enterprise_course_enrollment = get_enterprise_course_enrollment(user_id, course_id, enterprise_customer)
+    provided = enterprise_course_enrollment is not None and enterprise_course_enrollment.consent_available
```
brittneyexline: nit: why not just `enterprise_course_enrollment and enterprise_course_enrollment.consent_available`?

consent/helpers.py
```diff
@@ -0,0 +1,149 @@
...
+
+    if enterprise_course_enrollment is None:
```
brittneyexline: IIRC it's also possible that `enterprise_course_enrollment.consent_available` can be None, and in that case it also means that we should look at the audit table.

consent/helpers.py
```diff
@@ -0,0 +1,149 @@
...
+    :param course_id: The course for which consent is granted.
+    :param enterprise_customer_uuid: The consent requester.
+    :return: Whether consent is required for a course by an Enterprise customer from a user.
+    """
+    if consent_provided(user_id, course_id, enterprise_customer_uuid):
+        return False
+
+    enterprise_customer = get_enterprise_customer(enterprise_customer_uuid)
+    return enterprise_customer is not None and any([
+        enterprise_customer.enforces_data_sharing_consent('at_login'),
```
brittneyexline: I thought we were not planning to support 'at_login' for the new endpoint and are removing it later, @haikuginger @douglashall ?

UmanShahzad: Personally, I didn't even know about this until going through @haikuginger's recent PR. I will be deleting any and all things related to `at_login` in this PR; thanks for noting :)

consent/helpers.py
UmanShahzad: Ah, true! Yes I'll add to that.

consent/helpers.py
UmanShahzad: No special reason; I feel that reading it with "is not None" at the end there makes it more clear what I'm looking for.

UmanShahzad: Scratch that: I just remembered we're trying to return a boolean here so it has to be `is not None` or else it'd return an object.

UmanShahzad:
I've removed `at_login` and `optional` consent logic / tests related to this app that I shouldn't have added in the first place... and added more test cases.

I just have to add the appropriate permission class (staff only), which I'll do tomorrow. For that, I'll either straight drop in the code that's written in my PR summary, or use the one from the edx-platform (but I feel safer doing the former, even if it leads to some repeat code, because of the potential to become an independent app).

UmanShahzad:
@haikuginger @brittneyexline I've added the permission classes, coverage for them, and changed the name of the `enterprise_customer` key to `enterprise_customer_uuid` to make it more clear what we want as input. I've also made the tests cleaner and more easily changeable in case some key's names change, etc..

consent/helpers.py
```diff
@@ -43,7 +43,7 @@ def consent_provided(user_id, course_id, enterprise_customer_uuid):
     enterprise_course_enrollment = get_enterprise_course_enrollment(user_id, course_id, enterprise_customer)
     provided = enterprise_course_enrollment is not None and enterprise_course_enrollment.consent_available
 
-    if enterprise_course_enrollment is None:
+    if enterprise_course_enrollment is None or not enterprise_course_enrollment.consent_available:
```
brittneyexline: i think this should actually be `or enterprise_course_enrollment.consent_available is None`, since if it is false, that means they declined consent.

UmanShahzad: Good point, done

UmanShahzad:
@brittneyexline @haikuginger Let's have a final look?

consent/api/v1/views.py
```diff
@@ -0,0 +1,228 @@
...
+            response_status = HTTP_200_OK
+        except ConsentAPIRequestError:
+            response_status = HTTP_400_BAD_REQUEST
+        provided = consent_provided(user_id, course_id, enterprise_customer_uuid)
+        required = consent_required(user_id, course_id, enterprise_customer_uuid)
+        return Response({
+            self.REQUIRED_PARAM_USERNAME: username,
+            self.REQUIRED_PARAM_COURSE_ID: course_id,
+            self.REQUIRED_PARAM_ENTERPRISE_CUSTOMER: enterprise_customer_uuid,
+            self.CONSENT_EXISTS: response_status == 200,
```
UmanShahzad: Ah yep, did this in a rush, thanks!

consent/api/v1/views.py
```diff
@@ -0,0 +1,228 @@
...
+
+        """
+        if not (user_id and course_id and enterprise_customer_uuid):
+            raise ConsentAPIRequestError
+        enterprise_customer = get_enterprise_customer(enterprise_customer_uuid)
+        enterprise_customer_user = get_enterprise_customer_user(enterprise_customer, user_id)
+        enterprise_course_enrollment = EnterpriseCourseEnrollment.objects.get_or_create(
+            enterprise_customer_user=enterprise_customer_user,
+            course_id=course_id
+        )[0]
```
UmanShahzad: :+1:

consent/api/v1/views.py
```diff
@@ -0,0 +1,228 @@
...
+        which are the relevant query parameters for this API endpoint.
+
+        :param request: The request to this endpoint.
+        :return: The ``username``, ``course_id``, and ``enterprise_customer_uuid`` from the request.
+
+        """
+        if request.method == 'GET':
+            username = request.query_params.get(self.REQUIRED_PARAM_USERNAME, '')
+            course_id = request.query_params.get(self.REQUIRED_PARAM_COURSE_ID, '')
+            enterprise_customer_uuid = request.query_params.get(self.REQUIRED_PARAM_ENTERPRISE_CUSTOMER, None)
```
UmanShahzad: :+1:

consent/api/v1/views.py
```diff
@@ -0,0 +1,228 @@
...
+
+    REQUIRED_PARAM_USERNAME = 'username'
+    REQUIRED_PARAM_COURSE_ID = 'course_id'
+    REQUIRED_PARAM_ENTERPRISE_CUSTOMER = 'enterprise_customer_uuid'  # pylint: disable=invalid-name
+
+    CONSENT_EXISTS = 'exists'
+    CONSENT_GRANTED = 'consent_provided'
+    CONSENT_REQUIRED = 'consent_required'
+
+    def get_required_query_params(self, request):
```
UmanShahzad: Yep, a 400-series error is returned if any of them are missing. I've added tests now to also cover the case that `course_id` is missing.

tox.ini
UmanShahzad: Hmm, OK so I read this carefully:

```
setup.py
mypkg/
    ...
tests/
    __init__.py
    foo/
        __init__.py
        test_view.py
    bar/
        __init__.py
        test_view.py

Now pytest will load the modules as tests.foo.test_view and tests.bar.test_view, allowing you to have modules with the same name. But now this introduces a subtle problem: in order to load the test modules from the tests directory, pytest prepends the root of the repository to sys.path, which adds the side-effect that now mypkg is also importable. This is problematic if you are using a tool like tox to test your package in a virtual environment, because you want to test the installed version of your package, not the local code from the repository.

In this situation, it is strongly suggested to use a src layout where application root package resides in a sub-directory of your root:

setup.py
src/
    mypkg/
        __init__.py
        app.py
        view.py
tests/
    __init__.py
    foo/
        __init__.py
        test_view.py
    bar/
        __init__.py
        test_view.py

This layout prevents a lot of common pitfalls and has many benefits, which are better explained in this excellent blog post by Ionel Cristian Mărieș.
```

And for that reason, since I don't think I should be making a new `src/` directory (or some other appropriate name) as part of this PR, I'm moving `utils` back out to root as `test_utils` to avoid the `__init__.py`. In a future PR we can solve this problem.

consent/helpers.py
```diff
@@ -0,0 +1,146 @@
...
+    :param user_id: The user that grants consent.
+    :param course_id: The course for which consent is granted.
+    :param enterprise_customer_uuid: The consent requester.
+    :return: Whether consent is provided to the Enterprise customer by the user for a course.
+    """
+    enterprise_customer = get_enterprise_customer(enterprise_customer_uuid)
+    enterprise_course_enrollment = get_enterprise_course_enrollment(user_id, course_id, enterprise_customer)
+    provided = enterprise_course_enrollment is not None and enterprise_course_enrollment.consent_available
+
+    if enterprise_course_enrollment is None or enterprise_course_enrollment.consent_available is None:
```
UmanShahzad: :+1:

consent/api/v1/views.py
```diff
@@ -0,0 +1,228 @@
...
+            POST or DELETE.
+
+        **Notes**
+
+            `course_id` specifies a course key (course-v1:edX+DemoX),
+            and not a course run key (course-v1:edX+DemoX+Demo_Course).
+
+    """
+
+    permission_classes = [permissions.IsStaff, ]
```
UmanShahzad: Will give it a shot.

UmanShahzad:
Squashed on (`abfeae3`)[https://github.com/open-craft/edx-enterprise/commit/abfeae354c49d7a0b77e5261476c2c3712ae8b68].

UmanShahzad:
@haikuginger I hope this should have addressed your nits! See the commit in the above comment specifically for what addressed the review.

consent/api/v1/views.py
haikuginger: Ohhh. I must have been getting an error-code response, but the overall response was still being rendered fully. Maybe we should return an actual structurally-different response if the parameters are missing, rather than proceeding to do the database queries with the empty values but just flagging with a 400 code.

UmanShahzad: Sure, that's possible too. A body with just an error message?

UmanShahzad: I've changed it to now return an error message on 400's. But a 404 (for GET when params are good but object doesn't exist) will still just return a normal body.

UmanShahzad:
Squashed on [1f85456](https://github.com/open-craft/edx-enterprise/commit/1f85456a38fc9283738bff335532546d5238a3eb).

UmanShahzad:
@haikuginger Ok, let me know if the error-case is good now.

UmanShahzad:
@brittneyexline Possible to get a final review from you? :)

docs/conf.py
```diff
@@ -463,7 +463,7 @@
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3.5', None),
     'django': ('https://docs.djangoproject.com/en/1.10/', 'https://docs.djangoproject.com/en/1.10/_objects/'),
-    'model_utils': ('https://django-model-utils.readthedocs.io/en/latest/', None),
+    'model_utils': ('https://django-model-test_utils.readthedocs.io/en/latest/', None),
```
haikuginger: I think this is an accidental change.

UmanShahzad: Ugh. Yes it is... the IDE did it. My fault for not carefully reading the diff, though.

UmanShahzad:
Squashed from 815fbb30ddbec2b5c4c47be1cdc0d654f0a500f4.

PR #153, opened 2017-07-21
ENT-519: Create Audit enrollment in ecommerce system when user enrolls in the audit mode in enterprise landing page.

saleem-latif:
FYI -- @douglashall , @brittneyexline , @zubair-arbi , @asadiqbal08 

**Description:** 
Currently, when users come to course about pages in edx.org and click enroll, a call is sent to the ecommerce api that writes both the enrollment record and places an audit order in the ecommerce system. The same logic does not currently exist in the landing page flow, so we need to make sure an audit order is placed when users come to the enterprise landing page.

**JIRA:** [ENT-519](https://openedx.atlassian.net/browse/ENT-519)

**Installation instructions:** Simply update edx-enterprise version to this branch.

**Testing instructions:**

1. Create a course with an "audit" mode.
2. Go to enterprise landing page for some enterprise customer and the course you created above.
3. Click on the enroll button.
4. Learner should be enrolled in the course.
4. Open ecommerce admin page and go to "Orders" list page at "/admin/order/order/". There should be a completed order for the above enrollment.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag


enterprise/views.py
```diff
@@ -705,6 +705,26 @@ def get_final_price(self, mode, request):
...
+        """
+        try:
+            api = ecommerce_api_client(request.user)
+            response_data = api.baskets.post({
+                'products': [{'sku': sku}],
+                'checkout': True,
+            })
+        except HttpClientError:
+            logger.error(
+                "Failed to post audit enrollment for product '{sku}'".format(
```
mattdrayer: If there is an error we should also log the user id so that an enrollment record could be created for the learner after the fact.

saleem-latif:
@mattdrayer I have updated the log message, please take a look

mattdrayer:
Looks good -- thanks for making this change -- 👍

saleem-latif:
@zubair-arbi , @brittneyexline , @douglashall I have updated the PR so that learner is redirected to enterprise landing page with error message.

asadiqbal08:
👍

enterprise/views.py
```diff
@@ -523,6 +525,40 @@ def get_final_price(self, mode, request):
...
+            logger.error(
+                "Failed to post audit enrollment of user '{username}' in product '{sku}'.".format(
+                    sku=sku,
+                    username=request.user.username,
+                )
+            )
+            messages.error(
+                request,
+                _(
+                    "There was an error processing your request, please try again. "
```
brittneyexline: Since this will be a user facing message, it might be better to say something like "There was an error completing your enrollment in the course" for this sentence. @chris-mike thoughts? What error message should we show to the user if the enrollment call fails for some reason?

enterprise/views.py
```diff
@@ -714,6 +750,13 @@ def post(self, request, enterprise_uuid, course_id):
...
                     course_id=course_id,
                 )
+
+            try:
+                # Post to E-Commerce for audit enrollment.
+                self.post_order_to_ecommerce(request, selected_course_mode['sku'])
+            except (HTTPError, Timeout, HttpClientError):
+                return redirect('enterprise_course_enrollment_page', enterprise_uuid, course_id)
+
             client = EnrollmentApiClient()
```
brittneyexline: I'm wondering if we'll run into the same race condition that caused ENT-512 by keeping this code around. My understanding was that the ecommerce call was queuing a task that would subsequently do some work, including calling back to the enrollment api in lms. Because we were also directly calling the enrollment api in the same view, it was causing an error triggered by the potential race condition. @douglashall what do you think?

saleem-latif:
@brittneyexline I have incorporated your feedback, please take another look

saleem-latif:
@brittneyexline I have deployed the changes to [saleem-latif.sandbox.edx.og](https://saleem-latif.sandbox.edx.org) you can test it there. Here is a link to enterprise landing page https://saleem-latif.sandbox.edx.org/enterprise/1aab1e4f-5d85-4587-be2e-8107774f488e/course/course-v1:edX+TC101+2017_T2/enroll/?tpa_hint=saml-testshib

Note: Make sure that the user trying to enroll in the course is associated with an enterprise customer.

saleem-latif:
@brittneyexline I have deployed the changes on business sandbox also, here is a link that can be used for testing https://business.sandbox.edx.org/enterprise/2b5a2956-7746-4fc9-9587-bc887ba45973/course/course-v1:edX+DemoX+Demo_Course/enroll

douglashall:
@saleem-latif Is there a reason for the package upgrades here?

enterprise/utils.py
```diff
@@ -26,6 +29,11 @@
 except ImportError:
     Registry = None
 
+try:
+    from openedx.core.djangoapps.commerce.utils import ecommerce_api_client
+except ImportError:
+    ecommerce_api_client = None
```
douglashall: Instead of importing this from the edx-platform module, enterprise should have it's own ecommerce client defined, similar to how we have created clients for [discovery](https://github.com/edx/edx-enterprise/blob/master/enterprise/course_catalog_api.py) and [lms](https://github.com/edx/edx-enterprise/blob/master/enterprise/lms_api.py). It would also be nice to reorganize that API client code a bit, perhaps creating an `api_client` folder which could contain a file for each services' client code (lms, discovery, ecommerce).

enterprise/views.py
```diff
@@ -382,6 +385,29 @@ class HandleConsentEnrollment(View):
...
+                ).format(
+                    link_start='<a href="{support_link}" target="_blank">'.format(
+                        support_link=configuration_helpers.get_value('ENTERPRISE_SUPPORT_URL')
+                    ),
+                    platform_name=configuration_helpers.get_value('PLATFORM_NAME'),
+                    link_end='</a>',
+                ),
+            )
+            raise
+
```
douglashall: This would be better as a method on the ecommerce api client mentioned above.

enterprise/views.py
```diff
@@ -523,6 +550,29 @@ def get_final_price(self, mode, request):
...
```
douglashall: Move the addition of the message to the message cache to the method on the ecommerce client mentioned above. No need to separate this out into two separate locations in this file.

saleem-latif:
@douglashall packages are updated as result `make upgrade && make requirements` as is written in merge checklist.
I have incorporated your feedback, please take another look

tests/test_management.py
```diff
@@ -22,7 +22,7 @@
 from django.core.management.base import CommandError
 from django.utils import timezone
 
-from enterprise import lms_api
+from enterprise.api_client import lms
```
douglashall: This import can be:
```
from enterprise.api_client import lms as lms_api
```
Then you can revert the rest of the changes in this file.

tests/test_enterprise/api_client/test_lms.py
```diff
@@ -15,15 +15,15 @@
 
 from django.conf import settings
 
-from enterprise import lms_api
+from enterprise.api_client import lms
```
douglashall: `from enterprise.api_client import lms as lms_api`

tests/test_management.py
saleem-latif: Done

tests/test_enterprise/api_client/test_lms.py
saleem-latif: Done

saleem-latif:
Verified on business sandbox, going to merge

douglashall:
@saleem-latif Please create a PR to bump the version in edx-platform.

saleem-latif:
@douglashall I was waiting for it to be deployed to PyPI, here is the PR https://github.com/edx/edx-platform/pull/15720

schenedx:
@douglashall @saleem-latif 
As part of my effort on LEARNER-2222 and https://github.com/edx/edx-platform/pull/15811
We should stop creating baskets and orders for audit enrollments.
Please help create an enterprise jira ticket and schedule ASAP

saleem-latif:
@schenedx , @douglashall I have created https://openedx.atlassian.net/browse/ENT-608 to further discuss it and make the changes.

PR #154, opened 2017-07-24
ENT-472: Remove Macro use from swagger api config as it is not supported by AWS.

saleem-latif:
**Description:** I have removed usages of VTL macros as they are not supported by AWS API Gateway.

**JIRA:** [ENT-472](https://openedx.atlassian.net/browse/ENT-472)

instructions.

**Testing instructions:**
1. Execute the following to get JWT access token.
    `curl -H "Content-Type:application/x-www-form-urlencoded" -X POST -d "grant_type=client_credentials&client_id=<client-id>&client_secret=<client-secret>&token_type=jwt" "https://api.stage.edx.org/oauth2/v1/access_token"`
2. Execute to get the catalogs
    `curl -H "Authorization:JWT <jwt-token>"  -X GET "https://20enjrmi05.execute-api.us-west-2.amazonaws.com/edx_api/enterprise/v1/catalogs?limit=2&offset=2"`
3. Make sure `next` and `previous` in response has correct domain i.e. `20enjrmi05.execute-api.us-west-2.amazonaws.com` and correct path i.e. `/edx_api/enterprise/v1/catalogs`

**Merge checklist:**

- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

**Dependencies:** This PR can be tested when [ENT-554](https://openedx.atlassian.net/browse/ENT-554) is fixed. I will update testing instructions when bug is fixed and we able to test endpoint for `/enterprise/v1/catalogs/{id}/courses/`

### DIFF
--- a/api-compact.yaml
+++ b/api-compact.yaml
@@ -39,21 +39,17 @@ apigateway_responses_with_mapping_template: &apigateway_responses_with_mapping_t
       application/json: >
         #set($inputRoot = $input.path('$'))
         #set($host = $stageVariables.gateway_host)
+        #set($catalogId = $input.params('id'))
 
-        #macro (updateURL $url)
-            #if($url != '') {
-                $url.replaceAll("(^https?://)[^/]*[^?]*(.*$)", "$1$host$context.resourcePath$2")
-            } #else {
-                none
-            }
-            #end
-        #end
+        #set($URLMatchRegex = "(^https?://)[^/]*[^?]*(.*$)")
+        #set($updatedURL = "$1$host$context.resourcePath$2")
+        #set($resourceIdMatch = "{id}")
 
         {
           "count": $inputRoot.count,
-          "next": "$updateURL$inputRoot.next",
-          "previous": "$updateURL$inputRoot.previous",
-          "results": $inputRoot.results,
+          "next": "$inputRoot.next.replaceAll($URLMatchRegex, $updatedURL).replace($resourceIdMatch, $catalogId)",
+          "previous": "$inputRoot.previous.replaceAll($URLMatchRegex, $updatedURL).replace($resourceIdMatch, $catalogId)",
+          "results": $inputRoot.results
         }
   401:
     statusCode: "401"
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.39.3] - 2017-07-28
+---------------------
+
+* Remove Macro use from swagger api config as it is not supported by AWS.
+
 [0.39.2] - 2017-07-27
 ---------------------
 
--- a/api.yaml
+++ b/api.yaml
@@ -21,20 +21,13 @@ apigateway_responses_with_mapping_template:
   200:
     responseTemplates:
       application/json: |
-          #set($inputRoot = $input.path('$')) #set($host = $stageVariables.gateway_host)
-          #macro (updateURL $url)
-              #if($url != '') {
-                  $url.replaceAll("(^https?://)[^/]*[^?]*(.*$)", "$1$host$context.resourcePath$2")
-              } #else {
-                  none
-              }
-              #end
-          #end
+          #set($inputRoot = $input.path('$')) #set($host = $stageVariables.gateway_host) #set($catalogId = $input.params('id'))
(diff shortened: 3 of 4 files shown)


efagin:
It's still not going to work, though, because {{$context.resourcePath}} isn't dynamically evaluating resource IDs, it's just the static gateway configuration. (Try it to see for yourself.)

saleem-latif:
@efagin At the moment we have only added fixes for catalogs endpoint "/enterprise/v1/catalogs/". This endpoint does not contain resource ID. Enterprise courses API endpoint that contains resource ID will be handled as part of https://openedx.atlassian.net/browse/ENT-554, we could not do that here as there is a bug that is causing 500 error on "/enterprise/v1/catalogs/{id}/courses/" and was not possible to test the changes.

efagin:
@saleem-latif got it, thanks. I'm unsure how many of the issues I saw were just because I was testing in the API Gateway console, I wonder what this actually looks like on a deployed gateway.

saleem-latif:
@efagin I have deployed the current changes in a development environment where it can be tested, I have added testing instructions in the description (I am using stage as API backend client credentials should be the ones present on stage). Here is a sample response on my machine.  

**Stage Variables:**
```
gateway_host: 20enjrmi05.execute-api.us-west-2.amazonaws.com/edx_api
enterprise_host: courses.stage.edx.org
```

```
curl -H "Authorization:JWT <jwt-token>" -X GET "https://20enjrmi05.execute-api.us-west-2.amazonaws.com/edx_api/enterprise/v1/catalogs?limit=2&offset=2"

{
  "count": 15,
  "next": "https://20enjrmi05.execute-api.us-west-2.amazonaws.com/edx_api/enterprise/v1/catalogs?limit=2&offset=4",
  "previous": "https://20enjrmi05.execute-api.us-west-2.amazonaws.com/edx_api/enterprise/v1/catalogs?limit=2",
  "results" : [
    {
        "query":"*",
        "courses_count":2337,
        "id":17,
        "viewers": ["awais","mattdrayer","ajj"],
        "name":"test_cat"
    },
    {
        "query":"key:MattX+SD1 OR MattX+SF1",
        "courses_count":2,
        "id":16,
        "viewers":[],
        "name":"Arbisoft Demo Catalog"
    }
  ],
}
```
But I think it would be a good idea to have another set of eyes on this to make sure everything is working as expected.

douglashall:
Step 3 in the testing instructions seems wrong. I think we want the domain used in the next/previous URLs to be `api.stage.edx.org` not something like `20enjrmi05.execute-api.us-west-2.amazonaws.com`. Is this possible?

douglashall:
Nevermind. I see now that this is the value you have explicitly set for the domain of the test api gateway.

douglashall:
Is there something about this change that is limiting the next/previous URL rewriting to the /enterprise/v1/catalogs/ endpoint? I'm trying to figure out why the URL rewriting wouldn't also be getting applied to the /enterprise/v1/catalogs/{id}/courses/ endpoint.

api-compact.yaml
```diff
@@ -66,6 +59,41 @@ apigateway_responses_with_mapping_template: &apigateway_responses_with_mapping_t
...
+  200:
+    statusCode: "200"
+    responseTemplates:
+      # Response body mapping template, this template is used for updating
+      # "next" and "previous" urls (both host and url path) while preserving
+      # the protocol (http | https) and querystring parameters.
+      application/json: >
+        #set($inputRoot = $input.path('$'))
+        #set($host = $stageVariables.gateway_host)
+        #set($catalogId = $input.params('id'))
```
douglashall: Could you use `$context.resourceId` instead? Does it matter?

api-compact.yaml
```diff
@@ -66,6 +59,41 @@ apigateway_responses_with_mapping_template: &apigateway_responses_with_mapping_t
   500:
     statusCode: "500"
 
+apigateway_catalog_courses_response_with_mapping_template: &apigateway_catalog_courses_response_with_mapping_template
```
douglashall: Do we need two separate templates? Couldn't we run the same string replace on all endpoints?

api-compact.yaml
saleem-latif: @douglashall `$context.resourceId ` does not contain catalog id.

ref: http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html#context-variable-reference
> $context.resourceId : The identifier API Gateway assigns to your resource.

api-compact.yaml
saleem-latif: Done

PR #156, opened 2017-07-25
[ENT-464] Skip audit track data passback for enrollments by flag.

UmanShahzad:
**Description:** This PR allows audit track enrollment reporting (including data pass-back via SuccessFactors) to be configurable on a per-customer basis.

**JIRA:** [ENT-464](https://openedx.atlassian.net/browse/ENT-464)

**Dependencies:** None

**Merge deadline:** End of this week or July 31st

**Testing instructions:**

- The SAP test instance is basically that which is referenced at the beginning of https://openedx.atlassian.net/wiki/pages/viewpage.action?spaceKey=SOL&title=SAP+Testing+and+Configuration+notes
- The business sandbox has courses that I *believe* are not yet *exported* to the SAP test instance
- Once the courses are *exported* and then *imported* to the SAP instance, the SAP *admin* has to make them available for SAP *learners*.
- Once the courses are available for SAP *learners*, we can take the course through SAP (they have a UI for selecting these imported courses and taking them).
- Now we do the real "testing". We enroll in audit mode, finish the course.
- We enroll (possibly for another SAP test learner; there are many available) in verified mode.
- We make sure the enterprise customer has audit data reporting disabled.
- We run the learner data export command and, as far as I know, we will be able to view this data on the SAP test instance somehow (where exactly it'd appear is what I'll have to figure out still), if our logic was implemented correctly in the PR.
- We run the script again and see that nothing has changed because it won't send the same data twice.
- We now turn on audit data reporting for the customer and run the learner data export command again. This time, it will send audit data.
- We check to see if it was sent. If our PR's logic is good, it will work.

**Merge checklist:**

- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:** There was a lot of code-reading ramp-up I had to do here, so I _might_ have missed a spot. But then again, the tests build my confidence, so I think everything's OK.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=156&repo=edx%2Fedx-enterprise"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

tests/test_models.py
```diff
@@ -1025,11 +1031,13 @@ def test_channel_code(self):
...
     @mock.patch(
         'integrated_channels.sap_success_factors.transmitters.learner_data.SuccessFactorsLearnerDataTransmitter')
+    @mock.patch('integrated_channels.integrated_channel.learner_data.EnrollmentApiClient')
     @mock.patch('integrated_channels.integrated_channel.learner_data.CertificatesApiClient')
     @mock.patch('integrated_channels.integrated_channel.learner_data.CourseApiClient')
     @mock.patch('enterprise.lms_api.JwtBuilder', mock.Mock())
-    def test_transmit_learner_data(self, mock_course_api, mock_certificate, mock_transmitter, mock_sap_client):
-
+    def test_transmit_learner_data(
+            self, mock_course_api, mock_certificate, mock_enrollment, mock_transmitter, mock_sap_client
```
itsjeyd: @UmanShahzad Not directly related to the task at hand, but since you're touching this: I think it would be worth making parameter names consistent, i.e., append `_api` to any parameter that represents a patched `ApiClient`.

tests/test_sap_success_factors/test_learner_data.py
```diff
@@ -302,3 +338,61 @@ def get_course_grade(course_id, username):
         assert learner_data[2].course_completed
         assert learner_data[2].completed_timestamp == self.NOW_TIMESTAMP
         assert learner_data[2].grade == grade
+
+    @ddt.data(
+        (True, True, 'audit', 1), (True, True, 'verified', 1),
+        (True, False, 'audit', 0), (True, False, 'verified', 1),
+        (False, True, 'audit', 0), (False, True, 'verified', 1),
+        (False, False, 'audit', 0), (False, False, 'verified', 1),
```
itsjeyd: @UmanShahzad Nit: Matching items in these tuples to the signature of the test method would be easier if each tuple were listed on a line by itself.

tests/test_sap_success_factors/test_learner_data.py
```diff
@@ -302,3 +338,61 @@ def get_course_grade(course_id, username):
...
+    )
+    @ddt.unpack
+    @mock.patch('integrated_channels.integrated_channel.learner_data.EnrollmentApiClient')
+    @mock.patch('integrated_channels.integrated_channel.learner_data.GradesApiClient')
+    @mock.patch('integrated_channels.integrated_channel.learner_data.CourseApiClient')
+    def test_learner_data_audit_track_data_reporting_enabled(
+            self,
+            enable_audit_enrollment,
+            enable_reporting,
```
itsjeyd: @UmanShahzad Let's call this `enable_audit_data_reporting` to be as specific as possible.

tests/test_sap_success_factors/test_learner_data.py
```diff
@@ -302,3 +338,61 @@ def get_course_grade(course_id, username):
...
+        # Collect the learner data
+        with freeze_time(self.NOW):
+            learner_data = list(self.exporter.collect_learner_data())
+
+        assert len(learner_data) == expected_data_len
+        if self.enterprise_customer.enables_audit_data_reporting or mode == 'verified':
+            assert learner_data[0].enterprise_course_enrollment_id == enrollment.id
+            assert learner_data[0].course_id == self.course_id
+            assert learner_data[0].course_completed
+            assert learner_data[0].grade == BaseLearnerExporter.GRADE_PASSING
```
itsjeyd: @UmanShahzad Let's extract `learner_data[0]` into a variable to be more DRY (something like `report`).

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -97,28 +96,28 @@ def collect_learner_data(self):
...
 
             if course_details is None:
                 # Course not found, so we have nothing to report.
                 LOGGER.error("No course details found for enrollment %d: %s",
                              enterprise_enrollment.pk, course_id)
                 continue
 
+            # Since the course exists, see if the course enrollment's
+            # configuration may entail skipping data collection.
+            if self._should_skip_enrollment(enterprise_enrollment):
```
itsjeyd: @UmanShahzad `_should_skip_enrollment` is a bit confusing, shouldn't it be something like `_should_skip_data_collection`?

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -134,6 +133,34 @@ def collect_learner_data(self):
...
 
+    def _should_skip_enrollment(self, enterprise_enrollment):
+        """
+        Specifies whether to skip data collection of the enterprise enrollment due to any configuration reason.
+
+        :param enterprise_enrollment: The enterprise enrollment whose configuration is to be checked against.
+        :return: True if we should skip, False otherwise.
+        """
+        return any([
+            not enterprise_enrollment.consent_available,
```
itsjeyd: @UmanShahzad How does consent play into this? It would be worth explaining that in a comment, I think, or extending the docstring.

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -134,6 +133,34 @@ def collect_learner_data(self):
...
+        :return: True if we should skip, False otherwise.
+        """
+        if not self.enterprise_customer.enables_audit_data_reporting:
+            if self.course_enrollment_api is None:
+                self.course_enrollment_api = EnrollmentApiClient()
+            course_enrollment = self.course_enrollment_api.get_course_enrollment(
+                enterprise_enrollment.enterprise_customer_user.username,
+                enterprise_enrollment.course_id
+            )
+            return course_enrollment['mode'] in ['audit', 'honor'] if course_enrollment is not None else True
```
itsjeyd: @UmanShahzad I think I understand what you're trying to do here, but it wasn't obvious right away:

* If we're dealing with a `verified` enrollment, data collection should proceed as normal; this method should return `False`.
* If we're dealing with an `audit` or an `honor` enrollment, data collection should:
    * proceed as normal if the Enterprise customer enables audit data reporting; this method should return `False`.
    * be skipped if the Enterprise customer does *not* enable audit data reporting; this method should return `True`.

The current approach is one way of implementing this behavior (although strictly speaking the method returns `None` instead of a Boolean if the Enterprise customer enables audit data reporting), but maybe there's a way to write this that is a bit more accessible/easier to parse on first read? If not, we should at least extend the docstring to explain what's going on; it glosses over some important details at the moment.

UmanShahzad: Thanks for noting these points; I didn't like this function the way it currently is either but was hoping your review could help out. I'll refactor this to make it more readable!

integrated_channels/integrated_channel/learner_data.py
UmanShahzad: I say in the docstring that we skip data collection for an enterprise enrollment due to configuration reasons: since `consent_available` is a property (i.e. dependent on some configuration), it's one of the things we check for to determine the skip. But I will try to word the docstring a little better, yes.

integrated_channels/integrated_channel/learner_data.py
UmanShahzad: Sort of: we're in the middle of looping through an enrollment queryset here, which is itself part of a function that indicates data collection (`collect_learner_data`). So I target the nearest scope with this wording (namely, the looping through the enrollment queryset), saying "skip this element". `_should_skip_data_collection` doesn't signify for _what_ we're skipping data collection and comes off too general.

We can compromise and make it `_should_skip_enrollments_data_collection`?

tests/test_sap_success_factors/test_learner_data.py
UmanShahzad: Are you sure? This is the pattern used in all the other tests. If I change it here, maybe I should then change it in all the others?

tests/test_sap_success_factors/test_learner_data.py
UmanShahzad: Yep, sure. At first I had it as just 3 parameters (bool, bool, str) so it made a little more sense to have it this way since everything was equal in the first 2 parameters for each side except the string ('audit' vs 'verified'), but now with the numbers it makes sense to make it 1 column.

tests/test_models.py
UmanShahzad: Sure, that's something I'd like to do :)

enterprise/migrations/0023_audit_data_reporting_flag.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
```
UmanShahzad: :+1: :)

UmanShahzad:
@itsjeyd Thanks for the review :) About this:

> Calling this done based on the assumption that the "learner data management command" calls collect_learner_data, which I didn't double-check.

Get ready for a ride: See [this task](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/management/commands/transmit_learner_data.py#L54) which calls [`transmit_learner_data`](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/models.py#L95), a method in [`EnterpriseCustomerPluginConfiguration`](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/models.py#L54), which is subclassed by [`SAPSuccessFactorsEnterpriseCustomerConfiguration`](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/sap_success_factors/models.py#L54), which [returns a `BaseLearnerExporter` for learner data exporting](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/sap_success_factors/models.py#L128), which implements [`collect_learner_data`](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/learner_data.py#L73), which, finally, is called by [the transmit function](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/models.py#L101). Hope that was fun :)

tests/test_sap_success_factors/test_learner_data.py
itsjeyd: @UmanShahzad 

> This is the pattern used in all the other tests.

Ah, OK. I wasn't aware of that. Fine to leave as is, then. Better to keep the focus narrow here, and not make the task more complex than it is :)

UmanShahzad: Well, actually, too late :P I've already made the changes throughout this file! Next push will contain it.

integrated_channels/integrated_channel/learner_data.py
itsjeyd: @UmanShahzad I didn't consider the larger context initially; `_should_skip_enrollment` makes more sense from that perspective. But I'd also be OK with making the method name more specific. My suggestion for a compromise would be `_should_skip_data_collection_for_enrollment`.

UmanShahzad: @itsjeyd I'd go with `_should_skip_data_collection_for_enrollment`, but unfortunately pylint gives it an invalid-name error :\ `_should_skip_enrollments_data_collection` passes its test, though.

UmanShahzad:
@itsjeyd I attempted to address your review. In particular, I'm not sure you'll be fully happy with some of the new names for some functions, but let's see! I'm trying to take into account pylint's preferences too, so it's a little hard.

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -134,6 +133,52 @@ def collect_learner_data(self):
...
+
+        :param enterprise_enrollment: The enterprise enrollment whose configuration is to be checked against.
+        :return: True if we should skip, False otherwise.
+        """
+        return any([
+            not enterprise_enrollment.consent_available,
+            self._disabled_audit_reporting(enterprise_enrollment),
+        ])
+
+    def _disabled_audit_reporting(self, enterprise_enrollment):
```
itsjeyd: @UmanShahzad Nit: For a method that returns a Boolean, I'd slightly prefer `_audit_reporting_disabled`.

Also, we could consider making this a method/property of the `EnterpriseCourseEnrollment` model (not completely sure about this at the moment).

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -134,6 +133,52 @@ def collect_learner_data(self):
...
+        If it isn't enabled, we check to see if the course mode is of an audit type.
+
+        :param enterprise_enrollment: The enterprise enrollment whose reporting configuration is to be checked against.
+        :return: True if we should skip, False otherwise.
+        """
+        if not self.enterprise_customer.enables_audit_data_reporting:
+            return self._is_audit_enrollment(enterprise_enrollment)
+
+        # Since audit data reporting is enabled, we always return False here.
+        return False
```
itsjeyd: @UmanShahzad This is much better, thanks!

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -134,6 +133,52 @@ def collect_learner_data(self):
...
+
+    def _disabled_audit_reporting(self, enterprise_enrollment):
+        """
+        Specifies whether to skip data collection due to audit track data reporting not being enabled.
+
+        If it is enabled, we simply skip the check, regardless of the ``enterprise_enrollment``.
+        If it isn't enabled, we check to see if the course mode is of an audit type.
+
+        :param enterprise_enrollment: The enterprise enrollment whose reporting configuration is to be checked against.
+        :return: True if we should skip, False otherwise.
```
itsjeyd: @UmanShahzad A couple suggestions for this docstring:

```
Specifies whether audit track data reporting is disabled for `enterprise_enrollment`.

* If the current Enterprise customer enables audit track data reporting, simply return False.
* If the current Enterprise customer does not enable audit track data reporting,
  return True if we are dealing with an audit enrollment, and False otherwise.

:param: enterprise_enrollment: The enterprise enrollment whose reporting configuration is to be checked against.
:return: True if audit track data reporting is disabled, False otherwise.

File: integrated_channels/integrated_channel/learner_data.py [hunk 12]
```diff
@@ -134,6 +133,52 @@ def collect_learner_data(self):
...
+        :param enterprise_enrollment: The enterprise enrollment whose mode we check.
+        :return: Whether the course enrollment mode is of an audit type.
+        """
+        if self.course_enrollment_api is None:
+            self.course_enrollment_api = EnrollmentApiClient()
+        course_enrollment = self.course_enrollment_api.get_course_enrollment(
+            enterprise_enrollment.enterprise_customer_user.username,
+            enterprise_enrollment.course_id
+        )
+        return course_enrollment['mode'] in ['audit', 'honor']
```
itsjeyd: @UmanShahzad I think the list of modes to check against should be determined using

```python
audit_modes = getattr(settings, 'ENTERPRISE_COURSE_ENROLLMENT_AUDIT_MODES', ['audit', 'honor'])
```

I'll let @brittneyexline and/or @douglashall confirm, though.

tests/test_sap_success_factors/test_learner_data.py
```diff
@@ -280,25 +334,98 @@ def get_course_grade(course_id, username):
...
-        assert learner_data[1].completed_timestamp is None
-        assert learner_data[1].grade == BaseLearnerExporter.GRADE_INCOMPLETE
-
-        assert learner_data[2].enterprise_course_enrollment_id == enrollment2.id
-        assert learner_data[2].course_id == course_id2
-        assert learner_data[2].course_completed
-        assert learner_data[2].completed_timestamp == self.NOW_TIMESTAMP
-        assert learner_data[2].grade == grade
+
+        report1 = learner_data[0]
```
UmanShahzad: Not really, that's why we were just referring to them as `learner_data[x]` before. Notice above in the code we have `enrollment1`, `enrollment2`, etc..

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -97,28 +96,28 @@ def collect_learner_data(self):
...
 
             if course_details is None:
                 # Course not found, so we have nothing to report.
                 LOGGER.error("No course details found for enrollment %d: %s",
                              enterprise_enrollment.pk, course_id)
                 continue
 
+            # Since the course exists, see if the course enrollment's
+            # configuration may entail skipping data collection.
+            if self._should_skip_enrollments_data_collection(enterprise_enrollment):
```
UmanShahzad: Sure!

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -134,6 +133,52 @@ def collect_learner_data(self):
                 is_passing=is_passing,
             )
 
+    def _should_skip_enrollments_data_collection(self, enterprise_enrollment):
+        """
+        Specifies whether to skip data collection of the enterprise enrollment due to any configuration reason.
+
+        Reasons for skipping data collection may include data sharing consent not being granted,
+        or audit track data reporting being disabled for audit enrollments, etc..
```
UmanShahzad: Sure, I've adopted this format, however left out the last line because this isn't really a "public" function and that line reads like it's meant for that.

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -134,6 +133,52 @@ def collect_learner_data(self):
...
+
+        :param enterprise_enrollment: The enterprise enrollment whose configuration is to be checked against.
+        :return: True if we should skip, False otherwise.
+        """
+        return any([
+            not enterprise_enrollment.consent_available,
+            self._disabled_audit_reporting(enterprise_enrollment),
+        ])
```
UmanShahzad: Yep I've of course thought of that technique, but I do think we _could_ come up with reasons for adding to the list, that's why I made it this way.

integrated_channels/integrated_channel/learner_data.py
UmanShahzad: I've thought about making it a property, but it makes an API call, so not sure how much I want to couple the course enrollment API with the EnterpriseCourseEnrollment model.

I will rename it though!

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -134,6 +133,52 @@ def collect_learner_data(self):
...
+        :param enterprise_enrollment: The enterprise enrollment whose reporting configuration is to be checked against.
+        :return: True if we should skip, False otherwise.
+        """
+        if not self.enterprise_customer.enables_audit_data_reporting:
+            return self._is_audit_enrollment(enterprise_enrollment)
+
+        # Since audit data reporting is enabled, we always return False here.
+        return False
+
+    def _is_audit_enrollment(self, enterprise_enrollment):
```
UmanShahzad: See comment above about the fact that this would couple it to the course enrollment API. What do you think about that?

integrated_channels/integrated_channel/learner_data.py
UmanShahzad: :+1:

integrated_channels/integrated_channel/learner_data.py
UmanShahzad: I think that's probably correct, yes. Will wait on confirmation, too.

integrated_channels/integrated_channel/learner_data.py
UmanShahzad: Thanks!

integrated_channels/integrated_channel/learner_data.py
UmanShahzad: OK so I took a look and noticed that even the `EnterpriseCustomerUser` currently uses the `ThirdPartyAuthApiClient` for one of its methods (`get_remote_id`). In that case, I guess we can go ahead and do it.

integrated_channels/integrated_channel/learner_data.py
UmanShahzad: Nevermind:

> OK so I took a look and noticed that even the EnterpriseCustomerUser currently uses the ThirdPartyAuthApiClient for one of its methods (get_remote_id). In that case, I guess we can go ahead and do it.

UmanShahzad:
@brittneyexline @douglashall If either of you have time to review this today, that'd be excellent :) I've addressed @itsjeyd's review.

enterprise/models.py
```diff
@@ -808,6 +831,40 @@ def consent_needed(self):
...
+        return False
+
+    @property
+    def is_audit_enrollment(self):
+        """
+        Specify whether the course enrollment associated with this ``EnterpriseCourseEnrollment`` is in audit mode.
+
+        :return: Whether the course enrollment mode is of an audit type.
+        """
+        course_enrollment_api = EnrollmentApiClient()
```
brittneyexline: Although it makes the code in the caller more readable, it feels wrong to introduce a dependency on an external api within the context of a model.
I suppose that means either checking the enrollment mode in the caller's code, or formally introducing this as a field on the EnterpriseEnrollment record at the time of enrollment. I think the latter is something we should consider outside of the scope of this PR.
I can't say what would be the best way forward without knowing if anyone else has the same concern.

UmanShahzad: @brittneyexline This was previously not here, but @itsjeyd suggested this as an alternative for readability. Then again, we already have another model (`EnterpriseCustomerUser`) depending on the TPA API client.

And I think that yes, something like introducing a dedicated field is probably the cleaner way out but would be out of place in this PR.

integrated_channels/integrated_channel/learner_data.py
UmanShahzad: Nevermind, I went ahead with this.

enterprise/models.py
UmanShahzad: See https://github.com/edx/edx-enterprise/pull/156/files/8ead4196c7c632a82494f0ee268b25c38f53ccab#diff-6f7edc719ace33fecde94be26155c9c9R151 for where/how it previously was. Sorry for the comment spam in that diff!

itsjeyd: @umanshahzad @brittneyexline I hadn't thought of 

> formally introducing this as a field on the EnterpriseEnrollment record at the time of enrollment.

That seems like a good idea! To avoid blowing up the scope here, let's create a follow-up story to address this. In the meantime I think it would be fine to stick with the current solution. To me it does feel like an `EnterpriseEnrollment` should be responsible for determining whether it is an audit enrollment

enterprise/models.py
```diff
@@ -808,6 +831,40 @@ def consent_needed(self):
...
+        course_enrollment = course_enrollment_api.get_course_enrollment(
+            self.enterprise_customer_user.username,
+            self.course_id
+        )
+        return course_enrollment.get('mode') in getattr(
+            settings, 'ENTERPRISE_COURSE_ENROLLMENT_AUDIT_MODES', ['audit', 'honor']
+        )
```
itsjeyd: @UmanShahzad Nit: Let's break this up into two lines for readability, i.e.:

```python
audit_modes = getattr(settings, 'ENTERPRISE_COURSE_ENROLLMENT_AUDIT_MODES, ['audit', 'honor'])
return course_enrollment.get('mode') in audit_modes
```

UmanShahzad: :+1:

UmanShahzad:
@itsjeyd That sounds about right. Just need your :+1: and @brittneyexline :+1: then I'll cherry-pick a squashed commit onto the business branch.

itsjeyd:
@UmanShahzad I won't approve this explicitly until we've tested it. But with my last nit addressed (thanks!) I'm good with deploying this to the business sandbox. Before doing that though, let's see if @brittneyexline is OK with that as well.

UmanShahzad:
@itsjeyd Sure, that's fine. Yes, I'd like to know if @brittneyexline and @douglashall are OK with deploying this to the sandbox or not.

UmanShahzad:
Rebased on 72abc76a09deec8ab39dc8b375fed43298694f02

UmanShahzad:
Squashed on 68f978b2d59fa36c9eff8a3748585bd7424a72c3

itsjeyd:
@UmanShahzad We have the go-ahead from @brittneyexline for deploying this to the business sandbox!

UmanShahzad:
Rebased on `master`. Old commit was ee20f0b9851e2234ff63fe032b90dfda7e3d1d61

UmanShahzad:
Rebased from c4386f9411c8dfa93750af45dcbb57788f9752d9 on `master`.

enterprise/models.py
```diff
@@ -807,6 +830,39 @@ def consent_needed(self):
...
+
+        :return: Whether the course enrollment mode is of an audit type.
+        """
+        course_enrollment_api = EnrollmentApiClient()
+        course_enrollment = course_enrollment_api.get_course_enrollment(
+            self.enterprise_customer_user.username,
+            self.course_id
+        )
+        audit_modes = getattr(settings, 'ENTERPRISE_COURSE_ENROLLMENT_AUDIT_MODES', ['audit', 'honor'])
+        return course_enrollment.get('mode') in audit_modes
```
brittneyexline: I ran into the following error when testing, so I would suggest changing this to:
`return course_enrollment and core_enrollment.get('mode') in audit_modes`

But, this raises an interesting question about how this scenario might come about. I'm pretty sure what happened is that the user unenrolled from the course, so the record in student_courseenrollment was removed, but the enterprise_enterprisecourseenrollment record still exists (which is what I'm seeing in the database). I think the right thing to do is to not send any data back in this case, since the course can't be completed if the user unenrolled. We'll have to consider what to report back to indicate that they unenrolled, but that is for another time.

To move forward with testing, I deleted the two enterprise enrollment records. We can just recreate this case by going through the flow and then unenrolling.

```
2017-08-03 10:36:27,690 ERROR 4962 [enterprise.api_client.lms] lms.py:223 - no course enrollment details found for username=mtd_learner8, course=course-v1:SAPSuccessFactors+SAPSF101+2017_R1
  File "./manage.py", line 116, in <module>
    execute_from_command_line([sys.argv[0]] + django_args)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 346, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/core/management/base.py", line 394, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/core/management/base.py", line 445, in execute
    output = self.handle(*args, **options)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/integrated_channels/integrated_channel/management/commands/transmit_learner_data.py", line 50, in handle
    self.transmit_learner_data.delay(api_username, integrated_channel.channel_code(), integrated_channel.pk)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/integrated_channels/integrated_channel/management/commands/__init__.py", line 24, in no_delay
    func(*args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/integrated_channels/integrated_channel/management/commands/transmit_learner_data.py", line 60, in transmit_learner_data
    integrated_channel.transmit_learner_data(api_user)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/integrated_channels/integrated_channel/models.py", line 101, in transmit_learner_data
    for learner_data in exporter.collect_learner_data():
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/integrated_channels/integrated_channel/learner_data.py", line 118, in collect_learner_data
    if self._should_skip_enrollment(enterprise_enrollment):
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/integrated_channels/integrated_channel/learner_data.py", line 150, in _should_skip_enrollment
    enterprise_enrollment.audit_reporting_disabled,
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/enterprise/models.py", line 855, in audit_reporting_disabled
    return self.is_audit_enrollment
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/enterprise/models.py", line 873, in is_audit_enrollment
    return course_enrollment.get('mode') in audit_modes
AttributeError: 'NoneType' object has no attribute 'get'
```

UmanShahzad: @brittneyexline Thanks so much for testing this! I understand how you were able to reproduce this, but yeah, this is something we can consider another time. For now, I'll change it to take care of the NoneType case.

UmanShahzad:
@brittneyexline @itsjeyd @douglashall Rebased from 743cbe4d085f211956a010d24e7673cc27c11e30 and addressed what error Brittney ran into. I also updated i18n + requirements, and added the required geckodriver to get around the Jasmine builds failing from the updated requirements. (In that case, I'll close #161 when this gets merged).

itsjeyd:
@UmanShahzad @brittneyexline Thanks for the updates! Just to make sure we don't forget: Did we create a follow-up story for formally introducing `is_audit_enrollment` as a field on the `EnterpriseEnrollment` record at the time of enrollment, as mentioned [here](https://github.com/edx/edx-enterprise/pull/156/files#r129780208)?

CC @douglashall

UmanShahzad:
Rebased on 525a6b6c05c67fb7afebb01b7a92069c0c893014

@itsjeyd Since I'm not part of grooming the upstream tickets, I don't know if we did.

enterprise/models.py
douglashall: I'm not sure we should denormalize the course mode by adding it to the EnterpriseCourseEnrollment. I think it is fine to make an API request here as long as the response from that request is being cached appropriately.

itsjeyd: @UmanShahzad FYI, this API request is currently failing on stage; it seems to be using the wrong host/port. @brittneyexline will follow up with a fix some time today which she'll ask us to review.

PR #157, opened 2017-07-26
ENT-554: Fix 500 errors on catalog courses API endpoint for users that are not associated with any enterprise customer.

saleem-latif:
**Description:** 
If a client tries to access catalog courses through enterprise API, and the user associated with the request is not associated with any enterprise customer then the API returns a `500` error. This PR fixes that issue, now for API request with users not associated with any enterprise customer API will return a response with `200` status instead of `500`. 

**JIRA:** [ENT-554](https://openedx.atlassian.net/browse/ENT-554)

**Testing instructions:**

1. Fetch catalog courses via `http://localhost:8000/enterprise/api/v1/catalogs/{id}/courses/` request should be made using JWT credentials, use client-id and client-secret for a user that is not associated with any enterprise customer.
2. Catalog courses endpoint should return a proper response with `200` status code.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -278,7 +278,7 @@ class EnterpriseCatalogCoursesReadOnlySerializer(serializers.Serializer):
     previous = serializers.CharField(read_only=True, help_text=_("URL to fetch previous page of courses."))
     results = serializers.ListField(read_only=True, help_text=_("list of courses."))
 
-    def update_enterprise_courses(self, request, catalog_id):
+    def update_enterprise_courses(self, enterprise_customer, catalog_id):
         """
         This method adds enterprise specific metadata for each course.
 
@@ -286,7 +286,6 @@ class EnterpriseCatalogCoursesReadOnlySerializer(serializers.Serializer):
             tpa_hint: a string for identifying Identity Provider.
         """
         courses = []
-        enterprise_customer = utils.get_enterprise_customer_for_user(request.user)
 
         global_context = {
             'tpa_hint': enterprise_customer and enterprise_customer.identity_provider,
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.39.1] - 2017-07-27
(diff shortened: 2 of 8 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -364,7 +364,8 @@ def update_course_runs(self, course_runs, catalog_id, enterprise_customer):
                 query_parameters=query_parameters,
             )
 
-            enrollment_url = enterprise_customer.get_course_enrollment_url(course_run.get('key'))
+            enrollment_url = enterprise_customer and \
+                enterprise_customer.get_course_enrollment_url(course_run.get('key'))
```
douglashall: I would go with the following to make this a bit more explicit:
```
enrollment_url = None
if enterprise_customer:
    enrollment_url = enterprise_customer.get_course_enrollment_url(course_run.get('key'))
```

saleem-latif:
@douglashall I have incorporated your feedback, please take a look

mattdrayer:
Hmm, but don't we want the API user to be linked to the enterprise customer when accessing the catalog for said enterprise customer?  I almost feel like this is an AuthZ issue and if the user is not linked then we should return a 404.

douglashall:
@mattdrayer If we need to add permissions to the Enterprise API to support sales use cases I feel like that should be a separate story.

mattdrayer:
I guess I'd prefer that the 500 error continue to be returned, rather than a 200 response with data for that enterprise, if the user is not linked to the enterprise.

douglashall:
@mattdrayer Why?

mattdrayer:
I've been using it as a prompt that the configuration is not correct, because linking the user to the enterprise customer is an easy thing to forget.

douglashall:
Seems like there is a better solution to solving the configuration problem. Let me think about that.

douglashall:
@saleem-latif I have made some changes here. Could you give it a review?

saleem-latif:
@douglashall Changes look good 👍

PR #164, opened 2017-08-04
[ENT-341] Setup enterprise API gateway & service endpoint for Programs

UmanShahzad:
**Description:** This PR adds an Enterprise API service endpoint for Programs and Enterprise Catalogs which query course discovery, as well as an API gateway for those endpoints.

**JIRA:** [ENT-341](https://openedx.atlassian.net/browse/ENT-341)

**Merge deadline:** Coming Monday or slightly later.

**Testing instructions:**

Note: Anytime I say "send a GET request", the fastest way to do it is to just target the business sandbox through the browser (i.e. point to `https://business.sandbox.edx.org/enterprise/api/v1/enterprise-catalogs`). Make sure you're logged in as staff in the admin portal first, though.

- Testing `/enterprise/api/v1/enterprise-catalogs/`.

1. Send a GET request to `/enterprise/api/v1/enterprise-catalogs/` Note the results you get back.
1. Send a GET request to the discovery sandbox's `/api/v1/search/all/?q=` Note the results you get back.
1. Note that the results from the past 2 steps are equal.

- Testing `/enterprise/api/v1/enterprise-catalogs/{uuid}`.

1. Create an `EnterpriseCustomerCatalog` instance in the Django admin portal. Give it any query you want to test against (depending on the data available in the discovery sandbox). Note the UUID of the instance.
1. Send a GET request to `/enterprise/api/v1/enterprise-catalogs/{uuid}`, replacing `{uuid}` with the UUID you noted above. Note the results you get back.
1. Send a GET request to the discovery sandbox's `/api/v1/search/all/` with the query parameters equal to the query parameters you wrote in the instance. I.e. `/api/v1/search/all/?q=HelloWorld`. Note the results you get back.
1. Note that the results from the past 2 steps are equal.

- Testing `/enterprise/api/v1/programs/{uuid}`.

1. Send a GET request to `/enterprise/api/v1/programs/{uuid}`, replacing `{uuid}` with the UUID of an existing Program (search our discovery sandbox for some Program UUID, or create a new one and note the UUID). Note the results you get back.
1. Send a GET request to the discovery sandbox's `/api/v1/search/all/?q={uuid}` with `{uuid}` replaced with the Program UUID from above. Note the results you get back.
1. Note that the results from the past 2 steps are equal.

- Testing gateway

Note: I was given access to AWS by Matt Drayer. If you need access for any reason (although for testing this, you shouldn't unless you want to examine the actual gateway setup), ask him.

1. Get a JWT token from the business sandbox. You can do this using this cURL command:

`curl -H "Content-Type:application/x-www-form-urlencoded" -X POST -d "grant_type=client_credentials&client_id=<client-id>&client_secret=<client-secret>&token_type=jwt" "https://business.sandbox.edx.org/oauth2/access_token"`

Note to replace `<client-id>` and `<client-secret>` with the appropriate values (check for existing ones through the admin portal).

2. Use this JWT token in the following commands (replace `<jwt-token>` with the JWT token you generated above):

`curl -H "Authorization:JWT <jwt-token>" -X GET https://20enjrmi05.execute-api.us-west-2.amazonaws.com/edx_api/enterprise/v1/enterprise-catalogs/`
`curl -H "Authorization:JWT <jwt-token>" -X GET https://20enjrmi05.execute-api.us-west-2.amazonaws.com/edx_api/enterprise/v1/enterprise-catalogs/6b9f5528-23c9-4eff-aeb7-2755bf6780c1/`
`curl -H "Authorization:JWT <jwt-token>" -X GET https://20enjrmi05.execute-api.us-west-2.amazonaws.com/edx_api/enterprise/v1/programs/52ad909b-c57d-4ff1-bab3-999813a2479b/`

The above use UUIDs I setup before hand. You can play around by adding `EnterpriseCustomerCatalog`s and `Program`s as you like and use their UUIDs.

**Merge checklist:**

- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=164"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

UmanShahzad:
@douglashall Before I add tests, can you go over this to see whether you agree with the approach and whether it matches your vision of it?

api-compact.yaml
```diff
@@ -27,7 +30,7 @@ apigateway_responses: &apigateway_responses
   500:
     statusCode: "500"
 
-apigateway_responses_with_mapping_template: &apigateway_responses_with_mapping_template
+apigateway_responses_with_mapping_template_for_id: &apigateway_responses_with_mapping_template_for_id
```
UmanShahzad: I have little idea if this makes sense / is allowed, FYI.

enterprise/models.py
```diff
@@ -852,6 +852,54 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
 
 
 @python_2_unicode_compatible
+class EnterpriseCatalog(TimeStampedModel):
```
douglashall: On second thought, let's call this `EnterpriseCustomerCatalog` since it better reflects what this model actually is. You also won't have to refactor the names of the existing views/serializers.

douglashall:
@UmanShahzad Awesome! This looks like a solid start on the approach we have discussed. I had one note about the new catalog model name. I'll give this a more comprehensive view as we get closer to completion.

UmanShahzad:
@douglashall Great to hear! I very much agree with your note about the name change -- dealing with naming conflicts was a little annoying, but that should no longer be an issue.

enterprise/models.py
UmanShahzad: :+1: x100

UmanShahzad:
Rebased on 41063e7b7e4c06c7e7d966b0358f4743670a0463.

I just realized I opened this commit on a personal branch :\ Well, no worries, will be careful next time.

@douglashall @brittneyexline @itsjeyd Ready for review!

enterprise/api/v1/views.py
```diff
@@ -76,39 +108,34 @@ def courses(self, request, pk=None):
...
-        if not enterprise_customer.catalog:
-            error_message = (
-                "No catalog is associated with Enterprise {enterprise_name} from endpoint '{path}'.".format(
-                    enterprise_name=enterprise_customer.name,
-                    path=request.get_full_path()
-                )
+        self.ensure_data_exists(
+            request,
+            enterprise_customer.catalog,
+            "No catalog is associated with Enterprise {enterprise_name} from endpoint '{path}'.".format(
```
itsjeyd: @UmanShahzad Can we pass this string as a kwarg to make it even easier to tell what its purpose is?

enterprise/api/v1/views.py
```diff
@@ -248,14 +297,73 @@ class EnterpriseCustomerEntitlementViewSet(EnterpriseReadOnlyModelViewSet):
...
+
+        Arguments:
+            request (HttpRequest): Current request.
+            pk (string): Identifier for the EnterpriseCustomerCatalog instance that contains the necessary query.
+
+        Returns:
+            (Response): DRF response object containing a paginated result of searching the course discovery service.
+        """
+        if 'programs' in request.path:
+            # For the Programs endpoint, just search course discovery for the UUID.
```
itsjeyd: @UmanShahzad Nit: The UUID of what?

enterprise/api/v1/views.py
```diff
@@ -248,14 +297,73 @@ class EnterpriseCustomerEntitlementViewSet(EnterpriseReadOnlyModelViewSet):
...
+                    "No EnterpriseCustomerCatalog with uuid '{uuid}'; "
+                    "defaulting to empty query string.".format(uuid=pk)
+                )
+
+        catalog_api = CourseCatalogApiClient(request.user)
+        search_results = catalog_api.get_search_results(query)
+        self.ensure_data_exists(
+            request,
+            search_results,
+            "No search results returned with query '{query}'.".format(query=query)
```
itsjeyd: @UmanShahzad Same comment here about passing this as a kwarg; let's do this everywhere.

enterprise/api/v1/views.py
```diff
@@ -267,15 +375,10 @@ def list(self, request):
         Returns:
             (Response): DRF response object containing course catalogs.
         """
-        # fetch all course catalogs.
```
itsjeyd: @UmanShahzad Why remove this comment?

enterprise/api/v1/views.py
```diff
@@ -289,14 +392,9 @@ def retrieve(self, request, pk=None):  # pylint: disable=invalid-name
         Returns:
             (Response): DRF response object containing course catalogs.
         """
-        # fetch course catalog for the given catalog id.
```
itsjeyd: @UmanShahzad Why remove this comment?

enterprise/api/v1/views.py
```diff
@@ -289,14 +392,9 @@ def retrieve(self, request, pk=None):  # pylint: disable=invalid-name
...
         catalog_api = CourseCatalogApiClient(request.user)
         catalog = catalog_api.get_catalog(pk)
-
-        if not catalog:
-            logger.error("Unable to fetch API response for given catalog from endpoint '/catalog/%s/'.", pk)
-            raise NotFound("The resource you are looking for does not exist.")
-
+        self.ensure_data_exists(request, catalog)
```
itsjeyd: @UmanShahzad Let's preserve the custom error message here (pass it as `error_message`).

enterprise/api/v1/views.py
```diff
@@ -308,21 +406,10 @@ def courses(self, request, enterprise_customer, pk=None):  # pylint: disable=inv
...
-        # if API returned an empty response, that means pagination has ended.
-        # An empty response can also means that there was a problem fetching data from catalog API.
-        if not courses:
-            logger.error(
-                "Unable to fetch API response for catalog courses from endpoint '%s'.",
-                request.get_full_path(),
-            )
-            raise NotFound("The resource you are looking for does not exist.")
-
+        self.ensure_data_exists(request, courses)
```
itsjeyd: @UmanShahzad Let's preserve the custom error message here (pass it as `error_message`).

enterprise/api/v1/views.py
```diff
@@ -308,21 +406,10 @@ def courses(self, request, enterprise_customer, pk=None):  # pylint: disable=inv
...
         Only courses with active course runs are returned. A course run is considered active if it is currently
         open for enrollment, or will open in the future.
-        ---
-        serializer: serializers.CourseSerializerExcludingClosedRuns
         """
         catalog_api = CourseCatalogApiClient(request.user)
         courses = catalog_api.get_paginated_catalog_courses(pk, request.GET)
-
-        # if API returned an empty response, that means pagination has ended.
-        # An empty response can also means that there was a problem fetching data from catalog API.
```
itsjeyd: @UmanShahzad I think it would make sense to preserve this comment. Please replace "means" with "mean" in L318.

enterprise/api_client/discovery.py
```diff
@@ -72,6 +79,36 @@ def __init__(self, user):
...
+
+        Returns:
+            dict: API response with search results, as well as links to next and previous pages.
+
+        """
+        return self._load_data(
+            self.SEARCH_ALL_ENDPOINT,
+            default=[],
+            querystring=querystring,
+            traverse_pagination=False,
```
itsjeyd: @UmanShahzad What does `traverse_pagination` mean? (You didn't introduce that parameter so no need to change anything; I'm just asking because its name is a little confusing to me.)

enterprise/api_client/discovery.py
```diff
@@ -142,7 +195,11 @@ def get_course_details(self, course_key):
             dict: Details of the course in question.
 
         """
-        return self._load_data('courses', resource_id=course_key, many=False)
+        return self._load_data(
+            self.COURSES_ENDPOINT,
```
itsjeyd: @UmanShahzad Thanks for DRYing out the `_load_data` calls in this file.

enterprise/migrations/0024_enterprisecustomercatalog_historicalenterprisecustomercatalog.py
```diff
@@ -0,0 +1,55 @@
...
+            fields=[
+                ('created', model_utils.fields.AutoCreatedField(default=django.utils.timezone.now, verbose_name='created', editable=False)),
+                ('modified', model_utils.fields.AutoLastModifiedField(default=django.utils.timezone.now, verbose_name='modified', editable=False)),
+                ('uuid', models.UUIDField(default=uuid.uuid4, serialize=False, editable=False, primary_key=True)),
+                ('query', models.TextField(help_text='Query to the course discovery service. Leave empty for all results.')),
+                ('enterprise_customer', models.ForeignKey(related_name='enterprise_customer_catalog', to='enterprise.EnterpriseCustomer', on_delete=django.db.models.deletion.CASCADE)),
+            ],
+            options={
+                'verbose_name': 'Enterprise Catalog',
+                'verbose_name_plural': 'Enterprise Catalogs',
```
itsjeyd: @UmanShahzad Values should be "Enterprise Customer Catalog" and "Enterprise Customer Catalogs", respectively?

enterprise/migrations/0024_enterprisecustomercatalog_historicalenterprisecustomercatalog.py
```diff
@@ -0,0 +1,55 @@
...
+                ('history_id', models.AutoField(serialize=False, primary_key=True)),
+                ('history_date', models.DateTimeField()),
+                ('history_type', models.CharField(max_length=1, choices=[('+', 'Created'), ('~', 'Changed'), ('-', 'Deleted')])),
+                ('enterprise_customer', models.ForeignKey(related_name='+', on_delete=django.db.models.deletion.DO_NOTHING, db_constraint=False, blank=True, to='enterprise.EnterpriseCustomer', null=True)),
+                ('history_user', models.ForeignKey(related_name='+', on_delete=django.db.models.deletion.SET_NULL, to=settings.AUTH_USER_MODEL, null=True)),
+            ],
+            options={
+                'ordering': ('-history_date', '-history_id'),
+                'get_latest_by': 'history_date',
+                'verbose_name': 'historical Enterprise Catalog',
```
itsjeyd: @UmanShahzad "Historical Enterprise Customer Catalog"?

enterprise/models.py
```diff
@@ -908,6 +908,57 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
...
+        on_delete=models.deletion.CASCADE
+    )
+    query = models.TextField(
+        help_text=_('Query to the course discovery service. Leave empty for all results.')
+    )
+    history = HistoricalRecords()
+
+    class Meta(object):
+        verbose_name = _("Enterprise Catalog")
+        verbose_name_plural = _("Enterprise Catalogs")
```
itsjeyd: @UmanShahzad I should have put my comment about making the values of these attributes match the model name here... :)

* "Enterprise Customer Catalog"
* "Enterprise Customer Catalogs"

enterprise/api/permissions.py
```diff
@@ -61,3 +62,37 @@ def has_object_permission(self, request, view, obj):
...
+        """
+        Enforce that the user is authenticated. If the user isn't a staff user, or isn't linked to some common
+        Enterprise Customer as an Enterprise Customer user, deny the request.
+
+        Expects that the view has a ``cross_check_model`` member which serves as the cross-check model for the existence
+        of a common Enterprise Customer.
+        """
+        cross_check_enterprise = get_enterprise_customer_for_user(request.user)
+        cross_check_model = getattr(view, 'cross_check_model', None)
+        has_common_enterprise = cross_check_model.objects.filter(
```
UmanShahzad: :+1:

api-compact.yaml
```diff
@@ -199,3 +274,45 @@ endpoints:
           x-amazon-apigateway-integration:
             <<: *apigateway_integration_with_id_and_querystring_parameters
             uri: "https://${stageVariables.enterprise_host}/enterprise/api/v1/catalogs/{id}/courses/"
+
+    # /v1/enterprise_catalogs/
+    newEnterpriseCatalogs:
```
UmanShahzad: See above in the code -- it's taken :(

api.yaml
```diff
@@ -273,6 +352,142 @@ endpoints:
               statusCode: "400"
           type: http
           uri: "https://${stageVariables.enterprise_host}/enterprise/api/v1/catalogs/"
+    newEnterpriseCatalogs:
```
UmanShahzad: See above.

api.yaml
```diff
@@ -273,6 +352,142 @@ endpoints:
...
+      get:
+        operationId: new_get_enterprise_catalogs
```
UmanShahzad: It's taken.

enterprise/api/permissions.py
```diff
@@ -61,3 +62,37 @@ def has_object_permission(self, request, view, obj):
             return False
 
         return True
+
+
+class IsStaffUserOrLinkedToCommonEnterprise(permissions.IsAuthenticated):
+    """
+    This request is authenticated as a staff user, or is linked to the endpoint's appointed common Enterprise.
```
UmanShahzad: Sure!

enterprise/api/permissions.py
UmanShahzad: :+1:

enterprise/api/v1/serializers.py
```diff
@@ -129,10 +166,23 @@ def save(self):  # pylint: disable=arguments-differ
         )
 
 
+class EnterpriseCatalogSerializer(serializers.ModelSerializer):
```
UmanShahzad: Argh, not sure how this slipped through the cracks while I was renaming; thanks.

enterprise/api/v1/serializers.py
```diff
@@ -191,37 +244,35 @@ def save(self):  # pylint: disable=arguments-differ
...
-        Do not perform any operations for state changing requests.
-        """
-        pass
     user = UserSerializer(read_only=True)
     enterprise_customer = EnterpriseCustomerSerializer(read_only=True)
     data_sharing_consent = UserDataSharingConsentAuditSerializer(many=True, read_only=True)
 
 
-class EnterpriseCourseCatalogReadOnlySerializer(serializers.Serializer):
+class EnterpriseCustomerCatalogApiPaginatedResponseReadOnlySerializer(ResponsePaginationSerializer):
```
UmanShahzad: I would name it with `API` but, since I think `Api` tends to be the preferred style and any renaming done in this file would be pretty easy to fix across the codebase (since it's hardly used), I'll change the other one rather than this one; since you know I'll be doing it now, the review won't come as a surprise -- is that good?

enterprise/api/v1/urls.py
```diff
@@ -7,16 +7,15 @@
...
 router = DefaultRouter()  # pylint: disable=invalid-name
 router.register("site", views.SiteViewSet, 'site')
 router.register("auth-user", views.UserViewSet, 'auth-user')
-router.register(
-    "enterprise-course-enrollment",
-    views.EnterpriseCourseEnrollmentViewSet,
-    'enterprise-course-enrollment'
-)
+router.register("enterprise-catalog", views.EnterpriseCustomerCatalogViewSet, 'enterprise-catalog')
+router.register("enterprise_catalogs", views.EnterpriseCustomerCatalogApiViewSet, 'enterprise_catalogs')
```
UmanShahzad: Because that's the spec @douglashall layed out in the ticket -- to use `/enterprise_catalogs/{uuid}`, etc.

enterprise/api/v1/views.py
UmanShahzad: @itsjeyd Ah yes, I had thought of doing it but did a double-take. I'll do that now.

enterprise/api/v1/views.py
UmanShahzad: Ah this comment made more sense when I had `uuid` as `pk`, but had to change it to be *actually* `pk` because that's what the function apparently expects. I'll change it now.

enterprise/api/v1/views.py
UmanShahzad: @itsjeyd Ah yes, I had thought of doing it but did a double-take. I'll do that now.

enterprise/api/v1/views.py
UmanShahzad: Comments are obviously a good practice but imagine something akin to:

```python
# get the users
users = get_the_users()
```

Such comments can end up being awkward technical debt, interestingly. I deleted these because doing a quick read of the code I felt they didn't contribute much at all to readability -- did you feel differently?

enterprise/api/v1/views.py
UmanShahzad: Comments are obviously a good practice but imagine something akin to:

```python
# get the users
users = get_the_users()
```

Such comments can end up being awkward technical debt, interestingly. I deleted these because doing a quick read of the code I felt they didn't contribute much at all to readability -- did you feel differently?

enterprise/api/v1/views.py
UmanShahzad: I think the issue here is I am using the same error message to log *and* to throw, but here they use different ones. Which would you recommend to pass in? I don't know if I want to make an extra parameter for differentiating between logging/throwing, to keep things simple & consistent.

enterprise/api/v1/views.py
UmanShahzad: Sure! See above -- which one?

enterprise/api/v1/views.py
UmanShahzad: See above for some reasoning, but for this comment in particular -- it kind of just repeats what's in the error message. Unless you mean to not pass in a customized error message and leave the comment, then I understand that the comment could be useful.

enterprise/api_client/discovery.py
UmanShahzad: :+1:

enterprise/migrations/0024_enterprisecustomercatalog_historicalenterprisecustomercatalog.py
UmanShahzad: Woops! Not sure how this escaped naming changes :s I must've forgotten to remake migrations.

enterprise/migrations/0024_enterprisecustomercatalog_historicalenterprisecustomercatalog.py
UmanShahzad: Woops! Not sure how this escaped naming changes :s I must've forgotten to remake migrations.

enterprise/models.py
UmanShahzad: Woops! Not sure how this escaped naming changes :s I must've forgotten to remake migrations.

api-compact.yaml
UmanShahzad: I've renamed it to enterpriseCustomerCatalogs for now.

api.yaml
UmanShahzad: P.S. This is a generated file, so if I change api-compact.yml, this will update through the yamllint.com generator.

api.yaml
UmanShahzad: I've updated it to get_enterprise_customer_catalogs.

UmanShahzad:
@itsjeyd I addressed the parts of the review I agreed with you on -- on some I have questions so didn't make any changes for them. I'll await your review on that plus the remainder of the PR.

api-compact.yaml
itsjeyd: @UmanShahzad Do you know what the existing `enterpriseCatalogs` endpoint is for?

enterprise/api/v1/serializers.py
itsjeyd: @UmanShahzad Definitely fine going with the preferred style :) What about my comment about the length of the class name, though?

> since you know I'll be doing it now, the review won't come as a surprise -- is that good?

I'm not sure what you mean by this?

enterprise/api/v1/urls.py
itsjeyd: @UmanShahzad OK, but I'm still wondering if we can make these routes consistent? That is, if underscores are preferred, why not use them in L13 above?

enterprise/api/v1/views.py
itsjeyd: @UmanShahzad I know what you mean -- ideally, comments should describe **why** something needs to happen; if the code itself is well-written, it shouldn't be necessary to add comments that describe **what** happens.

I hadn't looked too closely at the body of this method, but now that I have I agree that it makes sense to remove this particular comment :)

enterprise/api/v1/views.py
itsjeyd: @UmanShahzad See above; I'm fine with removing this comment.

enterprise/api/v1/views.py
itsjeyd: @UmanShahzad Yeah, making an extra parameter seems like overkill. To not lose any info, how about combining the messages and passing in the resulting string?

enterprise/api/v1/views.py
itsjeyd: @UmanShahzad See above -- you could combine the messages.

enterprise/api/v1/views.py
itsjeyd: @UmanShahzad Interesting -- in this case I really do think that the comment adds some details that you can't glean from the code (especially now that the actual logic lives in a separate method. In particular, it mentions *why* a response might be empty. The error messages below don't do that.

So I vote for keeping this comment (and capitalizing the first `if`).

enterprise/api_client/discovery.py
itsjeyd: @UmanShahzad Ping :)

api-compact.yaml
```diff
@@ -199,3 +274,45 @@ endpoints:
...
+        get:
+          produces: *produces
+          parameters:
+            - *auth_header
+            - *uuid_parameter
+          operationId: "get_program_by_uuid"
+          responses: *responses
+          x-amazon-apigateway-integration:
+            <<: *apigateway_integration_with_uuid_parameter
+            uri: "https://${stageVariables.enterprise_host}/enterprise/api/v1/programs/{uuid}/"
```
brittneyexline: I'm in the same boat as @itsjeyd on this part of the change. Perhaps @saleem-latif can weigh in, as he has done a lot of work with the api gateway as well.

enterprise/api/v1/serializers.py
brittneyexline: I'm in favor of the shorter name proposed by @itsjeyd - mostly because it isn't as important to be explicit about the pagination, I'm ok with it being implied as it isn't an unreasonable implication.

enterprise/api/v1/urls.py
brittneyexline: Considering the route for all of the other endpoints registered here use dashes, let's stick with that for consistency.

enterprise/api/v1/views.py
```diff
@@ -76,39 +108,36 @@ def courses(self, request, pk=None):
 
         Only courses with active course runs are returned. A course run is considered active if it is currently
         open for enrollment, or will open in the future.
```
brittneyexline: I'm wondering what the implications of your change are for this endpoint, and by extension what they are for the way we currently link a Catalog to the EnterpriseCustomer. We still rely on that link for the SAP integration and have a story to back the catalog export script with this endpoint instead of hitting the discovery service directly.

enterprise/api/v1/views.py
```diff
@@ -308,21 +408,10 @@ def courses(self, request, enterprise_customer, pk=None):  # pylint: disable=inv
...
-        ---
```
brittneyexline: This is another view that I question the need for after this work is done, but I am fairly certain that it is in use, so maybe we can figure out from @mattdrayer who is using this and come up with a way to migrate them down the line.

enterprise/api/v1/views.py
UmanShahzad: @brittneyexline See https://openedx.atlassian.net/browse/ENT-341?focusedCommentId=278887&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-278887 and the comment before that -- as far as I know from @douglashall, this endpoint will make a transition in future stories.

enterprise/api/v1/views.py
UmanShahzad: See comment above.

enterprise/api/v1/views.py
UmanShahzad: Sure, let's go with keeping it and fixing the capitalization :P

enterprise/api_client/discovery.py
UmanShahzad: Hehe sorry, I didn't forget, I just need to go figure it out myself :P I copied this from the other function that gets a paginated response. If I had to guess, it just means "instead of traversing all the pages and giving me everything, *don't* traverse them and thus give me a paginated response instead" -- but I'll tell you whether that's the case in my next comment.

enterprise/api/v1/views.py
UmanShahzad: Sure, I'll try combining them into something meaningful.

enterprise/api/v1/views.py
UmanShahzad: :+1:

enterprise/api/v1/views.py
UmanShahzad: :+1:

enterprise/api/v1/views.py
UmanShahzad: :+1:

enterprise/api/v1/urls.py
UmanShahzad: @itsjeyd @brittneyexline Honestly I prefer dashes myself, because, well, I love `ctrl+delete`, and multiple words concatenated with underscores get insta-deleted with that keybind, but not with dashes. But since @douglashall wrote the original spec for this and he may have more insight into why he chose it (if it's not just an arbitrary choice), I'd like to hear it before I make the change, since it affects some tests I wrote, too.

enterprise/api/v1/serializers.py
UmanShahzad: @itsjeyd Sorry, that comment slipped by me -- yeah, the name is pretty long and we can remove the pagination part, no problem. CC @brittneyexline

api-compact.yaml
UmanShahzad: @itsjeyd @brittneyexline Then we're really in trouble because I'm in the same boat, too, even though I wrote this. Let's just say I kinda winged it hoping someone knowledgeable in review would point out my possibly incorrect assumptions :)

@saleem-latif Can you review these api-compact.yml changes and let me know if they make sense given the endpoints we're trying to put into the gateway?

enterprise/api_client/discovery.py
UmanShahzad: @itsjeyd Looks like my original idea was right:

`traverse_pagination (bool): Whether to traverse pagination or return paginated response..`

That's from the edx-platform's `openedx/core/lib/edx_api_utils.py.get_edx_api_data` function's docstring.

api-compact.yaml
```diff
@@ -62,9 +65,40 @@ apigateway_responses_with_mapping_template: &apigateway_responses_with_mapping_t
   500:
     statusCode: "500"
 
-produces: &produces
-  - "application/json"
-  - "application/csv"
+apigateway_responses_with_mapping_template_for_uuid: &apigateway_responses_with_mapping_template_for_uuid
```
saleem-latif: nit:

```
apigateway_responses_with_mapping_template_for_programs: &apigateway_responses_with_mapping_template_for_programs
```

api-compact.yaml
```diff
@@ -151,6 +195,37 @@ x-amazon-apigateway-integration-with-id-and-querystring-parameters: &apigateway_
     integration.request.querystring.page: "method.request.querystring.page"
     integration.request.querystring.page_size: "method.request.querystring.page_size"
 
+x-amazon-apigateway-integration-uuid-response: &apigateway_integration_uuid
```
saleem-latif: nit: 
```
x-amazon-apigateway-integration-with-mapping-template-for-programs: &apigateway_integration_with_mapping_template_for_programs
```

api-compact.yaml
```diff
@@ -167,7 +242,7 @@ endpoints:
           operationId: "get_enterprise_catalogs"
           responses: *responses
           x-amazon-apigateway-integration:
-            <<: *apigateway_integration
+            <<: *apigateway_integration_id
```
UmanShahzad: @saleem-latif Yep, but it may confuse someone seeing that it's not being used for the new `enterprise_catalogs` endpoint below :\

api-compact.yaml
UmanShahzad: @saleem-latif See above comment -- `uuid` is also used by the new `enterprise_catalogs/{uuid}` endpoint, so this may be confusing.

api-compact.yaml
```diff
@@ -120,8 +164,8 @@ page_size_qs_parameter: &page_size_qs_parameter
 
 # AWS API Gateway vendor extension point. This information is used
 #  by https://github.com/awslabs/aws-apigateway-importer.
-x-amazon-apigateway-integration: &apigateway_integration
-  responses: *apigateway_responses_with_mapping_template
+x-amazon-apigateway-integration-id-response: &apigateway_integration_id
```
UmanShahzad: @saleem-latif Yep, but same issue as above, unfortunately: it may be confusing to see this then not be used for the new `enterprise_catalogs` endpoint, which requires a `uuid`.

api-compact.yaml
UmanShahzad: @saleem-latif Unfortunately, this one is being used for the `enterprise_catalogs/{uuid}` endpoint, too :\

api-compact.yaml
UmanShahzad: @saleem-latif Makes sense -- but the thing is this could end up confusing someone who sees that it's not being used for the `/enterprise/api/enterprise_catalogs/{uuid}` endpoint, because that one doesn't use the `id` format.

UmanShahzad:
@saleem-latif Thanks a lot for the `api-compact.yml` review, it's very much appreciated.

The main issue seems to be that:

1. The old and new catalog endpoints share "catalog" in their name but use different parameters -- the former uses `id` and the latter uses `uuid`.
2. Both the new catalog endpoint and the programs endpoint use `uuid` as their detail parameters.

Due to the above reasons, it's hard to pick names that make sense for them unless we repeat code with the same body but different names, i.e. `apigateway_responses_with_mapping_template_for_catalogs` & `apigateway_responses_with_mapping_template_for_enterprise_customer_catalogs` with the same body, etc..

Any other recommendations? Or maybe my reasoning here isn't actually an issue?

enterprise/api_client/discovery.py
itsjeyd: @UmanShahzad Thanks for digging that up! The docstring doesn't do the best job of explaining the term IMHO, but your explanation matches my assumptions :)

UmanShahzad: :+1:

saleem-latif:
@UmanShahzad I see, I guess it makes sense using these. I can not come up with the anything that will fit the usage.

UmanShahzad:
@saleem-latif Great, thank you for giving your review, though!

tests/test_enterprise/api_client/test_discovery.py
```diff
@@ -314,6 +402,9 @@ def test_get_program_by_title_raise_multiple_match(self):
         (("c1", "c3", "c4"), [_make_run("c1"), _make_run("c2", "prof")], set()),
     )
     def test_get_common_course_modes(self, course_runs, response, expected_result):
+        """
+        Verify get_common_course_modes of CourseCatalogApiClient works as expected.
+        """
```
itsjeyd: @UmanShahzad Thanks for adding missing docstrings to the tests in this file!

enterprise/api/v1/views.py
```diff
@@ -308,21 +415,20 @@ def courses(self, request, enterprise_customer, pk=None):  # pylint: disable=inv
...
-        serializer: serializers.CourseSerializerExcludingClosedRuns
         """
         catalog_api = CourseCatalogApiClient(request.user)
         courses = catalog_api.get_paginated_catalog_courses(pk, request.GET)
 
-        # if API returned an empty response, that means pagination has ended.
+        # If the API returned an empty response, that means pagination has ended.
         # An empty response can also means that there was a problem fetching data from catalog API.
```
UmanShahzad: :+1:

test_utils/fake_catalog_api.py
```diff
@@ -5,8 +5,12 @@
 
 from __future__ import absolute_import, unicode_literals
 
+import uuid
+
 from six.moves import reduce as six_reduce
 
+FAKE_UUIDS = [str(uuid.uuid4()) for i in range(10)]  # pylint: disable=no-member
```
UmanShahzad: Well I tried to make it this way in case someone later needs to utilize many UUIDs where overlap isn't possible, but come to think of it I doubt someone would straight up need 10 different ones so I'll reduce it to 5.

test_utils/fake_catalog_api.py
```diff
@@ -566,6 +570,174 @@
...
+        }
+    ]
+}
+
+FAKE_SEARCH_ALL_COURSE_RESULT = {
+    "title": "edX Demonstration Course",
+    "min_effort": None,
+    "marketing_url": "course/course-v1:edX+DemoX+Demo_Course/about",
+    "image_url": "https://business.sandbox.edx.org/asset-v1:edX+DemoX+Demo_Course+type"
+                 "@asset+block@images_course_image.jpg",
```
UmanShahzad: I copied this straight from discovery-business.sandbox.edx.org/api-docs -- I didn't think it was harmful, as it's just a fake and temporary resource on the business sandbox itself. No bad-intentioned person would look at this and say, "found the weak spot".

test_utils/fake_catalog_api.py
```diff
@@ -566,6 +570,174 @@
...
+            "name": "",
+            "key": "edX",
+            "description": None,
+            "uuid": "12de950c-6fae-49f7-aaa9-778c2fbdae56",
+            "logo_image_url": None
+        }
+    ],
+    "staff_uuids": [],
+    "published": True,
+    "uuid": FAKE_UUIDS[-2],
```
UmanShahzad: @itsjeyd No good reason -- because I previously had 10 UUIDs I had to deal with, I decided to think of the ends rather than random numbers in the middle. But now with 5, I'll keep things positive.

test_utils/fake_catalog_api.py
```diff
@@ -566,6 +570,174 @@
...
+    "weeks_to_complete_min": None,
+    "type": "Professional Certificate",
+    "language": [
+        "English"
+    ],
+    "partner": "edx",
+    "subtitle": "Program Subtitle 1",
+    "status": "active",
+    "weeks_to_complete_max": None,
+    "aggregation_key": "program:" + FAKE_UUIDS[-2]
```
UmanShahzad: @itsjeyd No good reason -- because I previously had 10 UUIDs I had to deal with, I decided to think of the ends rather than random numbers in the middle. But now with 5, I'll keep things positive.

test_utils/fake_catalog_api.py
```diff
@@ -566,6 +570,174 @@
...
+        FAKE_SEARCH_ALL_COURSE_RESULT,
+        FAKE_SEARCH_ALL_SHORT_COURSE_RESULT,
+        FAKE_SEARCH_ALL_PROGRAM_RESULT,
+    ]
+}
+
+
+def get_search_results(querystring):
+    """
+    Fake implementation returning results based off of any object with keys or values that contain the query.
```
UmanShahzad: To be honest, the suggestion you gave tripped *me* up a little :P These things vary from person to person.

test_utils/fake_catalog_api.py
```diff
@@ -566,6 +570,174 @@
...
+        querystring (str): Query string to match against. Empty string returns all results.
+
+    Returns:
+        dict: Details of the object that matches the search results.
+    """
+    if querystring == '':
+        return FAKE_SEARCH_ALL_RESULTS
+
+    results = []
+    for result in FAKE_SEARCH_ALL_RESULTS.get('results', []):
```
UmanShahzad: But I believe if they did that, other tests down the line would fail and they'd notice :D

tests/test_enterprise/api/test_views.py
```diff
@@ -17,46 +17,18 @@
...
-        }
-    ]
-}
+from test_utils import TEST_USERNAME, APITest, factories, fake_catalog_api
+
+ENTERPRISE_CUSTOMER_COURSES_ENDPOINT = reverse('enterprise-customer-courses', (fake_catalog_api.FAKE_UUIDS[0],))
+ENTERPRISE_CUSTOMER_CATALOG_LIST_ENDPOINT = reverse('enterprise_catalogs-list')
+ENTERPRISE_CUSTOMER_CATALOG_DETAIL_ENDPOINT = reverse(
+    'enterprise_catalogs-detail',
+    (fake_catalog_api.FAKE_UUIDS[-1],)
```
UmanShahzad: Same thing as above - no good reason :) I'll go positive.

tests/test_enterprise/api/test_views.py
```diff
@@ -241,6 +213,21 @@ def test_user_view(self, factory, url, sorting_key, model_items, expected_json):
                 'consent_granted': True,
                 'course_id': 'course-v1:edX+DemoX+DemoCourse',
             }],
+        ),
+        (
+            factories.EnterpriseCustomerCatalogFactory,
+            reverse('enterprise-catalog-list'),
```
UmanShahzad: `ENTERPRISE_CUSTOMER_CATALOG_LIST_ENDPOINT` is not the same endpoint as this -- I actually tripped up here and forgot to change the model serialization endpoint to `enterprise-customer-catalog-list` :) No one caught it earlier but, alas!

tests/test_enterprise/api/test_views.py
```diff
@@ -627,20 +614,175 @@ def test_enterprise_customer_courses(
 
         assert response == expected
 
+    @ddt.data(
+        (
+            ENTERPRISE_CUSTOMER_CATALOG_DETAIL_ENDPOINT,
+            False, False,
```
UmanShahzad: :+1:

tests/test_enterprise/api/test_views.py
```diff
@@ -627,20 +614,175 @@ def test_enterprise_customer_courses(
...
+            True, True,
+            '',
+            fake_catalog_api.FAKE_SEARCH_ALL_RESULTS,
+        ),
+        (
+            ENTERPRISE_CUSTOMER_CATALOG_DETAIL_ENDPOINT,
+            True, False,
+            'no catalog for query so use empty string, i.e. get everything',
+            fake_catalog_api.FAKE_SEARCH_ALL_RESULTS,
+        ),
```
UmanShahzad: Oops, bad copy pasta.

tests/test_enterprise/api/test_views.py
```diff
@@ -627,20 +614,175 @@ def test_enterprise_customer_courses(
...
+            False, True,
+            '',
+            {
+                'detail': "User must be a staff user or "
+                          "associated with the endpoint's appointed common Enterprise Customer."
+            },
+        ),
+        (
+            ENTERPRISE_CUSTOMER_CATALOG_DETAIL_ENDPOINT,
+            True, False,  # Is a staff user, so it's allowed.
```
UmanShahzad: If no catalog exists however, then the user *must* be staff to get permission, because there would be nothing to cross-check with. So this test considers that case, and the code below is also thus required.

tests/test_enterprise/api/test_views.py
```diff
@@ -627,20 +614,175 @@ def test_enterprise_customer_courses(
...
+            'no catalog for query so use empty string, i.e. get everything',
```
UmanShahzad: :+1:

tests/test_enterprise/api/test_views.py
```diff
@@ -627,20 +614,175 @@ def test_enterprise_customer_courses(
...
+                'next': None,
+                'previous': None,
+                'results': [
+                    fake_catalog_api.FAKE_SEARCH_ALL_COURSE_RESULT,
+                    fake_catalog_api.FAKE_SEARCH_ALL_SHORT_COURSE_RESULT,
+                ],
+            },
+        ),
+        (
+            PROGRAMS_DETAIL_ENDPOINT,
```
UmanShahzad: Yes but similarly, one wouldn't expect the Programs client code to be under the `CourseCatalogApiClient` class (at least not me!), but it is. In the last PR, I had a significant refactor of this class into separate parts -- a class for targeting the Programs endpoint, one for the catalogs endpoint, etc. -- but Doug told to hold back on this, so I didn't bring that into this PR and decided to bite the nail and continue the current umbrella-ing of the Programs endpoint under anything "catalog".

tests/test_enterprise/api/test_views.py
```diff
@@ -627,20 +614,175 @@ def test_enterprise_customer_courses(
...
+                "previous": None,
+                "results": [fake_catalog_api.FAKE_SEARCH_ALL_PROGRAM_RESULT],
+            },
+        ),
+    )
+    @ddt.unpack
+    def test_enterprise_customer_catalogs_detail(self, url, has_linked_user, has_existing_catalog, querystring,
+                                                 expected_search_results):
+        """
+        Make sure the Enterprise Customer's Catalog view returns correctly detailed data.
```
UmanShahzad: Since you wrote it I'll take it, why not :P But normally these details would be explained in the actual code, I'd imagine, and the test's documentation should (usually) be the name of the test function itself, and usually that's plenty.

tests/test_enterprise/api/test_views.py
UmanShahzad: Ok from this I realized I have to change the test logic slightly: instead of `has_linked_user`, it should be `is_staff`, and the logic that creates the user should be inside of the `has_existing_catalog` branch.

tests/test_enterprise/api/test_views.py
```diff
@@ -627,20 +614,175 @@ def test_enterprise_customer_courses(
...
+            {
+                "count": 1,
+                "next": None,
+                "previous": None,
+                "results": [fake_catalog_api.FAKE_SEARCH_ALL_PROGRAM_RESULT],
+            },
+        ),
+    )
+    @ddt.unpack
+    def test_enterprise_customer_catalogs_detail(self, url, has_linked_user, has_existing_catalog, querystring,
```
UmanShahzad: I actually realized I have to change the test logic a little, and rename this to `is_staff` and move the EnterpriseCustomerUser creation down to the next `if` check.

tests/test_enterprise/api/test_views.py
```diff
@@ -627,20 +614,175 @@ def test_enterprise_customer_courses(
...
+                       "associated with the endpoint's appointed common Enterprise Customer."},
+        ),
+        (
+            ENTERPRISE_CUSTOMER_CATALOG_LIST_ENDPOINT,
+            False, True,
+            '',
+            {},
+            {'detail': "User must be a staff user or "
+                       "associated with the endpoint's appointed common Enterprise Customer."},
+        ),
```
UmanShahzad: Yep, plus a little change in the test logic to properly test this.

tests/test_enterprise/api/test_views.py
```diff
@@ -627,20 +614,175 @@ def test_enterprise_customer_courses(
...
+            {
+                'count': 3,
+                'next': 'http://testserver/enterprise/api/v1/enterprise_catalogs/?page=3&page_size=1',
+                'previous': 'http://testserver/enterprise/api/v1/enterprise_catalogs/?page_size=1',
+                'results': [fake_catalog_api.FAKE_SEARCH_ALL_COURSE_RESULT],
+            },
+        ),
+    )
+    @ddt.unpack
+    def test_enterprise_customer_catalogs_list(self, url, has_linked_user, has_existing_catalog, querystring,
```
UmanShahzad: See above -- changed logic a bit and the name entirely.

tests/test_enterprise/api/test_views.py
```diff
@@ -627,20 +614,175 @@ def test_enterprise_customer_courses(
...
+                enterprise_customer=enterprise_customer,
+                query=querystring
+            )
+        with mock.patch('enterprise.api.v1.views.CourseCatalogApiClient') as mock_catalog_api_client:
+            mock_catalog_api_client.return_value = mock.Mock(
+                get_paginated_search_results=mock.Mock(return_value=mocked_search_results),
+            )
+            response = self.client.get(url)
+            response = self.load_json(response.content)
+
```
UmanShahzad: There are many pre-existing tests in this file using this style. I think it's a fine style -- separating the assertion(s) from the data gathering.

tests/test_enterprise/api_client/test_discovery.py
```diff
@@ -24,26 +24,20 @@ class TestCourseCatalogApiInitialization(unittest.TestCase):
     @mock.patch('enterprise.api_client.discovery.CatalogIntegration')
     @mock.patch('enterprise.api_client.discovery.get_edx_api_data')
     def test_raise_error_missing_course_discovery_api(self, *args):  # pylint: disable=unused-argument
-        message = 'To get a Catalog API client, this package must be installed in an Open edX environment.'
-        with raises(NotConnectedToOpenEdX) as excinfo:
+        with self.assertRaises(NotConnectedToOpenEdX):
             CourseCatalogApiClient(mock.Mock(spec=User))
-        assert message == str(excinfo.value)
```
UmanShahzad: I've a pet peeve about this: there's no reason to check for the exact error message unless you're testing (for some reason) the exact error message you're getting. If all you're worried about *is that the exception's raised*, then only test for that. In this test (and honestly others in the codebase), what's being tested is that the error is raised (see the function name). Keeping the error string in there causes a potential maintenance headache.

tests/test_enterprise/api_client/test_discovery.py
UmanShahzad: :+1:

tests/test_enterprise/api/test_views.py
```diff
@@ -219,11 +191,11 @@ def test_user_view(self, factory, url, sorting_key, model_items, expected_json):
             reverse('enterprise-customer-entitlement-list'),
             itemgetter('enterprise_customer'),
             [{
-                'enterprise_customer__uuid': 'd1098bfb-2c78-44f1-9eb2-b94475356a3f',
+                'enterprise_customer__uuid': fake_catalog_api.FAKE_UUIDS[0],
                 'entitlement_id': 1
             }],
             [{
-                'enterprise_customer': 'd1098bfb-2c78-44f1-9eb2-b94475356a3f',
+                'enterprise_customer': fake_catalog_api.FAKE_UUIDS[0],
```
UmanShahzad: - It does read weird, and for that reason I think we should just import it directly rather than through `fake_catalog_api`. Besides, we just care that they're fake UUIDs, not where they came from, so that should be fine. With that said, I'm going to move `FAKE_UUIDS` to `test_utils/__init__.py` instead -- makes more sense to be there.
- It's not about being more DRY, it's that it's easier to manage which UUIDs you're using where. It's easier to see the difference between `FAKE_UUIDS[0]` and `FAKE_UUIDS[1]` vs. `d1098bfb-2c78-44f1-9eb2-b94475356a3f` and `d3098bfb-2c78-44f1-9eb2-b94475356a3f`. This is important when conflicts matter.
- Using different UUIDs is all about preventing conflict -- they're *supposed* to be, well, universally unique :P

What I'll do is move `FAKE_UUIDS` so they're portable from `test_utils` directly, thus handling the readability concern, and I think changing things back to `DEFAULT_ENTERPRISE_UUID` will cause an inconsistency that'll cause detecting conflicts difficult when writing new tests that require multiple UUIDs, including `DEFAULT_ENTERPRISE_UUID`.

UmanShahzad:
@itsjeyd @brittneyexline I've added some pretty detailed testing instructions in the PR post.

tests/test_enterprise/api/test_views.py
```diff
@@ -216,16 +217,16 @@ def test_user_view(self, factory, url, sorting_key, model_items, expected_json):
         ),
         (
             factories.EnterpriseCustomerCatalogFactory,
-            reverse('enterprise-catalog-list'),
+            reverse('enterprise-customer-catalog-list'),
```
itsjeyd: @UmanShahzad It might make sense to introduce a constant for this as well. That way, we'd have all relevant endpoints listed at the top. Naming them might be a bit awkward, but mentioning all relevant endpoints at the top of the file would make the duplication that we currently have (and want to get rid of) more visible.

test_utils/fake_catalog_api.py
itsjeyd: @UmanShahzad Fair enough, but can we try and find a wording that we're both OK with? That would increase the chances of others *not* getting confused, I think.

How about

> Fake implementation that returns result consisting of all objects whose keys or values contain ``querystring``.

That describes what the function does pretty accurately?

tests/test_enterprise/api/test_views.py
itsjeyd: @UmanShahzad Ah, OK. My main concern was to make whitespace consistent, so I agree with this. However, if a blank line above assertions is the predominant style, you should add a blank line above L714 above. (I had been looking at that assertion and this one, but not the remaining tests; so I just chose one of the two variants and suggested to change the other to make things consistent :))

UmanShahzad:
So I'm testing this on the sandbox, and although listing everything works, retrieving doesn't because apparently `querystring` isn't really expected to be a string.. :) It's supposed to be a dict because it's passed into the endpoint as keyword arguments like `**querystring`. Yay for naming.

I'll update things soon.

tests/test_enterprise/api/test_views.py
UmanShahzad: @itsjeyd I'm all for it!

tests/test_enterprise/api/test_views.py
```diff
@@ -678,24 +676,30 @@ def test_enterprise_customer_courses(
...
+    def test_enterprise_customer_catalogs_detail(self, url, is_staff, has_existing_catalog, querystring,
                                                  expected_search_results):
         """
-        Make sure the Enterprise Customer's Catalog view returns correctly detailed data.
+        Make sure the Enterprise Customer's Catalog view correctly returns details about specific catalogs based on
+        ``querystring``.
+
+        Search results should also take into account whether
+
+        * requesting user is linked to enterprise customer.
```
UmanShahzad: @itsjeyd You're right, done!

test_utils/fake_catalog_api.py
UmanShahzad: Yes, that's all good :) I had to change it just slightly with the new realization that `querystring` isn't supposed to be a string; see my newest comment on the PR.

tests/test_enterprise/api/test_views.py
UmanShahzad: Thanks, I didn't catch L714. :+1:

UmanShahzad:
Welp, that didn't fix things. I'll have to look at this tomorrow.

CHANGELOG.rst
```diff
@@ -14,6 +14,14 @@ Change Log
 Unreleased
 ----------
 
+[0.40.0] - 2017-08-08
+---------------------
+
+* Add Enterprise API Gateway for new Enterprise Catalogs and Programs endpoint.
```
itsjeyd: @UmanShahzad Typo: "endpoint" should be "endpoints" here.

CHANGELOG.rst
```diff
@@ -14,6 +14,14 @@ Change Log
...
+* Add /enterprise/api/v1/enterprise_catalogs/ endpoint.
+* Add /enterprise/api/v1/enterprise_catalogs/{uuid}/ endpoint.
```
itsjeyd: @UmanShahzad We still have underscores here. Should probably turn them into dashes after fbdf6222b8925de0ed2b4d2d302d014fb03fbdd4.

api-compact.yaml
```diff
@@ -10,6 +10,9 @@
 #   /enterprise/api/v1/catalogs/
 #   /enterprise/api/v1/catalogs/{id}/
 #   /enterprise/api/v1/catalogs/{id}/courses/
+#   /enterprise/api/v1/enterprise_catalogs/
+#   /enterprise/api/v1/enterprise_catalogs/{uuid}/
```
itsjeyd: @UmanShahzad We still have underscores here. Should probably turn them into dashes after fbdf6222b8925de0ed2b4d2d302d014fb03fbdd4.

enterprise/api/v1/views.py
```diff
@@ -138,10 +168,26 @@ def get_serializer_class(self):
         return serializers.EnterpriseCourseEnrollmentWriteSerializer
 
 
+class EnterpriseCustomerCatalogViewSet(EnterpriseReadOnlyModelViewSet):
+    """
+    API views for the ``enterprise_catalog`` API endpoint.
```
itsjeyd: @UmanShahzad We still have an underscore here. Should probably turn it into a dash after fbdf6222b8925de0ed2b4d2d302d014fb03fbdd4.

api-compact.yaml
```diff
@@ -199,3 +274,45 @@ endpoints:
           x-amazon-apigateway-integration:
             <<: *apigateway_integration_with_id_and_querystring_parameters
             uri: "https://${stageVariables.enterprise_host}/enterprise/api/v1/catalogs/{id}/courses/"
+
+    # /v1/enterprise_catalogs/
```
itsjeyd: @UmanShahzad We still have underscores in L278, L292, L294, L305. Should probably turn them into dashes after fbdf6222b8925de0ed2b4d2d302d014fb03fbdd4.

enterprise/api/v1/views.py
UmanShahzad: :+1:

api-compact.yaml
UmanShahzad: :+1:

api-compact.yaml
UmanShahzad: :+1:

CHANGELOG.rst
UmanShahzad: :+1:

CHANGELOG.rst
UmanShahzad: :+1:

UmanShahzad:
@itsjeyd Addressed the old naming issue; thanks for catching them. Since some changes were made to api.yml for this and we intend to test it on stage, I'll also just update the sandbox with this, in case. (Will do it later when I get access to AWS).

itsjeyd:
@UmanShahzad Looking good! You can squash again before deploying to the business sandbox. No need to keep the changes from 47c1e1ee4bd43120d9e10a3a0dab502b36a3fa08 ("Address review") in a separate commit as far as I'm concerned.

UmanShahzad:
@itsjeyd Yeah I'll be squashing these things before I send them to the sandbox (although the stuff that remains doesn't really need to be there except the api.yml changes).

itsjeyd:
@UmanShahzad I'm done testing the changes on the sandbox, all good! 🎉  I'll formally approve the PR once the API gateway changes have been verified. (Let me know if we need to merge and go to stage in order to do that. In that case I'll approve right away.)

CC @brittneyexline @saleem-latif

UmanShahzad:
@itsjeyd Great to hear! As for the gateway changes, @saleem-latif says @mattdrayer has to create a dev account for me on AWS to get access to their API Manager. I've pinged here now and on HipChat, awaiting access.

UmanShahzad:
@itsjeyd Sorry I didn't answer your question: Yes, this has to be tested on stage, so let's get this merged.

saleem-latif:
It seems like `limit`, `offset` query parameters are not working (on API Gateway or business sandbox).

for reproducing on gateway
```
curl -H "Authorization:JWT <jwt-key>" -X GET https://20enjrmi05.execute-api.us-west-2.amazonaws.com/edx_api/enterprise/v1/enterprise-catalogs/?limit=2&offset=2
```

for reproducing on sandbox

```
curl -H "Authorization:JWT <jwt-key>" -X https://business.sandbox.edx.org/enterprise/api/v1/enterprise-catalogs/?limi=2&offset=3
```

`page` and `page_size` querystring parameters are working as expected.

UmanShahzad:
@saleem-latif Thanks so much for helping out here. The issue is that `/search/all/` doesn't take `limit` and `offset`. You can confirm by comparing these two:

https://discovery-business.sandbox.edx.org/api/v1/search/all/?page=2&page_size=2&limit=2&offset=2

https://discovery-business.sandbox.edx.org/api/v1/search/all/?page=2&page_size=2

For me they return the same (except obviously different next/previous links because they keep any other query params).

So this is my fault -- I had included them inside the `api-compact.yml` file due to bad copy pasta. I removed them now.

CC @itsjeyd @brittneyexline

api-compact.yaml
```diff
@@ -214,18 +212,6 @@ x-amazon-apigateway-integration-with-uuid: &apigateway_integration_with_uuid_par
...
-  httpMethod: "GET"
-  type: "http"
-  requestParameters:
-    integration.request.header.Authorization: "method.request.header.Authorization"
-    integration.request.path.uuid: "method.request.path.uuid"
-    integration.request.querystring.limit: "method.request.querystring.limit"
-    integration.request.querystring.offset: "method.request.querystring.offset"
-    integration.request.querystring.page: "method.request.querystring.page"
-    integration.request.querystring.page_size: "method.request.querystring.page_size"
-
```
UmanShahzad: So this was unused.

UmanShahzad:
@saleem-latif Thanks! I've squashed from 97552a38a7c481ae7a57427f184911af59a3108f

@itsjeyd If you can give us formal approval here and @brittneyexline can give it one last look, we can merge.

UmanShahzad:
@brittneyexline @itsjeyd I've updated testing instructions to include testing the gateway. Currently @saleem-latif set us up with the `/enterprise-catalogs/` endpoint only, so I'll try to do the others.

UmanShahzad:
@saleem-latif I made one last addition to the API gateway: checking for page/page_size in the other endpoints. Since they all come in paginated data format (even the programs endpoint because it also points to `/search/all/`), these needed to be there.

UmanShahzad:
@brittneyexline Great! @itsjeyd will get to this tomorrow and we can go for it.

douglashall:
Nice work @UmanShahzad! The last step here is to add the new endpoints to the edx/api-manager swagger doc. This will allow us to make these new endpoints available via the API gateway. I have opened a PR [here](https://github.com/edx/api-manager/pull/62).

UmanShahzad:
@douglashall Ah, thanks, I didn't know about that. I've reviewed the PR!

enterprise/api/v1/urls.py
```diff
@@ -7,16 +7,15 @@
...
 router = DefaultRouter()  # pylint: disable=invalid-name
 router.register("site", views.SiteViewSet, 'site')
 router.register("auth-user", views.UserViewSet, 'auth-user')
-router.register(
-    "enterprise-course-enrollment",
-    views.EnterpriseCourseEnrollmentViewSet,
-    'enterprise-course-enrollment'
-)
+router.register("enterprise-customer-catalog", views.EnterpriseCustomerCatalogViewSet, 'enterprise-customer-catalog')
+router.register("enterprise-catalogs", views.EnterpriseCustomerCatalogApiViewSet, 'enterprise-catalogs')
```
douglashall: @UmanShahzad Why did we add two separate routes/ViewSets here?

`/enterprise/api/v1/enterprise-catalogs` should return the list of EnterpriseCatalog's that the requesting user has access to. As currently implemented, it returns the results of an unfiltered query to the catalog's search endpoint.

UmanShahzad: @douglashall The `enterprise-customer-catalog` endpoint is for serializing the _model_ -- `EnterpriseCustomerCatalog`. The `enterprise-catalogs` endpoint is the service endpoint which wraps over discovery.

douglashall: Right, apologies if this was not clear in our previous discussions. `/enterprise/api/v1/enterprise-catalogs` should return the serialized list of `EnterpriseCustomerCatalog` models that the requesting user has access to. We have no need for an endpoint that returns the results of an unfiltered query to the catalog's search endpoint.

How should we proceed? Is this something you can address as a part of the program landing page work or should I file a separate ticket?

FYI @brittneyexline

UmanShahzad: @douglashall Hmm, yeah from the ticket and the discussions I thought this endpoint would be the "list" version of its corresponding more detailed `/enterprise/api/v1/enterprise-catalogs/{uuid}/` endpoint, which would filter things (i.e. for details).

In terms of implementation, this would be a matter of changing the `list` function at https://github.com/edx/edx-enterprise/blob/master/enterprise/api/v1/views.py#L347 -- instead of querying the catalog client, we'd filter `EnterpriseCustomerCatalog` instances which the requesting user has access to, serialize them, and return the response.

I can do that in the Programs ticket but it's getting bigger and bigger and it'll slow me down a bit. *If* we have capacity for someone else to pick it up this sprint that'd be great, otherwise I'll fit it in as a separate commit, yeah. I've already refactored the Programs endpoint to point to its own dedicated view set, and that was significant in and of itself, so it might make sense to allocate this change to me as well.

douglashall: Awesome. Yes, let's incorporate this change into the program landing page work as a separate commit. Feel free to reach out for help if you need it. Thanks.

Also, I'll record something about this conversation on the program landing page ticket.

douglashall: On second thought, I'm going to create a follow up ticket linked to ENT-341 and pull it into the sprint. I think that is more appropriate.

UmanShahzad: @douglashall Sounds good to me :)

PR #167, opened 2017-08-08
[ENT-495] Create a unified Data Sharing Consent model

UmanShahzad:
**Description:** This ticket creates a new `DataSharingConsent` model that represents the abstract concept of consent in preparation for use in future tickets, including [ENT-496](https://openedx.atlassian.net/browse/ENT-496).

**JIRA:** [ENT-495](https://openedx.atlassian.net/browse/ENT-495)

**Dependencies:** [ENT-496](https://openedx.atlassian.net/browse/ENT-496) depends on this and is in the same sprint.

**Merge deadline:** ASAP due to the dependency above.

**Testing instructions:**

1. See tests & test coverage for confidence.
2. See that the model(s) match the spec in the ticket.

There's not much else that can be done to test this because it's just adding models and nothing else -- it is [ENT-496](https://openedx.atlassian.net/browse/ENT-496) that will be using these models.

**Merge checklist:**

- [ ] All reviewers approved
- [x] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:** I'm worried the `ProxyDataSharingConsent` "model" doesn't really match the spec, so I'll need @haikuginger to double check for any mistakes there.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=167"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

consent/models.py
```diff
@@ -3,4 +3,150 @@
...
+        app_label = 'consent'
+        unique_together = (("enterprise_customer", "username"),)
+
+    enterprise_customer = models.ForeignKey(
+        EnterpriseCustomer,
+        blank=False,
+        null=False,
+        related_name='enterprise_customer_consent',
+        on_delete=models.deletion.CASCADE
+    )
```
brittneyexline: is there a reason we don't use a foreign key to EnterpriseCustomerUser instead of EnterpriseCustomer + username?

consent/models.py
```diff
@@ -3,4 +3,150 @@
...
+    """
+
+    class Meta:
+        """
+        Meta class for the ``Consent`` model.
+        """
+
+        abstract = True
+        app_label = 'consent'
+        unique_together = (("enterprise_customer", "username"),)
```
brittneyexline: if we make this unique, won't we not be able to have different records per course?

consent/models.py
brittneyexline: I realize that this was discussed on the jira ticket, but it's still not clear to me, so maybe @haikuginger can shed some light.

UmanShahzad: @brittneyexline Indeed @haikuginger can answer this better than I can since he originally wrote the spec.

consent/models.py
UmanShahzad: @brittneyexline That's true; I'll have to remove that. Should `username` and `course_id` be unique together, instead?

consent/models.py
haikuginger: Flexibility in the interest of being able to spin this into an IDA in future. Username is a unique indexed field on the Django user, so we don't lose much speed by not referencing the ID directly in the form of a ForeignKey.

consent/models.py
haikuginger: I would make all three fields unique together.

UmanShahzad: @haikuginger :+1:

UmanShahzad:
@haikuginger I've added some admin portal detail as well now. Take a look at and let me know if this matches your vision of it, give or take a bit :)

UmanShahzad:
@haikuginger Just need your approval and we can go for it! (Although at this point I hope we merge #164 first because it also has a migration, and I don't want to have to keep updating that ticket :D).

consent/admin/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+
+    search_fields = (
+        'enterprise_customer__name',
+        'enterprise_customer__uuid',
+        'username',
+        'course_id',
+    )
+
+    inlines = (
+        HistoricalDataSharingConsentInlineAdmin,
```
haikuginger: There isn't actually a ForeignKey between the DSC and hDSC, so we either need to do some fancy magic here to get the right queryset, or we need to skip this entirely. For now, I recommend the latter. This is actually preventing the migration from working on my devstack.

consent/models.py
```diff
@@ -3,4 +3,150 @@
...
+        on_delete=models.deletion.CASCADE
+    )
+    username = models.CharField(
+        max_length=255,
+        blank=False,
+        null=False,
+        help_text=_("Name of the user whose consent state is stored.")
+    )
+    granted = models.NullBooleanField(help_text=_("Whether consent is granted."))
+    exists = models.BooleanField(default=True, editable=False)
```
haikuginger: We don't need a database field for this - we can assume `obj.exists` if `obj.pk`.

consent/models.py
```diff
@@ -3,4 +3,150 @@
...
+    """
+
+    class Meta(Consent.Meta):
+        """
+        Meta class for the ``DataSharingConsent`` model.
+        """
+
+        abstract = False
+        verbose_name = _("Data Sharing Consent")
+        verbose_name_plural = _("Data Sharing Consents")
```
haikuginger: Nit: `Data Sharing Consent Record` and `Data Sharing Consent Records`?

tests/test_models.py
```diff
@@ -864,6 +867,120 @@ def test_string_conversion(self, method):
...
+
+    def test_commit(self):
+        """
+        Test that ``ProxyDataSharingConsent``'s ``commit`` method properly creates/saves/returns a new
+        ``DataSharingConsent`` instance, or returns ``None`` for any validation errors (i.e. conflict).
+        """
+        new_dsc = self.proxy_dsc.commit()
+        no_dsc = self.proxy_dsc.commit()
+        dsc = DataSharingConsent.objects.all()
+        self.assertEqual(len(dsc), 1)
```
haikuginger: I'd do `assert DataSharingConsent.objects.count() == 1`. Note that we use py.test, so we can use bare asserts and get the test metadata that stuff like `self.assertEqual` usually gives us.

tests/test_models.py
```diff
@@ -864,6 +867,120 @@ def test_string_conversion(self, method):
...
+        self.assertIsNone(no_dsc)
+
+    @ddt.data(
+        str, repr
+    )
+    def test_string_conversion(self, method):
+        """
+        Test ``ProxyDataSharingConsent`` conversion to string
+        """
+        expected_to_str = (
```
haikuginger: Given this is an expectation, and that we're only testing with the one set of values, I'd prefer writing out the raw string literal.

tests/test_models.py
```diff
@@ -864,6 +867,120 @@ def test_string_conversion(self, method):
...
+        Test ``DataSharingConsent`` conversion to string
+        """
+        dsc = DataSharingConsentFactory(
+            enterprise_customer=EnterpriseCustomerFactory(
+                name='rich_enterprise'
+            ),
+            username='lowly_bob',
+            course_id='hard_course_2017'
+        )
+        expected_to_str = (
```
haikuginger: As above.

UmanShahzad:
@haikuginger Thanks for the review -- I've made the changes.

consent/models.py
```diff
@@ -3,4 +3,153 @@
...
+        """
+        Perform the query and returns a single object matching the given keyword arguments.
+
+        This customizes the queryset to return an instance of ``ProxyDataSharingConsent`` when
+        the searched-for ``DataSharingConsent`` instance does not exist.
+        """
+        try:
+            return super(DataSharingConsentQuerySet, self).get(*args, **kwargs)
+        except DataSharingConsent.DoesNotExist:
+            return ProxyDataSharingConsent(**kwargs)
```
haikuginger: Realizing in testing that we may run into issues here - specifically, we might make a composite query like `DataSharingConsent.objects.get(username='john', course_id='course', enterprise_customer__uuid='this-is-a-real-uuid-promise')`. I think we can probably defer that for right now and handle as part of the API buildup, though.

consent/models.py
```diff
@@ -3,4 +3,156 @@
...
+        help_text=_("Name of the user whose consent state is stored.")
+    )
+    granted = models.NullBooleanField(help_text=_("Whether consent is granted."))
+
+    @property
+    def exists(self):
+        """
+        Return whether the instance exists or not.
+        """
+        return self.pk
```
haikuginger: Nit: cast this to a bool.

UmanShahzad: Good catch.

consent/models.py
UmanShahzad: That's something I've already run into as part of the refactor, in fact, yeah. Have to think of a way around it.

UmanShahzad:
@haikuginger Addressed nit and squashed from 5c59df57acbc62e2e03b8b7ea7bd2ecf1e5b8cf6

consent/models.py
UmanShahzad: @haikuginger Here's a temporary solution I have:

```python
    def __init__(self, enterprise_customer=None, username='', course_id='', granted=False, **kwargs):
        """
        Initialize a proxy version of ``DataSharingConsent`` which behaves similarly but does not exist in the DB.
        """
        ec_keys = {}
        for key in kwargs:
            if str(key).startswith('enterprise_customer__'):
                ec_keys.update({key[len('enterprise_customer__'):]: kwargs[key]})

        enterprise_customer_from_db = None
        if ec_keys:
            try:
                enterprise_customer_from_db = EnterpriseCustomer.objects.get(**ec_keys)
            except EnterpriseCustomer.DoesNotExist:
                enterprise_customer_from_db = None

        self.enterprise_customer = enterprise_customer_from_db or enterprise_customer
        self.username = username
        self.course_id = course_id
        self.granted = granted
        self.exists = False
```

UmanShahzad: @haikuginger This test passes:

```python
    def test_create_new_proxy_with_composite_query(self):
        """
        Test that we can use composite queries for enterprise customers.
        """
        proxy_dsc = ProxyDataSharingConsent(
            enterprise_customer__name='rich_enterprise',
            enterprise_customer__enable_data_sharing_consent=True
        )
        the_only_enterprise_customer = EnterpriseCustomer.objects.all().first()
        assert the_only_enterprise_customer == proxy_dsc.enterprise_customer
```

UmanShahzad:
Rebased on master from 6f63b913ab9b2cd171324b73cffbf24beba96ee5

PR #179, opened 2017-09-05
ENT-615 Discount Text on Course Enrollment Landing Page

asadiqbal08:
**Description:** Whenever the enterprise provides a discount to learners, that should be indicated in close proximity to the discount. 

@brittneyexline @zubair-arbi @douglashall can u take a look on this small change. 

screen shot 2017-09-05 at 2 33 49 pm

screen shot 2017-09-06 at 3 15 39 pm

**JIRA:** [ENT-615](https://openedx.atlassian.net/browse/ENT-615)

**Testing instructions:**

1. Apply a discount to an enterprise verified course.
2. Visit the enterprise enrollment landing page.
3. See the text "Discount provided by [Enterprise]" below the strike value as shown in the attachment.

- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag


enterprise/views.py
```diff
@@ -600,6 +600,7 @@ class CourseEnrollmentView(View):
         'expected_learning_items_text': _("What you'll learn"),
         'course_full_description_text': _('About This Course'),
         'staff_text': _('Course Staff'),
+        'discount_text': _('Discount provided by')
```
UmanShahzad: Not sure how much it matters, but we can probably format `_('Discount provided by {enterprise_customer_name}').format(enterprise_customer_name=enterprise_customer.name)` in context inside the method and make the template slightly simpler.

enterprise/templates/enterprise/enterprise_course_enrollment_page.html
```diff
@@ -72,6 +72,7 @@ <h2 class="course-confirmation-title">{{ confirmation_text }}</h2>
                 {{ price_text }}:
                 {% if course_mode.final_price and course_mode.original_price != course_mode.final_price %}
                   <strike>{{ course_mode.original_price }}</strike> {{ course_mode.final_price }}
+                  <div>{{discount_text}} <b>{{enterprise_customer.name}}</b></div>
```
UmanShahzad: See below: we can probably make this just `{discount_text}` if we pass in the full string at once.

enterprise/views.py
asadiqbal08: It is a good idea but I added the string to `STATIC_TEXT_FORMAT` that does not have the `enterprise_customer` info in it. otherwise we need to update it out of that block

UmanShahzad: Yep, that's what I meant -- to put it into a context where `enterprise_customer` is reachable. But the new version looks good to me :)

UmanShahzad:
@asadiqbal08 Thanks! I actually noticed that we don't have a check for this in the tests' context. Maybe add something there if you have time?

UmanShahzad:
@brittneyexline I think he has done that at https://github.com/edx/edx-enterprise/pull/179/files#diff-a3975c533f09d91598f0b77b55f009a0R127 -- so we should be good unless there's another place to add this?

brittneyexline:
ah thanks @UmanShahzad - somehow i had missed that :woman_facepalming:

PR #180, opened 2017-09-05
ENT-496 Posting to enterprise-course-enrollment shouldn't write consent_granted anymore

brittneyexline:
**Description:** I discovered an error on the business sandbox where the user wasn't being enrolled in the course when doing a verified purchase. The root cause was that we were calling the enterprise-course-enrollment endpoint attempting to write the consent_granted field. However, the record existed with consent_granted set to NULL, as expected from the changes from ENT-496, but the api save method used get_or_create with that and threw an DB integrity error. There is an associated edx-platform PR to address how we call this (https://github.com/edx/edx-platform/pull/15950)

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

Pre these changes:
Go through the enterprise landing page and select the verified track. After purchasing, return to the dashboard and see that the user is not enrolled in the course.

Apply these changes:
Go through the same flow as above, but when you come to the dashboard, you should be enrolled in the course.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -161,12 +161,10 @@ class EnterpriseCourseEnrollmentWriteSerializer(serializers.ModelSerializer):
         Save the model with the found EnterpriseCustomerUser.
         """
         course_id = self.validated_data['course_id']
-        consent_granted = self.validated_data['consent_granted']
 
         models.EnterpriseCourseEnrollment.objects.get_or_create(
             enterprise_customer_user=self.enterprise_customer_user,
             course_id=course_id,
-            consent_granted=consent_granted,
         )
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.43.1] - 2017-09-05
+---------------------
+
+* Remove support for writing consent_granted in enterprise-course-enrollment api.
+
 [0.43.0] - 2017-08-31
 ---------------------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.43.0"
+__version__ = "0.43.1"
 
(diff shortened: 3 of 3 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -161,12 +161,10 @@ def save(self):  # pylint: disable=arguments-differ
         Save the model with the found EnterpriseCustomerUser.
         """
         course_id = self.validated_data['course_id']
-        consent_granted = self.validated_data['consent_granted']
 
         models.EnterpriseCourseEnrollment.objects.get_or_create(
             enterprise_customer_user=self.enterprise_customer_user,
             course_id=course_id,
-            consent_granted=consent_granted,
         )
```
douglashall: Do we still need to set the `consent_granted` flag here though if it was passed?

brittneyexline: I am guessing no because we aren't using this field anymore to indicate consent, so I think it should always be written as NULL or ideally removed altogether from the model. That was the intent I gleaned from the migration to the new consent architecture, but I could be wrong. Otherwise, we should change this call to update_or_create to allow updates to consent_granted which fail when the record exists for a user/course.

douglashall: Makes sense. Do we have a Jira ticket for removing the consent_granted field from the EnterpriseCourseEnrollment record in the backlog? @haikuginger

UmanShahzad: The intent of the migration to the new consent architecture included deprecating reliance on consent through the enrollment model, but I had spoken with @haikuginger about completely removing this and related to-be-deprecated consent features, but it was decided that that should be considered in a later story.

brittneyexline:
I've deployed this and the edx-platform change to the business sandbox and verified that the behavior is fixed.

enterprise/api/v1/serializers.py
haikuginger: Yeah, to my knowledge, we do not have the needed followup story.

PR #182, opened 2017-09-06
[ENT-587] More gracefully handle errors from hitting the Ecommerce API (and others)

UmanShahzad:
**Description:** This PR moves error handling logic for our external API clients into the clients themselves, rather than having try/except blocks all over the codebase.

**JIRA:** [ENT-587](https://openedx.atlassian.net/browse/ENT-587)

**Merge deadline:** Before the end of the week.

**Testing instructions:**

1. Since this is just a refactor and doesn't introduce any new functionality (but does add test coverage that was lost due to moving logic around), you can either rely on the tests or just give the course enrollment landing page a spin; see if the API calls are working properly and that in exceptional cases like missing certain data, we get the same behavior as last time.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:** I do wonder if the defaults I return in some cases are appropriate. So please do a double-check there.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=182&repo=edx%2Fedx-enterprise"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

UmanShahzad:
I basically decided to cover all of our external API calls in `enterprise.views.py` to put enough work into this to match the estimated story points... :P

enterprise/api_client/discovery.py
```diff
@@ -377,13 +383,16 @@ def _load_data(self, resource, default=DEFAULT_VALUE_SAFEGUARD, **kwargs):
...
-        return result or default_val
+        try:
+            return get_edx_api_data(
+                api_config=CatalogIntegration.current(),
+                resource=resource,
+                api=self.client,
+                **kwargs
+            ) or default_val
+        except HttpClientError as exc:
+            LOGGER.error('Failed to load data from resource %s with kwargs %s due to: %s', resource, kwargs, str(exc))
```
douglashall: Can we change these to `LOGGER.exception` calls instead so we get the full stack trace in the logs?

UmanShahzad: Yes, sure!

enterprise/api_client/discovery.py
```diff
@@ -377,13 +383,16 @@ def _load_data(self, resource, default=DEFAULT_VALUE_SAFEGUARD, **kwargs):
...
-        )
-        return result or default_val
+        try:
+            return get_edx_api_data(
+                api_config=CatalogIntegration.current(),
+                resource=resource,
+                api=self.client,
+                **kwargs
+            ) or default_val
+        except HttpClientError as exc:
```
douglashall: Does this need to be the `(ConnectionError, SlumberBaseException, Timeout)` trifecta like you have in the [ecommerce](https://github.com/edx/ecommerce/pull/1468/files#diff-0aa9e695700bf5edb4df6ef6ac088d71R87) PR?

UmanShahzad: I'm honestly not sure about it. @haikuginger has more knowledge about Slumber -- he would know.

enterprise/api_client/discovery.py
```diff
@@ -411,3 +420,13 @@ def __init__(self):
                 raise ImproperlyConfigured(_("The configured CatalogIntegration service user does not exist."))
         else:
             raise ImproperlyConfigured(_("There is no active CatalogIntegration."))
+
+    @classmethod
```
douglashall: I don't think you want the `@classmethod` decorator here.

UmanShahzad: I wanted to encapsulate handling the `ImproperlyConfigured` error into this function if possible to make the code in our views easy to read. For that, I needed to instantiate it here. Or do you mean make it a static method?

douglashall: Ah, I see now. Makes sense. Thanks for the clarification.

enterprise/api_client/discovery.py
douglashall: @haikuginger What do you think? Seems like we should at least change this to `SlumberBaseException`.

haikuginger: @douglashall, as far as I can tell, Slumber doesn't catch those other two, so let's include them.

UmanShahzad: :+1:

UmanShahzad:
@haikuginger @douglashall Added more exceptions for catching. Anything in particular that's left here?

UmanShahzad:
@douglashall @brittneyexline Look good to you guys? I'd like to rebase the programs landing page on this to benefit from it if possible before we merge that.

PR #187, opened 2017-09-13
ENT-594 Automate Enterprise Data Reporting Configuration

brittneyexline:
**Description:** This PR introduces a new django app called enterprise_reporting which is meant to hold functionality pertaining to generating and sending reports to enterprise customers without any particular integration point (which would live in integrated_channels). This also introduces the EnterpriseCustomerReportingConfiguration model, which can be created in django admin to specify information needed for sending the report via email to the enterprise customer.

There is also an api endpoint for reading the model that behaves as other read only endpoints for enterprise: <base_lms_url>/enterprise/api/v1/enterprise_customer_reporting

It can be filtered by enterprise customer by appending the query parameter: enterprise_customer=<enterprise_uuid>

**JIRA:** https://openedx.atlassian.net/browse/ENT-594

**Dependencies:** 

**Merge deadline:** 

**Installation instructions:** As expected, but make sure to run migrations.

**Testing instructions:**
There should now be an enterprise reporting section in django admin with one model, Enterprise Customer Reporting Configuration. You should be able to create one linked to an existing enterprise customer. Based on the frequency that is configured, the day_of_month and day_of_week fields will be validated, and set to none if not relevant for the frequency that is configured. The password field is not editable and is generated once the model is saved, at which point it is visible.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


enterprise_reporting/models.py
```diff
@@ -0,0 +1,131 @@
...
+        elif self.frequency == self.FREQUENCY_TYPE_WEEKLY:
+            if not self.day_of_week:
+                raise ValidationError({'day_of_week': _('Day of week must be set if the frequency is weekly.')})
+            self.day_of_month = None
+        elif self.frequency == self.FREQUENCY_TYPE_MONTHLY:
+            if not self.day_of_month:
+                raise ValidationError({'day_of_month': _('Day of month must be set if the frequency is monthly.')})
+            self.day_of_week = None
+        else:
+            raise ValidationError('Frequency must be set to either daily, weekly, or monthly.')
```
UmanShahzad: Should we translate this string too?

tox.ini
```diff
@@ -65,17 +65,19 @@ commands =
...
+    pylint --py3k enterprise_reporting
     pylint --py3k tests
     pylint --py3k test_utils
     rm tests/__init__.py
-    pycodestyle enterprise consent integrated_channels tests test_utils
-    pydocstyle enterprise consent integrated_channels tests test_utils
-    isort --check-only --recursive tests test_utils enterprise consent manage.py setup.py test_settings.py
+    pycodestyle enterprise consent integrated_channels enterprise_reporting tests test_utils
+    pydocstyle enterprise consent integrated_channels enterprise_reporting tests test_utils
+    isort --check-only --recursive tests test_utils enterprise consent integrated_channels enterprise_reporting manage.py setup.py test_settings.py
```
UmanShahzad: Ah wow, didn't know we weren't checking for this. Can you also add these to the `isort` make target?

UmanShahzad:
I know this WIP, but so far looks good to me, although I haven't manually tested it yet according to the current test instructions.

enterprise_reporting/client.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+"""
+Client for connecting to SAP SuccessFactors.
```
haikuginger: Docstring?

enterprise_reporting/client.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def fetch_results(self, query):
+        """
+        Fetches all the results for a query using the current connection.
+        """
+
+    def stream_results(self, query):
+        """
+        Streams the results for a query using the current connection.
+        """
```
haikuginger: Newline

enterprise_reporting/client.py
```diff
@@ -0,0 +1,37 @@
...
+class VerticaClient(object):
+    """
+    Client for connecting to Vertica.
+    """
+
+    def __init__(self):
+        """
+        Instantiate a new client using the Django settings to determine the vertica credentials.
+        If there are none configured, throw an exception.
+        """
```
haikuginger: Insert either `pass` or `raise NotImplemented`?

enterprise_reporting/management/commands/send_enteprise_reports.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+        super(Command, self).add_arguments(parser)
+
+    def handle(self, *args, **options):
+        """
+        Send csv reports to the EnterpriseCustomer(s) configured to receive data.
+        """
+        enterprise_customer = options['enterprise_customer']
+
+        if enterprise_customer:
```
haikuginger: I think this could probably be simplified a bit overall:

```python
if enterprise_customer:
    filter_params = {'enterprise_customer__uuid': enterprise_customer, 'active': True}
else:
    filter_params = {'active': True}

reporting_configs = EnterpriseCustomerReportingConfiguration.objects.filter(**filter_params)
if enterprise_customer and not reporting_configs.exists():
    raise CommandError('The enterprise {} does not have a reporting configuration.'.format(enterprise_customer))

for reporting_config in reporting_configs:
    if enterprise_customer or reporting_config.is_current_time_in_schedule():
        send_data_task.delay(reporting_config)
```

Do your own thing, of course, but I think it'd be good to not duplicate the database query and task-launching logic.

enterprise_reporting/models.py
```diff
@@ -0,0 +1,157 @@
...
+    )
+    hour_of_day = models.SmallIntegerField(
+        blank=False,
+        null=False,
+        verbose_name=_("Hour of Day"),
+        help_text=_("The hour of the day to send the report, in Eastern Standard Time (EST). "
+                    "This is required for all frequency settings."),
+        validators=[MinValueValidator(0), MaxValueValidator(23)]
+    )
+    password = models.CharField(
```
haikuginger: What is this password _for_? I'm _very_ squeamish about having any sort of password stored in the database as plain text.

brittneyexline: Great question, open to implementing differently, especially while it's still easy to change. Essentially we need a password for encrypting the csv that can be shared with the customer separately. This job needs access to that field and it needs to be configured per enterprise customer.

haikuginger: Could we do offline generation of a public/private keypair? Then, we can store the public key in the DB and use it for encryption operations, and provide the private key to the customer to safeguard.

brittneyexline: The implementation we agreed upon for this initial story was to send password protected zip files. I'm planning to use this package to do that: https://github.com/smihica/pyminizip
Maybe this is something we can talk about after standup tomorrow, because I agree this is not ideal but my impression from how we groomed this story was that we were not going to go the full public/private key route, but I would like to not store a password in the database.

haikuginger: Gotcha; that makes sense. Yeah, I don't want to explode the scope too much - does @fredsmith have any thoughts on specific implementation?

enterprise_reporting/management/commands/send_enteprise_reports.py
```diff
@@ -0,0 +1,86 @@
+# -*- coding: utf-8 -*-
```
haikuginger: Filename. Needs to be changed to `send_enterprise_reports.py`.

brittneyexline:
@douglashall yea, i was just thinking the same thing as i started to get into the api work. will do!

enterprise/api/v1/serializers.py
```diff
@@ -391,3 +391,16 @@ def to_representation(self, instance):
...
+class EnterpriseCustomerReportingConfigurationSerializer(serializers.ModelSerializer):
+    """
+    Serializer for EnterpriseCustomerReportingConfiguration model.
+    """
+
+    class Meta:
+        model = models.EnterpriseCustomerReportingConfiguration
+        fields = (
+            'enterprise_customer', 'active', 'delivery_method', 'email', 'frequency', 'day_of_month', 'day_of_week',
+            'hour_of_day', 'password'
```
brittneyexline: so this password thing (as previously discussed) feels wrong. I'm thinking maybe there is a shared secret we can configure between enterprise/edxapp and jenkins so that this can be encrypted and then decrypted for use when the script runs?

enterprise/api/v1/urls.py
```diff
@@ -24,3 +24,8 @@
     'enterprise-customer-entitlement',
 )
 router.register("catalogs", views.EnterpriseCourseCatalogViewSet, 'catalogs')
+router.register(
+    "enterprise-customer-reporting",
+    views.EnterpriseCustomerReportingConfigurationViewSet,
+    'enterprise-customer-reporting',
```
UmanShahzad: Since we're going to be doing underscores anyway, we should probably go ahead and do it here too.

enterprise/api/v1/serializers.py
douglashall: How will the Enterprise administrator get access to the password so that they can share it with the Enterprise customer receiving the data?

brittneyexline: i guess we could add a property to the django admin that would decrypt the password for viewing if someone goes into the detail view?

douglashall: Oh nevermind. You are talking specifically about making the password available in the data returned from the API endpoint.

douglashall: Yeah, you could store the shared secret in [edx-ops/business-secure](https://github.com/edx-ops/business-secure). I guess if we are going to do that, why store the password in the database at all?

brittneyexline: yea, we could just store the encrypted password, which would be better. i'll start exploring that route.

enterprise/models.py
```diff
@@ -1028,3 +1030,143 @@ def __repr__(self):
...
+    )
+    hour_of_day = models.SmallIntegerField(
+        blank=False,
+        null=False,
+        verbose_name=_("Hour of Day"),
+        help_text=_("The hour of the day to send the report, in Eastern Standard Time (EST). "
+                    "This is required for all frequency settings."),
+        validators=[MinValueValidator(0), MaxValueValidator(23)]
+    )
+    iv = models.BinaryField(   # pylint: disable=invalid-name
```
douglashall: What does this do?

enterprise/models.py
```diff
@@ -1028,3 +1030,143 @@ def __repr__(self):
...
+        blank=True,
+        null=True,
+        choices=DAYS_OF_WEEK,
+        verbose_name=_("Day of Week"),
+        help_text=_("The day of the week to send the report. "
+                    "This field is required and only valid when the frequency is weekly."),
+    )
+    hour_of_day = models.SmallIntegerField(
+        blank=False,
+        null=False,
```
douglashall: You can remove
```
blank=False,
null=False
```
as they are the default values.

enterprise/models.py
douglashall: Nevermind. I get it. It would be good to add `help_text` or a comment explaining what this field is used for. Also, I would go with `password_encryption_initialization_vector` as a field name.

PR #188, opened 2017-09-13
Make enrollment emails marketing site aware

tuchfarber:
**Description:** 
Currently the enterprise emails assume they are being ran with a
marketing site. This allows for instances without a marketing site
(e.g. devstack, sandbox, WL) to send emails correctly.

**JIRA:**
[WL-1124]

None

None. Sooner the better?

**Installation instructions:** List any non-trivial installation instructions.
None

**Testing instructions:**
1. Open the enterprise customer admin page in LMS
2. Click "Manage Learners" in the top right
3. Add a leaner to a program or course
4. Note in email how the link is formatted correctly based on whether the site has an admin page or not.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.
It's goddamn perfect.

### DIFF
--- a/enterprise/admin/views.py
+++ b/enterprise/admin/views.py
@@ -19,6 +19,7 @@ from django.core.exceptions import ValidationError
 from django.db.models import Q
 from django.http import HttpResponse, HttpResponseRedirect
 from django.shortcuts import get_object_or_404, render
+from django.utils.http import urlquote
 from django.utils.translation import ugettext as _
 from django.utils.translation import ungettext
 from django.views.generic import View
@@ -43,7 +44,7 @@ from enterprise.models import (
     PendingEnrollment,
     PendingEnterpriseCustomerUser,
 )
-from enterprise.utils import get_reversed_url_by_site, send_email_notification_message
+from enterprise.utils import get_configuration_value_for_site, send_email_notification_message
 
 
 class TemplatePreviewView(View):
@@ -508,16 +509,13 @@ class EnterpriseCustomerManageLearnersView(View):
                 ).format(course_id)
             )
             return
-        course_url = course_details.get('marketing_url')
-        if course_url is None:
-            # If we didn't get a useful path to the course on a marketing site from the catalog API,
-            # then we should build a path to the course on the LMS directly.
-            course_url = get_reversed_url_by_site(
-                request,
-                enterprise_customer.site,
-                'about_course',
-                args=(course_id,),
-            )
+        course_path = urlquote('/courses/{course_id}/course'.format(course_id=course_id))
+        lms_root_url = get_configuration_value_for_site(enterprise_customer.site, 'LMS_ROOT_URL')
+        destination_url = '{site}/{login_or_register}?next={course_path}'.format(
+            site=lms_root_url,
+            login_or_register='{login_or_register}',  # We don't know the value at this time
+            course_path=course_path
+        )
         course_name = course_details.get('title')
 
         try:
@@ -525,18 +523,20 @@ class EnterpriseCustomerManageLearnersView(View):
         except (TypeError, ValueError):
             course_start = None
             logging.exception(
-                "None or empty value passed as course start date.\nCourse Details:\n{course_details}".format(
+                'None or empty value passed as course start date.\nCourse Details:\n{course_details}'.format(
                     course_details=course_details,
                 )
             )
 
         with mail.get_connection() as email_conn:
             for user in users:
+                login_or_register = 'register' if isinstance(user, PendingEnterpriseCustomerUser) else 'login'
+                destination_url = destination_url.format(login_or_register=login_or_register)
                 send_email_notification_message(
                     user=user,
                     enrolled_in={
                         'name': course_name,
-                        'url': course_url,
+                        'url': destination_url,
                         'type': 'course',
                         'start': course_start,
                     },
@@ -556,17 +556,27 @@ class EnterpriseCustomerManageLearnersView(View):
         """
         program_name = program_details.get('title')
         program_branding = program_details.get('type')
-        program_url = program_details.get('marketing_url')
+        program_uuid = program_details.get('uuid')
+
+        lms_root_url = get_configuration_value_for_site(enterprise_customer.site, 'LMS_ROOT_URL')
+        program_path = urlquote('/dashboard/programs/{program_uuid}/'.format(program_uuid=program_uuid))
+        destination_url = '{site}/{login_or_register}?next={program_path}'.format(
+            site=lms_root_url,
+            login_or_register='{login_or_register}',
+            program_path=program_path
+        )
         program_type = 'program'
         program_start = get_earliest_start_date_from_program(program_details)
 
         with mail.get_connection() as email_conn:
             for user in users:
+                login_or_register = 'register' if isinstance(user, PendingEnterpriseCustomerUser) else 'login'
+                destination_url = destination_url.format(login_or_register=login_or_register)
                 send_email_notification_message(
                     user=user,
                     enrolled_in={
                         'name': program_name,
-                        'url': program_url,
+                        'url': destination_url,
                         'type': program_type,
                         'start': program_start,
                         'branding': program_branding,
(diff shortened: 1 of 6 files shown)


tuchfarber:
According to the number of tests that failed, it's apparently NOT perfect. Can you not use openedx libraries in enterprise?

enterprise/admin/views.py
```diff
@@ -45,6 +45,8 @@
 )
 from enterprise.utils import get_reversed_url_by_site, send_email_notification_message
 
+from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
```
douglashall: Take a look at how we import edx-platform modules in other files:

https://github.com/edx/edx-enterprise/blob/master/enterprise/views.py#L52
https://github.com/edx/edx-enterprise/blob/master/enterprise/views.py#L66
https://github.com/edx/edx-enterprise/blob/master/enterprise/views.py#L424

This allows us to use edx-platform code in the LMS runtime, but also be able to mock calls to those modules in the edx-enterprise tests.

tuchfarber: Ahh, so the tests don't pull in the rest of edx-platform.. Didn't realize. Good to know, will update!

enterprise/admin/views.py
```diff
@@ -508,7 +513,12 @@ def notify_enrolled_learners(cls, enterprise_customer, request, course_id, users
                 ).format(course_id)
             )
             return
-        course_url = course_details.get('marketing_url')
+
+        marketing_site_enabled = configuration_helpers.get_value(
+            'ENABLE_MKTG_SITE',
+            settings.FEATURES.get('ENABLE_MKTG_SITE', False)
+        )
+        course_url = course_details.get('marketing_url') if marketing_site_enabled else None
```
douglashall: Yes, you were correct. The case when `course_url` is None was not covered before your change here which is why the tests were passing (we never hit the `reverse` call). The think the way forward here is actually to not send them to the `course_about` URL at all and instead construct the [course info URL](https://courses.edx.org/courses/course-v1:MITx+CTL.SC3x+2T2017/course/) without using the Django reverse function. For program enrollment, the link should take them to the [program detail dashboard in LMS](https://courses.edx.org/dashboard/programs/f5448140-88fc-451c-82cf-976504bdfa8d/).

enterprise/admin/views.py
```diff
@@ -508,16 +508,11 @@ def notify_enrolled_learners(cls, enterprise_customer, request, course_id, users
...
-            # then we should build a path to the course on the LMS directly.
-            course_url = get_reversed_url_by_site(
-                request,
-                enterprise_customer.site,
-                'about_course',
-                args=(course_id,),
-            )
+
+        course_url = "http://{site}/courses/{course_id}/course/".format(
+            site=enterprise_customer.site,
```
douglashall: Sorry, I missed this. You will want to construct this URL like this:
```
lms_root_url = enterprise_customer.site.siteconfiguration.get_value('LMS_ROOT_URL', settings.LMS_ROOT_URL)
next_url = urljoin(lms_root_url, '/courses/', course_id, '/course/')
login_url = '{login_url}?{query}'.format(
    login_url=urljoin(lms_root_url, '/login'),
    query=urlencode({'next': next_url})
)
```

enterprise/admin/views.py
```diff
@@ -532,11 +525,17 @@ def notify_enrolled_learners(cls, enterprise_customer, request, course_id, users
 
         with mail.get_connection() as email_conn:
             for user in users:
+                login_or_register = 'register' if isinstance(user, PendingEnterpriseCustomerUser) else 'login'
+                destination_url = logistration_url.format(
+                    site=enterprise_customer.site,
```
douglashall: This is looking good. I think you still want to use `enterprise_customer.site.siteconfiguration` to get the `LMS_ROOT_URL` instead of hard-coding `http` above.

douglashall: Also, you only need to format the `login_or_register` value inside the loop here, the rest could be formatted outside the loop.

tuchfarber:
@douglashall Look good now?

enterprise/admin/views.py
```diff
@@ -508,16 +509,13 @@ def notify_enrolled_learners(cls, enterprise_customer, request, course_id, users
...
-            # then we should build a path to the course on the LMS directly.
-            course_url = get_reversed_url_by_site(
-                request,
-                enterprise_customer.site,
-                'about_course',
-                args=(course_id,),
-            )
+        course_path = urlquote("/courses/{course_id}/info".format(course_id=course_id))
+        lms_root_url = get_configuration_value_for_site(enterprise_customer.site, 'LMS_ROOT_URL')
+        destination_url = "{site}/{login_or_register}?next={course_path}".format(
```
douglashall: Nit: Keep quoting consistent, single vs double. Most of this file  uses single quotes.

enterprise/admin/views.py
```diff
@@ -508,16 +509,13 @@ def notify_enrolled_learners(cls, enterprise_customer, request, course_id, users
...
-        if course_url is None:
-            # If we didn't get a useful path to the course on a marketing site from the catalog API,
-            # then we should build a path to the course on the LMS directly.
-            course_url = get_reversed_url_by_site(
-                request,
-                enterprise_customer.site,
-                'about_course',
-                args=(course_id,),
-            )
+        course_path = urlquote("/courses/{course_id}/info".format(course_id=course_id))
```
douglashall: This needs to be `/courses/{course_id}/course`. I think that changed recently.

enterprise/utils.py
```diff
@@ -594,3 +594,19 @@ def format_price(price, currency='$'):
     if int(price) == price:
         return '{}{}'.format(currency, int(price))
     return '{}{:0.2f}'.format(currency, price)
+
+
+def get_configuration_value_for_site(site, key):
```
douglashall: Add a `default=None` kwarg.

enterprise/utils.py
```diff
@@ -594,3 +594,19 @@ def format_price(price, currency='$'):
...
+    Useful for testing when no Site Configuration exists in edx-enterprise or if a site in LMS doesn't have
+    a configuration tied to it.
+
+    :param site: A Site model object
+    :param key: The name of the value to retrieve
+    :return: The value located at that key in the site configuration or settings file.
+    """
+    if hasattr(site, 'configuration'):
+        return site.configuration.get_value(key, getattr(settings, key))
+    return getattr(settings, key)
```
douglashall:
```
if hasattr(site, 'configuration'):
    return site.configuration.get_value(key, default)
return default
```

enterprise/admin/views.py
tuchfarber: From what I found Info and Course are two separate tabs. If you go to 'info' it redirects to course if it doesn't exist. Wouldn't we want to drop them into the course description tab if it exists?

tuchfarber:
Whoops. Rebase blew away the comments...

enterprise/admin/views.py
douglashall: Maybe check with @andy-armstrong on the Learner team about those courseware URLs. It looks like they are somewhat in flux.

andy-armstrong: Thanks @douglashall. The info page is no longer shown in production, but instead it shows the unified '/course' page. Note that there is a redirect from /info so your code will continue to work.

We have a helper function in the LMS but that won't help you here. I would recommend that you use ``reverse`` rather than hard-coding the URL path. Something like:

    reverse('info', kwargs={'course_id': course_id})

The new course page equivalent would be:

    reverse('openedx.course_experience.course_home', kwargs={'course_id': course_id})

tuchfarber:
@douglashall With URL changed am I good to merge?

enterprise/admin/views.py
tuchfarber: Thanks for the info Andy. That explains what I was seeing with the /info and /course behaviour.

UmanShahzad:
@douglashall @tuchfarber This was missing some coverage :\

```
enterprise/utils.py  96%   309-316, 620, 619->620
```

PR #189, opened 2017-09-14
Allow multi-course enrollment for enterprise users in admin

bill-filler:
This adds functionality to allow an existing enterprise customer to be
enrolled in multiple courses. The current behavior would incorectly
limit the user to only be enrolled once via the django-admin page.

WL-1221

**Dependencies:** None

**Merge deadline:** None

**Installation instructions:** Install branch

**Testing instructions:**

1. As admin, Open Enterprise Customer->Manage Learners page
2. As admin, enroll new User A into Course A, verify enrollment email sent and User A put into Pending Linked Learner
3. As admin, Enroll User A into Course B, verify enrollment email sent and User A still in Pending List
4. As User A, create an account on edX
5. As User A, verify that Course A and Course B show up in Courses Dashboard
6. As admin, verify that User A was moved from Pending to Linked Learner list
7. As admin, enroll User A into Course C, verify enrollment email sent
8. As User A, verify that Course A, B, C show up in Courses dashboard

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:** None

### DIFF
--- a/enterprise/admin/views.py
+++ b/enterprise/admin/views.py
@@ -205,7 +205,7 @@ class EnterpriseCustomerManageLearnersView(View):
         form_field_value = manage_learners_form.cleaned_data[ManageLearnersForm.Fields.EMAIL_OR_USERNAME]
         email = email_or_username__to__email(form_field_value)
         try:
-            validate_email_to_link(email, form_field_value, ValidationMessages.INVALID_EMAIL_OR_USERNAME)
+            validate_email_to_link(email, form_field_value, ValidationMessages.INVALID_EMAIL_OR_USERNAME, True)
         except ValidationError as exc:
             manage_learners_form.add_error(ManageLearnersForm.Fields.EMAIL_OR_USERNAME, exc)
         else:
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -302,9 +302,10 @@ class EnterpriseCustomerUserManager(models.Manager):
         """
         try:
             existing_user = User.objects.get(email=user_email)
-            self.create(enterprise_customer=enterprise_customer, user_id=existing_user.id)
+            self.get_or_create(enterprise_customer=enterprise_customer, user_id=existing_user.id)
         except User.DoesNotExist:
-            PendingEnterpriseCustomerUser.objects.create(enterprise_customer=enterprise_customer, user_email=user_email)
+            PendingEnterpriseCustomerUser.objects.get_or_create(enterprise_customer=enterprise_customer,
+                                                                user_email=user_email)
 
     def unlink_user(self, enterprise_customer, user_email):
         """
--- a/AUTHORS
+++ b/AUTHORS
@@ -11,3 +11,4 @@ Douglas Cerna <douglas@opencraft.com>
 Tim Krones <tim@opencraft.com>
 Matt Drayer <mattdrayer@edx.org>
 Lucas Tadeu Teixeira <lucas@opencraft.com>
+Bill Filler <bfiller@edx.org>
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,10 @@ Change Log
 
 Unreleased
 ----------
+[0.46.0] - 2017-09-15
+---------------------
+
+* Allow multi-course enrollment for enterprise users in admin. 
 
 [0.45.0] - 2017-09-14
 ---------------------
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.45.0"
+__version__ = "0.46.0"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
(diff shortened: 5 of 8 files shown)


enterprise/admin/views.py
```diff
@@ -205,11 +205,13 @@ def _handle_singular(cls, enterprise_customer, manage_learners_form):
         form_field_value = manage_learners_form.cleaned_data[ManageLearnersForm.Fields.EMAIL_OR_USERNAME]
         email = email_or_username__to__email(form_field_value)
         try:
-            validate_email_to_link(email, form_field_value, ValidationMessages.INVALID_EMAIL_OR_USERNAME)
+            validate_email_to_link(email, form_field_value, ValidationMessages.INVALID_EMAIL_OR_USERNAME, True)
```
douglashall: @haikuginger @e-kolpakov Do either of you know what the purpose of the `ignore_existing` kwarg is/was on the [validate_email_to_link](https://github.com/edx/edx-enterprise/blame/master/enterprise/admin/utils.py#L131) function? I can't think of a reason to ever need to raise an exception if there is an existing `EnterpriseCustomerUser`.

enterprise/admin/views.py
```diff
@@ -205,11 +205,13 @@ def _handle_singular(cls, enterprise_customer, manage_learners_form):
...
         except ValidationError as exc:
             manage_learners_form.add_error(ManageLearnersForm.Fields.EMAIL_OR_USERNAME, exc)
         else:
-            EnterpriseCustomerUser.objects.link_user(enterprise_customer, email)
+            if not (EnterpriseCustomerUser.objects.is_user_linked(enterprise_customer, email) or
+                    EnterpriseCustomerUser.objects.is_user_pending(enterprise_customer, email)):
+                EnterpriseCustomerUser.objects.link_user(enterprise_customer, email)
```
douglashall: I think it would be better to modify the [link_user](https://github.com/edx/edx-enterprise/blob/master/enterprise/models.py#L305) function to call [get_or_create](https://docs.djangoproject.com/en/dev/ref/models/querysets/#get-or-create) instead of create.

enterprise/admin/views.py
```diff
@@ -261,7 +263,9 @@ def _handle_bulk_upload(cls, enterprise_customer, manage_learners_form, request,
 
         # There were no errors. Now do the actual linking:
         for email in emails:
-            EnterpriseCustomerUser.objects.link_user(enterprise_customer, email)
+            if not (EnterpriseCustomerUser.objects.is_user_linked(enterprise_customer, email) or
+                    EnterpriseCustomerUser.objects.is_user_pending(enterprise_customer, email)):
+                EnterpriseCustomerUser.objects.link_user(enterprise_customer, email)
```
douglashall: Same here. Modify `link_user` to call `get_or_create`.

enterprise/models.py
```diff
@@ -332,6 +333,36 @@ def unlink_user(self, enterprise_customer, user_email):
...
+        Check if user is already a pending enterprise customer.
+        """
+        try:
+            pending_link = PendingEnterpriseCustomerUser.objects.get(
+                enterprise_customer=enterprise_customer, user_email=user_email
+            )
+            return bool(pending_link)
+
+        except PendingEnterpriseCustomerUser.DoesNotExist:
+            return False
```
douglashall: We won't need these if we modify `link_user` as suggested above.

enterprise/models.py
```diff
@@ -323,7 +323,8 @@ def unlink_user(self, enterprise_customer, user_email):
         try:
             existing_user = User.objects.get(email=user_email)
             # not capturing DoesNotExist intentionally to signal to view that link does not exist
-            link_record = self.get(enterprise_customer=enterprise_customer, user_id=existing_user.id)
+            link_record = self.get(enterprise_customer=enterprise_customer,
+                                   user_id=existing_user.id)
```
douglashall: Nit: No need to wrap here.

enterprise/admin/views.py
bill-filler: will give that a try, didn't know about that method. agreed would be cleaner

douglashall:
@bill-filler Looks good to go once the checks pass.

Before merging we generally, increment the [package version](https://github.com/edx/edx-enterprise/blob/master/enterprise/__init__.py#L7) and add a [changelog entry](https://github.com/edx/edx-enterprise/blob/master/CHANGELOG.rst). After merge, we create a git tag, push the tag to github, wait for the updated package version to publish to [pypi](https://pypi.python.org/pypi/edx-enterprise), and then create an [edx-platform PR](https://github.com/edx/edx-platform/pull/16024) to upgrade the edx-enterprise package version.

enterprise/admin/views.py
e-kolpakov: @douglashall If I recall right at that point the decision was that customer can only be linked to single enterprise customer, so this is a mechanism to indicate that a student is already linked to enterprise customer.

In the original PR, it was set to ignore bulk uploads (for the purposes of allowing same file loaded multiple times), but report it during adding single student. At that point it was considered as malformed request and reported as validation error.

PR #191, opened 2017-09-15
ENT-592 Add exception handling for transmit course data task

zubair-arbi:
@asadiqbal08 @saleem-latif @douglashall @mattdrayer 

**Description:** Add exception handling for transmit course data task so that course data transmission for multiple integrated channels continues even if any of these integrated channels fails to transmit its catalog data.

**JIRA:** [ENT-592](https://openedx.atlassian.net/browse/ENT-592)

**Dependencies:** N/A

**Merge deadline:** 22 September, 2017

**Installation instructions:** Install `edx-enterprise` from this branch `zub/ENT-592-ocn-catalog-push-command-failure-tolerance` in `edx-platform`.

**Testing instructions:**

1. Create enterprise with a course catalog
2. Create two SAP integrated channels: First integrated channel with some bad configuration so that the management command `transmit_courseware_data` fails to transmit its data and a second SAP integrated channel with proper configuration: https://openedx.atlassian.net/wiki/display/SOL/SAP+Testing+and+Configuration+notes
3. Run the management command transmit_courseware_data for the enterprise like:
```
./manage.py lms --settings=devstack transmit_courseware_data --catalog_user staff
```
4. Verify the transmission of first integrated channel fails with the log message `Transmission of course data failed for user "C-3PO" and for integrated channel with code "SAP" and id "1".` but the second integrated channel successfully transmits its courses data, ignoring the error in transmission of first integrated channel.

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
+++ b/integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
@@ -66,9 +66,10 @@ def send_data_task(username, channel_code, channel_pk):
     """
     Task to send course data to each linked integrated channel
 
-    Args:
-        user: The user whose access token to use for course discovery API access
-        enterprise_customer: The EnterpriseCustomer whose integrated channels we need to send.
+    Arguments:
+        channel_code (str): Capitalized identifier for the integrated channel
+        channel_pk (str): Primary key for identifying integrated channel
+
     """
     user = User.objects.get(username=username)
     channel = INTEGRATED_CHANNEL_CHOICES[channel_code].objects.get(pk=channel_pk)
@@ -78,4 +79,13 @@ def send_data_task(username, channel_code, channel_pk):
         channel,
     )
 
-    channel.transmit_course_data(user)
+    try:
+        channel.transmit_course_data(user)
+    except Exception:  # pylint: disable=broad-except
+        exception_message = 'Transmission of course metadata failed for user "{username}" and for integrated ' \
+                            'channel with code "{channel_code}" and id "{channel_pk}".'.format(
+                                username=username,
+                                channel_code=channel_code,
+                                channel_pk=channel_pk,
+                            )
+        LOGGER.exception(exception_message)
--- a/test_utils/fake_enterprise_api.py
+++ b/test_utils/fake_enterprise_api.py
@@ -177,6 +177,18 @@ class EnterpriseMockMixin(object):
                 content_type='application/json',
             )
 
+    def mock_ent_courses_api_with_error(self, enterprise_uuid):
+        """
+        DRY function to register enterprise courses API to return error response.
+        """
+        responses.add(
+            responses.GET,
+            url=self.build_enterprise_api_url('enterprise-customer-courses', enterprise_uuid),
+            json={},
+            status=500,
+            content_type='application/json',
+        )
+
     def mock_empty_response(self, resource, *args, **kwargs):
         """
         DRY function to register an empty response from some Enterprise API endpoint.
--- a/tests/test_management.py
+++ b/tests/test_management.py
@@ -89,6 +89,89 @@ class TestTransmitCoursewareDataManagementCommand(unittest.TestCase, EnterpriseM
     @responses.activate
     @override_switch('SAP_USE_ENTERPRISE_ENROLLMENT_PAGE', active=True)
     @mock.patch('enterprise.api_client.lms.JwtBuilder', mock.Mock())
+    @mock.patch('enterprise.models.configuration_helpers')
+    @mock.patch('integrated_channels.sap_success_factors.utils.reverse')
+    @mock.patch('integrated_channels.sap_success_factors.transmitters.SAPSuccessFactorsAPIClient')
+    def test_transmit_courseware_task_with_error(
+            self,
+            fake_sap_client,
+            track_selection_reverse_mock,
+            fake_config_helpers,
+    ):
+        """
+        Verify the data transmission task for integrated channels with error.
+
+        Test that the management command `transmit_courseware_data` transmits
+        courses metadata related to other integrated channels even if an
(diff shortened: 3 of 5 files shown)


mattdrayer:
It may be outside of the scope of this PR, but the name "transmit_courseware_data" is inaccurate -- it really should be "transmit_course_metadata" or "transmit_catalog_data", etc.

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
```diff
@@ -78,4 +78,14 @@ def send_data_task(username, channel_code, channel_pk):
         channel,
     )
 
-    channel.transmit_course_data(user)
+    # channel.transmit_course_data(user)
```
asadiqbal08: need to remove ?

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
```diff
@@ -78,4 +78,14 @@ def send_data_task(username, channel_code, channel_pk):
...
+    try:
+        channel.transmit_course_data(user)
+    except Exception:  # pylint: disable=broad-except
+        exception_message = 'Transmission of course data failed for user "{username}" and for integrated channel ' \
```
asadiqbal08: `course metadata ...`

tests/test_management.py
```diff
@@ -89,6 +89,90 @@ def test_working_user(self, mock_data_task):
...
+            self,
+            fake_sap_client,
+            track_selection_reverse_mock,
+            fake_config_helpers,
+    ):
+        """
+        Verify that course data transmission works for all channels.
+
+        Test that the management command `transmit_courseware_data` logs
+        exception for an integrated channel and continues for other remaining
```
asadiqbal08: The above line `Verify that course data transmission works for all channels.` seems me confusing.

test_utils/fake_enterprise_api.py
```diff
@@ -177,6 +177,18 @@ def mock_ent_courses_api_with_pagination(self, enterprise_uuid, course_run_ids):
                 content_type='application/json',
             )
 
+    def mock_ent_courses_api_with_error(self, enterprise_uuid):
```
asadiqbal08: nit: `ent ~ enterprise`

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
```diff
@@ -78,4 +78,14 @@ def send_data_task(username, channel_code, channel_pk):
         channel,
     )
 
-    channel.transmit_course_data(user)
+    # channel.transmit_course_data(user)
+    try:
+        channel.transmit_course_data(user)
+    except Exception:  # pylint: disable=broad-except
```
UmanShahzad: We should really narrow down what the common errors are here, no?

asadiqbal08: Actually, We did not know what exactly the exception occurring here.

UmanShahzad: Yep I assumed so, but it would be nice to know what it is.

brittneyexline: i think we should catch this general exception, but we need to add logging information about this exception below. then we can start adding specific cases as we find them.

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
```diff
@@ -78,4 +78,14 @@ def send_data_task(username, channel_code, channel_pk):
...
+        exception_message = 'Transmission of course data failed for user "{username}" and for integrated channel ' \
+                        'with code "{channel_code}" and id "{channel_pk}".'.format(
+                            username=username,
+                            channel_code=channel_code,
+                            channel_pk=channel_pk,
+                        )
+        LOGGER.exception(exception_message)
```
brittneyexline: we need to also log information about the exception that was caught.

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
```diff
@@ -78,4 +78,14 @@ def send_data_task(username, channel_code, channel_pk):
         channel,
     )
 
-    channel.transmit_course_data(user)
+    # channel.transmit_course_data(user)
+    try:
+        channel.transmit_course_data(user)
```
brittneyexline: So I was digging around in this management command as part of the data passback work and noticed that the reason the entire job fails is because we are creating celery tasks to do work by channel, not by enterprise customer. So that is why if one enterprise customer fails, everything fails, because they are all being done in the same thread. I think we need to refactor to kick off a celery task per enterprise customer, which was the original intent from what I remember. Is there a reason doing these celery tasks by enterprise customer would be problematic @haikuginger and @douglashall?

integrated_channels/integrated_channel/management/commands/transmit_courseware_data.py
zubair-arbi: LOGGER.exception will include traceback for exception.

test_utils/fake_enterprise_api.py
zubair-arbi: @asadiqbal08 the quality will fail due to long variable name:
```
quality runtests: commands[5] | pylint test_utils
************* Module test_utils.fake_enterprise_api
C:110, 4: Invalid method name "mock_enterprise_courses_api_with_pagination" (invalid-name)
```

zubair-arbi:
@asadiqbal08  @brittneyexline Please have another look.

asadiqbal08:
👍

PR #190, opened 2017-09-15
[ENT-428] Remove hard requirement on `configuration_helpers`.

UmanShahzad:
**Description:** This PR removes our hard requirement on `configuration_helpers` by abstracting all uses of it into a single utility function which can be easily modified in the future in case we decide to officially drop the need for `configuration_helpers`.

**JIRA:** [ENT-428](https://openedx.atlassian.net/browse/ENT-428)

**Merge deadline:** N/A

**Testing instructions:**

Just see that the abstraction is occurring properly and that tests are still passing.

To be safe, go to the course enrollment landing page with this PR's code installed, and make sure the platform name and other values we get through `configuration_helpers` are still gotten.

**Merge checklist:**

- [ ] Check that the versions of the requirements in the `platform-master.in` file match edx-platform.
- [ ] New requirements are in the right place (`base.in` if only used in enterprise; in the correct `platform-****.in` files if they're hosted in edx-platform)
- [ ] Regenerate requirements with `make upgrade && make requirements` (and make sure to fix any errors).
  **DO NOT** just add dependencies to `requirements/*.txt` files.
- [ ] All reviewers approved
- [x] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Documentation updated (not only docstrings)
- [ ] Commits are (reasonably) squashed
- [ ] Translations are updated
- [ ] PR author is listed in AUTHORS

**Post merge:**
- [ ] Create a tag
- [ ] Check new version is pushed to PyPi after tag-triggered build is finished.
- [ ] Delete working branch (if not needed anymore)
- [ ] edx-platform PR (be sure to include edx-platform requirements upgrades that were present in this PR)

PR #192, opened 2017-09-15
[ENT-644] Remove account-level & `EnterpriseCourseEnrollment` consent.

UmanShahzad:
**Description:** This removes any remaining traces of account-level consent and consent for courses tracked by the `EnterpriseCourseEnrollment` model.

**JIRA:** [ENT-644](https://openedx.atlassian.net/browse/ENT-644)

**Merge deadline:** N/A

**Testing instructions:**

Since account-level consent and course-level consent through `EnterpriseCourseEnrollment` were both deprecated in favor of `DataSharingConsent`, this should not have any effect on functionality. However, to be safe, it may be recommended to try out our data sharing consent flows to make sure everything still works as intended.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/tests/test_enterprise/api/test_views.py
+++ b/tests/test_enterprise/api/test_views.py
@@ -14,8 +14,7 @@ from django.conf import settings
 from django.test import override_settings
 from django.utils import timezone
 
-from enterprise.decorators import ignore_warning
-from enterprise.models import EnterpriseCustomer, EnterpriseCustomerIdentityProvider, UserDataSharingConsentAudit
+from enterprise.models import EnterpriseCustomer, EnterpriseCustomerIdentityProvider
 from six.moves.urllib.parse import urljoin  # pylint: disable=import-error
 from test_utils import FAKE_UUIDS, TEST_COURSE, TEST_USERNAME, APITest, factories, fake_catalog_api, fake_enterprise_api
 
@@ -42,7 +41,6 @@ ENTERPRISE_CUSTOMER_ENTITLEMENT_LIST_ENDPOINT = reverse('enterprise-customer-ent
 ENTERPRISE_CUSTOMER_LIST_ENDPOINT = reverse('enterprise-customer-list')
 ENTERPRISE_LEARNER_LIST_ENDPOINT = reverse('enterprise-learner-list')
 SITE_LIST_ENDPOINT = reverse('site-list')
-USER_DATA_SHARING_CONSENT_LIST_ENDPOINT = reverse('user-data-sharing-consent-list')
 
 
 @ddt.ddt
@@ -62,7 +60,7 @@ class TestEnterpriseAPIViews(APITest):
 
     @ddt.data(
         (
-            True, EnterpriseCustomer.AT_ENROLLMENT, UserDataSharingConsentAudit.ENABLED,
+            True, EnterpriseCustomer.AT_ENROLLMENT, True,
             [1, 2, 3], {"entitlements": [
                 {"entitlement_id": 1, "requires_consent": False},
                 {"entitlement_id": 2, "requires_consent": False},
@@ -70,7 +68,7 @@ class TestEnterpriseAPIViews(APITest):
             ]},
         ),
         (
-            True, EnterpriseCustomer.AT_ENROLLMENT, UserDataSharingConsentAudit.DISABLED,
+            True, EnterpriseCustomer.AT_ENROLLMENT, False,
             [1, 2, 3], {"entitlements": [
                 {"entitlement_id": 1, "requires_consent": True},
                 {"entitlement_id": 2, "requires_consent": True},
@@ -78,15 +76,7 @@ class TestEnterpriseAPIViews(APITest):
             ]},
         ),
         (
-            True, EnterpriseCustomer.AT_ENROLLMENT, UserDataSharingConsentAudit.NOT_SET,
-            [1, 2, 3], {"entitlements": [
-                {"entitlement_id": 1, "requires_consent": True},
-                {"entitlement_id": 2, "requires_consent": True},
-                {"entitlement_id": 3, "requires_consent": True},
-            ]},
-        ),
-        (
-            False, EnterpriseCustomer.AT_ENROLLMENT, UserDataSharingConsentAudit.ENABLED,
+            False, EnterpriseCustomer.AT_ENROLLMENT, True,
             [1, 2, 3], {"entitlements": [
                 {"entitlement_id": 1, "requires_consent": False},
                 {"entitlement_id": 2, "requires_consent": False},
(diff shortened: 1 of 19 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=192&repo=edx%2Fedx-enterprise"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

enterprise/decorators.py
```diff
@@ -17,58 +15,6 @@
...
-        """
-        @wraps(func)
-        def wrapper(*args, **kwargs):
-            """
-            Wrap the function.
-            """
-            warnings.simplefilter('ignore', warning)
-            return func(*args, **kwargs)
-        return wrapper
-    return decorator
```
UmanShahzad: I don't mind keeping them but they're unused, and deleting code is fun :]

tests/test_enterprise/api/test_views.py
```diff
@@ -220,33 +185,28 @@ def test_enterprise_learner_entitlements(
...
-        Arguments:
-            enable_data_sharing_consent (bool): True if enterprise customer enables data sharing consent,
-                False it does not.
-            enforce_data_sharing_consent (str): string for the location at which enterprise customer enforces
-                data sharing consent, possible values are 'at_enrollment' and 'externally_managed'.
-            learner_consent_state (str): string containing the state of learner consent on data sharing,
-                possible values are 'not_set', 'enabled' and 'disabled'.
-            entitlements (list): A list of integers pointing to voucher ids generated in E-Commerce CAT tool.
-            expected_json (dict): A dict with structure and values corresponding to
-                the expected json from API endpoint.
```
UmanShahzad: I didn't think it was useful to maintain an argument list doc for a test function. The names should speak for themselves.

tests/test_enterprise/api/test_views.py
```diff
@@ -220,33 +185,28 @@ def test_enterprise_learner_entitlements(
...
         """
-        user_id = 1
+        user_id = self.user.id + 1
```
UmanShahzad: Prevent conflicts => less flaky tests.

tests/test_enterprise/api/test_views.py
```diff
@@ -220,33 +185,28 @@ def test_enterprise_learner_entitlements(
...
+            enterprise_customer=enterprise_customer,
+            granted=learner_consent_state,
         )
         for entitlement in entitlements:
             factories.EnterpriseCustomerEntitlementFactory(
                 enterprise_customer=enterprise_customer,
                 entitlement_id=entitlement,
             )
-        url = reverse('enterprise-learner-entitlements', (user_id, ))
+        url = reverse('enterprise-learner-entitlements', (ecu.id, ))
```
UmanShahzad: The ID being used here is for the learner instance, not the user instance.

tests/test_migrations.py
```diff
@@ -77,97 +70,3 @@ def model_label(self):
...
-            granted=False,
-        )
-
-        self.enterprise_course_enrollment.consent_granted = True
-        self.enterprise_course_enrollment.save()
-
-        self.migrate_to_dest()
-
-        data_sharing_consent = DataSharingConsent.objects.all().first()
-        assert data_sharing_consent.granted
```
UmanShahzad: Although the migration this was testing obviously still exists, there should be no issue for new users of edx-enterprise, so there's no need to test that code anymore.

enterprise/decorators.py
```diff
@@ -17,58 +15,6 @@
 from six.moves.urllib.parse import parse_qs, urlencode, urlparse, urlunparse  # pylint: disable=import-error
 
 
-def deprecated(extra):
```
brittneyexline: this isn't specific to the old consent stuff although it brought this up as being useful - i would argue we should keep this because i'm sure we'll be deprecating other stuff. unless there is another way we should be doing that going forward?

UmanShahzad: Hmm, I think the decorators did help me remember what it was that needed to be deleted (that's why I put 'em in the first place -- and to bug us in the test logs into doing so). But I think now that it's unused, I'll have to add unit tests for coverage D: But no worries I'll do that :+1:

UmanShahzad:
@brittneyexline It is not deployed yet, no. I'll do that early Monday my time along with #190.

UmanShahzad:
This is now on the sandbox.

enterprise/decorators.py
UmanShahzad: @brittneyexline I've added these back and added unit tests.

UmanShahzad:
Note that Zeshan has tested this on the sandbox and found no regressions.

tests/test_enterprise/api/test_views.py
```diff
@@ -220,33 +185,28 @@ def test_enterprise_learner_entitlements(
...
-        factories.UserDataSharingConsentAuditFactory(
-            user__id=user_id,
-            user__enterprise_customer=enterprise_customer,
-            state=learner_consent_state,
+        user = factories.UserFactory(id=user_id)
+        ecu = factories.EnterpriseCustomerUserFactory(
+            user_id=user.id,
+            enterprise_customer=enterprise_customer,
+        )
+        factories.DataSharingConsentFactory(
```
haikuginger: Is this still doing something meaningful? Why aren't we passing a course ID (which should be required for this model), and why is the result working without a specific course ID? I might be missing something here.

UmanShahzad: The endpoint's function attempts to [get entitlements](https://github.com/edx/edx-enterprise/blob/master/enterprise/api/v1/views.py#L220) which then only [looks to see](https://github.com/edx/edx-enterprise/blob/master/enterprise/models.py#L417) if any DSC records exist between the learner and the customer, and if one does and DSC is enforced, the logic says consent isn't required.

I don't know if this is how DSC requirements for entitlements should actually be working, so please let me know and we'd have to fix that fast.

haikuginger: @douglashall, thoughts? I think we might have missed an edge case here; I think the appropriate fix would be to mark consent as required on the entitlements based _solely_ on the Enterprise setting, and then put logic in ecom to check for actual consent on the relevant courses. Or, you could even modify the existing consent views to redirect the user back to the success URL if consent straight-up isn't required.

douglashall: Is this something that would break our current coupon-voucher-based implementation of auto-discounts for enterprise learners? If so, we should fix it. Even though eventually, I think the entitlements endpoint can be removed entirely once we migrate over to using Enterprise Offers for the auto-enterprise-discount.

haikuginger: @douglashall I think it might. @UmanShahzad, see my thoughts on the implementation above.

UmanShahzad: @haikuginger @douglashall I will look into this, but I'm a little time-strapped with the programs landing page & the endpoint stuff. This issue was of course quite unexpected. I might have to ask @itsjeyd to help if I realize this may be a little too much for me alone right now. I'll let you guys know by tomorrow morning.

UmanShahzad: See https://openedx.atlassian.net/browse/ENT-644?focusedCommentId=286220&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-286220 for further discussion on this.

UmanShahzad:
Rebased on `master` from 609666adac4058b37dcf1bbebf8b2009e729f3bb.

UmanShahzad:
Ugh, coverage is failing due to `master`. I'll try and add something to this PR.

UmanShahzad:
Ok, fixed coverage from `master`.

tests/test_utilities.py
```diff
@@ -912,21 +904,38 @@ def test_ungettext_min_max(self, singular, plural, range_text, min_val, max_val,
...
+        """
+        site = SiteFactory()
+        site.configuration = mock.MagicMock(get_value=mock.MagicMock(return_value='value'))
+        assert utils.get_configuration_value_for_site(site, 'key', 'default') == 'value'
+
+    def test_get_configuration_value_for_site_without_configuration(self):
+        """
+        ``get_configuration_value_for_site`` returns the default because of no site configuration.
+        """
+        assert utils.get_configuration_value_for_site(SiteFactory(), 'key', 'default') == 'default'
```
douglashall: FYI @tuchfarber

UmanShahzad:
Rebased on `master` from d248130e29f359cf54d863fa0182dc2b92f023f5.

PR #195, opened 2017-09-20
asadiqbal08/ENT-591 - Added some log messages to get the actual reasons for 404

asadiqbal08:
**Description:** Added some log messages to trace possible reasons for 404.

[ENT-591](https://openedx.atlassian.net/browse/ENT-591)

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.46.6] - 2017-09-21
+---------------------
+
+* Added some log messages to trace possible 404 issue.
+
 [0.46.5] - 2017-09-21
 ---------------------
 
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -654,12 +654,18 @@ class CourseEnrollmentView(NonAtomicView):
             raise Http404
 
         if not course or not course_run:
+            LOGGER.warning('Failed to fetch course "{course}" or course run "{course_run}" details'.format(
+                course=course, course_run=course_run
+            ))
             raise Http404
 
         enterprise_customer = get_enterprise_customer_or_404(enterprise_uuid)
 
         modes = EnrollmentApiClient().get_course_modes(course_run_id)
         if not modes:
+            LOGGER.warning('Unable to get course modes for course run id {course_run_id}.'.format(
+                course_run_id=course_run_id
+            ))
             raise Http404
 
         course_modes = []
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.46.5"
+__version__ = "0.46.6"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name


enterprise/views.py
```diff
@@ -650,16 +650,23 @@ def get_base_details(self, enterprise_uuid, course_run_id):
         """
         try:
             course, course_run = CourseCatalogApiServiceClient().get_course_and_course_run(course_run_id)
-        except ImproperlyConfigured:
+        except ImproperlyConfigured as error:
```
UmanShahzad: So I don't think prod has an improper configuration for the course catalog API, so this part might not be necessary, otherwise it would happen for every single course.

CHANGELOG.rst
```diff
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 ----------
 
+[0.46.5] - 2017-09-20
+---------------------
+
+* Added some log messages.
+
+
```
UmanShahzad: Super nitty: most of us end up deleting this extra space :P

UmanShahzad:
CC @mattdrayer -- would this be good for you?

enterprise/views.py
asadiqbal08: So what do u think, only raising 404 in `ImproperlyConfigured` would be sufficient if this exception raised due to any reason in future?

UmanShahzad: This is only raised I believe if in the LMS Django admin we haven't set up a catalog integration. But if that's the case, it would *always* happen, on every single call.

UmanShahzad: If the call fails somewhere inside of `get_course_and_course_run`, we already log the error for that in `_load_data`.

asadiqbal08:
@brittneyexline

CHANGELOG.rst
```diff
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 ----------
 
+[0.46.5] - 2017-09-20
+---------------------
+
+* Added some log messages.
```
brittneyexline: I would be a little more specific about what these log messages are for.

enterprise/views.py
UmanShahzad: tl;dr this log is actually unnecessary and the traceback would give more than enough information (it would explicitly say "Catalog Integration does not exist" or something).

enterprise/views.py
```diff
@@ -654,12 +654,18 @@ def get_base_details(self, enterprise_uuid, course_run_id):
             raise Http404
```
asadiqbal08: @UmanShahzad I removed the log from here but it still seems me necessary to add log here too. Actual exception is `ImproperlyConfigured` and we are raising 404 inside it.

UmanShahzad: @asadiqbal08 The exception is raised at https://github.com/edx/edx-enterprise/blob/master/enterprise/api_client/discovery.py#L424 and the line below it. If you have an improperly configured catalog integration, this exception will constantly happen, until you configure one. Prod obviously has this configured so there's no need to log it. Moreover, the resulting stack-trace from such an error is quite unambiguous.

PR #196, opened 2017-09-21
[ENT-644] Remove migrations of old consent for a safe deployment. (Step 1)

UmanShahzad:
**Description:** This removes the migration involved in removal of the old account-level consent feature, so that we can deploy the code safely.

**JIRA:** [ENT-644](https://openedx.atlassian.net/browse/ENT-644)

**Merge deadline:** asap

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=196"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

UmanShahzad:
@douglashall Not sure why the tests are failing. Is this the proper way of going about this?

enterprise/migrations/0026_remove_old_consent.py
```diff
@@ -1,48 +0,0 @@
...
-            name='user',
-        ),
-        migrations.RemoveField(
-            model_name='enterprisecourseenrollment',
-            name='consent_granted',
-        ),
-        migrations.RemoveField(
-            model_name='enterprisecustomer',
-            name='require_account_level_consent',
-        ),
```
douglashall: I think before we entirely remove this field from code, we must first mark it as nullable and deploy the migration associated with that change. We can then remove the field entirely from code and deploy, then deploy the migration to remove it from the database.

UmanShahzad: That sounds pretty complex, and I'm starting to wonder if this is necessary for work outside of the platform (I wouldn't imagine we could *ever* do this for external dependencies that made such a migration, for example, but things work out fine). The main reason I'm doubtful is that, in order for this migration to come up *in the first place*, we have to have installed the latest edx-enterprise version, which comes with the latest code. I could still be wrong so maybe we should double-check with ops?

douglashall: It is complex, but it is necessary. I imagine that if we have done something similar in the past that the feature which touched the models with removed fields was broken for the time between when we run the migrations and when the servers with the new code start servicing users.

UmanShahzad: Alright, so I'll add back the fields in this PR and make them nullable, and generate the migration for that and we can deploy it. In #197 I'll remove these fields only but not make any new migration. In a new PR we'll finally make the migration. Jeez :P

UmanShahzad:
@douglashall Alrighty, can you double-check that this ok as "step 1" in this process?

UmanShahzad:
@haikuginger Can I get your quick +1 on this?

haikuginger:
@douglashall, do we need to be similarly squeamish about the consent field that exists on the EnterpriseCourseEnrollment?

douglashall:
@haikuginger The `EnterpriseCourseEnrollment.consent` field is currently nullable, so we should be good there. We will need to deploy the removal from code of that field first, before deploying the migration that removes it from the database.

UmanShahzad:
@douglashall @haikuginger Just a note on the priority of this: this safe deployment is blocking the edx-enterprise version bump pipeline we've got going on over at edx-platform. We can't bring any more changes here onto stage and prod until we safely deploy this because 0.46.5 contains the dangerous migration.

UmanShahzad:
@haikuginger Ah yeah, sorry if that was confusing, we've been going back and forth with this but it's clear now what the path is. Thanks a lot!

PR #203, opened 2017-09-25
[ENT-616] Fix course modal UI issues (...with Bootstrap & SASS/SCSS)

UmanShahzad:
**Description:** This PR primarily seeks to fix the course modal (and the DSC confirmation modal) that we have in our enterprise learner flow. In the process, it does the following:

1. Uses [edx-bootstrap](https://www.github.com/edx/edx-bootstrap) for our course and DSC confirmation modals; this prepares us to use the latest, most modern technology available and stay up-to-date with the direction edX is taking in front-end technologies, and also makes implementation and maintenance of our modals easier and much more guaranteed to be mobile responsive.
1. Rewrites all of our CSS in SASS/SCSS, helping us lessen code and consolidating our styles further. This is required if we want to take advantage of the features in `edx-bootstrap` (as it's fully in SASS/SCSS itself because Bootstrap 4 is, too).
1. Sets up a Webpack SASS/SCSS bundling pipeline, required for easily compiling our new SASS/SCSS into CSS for use; this also prepares us for when we become a service and need to bundle our static content optimally on provision.
1. Updates any necessary JS to account for the above changes.

**JIRA:** [ENT-616](https://openedx.atlassian.net/browse/ENT-616)

**Merge deadline:** Before end of sprint.

**Installation instructions:** If you want to develop on this, you **musn't** forget to `npm install` initially, and do `make static` when changing static content. Right now this matters, but in the future when we become a service, static asset generation should happen automatically when running the service in a container. But since we're currently being installed through `pip`, we have to do it manually.

**Testing instructions:**

Since this rewrites all of our CSS in SASS/SCSS, it would be wise to go through our landing pages, the DSC page, and the manage-learners admin portal to make sure everything looks and feels the same.

**NOTE**: Bootstrap was *only* used for the modal, so this does not necessarily make our overall pages any more mobile responsive; it just really applies to the modals.

For the **course enrollment landing page (1)**, **program enrollment landing page (2)**, **DSC confirmation page (3)**, and the **manage learners admin portal (4)**, ensure the following where they apply:
- Go to the relevant page.
- Make sure that styling looks the same as it did before this PR.
- Ensure that the page is as mobile responsive as it used to be.
- If a modal exists, open it, and see that it covers any issues enlisted on the ticket, and especially that it is mobile responsive.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=203"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

enterprise/templates/enterprise/admin/manage_learners.html
```diff
@@ -2,8 +2,9 @@
 {% load i18n static admin_urls %}
 
 {% block extrastyle %}
-<link rel="stylesheet" type="text/css" href="{% static 'enterprise/admin/manage_learners.css' %}"/>
-<link rel="stylesheet" type="text/css" href="{% static 'admin/css/changelists.css' %}"/>
+  <link rel="stylesheet" type="text/css" href="{% static 'admin/css/changelists.css' %}"/>
+  {# TODO: When we move to be a service, use the webpack loader's `render_bundle`. #}
+  <link rel="stylesheet" href="{% static 'enterprise/bundles/main-admin.style.css' %}"/>
```
UmanShahzad: Note to self: add `type="text/css"` and move it above the `admin/css/changelists.css` import.

enterprise/templates/enterprise/base.html
```diff
@@ -36,5 +41,7 @@
     {% endblock %}
 
     {% block contents %}{% endblock %}
+
+  <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-alpha.6/js/bootstrap.min.js"></script>
```
UmanShahzad: Note to self: I just found that we have Bootstrap v4.0.0-beta importable from the `edx-platform` given the static-based folder path `common/js/vendor/bootstrap.js`.

UmanShahzad:
Please don't be put off by the large diff. Most of the additions are from `package-lock.json` which is just generated by the more later versions of `npm` on each install.

UmanShahzad:
We also forgot to update translations for some of the previous PRs, and it will be necessary :\ So let's make sure we do that here or in whatever upcoming PR we merge.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -22,6 +22,7 @@ instructions.
 - [ ] New requirements are in the right place (`base.in` if only used in enterprise; in the correct `platform-****.in` files if they're hosted in edx-platform)
 - [ ] Regenerate requirements with `make upgrade && make requirements` (and make sure to fix any errors).
   **DO NOT** just add dependencies to `requirements/*.txt` files.
+- [ ] Called `make static` for webpack bundling if any static content was updated.
```
haikuginger: Hah! I was gonna ask for this if you hadn't done it!

LICENSE
```diff
@@ -659,3 +659,13 @@ specific requirements.
 if any, to sign a "copyright disclaimer" for the program, if necessary.
 For more information on this, and how to apply and follow the GNU AGPL, see
 <http://www.gnu.org/licenses/>.
+
+EdX Inc. wishes to state, in clarification of the above license terms, that
```
haikuginger: Why are we messing with the license files? And where did this additional text specifically come from?

Makefile
```diff
@@ -67,39 +88,35 @@ upgrade: ## update the requirements/*.txt files with the latest packages satisfy
...
 test: clean ## run tests in the current virtualenv
 	pip install -qr requirements/test.txt --exists-action w
 	py.test
 
 diff_cover: test
 	diff-cover coverage.xml
 
-test-all: clean jshint quality jasmine
-	## run tests on every supported Python/Django combination
+test-all: clean jshint ## run tests on every supported Python/Django combination
```
haikuginger: Should we `make static` here? Otherwise, the Jasmine tests will be using an old static build.

Makefile
```diff
@@ -67,39 +88,35 @@ upgrade: ## update the requirements/*.txt files with the latest packages satisfy
...
+validate: test ## run tests and quality checks
+	tox -e quality
 
-isort: ## Call isort on packages/files that are checked in quality tests
-	isort --recursive tests test_utils enterprise consent manage.py setup.py test_settings.py
+isort: ## call isort on packages/files that are checked in quality tests
+	isort --recursive tests test_utils enterprise consent manage.py setup.py
+
+.PHONY: clean compile_translations coverage docs dummy_translations extract_translations \
+	fake_translations help pull_translations push_translations requirements test test-all upgrade validate isort
```
haikuginger: The static make commands should be in here too, correct?

enterprise/static/enterprise/js/grant_data_sharing_permissions.js
```diff
@@ -47,49 +26,29 @@ function loadConsentPage() {
...
     });
-    $("#consent-confirmation-modal").click(function (event) {
-        hideConsentConfirmationModal();
-        $("#failure-link").focus();
-    });
     $("#data-consent-checkbox").change(function (event) {
         $("#consent-button").attr("disabled", !(this.checked));
     });
-    $(document).keydown(function (event) {
-        if ((event.keyCode === 27) && ($("#consent-confirmation-modal").is(":visible"))) {
```
haikuginger: Bootstrap handles this for us now?

enterprise/static/enterprise/js/grant_data_sharing_permissions.js
```diff
@@ -47,49 +26,29 @@ function loadConsentPage() {
...
-        if (event.keyCode == 9 && !event.shiftKey){
+    $reviewPolicyLink.keydown(function (event) {
+        if (event.keyCode === 9 && !event.shiftKey){
             // Catch the tab keydown event when leaving the last control in the modal,
             // and move the focus to the first control in the modal.
             event.preventDefault();
             $("#modal-close-button").focus();
         }
     });
-    $("#modal-close-button").keydown(function (event) {
```
haikuginger: I see we're still manually handling the forward wraparound to the top of the modal; why are we removing the backward wraparound to the end of the modal?

setup.cfg
```diff
@@ -1,12 +0,0 @@
-[isort]
```
haikuginger: Why are we removing `setup.cfg`?

spec/javascripts/data_sharing_consent_spec.js
```diff
@@ -27,51 +25,32 @@ describe("Data sharing consent page", function () {
 
     describe("modal confirmation div", function () {
         it("is not shown when the page is loaded", function () {
-            expect($("#consent-confirmation-modal")).toBeHidden();
-        });
-        it("is shown when the failure link is clicked", function () {
```
haikuginger: I assume we're removing these specs because they're handled by Bootstrap; shouldn't we still want to ensure the behavior, though?

.gitignore
```diff
@@ -77,3 +77,10 @@ default.db
 
 # Geckodriver for Jasmine
 geckodriver/
+
+# Static assets
+enterprise/assets/
```
haikuginger: Isn't this where Webpack is going to put the built assets? We want those in version control, right?

.github/PULL_REQUEST_TEMPLATE.md
UmanShahzad: Haha yep, I've been spamming it during development.

LICENSE
UmanShahzad: For some odd reason we had two LICENSE files, one with `.txt` and one without. This additional text is from my lawyer.

Just kidding, it's copied from the other services (ecommerce/discovery) LICENSE files. I think I did this because I was doing drive-by changes/fixes to files in our root. Sorry if it's a little uncalled for.

Makefile
UmanShahzad: Yes, good idea :+1:

Makefile
UmanShahzad: Yep, good catch :+1:

enterprise/static/enterprise/js/grant_data_sharing_permissions.js
UmanShahzad: Yep, Bootstrap 4 handles almost everything JS side for modals as long as you provide the appropriate attributes, which I've made sure to do. You can do extra fancy stuff like catching events but we don't need that right now.

enterprise/static/enterprise/js/grant_data_sharing_permissions.js
UmanShahzad: You're right. Not sure why I erased this. Thanks for pointing this out!

setup.cfg
UmanShahzad: Not a great reason, but here's what I had in mind: when we move to be a service, we won't need `setup.py` (correct me if I'm wrong). I know `setup.cfg` is not necessarily linked to it, but I felt it'd be a good idea to have all this stuff in 1 central configuration file, namely `tox.ini` since that's where we already store stuff, rather than spread out all over the place almost arbitrarily. Again this change isn't necessarily called for in this PR, but I made it a "drive-by" change while changing other stuff in the root folder.

spec/javascripts/data_sharing_consent_spec.js
UmanShahzad: When we use JS features handled by Bootstrap like Bootstrap's `fade` animation on modals, it's hard to test these things exactly, because they don't always update the values we're looking for immediately. There probably is some way to test these things safely, but I'm not too knowledgeable about it right now, and I also don't think we should bother verifying that Bootstrap JS works -- I'd leave it to Bootstrap to verify that (same idea applies for 3rd party Python libs; we wouldn't bother verifying they behave correctly, unless there's some really obvious upstream issue we need to work around).

.gitignore
UmanShahzad: Webpack actually only generates the `enterprise/static/enterprise/bundles` folder. This `assets` folder is generated by `python manage.py collectstatic --noinput` inside of the `static` make target. In the future we'll want to be keeping all of these generated files out of version control as they're only necessary while provisioning. Since we're installed through `pip` right now because we aren't a service, I'm only keeping the `bundles` folder around as a place to grab CSS inside of our templates. Eventually we'd ignore even this and pick up assets through the Webpack loader.

Makefile
UmanShahzad: Actually I realized that at least for now, this doesn't matter, but will in the future. Right now we're just using Webpack for SASS/SCSS -> CSS.

chris-mike:
@UmanShahzad Some comments re: modal.  First, I'm only viewing through the program LP. Let me know if I ought to review the modal through the course LP or elsewhere, at this point.

1. There's some inconsistency with regards to images of course staff. The real photo is much smaller than the rest. Can we make all photos the same size as the first (110px)
2. Also, for testing purposes, can you display one course staff bio w/o an image. I would expect the text to left align with the images in that case.
3. If I tab through, focus hits the "close", the browser URL field, and something behind the modal, on the page, but I can't see what. We ought to keep focus in the modal.

Titles look good, sections look good, spacing looks good.  Thanks!

UmanShahzad:
@chris-mike Thanks for the review :)

1. Yes; just to confirm, you mean 110px *max* for width & height, or force them to be 110px by 110px?
1. Ah, currently they are not left-aligning when the picture is removed because the image tag stays and shows an alternative text instead. I can fix that.
1. Sure, a little JavaScript can fix that.

UmanShahzad:
@chris-mike Addressed the review. Will put this up on the sandbox by tomorrow before your start of day.

UmanShahzad:
@chris-mike This is now on the sandbox.

chris-mike:
For course staff images, we request in Drupal: 110 x 110 and the size must be less than 256KB. 

Also, I noticed that the modal has no height. The expectation is that it would take up about 75% of the screen.

LICENSE
haikuginger: No problem. Pinging @douglashall to make sure this is okay.

haikuginger:
Code looks good, and nothing significant (outside the ticket requirements) appears changed, with two exceptions:
* The "Are you aware..." text in the Consent declination popup - it now shows as plain text, whereas previously it showed as a bolded, larger header.
* The course popup dialog uses a different "close" button icon than the consent declination popup; we should pick one.

chris-mike:
@UmanShahzad Let's leave the scroll / modal height as is. I like the way it functions currently and I see no UX issues.

UmanShahzad:
@chris-mike Very happy to hear that :) I also think this Bootstrap default is really nice.

@haikuginger I'll force 110x110 dimensions and fix that "Are you aware..." text. I also don't have a preference for the close button icon. Any suggestion @chris-mike? (use the current course modal one vs DSC one, for both?)

chris-mike:
I don't recall what we're using as a close-icon in the DSC modal. I assume it's fa-close (alias) [&#xf00d;].  If so, I prefer no circle b/c this icon is more prevalent across edx.

http://fontawesome.io/icon/times/

UmanShahzad:
@chris-mike Alrighty, let's use that (fa-times).

UmanShahzad:
@haikuginger @chris-mike Addressed review:

1. Close icons for modals are all now `fa-times`. I realized it does actually look better.
1. Staff profile images are forced to be 110x110 dimensions.
1. The "Are you aware..." text in the DSC modal is now the same style as it used to be (bigger and bolder).

This change is now on the sandbox.

ianderson96:
@UmanShahzad The modal updates look great! 

Just one thing. I may not be viewing the correct version of the sandbox, but staff profile images appear stretched to me (not 110x110). Example is attached.

screen shot 2017-09-29 at 11 21 20 am

UmanShahzad:
@ianderson96 Oh jeez, I didn't realize that the padding value was included in those pixels. I'll just change this to be a margin and it'll fix, sorry!

ianderson96:
@UmanShahzad Thanks! Everything else looks awesome, great work!

UmanShahzad:
@brittneyexline Thanks :)

@haikuginger Your +1 and we can go for it.

UmanShahzad:
Rebased from `master` on bbbe8a67314d03b1eeb30af51a43575a2223fb4c.

UmanShahzad:
@douglashall @brittneyexline Can either of you guys take a look at https://github.com/edx/edx-enterprise/pull/203#discussion_r141030371 just to make sure that's OK? Sorry for messing with the license files :\

LICENSE
brittneyexline: I talked with @douglashall and we don't see any issue since this is consistent with our other services.

UmanShahzad: Thanks!

UmanShahzad:
Squashed from f73c432744b449eb93cbc4e9c13de6b71fb2471a

PR #205, opened 2017-09-27
ENT-649 Only attempt to transmit SAP SuccessFactors details for which a currently-linked SSO user exists

haikuginger:
**Description:**

This pull request changes the user data transmission routine to only attempt to transmit and log user details in SAP SuccessFactors for users who are currently linked via SSO to the relevant provider; if there isn't a linked provider, then there's no real way for us to transmit data back anyway.

**JIRA:** [ENT-649](https://openedx.atlassian.net/browse/ENT-649)

**Dependencies:** None

**Merge deadline:** 9 October 2017

**Installation instructions:** N/A

**Testing instructions:**

1. Create ECU linked to an EnterpriseCustomer set up for SAPSF (real details not necessary)
2. Run data transmission management command
    - Note that no data transmision is attempted
3. Log into that user, and authenticate with the SSO provider linked to the EC
4. Run data transmission management command
    - Note that data transmission is attempted on that user (verify by checking the data transmission audit table)

To test improved login requirement robustness:

1. While fully logged-out, navigate to a program or course enterprise enrollment page.
2. Rather than logging in with SSO, as prompted, navigate directly to the platform login page and log in.
3. Again, navigate directly to a program or course enterprise enrollment page
4. Verify that your dropped back to an SSO-based login, rather than the current state of being allowed to enroll in the course/program.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=205"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

integrated_channels/sap_success_factors/models.py
```diff
@@ -117,14 +119,17 @@ def get_learner_data_record(self, enterprise_enrollment, completed_date=None, gr
...
 
-        return LearnerDataTransmissionAudit(
-            enterprise_course_enrollment_id=enterprise_enrollment.id,
-            sapsf_user_id=enterprise_enrollment.enterprise_customer_user.get_remote_id(),
-            course_id=enterprise_enrollment.course_id,
-            course_completed=course_completed,
-            completed_timestamp=completed_timestamp,
-            grade=grade,
-        )
+        sapsf_user_id = enterprise_enrollment.enterprise_customer_user.get_remote_id()
```
UmanShahzad: As per the test instructions, just authenticating through SAPSF wasn't enough, as the transmission command would skip the user each time since an enrollment did not exist. So just a nit to update the test instructions to reflect that we need to be enrolled in something first.

brittneyexline: I would still like there to be a log message that says we weren't able to find the remote id - I think it will be helpful for @mattdrayer when setting up and testing SAP with customers

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -132,12 +132,14 @@ def collect_learner_data(self):
...
                 completed_date, grade, is_passing = self._collect_grades_data(enterprise_enrollment, course_details)
 
-            yield self.get_learner_data_record(
+            record = self.get_learner_data_record(
                 enterprise_enrollment=enterprise_enrollment,
                 completed_date=completed_date,
                 grade=grade,
                 is_passing=is_passing,
             )
+            if record:
```
mattdrayer: Nit: Might be useful to add a comment here explaining to future developers why we need to differentiate between when a record is returned and when it isn't.

integrated_channels/sap_success_factors/models.py
mattdrayer: I agree with @brittneyexline -- can't hurt to write a log entry -- let's be sure we indicate the proper log level so the alerts don't become noisy.

haikuginger:
@brittneyexline, @mattdrayer, @UmanShahzad, I've addressed review comments. I've also added code to handle a minor edge case I turned up in the session-freshness-ensurer; I don't think that it would have had a major impact based on the methods to reproduce, but to test it, do the following:

1. While fully logged-out, navigate to a program or course enterprise enrollment page.
2. Rather than logging in with SSO, as prompted, navigate directly to the platform login page and log in.
3. Again, navigate directly to a program or course enterprise enrollment page
4. Verify that your dropped back to an SSO-based login, rather than the current state of being allowed to enroll in the course/program.

integrated_channels/integrated_channel/learner_data.py
```diff
@@ -132,12 +132,20 @@ def collect_learner_data(self):
...
+                # There are some cases where we won't receive a record from the above
+                # method; right now, that should only happen if we have an Enterprise-linked
+                # user for the integrated channel, and transmission of that user's
+                # data requires an upstream user identifier that we don't have (due to a
+                # failure of SSO or similar). In sucha a case, `get_learner_data_record`
```
UmanShahzad: Nit: "In sucha a..."

UmanShahzad:
@haikuginger Thanks, I put the latest updates on the sandbox. I've tested the session freshness case on there, and it works fine for the enterprises using TestShib, but not some of the others.

haikuginger:
@UmanShahzad, can you specify what you're seeing on the other providers? The logic should hold for all of them.

UmanShahzad:
@haikuginger For "Aviato", when originally going to the program landing page, I get a dialog to sign in using Aviato or some other way of signing in. I go to `/login/` as per test instructions, login, go back to the program landing page URL, but don't get shown an SSO-based login page, I just see the program landing page. This doesn't happen with Enterprises using TestShib; I get shown the TestShib login page instead of being directed to the landing page.

haikuginger:
@UmanShahzad, ah, that's fine. You can see by looking in the Django admin that Aviato doesn't set the "drop existing sessions" flag, so it wouldn't trigger this middleware.

UmanShahzad:
@haikuginger Ah ok then, we should be good :)

haikuginger:
Rebased and prepped for merge from 0698509.

PR #206, opened 2017-09-28
[ENT-653] Added UTM parameters to marketing/enrollment URLs

asadiqbal08:
@zubair-arbi @saleem-latif can you please take a look on that PR?
**Description:** Added UTM parameters to marketing/enrollment URLs returned by Enterprise API.

**JIRA:** [ENT-653](https://openedx.atlassian.net/browse/ENT-653)

**Merge checklist:**

- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.51.4] - 2017-10-11
+---------------------
+
+* Added UTM parameters to marketing, track selection, and course/program enrollment URLs returned by Enterprise API.
+
 [0.51.3] - 2017-10-10
 ---------------------
 
--- a/enterprise/api/v1/mixins.py
+++ b/enterprise/api/v1/mixins.py
@@ -60,7 +60,8 @@ class EnterpriseCourseContextSerializerMixin(object):
         # Update marketing urls in course metadata to include enterprise related info (i.e. our global context).
         marketing_url = course.get('marketing_url')
         if marketing_url:
-            course.update({"marketing_url": utils.update_query_parameters(marketing_url, enterprise_context)})
+            query_parameters = dict(enterprise_context, **utils.get_enterprise_utm_context(enterprise_customer))
+            course.update({'marketing_url': utils.update_query_parameters(marketing_url, query_parameters)})
 
         # Finally, add context to the course as a whole.
         course.update(enterprise_context)
@@ -82,8 +83,9 @@ class EnterpriseCourseContextSerializerMixin(object):
         for course_run in course_runs:
             track_selection_url = utils.get_course_track_selection_url(
                 course_run=course_run,
-                query_parameters=enterprise_context,
+                query_parameters=dict(enterprise_context, **utils.get_enterprise_utm_context(enterprise_customer)),
             )
+
             enrollment_url = enterprise_customer.get_course_run_enrollment_url(course_run.get('key'))
 
             course_run.update({
@@ -94,7 +96,8 @@ class EnterpriseCourseContextSerializerMixin(object):
             # Update marketing urls in course metadata to include enterprise related info.
             marketing_url = course_run.get('marketing_url')
             if marketing_url:
-                course_run.update({"marketing_url": utils.update_query_parameters(marketing_url, enterprise_context)})
+                query_parameters = dict(enterprise_context, **utils.get_enterprise_utm_context(enterprise_customer))
+                course_run.update({'marketing_url': utils.update_query_parameters(marketing_url, query_parameters)})
 
             # Add updated course run to the list.
             updated_course_runs.append(course_run)
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -12,9 +12,8 @@ from django.contrib.auth.models import User
 from django.contrib.sites.models import Site
 from django.utils.translation import ugettext_lazy as _
 
-from enterprise import models
+from enterprise import models, utils
 from enterprise.api.v1.mixins import EnterpriseCourseContextSerializerMixin
-from enterprise.utils import update_query_parameters
 
 
 class ImmutableStateSerializer(serializers.Serializer):
@@ -207,9 +206,14 @@ class EnterpriseCustomerCatalogDetailSerializer(EnterpriseCustomerCatalogSeriali
         count = paginated_content['count']
         search_results = paginated_content['results']
 
-        # Add the Enterprise enrollment URL to each content item returned from the discovery service.
         for item in search_results:
             content_type = item['content_type']
+            marketing_url = item.get('marketing_url')
+            if marketing_url:
+                item['marketing_url'] = utils.update_query_parameters(
+                    marketing_url, utils.get_enterprise_utm_context(enterprise_customer)
+                )
+            # Add the Enterprise enrollment URL to each content item returned from the discovery service.
             if content_type == 'courserun':
                 item['enrollment_url'] = enterprise_customer.get_course_run_enrollment_url(item['key'])
             elif content_type == 'program':
@@ -221,9 +225,9 @@ class EnterpriseCustomerCatalogDetailSerializer(EnterpriseCustomerCatalogSeriali
         page = int(request.GET.get('page', '1'))
         request_uri = request.build_absolute_uri()
         if paginated_content['previous']:
-            previous_url = update_query_parameters(request_uri, {'page': page - 1})
+            previous_url = utils.update_query_parameters(request_uri, {'page': page - 1})
         if paginated_content['next']:
-            next_url = update_query_parameters(request_uri, {'page': page + 1})
+            next_url = utils.update_query_parameters(request_uri, {'page': page + 1})
 
         representation['count'] = count
         representation['previous'] = previous_url
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -19,6 +19,7 @@ from django.core.exceptions import ObjectDoesNotExist
 from django.core.urlresolvers import reverse
 from django.http import Http404
 from django.template.loader import render_to_string
+from django.utils.text import slugify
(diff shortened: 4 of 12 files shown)


enterprise/api/v1/mixins.py
```diff
@@ -84,7 +91,10 @@ def update_course_runs(self, course_runs, enterprise_customer, enterprise_contex
                 course_run=course_run,
                 query_parameters=enterprise_context,
             )
-            enrollment_url = enterprise_customer.get_course_run_enrollment_url(course_run.get('key'))
+            #update enrollment url parameters with enterprise utm context
```
zubair-arbi: nit:
```
# Update querystring of enterprise course enrollment url with utm
# context.
```

enterprise/api/v1/mixins.py
```diff
@@ -51,22 +51,29 @@ def update_course(self, course, enterprise_customer, enterprise_context):
...
             enterprise_context=enterprise_context,
+            enterprise_utm_context=enterprise_utm_context
         )
 
         # Update marketing urls in course metadata to include enterprise related info (i.e. our global context).
         marketing_url = course.get('marketing_url')
         if marketing_url:
-            course.update({"marketing_url": utils.update_query_parameters(marketing_url, enterprise_context)})
+            query_parameters = dict(enterprise_context, **enterprise_utm_context)
+            course.update({"marketing_url": utils.update_query_parameters(marketing_url, query_parameters )})
```
zubair-arbi: nits: use single quotes also remove space before closing the bracket

enterprise/api/v1/mixins.py
```diff
@@ -51,22 +51,29 @@ def update_course(self, course, enterprise_customer, enterprise_context):
         Returns:
             (dict): Updated course metadata
         """
+        enterprise_utm_context = {
```
saleem-latif: I think it would be better to add utm context in the `enterprise_context` dict.

```
enterprise_context.update({
    'utm_medium': 'enterprise',
    'utm_source': enterprise_customer.name
})
``` 
That way it will be available to all other methods.

enterprise/api/v1/mixins.py
```diff
@@ -51,22 +51,29 @@ def update_course(self, course, enterprise_customer, enterprise_context):
...
-            course.update({"marketing_url": utils.update_query_parameters(marketing_url, enterprise_context)})
+            query_parameters = dict(enterprise_context, **enterprise_utm_context)
+            course.update({'marketing_url': utils.update_query_parameters(marketing_url, query_parameters)})
 
         # Finally, add context to the course as a whole.
         course.update(enterprise_context)
         return course
 
-    def update_course_runs(self, course_runs, enterprise_customer, enterprise_context):
+    def update_course_runs(self, course_runs, enterprise_customer, enterprise_context, enterprise_utm_context):
```
saleem-latif: We do not need the new argument, we can add/remove items in the `enterprise_context` dict.

enterprise/api/v1/mixins.py
```diff
@@ -84,7 +91,10 @@ def update_course_runs(self, course_runs, enterprise_customer, enterprise_contex
                 course_run=course_run,
                 query_parameters=enterprise_context,
             )
-            enrollment_url = enterprise_customer.get_course_run_enrollment_url(course_run.get('key'))
+            # Update querystring of enterprise course enrollment url with utm context.
```
saleem-latif: Here we can use the same approach as used for updating `marketing_url `

```
enrollment_url = enterprise_customer.get_course_run_enrollment_url(course_run.get('key'))
course_run.update({'enrollment_url': utils.update_query_parameters(marketing_url, enterprise_context)})
```

saleem-latif: I do not think updating query parameters of `enrollment_url ` using `enterprise_context ` would cause unwanted results but if they do then I think you can use `enterprise_context` to create another dict that only contains the desired key value pairs in this method.

enterprise/api/v1/mixins.py
asadiqbal08: Actually, I was thinking the same but `enterprise_context` is in use by some other methods [e.g.](https://github.com/edx/edx-enterprise/blob/master/enterprise/api/v1/mixins.py#L85 )that I think, at that point do not require to have UTM information in query parameters.

enterprise/api/v1/mixins.py
asadiqbal08: see the above comment

enterprise/api/v1/mixins.py
brittneyexline: I agree with @saleem-latif - it makes sense to me that this is generally available because we want to add this tracking to any link we are providing to customers to which they will potentially send traffic.

enterprise/api/v1/mixins.py
```diff
@@ -84,7 +91,10 @@ def update_course_runs(self, course_runs, enterprise_customer, enterprise_contex
                 course_run=course_run,
```
brittneyexline: we should update the track selection url since we're still returning it in the api

enterprise/api/v1/mixins.py
asadiqbal08: @brittneyexline If we update the `enterprise_context` dict with the `UTM` information then we will get the `UTM` information in course as we are updating the course with `enterprise_context` values [HERE](https://github.com/edx/edx-enterprise/blob/master/enterprise/api/v1/mixins.py#L66)
So we will get the `UTM` information in [Catalog Response](https://github.com/edx/edx-enterprise/blob/master/tests/test_enterprise/api/test_views.py#L486)

asadiqbal08:
@brittneyexline Do u mean that we should update the program_enrollment_url?
[get_program_enrollment_url](https://github.com/edx/edx-enterprise/blob/master/enterprise/models.py#L211) in that PR? or anything else also?

enterprise/api/v1/mixins.py
asadiqbal08: Actually, It was not mentioned in the story.

brittneyexline: My understanding of the story was any link that we publish relating to marketing a course or enrolling in a course, and the track selection page falls under that category. We want any traffic coming from enterprise's that are using our api to be tracked, so that includes this url.

brittneyexline:
@asadiqbal08 - I would interpret this story more broadly that any link we publish in the api at any level for courses and programs should have tracking information about the enterprise customer. The goal of this story is to have consistent tracking of traffic coming from direct integration customers so if one url where users can be sent that is published does not have tracking info, then that wouldn't be useful to us.

enterprise/api/v1/mixins.py
brittneyexline: alright, i see what you are saying. i think that is fine as long as we are updating all of the urls as described in the other comments.

asadiqbal08:
@douglashall I have updated the [function](https://github.com/edx/edx-enterprise/pull/206/files#diff-4bcae564bc272f5c2e6753dc6021eb1cR215) with UTM parameters, So I do not think make an explicit call here. anything else am I missing?

douglashall:
@asadiqbal08 Ah right, sorry I missed that.

CHANGELOG.rst
```diff
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.50.2] - 2017-10-04
+---------------------
+
+* Added UTM parameters to marketing,track_selection and course/programs enrollment URLs returned by Enterprise API.
```
douglashall: `Added UTM parameters to marketing, track selection, and course/program enrollment URLs returned by Enterprise API.`

douglashall:
@asadiqbal08 On second thought, for EnterpriseCustomerCatalogs this is only going to add the UTM parameters onto the course/program enrollment URLs. The marketing and track selection URLs also need to be updated.

asadiqbal08:
@douglashall track_selection and marketing_urls have been updated. Let me know if anything else is required in that PR to update?

douglashall:
@asadiqbal08 I'm still not seeing the marketing URLs getting updated for EnterpriseCustomerCatalogs. What am I missing?

asadiqbal08:
@douglashall I updated the `marketing_url` [HERE](https://github.com/edx/edx-enterprise/pull/206/files#diff-9b920c17601896b079cd031907580cb8R214). Are you pointing to that one?

enterprise/utils.py
```diff
@@ -636,3 +636,13 @@ def get_program_type_description(program_type):
...
     return PROGRAM_TYPE_DESCRIPTION.get(program_type, '')
+
+
+def get_enterprise_utm_context(enterprise_customer):
+    """
+    Get the UTM context for the enterprise.
+    """
+    return {
+        'utm_medium': 'enterprise',
+        'utm_source': enterprise_customer.name
```
douglashall: We should call `slugify` here to format the EntepriseCustomer name as a query parameter value like @saleem-latif  is doing [here](https://github.com/edx/course-discovery/pull/1145/files#diff-651fda19c2550b81fbdc3c7de19a0dbbR168).

enterprise/api/v1/serializers.py
```diff
@@ -210,6 +209,11 @@ def to_representation(self, instance):
         # Add the Enterprise enrollment URL to each content item returned from the discovery service.
         for item in search_results:
             content_type = item['content_type']
+            marketing_url = item.get('marketing_url')
+            if marketing_url:
+                item['marketing_url'] = utils.update_query_parameters(
+                    marketing_url, utils.get_enterprise_utm_context(enterprise_customer)
+                )
             if content_type == 'courserun':
```
douglashall: Move the comment above the `for` block down here closer to where we are populating the enrollment URL.

asadiqbal08:
@douglashall Are we good to merge it?

PR #213, opened 2017-10-10
[ENT-643] Direct enrollment to Audit Track

UmanShahzad:
**Description:** This PR adds the ability to be directly enrolled into a course in audit mode through the course enrollment landing page, given some conditions:

1. The query parameter `audit` exists in the link.
1. The course is in the enterprise customer's catalog.

**JIRA:** [ENT-643](https://openedx.atlassian.net/browse/ENT-643)

**Merge deadline:** End of sprint

## Testing Instructions

1. Setup an Enterprise Customer with DSC enabled.
1. Go to the course enrollment landing page without any query parameters.
1. Enroll in the course in whatever mode you want, and confirm everything works as normal as before.
1. With an unenrolled user, go to the course enrollment landing page with the `audit` query parameter (it can take any value).
1. Confirm that you do not see the landing page, but see the DSC page immediately. Functionality from hereon should match the DSC workflows available in the `edx-platform`.
1. Setup the Enterprise Customer to disable DSC.
1. With an unerolled user, go to the landing page, but do not see the DSC page. Rather, you should be immediately directed to the LMS courseware page for that course.

#### To ensure no regressions:

1. Go to the program enrollment landing page and ensure our workflows are as normal. In particular, **nothing needs to be tested further** if you can confirm that the page renders correctly.

#### Segment

To test the fired track event to segment, we need to access the segment account linked to our business sandbox. This is located at https://segment.com/business-sandbox/sources/javascript/debugger -- ask the team if you need credentials.

1. Go to the course enrollment landing page with the `audit` query parameter attached. You can use something like https://business.sandbox.edx.org/enterprise/1b542b06-d353-45e6-9867-9579d9e12745/course/course-v1:edX+Test101+course/enroll/?audit=yes
1. See that once you get redirected to courseware (or DSC), an event is tracked at https://segment.com/business-sandbox/sources/javascript/debugger. Look for the event `edx.bi.user.enterprise.enrollment.course`. It should show something like this on the sidebar when you click it:

```python
analytics.track('394', 'edx.bi.user.enterprise.enrollment.course', {
    'category': 'enterprise',
    'enterprise_customer_name': 'Aviato',
    'enterprise_customer_uuid': '1b542b06-d353-45e6-9867-9579d9e12745',
    'label': 'course-v1:edX+Test101+course',
    'mode': 'audit'
})
```

---

**Merge checklist:**

- [ ] All reviewers approved
- [x] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=213&repo=edx%2Fedx-enterprise"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

UmanShahzad:
@haikuginger Ready for review, but this is not yet on the sandbox; @douglashall if you'll be making the new sprint branch, let me know when and I can put this onto it unless you do first.

enterprise/apps.py
```diff
@@ -37,6 +39,9 @@ def ready(self):
         post_save.connect(handle_user_post_save, sender=self.auth_user_model, dispatch_uid=USER_POST_SAVE_DISPATCH_UID)
         pre_migrate.connect(self._disconnect_user_post_save_for_migrations)
 
+        # Segment initialization.
+        analytics.write_key = getattr(settings, 'LMS_SEGMENT_KEY')
```
douglashall: @UmanShahzad Do we need to do this here since we are already doing it in [LMS startup](https://github.com/edx/edx-platform/blob/master/lms/startup.py#L63)? This is causing an issue with running edxapp on a sandbox because during CMS startup this code runs and CMS settings do not contain an `LMS_SEGMENT_KEY` (instead it is `CMS_SEGMENT_KEY`). If we do need to do this here we should at least fail gracefully if `LMS_SEGMENT_KEY` does not exist.

UmanShahzad: I think you can erase it actually, yeah.

UmanShahzad: I've pushed the deletion now on this branch, too.

UmanShahzad:
@haikuginger This is now on the sandbox, and I can at least personally say that my own manual testing shows that this works. See test instructions.

enterprise/api_client/lms.py
```diff
@@ -229,6 +229,20 @@ def get_course_enrollment(self, username, course_id):
 
         return result
 
+    def is_enrolled(self, username, course_id):
```
douglashall: Can we move towards standardizing on `course_run_id` when referring to the unique string identifier for a course run?

enterprise/views.py
```diff
@@ -1137,3 +1138,136 @@ def post(self, request, enterprise_uuid, program_uuid):
...
+            course_id
+        )
+        customer_catalog_contains_course = enterprise_customer.catalog_contains_course(course_id)
+        return request_has_audit_query_param and is_course_enrollment_url and customer_catalog_contains_course
+
+    def enroll_learner(self, request, enterprise_customer, course_id, mode):
+        """
+        Enroll a user into a course track, and register enterprise course enrollment & data sharing consent objects.
+        """
+        with transaction.atomic():
```
douglashall: I think you can get rid of this atomic block since you are only performing a single DB update inside of it.

enterprise/views.py
```diff
@@ -1137,3 +1138,136 @@ def post(self, request, enterprise_uuid, program_uuid):
...
+
+        enrollment_api_client = EnrollmentApiClient()
+
+        # Check to see if the user's already enrolled and we have an enterprise course enrollment to track it.
+        enrolled_in_course = enrollment_api_client.is_enrolled(request.user.username, course_id)
+        enterprise_course_enrollment_exists = EnterpriseCourseEnrollment.objects.filter(
+            enterprise_customer_user__enterprise_customer=enterprise_customer,
+            enterprise_customer_user__user_id=request.user.id,
+            course_id=course_id
+        ).exists()
```
douglashall: Instead of this `exists` statement, replace it with an `EnterpriseCourseEnrollment.objects.get_or_create`.

enterprise/views.py
```diff
@@ -1137,3 +1138,136 @@ def post(self, request, enterprise_uuid, program_uuid):
...
+        if not (enrolled_in_course and enterprise_course_enrollment_exists):
```
douglashall: Just check `enrolled_in_course` here.

enterprise/views.py
```diff
@@ -1137,3 +1138,136 @@ def post(self, request, enterprise_uuid, program_uuid):
...
+            # Directly enroll into the audit track.
+            EnterpriseCourseEnrollment.objects.update_or_create(
+                enterprise_customer_user=enterprise_customer_user,
+                course_id=course_id,
+            )
```
douglashall: Remove this `update_or_create` call.

enterprise/views.py
```diff
@@ -1137,3 +1138,136 @@ def post(self, request, enterprise_uuid, program_uuid):
...
+        """
+        request_has_audit_query_param = request.GET.get('audit')
+        is_course_enrollment_url = request.path == self.COURSE_ENROLLMENT_VIEW_URL.format(
+            enterprise_customer.uuid,
+            course_id
+        )
+        customer_catalog_contains_course = enterprise_customer.catalog_contains_course(course_id)
+        return request_has_audit_query_param and is_course_enrollment_url and customer_catalog_contains_course
+
+    def enroll_learner(self, request, enterprise_customer, course_id, mode):
```
douglashall: It feels like this function should be in the `utils` module or perhaps a method on `EnterpriseCustomer`.

enterprise/views.py
```diff
@@ -1137,3 +1138,136 @@ def post(self, request, enterprise_uuid, program_uuid):
...
+            self.track_event(request.user.id, 'edx.bi.user.enterprise.enrollment.course', {
+                'category': 'enterprise',
+                'label': course_id,
+                'enterprise_customer_uuid': str(enterprise_customer.uuid),
+                'enterprise_customer_name': enterprise_customer.name,
+                'mode': mode,
+            })
+
+        # The courseware view logic will check for DSC requirements, and route to the DSC page if necessary.
+        return redirect(LMS_COURSEWARE_URL.format(course_id=course_id))
```
douglashall: Pass the `request.user` into this function and just have it do the enrollment. Let the `get` method handle the redirecting.

enterprise/views.py
```diff
@@ -1137,3 +1138,136 @@ def post(self, request, enterprise_uuid, program_uuid):
...
+
+    def redirect(self, request, *args, **kwargs):
+        """
+        Redirects to the appropriate view depending on where the user came from.
+        """
+        enterprise_customer_uuid, resource_id = RouterView.get_path_variables(**kwargs)
+        for path, view_class in self.VIEWS.items():
```
douglashall: It would be better if we didn't have to loop through all the views here. Could we use a regular expression to replace the `enterprise_customer_uuid` and `resource_id` values in the path with `{}`, then use the result as the key on the `VIEWS` dict?

enterprise/views.py
```diff
@@ -1137,3 +1138,136 @@ def post(self, request, enterprise_uuid, program_uuid):
...
+        return self.redirect(request, *args, **kwargs)
+
+    @method_decorator(enterprise_login_required)
+    def post(self, request, *args, **kwargs):
+        """
+        Run some custom POST logic for Enterprise workflows before routing the user through existing views.
+        """
+        return self.redirect(request, *args, **kwargs)
+
+    def track_event(self, user_id, event_name, properties):
```
douglashall: Move this to the `utils` module.

enterprise/api_client/lms.py
UmanShahzad: Yes, good idea.

enterprise/views.py
UmanShahzad: Or on `EnterpriseCustomerUser`, as then we'd just have to pass a `course_run_id` and `mode`.

enterprise/views.py
UmanShahzad: :+1:

enterprise/views.py
UmanShahzad: :+1:

enterprise/views.py
UmanShahzad: :+1:

enterprise/views.py
UmanShahzad: :+1: but I will be moving the new `get_or_create` all the way to the bottom, to ensure we safely enroll the user in the course itself first before creating the enterprise course enrollment object.

enterprise/views.py
UmanShahzad: :+1:

enterprise/views.py
UmanShahzad: :+1:

enterprise/views.py
UmanShahzad: Yep, great idea.

enterprise/views.py
UmanShahzad: Actually, apparently we need this: https://github.com/edx/edx-enterprise/blob/master/enterprise/views.py#L763-L766

douglashall: Ah, right. Makes sense.

UmanShahzad:
@douglashall Review addressed.

UmanShahzad:
Rebased on `master` from0611973c6af7b658999bf88ea00d25db34a607f4.

enterprise/views.py
```diff
@@ -1137,3 +1119,91 @@ def post(self, request, enterprise_uuid, program_uuid):
...
+        return request.GET.get('audit') and \
+            request.path == self.COURSE_ENROLLMENT_VIEW_URL.format(enterprise_customer.uuid, course_run_id) and \
+            enterprise_customer.catalog_contains_course(course_run_id)
+
+    def redirect(self, request, *args, **kwargs):
+        """
+        Redirects to the appropriate view depending on where the user came from.
+        """
+        enterprise_customer_uuid, resource_id = RouterView.get_path_variables(**kwargs)
+        path = re.sub('{}|{}'.format(enterprise_customer_uuid, re.escape(resource_id)), '{}', request.path)
```
haikuginger: This line is a bit opaque to me; can you comment it out?

enterprise/views.py
```diff
@@ -1137,3 +1119,91 @@ def post(self, request, enterprise_uuid, program_uuid):
...
+        - `enterprise_uuid` - the UUID of the enterprise customer.
+        - `course_run_id` - the ID of the course, if applicable.
+        - `program_uuid` - the UUID of the program, if applicable.
+        """
+        enterprise_customer_uuid = kwargs.get('enterprise_uuid', '')
+        course_run_id = kwargs.get('course_id', '')
+        program_uuid = kwargs.get('program_uuid', '')
+        return enterprise_customer_uuid, course_run_id or program_uuid
+
+    def eligible_for_direct_audit_enrollment(self, request, enterprise_customer, course_run_id):
```
haikuginger: Can we add a check to make sure that `audit` is a valid enrollment mode for the given course?

enterprise/views.py
UmanShahzad: Yes sure. Essentially we're turning the request's path into something that can match one in the class's `VIEWS`, in order to directly and easily match a corresponding view class. Before, we were looping over the dictionary to find a match, but regex is scalable (albeit harder to read, so I'll add that comment).

enterprise/views.py
UmanShahzad: Yes :+1:

enterprise/views.py
UmanShahzad: Addressed with 9f7a6fb745c5d3c7abf602745ba170459d1e62ef

enterprise/views.py
UmanShahzad: Addressed with c9d516a4f7b0ce80bacfb93cf6088648973887f2

UmanShahzad:
@haikuginger Thanks for the review :) I've now addressed it, however the new check for whether the course has an audit mode is currently not on the sandbox.

enterprise/views.py
UmanShahzad: Forgot to update docstring, though.

enterprise/api_client/lms.py
UmanShahzad: Ah I forgot to change it in this function; only changed things in the `RouterView`. Updating...

enterprise/api_client/lms.py
```diff
@@ -229,6 +243,20 @@ def get_course_enrollment(self, username, course_id):
 
         return result
 
+    def is_enrolled(self, username, course_run_id):
```
haikuginger: There's a bug in this function - an enrollment can exist, but be inactive. This would appear to the user as if they were unenrolled, but the API will still return a record. If a record isn't returned, there's necessarily no enrollment, but if a record is returned, you'll need to look for the `is_active` key, as below:

```json
{
    "created":"2017-10-13T14:09:22.176782Z",
    "mode":"audit",
    "is_active":false,
    "course_details":{"course_id":"course-v1:edX+DemoX+Demo_Course","course_name":"edX Demonstration Course","enrollment_start":null,"enrollment_end":null,"course_start":"2013-02-05T05:00:00Z","course_end":null,"invite_only":false,"course_modes":[{"slug":"audit","name":"Audit","min_price":0,"suggested_prices":"","currency":"usd","expiration_datetime":null,"description":null,"sku":null,"bulk_sku":null}]},
    "user":"superegoe0df"
}
```

Hooray for manual testing ferreting this out!

UmanShahzad: Ah, nice. I'm going to push and update the sandbox in a minute, when my local tests pass.

UmanShahzad: This'll be on the sandbox in 5-10 minutes.

UmanShahzad: It's up.

enterprise/views.py
```diff
@@ -1137,3 +1119,95 @@ def post(self, request, enterprise_uuid, program_uuid):
...
+        # Ensure that the link is saved to the database prior to making some call in a downstream view
+        # which may need to know that the user belongs to an enterprise customer.
+        with transaction.atomic():
+            enterprise_customer_user, __ = EnterpriseCustomerUser.objects.get_or_create(
+                enterprise_customer=enterprise_customer,
+                user_id=request.user.id
+            )
+
+        # Directly enroll in audit mode if the request in question has full direct audit enrollment eligibility.
+        if self.eligible_for_direct_audit_enrollment(request, enterprise_customer, resource_id):
```
brittneyexline: one thing i'm wondering about with regards to firing an event in this case is whether it should live in that enroll method or here in the view. when we were discussing the story i had thought of this event as something that is recording the decision the router view was making, and as such, that it would live in the router view.
i don't think you should change your implementation but curious to hear thoughts on this.

UmanShahzad: Yes that's also what I had read in the description actually, so good point. But it turns out that the actual track event is so generic to enterprise course enrollments (it doesn't contain any special information about having gone through *this* view in particular), that it may just make sense to keep it in the `enroll` function. But I'm open to any other viewpoint!

UmanShahzad: ^ Examples of specific cases would be: `'url': request.path`.

douglashall: I like recording the general enterprise enrollment event here. In subsequent work, it would also be nice to track a similar event for all enterprise enrollments.

Perhaps we do need to add another event as well though that tracks the fact that the user came through the direct-to-audit pathway. @UmanShahzad Can you add that?

UmanShahzad: @douglashall Do you mean to keep the event tracking function call inside `enroll` but fire another event here with information about this pathway?

douglashall: Almost, I think the other event relevant to the pathway should be in the view as @brittneyexline suggested.

brittneyexline: agreed. it makes sense to keep the event in enroll, but other things might call that in the future. conversely, we're also interested in the decisions the router view makes, so two events makes sense.

UmanShahzad: @brittneyexline Ok that makes sense!

UmanShahzad: @douglashall @brittneyexline I've added that specific event now before the redirect in this pathway.

UmanShahzad: (805db539b4cef020dac0b29315a8d2f9b8746beb)

UmanShahzad: Hopefully I didn't misunderstand something here, though.

PR #214, opened 2017-10-10
fix null site when creating new enterprise_customer

bill-filler:
**Description:** fix regression caused by previous fix. Now check for null site when creating a new enterprise_customer. Added a test for this as well.

**Testing instructions:**

1. Create a new enterprise_customer via http://localhost:18000/admin/enterprise/enterprisecustomer/add/
2. Make sure the form is properly displayed

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/admin/forms.py
+++ b/enterprise/admin/forms.py
@@ -332,7 +332,16 @@ class EnterpriseCustomerAdminForm(forms.ModelForm):
         Once retrieved, these name pairs can be used directly as a value
         for the `choices` argument to a ChoiceField.
         """
-        catalog_api = CourseCatalogApiClient(self.user, self.instance.site)
+        # TODO: We will remove the discovery service catalog implementation
+        # once we have fully migrated customer's to EnterpriseCustomerCatalogs.
+        # For now, this code will prevent an admin from creating a new
+        # EnterpriseCustomer with a discovery service catalog. They will have to first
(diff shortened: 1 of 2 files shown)


enterprise/admin/forms.py
```diff
@@ -332,7 +332,10 @@ def get_catalog_options(self):
         Once retrieved, these name pairs can be used directly as a value
         for the `choices` argument to a ChoiceField.
         """
-        catalog_api = CourseCatalogApiClient(self.user, self.instance.site)
+        if hasattr(self.instance, 'site'):
```
douglashall: Can you add the following comment here?
```
# TODO: We will remove the discovery service catalog implementation
# once we have fully migrated customer's to EnterpriseCustomerCatalogs.
# For now, this code will prevent an admin from creating a new
# EnterpriseCustomer with a discovery service catalog. They will have to first
# save the EnterpriseCustomer admin form and then edit the EnterpriseCustomer
# to add a discovery service catalog.

bill-filler: done

douglashall:
@bill-filler Can you bump the version and changelog?

PR #217, opened 2017-10-13
[ENT-697]: Update EnterpriseApiClient.get_enterprise_courses to account for EnterpriseCustomerCatalogs

saleem-latif:
**Description:** 
EnterpriseApiClient.get_enterprise_courses is the Enterprise API client method used to get all of the course runs that exist in the EnterpriseCustomer's catalog. Currently, this method only accounts for the legacy discovery service catalog implementation. We need to modify this method to also collect the course runs that exist in any EnterpriseCustomerCatalogs associated with the EnterpriseCustomer.
The result set returned from EnterpriseApiClient.get_enterprise_courses should contain the set of unique course runs from both the discovery service catalog implementation and all EnterpriseCustomerCatalogs associated with the EnterpriseCustomer.

**JIRA:** [ENT-697](https://openedx.atlassian.net/browse/ENT-697)

**Installation instructions:** Simply install `edx-enterprise` using this branch.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag


enterprise/api_client/enterprise.py
```diff
@@ -28,20 +30,53 @@ class EnterpriseApiClient(JwtLmsApiClient):
...
+            )
+
+        for enterprise_customer_catalog in enterprise_customer.enterprise_customer_catalogs.all():
+            catalog_courses = discovery_api.get_search_results(enterprise_customer_catalog.content_filter)
+            response = self._merge_search_all_response(response, catalog_courses)
+
+        return response
+
+    @staticmethod
+    def _merge_search_all_response(first, search_results):
```
zubair-arbi:
```python
def _update_discovery_api_search_results(all_api_search_results, search_results):
```

enterprise/api_client/enterprise.py
```diff
@@ -28,20 +30,53 @@ class EnterpriseApiClient(JwtLmsApiClient):
...
+        if not search_results:
+            return first
+
+        existing_course_keys = {item['key'] for item in first['results']} if first else set()
+
+        count = 0
+        for item in search_results:
+            if item['key'] not in existing_course_keys:
+                count += 1
```
zubair-arbi: Instead of using placeholder `count` can we do this?
```python
first['count'] += count
```

integrated_channels/integrated_channel/course_metadata.py
```diff
@@ -51,7 +51,6 @@ def get_course_runs(user, enterprise_customer):
 
     enterprise_courses = client.get_enterprise_courses(
         enterprise_customer,
-        traverse_pagination=True
```
zubair-arbi: removal to this flag is intentional? Perhaps we could pass this flag `traverse_pagination` to the method itself.

saleem-latif: Yes, we have to remove this as we are getting all course runs by default. This method is not being used anywhere else, so it is safe to remove the parameter.

saleem-latif:
@zubair-arbi I have made the changes you suggested, can you take another look

enterprise/api_client/enterprise.py
```diff
@@ -28,20 +29,57 @@ class EnterpriseApiClient(JwtLmsApiClient):
...
+            response = self._load_data(
+                self.ENTERPRISE_CUSTOMER_ENDPOINT,
+                detail_resource='courses',
+                resource_id=str(enterprise_customer.uuid),
+                traverse_pagination=True,
+            )
+
+        for enterprise_customer_catalog in enterprise_customer.enterprise_customer_catalogs.all():
+            catalog_courses = discovery_api.get_search_results(enterprise_customer_catalog.content_filter)
+            response = self._update_discovery_api_search_results(response, catalog_courses)
```
douglashall: We want the EnterpriseCustomerCatalog data to override the discovery service catalog data.
```
for enterprise_customer_catalog in enterprise_customer.enterprise_customer_catalogs.all():
    for course_run in enterprise_customer_catalog.course_runs:
        course_runs[course_run['key']] = course_run
```

enterprise/api_client/enterprise.py
```diff
@@ -28,20 +29,57 @@ class EnterpriseApiClient(JwtLmsApiClient):
...
+        if enterprise_customer.catalog:
+            response = self._load_data(
+                self.ENTERPRISE_CUSTOMER_ENDPOINT,
+                detail_resource='courses',
+                resource_id=str(enterprise_customer.uuid),
+                traverse_pagination=True,
+            )
+
+        for enterprise_customer_catalog in enterprise_customer.enterprise_customer_catalogs.all():
+            catalog_courses = discovery_api.get_search_results(enterprise_customer_catalog.content_filter)
```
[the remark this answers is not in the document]
douglashall: Scratch that. This is the Enterprise API client. We should be calling the Enterprise API's `enterprise_customer_catalogs` endpoint here to retrieve the course run data. This will ensure that the course enrollment URL is populated.

PR #220, opened 2017-10-16
[ENT-648] Change log level from `error` to `info`

UmanShahzad:
**Description:** Change the log level from `error` to `info`.

**JIRA:** [ENT-648](https://openedx.atlassian.net/browse/ENT-648)

**Testing instructions:** Ensure there's no typos :)

### DIFF
--- a/enterprise/api_client/lms.py
+++ b/enterprise/api_client/lms.py
@@ -238,7 +238,7 @@ class EnrollmentApiClient(LmsApiClient):
         # This enrollment data endpoint returns an empty string if the username and course_id is valid, but there's
         # no matching enrollment found
         if not result:
-            LOGGER.error('no course enrollment details found for username=%s, course=%s', username, course_id)
+            LOGGER.info('Failed to find course enrollment details for user [%s] and course [%s]', username, course_id)
             return None
 
         return result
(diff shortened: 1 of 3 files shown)


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=220&repo=edx%2Fedx-enterprise"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

enterprise/api_client/lms.py
```diff
@@ -238,7 +238,7 @@ def get_course_enrollment(self, username, course_id):
         # This enrollment data endpoint returns an empty string if the username and course_id is valid, but there's
         # no matching enrollment found
         if not result:
-            LOGGER.error('no course enrollment details found for username=%s, course=%s', username, course_id)
+            LOGGER.info('no course enrollment details found for username=%s, course=%s', username, course_id)
```
UmanShahzad: @mattdrayer There's one at https://github.com/edx/edx-enterprise/pull/220/files#diff-c06e09cde492a184aeafc6e12c32fe49R232 right above this too -- would you want that to be `info`?

douglashall: @UmanShahzad The error log message you pointed out above should stay as error since that is an unexpected scenario.

Nit: While you are in here perhaps you could improve the message a bit:
```
LOGGER.info('Failed to find course enrollment details for user [%s] and course [%s]', username, course_id)
```

PR #227, opened 2017-10-24
ENT-726 Fix alteration in querystring parameters for enterprise login decorator

zubair-arbi:
@asadiqbal08 @saleem-latif @douglashall 

**Description:** Fix alteration in querystring parameters for decorator "enterprise_login_required" when learner is redirected from `SSO`.

**JIRA:** [ENT-726](https://openedx.atlassian.net/browse/ENT-726)

**Dependencies:** N/A

**Merge deadline:** End of sprint `V`.

**Installation instructions:** Install `edx-enterprise` from this branch `zub/ENT-726-fix-enterprise-login-querysting-alteration` in `edx-platform`.

**Testing instructions:**

1. Create a SAML IDP for while enabling the setting "Drop existing session"
2. Create enterprise with the above IDP
3. Create a related catalog in model `EnterpriseCustomerCatalog`
4. Access the enterprise course enrollment url with the catalog query parameter (value should be the UUID of above created `EnterpriseCustomerCatalog` records) with url like:
```
http://zubair.localhost:8000/enterprise/72416e52-8c77-4860-9584-15e5b06220fb/course/course-v1:edX+DemoX+Demo_Course/enroll/?catalog=6eca3efb-f3a0-4c08-806f-c6e6b65d61cb
```
5. Now verify that you can view the course enrollment page after redirecting from SSO (valid credentials). Note that the value for `catalog` query parameter in the url is string instead of a list.

**Merge checklist:**

- [x] All reviewers approved
- [ ] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/decorators.py
+++ b/enterprise/decorators.py
@@ -141,7 +141,7 @@ def enterprise_login_required(view):
             })
             next_url = '{current_path}?{query_string}'.format(
                 current_path=quote(parsed_current_url.path),
-                query_string=urlencode(parsed_query_string)
+                query_string=urlencode(parsed_query_string, doseq=True)
             )
             return redirect(
                 '{login_url}?{params}'.format(
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.53.1] - 2017-10-24
+----------------------
+
+* Fix alteration in querystring parameters for decorator "enterprise_login_required".
+
 [0.53.0] - 2017-10-24
 ----------------------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.53.0"
+__version__ = "0.53.1"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/tests/test_decorators.py
+++ b/tests/test_decorators.py
@@ -140,6 +140,40 @@ class TestEnterpriseDecorators(unittest.TestCase):
         assert 'new_enterprise_login%3Dyes' in response.url
         assert 'tpa_hint' in response.url
 
+    def test_enterprise_login_required_redirects_for_anonymous_users_with_querystring(self):
+        """
+        Test that the decorator `enterprise_login_required` returns Http
+        Redirect for anonymous users while keeping the format of query
+        parameters unchanged.
+        """
+        view_function = mock_view_function()
+        course_id = 'course-v1:edX+DemoX+Demo_Course'
+        course_enrollment_url = reverse(
+            'enterprise_course_enrollment_page',
+            args=[self.customer.uuid, course_id],
+        )
+        querystring = 'catalog=dummy-catalog-uuid'
+        course_enrollment_url = '{course_enrollment_url}?{querystring}'.format(
+            course_enrollment_url=course_enrollment_url, querystring=querystring
+        )
+        request = self._prepare_request(course_enrollment_url, AnonymousUser())
+
+        response = enterprise_login_required(view_function)(
+            request, enterprise_uuid=self.customer.uuid, course_id=course_id
+        )
+
+        # Assert that redirect status code 302 is returned when an anonymous
+        # user tries to access enterprise course enrollment page.
+        assert response.status_code == 302
+
+        # Now verify that the query parameters in the querystring of next url
+        # are unchanged
+        next_url = parse_qs(urlparse(response.url).query)['next'][0]
+        next_url_querystring = unquote(urlparse(next_url).query)
+        assert 'new_enterprise_login=yes' in next_url_querystring
+        assert 'tpa_hint' in next_url_querystring
+        assert querystring in next_url_querystring
+
     def test_enterprise_login_required(self):
         """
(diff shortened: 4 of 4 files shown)


enterprise/decorators.py
```diff
@@ -134,11 +134,15 @@ def wrapper(request, *args, **kwargs):
...
-            parsed_query_string.update({
+            parsed_query_string = parse_qsl(parsed_current_url.query)
+
+            additional_query_parameters = {
                 'tpa_hint': enterprise_customer.identity_provider,
                 FRESH_LOGIN_PARAMETER: 'yes'
-            })
+            }
+            for query_parameter_key, query_parameter_value in additional_query_parameters.iteritems():
+                parsed_query_string.append((query_parameter_key, query_parameter_value))
```
saleem-latif: use 
```
from six import iteritems
```
and use that to get items iterator for dict.

zubair-arbi: Done

enterprise/decorators.py
```diff
@@ -134,11 +140,15 @@ def wrapper(request, *args, **kwargs):
...
-            parsed_query_string = parse_qs(parsed_current_url.query)
-            parsed_query_string.update({
+            parsed_query_string = parse_qsl(parsed_current_url.query)
+
+            additional_query_parameters = {
                 'tpa_hint': enterprise_customer.identity_provider,
                 FRESH_LOGIN_PARAMETER: 'yes'
-            })
+            }
+            for query_parameter_key, query_parameter_value in iteritems(additional_query_parameters):
```
asadiqbal08: Not just `key` and `value` would be nice instead of `query_parameter_key/value` ?

asadiqbal08:
👍

enterprise/decorators.py
```diff
@@ -134,11 +140,15 @@ def wrapper(request, *args, **kwargs):
         # Enterprise-linked IdP and the pipeline will get them back here.
         if not request.user.is_authenticated():
             parsed_current_url = urlparse(request.get_full_path())
-            parsed_query_string = parse_qs(parsed_current_url.query)
-            parsed_query_string.update({
+            parsed_query_string = parse_qsl(parsed_current_url.query)
```
douglashall: How does parsing this to a list of key/value pairs instead of a dictionary solve the problem?

zubair-arbi: @douglashall previously the `parse_qs` was converting the values of all the existing query params to list so the `catalog=6eca3efb-f3a0-4c08-806f-c6e6b65d61cb` would become `{'catalog': ['6eca3efb-f3a0-4c08-806f-c6e6b65d61cb'] }`. `parse_qs` does this to combine the values of same params.

`parse_qsl` on the other hand converts the query params to tuples this way the query param values remain unchanged.

https://stackoverflow.com/questions/1024143/how-to-stop-python-parse-qs-from-parsing-single-values-into-lists

enterprise/decorators.py
zubair-arbi: I think that verbose variable names make the statement more readable.

enterprise/decorators.py
douglashall: @zubair-arbi Got it. Makes sense. Thanks.

douglashall: Actually, no. There's simpler fix here. See below.

enterprise/decorators.py
```diff
@@ -134,11 +140,15 @@ def wrapper(request, *args, **kwargs):
...
                 'tpa_hint': enterprise_customer.identity_provider,
                 FRESH_LOGIN_PARAMETER: 'yes'
-            })
+            }
+            for query_parameter_key, query_parameter_value in iteritems(additional_query_parameters):
+                parsed_query_string.append((query_parameter_key, query_parameter_value))
+
             next_url = '{current_path}?{query_string}'.format(
                 current_path=quote(parsed_current_url.path),
                 query_string=urlencode(parsed_query_string)
```
douglashall: Instead of using `parse_qsl` above, just add `doseq=True` to the call to `urlencode` here.

enterprise/decorators.py
zubair-arbi: @douglashall I intentionally didn't used that approach as it says ` Please note that if there are multiple values for the same key, you'll only get the last one.` I know we aren't currently using multiple values for the same key but I wanted to implement a flexible solution. Please let me know your thoughts.

enterprise/decorators.py
asadiqbal08: agree but For local variables that are only relevant for a short number of lines, having long verbose names actually makes them less efficient :)

enterprise/decorators.py
douglashall: @zubair-arbi That comment you are referencing from the stackoverflow article you linked to has nothing to do with the `doseq` approach.

enterprise/decorators.py
zubair-arbi: @douglashall Thanks for this nice suggestion: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode

zubair-arbi:
@douglashall I have update the PR to use `doseq=True` kwarfs to the call to `urlencode` instead of using `parse_qsl`. Please have another look.

zubair-arbi:
@douglashall I have rebased the branch onto master and squashed it after adding version. You can merge if when the build is green.

douglashall:
Thanks @zubair-arbi! Will do.

PR #228, opened 2017-10-24
PLAT-1780 Remove dependency on django-extensions

jmbowman:
**Description:** Removes the dependency on django-extensions, which doesn't seem to actually be used.  Trying to upgrade it in edx-platform to a version which fully supports Django 1.11 is causing errors in seemingly unrelated edx-enterprise migrations, so I'm attempting removal instead.

**JIRA:** [PLAT-1780](https://openedx.atlassian.net/browse/PLAT-1780)

**Dependencies:** None

**Merge deadline:** None

**Testing instructions:**

1. Verify that automated tests pass

**Merge checklist:**

- [x] Check that the versions of the requirements in the `test-master.in` file match edx-platform.
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:** It looks like this dependency was never actually used; it may have been added in confusion because of `TimeStampedModel`, which exists in both django-extensions and django-model-utils, but only code from the latter is used here.  If there's another legitimate reason for keeping the dependency, please let me know.  This shouldn't require a new version of edx-enterprise, since the only changes are in requirements files for testing and local development.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.53.2] - 2017-10-24
+---------------------
+
+* Remove unused dependency on django-extensions
+
 [0.53.1] - 2017-10-24
 ----------------------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.53.1"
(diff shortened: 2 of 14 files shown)


jmbowman:
@douglashall Done.  I don't think we actually need an edx-platform update on this one (the changes don't really propagate into the uploaded package), but I can open a PR to upgrade to this new edx-enterprise version if you'd like.

PR #229, opened 2017-10-26
Applied appropriate content filtering to the EnterpriseCustomerCatalog detail endpoint

asadiqbal08:
**Description:** Applied appropriate content filtering for courseruns to the EnterpriseCustomerCatalog detail endpoint

**JIRA:** [ENT-687](https://openedx.atlassian.net/browse/ENT-687)

#### NOTE: Kindly ignore the failing tests at that point which are failing due to programs content ids.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -915,10 +915,25 @@ class EnterpriseCustomerCatalog(TimeStampedModel):
         Returns:
             dict: The paginated discovery service search results.
         """
+        results = []
         query = self.content_filter.copy()
         query.update(query_parameters)
         catalog_client = CourseCatalogApiServiceClient(self.enterprise_customer.site)
-        return catalog_client.get_search_results(query, traverse_pagination=False)
+        search_results = catalog_client.get_search_results(query, traverse_pagination=False)
+        for content in search_results['results']:
+            if content['content_type'] == 'courserun' and content['has_enrollable_seats']:
+                results.append(content)
+            elif content['content_type'] == 'program' and content['is_program_eligible_for_one_click_purchase']:
+                results.append(content)
+
+        response = {
+            'count': len(results),
+            'next': search_results['next'],
+            'previous': search_results['previous'],
+            'results': results,
+            }
+
+        return response
 
     def contains_content_items(self, content_id_field_name, content_id_values):
         """
@@ -942,7 +957,11 @@ class EnterpriseCustomerCatalog(TimeStampedModel):
             updated_content_filter[content_id_field_name] = content_id_values
             results = CourseCatalogApiServiceClient().get_search_results(updated_content_filter)
             if results:
-                content_ids_in_catalog = {content[content_id_field_name] for content in results}
+                for content in results:
+                    if content['content_type'] == 'courserun' and content['has_enrollable_seats']:
+                        content_ids_in_catalog.add(content[content_id_field_name])
+                    elif content['content_type'] == 'program' and content['is_program_eligible_for_one_click_purchase']:
+                        content_ids_in_catalog.add(content[content_id_field_name])
 
         # Diff the content IDs found in the catalog with the set of
         # IDs which we are checking for existence.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.53.9] - 2017-11-02
+---------------------
+
+* Apply appropriate content filtering to the EnterpriseCustomerCatalog detail endpoints.
+
 [0.53.8] - 2017-11-02
 ---------------------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.53.8"
+__version__ = "0.53.9"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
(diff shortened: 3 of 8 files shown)


asadiqbal08:
@douglashall It is a WIP PR at that point.

enterprise/models.py
```diff
@@ -942,7 +954,12 @@ def contains_content_items(self, content_id_field_name, content_id_values):
             updated_content_filter[content_id_field_name] = content_id_values
             results = CourseCatalogApiServiceClient().get_search_results(updated_content_filter)
```
douglashall: Instead of filtering the results below, just add the `has_enrollable_seats` and `is_program_eligible_for_one_click_purchase` flags to the `updated_content_filter`.

enterprise/models.py
```diff
@@ -915,10 +915,22 @@ def get_paginated_content(self, query_parameters):
...
+        for content in search_results['results']:
+            if content.get('content_type', '') == 'courserun' and content.get('has_enrollable_seats', False):
+                results.append(content)
+            elif content.get('content_type', '') == 'program' and \
+                    content.get('is_program_eligible_for_one_click_purchase', False):
+                results.append(content)
+            elif content.get('content_type', '') == 'course':
+                results.append(content)
+
+        search_results['results'] = results
```
asadiqbal08: updating that code in next commit.

enterprise/models.py
asadiqbal08: @douglashall do you mean that we should update the OrderDict `updated_content_filter` with flags e.g. 

```
updated_content_filter.update({
            "has_enrollable_seats": True,
            "is_program_eligible_for_one_click_purchase": True
}) ???
Do you think will that work as course and programs have different flags values in them. I think, It will not fetch the data. correct me If I am missing your point. 
```

douglashall: Ah, good point. I guess we can't do that unfortunately.

enterprise/models.py
```diff
@@ -942,7 +954,12 @@ def contains_content_items(self, content_id_field_name, content_id_values):
             updated_content_filter[content_id_field_name] = content_id_values
             results = CourseCatalogApiServiceClient().get_search_results(updated_content_filter)
             if results:
-                content_ids_in_catalog = {content[content_id_field_name] for content in results}
+                for content in results:
+                    if content.get('content_type', '') == 'courserun' and content.get('has_enrollable_seats', False):
```
douglashall: No need to use `get` here, course run items will always have the `content_type` and `has_enrollable_seats` keys. Same thing below with programs.

enterprise/models.py
```diff
@@ -915,10 +915,22 @@ def get_paginated_content(self, query_parameters):
         Returns:
             dict: The paginated discovery service search results.
         """
+        results = []
         query = self.content_filter.copy()
         query.update(query_parameters)
         catalog_client = CourseCatalogApiServiceClient(self.enterprise_customer.site)
-        return catalog_client.get_search_results(query, traverse_pagination=False)
+        search_results = catalog_client.get_search_results(query, traverse_pagination=False)
+        for content in search_results['results']:
```
douglashall: Save yourself a variable.
```
for content in catalog_client.get_search_results(query, traverse_pagination=False):
```

enterprise/models.py
```diff
@@ -915,10 +915,22 @@ def get_paginated_content(self, query_parameters):
...
+            if content.get('content_type', '') == 'courserun' and content.get('has_enrollable_seats', False):
```
douglashall: Same thing here. No need for `get`.

enterprise/models.py
```diff
@@ -915,10 +915,22 @@ def get_paginated_content(self, query_parameters):
...
+                results.append(content)
+            elif content.get('content_type', '') == 'program' and \
+                    content.get('is_program_eligible_for_one_click_purchase', False):
+                results.append(content)
+            elif content.get('content_type', '') == 'course':
```
douglashall: Let's leave this `else` clause out for now and only include course runs and programs.

enterprise/models.py
```diff
@@ -915,10 +915,27 @@ def get_paginated_content(self, query_parameters):
...
         query.update(query_parameters)
         catalog_client = CourseCatalogApiServiceClient(self.enterprise_customer.site)
-        return catalog_client.get_search_results(query, traverse_pagination=False)
+        search_results = catalog_client.get_search_results(query, traverse_pagination=False)
+        for content in search_results['results']:
+            if content['content_type'] == 'courserun' and content['has_enrollable_seats']:
+                results.append(content)
+            elif content['content_type'] == 'program' and content['is_program_eligible_for_one_click_purchase']:
+                results.append(content)
+            elif content['content_type'] == 'course':
```
asadiqbal08: I have update the PR. @douglashall I think, As we are filtering the `programs` and `courserun` but we should keep the `course` contents in pagination detail results at that point that may be useful to get some course information? thoughts?

douglashall: @asadiqbal08 As of now we are only supporting programs and course runs as content items in enterprise catalogs. This may change in the future, but for now we should remove this.

asadiqbal08:
DONE

enterprise/models.py
douglashall: You don't need the results variable.
```
search_results = []
for content in catalog_client.get_search_results(query, traverse_pagination=False):
```
And then append valid content items to the `search_results` list.

asadiqbal08: @douglashall I am using that variable [HERE](https://github.com/edx/edx-enterprise/pull/229/files#diff-4bcae564bc272f5c2e6753dc6021eb1cR931). Is that variable making the statement complex?

douglashall: Oh, nevermind. I see why you need it now. Thanks.

PR #231, opened 2017-10-27
ENT-735: Fetch catalog courses in large chunks to avoid API limit

saleem-latif:
**Description:** 
`transmit_courseware_data` management command errors out with 429. This PR fixes this issue.

**JIRA:**  [ENT-735](https://openedx.atlassian.net/browse/ENT-735)

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/enterprise/api_client/enterprise.py
+++ b/enterprise/api_client/enterprise.py
@@ -56,6 +56,8 @@ class EnterpriseApiClient(JwtLmsApiClient):
                 self.ENTERPRISE_CUSTOMER_CATALOGS_ENDPOINT,
                 resource_id=str(enterprise_customer_catalog.uuid),
                 traverse_pagination=True,
+                # we need to fetch data in large chunks so that we do not hit api limit.
+                querystring={'page_size': 1000},
             )
 
             course_runs.update(self.get_course_runs_from_search_results(response['results']))
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.53.5] - 2017-10-26
+---------------------
+
+* Fetch catalog courses in large chunks to avoid API limit.
+
 [0.53.4] - 2017-10-26
 ---------------------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.53.4"
+__version__ = "0.53.5"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name


enterprise/api_client/enterprise.py
```diff
@@ -56,6 +56,7 @@ def get_enterprise_course_runs(self, enterprise_customer):
                 self.ENTERPRISE_CUSTOMER_CATALOGS_ENDPOINT,
                 resource_id=str(enterprise_customer_catalog.uuid),
                 traverse_pagination=True,
+                querystring=500,  # we need to fetch data in large chunks so that we do not hit api limit.
```
douglashall: This should be `querystring={'page_size': 5000}`, no?

douglashall: The tricky part is going to be getting the EnterpriseCustomerCatalog detail endpoint to use the page_size parameter when pulling the content items from the discovery service [here](https://github.com/edx/edx-enterprise/blob/master/enterprise/api/v1/serializers.py#L205).

saleem-latif: you mean `querystring={'page_size': 500}`  right ?. 5000 may cause `Timeout`.

saleem-latif: >The tricky part is going to be getting the EnterpriseCustomerCatalog detail endpoint to use the page_size parameter when pulling the content items from the discovery service [here](https://github.com/edx/edx-enterprise/blob/master/enterprise/api/v1/serializers.py#L205).

I think it should work by just this update as `request.GET` would already have `page_size` parameter. I tested this locally and with a very small data set, it is working. I will deploy it the sandbox for further testing but I think more accurate testing will be performed only on stage where we have large data sets.

saleem-latif: never mind, I saw you comment on the ticket

saleem-latif:
@zubair-arbi , @douglashall Build is green and also tested it on business sandbox. Please take a look

enterprise/api_client/enterprise.py
```diff
@@ -56,6 +56,8 @@ def get_enterprise_course_runs(self, enterprise_customer):
                 self.ENTERPRISE_CUSTOMER_CATALOGS_ENDPOINT,
                 resource_id=str(enterprise_customer_catalog.uuid),
                 traverse_pagination=True,
+                # we need to fetch data in large chunks so that we do not hit api limit.
+                querystring={'page_size': 1000},
```
douglashall: Isn't [this](https://github.com/edx/edx-enterprise/blob/54003d767ae81b42de1357b464c2c86ea64490dc/enterprise/models.py#L921) going to prevent us from traversing pagination?

saleem-latif: Actually we are traversing pagination [here](https://github.com/edx/edx-enterprise/blob/master/enterprise/api_client/enterprise.py#L126).

[This](https://github.com/edx/edx-enterprise/blob/54003d767ae81b42de1357b464c2c86ea64490dc/enterprise/models.py#L921) part of the code causes enterprise API to return paginated response and [this](https://github.com/edx/edx-enterprise/blob/master/enterprise/api_client/enterprise.py#L126) part makes sure to traverse pagination of enterprise API and return the all the content.

douglashall: Got it. Makes sense. Thanks.

Can we move the addition of the `page_size` query parameter to the `utils.traverse_pagination function`? If we are traversing pagination, seems like we should always be doing that in the largest chunks possible.

saleem-latif: `utils.traverse_pagination` [follows pagination based on `next`](https://github.com/edx/edx-enterprise/blob/master/enterprise/utils.py#L541) url in the response dict which already contain `page`, `page_size`, `offset`, `limit` etc. parameters. 

So, I am not sure if it is possible to add `page_size` argument to `utils.traverse_pagination`. The reason is that `utils.traverse_pagination` is designed to be independent of pagination style (e.g. page, page_size vs limit, offset).

douglashall: Makes sense. Thanks for clarifying.

PR #235, opened 2017-10-30
ENT-742 Enrollment API

brittneyexline:
**Description:** 
Introduces a new endpoint for enrolling a single user in a single course run for a given enterprise.

See https://github.com/edx/edx-enterprise/pull/237 for the initial prototype this is based on.

**JIRA:** https://openedx.atlassian.net/browse/ENT-742

**Merge deadline:** 

**Testing instructions:**
The endpoint can be accessed using POST calls only at: {base_url}/enterprise/api/v1/enterprise-customer/{enterprise_uuid}/course_enrollments
Only users that have create access to Enterprise Customer models can hit this endpoint. This is temporary until the Authorization story is handled.

The fields to be sent in the POST data are as follows:
lms_user_id - the edX user id
tpa_user_id - the IdP user id
user_email - the user's email address
course_run_id - the course run ID to enroll the user in
course_mode - the course mode to enroll the user in (audit, verified, professional)
email_students - whether or not to email the students to notify them of the enrollment. defaults to false

The endpoint accepts a list of payloads with the above params. So for an individual enrollment call, an example payload would be:
[{
	"course_run_id":"course-v1:edX+DemoX+Demo_Course",
	"course_mode":"verified",
	"user_email":"bexline+enapitest4@edx.org",
	"email_students": true
}]

Multiple would be sent like so:
[
  {
	"course_run_id":"course-v1:edX+DemoX+Demo_Course",
	"course_mode":"verified",
	"user_email":"bexline+enapitest4@edx.org",
	"email_students": true
  },
  {
	"course_run_id":"course-v1:edX+DemoX+Demo_Course",
	"course_mode":"audit",
	"tpa_user_id":"abcdefg"
  }
]

The corresponding response will return an array with dictionaries mapping to each item in the original request. Successful responses will appear as {'detail': 'success'}, and unsuccessful responses will have a message indicating what the failure was for that payload.

Testing basic field validation - course_run_id and course_mode are always required, and at least one of lms_user_id, tpa_user_id, and user_email must be sent. The course_run_id must be part of the Enterprise Customer's Catalog. Making calls that do not meet these requirements should return responses that detail what is missing from the request data.

Note that for calls with multiple payloads, validation is processed serially. If one fails, all other payloads after it in the sequence will not be processed and return an empty value in the response.

Testing user identifier field validation - because the user identifier fields require at least one be specified, and multiple user identifier fields can be sent in one request, ie sending both lms_user_id and tpa_user_id together, there is some additional complexity in the validation. If more than one is specified, the field which results in a successful lookup of an EnterpriseCustomerUser is used, with preference given to lms_user_id, then tpa_user_id, then user_email. If user_email is given and no user is associated with it, and the other fields are not given or are also not associated with a user, a Pending Enterprise User and Pending Enrollment will be created. This fall back does not exist for the other two fields - if they are specified, no user can be found, and user_email is not specified, then the request will return an error indicating this.

Additionally, if the user is enrolled in a paid mode (verified or professional) and the request is attempting to enroll in audit, there will be a message specific to that and no change to enrollment will be performed. This will not impact later payloads if there are multiple sent. The reverse of this is allowed and will affect the enrollment, meaning a user already enrolled in audit or honor can be upgraded to verified/professional through this endpoint. Enrollment calls that attempt to enroll the user in their same mode will behave as if successful.

Testing enrolled user experience - this should behave as it does for users enrolled through the django admin interface. When they log in, they will see the course on their dashboard. If data sharing consent is enabled, they will be prompted for consent when they try to access the course for the first time. For pending users, this will happen after they register with the email they were enrolled with.

Testing email_students flag - if this is set to True, it will behave as the django admin interface and send an email using the same template to new or existing users about their enrollment.

Example Calls and their expected behaviors:
using a user that does not have add EnterpriseCustomer model access - Should give a message about insufficient permission.
no parameters - Should give a message about course_run_id and course_mode being required.
call with course_mode and course_run_id only - Should give a message about specifying one of the user identifier fields.
call with course_mode, course_run_id, and lms_user_id and/or tpa_user_id without an existing user - Should give a message indicating that no user could be found and user_email not being present.
call with course_mode, course_run_id, and user_email without an existing user - Should return a successful response, with a corresponding new PendingEnterpriseCustomerUser and PendingEnrollment created.
call with course_mode, course_run_id, and lms_user_id/tpa_user_id/user_mail with an existing user - Should return a successful response, and a new EnterpriseCourseEnrollment with the corresponding user and course should be created.

Remaining items:
*Links from the email do not trigger SSO for the Enterprise Customer. Depends on work from https://openedx.atlassian.net/browse/ENT-729. Once that is completed, we can change the url sent in emails to learners to include a tpa_hint. This might land separately from this PR depending on timing.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -6,7 +6,9 @@ from __future__ import absolute_import, unicode_literals
 
 import copy
 
+from edx_rest_api_client.exceptions import HttpClientError
 from rest_framework import serializers
+from rest_framework.settings import api_settings
 
 from django.contrib.auth.models import User
 from django.contrib.sites.models import Site
@@ -14,6 +16,7 @@ from django.utils.translation import ugettext_lazy as _
 
 from enterprise import models, utils
 from enterprise.api.v1.mixins import EnterpriseCourseContextSerializerMixin
+from enterprise.api_client.lms import ThirdPartyAuthApiClient
 
 
 class ImmutableStateSerializer(serializers.Serializer):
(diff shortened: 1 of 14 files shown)


mattdrayer:
Would it be possible to rename this endpoint to something resource-oriented, versus action-oriented?  For example if the "enrollment" is considered a resource it could be `/enrollment`.  See the [edX REST API Conventions](https://openedx.atlassian.net/wiki/spaces/OpenDev/pages/18350757/edX+REST+API+Conventions) document for more context.

brittneyexline:
@mattdrayer @douglashall so there is the EnterpriseCourseEnrollment set of endpoints, however the post method implemented right now directly creates the object without any other considerations of enrollment elsewhere. would this be better as a detail route for that viewset?

douglashall:
Yes, I think it makes sense to override the `create` method of the [EnterpriseCourseEnrollmentViewSet](https://github.com/edx/edx-enterprise/blob/master/enterprise/api/v1/views.py#L171) in order to implement this endpoint. Do we currently make POST requests to `enterprise/api/v1/enterprise_course_enrollments` from anywhere?

brittneyexline:
We do make calls from the lms... https://github.com/edx/edx-platform/blob/master/common/djangoapps/enrollment/views.py#L603

douglashall:
Hmm, okay. What if we made this a detail route on `EnterpriseCustomer`? The API endpoint path would look like `enterprise/api/v1/enterprise_customer/{ec_uuid}/enrollments`.

brittneyexline:
@douglashall sounds good to me. I've tested the lms_user_id and user_email cases on devstack, but I'd like to deploy this to the sandbox to test the tpa case since setting up SSO is kind of a pain. I will include the detail route note in the ticket for implementation.

douglashall:
Is there any reason we wouldn't start implementing now?

brittneyexline:
No reason - but might be easier if we plan to give part of this work to someone else. I'd at least like to write out a ticket with what I plan to do just to have it somewhere and to give some transparency to others, then we can pull it into the sprint.

douglashall:
You could take the approach that @UmanShahzad has taken with the [Degreed integration](https://github.com/edx/edx-enterprise/pull/232), spelling out the changes in the PR description.

enterprise/models.py
```diff
@@ -489,10 +579,25 @@ def enroll(self, course_run_id, mode):
...
+                    course_run_id=course_run_id,
+                    current_mode=course_enrollment.get('mode'),
+                    given_mode=mode,
+                )
+            )
+        else:
+            # The user was otherwise already enrolled.
+            raise UserAlreadyEnrolledInCourse('The user is already enrolled in the course {course_run_id}'.format(
+                course_run_id=course_run_id
+            ))
```
douglashall: Do we need to return an error for these conditions where the learner is already enrolled? I'm not sure what the caller of the API endpoint would gain from receiving this information.

enterprise/api/v1/serializers.py
```diff
@@ -410,3 +411,147 @@ class Meta:
...
+                course_mode,
+                course_run_id
+            )
+
+        if email_students:
+            enterprise_customer.notify_enrolled_learners(
+                self.context.get('request_user'),
+                course_run_id,
+                [self.enterprise_customer_user]
+            )
```
douglashall: Do we want to notify if this was just an upgrade?

enterprise/api/v1/views.py
```diff
@@ -167,6 +169,28 @@ def courses(self, request, pk=None):  # pylint: disable=invalid-name,unused-argu
...
+        Creates a course enrollment for an EnterpriseCustomerUser.
+        """
+        enterprise_customer = self.get_object()
+        serializer = serializers.EnterpriseCustomerCourseEnrollmentsSerializer(
+            data=request.data,
+            context={
+                'enterprise_customer': enterprise_customer,
+                'request_user': request.user,
+            }
+        )
```
douglashall: It would be nice if this endpoint could take a list of course enrollments to create/update instead of just one. It [looks like](http://www.django-rest-framework.org/api-guide/serializers/#dealing-with-multiple-objects) you can do this by passing a `many=True` keyword argument to the `EnterpriseCustomerCourseEnrollmentsSerializer` constructor.

enterprise/api/v1/serializers.py
brittneyexline: I don't have a strong opinion, but I remember our conversation around Guild being that we would notify learners when they were upgraded. In any case, they still have the option to not notify learners if they know ahead of time they are performing an upgrade.

enterprise/api/v1/views.py
brittneyexline: I'll look into it.

enterprise/api/v1/views.py
```diff
@@ -167,6 +169,28 @@ def courses(self, request, pk=None):  # pylint: disable=invalid-name,unused-argu
...
+        if serializer.is_valid():
+            try:
+                serializer.save()
+            except UserAlreadyEnrolledInCourse as e:  # pylint: disable=invalid-name
```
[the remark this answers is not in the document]
brittneyexline: I'll look into it.

enterprise/models.py
brittneyexline: I guess I'm erring on the side of being explicit, just to say that no action was performed because they were already enrolled, or the other case which was they were attempting to "downgrade" the enrollment. I could still catch this exception at the top level view and return success, but I like having the exception because it will short circuit the rest of the logic that doesn't need to be performed.

douglashall: Okay, just seems like a lot of extra code and logic paths for no real value.

brittneyexline: I see where you're coming from, and I don't disagree. I think the only case that is really an error according to business logic is the downgrade case - it seems like the caller should know why the enrollment is in a different mode than what they specified in their call, or that they shouldn't expect switching from verified to audit to work if they tried purposefully. Or maybe it's something we should allow?

douglashall: We should definitely prevent the downgrade and I agree that we should return an error status and message as the response to a request that attempts to do so.

enterprise/api/v1/views.py
brittneyexline: Unfortunately this is not as simple as adding that param. I've been trying to get it right for the entire morning but want to ask how important it is to support bulk before I go further into the rabbit hole :-)

douglashall: Maybe we can take a look at it together tomorrow morning and you can show me what is going on.

brittneyexline:
@douglashall i think this is ready for a final review, though i'm working out a few kinks in the tests...

enterprise/models.py
```diff
@@ -478,9 +563,14 @@ def enroll(self, course_run_id, mode):
...
-        enrolled_in_course = enrollment_api_client.is_enrolled(self.username, course_run_id)
-        if not enrolled_in_course:
-            # Directly enroll into the audit track.
+        course_enrollment = enrollment_api_client.get_course_enrollment(self.username, course_run_id) or {}
+        enrolled_in_course = course_enrollment and course_enrollment.get('is_active', False)
+
+        audit_modes = getattr(settings, 'ENTERPRISE_COURSE_ENROLLMENT_AUDIT_MODES', ['audit', 'honor'])
+        paid_modes = ['verified', 'professional']
+        is_upgrading = mode in paid_modes and course_enrollment.get('mode') in audit_modes
+        if not enrolled_in_course or is_upgrading:
```
douglashall: I think we probably want to modify the `edx.bi.user.enterprise.enrollment.course` tracking event below to identify mode changes, i.e. old mode vs. new mode.

PR #244, opened 2017-11-13
PLAT-1199 Replace PyCrypto with cryptography

jmbowman:
**Description:** Switch from `PyCrypto` (which hasn't been maintained in years) to the `cryptography` package, which provides equivalent functionality and has been used in edx-platform since at least Ficus.

**JIRA:** [PLAT-1199](https://openedx.atlassian.net/browse/PLAT-1199)

**Dependencies:** None

**Merge deadline:** None

**Testing instructions:**

1. A newly generated `EnterpriseCustomerReportingConfiguration` should still get an encrypted random password
2. The correct decrypted password should still be shown in Django admin

**Merge checklist:**

- [x] All reviewers approved
- [ ] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:**

* I'm a little fuzzy on the exact testing instructions, but the features above are the only ones where the change should affect this package.  There are unit tests covering this.
* The requirements file changes look significant because `cryptography` has a few dependencies, but we already install these in edx-platform anyway (and don't pin most of them yet, I'll be fixing that soon).
* This isn't quite a standard operation recommended by `cryptography` (hence the "hazmat" imports), but it has to work this way to keep compatibility with already saved encrypted passwords.
* `cryptography` [recommends using `os.urandom`](https://cryptography.io/en/latest/random-numbers/) on Python versions below 3.6, hence the switch away from a custom random number generator in this PR.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -13,3 +13,4 @@ Matt Drayer <mattdrayer@edx.org>
 Lucas Tadeu Teixeira <lucas@opencraft.com>
 Bill Filler <bfiller@edx.org>
 Matt Tuchfarber <mtuchfarber@edx.org>
+Jeremy Bowman <jbowman@edx.org>
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.53.15] - 2017-11-16
+----------------------
+
+* Use the cryptography package instead of the unmaintained pycrypto.
+
 [0.53.14] - 2017-11-14
 ----------------------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.53.14"
+__version__ = "0.53.15"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -2,21 +2,24 @@
 """
 Utility functions for enterprise app.
 """
-from __future__ import absolute_import, unicode_literals
+from __future__ import absolute_import, division, unicode_literals
 
 import datetime
 import hashlib
 import logging
+import os
 import re
 from uuid import UUID
 
 import analytics
 import pytz
-from Crypto import Random
-from Crypto.Cipher import AES
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.ciphers import Cipher
+from cryptography.hazmat.primitives.ciphers.algorithms import AES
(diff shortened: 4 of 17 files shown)


brittneyexline:
fyi @douglashall 
We'll also need to make a change here (https://github.com/edx/edx-enterprise-data/blob/master/enterprise_reporting/utils.py#L106) where the password gets used in the job for sending the reports.
I'm not sure what's up with the test coverage... I feel like this has happened before and I don't remember what we did about it.

jmbowman:
I don't see edx-enterprise-data among the edx-platform dependencies; is that something that's typically installed in its own virtualenv?  If so, the timing on upgrading it is pretty flexible; not hard to tackle it now, but it wouldn't block upgrading edx-enterprise and/or edx-platform.

I half wonder if we have coverage.py misconfigured, and it's mixing coverage data for `enterprise/utils.py` and `integrated_channels/sap_success_factors/utils.py`.  The lines marked in the latter as being not covered make no sense; they're often directly between lines which were covered, with no conditionals involved.

jmbowman:
Any feedback on this?  Should I rebase and rerun tests to see if the coverage fixes itself?

brittneyexline:
yea, i would rebase and try again. edx-enterprise-data is not installed here - it gets run separately from edxapp. as long as the algorithm is the same and it is able to decode successfully, then we can update it whenever as you say.

douglashall:
I would ignore the test coverage failure.

douglashall:
@jmbowman @brittneyexline It looks like this change is preventing the decrypted password from being displayed in Django admin. Did you test this?

douglashall:
FYI @georgebabey ^^

jmbowman:
@douglashall I wasn't sure how to test the UI directly (I've never used the enterprise functionality before), so I was hoping that a combination of the unit tests and code review would cover that.  If someone can give me steps for testing the functionality, I'll debug it and create a PR (including a new regression test for this).

douglashall:
Thanks @jmbowman. You can test by visiting the Django admin form for the `EnterpriseCustomerReportingConfiguration` model. On stage, that's [here](https://courses.stage.edx.org/admin/enterprise/enterprisecustomerreportingconfiguration/1/). I just created that `EnterpriseCustomerReportingConfiguration` on stage and it appears to be displaying the password as expected. You may need to revert this code, create a `EnterpriseCustomerReportingConfiguration`, and then roll forward in order to reproduce this issue.

douglashall:
@jmbowman I re-saved one of the `EntepriseCustomerReportingConfiguration` we have in production and the problem resolved itself. I'm not sure you need to take any further action.

jmbowman:
@douglashall Ok, it looks like that's because the random password and initialization vector are re-generated on each save.  The new encryption/decryption routine is internally consistent, but apparently a little out of sync with the old one.  If that's ok (i.e., we aren't counting on users actually depending on the password staying constant over time), then great.  Let me know if you do need any further help.

PR #245, opened 2017-11-14
ENT-740: Add identity_provider to enterprise customer API  response.

saleem-latif:
**Description:** This PR is related to https://github.com/edx/edx-platform/pull/16556 , We needed to access `identity_provider` from enterprise customer API response. This PR makes sure to provide `identity_provider` in the API response.

**JIRA:** [ENT-740](https://openedx.atlassian.net/browse/ENT-740)

**Installation instructions:** Simply install edx-enterprise using this branch
**Testing instructions:**

1. Open page enterprise customer details page (/enterprise/api/v1/enterprise-customer/<enterprise-customer-uuid>)
2. Make sure identity_provider field is present with appropriate value.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/tests/test_enterprise/api/test_views.py
+++ b/tests/test_enterprise/api/test_views.py
@@ -597,7 +597,7 @@ class TestEnterpriseAPIViews(APITest):
                 'catalog': 1, 'active': True, 'enable_data_sharing_consent': True,
                 'enforce_data_sharing_consent': 'at_enrollment',
                 'branding_configuration': None, 'enterprise_customer_entitlements': [],
-                'enable_audit_enrollment': False,
+                'enable_audit_enrollment': False, 'identity_provider': None,
                 'site': {
                     'domain': 'example.com', 'name': 'example.com'
                 },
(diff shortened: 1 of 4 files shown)


tests/test_enterprise/api/test_views.py
```diff
@@ -624,7 +624,7 @@ def test_enterprise_customer_courses(
                     'catalog': 1, 'active': True, 'enable_data_sharing_consent': True,
                     'enforce_data_sharing_consent': 'at_enrollment',
                     'branding_configuration': None, 'enterprise_customer_entitlements': [],
-                    'enable_audit_enrollment': False,
+                    'enable_audit_enrollment': False, 'identity_provider': None,
```
douglashall: We should also have a test to verify that the correct IdP provider_id is returned when the `EnterpriseCustomer` has an associated IdP.

saleem-latif: Done

zubair-arbi:
@douglashall Can we merge this PR? this way we would be able to get identity provider against a learner without separately querying db.

douglashall:
I'm fine with merging this, but is it actually going to get you what you need? This change adds only the IdP configuration ID to the API response data. Wouldn't you have to use that ID to query the database anyway?

zubair-arbi:
@douglashall you are right that this change will just provide id of related IdP but this way we will have to just query LMS db models instead of querying on enterprise models too.

asadiqbal08:
@saleem-latif Changes looked good to me 👍. please fix the travis issue

PR #246, opened 2017-11-17
[ENT-744] Track enterprise course enrollment events.

mtyaka:
**Description:** This patch adds `track_event` calls to all call sites where `EnterpriseCourseEnrollment` instances get created.

**JIRA:** [ENT-744](https://openedx.atlassian.net/browse/ENT-744)

**Dependencies:** None

**Testing instructions:**

1. Install this patch & run the migration.
2. Create some enterprise course enrollments (direct from admin, pending from admin, course landing page...).
4. Verify tracking events are emitted whenever course enrollment objects get created. One way to do that is to temporarily patch the `enterprise.utils.track_event` method to log its calls.

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:** 

There are many different flows and it is hard to capture all of them without threading a lot of parameters through multiple systems edx-enterprise, edx-platform, ecommerce. For example an enrollment may start on the course landing page, go via ecommerce and finally be created on the data sharing consent page.

### DIFF
--- a/enterprise/management/commands/create_enterprise_course_enrollments.py
+++ b/enterprise/management/commands/create_enterprise_course_enrollments.py
@@ -118,7 +118,6 @@ class Command(BaseCommand):
                     enterprise_customer_user=enterprise_learner,
                     course_id=course_id,
                 )
-
                 if created:
                     ent_course_enrollments_count += 1
 
--- a/enterprise/signals.py
+++ b/enterprise/signals.py
@@ -8,6 +8,7 @@ from logging import getLogger
 
 from enterprise.decorators import disable_for_loaddata
 from enterprise.models import EnterpriseCourseEnrollment, EnterpriseCustomerUser, PendingEnterpriseCustomerUser
+from enterprise.utils import track_enrollment
 
 logger = getLogger(__name__)  # pylint: disable=invalid-name
 
@@ -55,8 +56,9 @@ def handle_user_post_save(sender, **kwargs):  # pylint: disable=unused-argument
         # actually exist in the system; in such a case, the enrollment for each such
         # course is finalized when the user registers with the OpenEdX platform.
         enrollment.complete_enrollment()
-        EnterpriseCourseEnrollment.objects.get_or_create(
+        EnterpriseCourseEnrollment.objects.create(
             enterprise_customer_user=enterprise_customer_user,
-            course_id=enrollment.course_id
+            course_id=enrollment.course_id,
         )
+        track_enrollment('pending-admin-enrollment', user_instance.id, enrollment.course_id)
     pending_ecu.delete()
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.55.1] - 2017-11-30
+---------------------
+
+* Track enterprise course enrollment events.
+
 [0.55.0] - 2017-11-29
 ---------------------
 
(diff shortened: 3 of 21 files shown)


edx-webhook:
Thanks for the pull request, @mtyaka! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fedx-enterprise&number=246"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

enterprise/management/commands/create_enterprise_course_enrollments.py
```diff
@@ -118,8 +119,8 @@ def create_enterprise_course_enrollments(self, course_id, enterprise_learners):
                     enterprise_customer_user=enterprise_learner,
                     course_id=course_id,
                 )
-
                 if created:
                     ent_course_enrollments_count += 1
+                    track_enrollment('management-command-enrollment', enterprise_learner.user_id, course_id)
```
UmanShahzad: For some reason this one doesn't show up, and I'm not entirely sure why.

UmanShahzad: Actually since this is a management command to handle missing enrollments, I'm wondering if we even need this? cc @douglashall @brittneyexline

UmanShahzad:
@douglashall can you check this one out or will we have to wait for @brittneyexline ?

enterprise/management/commands/create_enterprise_course_enrollments.py
douglashall: @UmanShahzad Agreed, I don't think we should fire an event here.

enterprise/signals.py
```diff
@@ -55,8 +56,9 @@ def handle_user_post_save(sender, **kwargs):  # pylint: disable=unused-argument
         # actually exist in the system; in such a case, the enrollment for each such
         # course is finalized when the user registers with the OpenEdX platform.
         enrollment.complete_enrollment()
-        EnterpriseCourseEnrollment.objects.get_or_create(
+        EnterpriseCourseEnrollment.objects.create(
```
douglashall: Why change this? I think we should revert.

mtyaka: @douglashall I changed it because the `EnterpriseCustomerUser` associated with this course enrollment is only created [several lines above](https://github.com/edx/edx-enterprise/blob/7ad0b8bced8c033805d678e1c4413e0819236c3f/enterprise/signals.py#L50-L53), so an `EnterpriseCourseEnrollment` linked to the new `EnterpriseCustomerUser` instance cannot possibly exist at this stage in the code.

mtyaka:
Rebased from 7ad0b8bced8c033805d678e1c4413e0819236c3f

enterprise/management/commands/create_enterprise_course_enrollments.py
mtyaka: Thanks, removed.

mtyaka:
Thanks for the review @douglashall! I addressed your comments. Can you please check again?

mtyaka:
Rebased/squashed from 77c966350d8e2767723c617c13f8d9aaf35bea2d.

mtyaka:
Version bump PR at https://github.com/edx/edx-platform/pull/16716.

PR #247, opened 2017-11-23
Added signal and added enterprise related information in SailThru Variables

asadiqbal08:
@douglashall can you give some initial review on that changes ? 

RELATES TO [PR#16665 ](https://github.com/edx/edx-platform/pull/16665)

**Description:** 
- Introduced a signal what will be triggered when an enterprise learner is created into the model. 
- The corresponding receiver is defined in [PR#16665 ](https://github.com/edx/edx-platform/pull/16665) that will update the `is_enterprise_leaner` variable SailThru Var for that user. 

**JIRA:** [ENT-751](https://openedx.atlassian.net/browse/ENT-751)

### DIFF
--- a/enterprise/tpa_pipeline.py
+++ b/enterprise/tpa_pipeline.py
@@ -4,6 +4,7 @@ Module provides elements to be used in third-party auth pipeline.
 from __future__ import absolute_import, unicode_literals
 
 from enterprise.models import EnterpriseCustomer, EnterpriseCustomerUser
+from enterprise.utils import REGISTER_ENTERPRISE_USER
 
 try:
     from social_core.pipeline.partial import partial
@@ -62,7 +63,10 @@ def handle_enterprise_logistration(backend, user, **kwargs):
         return
 
     # proceed with the creation of a link between the user and the enterprise customer, then exit.
-    EnterpriseCustomerUser.objects.update_or_create(
+    __, created = EnterpriseCustomerUser.objects.update_or_create(
         enterprise_customer=enterprise_customer,
         user_id=user.id
     )
+    if created:
+        # Announce the registration
+        REGISTER_ENTERPRISE_USER.send(sender=None, user=user)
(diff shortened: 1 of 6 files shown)


enterprise/utils.py
```diff
@@ -734,3 +738,21 @@ def decrypt_string(string, iv):   # pylint: disable=invalid-name
...
+    :param user:
+    :return: enterprise_customer_user instance and created state
+    """
+    enterprise_customer_user, created = EnterpriseCustomerUser.objects.get_or_create(
+        enterprise_customer=enterprise_customer,
+        user_id=user.id
+    )
+    if created:
+        # Announce registration
+        REGISTER_ENTERPRISE_USER.send(sender=None, user=user)
```
asadiqbal08: Another approach is to use the `post_save` signal/receiver instead of using the custom one as doing here.

enterprise/tpa_pipeline.py
```diff
@@ -62,7 +63,10 @@ def handle_enterprise_logistration(backend, user, **kwargs):
...
 
     # proceed with the creation of a link between the user and the enterprise customer, then exit.
-    EnterpriseCustomerUser.objects.update_or_create(
+    __, created = EnterpriseCustomerUser.objects.update_or_create(
         enterprise_customer=enterprise_customer,
         user_id=user.id
     )
+    if created:
+        # Announce the registration
+        REGISTER_ENTERPRISE_USER.send(sender=None, user=user)
```
asadiqbal08: Another approach is to use the post_save signal/receiver instead of using the custom one as doing here.

asadiqbal08:
@douglashall Can you share your thoughts on that two PR's ?

enterprise/tpa_pipeline.py
douglashall: Yes, no need for the custom signal.

PR #249, opened 2017-11-27
Do not change EnterpriseCustomerReportingConfiguration.password on update.

douglashall:
We do not want to generate a new password every time someone saves
an EnterpriseCustomerReportingConfiguration model.

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -1240,12 +1240,13 @@ class EnterpriseCustomerReportingConfiguration(TimeStampedModel):
         """
         Override of model save method to dynamically generate the password field and perform additional validation.
         """
-        self.initialization_vector = utils.generate_aes_initialization_vector()
-        self.password = utils.encrypt_string(
-            get_random_string(length=32),
-            self.initialization_vector
-        )
-        self.full_clean()
+        if not self.pk:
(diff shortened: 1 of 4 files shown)


enterprise/models.py
```diff
@@ -1241,11 +1241,12 @@ def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
...
-            get_random_string(length=32),
-            self.initialization_vector
-        )
-        self.full_clean()
+        if not self.pk:
+            self.password = utils.encrypt_string(
+                get_random_string(length=32),
+                self.initialization_vector
+            )
+            self.full_clean()
```
georgebabey: Do you not need full_clean() for validation on updates?

douglashall: `full_clean` is a wrapper function that calls a set of other validation functions that are called by the Model admin form. I think we are calling `full_clean` here because we added the password and want to validate the model again. So, no I don't think we want to call it if we aren't changing anything inside the `save` function here.

georgebabey: I think this will also create a new initialization_vector on update, but not password with that IV, so those would now be out of sync.

georgebabey: I think you're changing the iv though, probably should all go inside the if

douglashall: Good catch.

PR #251, opened 2017-11-30
Change `url` to `Url`, which is what Degreed expects.

UmanShahzad:
Degreed expects `Url` rather than `url`. Yeah.


edx-webhook:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=251&repo=edx%2Fedx-enterprise"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

UmanShahzad:
@douglashall Actually this doesn't seem to be working either. The docs say "Url" in the table, but "url" in the example. The enrollment URLs themselves work, so I'm thinking this is a Degreed-side issue, and we should clarify the `url` vs. `Url` thing.

UmanShahzad:
@douglashall In case Branden doesn't get back, one idea is to modify the description we send back to prepend the enrollment URL as a clickable "Enroll now." text or something.

UmanShahzad:
Ah :)

PR #254, opened 2017-12-04
ENT-767 Check embargo restrictions when rendering enterprise landing pages.

douglashall:
Redirect the user to the embargo restriction page if their access
to the given course run/program is restricted.

**JIRA:** https://openedx.atlassian.net/browse/ENT-767

**Testing Instructions:**
1. Visit https://business.sandbox.edx.org/enterprise/2b5a2956-7746-4fc9-9587-bc887ba45973/course/course-v1:edX+DemoX+Demo_Course/enroll/?catalog=cff09318-1358-4703-a277-bc2596480ed5&utm_medium=enterprise&utm_source=pied-piper.
2. Register a new account. Choose "Canada" as your country.
3. Verify that the Embargo restriction message page is displayed.

screen shot 2017-12-05 at 12 43 27 pm

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -11,6 +11,7 @@ from uuid import UUID
 from consent.helpers import get_data_sharing_consent
 from consent.models import DataSharingConsent
 from dateutil.parser import parse
+from ipware.ip import get_ip
 
 from django.conf import settings
 from django.contrib.auth.decorators import login_required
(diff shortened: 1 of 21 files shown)


enterprise/views.py
```diff
@@ -1178,6 +1184,15 @@ def get(self, request, enterprise_uuid, program_uuid):
             # The user is already enrolled in the program, so redirect to the program's dashboard.
             return redirect(LMS_PROGRAMS_DASHBOARD_URL.format(uuid=program_uuid))
 
+        # Check to see if access to any of the course runs in the program are restricted for this user.
+        course_run_ids = []
+        for course in program_details['courses']:
+            for course_run in course['course_runs']:
+                course_run_ids.append(course_run['key'])
```
georgebabey: Any edX recommended guidance on for loops versus list comprehensions?
ie. I think this something like this would work as well but might be less readable: `course_run_ids = [course_run['key'] for course in program_details['courses'] for course_run in course['course_runs']]`

douglashall: Yeah, I would op for readability in this case.

PR #255, opened 2017-12-05
Added the integrated channel configuration to data task logging.

douglashall:


douglashall:
@mattdrayer Indeed it does.

PR #259, opened 2017-12-12
ENT-738 Added publish_audit_enrollment_url flag

georgebabey:
If publish_audit_enrollment_url flag is True on EnterpriseCustomerCatalog
courses will be published with the direct-to-audit querystring url (?audit=true)

https://openedx.atlassian.net/browse/ENT-738

**Testing instructions:**

1. In django-admin, update an EnterpriseCustomerCatalog to set the publish_audit_enrollment_url flag
2. Run the transmit_course_metadata admin job
3. From the LMS set up to use the catalog from step 1, verify that when a user clicks through a course they are taken directly to the course (based on direct-to-audit work done in https://openedx.atlassian.net/browse/ENT-738)

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/test_utils/__init__.py
+++ b/test_utils/__init__.py
@@ -167,7 +167,7 @@ def assert_url(first, second):
 
 def assert_url_contains_query_parameters(url, query_params):
     """
-    Asserts that a url string contains the given query parameters
+    Assert that a url string contains the given query parameters.
 
     Args:
         url: Full url string to check
--- a/test_utils/__init__.py
+++ b/test_utils/__init__.py
@@ -19,8 +19,7 @@ import mock
 import six
 from pytest import mark
 from rest_framework.test import APITestCase, APIClient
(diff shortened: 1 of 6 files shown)


test_utils/__init__.py
```diff
@@ -160,6 +159,17 @@ def assert_url(first, second):
 
     assert first == second
 
+def assert_url_contains_query_parameters(url, query_params):
+    """
+    Asserts that a url string contains the given query parameters
+    :param url: Full url string to check
+    :param query_params: Dict of query string key/value pairs
+    :raises: Assertion error if the params do not exist in the given url
```
douglashall: We try to use [Google style docstrings](https://google.github.io/styleguide/pyguide.html#Comments).

PR #261, opened 2017-12-13
ENT-809 Add swagger spec for course_enrollments API endpoint.

douglashall:
This adds the swagger specification for the course_enrollments API
endpoint so that this endpoint can be made accessible via the
API gateway.

The only feasible way to test this is to get it merged, update the api-manager swagger doc to point to the new git hash, and then deploy api-manager to stage. I'm going to merge this when the tests pass.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.56.2] - 2017-12-13
+---------------------
+
+* Add course_enrollments API endpoint to swagger specification.
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.rst
```diff
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.56.2 - 2017-12-13
```
brittneyexline: missing a square bracket here

douglashall:
@brittneyexline Good catch. Thanks.

PR #263, opened 2017-12-18
ENT-775 Add sftp configuration support for EnterpriseCustomerReportingConfig

brittneyexline:
**Description:** This PR adds the ability to configure the EnterpriseCustomerReportingConfig for SFTP. It finally utilizes the delivery method field, which now has sftp in addition to email. It also adds 5 new fields for sftp: hostname, port, username, password, and file path. These fields are required when sftp is selected as the delivery method. The sftp password field is stored encrypted in the database, just as the existing password field is.

Additionally, this PR now enables editing of the existing password field as well as the sftp password. After migrations are run, the existing decrypted password display will be replaced by a text box that displays the decrypted password, and making changes to the password in the text box will be updated on save just as any other field.

Finally, the new fields have been added to the serializer so that they get returned in the api endpoint for the object.

**JIRA:** https://openedx.atlassian.net/browse/ENT-775

**Dependencies:** https://github.com/edx/edx-enterprise-data/pull/15 depends on this. 

**Merge deadline:** 

**Installation instructions:** After installing, be sure to run migrations!

**Testing instructions:**

Verify that existing configurations have their data in tact. Please note that on the business sandbox, something about the passwords for earlier configurations got corrupted due to either the secret or initialization vectors changing, so they may display as blank. They can be edited and new changes will work as expected, and we shouldn't run into this issue in production.

Verify that if you select sftp as the delivery method, you must enter the other sftp fields, excluding password (blank passwords are technically valid, plus this one was more complicated with the encryption logic) and port (has a default value of 22, so it is always set). Verify that the existing validation has remained the same.

Verify that if you hit the api endpoint for reporting configs (https://business.sandbox.edx.org/enterprise/api/v1/enterprise_customer_reporting) that the new sftp fields come through.

Verify that both the password and sftp_password field display in plain text in the configuration but are saved encrypted in the database.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -1264,12 +1264,11 @@ class EnterpriseCustomerReportingConfiguration(TimeStampedModel):
     )
 
     DELIVERY_METHOD_EMAIL = 'email'
-    # We are adding this field preemptively as we plan to support FTP, but don't have the details yet.
-    DELIVERY_METHOD_FTP = 'ftp'
+    DELIVERY_METHOD_SFTP = 'sftp'
 
     DELIVERY_METHOD_CHOICES = (
         (DELIVERY_METHOD_EMAIL, DELIVERY_METHOD_EMAIL),
-        (DELIVERY_METHOD_FTP, DELIVERY_METHOD_FTP),
+        (DELIVERY_METHOD_SFTP, DELIVERY_METHOD_SFTP),
     )
 
     DAYS_OF_WEEK = (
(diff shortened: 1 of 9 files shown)


enterprise/models.py
```diff
@@ -1360,15 +1393,28 @@ def __repr__(self):
...
             )
-            self.full_clean()
+
+        if decrypted_sftp_password:
+            self.sftp_password = utils.encrypt_string(
+                decrypted_sftp_password,
+                self.initialization_vector
+            )
+
+        self.full_clean()
```
georgebabey: nit: We will now call self.full_clean() on each save even if we don't manually update the password or sftp password. Probably not a problem but slight overhead to recheck validation. Could wrap in a `if decrypted_password or decrypted_sftp_password`

PR #268, opened 2017-12-21
ENT-763 Show pricing in local currency

georgebabey:
When the waffle switch for localized currency is enabled, this will update the display price on the enterprise landing page so it displays in the users local currency based on their ip address

https://openedx.atlassian.net/browse/ENT-763

**Testing instructions:**

1. Using a new session (log out of any existing lms session) utilize a VPN to access a course through one of the LMS channels (ie. SuccessFactors) from another country (ie. Canada)
2. Verify the landing page has the price updated in the users local currency.
3. Turn off waffle switch and verify the currency is in USD
4. Turn on waffle switch and disable the VPN, verify the currency shows in USD.

Requires https://github.com/edx/edx-platform/pull/17153

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -5,10 +5,10 @@ User-facing views for the Enterprise app.
 from __future__ import absolute_import, unicode_literals
 
 import re
-import waffle
 from logging import getLogger
 from uuid import UUID
 
+import waffle
 from consent.helpers import get_data_sharing_consent
 from consent.models import DataSharingConsent
 from dateutil.parser import parse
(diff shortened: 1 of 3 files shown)


georgebabey:
Will require https://github.com/edx/edx-platform/pull/17001

douglashall:
Wouldn't it be simpler to do this all client-side like we are doing in [ecommerce](https://github.com/edx/ecommerce/blob/d01a81391c5fd1c510efb1ef49d1971c33e8c1d3/ecommerce/static/js/pages/basket_page.js#L259)?

enterprise/views.py
```diff
@@ -98,6 +106,34 @@ def get_global_context(request):
         'header_logo_alt_text': _('{platform_name} home page').format(platform_name=platform_name),
     }
 
+def get_localized_price_text(price, request):
```
douglashall: Seems like this would be better in edx-platform or some other python package where it could be reused.

georgebabey: Agreed - probably with edx-platform since that is where the middleware that sets country_code lives.

georgebabey:
@douglashall  This has been reworked, can you check it again

enterprise/views.py
```diff
@@ -99,6 +105,18 @@ def get_global_context(request):
...
 
 
+def get_price_text(price, request):
+    """
+    Returns the localized converted price as string (ex. '$150 USD')
+
+    If the local_currency switch is enabled and the users location has been determined this will convert the
+    given price based on conversion rate from the Catalog service and return a localized string
+    """
+    if waffle.switch_is_active('local_currency') and get_localized_price_text:
```
douglashall: You might need test coverage here.

PR #270, opened 2018-01-03
[ENT-431]: SuccessFactors Admin Update: Enterprise Customer Configuration

saleem-latif:
**Description:** 
The Django Admin for SuccessFactors Enterprise Customer Configurations is confusing:

**List View**
1. The Client Secret values should not be displayed on the List view.
2. We can just display checkboxes instead of values for Active and Access Token
3. Fields to display should be Name, Active, Base URL, Access Token, Date Modified, in that order. This will provide a quick visual reference and whether the configuration can generate a valid access token.
4. Please order the configuration records by name ascending, and not by ID descending.

**Detail View:**
1. There is an Enterprise Customer dropdown field, and an "Enterprise customer name" field. Unsure why we have both, so we can probably remove the read-only name field.
2. The Enterprise Customer dropdown field should show the name of the customer, versus the object string
3. The order of the fields should be: Enterprise Customer, Active, Base URL, Company ID, Client ID, Client Secret, User Id, User Type, Has Access Token.
4. SAP Base URL and SAP User Type should be in regular format, not bold
5. All fields should have associated help text

**JIRA:** [ENT-431](https://openedx.atlassian.net/browse/ENT-431)

**Installation instructions:** Simply install edx-enterprise using `saleem-latif/ENT-431` branch.

**Testing instructions:**

1. Open success factors configuration admin page and verify list and details views are according to the description above.

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag


saleem-latif:
@zubair-arbi `edx-enterprise` package requirements come from `requirements/base.txt` (see [this](https://github.com/edx/edx-enterprise/blob/master/setup.py#L56) and [this](https://github.com/edx/edx-enterprise/blob/master/setup.py#L37)) I do not think requirements updates will effect edx-enterprise installation.

saleem-latif:
@georgebabey PR is ready for you review. Please take a look

integrated_channels/sap_success_factors/models.py
```diff
@@ -63,17 +64,27 @@ class SAPSuccessFactorsEnterpriseCustomerConfiguration(EnterpriseCustomerPluginC
         (USER_TYPE_ADMIN, 'Admin'),
     )
 
-    key = models.CharField(max_length=255, blank=True, verbose_name="Client ID")
-    sapsf_base_url = models.CharField(max_length=255, verbose_name="SAP Base URL")
-    sapsf_company_id = models.CharField(max_length=255, blank=True, verbose_name="SAP Company ID")
-    sapsf_user_id = models.CharField(max_length=255, blank=True, verbose_name="SAP User ID")
-    secret = models.CharField(max_length=255, blank=True, verbose_name="Client Secret")
+    key = models.CharField(
+        max_length=255, verbose_name="Client ID", help_text=_("OAuth client identifier.")
```
brittneyexline: nit: now that we have a longer list of arguments, we could put each one on one line for better readability.

PR #278, opened 2018-01-19
ENT-854 ENT-853 Improved handling of password fields.

douglashall:
This modifies the `EnterpriseCustomerReportingConfiguration` model in the following ways:

1. Adds two new fields, `decrypted_password` and `decrypted_sftp_password`, which will be used going forward to store the password used to lock the reporting zip file that is emailed to enterprises and the password used to make an SFTP connection for transmitting the reporting zip file respectively. These fields make use of the `EncryptedCharField` made available by the django-fernet-fields package.

2. Modifies the `EnterpriseCustomerReportingConfiguration` admin form to display the password fields as type="password" input elements. Also, the password fields will only be displayed on the create form and not on the update form.

**Related PRs:**
https://github.com/edx-ops/business-secure/pull/2
https://github.com/edx/jenkins-job-dsl-internal/pull/155
https://github.com/edx/edx-enterprise-data/pull/16


enterprise/migrations/0038_auto_20180122_1427.py
```diff
@@ -0,0 +1,27 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('enterprise', '0037_auto_20180110_0450'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='enterprisecustomerreportingconfiguration',
+            name='decrypted_password',
```
georgebabey: I worry a little naming these `decrypted` will confuse people when they are in fact encrypted at rest.  Although describing it as decrypted in the model will caution people from displaying it - is that the intent?

enterprise/migrations/0039_auto_20180122_0934.py
```diff
@@ -0,0 +1,39 @@
...
+        'EnterpriseCustomerReportingConfiguration'
+    )
+    for config in EnterpriseCustomerReportingConfiguration.objects.all():
+        if config.password:
+            config.decrypted_password = decrypt_string(
+                config.password,
+                config.initialization_vector
+            )
+        if config.sftp_password:
+            config.decrypted_sftp_password = decrypt_string(
```
georgebabey: can we remove the utils `encrypt_string` / `decrypt_string` once this is merged and the migration happens?

douglashall: Yes, that is my plan, along with the old password fields.

enterprise/migrations/0038_auto_20180122_1427.py
douglashall: Yeah, I went back and forth on this as well. I think this is the right name.

PR #279, opened 2018-01-26
ENT-823 Add course detail endpoint to EnterpriseCustomerCatalog API.

douglashall:
This will enable us to define EnterpriseCustomerCatalogs with a `"content_type": "course"` content filter and provide a way for direct integrators to pull the full details for each course in the catalog (including course run data).

I have a [WIP branch](https://github.com/edx/edx-enterprise/pull/293) which refactors the `integrated_channels` codebase to allow it to manage content metadata for more than just course runs (which is what it is limited to currently). I'd like to get these changes merged as I continue to work on that refactoring.

## Testing Instructions ##
1. Visit https://business.sandbox.edx.org/enterprise/api/v1/enterprise_catalogs/f8f13d0b-58a1-4708-b1be-6accc88acfb1/. This is the detail endpoint for an EnterpriseCustomerCatalog which uses the `"content_type": "course"` content filter.
2. Grab a course key for one of the courses and attempt to hit the courses detail endpoint. For example, https://business.sandbox.edx.org/enterprise/api/v1/enterprise_catalogs/f8f13d0b-58a1-4708-b1be-6accc88acfb1/courses/edX+DemoX/.

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -27,6 +27,7 @@ from enterprise.api.throttles import ServiceUserThrottle
 from enterprise.api.v1 import serializers
 from enterprise.api.v1.decorators import enterprise_customer_required, require_at_least_one_query_parameter
 from enterprise.api_client.discovery import CourseCatalogApiClient
+from enterprise.constants import COURSE_KEY_URL_PATTERN
 from six.moves.urllib.parse import quote_plus, unquote  # pylint: disable=import-error,ungrouped-imports
 
 LOGGER = getLogger(__name__)
@@ -119,8 +120,8 @@ class EnterpriseCustomerViewSet(EnterpriseReadWriteModelViewSet):
 
         contains_content_items = False
         for catalog in enterprise_customer.enterprise_customer_catalogs.all():
-            contains_course_runs = not course_run_ids or catalog.contains_content_items('key', course_run_ids)
-            contains_program_uuids = not program_uuids or catalog.contains_content_items('uuid', program_uuids)
+            contains_course_runs = not course_run_ids or catalog.contains_courses(course_run_ids)
+            contains_program_uuids = not program_uuids or catalog.contains_programs(program_uuids)
             if contains_course_runs and contains_program_uuids:
                 contains_content_items = True
                 break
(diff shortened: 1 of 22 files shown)


enterprise/api/v1/views.py
```diff
@@ -326,15 +327,33 @@ def contains_content_items(self, request, pk, course_run_ids, program_uuids):
...
+
+        The course needs to be included in the specified EnterpriseCustomerCatalog
+        in order for metadata to be returned from this endpoint.
+        """
+        enterprise_customer_catalog = self.get_object()
+        course = enterprise_customer_catalog.get_course(course_key)
+        if not course:
+            raise Http404
+
+        context = self.get_serializer_context()
```
brittneyexline: any reason not to override get_serializer_context for this view set to always include enterprise_customer_catalog instead of adding it for each detail_route?

enterprise/models.py
```diff
@@ -1055,38 +1082,65 @@ def get_paginated_content(self, query_parameters):
...
+            content_ids_in_catalog = self._filter_members('key', course_keys)
+
+        return set(content_ids).issubset(content_ids_in_catalog) or course_keys.issubset(content_ids_in_catalog)
+
+    def contains_programs(self, program_uuids):
+        """
+        Return true if this catalog contains the given programs.
+        """
+        content_ids_in_catalog = self.content_filter_ids
+        if not content_ids_in_catalog:
```
brittneyexline: does this mean that an empty catalog is the same as a catalog with all content when determining if some program/course is part of the catalog?

enterprise/utils.py
```diff
@@ -707,3 +694,66 @@ def is_course_run_enrollable(course_run):
...
            (not enrollment_end or enrollment_end > now)
+
+
+def is_course_run_upgradeable(course_run):
+    """
+    Return true if the course run has a verified seat with an unexpired upgrade deadline, false otherwise.
+    """
+    now = datetime.datetime.now(pytz.UTC)
+    for seat in course_run.get('seats', []):
+        if seat.get('type') == 'verified':
```
brittneyexline: should we also check for professional and no_id_professional?

enterprise/models.py
douglashall: No, this is checking to see if the content_filter was defined with any content IDs (i.e. program uuids or course/courserun keys). If it was, then we can just check that list of ids to see if the provided program UUIDs are included in that list. If not, then we we add the give program UUIDs to the content_filter and make the search request to the discovery service.

enterprise/utils.py
douglashall: Hmm, no, not here, but maybe somewhere else.

enterprise/api/v1/views.py
douglashall: Good idea.

douglashall: On second thought, that would require multiple calls to `self.get_object` resulting in multiple calls to the DB. I'm going to stick with this for now.

enterprise/utils.py
douglashall: `professional` and `no_id_professional` courses will be handled appropriately [here](https://github.com/edx/edx-enterprise/pull/279/files#diff-ce3a6d48e7f42961c07cd219f505d179R737). 
Basically, no "upgradeable" course runs will be found because professional/no_id_professional course runs are never "upgradeable" and we will default to including all course runs in the calculation to determine the current course run.

douglashall:
@brittneyexline @georgebabey Is this ready for merge after version bump/changelog entry?

PR #280, opened 2018-01-29
ENT-854 Remove old password fields.

douglashall:
This removes code references to the old password fields. There will be a follow-up PR for adding the migration to delete these fields from the database.


enterprise/api/v1/serializers.py
```diff
@@ -412,7 +412,7 @@ class Meta:
         model = models.EnterpriseCustomerReportingConfiguration
         fields = (
             'enterprise_customer', 'active', 'delivery_method', 'email', 'frequency', 'day_of_month', 'day_of_week',
-            'hour_of_day', 'initialization_vector', 'encrypted_password', 'sftp_hostname', 'sftp_port', 'sftp_username',
+            'hour_of_day', 'encrypted_password', 'sftp_hostname', 'sftp_port', 'sftp_username',
```
brittneyexline: i think you have to remove encrypted password as well, unless that is what the new password field will be populating going forward?

enterprise/models.py
```diff
@@ -1326,20 +1326,6 @@ class EnterpriseCustomerReportingConfiguration(TimeStampedModel):
...
     )
-    initialization_vector = models.BinaryField(
-        max_length=32,
-        blank=False,
-        null=False,
-        editable=False,
-        verbose_name=_("Password Encryption Initialization Vector"),
-        help_text=_("This is the initialization vector used to encrypt the password using AESencr")
-    )
-    password = models.BinaryField(
```
brittneyexline: do you need to remove the sftp_password field as well?

enterprise/api/v1/serializers.py
douglashall: Yes. The new field is called `encrypted_password` and it specified as such here in the serializer configuration.

douglashall:
@brittneyexline Good to go now?

PR #282, opened 2018-01-30
asadiqbal08/ENT-849 enterprise catalog API in XML format.

asadiqbal08:
**Description:** Provided option for enterprise to pull enterprise/catalog API in XML format (not just JSON)

**JIRA:** [ENT-849](https://openedx.atlassian.net/browse/ENT-849)

**Testing instructions:**

- Hit [enterprise catalog api](https://business.sandbox.edx.org/enterprise/api/v1/enterprise_catalogs/) on business sandbox.
- By default, JSON response would be returned. 
- To get an xml response, Add the suffix **.xml** at the end of URL --> e.g. https://business.sandbox.edx.org/enterprise/api/v1/enterprise_catalogs.xml . An xml response would be returned

### DIFF
--- a/enterprise/api/renderers.py
+++ b/enterprise/api/renderers.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+"""
+Renderer(s) for api response.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from rest_framework.negotiation import DefaultContentNegotiation
+from rest_framework.settings import api_settings
+
+
+class IgnoreClientContentNegotiation(DefaultContentNegotiation):
+    """
+    Class for client content negotiation.
+    """
+
+    def select_renderer(self, request, renderers, format_suffix):  # pylint: disable=signature-differs
+        # Allow URL style format override.  eg. "?format=json or http://example.com/v1/organizations.xml/
+        render_format = format_suffix or request.query_params.get(api_settings.FORMAT_SUFFIX_KWARG)
+        if render_format:
+            return DefaultContentNegotiation.select_renderer(self, request, renderers, render_format)
+
+        # Select the first renderer in the `.renderer_classes` list.
+        return (renderers[0], renderers[0].media_type)
--- a/tests/test_enterprise/api/test_views.py
+++ b/tests/test_enterprise/api/test_views.py
@@ -1998,3 +1998,14 @@ class TestEnterpriseAPIViews(APITest):
         response = self.load_json(response.content)
 
         self.assertListEqual(response, expected_response)
+
+    def test_enterprise_customer_catalogs_response_formats(self):
+        """
+        ``enterprise_catalogs``'s xml and json responses verification.
+        """
+        response_xml_1 = self.client.get('{}?format=xml'.format(ENTERPRISE_CATALOGS_LIST_ENDPOINT))
+        response_xml_2 = self.client.get('/enterprise/api/v1/enterprise_catalogs.xml')
+        response_json = self.client.get('{}?format=json'.format(ENTERPRISE_CATALOGS_LIST_ENDPOINT))
+        self.assertTrue(response_xml_1['content-type'] == 'application/xml; charset=utf-8')
+        self.assertTrue(response_xml_2['content-type'] == 'application/xml; charset=utf-8')
+        self.assertTrue(response_json['content-type'] == 'application/json')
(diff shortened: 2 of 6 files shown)


asadiqbal08:
@georgebabey can you please take a look today ?

asadiqbal08:
@georgebabey I am not sure about the docs in enterprise repo. If there is that need to be add then let me know, I will create another PR for that one. thanks

enterprise/api/renderers.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+Renderer(s) for api response.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from rest_framework.negotiation import DefaultContentNegotiation
+from rest_framework.settings import api_settings
+
+
+class IgnoreClientContentNegotiation(DefaultContentNegotiation):
```
douglashall: What does this class do? What does "IgnoreClient" mean? I think the docstrings could have been better here.

tests/test_enterprise/api/test_views.py
```diff
@@ -1998,3 +1998,14 @@ def test_enterprise_customer_course_enrollments_detail_multiple(
         response = self.load_json(response.content)
 
         self.assertListEqual(response, expected_response)
+
+    def test_enterprise_customer_catalogs_response_formats(self):
+        """
+        ``enterprise_catalogs``'s xml and json responses verification.
+        """
+        response_xml_1 = self.client.get('{}?format=xml'.format(ENTERPRISE_CATALOGS_LIST_ENDPOINT))
```
mattdrayer: Now that we are introducing multiple formats it might make sense to add a line here that requests the default representation and confirms that it is JSON.  Or maybe it's safe enough to assume that the system always returns JSON unless a special case is added, and so we need to only test the special case and not verify the default format has not changed?

enterprise/api/renderers.py
asadiqbal08: It was referenced in [django rest framework](http://www.django-rest-framework.org/api-guide/content-negotiation/#custom-content-negotiation).

tests/test_enterprise/api/test_views.py
asadiqbal08: By default, the content renderer is JSON. Existing tests were validating a json response in test repo. However we can add our test too here to show up the default JSON behavior.

enterprise/api/renderers.py
georgebabey: In that case they were skipping all client requested negotiation and just returning the first. This one introduced allows the format_suffix. Maybe this should be renamed to something like `AllowSuffixContentNegotiation` or more generic `CustomContentNegotiation` 

Good catch @douglashall

asadiqbal08: I will update the naming issue soon. thanks

douglashall: Why do we need to override here at all? Shouldn't the `DefaultContentNegotiation` class do what we need to do? What exactly is this override making possible that the `DefaultContentNegotiation` class does not? If we do need to override, shouldn't we also be overriding `select_parser`?

georgebabey: From reading the code, my understanding was that this would either handle the format_suffix passed in or pull it from the query params. Since most APIs aren't use to accessing `v1/organizations.xml/` but `?format=xml` is more common this seems like a good override (if my assumptions are correct)

Its still not clear to me if an `accepts` header is handled natively in DRF or if we would need to also sent render_format based on header. I would assume (hope) that DRF handles that negotiation out-of-box. @asadiqbal08  did you test that?

asadiqbal08: @georgebabey you are right with the above statement. Further I am sharing my findings that were examined during implementation. 

- DRF was ignoring the order in [render_classes](https://github.com/edx/edx-enterprise/pull/282/files#diff-4495418c366e5831b5e2daaea415aa56R305), may be due to djangorestframework-xml library. It is giving precedence to xml always. so that’s why I override the method. 

- `?format=json` was not working, DRF returns the xml response ignore the url format pattern but working fine when add the `.xml` and `.json` suffix in URL.

- I did not override the parser, actually, REST framework examine the Content-Type header on the incoming request, and determine which parser to use to parse the request content. Please verify the behavior on [business-sandbox](https://business.sandbox.edx.org/enterprise/api/v1/enterprise_catalogs/)  and let me know If there is something that I missed in PR. 

Many Thanks.

douglashall: We should make note of these things in the docstring. When I read the DRF code it seemed like all of this should already by handled without customization.

PR #286, opened 2018-02-05
Decrease SAPSF chunk size from 1000 to 500

mattdrayer:
Per SAP technical support, while the SuccessFactors Learning API does support payloads up to 1000 records per request, this is too much data for most tenants to handle in a single request.  SAPSF recommends sending no more than 500 records per request.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.65.3] - 2018-02-06
+---------------------
+
+* Decreased SuccessFactors course metadata chunk size from 1000 to 500, per SAP's recommendation.
+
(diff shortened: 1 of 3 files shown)


mattdrayer:
@georgebabey I included the version bump for this PR -- good to go?

georgebabey:
Looks good - :shipit:

douglashall:
Have we stopped using CHANGELOG.rst?

mattdrayer:
:man_shrugging: -- I can updated if we are still using it

georgebabey:
I like it - shows in PyPi. For a change this small most probably wouldn't care but then they'll be left wondering *what* changed

mattdrayer:
Went ahead and added a changelog entry

PR #287, opened 2018-02-06
ENT-849 post merge suggestions.

asadiqbal08:
Addressed the suggestion that received on PR -- https://github.com/edx/edx-enterprise/pull/282


enterprise/api/renderers.py
```diff
@@ -8,13 +8,18 @@
 from rest_framework.settings import api_settings
 
 
-class IgnoreClientContentNegotiation(DefaultContentNegotiation):
+class CustomContentNegotiation(DefaultContentNegotiation):
     """
-    Class for client content negotiation.
+    DRF was ignoring the order in render_classes, may be due to djangorestframework-xml library.
+    By Default, It was giving precedence to xml by default so we need to override the select_renderer method.1
```
zubair-arbi: nit: `By Default, It was giving precedence to xml by default so we need to override the select_renderer method.1` -> `By Default, It was giving precedence to XML so we need to override the select_renderer method`

enterprise/api/renderers.py
```diff
@@ -8,13 +8,18 @@
...
     """
 
     def select_renderer(self, request, renderers, format_suffix):  # pylint: disable=signature-differs
-        # Allow URL style format override.  eg. "?format=json or http://example.com/v1/organizations.xml/
+        """
+        Allow URL style format override. Either you can specify the querystring parameter (/?format=json
+        or /?format=xml) or XML/JSON suffix at end of URL e.g. shttp://example.com/v1/organizations.xml/
```
zubair-arbi: nit: `or XML/JSON suffix at end of URL e.g. shttp://example.com/v1/organizations.xml/` -> 
`or add XML/JSON suffix at end of URL, e.g. http://example.com/v1/organizations.xml`

georgebabey:
Thanks for updating the docstrings - watch the version bump with Matt's PR https://github.com/edx/edx-enterprise/pull/286

CHANGELOG.rst
```diff
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.65.3] - 2018-02-06
+---------------------
+
+* Refactor the code.
```
douglashall: We need a meaningful sentence here, otherwise we should just stop using this file all together.

enterprise/api/renderers.py
```diff
@@ -8,13 +8,18 @@
 from rest_framework.settings import api_settings
 
 
-class IgnoreClientContentNegotiation(DefaultContentNegotiation):
+class CustomContentNegotiation(DefaultContentNegotiation):
```
douglashall: Something that tries to describe what this custom content negotiation class does would be better.

enterprise/api/renderers.py
douglashall: Was it giving you XML even when you specified the appropriate headers `Accept: application/json`?

enterprise/api/renderers.py
```diff
@@ -8,13 +8,18 @@
...
+    By Default, It was giving precedence to xml by default so we need to override the select_renderer method.1
     """
 
     def select_renderer(self, request, renderers, format_suffix):  # pylint: disable=signature-differs
-        # Allow URL style format override.  eg. "?format=json or http://example.com/v1/organizations.xml/
+        """
+        Allow URL style format override. Either you can specify the querystring parameter (/?format=json
+        or /?format=xml) or XML/JSON suffix at end of URL e.g. shttp://example.com/v1/organizations.xml/
+        """
+
```
douglashall: I still don't understand why this is needed. Why aren't we using [format_suffix_patterns](http://www.django-rest-framework.org/api-guide/format-suffixes/) to do this?

enterprise/api/renderers.py
douglashall: Can we make an upstream PR to allow the default content type returned to be configurable?

enterprise/api/renderers.py
georgebabey: Personally I dislike the format suffix, but that page does indicate support by default for [query params](http://www.django-rest-framework.org/api-guide/format-suffixes/#query-parameter-formats)

Based on the last paragraph on that page, it looks like the DRF have something against accept headers so I'm not surprised that isn't supported.

asadiqbal08:
I also followed the DRF documentation to handle xml and json content negotiation as I shared the link above for DRF.

enterprise/api/renderers.py
douglashall: Do we know that passing an appropriate accept header does not work? I would expect that it would.

enterprise/api/renderers.py
asadiqbal08: Seems good but creating a new ticket would be helpful.

asadiqbal08:
@douglashall , Let me share the reasons for overriding `select_renderer` in enterprise repo. 

1- I can remove the `CustomContentNegotiation` class (as per your request) and corresponding file (renderer.py) if you do not prefer the [query parameter format](http://www.django-rest-framework.org/api-guide/format-suffixes/#query-parameter-formats) and allow user to get JSON/XML response for that particular view using .xml and .json suffix at the end of requested URL  e.g. 
  - `http://example.com/enterprise/api/v1/enterprise_catalogs.xml`
  - `http://example.com/enterprise/api/v1/enterprise_catalogs.json` 
      ( Personally, I do not like that syntax) 

2- After removing the `CustomContentNegotiation` class, Default renderer would be `xml` instead of `json` for that particular view due to [djangorestframework-xml](https://pypi.python.org/pypi/djangorestframework-xml/1.3.0) library. DRF ignores the order in [renderer_classes](https://github.com/edx/edx-enterprise/blob/master/enterprise/api/v1/views.py#L305)

I did not have any other choices to give support for query parameters + to default JSON contents when using the external xml library. 

@georgebabey FYI

zubair-arbi:
@asadiqbal08 We can try setting default renderer to json in settings as mentioned in the docs: https://github.com/jpadilla/django-rest-framework-xml#example

asadiqbal08:
some past discussions on [stackoverflow](https://stackoverflow.com/a/29459782) would be helpful.

asadiqbal08:
@douglashall I came to know that, You are taking off tomorrow. So do you want to share your thoughts on that? If you have any other solution to which I look around then let me know. Thanks.

douglashall:
@asadiqbal08 I would dig into why exactly setting the default renderer appropriately does not work.

asadiqbal08:
Thanks @douglashall . Today, I looked around the code of select_renderer in django-rest-framework further to investigate why exactly setting the default renderer appropriately does not work?

So, You will get the right answer inside [django-rest-framework](https://github.com/edx/django-rest-framework/blob/master/rest_framework) in [select_renderer](https://github.com/edx/django-rest-framework/blob/master/rest_framework/negotiation.py#L58) method. 
in that method, [renderers](https://github.com/edx/django-rest-framework/blob/master/rest_framework/negotiation.py#L47) contain a list of renderers defined in [renderer_classes](https://github.com/edx/edx-enterprise/blob/master/enterprise/api/v1/views.py#L305) e.g. [JSON, XML] and [order_by_precedence(accepts)](https://github.com/edx/django-rest-framework/blob/master/rest_framework/negotiation.py#L55) returns a list of set: 

`[set([u'application/xml;q=0.9', u'text/html', u'application/xhtml+xml']), set([u'*/*;q=0.8'])]` and iterate over each:

The inner loop over [renderers](https://github.com/edx/django-rest-framework/blob/master/rest_framework/negotiation.py#L56) look for each renderer in the first [media_type_set](https://github.com/edx/django-rest-framework/blob/master/rest_framework/negotiation.py#L55) and return XML.

douglashall:
@asadiqbal08 Giving precedence to the `Accepts` header makes sense to me. If a client provides an `Accepts` header, it should be respected. Only when the client does not make it clear which format they want (either through an `Accepts` header, URL suffix, or format query parameter), should content negotiation default to code-defined content types. I believe this is how DRF works by default without any customizations in edx-enterprise.

asadiqbal08:
So do you mean that, Default XML response for that view is acceptable? If client want specific json format then client will add the corresponding suffix at the end of URL e.g. /api/endpoint.json

asadiqbal08:
So do you mean that, Default XML response for that view is acceptable? If client want specific json format then client will add the corresponding suffix at the end of URL e.g. /api/endpoint.json

douglashall:
The client should be able to specify a format in one of three ways:

1. Accepts header
2. format query parameter
3. URL extension

If the client for some reason does not specify format in one of these three ways, we should be able to set the default content type returned via `renderer_classes` (which for our purposes should be JSON).

asadiqbal08:
Doug, I just want to understand your comment more closely. 

1- According to me, Client hit the enterprise catalog API endpoint via browser so I did not understand the first way as you have mentioned about "Accept Header Format" , Do u mean that user will hit the endpoint using CURL and give Accept header there ?

2- Format query parameter is disabled in common.py (As I know you are not liking to override select_renderer method in renderer.py file) so do u mean, we should enable the query parameter format in [REST_FRAMEWORK Settings](https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L2384), Currently It has None value ? 

3- If the client for some reason does not specify any format then how can we set the content_type without overriding in DRF ? Is there any other way to set the renderer without overriding?

asadiqbal08:
@douglashall @georgebabey I have removed the renderer.py and corresponding override functionality. 

Following things would be happened with current DRF implementation. 

1- If client specify the Accept-Header & Content-Type in request e.g. using cURL, then DRF will respond according to the Header content as requested.
2- URL extension .xml/.json will work fine as expected by DRF itself. 
3- Query parameter format `?format=xml` is currently disabled in system in [REST_FRAMEWORK_SETTINGS](https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L2384). If it is essential, Then we can enable it. (your suggestions)
4- Default response is `JSON` if hit the api endpoint using cURL but in browser it is returning `XML` as browser set the Accept-Header to XML and prioritize it. e.g.  `Accept:text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8`

I think, We are in good shape with the removal of renderer file and using the default DRF behavior for handing incoming requests.

PR #293, opened 2018-02-28
ENT-823 Refactor Integrated Channels to allow for multiple content types.

douglashall:


integrated_channels/sap_success_factors/exporters/content_metadata.py
```diff
@@ -0,0 +1,95 @@
...
+        Return the status of the content item.
+        """
+        return 'ACTIVE'
+
+    def transform_title(self, content_metadata_item):
+        """
+        Return the title of the content item.
+        """
+        return [{
+            'locale': 'English',
```
brittneyexline: i think we had support before for checking the language of the course and adjusting this based on the language.

integrated_channels/sap_success_factors/transmitters/content_metadata.py
```diff
@@ -30,9 +33,25 @@ def transmit(self, payload, **kwargs):
         Transmit content metadata items to the integrated channel.
         """
         items_to_create, items_to_update, items_to_delete, transmission_map = self._partition_items(payload)
-        self._transmit_delete(items_to_delete)
-        self._transmit_create(items_to_create)
-        self._transmit_update(items_to_update, transmission_map)
+        serialized_items = (
+            self._serialize_items_for_update(list(items_to_create.values())) +
```
georgebabey: Looks like this is just using the base implementation and I know create/update do the same but I'd still recommend using `_serialize_items_for_create` for clarity and to just in case the base implementations someday diverge

douglashall: Definitely. This is a typo, should have been `_serialize_items_for_create`.

PR #296, opened 2018-03-15
Initial model for data sharing page texts

ziafazal:
**Description:** Model to persist data sharing consent page texts

**JIRA:** https://openedx.atlassian.net/browse/WL-1246

### DIFF
--- a/consent/models.py
+++ b/consent/models.py
@@ -292,11 +292,6 @@ class DataSharingConsentPage(TimeStampedModel):
         default=_('Data sharing consent required'),
         help_text=_('Title of page')
     )
-    left_sidebar_header = models.CharField(
-        max_length=255,
-        default=_('Welcome to {platform_name}.'),
-        help_text=_('Header text in left sidebar. platform_name variable may be available')
-    )
     left_sidebar_text = models.TextField(
         null=True,
         blank=True,
@@ -305,12 +300,6 @@ class DataSharingConsentPage(TimeStampedModel):
             "opportunities from the world's best universities."
         ),
         help_text=SIDEBAR_PARAGRAPH_HELP_TEXT
-    )
-    message_header = models.TextField(
-        null=True,
-        blank=True,
-        default=_('Consent to share your data'),
-        help_text=_('Heading of page')
     )
     top_paragraph = models.TextField(
         null=True,
--- a/consent/models.py
+++ b/consent/models.py
@@ -5,6 +5,7 @@ Models for edX Enterprise's Consent application.
 
 from __future__ import absolute_import, unicode_literals
 
+import six
 from consent.errors import InvalidProxyConsent
 from consent.mixins import ConsentModelMixin
 from opaque_keys import InvalidKeyError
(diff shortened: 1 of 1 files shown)


consent/models.py
```diff
@@ -239,3 +246,150 @@ class Meta(Consent.Meta):
...
+        help_text=_('Header text in left sidebar. platform_name variable may be available')
+    )
+    left_sidebar_text = models.TextField(
+        null=True,
+        blank=True,
+        default=_(
+            "{enterprise_customer_name} has partnered with {platform_name} to offer you high-quality learning "
+            "opportunities from the world's best universities."
+        ),
+        help_text=SIDEBAR_PARAGRAPH_HELP_TEXT
```
bill-filler: wondering if we could combine left_sidebar_header and left_sidebar_text into one field: left_sidebar_paragraph. To achieve the formatting that is in the default template the paragraph text would need to have markup built in to get <p> between the header and the text

consent/models.py
```diff
@@ -239,3 +246,150 @@ class Meta(Consent.Meta):
...
+        blank=True,
+        default=_('Consent to share your data'),
+        help_text=_('Heading of page')
+    )
+    top_paragraph = models.TextField(
+        null=True,
+        blank=True,
+        default=_('First paragraph of page'),
+        help_text=TOP_PARAGRAPH_HELP_TEXT
+    )
```
bill-filler: same comment about message_header and top_paragraph, maybe combine into one field?

consent/models.py
```diff
@@ -239,3 +246,150 @@ class Meta(Consent.Meta):
...
+        default=_('No, take me back.'),
+        help_text=_('Text of decline link')
+    )
+    policy_dropdown_header = models.CharField(
+        max_length=255,
+        null=True,
+        blank=True,
+        default=_('Data Sharing Policy'),
+        help_text=_('Text of policy drop down')
+    )
```
bill-filler: We might needy another field like show_policy_details that is a boolean. If true, then we allow displaying policy_dropdown_header and policy_paragraph and if false we do not show those fields. We need this for the Boeing/MIT use case where they do not show the policy dropdown or text at all (see attached screenshot)
boeingdsc

consent/models.py
```diff
@@ -239,3 +246,150 @@ class Meta(Consent.Meta):
...
+        max_length=255,
+        default=_('View the data sharing policy'),
+        help_text=_('Text of abort decline link on confirmation dialog box')
+    )
+    enterprise_customer = models.OneToOneField(
+        EnterpriseCustomer,
+        related_name="enterprise_consent_page",
+        on_delete=models.deletion.CASCADE
+    )
+    history = HistoricalRecords()
```
bill-filler: is history necessary? not sure, maybe @douglashall has an opinion

consent/models.py
ziafazal: I was thinking of displaying policy section only if both `policy_dropdown_header ` and `policy_paragraph` are not empty.

johnnagro:
I am closing old, inactive PRs. If you feel this has been closed in error please comment.

PR #298, opened 2018-03-22
Updated isort

ziafazal:
**Description:** This PR updates isort package to it latest version(4.3.4) and sorts order of imports using it.

**Merge checklist:**

- [ ] All reviewers approved
- [x] CI build is green
- [ ] Version bumped
- [x] Changelog record added

**Post merge:**

migrations, etc.


georgebabey:
@ziafazal Please note the isort config we use is in our [tox.ini file](https://github.com/edx/edx-enterprise/blob/master/tox.ini#L34) so if you run isort globally you'll get a different result than from within tox.

You'll want to either run `make validate` or run isort within the .tox/ folder  There is no need to change all of the import ordering

ziafazal:
@georgebabey Thanks for the info. It appears new version of `isort` treats `consent` and `integrated` apps as first party apps instead of third party apps and that is in fact true. This PR has imports sorted that way by `isort`. Could you please review?

georgebabey:
@ziafazal That's probably due to the change in tox.ini here - https://github.com/edx/edx-enterprise/pull/298/files#diff-b91f3d5bd63fcd17221b267e851608e8R39  But I agree those should be first party. Thanks for updating!

ziafazal:
@bill-filler can you review this too?. It appears I need 2 thumbs up to before merging it.

PR #299, opened 2018-03-23
ENT-916 Temporarily disable linked learners list on manage learners Django admin page for performance reasons.

douglashall:
**Sandox**
https://business.sandbox.edx.org/admin/enterprise/enterprisecustomer/2b5a2956-7746-4fc9-9587-bc887ba45973/manage_learners

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.66.1] - 2018-03-23
+---------------------
+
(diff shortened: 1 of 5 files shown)


mattdrayer:
Whoa -- how temporary is temporary?  Seems liek we could just avoid the couple of customers with lots of linked learner records for now?  For example, I need access to this screen for troubleshooting ENT-925

mattdrayer:
Ah, looking at the code I'm seeing we can still get to the screen we just have to provide a search keyword -- ok, i'm good w/ it :)

PR #300, opened 2018-04-03
ENT-952 Integrated channel refactoring cleanup.

douglashall:
Add chunking to SAPSF content metadata item transmission.
Add migration for deletion of CatalogTransmissionAudit.

### DIFF
--- a/integrated_channels/integrated_channel/migrations/0006_delete_catalogtransmissionaudit.py
+++ b/integrated_channels/integrated_channel/migrations/0006_delete_catalogtransmissionaudit.py
@@ -1,0 +1,17 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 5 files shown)


douglashall:
@georgebabey I don't think we've tested the limits of the Degreed API and I don't remember seeing anything about them in the docs. However, the Degreed integrated channel is currently set up to [chunk](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/transmitters/content_metadata.py#L135) for each create/read/delete operation. Note that this chunking happens in the base integrated channel implementation so future integrated channels can also take advantage of this. We needed to retrofit this to the SAPSF integrated channel because the 1-request-per-5-minutes limitation prevents us from taking advantage of the base class chunking linked about since we can't send consecutive API requests for the create/update/delete operations.

PR #305, opened 2018-04-20
[ENT-962] Update sessions whenever Enterprise Customer is created.

UmanShahzad:
**Description:** This PR will allow the request session to update with the latest Enterprise Customer for the user that just became an Enterprise user.

**JIRA:** https://openedx.atlassian.net/browse/ENT-962

**Dependencies:** https://github.com/edx/edx-platform/pull/17969

**Merge deadline:** ASAP.

**Testing instructions:**

See https://github.com/edx/edx-platform/pull/17969

However in addition, you'll want to see that your request session is updated when your user becomes linked to an Enterprise through some request.

1. Go to `/enterprise/{enterprise_uuid}/course/{course_run_key}/enroll/`.
2. After going through the process, your user should be linked and that data should be in the session. If the customer had `replace_sensitive_sso_username` on, this'll give you a new username on the top-right corner, for example, without needing to relog. Go to the dashboard to confirm.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.67.6] - 2018-04-20
+---------------------
+
+* Update user session when they become an Enterprise learner.
+
 [0.67.5] - 2018-04-18
 ---------------------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.67.5"
+__version__ = "0.67.6"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -212,6 +212,12 @@ class EnterpriseCustomer(TimeStampedModel):
         """
         return self.enable_audit_enrollment and self.enable_audit_data_reporting
 
+    @property
+    def serialized(self):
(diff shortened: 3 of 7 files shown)


UmanShahzad:
This is on the business sandbox.

pomegranited:
:+1: @UmanShahzad  Works as described 😄 

- [x] I tested this on the business sandbox using the PR test instructions
- [x] I read through the code
- [x] <strike>I checked for accessibility issues</strike> N/A
- [x] <strike>Includes documentation</strike> bugfix, N/A

UmanShahzad:
@pomegranited Thanks a lot :)

georgebabey:
@UmanShahzad Any reason not to do this as a post_save signal?

UmanShahzad:
@georgebabey I tried that first, but saving doesn't send the request object which we need to update the session.

UmanShahzad:
@georgebabey Addition to comment above: I spent a good 1-2 hours trying to find ways to get the request object, but ultimately I think it needs the cookie from the user's browser, so it just won't work in any scenario where the request object isn't available and loaded with that session.

georgebabey:
@UmanShahzad Ah good call. That makes sense, thanks!

edx-webhook:
Although this pull request is already merged, I've created [OSPR-5537](https://openedx.atlassian.net/browse/OSPR-5537) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks again for your contribution.

PR #307, opened 2018-04-25
[ENT-958]Update enrollment api authorization to check group permissions

asadiqbal08:
**Description:** Update enrollment api authorization to check group permissions

**JIRA:** [ENT-958](https://openedx.atlassian.net/browse/ENT-958)

Testing Instructions: 

1. Go to admin portal on [sandbox ](http://business.sandbox.edx.org/admin/) and add a new group with name **enterprise_enrollment_api_access**.
2. Hit the [course_enrollments](https://business.sandbox.edx.org/enterprise/api/v1/enterprise-customer/0a35930c-168c-4955-886c-db99eb699007/course_enrollments/) API endpoint via using Postman tool or the normal flow. 
3. Try to execute the following test cases. 
    - If a user is not staff or not belong to 'enterprise_enrollment_api_access' group then permission denied should be raised with a message u"User is not allowed to access the view."
    - A staff user or a user belong to 'enterprise_enrollment_api_access' group should be allowed to access the api endpoint.

### DIFF
--- a/enterprise/api/v1/permissions.py
+++ b/enterprise/api/v1/permissions.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+
+""" Custom API permissions. """
+
+from __future__ import absolute_import, unicode_literals
+
+from rest_framework import permissions
+
+
+class IsAdminUserOrInGroup(permissions.IsAdminUser):
+    """
+    Find the requesting user is either staff or belong to specific group.
+    It will return a 403 forbidden response with a message if the user is not authorized to access the view.
+    """
+    ALLOWED_API_GROUPS = []  # pylint: disable=invalid-name
+    message = u'User is not allowed to access the view.'
+
+    def has_permission(self, request, view):
+        """
+        Returns True if the requesting user is either 'staff' or belong to specific group, False otherwise.
+        """
+        return super(IsAdminUserOrInGroup, self).has_permission(request, view) or request.user.groups.filter(
+            name__in=self.ALLOWED_API_GROUPS
+        ).exists()
+
+
+class HasEnterpriseEnrollmentAPIAccess(IsAdminUserOrInGroup):
+    """
+    Find the requesting user is either staff or belong to enterprise_enrollment_api_access group.
+    It will return a 403 forbidden response with a message if the user is not authorized to access the view.
+    """
+    def __init__(self):
+        """ Initialize the class with a API_ALLOWED_GROUPS """
+        self.ALLOWED_API_GROUPS = [u'enterprise_enrollment_api_access', ]  # pylint: disable=invalid-name
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -27,6 +27,7 @@ from enterprise.api.pagination import get_paginated_response
 from enterprise.api.throttles import ServiceUserThrottle
 from enterprise.api.v1 import serializers
 from enterprise.api.v1.decorators import enterprise_customer_required, require_at_least_one_query_parameter
+from enterprise.api.v1.permissions import HasEnterpriseEnrollmentAPIAccess
 from enterprise.api_client.discovery import CourseCatalogApiClient
 from enterprise.constants import COURSE_KEY_URL_PATTERN
 
@@ -171,8 +172,12 @@ class EnterpriseCustomerViewSet(EnterpriseReadWriteModelViewSet):
         serializer.update_enterprise_courses(enterprise_customer, catalog_id=enterprise_customer.catalog)
         return get_paginated_response(serializer.data, request)
 
-    @detail_route(methods=['post'])
-    def course_enrollments(self, request, pk):  # pylint: disable=invalid-name,unused-argument
+    @detail_route(methods=['post'], permission_classes=[
+        permissions.IsAuthenticated,
+        HasEnterpriseEnrollmentAPIAccess,
+    ])
+    # pylint: disable=invalid-name,unused-argument
+    def course_enrollments(self, request, pk):
         """
         Creates a course enrollment for an EnterpriseCustomerUser.
         """
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,10 @@ Change Log
 Unreleased
 ----------
 
+[0.68.3] - 2018-05-11
+---------------------
+
+* Update enrollment api authorization to check group permissions.
 
 [0.68.2] - 2018-05-10
 ---------------------
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.68.2"
+__version__ = "0.68.3"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/test_utils/factories.py
+++ b/test_utils/factories.py
@@ -10,7 +10,7 @@ from uuid import UUID
 import factory
 from faker import Factory as FakerFactory
 
-from django.contrib.auth.models import User
+from django.contrib.auth.models import Group, User
 from django.contrib.sites.models import Site
 from django.utils import timezone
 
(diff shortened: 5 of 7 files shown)


enterprise/api/v1/decorators.py
```diff
@@ -89,3 +89,41 @@ def wrapper(request, *args, **kwargs):
             return view(request, *args, **kwargs)
         return wrapper
     return outer_wrapper
+
+def is_user_authorised(view):
+    """
+    Ensure the user making the API request is staff or belong to enterprise_enrollment_api_access group.
+
+    This decorator attempts to find the requesting user is either staff or belong to 'enterprise_enrollment_api_access'
```
zubair-arbi: Convert this section to `Raises`
```python
"""
Ensure that the user making the API request is either staff or belongs to the 
"enterprise_enrollment_api_access" group.

Raises:
    PermissionDenied: User neither staff nor authorized enterprise user.

Usage:
    ....
```
"""

enterprise/api/v1/views.py
```diff
@@ -26,7 +26,9 @@
 from enterprise.api.pagination import get_paginated_response
 from enterprise.api.throttles import ServiceUserThrottle
 from enterprise.api.v1 import serializers
-from enterprise.api.v1.decorators import enterprise_customer_required, require_at_least_one_query_parameter
+from enterprise.api.v1.decorators import enterprise_customer_required, is_user_authorised, \
```
zubair-arbi:
```python
from enterprise.api.v1.decorators import (
    enterprise_customer_required,
    is_user_authorised,
    require_at_least_one_query_parameter,
)
```

asadiqbal08:
@brittneyexline, It also seems me an elegant way to use permission_classes decorator but I was looking into the [@permission_classes](http://www.django-rest-framework.org/api-guide/permissions/) decorator [over the net](https://code.i-harness.com/en/q/224df9a) and tried to implement it but It seems me, It is not helpful on class based views ( as in our case ~ to specific endpoint). 

So in the PR, I followed the same approach that was used for other views via using decorators wrapper. Please [look here ](https://github.com/edx/edx-enterprise/blob/master/enterprise/api/v1/decorators.py)

brittneyexline:
@asadiqbal08, it would be good to dig into this further and understand why it's not working and the options we have to use it. I also think that it's likely that we will want to check users against different groups, so it would be nice if we could make whatever gets built extensible to other group checks.

asadiqbal08:
@brittneyexline The permission_classes and authentication_classes decorators are designed for function based views. [Reference](https://stackoverflow.com/questions/35239638/method-decorator-in-class-based-view-is-not-working)
However, We can pass the list of groups as `allowed_groups` to `is_user_authorised` decorator then can check the existence of user in different groups as you mentioned above.

asadiqbal08:
I will dig into more tomorrow, May be we can use `permission_classes` inside `detail_route` instead of the decorator. post final suggestion tomorrow.

asadiqbal08:
@brittneyexline Kindly look into second commit here. I try to use the permission_classes for the `/course_enrollments/` endpoint.

enterprise/api/v1/permissions.py
```diff
@@ -0,0 +1,16 @@
+""" Custom API permissions. """
+
+from rest_framework import permissions
+
+
+class IsAdminUserOrInGroup(permissions.BasePermission):
```
brittneyexline: As is, this class is specific to the enterprise api group. I would recommend making this an abstract parent class, and having a class specific to the group that inherits from it. Then the constant for the enrollment api access group that is defined below can be moved into the permission class.

enterprise/api/v1/views.py
```diff
@@ -171,7 +174,7 @@ def courses(self, request, pk=None):  # pylint: disable=invalid-name,unused-argu
         serializer.update_enterprise_courses(enterprise_customer, catalog_id=enterprise_customer.catalog)
         return get_paginated_response(serializer.data, request)
 
-    @detail_route(methods=['post'])
+    @detail_route(methods=['post'],  permission_classes=[IsAdminUserOrInGroup])
```
brittneyexline: do we know if using this overrides the permission classes for the view set or is this checked additionally?

enterprise/api/v1/permissions.py
asadiqbal08: @brittneyexline I am using a Rest-Framework class hierarchy in which our classes are inherit from permissions.BasePermission and provide the implementation of has_permission method. 
Making an abstract base class does not look good to me here. 
Simply, I would suggest something like that: 
```
class IsEnterpriseEnrollmentAPIAccessed(permissions.IsAdminUser):
    ENTERPRISE_API_ALLOWED_GROUPS = ['enterprise_enrollment_api_access', ]

    message = 'User is not allowed to access the view.'
    def has_permission(self, request, view, *args):
        return super(IsEnterpriseEnrollmentAPIAccessed, self).has_permission(request, view) or \
               request.user.groups.filter(name__in=self.ENTERPRISE_API_ALLOWED_GROUPS).exists()

```

enterprise/api/v1/views.py
asadiqbal08: I was thinking that, DRF first checks for global permissions (those from the settings file), then for view permissions (declared in permission_classes -- if these exist, they will override global permissions) and only after that for method permissions (declared with the decorator @permission_classes) **but** It does not work like that. So here, It will override the permission_classes set on it before. 
So, We can define something like that for our POST call to `\course_enrollments\` endpoint. 

 ```
@detail_route(methods=['post'],  permission_classes=[
        permissions.IsAuthenticated,
        permissions.DjangoModelPermissions,
        IsEnterpriseEnrollmentAPIAccessed,
    ])
 def course_enrollments(self, request, pk):  # pylint: disable=invalid-name,unused-argument
    .....



```

Comment (brittneyexline) [reply enterprise/api/v1/views.py]: I think we should add permissions.IsAuthenticated, but leave out permissions.DjangoModelPermissions. Thanks for confirming.

File: enterprise/api/v1/permissions.py (hunk 3 above)
Comment (brittneyexline) [reply enterprise/api/v1/permissions.py]: At a high level, it would be good to separate the common logic of check if this user is in a certain group away from checking for a specific group, so that it can be reused later. It doesn't have to be an abstract class, it could be a mixin or something else. All I care about is reusability of the group checking logic and naming the class responsible for this particular group's check in a way that is specific to that group.

File: enterprise/api/v1/permissions.py [hunk 5]
```diff
@@ -0,0 +1,30 @@
...
+    """
+    ALLOWED_API_GROUPS = []
+    message = 'User is not allowed to access the view.'
+
+    def has_permission(self, request, view):
+        """Returns True if the requesting user is either 'staff' or belong to specific group, False otherwise."""
+        return super(IsAdminUserOrInGroup, self).has_permission(request, view) or \
+               request.user.groups.filter(name__in=self.ALLOWED_API_GROUPS).exists()
+
+class IsEnterpriseEnrollmentAPIAccessed(IsAdminUserOrInGroup):
```
brittneyexline: nit: rename to HasEnterpriseEnrollmentAPIAccess

enterprise/api/v1/permissions.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    Find the requesting user is either staff or belong to 'enterprise_enrollment_api_access' group.
+    It will return a 403 forbidden response with a message if the user is not authorized to access the view.
+    """
+    def __init__(self):
+        """ Initialize the class with a API_ALLOWED_GROUPS """
+        self.ALLOWED_API_GROUPS = ['enterprise_enrollment_api_access', ]
+
+    def has_permission(self, request, view):
```
brittneyexline: If you are just calling the parent class, then you shouldn't need this defined.

enterprise/api/v1/permissions.py
asadiqbal08: ah, yes make sense. Thanks

enterprise/api/v1/permissions.py
asadiqbal08: ah, yes make sense. Thanks

enterprise/api/v1/views.py
```diff
@@ -171,8 +172,13 @@ def courses(self, request, pk=None):  # pylint: disable=invalid-name,unused-argu
         serializer.update_enterprise_courses(enterprise_customer, catalog_id=enterprise_customer.catalog)
         return get_paginated_response(serializer.data, request)
 
-    @detail_route(methods=['post'])
-    def course_enrollments(self, request, pk):  # pylint: disable=invalid-name,unused-argument
+    @detail_route(methods=['post'], permission_classes=[
+        permissions.IsAuthenticated,
+        permissions.DjangoModelPermissions,
```
brittneyexline: we need to remove this particular permission check for this endpoint

asadiqbal08: I removed but a certain (pre-existing) test was failing on that. So I added that to that end point.

brittneyexline: We should adjust the test then. This new permission class is meant to replace this one for this endpoint.

PR #309, opened 2018-05-03
ENT-965: Added order_by on queries inside view set to resolve Pagination warning.

saleem-latif:
**Description:** 
Since updating to Django 1.11 we are getting a number of warnings in the log:

UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: `<class 'enterprise.models.EnterpriseCustomerUser'>` QuerySet

This PR fixes this by adding ordering to our ViewSets that don't include it already.

**JIRA:** [ENT-965](https://openedx.atlassian.net/browse/ENT-965)

Simply Install edx-enterprise with this branch.

**Testing instructions:**

1. Open all the enterprise API endpoints one by one, API endpoints are listed below
   - https://business.sandbox.edx.org/enterprise/api/v1/enterprise_catalogs/
   - https://business.sandbox.edx.org/enterprise/api/v1/enterprise-course-enrollment/
   - https://business.sandbox.edx.org/enterprise/api/v1/enterprise-customer/
   - https://business.sandbox.edx.org/enterprise/api/v1/enterprise-learner/
   - https://business.sandbox.edx.org/enterprise/api/v1/enterprise-customer-branding/
   - https://business.sandbox.edx.org/enterprise/api/v1/enterprise-customer-entitlement/
   - https://business.sandbox.edx.org/enterprise/api/v1/enterprise_customer_reporting/ 
2. Make sure `UnorderedObjectListWarning` does not appear in the logs.

**Merge checklist:**

- [x] All reviewers approved
- [ ] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.67.8] - 2018-05-04
+---------------------
+
+* Added ordering to resolve warnings of probable invalid pagination data.
+
 [0.67.7] - 2018-04-23
 ---------------------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.67.7"
+__version__ = "0.67.8"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/enterprise/migrations/0043_auto_20180507_0138.py
+++ b/enterprise/migrations/0043_auto_20180507_0138.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('enterprise', '0042_replace_sensitive_sso_username'),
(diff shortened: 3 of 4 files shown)


enterprise/api/v1/views.py
```diff
@@ -197,7 +197,7 @@ class EnterpriseCourseEnrollmentViewSet(EnterpriseReadWriteModelViewSet):
     API views for the ``enterprise-course-enrollment`` API endpoint.
     """
 
-    queryset = models.EnterpriseCourseEnrollment.objects.all()
+    queryset = models.EnterpriseCourseEnrollment.objects.order_by("created").all()
```
zubair-arbi: Can we set ordering on model level: https://docs.djangoproject.com/en/2.0/ref/models/options/#django.db.models.Options.ordering ?
nit: use single quotes

saleem-latif: We can use the ordering at the module level, but then it will affect everywhere the models are being used. I am not sure if that would be a good idea.

zubair-arbi: Please try to use model level ordering and see if any test fails with the change. If no issues are detected I would suggest that we use model level ordering.

saleem-latif:
@UmanShahzad good catch, I did not think we needed migrations for Meta options. I have added the missing migrations. Can you take another look.

PR #310, opened 2018-05-09
drop sap_success_factors_historicalsapsuccessfactorsenterprisecus80ad

sanfordstudent:
**Description:** Drops the table specified for https://openedx.atlassian.net/browse/EDUCATOR-2681 . 

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/integrated_channels/sap_success_factors/migrations/0014_drop_historical_table.py
+++ b/integrated_channels/sap_success_factors/migrations/0014_drop_historical_table.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import connection, migrations
+
+
+def dropHistoricalTable(apps, schema_editor):
+    """
+    Drops the historical sap_success_factors table named herein.
+    """
+    table_name = 'sap_success_factors_historicalsapsuccessfactorsenterprisecus80ad'
+    if table_name in connection.introspection.table_names():
+        migrations.DeleteModel(
+            name=table_name,
+        )
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('sap_success_factors', '0013_auto_20180306_1251'),
+    ]
+
+    operations = [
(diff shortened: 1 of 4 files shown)


integrated_channels/sap_success_factors/migrations/0014_drop_historical_table.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import connection, migrations
+
+
+def dropHistoricalTable(apps, schema_editor):
```
brittneyexline: i'm wondering if you need to remove the historical records here too? https://github.com/edx/edx-enterprise/blob/master/integrated_channels/sap_success_factors/models.py#L101

also i'm curious about what the PII is in this table?

sanfordstudent: You're right. I should remove that line. I think the PII in this table is the success factors user ID.

iloveagent57: I agree, sorry I missed that.

brittneyexline: and just to double check, would running makemigrations now create the migration to delete the table without having to do the custom migration to delete it?

sanfordstudent: I'm not sure, because I can't actually get edx-enterprise up and running locally. `make requirements` fails consistently. I'll give it a bit more of a shot before I merge this though.

brittneyexline: hmm, feel free to drop by if you are still having trouble. i'm on the 10th floor now, happy to help!

PR #316, opened 2018-05-21
[ENT-980] WIP: Version bump for catalog API endpoint.

UmanShahzad:
**Description:** A WIP for what is described as a solution in https://openedx.atlassian.net/browse/ENT-980.

**JIRA:** https://openedx.atlassian.net/browse/ENT-980


edx-webhook:
Although this pull request is already closed, I've created [OSPR-5532](https://openedx.atlassian.net/browse/OSPR-5532) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #322, opened 2018-05-29
[ENT-995] - Fix the Direct-to-Audit enrollment issue in case of course.

asadiqbal08:
**Description:** 

The direct-to-audit enrollment feature does not appear to work for enrollment URLs containing a Course identifier. After doing some investigation, I came to know that, the following condition is returning False in case of course key instead of course_run_id. [eligible_for_direct_audit_enrollment](https://github.com/edx/edx-enterprise/blob/master/enterprise/views.py#L1386)

> request.path == self.COURSE_ENROLLMENT_VIEW_URL.format(enterprise_customer.uuid, course_run_id)

**JIRA:** [ENT-995](https://openedx.atlassian.net/browse/ENT-995)

#### NOTE 
@georgebabey I still need to test this functionality on sandbox as I am getting some local elasticsearch issue in my local Devstack.

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -1383,6 +1383,27 @@ class RouterView(NonAtomicView):
 
         return enterprise_customer_uuid, course_run_id, course_key, program_uuid
 
+    @staticmethod
+    def get_course_run_id(enterprise_customer, course_key):
+        """
+        User is requesting a course, we need to translate that into the current course run.
+
+        :param enterprise_customer:
+        :param course_key:
+        :return: course_run_id
+        """
+        try:
+            course = CourseCatalogApiServiceClient(enterprise_customer.site).get_course_details(course_key)
+        except ImproperlyConfigured:
+            raise Http404
+
+        course_run = get_current_course_run(course)
+        if course_run:
+            course_run_id = course_run['key']
+            return course_run_id
+        else:
+            raise Http404
+
     def eligible_for_direct_audit_enrollment(self, request, enterprise_customer, course_run_id):
         """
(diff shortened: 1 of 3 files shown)


enterprise/views.py
```diff
@@ -1453,7 +1454,7 @@ def get(self, request, *args, **kwargs):
 
         # Directly enroll in audit mode if the request in question has full direct audit enrollment eligibility.
         resource_id = course_run_id or program_uuid
```
UmanShahzad: I think if you change this to

```
resource_id = course_run_id or program_uuid or course_key
```

then no other code change here is needed.

UmanShahzad: Nevermind, this may not work for the actual `enroll` function or similar right below it though.

enterprise/views.py
```diff
@@ -1393,9 +1393,10 @@ def eligible_for_direct_audit_enrollment(self, request, enterprise_customer, cou
         - The customer's catalog contains the course in question.
         - The audit track is an available mode for the course.
         """
+        course_identifier = course_key if course_key else course_run_id
```
UmanShahzad: Simplify to `course_identifier = course_key or course_run_id`.

UmanShahzad:
@asadiqbal08 My intuition tells me this'll fix the issue, and if Zeshan is already testing this then I'll let 'em do it. I've also restarted any flaky tests.

asadiqbal08:
@georgebabey @UmanShahzad Can you please approve the PR if everything is OK ?

asadiqbal08:
I did not close the PR & where are the changes, How is it happened ? Strange

PR #325, opened 2018-06-07
[ENT-995] Fix the Direct-to-Audit enrollment issue in case of course.

asadiqbal08:
**Description:** 

The direct-to-audit enrollment feature does not appear to work for enrollment URLs containing a Course identifier. After doing some investigation, I came to know that, the following condition is returning False in case of course key instead of course_run_id. [eligible_for_direct_audit_enrollment](https://github.com/edx/edx-enterprise/blob/master/enterprise/views.py#L1386)

> request.path == self.COURSE_ENROLLMENT_VIEW_URL.format(enterprise_customer.uuid, course_run_id)

**JIRA:** [ENT-995](https://openedx.atlassian.net/browse/ENT-995)

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -1404,21 +1404,24 @@ class RouterView(NonAtomicView):
         else:
             raise Http404
 
-    def eligible_for_direct_audit_enrollment(self, request, enterprise_customer, course_run_id):
+    def eligible_for_direct_audit_enrollment(self, request, enterprise_customer, resource_id, course_key=None):
         """
         Return whether a request is eligible for direct audit enrollment for a particular enterprise customer.
 
+        'resource_id' can be either course_run_id or program_uuid.
         We check for the following criteria:
         - The `audit` query parameter.
         - The user's being routed to the course enrollment landing page.
         - The customer's catalog contains the course in question.
         - The audit track is an available mode for the course.
         """
+        course_identifier = course_key or resource_id
+
         # Return it in one big statement to utilize short-circuiting behavior. Avoid the API call if possible.
         return request.GET.get('audit') and \
-            request.path == self.COURSE_ENROLLMENT_VIEW_URL.format(enterprise_customer.uuid, course_run_id) and \
-            enterprise_customer.catalog_contains_course(course_run_id) and \
-            EnrollmentApiClient().has_course_mode(course_run_id, 'audit')
+            request.path == self.COURSE_ENROLLMENT_VIEW_URL.format(enterprise_customer.uuid, course_identifier) and \
+            enterprise_customer.catalog_contains_course(resource_id) and \
+            EnrollmentApiClient().has_course_mode(resource_id, 'audit')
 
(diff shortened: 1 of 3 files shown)


enterprise/views.py
```diff
@@ -1462,7 +1464,7 @@ def get(self, request, *args, **kwargs):
 
         # Directly enroll in audit mode if the request in question has full direct audit enrollment eligibility.
         resource_id = course_run_id or program_uuid
-        if self.eligible_for_direct_audit_enrollment(request, enterprise_customer, resource_id):
+        if self.eligible_for_direct_audit_enrollment(request, enterprise_customer, resource_id, course_key):
```
zubair-arbi: Can we just update above line with 
```python
resource_id = course_key or course_run_id or program_uuid
```
instead of updating the method `eligible_for_direct_audit_enrollment`?

asadiqbal08: No, Course_run_id is in use in that method. this may not work for the actual enroll function or similar right below it though.

georgebabey: This code is very confusing given that we're passing `resource_id` which may be a `program_uuid`, but in `eligible_for_direct_audit_enrollment` the parameter is just called `course_run_id`

I know you didn't update this but I'm actually surprised the other 2 method calls inside `eligible_for_direct_audit_enrollment ` that use this work with programs: (`enterprise_customer.catalog_contains_course` and `EnrollmentApiClient().has_course_mode`)

If we are passing around course_run_id or program_uuid I'd recommend updating the variable name in `eligible_for_direct_audit_enrollment` to `resource_id` with a comment in the method that `resource_id` is `course_run_id` or `program_uuid` to avoid this confusion in the future.  It doesn't look like this gets used anywhere other than this call.

asadiqbal08:
@georgebabey please take another look.

asadiqbal08:
@zubair-arbi As you are the reviewer on the PR, Please take another look. I am going to merge it then. Thanks

UmanShahzad:
@asadiqbal08 I haven't been following the PR any longer and likely won't have time to thoroughly test it, so @zubair-arbi's :+1: is enough.

enterprise/views.py
```diff
@@ -1404,21 +1404,24 @@ def get_course_run_id(enterprise_customer, course_key):
         else:
             raise Http404
 
-    def eligible_for_direct_audit_enrollment(self, request, enterprise_customer, course_run_id):
+    def eligible_for_direct_audit_enrollment(self, request, enterprise_customer, resource_id, course_key=None):
         """
         Return whether a request is eligible for direct audit enrollment for a particular enterprise customer.
 
+        'resource_id' can be either course_run_id or program_uuid.
```
zubair-arbi: nit: use backquotes while quoting variables in docstrings 
```"`resource_id` can be either `course_run_id` or `program_uuid`."```
just like 2 lines below 
```
The `audit` query parameter.
```

PR #327, opened 2018-06-13
Paginate linked learners on manage learners admin view - ENT-933

mushtaqak:
**Description:** The purpose of this PR is to add pagination support to linked learners list on manage learners admin view.

**Testing instructions:**

1. Open Enterprise customer detail view.
2. Click manage learners
3. Expect linked learners list to be paginated.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.70.0"
+__version__ = "0.70.1"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/enterprise/admin/paginator.py
+++ b/enterprise/admin/paginator.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+"""
+Custom paginator to implement smart pagination.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from django.core.paginator import Paginator
+
+
+class CustomPaginator(Paginator):
+    """
+    Adopted from django/core/paginator
+    so as to implement smart links pagination in custom views.
+    """
+    _page_range = []
+
+    @property
+    def page_range(self):
+        """
+        We have customized the getter so that it can return the value of the page_range property
+        instead of always calculating the result.
+        """
+        return self._page_range or list(range(1, self.num_pages + 1))
+
+    @page_range.setter
+    def page_range(self, value):
+        """
+        We have introduced a setter method here, so as to set value for page_range property.
+        This was not present in Paginator class.
+        """
+        self._page_range = value
--- a/enterprise/admin/utils.py
+++ b/enterprise/admin/utils.py
@@ -8,11 +8,17 @@ import unicodecsv
 
 from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
+from django.core.paginator import EmptyPage, PageNotAnInteger
 from django.core.validators import validate_email
 from django.utils.translation import ugettext as _
 
+from enterprise.admin.paginator import CustomPaginator
 from enterprise.api_client.lms import parse_lms_api_datetime
 from enterprise.models import EnterpriseCustomerUser
+
+DOT = '.'
+PAGES_ON_EACH_SIDE = 3
+PAGES_ON_ENDS = 2
 
 
 class ProgramStatuses(object):
@@ -216,3 +222,55 @@ def split_usernames_and_emails(email_field):
     provide a list of email addresses or usernames if it is.
     """
     return [name.strip() for name in email_field.split(',')]
+
+
+# pylint: disable=range-builtin-not-iterating
+def paginated_list(object_list, page, page_size=25):
+    """
+    Returns paginated list.
+
+    Arguments:
+        object_list (QuerySet): A list of records to be paginated.
+        page (int): Current page number.
+        page_size (int): Number of records displayed in each paginated set.
+        show_all (bool): Whether to show all records.
+
+    Adopted from django/contrib/admin/templatetags/admin_list.py
+    https://github.com/django/django/blob/1.11.1/django/contrib/admin/templatetags/admin_list.py#L50
+    """
+    paginator = CustomPaginator(object_list, page_size)
+    try:
+        object_list = paginator.page(page)
+    except PageNotAnInteger:
+        object_list = paginator.page(1)
+    except EmptyPage:
+        object_list = paginator.page(paginator.num_pages)
+
+    page_range = []
+    page_num = object_list.number
+
+    # If there are 10 or fewer pages, display links to every page.
+    # Otherwise, do some fancy
+    if paginator.num_pages <= 10:
+        page_range = range(paginator.num_pages)
+    else:
+        # Insert "smart" pagination links, so that there are always ON_ENDS
+        # links at either end of the list of pages, and there are always
+        # ON_EACH_SIDE links at either end of the "current page" link.
+        if page_num > (PAGES_ON_EACH_SIDE + PAGES_ON_ENDS + 1):
+            page_range.extend(range(1, PAGES_ON_ENDS + 1))
+            page_range.append(DOT)
+            page_range.extend(range(page_num - PAGES_ON_EACH_SIDE, page_num + 1))
+        else:
+            page_range.extend(range(1, page_num + 1))
+        if page_num < (paginator.num_pages - PAGES_ON_EACH_SIDE - PAGES_ON_ENDS):
+            page_range.extend(range(page_num + 1, page_num + PAGES_ON_EACH_SIDE + 1))
+            page_range.append(DOT)
+            page_range.extend(range(paginator.num_pages + 1 - PAGES_ON_ENDS, paginator.num_pages + 1))
+        else:
+            page_range.extend(range(page_num + 1, paginator.num_pages + 1))
+
+        # Override page range to implement custom smart links.
+        object_list.paginator.page_range = page_range
+
+    return object_list
--- a/enterprise/admin/views.py
+++ b/enterprise/admin/views.py
@@ -33,11 +33,13 @@ from enterprise.admin.utils import (
     email_or_username__to__email,
     get_course_runs_from_program,
     get_earliest_start_date_from_program,
+    paginated_list,
     parse_csv,
     split_usernames_and_emails,
     validate_email_to_link,
 )
 from enterprise.api_client.lms import EnrollmentApiClient
+from enterprise.constants import PAGE_SIZE
 from enterprise.models import (
     EnrollmentNotificationEmailTemplate,
     EnterpriseCourseEnrollment,
@@ -221,7 +223,7 @@ class EnterpriseCustomerManageLearnersView(View):
         enterprise_customer = EnterpriseCustomer.objects.get(uuid=customer_uuid)  # pylint: disable=no-member
 
         search_keyword = self.get_search_keyword(request)
-        linked_learners = self.get_enterprise_customer_user_queryset(search_keyword, customer_uuid)
+        linked_learners = self.get_enterprise_customer_user_queryset(request, search_keyword, customer_uuid)
         pending_linked_learners = self.get_pending_users_queryset(search_keyword, customer_uuid)
 
         context = {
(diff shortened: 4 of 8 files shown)


enterprise/__init__.py
```diff
@@ -4,6 +4,6 @@
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.69.1"
+__version__ = "0.69.2"
```
zubair-arbi: @mushtaqak generally we add version update related changes after we have verified our changes on our combined business sandbox (http://business.sandbox.edx.org/) and just before we are going to merge. This make us avoid unnecessary conflicts for version updates and all enterprise team members use this sandbox.

enterprise/admin/views.py
```diff
@@ -241,7 +242,7 @@ def get_search_keyword(self, request):
         """
         return request.GET.get('q', None)
 
-    def get_enterprise_customer_user_queryset(self, search_keyword, customer_uuid):
+    def get_enterprise_customer_user_queryset(self, request, search_keyword, customer_uuid):
```
zubair-arbi: nit: please add `request` in `Args` below and also rename `Args` to complete word `Arguments` http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#docstrings

enterprise/admin/views.py
```diff
@@ -250,11 +251,13 @@ def get_enterprise_customer_user_queryset(self, search_keyword, customer_uuid):
             customer_uuid (str): A unique identifier to filter down to only users linked to a
             particular EnterpriseCustomer.
         """
-        # TODO: Add paging so the admin view does not time out.
-        # For now, we will allow search, but will return an empty list
-        # if no search keyword is provided.
-        learners = []
-        if search_keyword is not None:
+        paginate_by = 10
```
zubair-arbi: lets pass this variable as optional argument in the method (also include it in arguments docstring) instead of hardcoding it here
```python
 def get_enterprise_customer_user_queryset(self, request, search_keyword, customer_uuid, page_size=10):

paginate_by = page_size
```

enterprise/templates/enterprise/admin/manage_learners.html
```diff
@@ -120,7 +118,18 @@ <h1>Linked learners</h1>
...
       </tbody>
     </table>
-
+    {% if learners.has_other_pages %}
+      <p class="paginator">
+        {% for i in learners.paginator.page_range %}
+          {% if learners.number == i %}
+          <span class="this-page"> {{ i }}</span>
+          {% else %}
+          <a href="?p={{ i }}">{{ i }}</a>
```
zubair-arbi: nit: we can use more verbose name here like `page` instead of `p`

enterprise/templates/enterprise/admin/manage_learners.html
```diff
@@ -120,7 +118,18 @@ <h1>Linked learners</h1>
       {% endfor %}
       </tbody>
     </table>
-
+    {% if learners.has_other_pages %}
+      <p class="paginator">
+        {% for i in learners.paginator.page_range %}
```
zubair-arbi: nit `i` -> `page_number`

enterprise/templates/enterprise/admin/manage_learners.html
```diff
@@ -120,7 +118,18 @@ <h1>Linked learners</h1>
...
+          {% if learners.number == i %}
+          <span class="this-page"> {{ i }}</span>
+          {% else %}
+          <a href="?p={{ i }}">{{ i }}</a>
+          {% endif %}
+        {% endfor %}
+        {{learners.paginator.count}} learners <a href="?all=" class="showall">Show all</a>
```
zubair-arbi: nit: need spacing  `{{ learners.paginator.count }} `

enterprise/admin/views.py
```diff
@@ -264,6 +267,14 @@ def get_enterprise_customer_user_queryset(self, search_keyword, customer_uuid):
             matching_user_ids = matching_users.values_list('pk', flat=True)
             learners = learners.filter(user_id__in=matching_user_ids)
 
+        if all_learners is None:
```
zubair-arbi: Can we combine this code with the condition above `if all_learners or search_keyword is None:` ?

enterprise/__init__.py
zubair-arbi: As part of version update we will also need to add changes in https://github.com/edx/edx-enterprise/blob/master/CHANGELOG.rst

mushtaqak: Right on, I will make sure this happens

enterprise/templates/enterprise/admin/manage_learners.html
mushtaqak: Just trying to go along with current django approach, but I am fine whatever name.

enterprise/admin/utils.py
```diff
@@ -216,3 +221,56 @@ def split_usernames_and_emails(email_field):
...
+    except PageNotAnInteger:
+        object_list = paginator.page(1)
+    except EmptyPage:
+        object_list = paginator.page(paginator.num_pages)
+
+    page_range = []
+    page_num = object_list.number
+
+    # If there are 10 or fewer pages, display links to every page.
+    # Otherwise, do some fancy
```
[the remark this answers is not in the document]
mushtaqak: This is written in django as well. And this helps in undrstanding when to show smart links, so I am thinking to keep it.

mushtaqak:
@zubair-arbi Addressed feedback.

enterprise/admin/paginator.py
```diff
@@ -0,0 +1,96 @@
...
+            if self.count == 0 and not self.allow_empty_first_page:
+                self._num_pages = 0
+            else:
+                hits = max(1, self.count - self.orphans)
+                self._num_pages = int(ceil(hits / float(self.per_page)))
+        return self._num_pages
+    num_pages = property(_get_num_pages)
+
+    @property
+    def page_range(self):
```
mushtaqak: The purpose of  adopting the `Paginator` was to make sure `page_range` property gets set correctly and we get the latest value instead of fix old value.

```
def _get_page_range(self):
        """
        Returns a 1-based range of pages for iterating through within
        a template for loop.
        """
        return list(six.moves.range(1, self.num_pages + 1))
page_range = property(_get_page_range)
```

muhammad-ammar:
@mushtaqak Some questions: 
1. Why we can not use the builtin Paginator class? Can you please explain what problem you faced in using the existing class? 
2. From which django version you adopted the Paginator class?

I think we should find a way to use/extend the builtin Paginator class.

enterprise/admin/paginator.py
mushtaqak: @muhammad-ammar  ^^

mushtaqak:
@muhammad-ammar I have extended Paginator class and overridden `page_range` methods.

muhammad-ammar:
@georgebabey @mushtaqak @zubair-arbi [Sandbox](https://business.sandbox.edx.org/admin/enterprise/enterprisecustomer/515d50ee-88a4-4d77-b463-0b043ff10807/manage_learners) is updated with changes in this PR. Please have a look at the pagination.

enterprise/admin/views.py
```diff
@@ -241,20 +244,23 @@ def get_search_keyword(self, request):
...
             particular EnterpriseCustomer.
+            page_size (int): Number of learners displayed in each paginated set.
         """
-        # TODO: Add paging so the admin view does not time out.
-        # For now, we will allow search, but will return an empty list
-        # if no search keyword is provided.
-        learners = []
-        if search_keyword is not None:
+        page = request.GET.get('page', 1)
+        show_all = request.GET.get('all', None)
```
muhammad-ammar: Why we need to support`all`?  This is not working. 
Below url is getting all the results without pagination.
`https://business.sandbox.edx.org/admin/enterprise/enterprisecustomer/515d50ee-88a4-4d77-b463-0b043ff10807/manage_learners?all`

We are setting `all` only when results are less than `10`, So I would suggest we should show all results by default in this case. On a second though we don't need this.

enterprise/admin/views.py
```diff
@@ -241,20 +244,23 @@ def get_search_keyword(self, request):
...
+
+        if show_all or search_keyword is None:
+            learners = EnterpriseCustomerUser.objects.all()
```
muhammad-ammar: Why we need this `if` block? As per my understanding we only need the code inside below `else` block plus `learners = paginated_list(learners, page, page_size)`

enterprise/admin/views.py
```diff
@@ -264,6 +270,9 @@ def get_enterprise_customer_user_queryset(self, search_keyword, customer_uuid):
             matching_user_ids = matching_users.values_list('pk', flat=True)
             learners = learners.filter(user_id__in=matching_user_ids)
 
+        if show_all is None:
```
muhammad-ammar: I think we should not handle show_all

enterprise/admin/views.py
```diff
@@ -241,20 +244,23 @@ def get_search_keyword(self, request):
         """
         return request.GET.get('q', None)
 
-    def get_enterprise_customer_user_queryset(self, search_keyword, customer_uuid):
+    def get_enterprise_customer_user_queryset(self, request, search_keyword, customer_uuid, page_size=10):
```
muhammad-ammar: I think default `page_size` should be `25`. It would be nice if we can store it as constant.

enterprise/templates/enterprise/admin/manage_learners.html
```diff
@@ -120,7 +118,21 @@ <h1>Linked learners</h1>
       {% endfor %}
       </tbody>
     </table>
-
+    {% if learners.has_other_pages %}
```
muhammad-ammar: Where this is coming from?

enterprise/templates/enterprise/admin/manage_learners.html
```diff
@@ -120,7 +118,21 @@ <h1>Linked learners</h1>
...
+      <p class="paginator">
+        {% for page_number in learners.paginator.page_range %}
+          {% if page_number == '.' %}
```
muhammad-ammar: This is weird. Comparing an integer variable with string?

enterprise/admin/views.py
muhammad-ammar: @mushtaqak  To be clear we don't need to support `all` parameter and its functionality.

enterprise/templates/enterprise/admin/manage_learners.html
mushtaqak: learners is a Page object now, so it has `has_other_pages` property.

enterprise/templates/enterprise/admin/manage_learners.html
mushtaqak: Adopted from https://github.com/django/django/blob/94a180402c9e52bca2e911b6427c88f436e95449/django/contrib/admin/templatetags/admin_list.py#L34

enterprise/admin/views.py
```diff
@@ -241,20 +244,23 @@ def get_search_keyword(self, request):
...
-        if search_keyword is not None:
+        page = request.GET.get('page', 1)
+        show_all = request.GET.get('all', None)
+
+        if show_all or search_keyword is None:
+            learners = EnterpriseCustomerUser.objects.all()
+        else:
             learners = EnterpriseCustomerUser.objects.filter(enterprise_customer__uuid=customer_uuid)
             user_ids = learners.values_list('user_id', flat=True)
             matching_users = User.objects.filter(
```
muhammad-ammar: @mushtaqak  
```python
matching_users = User.objects.filter(pk__in=user_ids)
if search_keyword is not None:
    matching_users.filter(Q(email__icontains=search_keyword) | Q(username__icontains=search_keyword))
```

enterprise/admin/utils.py
```diff
@@ -216,3 +220,56 @@ def split_usernames_and_emails(email_field):
...
+    Returns paginated list.
+
+    Arguments:
+        object_list (QuerySet): A list of records to be paginated.
+        page (int): Current page number.
+        page_size (int): Number of records displayed in each paginated set.
+        show_all (bool): Whether to show all records.
+
+    Adopted from django/contrib/admin/templatetags/admin_list.py
+    https://github.com/django/django/blob/94a180402c9e52bca2e911b6427c88f436e95449/django/contrib/admin/templatetags/admin_list.py#L45
```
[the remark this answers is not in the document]
mushtaqak: @muhammad-ammar  I have added the link to the method, the method is almost same except the page_range and unpacking (which we have for python 2.7) and **we are are setting page_range** here

mushtaqak:
@zubair-arbi @muhammad-ammar Feedback addressed.

enterprise/admin/utils.py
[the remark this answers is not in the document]
zubair-arbi: @mushtaqak can you please update this reference link for a specific version of django, probably for version 1.11

enterprise/admin/utils.py
```diff
@@ -8,12 +8,18 @@
...
+from django.core.paginator import EmptyPage, PageNotAnInteger
 from django.core.validators import validate_email
 from django.utils.translation import ugettext as _
 
+from enterprise.admin.paginator import CustomPaginator
 from enterprise.api_client.lms import parse_lms_api_datetime
 from enterprise.models import EnterpriseCustomerUser
 
+DOT = '.'
+ON_EACH_SIDE = 3
```
zubair-arbi: Lets rename this to `EXTRA_PAGES_ON_EACH_SIDE`

enterprise/admin/utils.py
```diff
@@ -8,12 +8,18 @@
...
+ON_ENDS = 2
```
zubair-arbi: lets rename this to `PAGES_ON_ENDS`

tests/test_admin/test_view.py
```diff
@@ -245,28 +272,76 @@ def test_get_empty_links(self):
...
     def test_get_existing_and_pending_links(self):
         self._login()
 
-        EnterpriseCustomerUserFactory(enterprise_customer=self.enterprise_customer)
-        EnterpriseCustomerUserFactory(enterprise_customer=self.enterprise_customer)
-        EnterpriseCustomerUserFactory(enterprise_customer=self.enterprise_customer)
+        linked_learners = [
+            EnterpriseCustomerUserFactory(
+                enterprise_customer=self.enterprise_customer,
+                user_id=UserFactory(
```
zubair-arbi: nit: `user_id=UserFactory().id`

tests/test_admin/test_view.py
```diff
@@ -317,6 +392,69 @@ def test_get_with_search_param(self):
...
+        response = self.client.get(self.view_url)
+        self._verify_pagination(
+            response.context[self.context_parameters.LEARNERS],
+            linked_learners,
+            linked_learners[:25],
+            current_page_number=current_page_number
+        )
+
+        # Verify we get the records for 2nd page only.
+        current_page_number = 2
```
zubair-arbi: lets try to reduce the code by using `ddt` for different pages

mushtaqak:
@zubair-arbi addressed your concerns.

PR #332, opened 2018-06-26
ENT-1007 Adding slug field to enterprise customer model

brittneyexline:
**Description:** Adding slug field to enterprise customer model

**JIRA:** https://openedx.atlassian.net/browse/ENT-1007

**Dependencies:** n/a

**Merge deadline:** 

**Testing instructions:**
Go to the business sandbox, where the migrations have been run, and see that the slugs are populated.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/migrations/0051_add_enterprise_slug.py
+++ b/enterprise/migrations/0051_add_enterprise_slug.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+from django.utils.text import slugify
+
+
+def populate_slug(apps, schema_editor):
+    EnterpriseCustomer = apps.get_model('enterprise', 'EnterpriseCustomer')
+
+    enterprises = EnterpriseCustomer.objects.all()
+    for enterprise in enterprises:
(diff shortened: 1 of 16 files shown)


enterprise/migrations/0051_add_enterprise_slug.py
```diff
@@ -0,0 +1,35 @@
...
+from django.db import migrations, models
+from django.utils.text import slugify
+
+
+def populate_slug(apps, schema_editor):
+    EnterpriseCustomer = apps.get_model('enterprise', 'EnterpriseCustomer')
+
+    enterprises = EnterpriseCustomer.objects.all()
+    for enterprise in enterprises:
+        enterprise.slug = slugify(enterprise.name)
```
georgebabey: Not sure we need to pre-populate these, especially if we're allowing the field to be blank

georgebabey: Nevermind - I see now the subsequent migration makes this field unique, so they can't all be blank.

PR #334, opened 2018-06-27
ENT-1008 Modify branding config api to look up by enterprise slug

brittneyexline:
**Description:** We need an api that takes the enterprise slug and returns the branding config. The existing branding config api isn't used and lists details by id, but we would rarely ever know the specific branding config id, so I changed it to use enterprise slug.

**JIRA:** https://openedx.atlassian.net/browse/ENT-1008

**Dependencies:** n/a

**Merge deadline:** 

**Testing instructions:**
Hit the following endpoint on the business sandbox using JWT authentication: https://business.sandbox.edx.org/enterprise/api/v1/enterprise-customer-branding/
This should list all branding configs and now include the enterprise slug.
Then hit https://business.sandbox.edx.org/enterprise/api/v1/enterprise-customer-branding/aviato
(or use any enterprise slug value you wish)
This should return just the branding config details for the appropriate enterprise.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/test_utils/factories.py
+++ b/test_utils/factories.py
@@ -18,6 +18,7 @@ from consent.models import DataSharingConsent, DataSharingConsentTextOverrides
 from enterprise.models import (
     EnterpriseCourseEnrollment,
     EnterpriseCustomer,
+    EnterpriseCustomerBrandingConfiguration,
     EnterpriseCustomerCatalog,
     EnterpriseCustomerEntitlement,
     EnterpriseCustomerIdentityProvider,
@@ -205,7 +206,7 @@ class EnterpriseCustomerEntitlementFactory(factory.django.DjangoModelFactory):
     EnterpriseCustomerEntitlement factory.
 
     Creates an instance of EnterpriseCustomerEntitlement with minimal boilerplate - uses this class' attributes as
-    default parameters for EnterpriseCustomerBrandingFactory constructor.
+    default parameters for EnterpriseCustomerEntitlementFactory constructor.
     """
 
     class Meta(object):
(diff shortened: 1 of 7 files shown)


test_utils/factories.py
```diff
@@ -220,6 +221,26 @@ class Meta(object):
     enterprise_customer = factory.SubFactory(EnterpriseCustomerFactory)
 
 
+class EnterpriseCustomerBrandingConfigurationFactory(factory.django.DjangoModelFactory):
+    """
+    EnterpriseCustomerBrandingConfiguration factory.
+
+    Creates an instance of EnterpriseCustomerBrandingConfiguration with minimal boilerplate - uses this class'
+     attributes as default parameters for EnterpriseCustomerBrandingFactory constructor.
```
adamstankiewicz: nit: Is this an extra space before `attributes`?

tests/test_enterprise/api/test_views.py
```diff
@@ -805,6 +825,34 @@ def test_enterprise_customer_reporting_list(self):
         assert result_item['encrypted_password'] is not None
         assert result_item['encrypted_sftp_password'] is not None
 
+    def test_enterprise_customer_branding_detail(self):
+        """
+        ``enterprise_customer_branding``'s get endpoint should get the config by looking up the enterprise slug and
+         serialize the ``EnterpriseCustomerBrandingConfig``.
```
adamstankiewicz: nit: Is this an extra space before `serialize`?

PR #338, opened 2018-07-03
ENT-1074 Attempt to send SAPSF learner data transmissions with course keys first

brittneyexline:
**Description:** Since we have started sending courses to SuccessFactors instead of course runs, we need to attempt to send transmissions with course keys and course run ids in order to ensure that we account for whether courses or course runs exist in the SuccessFactors instance. If the transmission with the course key succeeds, the next one will get skipped. If it fails, the one with the course run id will be attempted and (presumably) succeed.

**JIRA:** https://openedx.atlassian.net/browse/ENT-1074

**Dependencies:** n/a

**Merge deadline:** ASAP

**Testing instructions:** 
I've _sort of_ tested this on the business sandbox, but I am having issues testing the successful transmission of course key because a content object with a course key doesn't seem to exist on the success factors instance and I can't figure out how to get it there. However, I feel reasonably confident because I have written specific unit test cases for it.
I used the following enrollment link and the user mtd_learner9: https://business.sandbox.edx.org/enterprise/d583b097-7bd9-4a7e-a98d-c15dea220354/course/NYIF+CM2.x/enroll/?catalog=573f379d-9238-4a10-9064-22b7b03cda21&utm_medium=enterprise&utm_source=successfactors-company
Note that if it takes you to login with an existing account, you can go to /register to create a new account instead, just make sure you don't have an active ecommerce session on the sandbox.
I completed the course, and ran the following command on the business sandbox:
```
$ edxapp
$ ./manage.py lms --settings=aws transmit_learner_data --enterprise_customer d583b097-7bd9-4a7e-a98d-c15dea220354 --api_user staff
```
Then, look in mysql, you should see the course transmission attempts with the course run id one succeeding.
```
mysql> select * from sap_success_factors_sapsuccessfactorslearnerdatatransmission3ce5 order by id desc limit 5;
```

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/integrated_channels/degreed/exporters/learner_data.py
+++ b/integrated_channels/degreed/exporters/learner_data.py
@@ -11,6 +11,7 @@ from logging import getLogger
 
 from django.apps import apps
 
+from enterprise.utils import parse_course_key
 from integrated_channels.integrated_channel.exporters.learner_data import LearnerExporter
 
 LOGGER = getLogger(__name__)
@@ -21,7 +22,7 @@ class DegreedLearnerExporter(LearnerExporter):
     Class to provide a Degreed learner data transmission audit prepared for serialization.
     """
 
-    def get_learner_data_record(
+    def get_learner_data_records(
             self,
             enterprise_enrollment,
             completed_date=None,
@@ -42,13 +43,24 @@ class DegreedLearnerExporter(LearnerExporter):
                 'degreed',
                 'DegreedLearnerDataTransmissionAudit'
             )
-            return DegreedLearnerDataTransmissionAudit(
-                enterprise_course_enrollment_id=enterprise_enrollment.id,
-                degreed_user_email=enterprise_enrollment.enterprise_customer_user.user_email,
-                course_id=enterprise_enrollment.course_id,
-                course_completed=completed_date is not None and is_passing,
-                completed_timestamp=completed_timestamp,
-            )
+            # We return two records here, one with the course key and one with the course run id, to account for
+            # uncertainty about the type of content (course vs. course run) that was sent to the integrated channel.
+            return [
+                DegreedLearnerDataTransmissionAudit(
+                    enterprise_course_enrollment_id=enterprise_enrollment.id,
+                    degreed_user_email=enterprise_enrollment.enterprise_customer_user.user_email,
+                    course_id=parse_course_key(enterprise_enrollment.course_id),
+                    course_completed=completed_date is not None and is_passing,
+                    completed_timestamp=completed_timestamp,
+                ),
+                DegreedLearnerDataTransmissionAudit(
+                    enterprise_course_enrollment_id=enterprise_enrollment.id,
+                    degreed_user_email=enterprise_enrollment.enterprise_customer_user.user_email,
+                    course_id=enterprise_enrollment.course_id,
+                    course_completed=completed_date is not None and is_passing,
+                    completed_timestamp=completed_timestamp,
+                )
+            ]
         else:
             LOGGER.debug(
                 'No learner data was sent for user [%s] because a Degreed user ID could not be found.',
--- a/integrated_channels/sap_success_factors/exporters/learner_data.py
+++ b/integrated_channels/sap_success_factors/exporters/learner_data.py
@@ -10,6 +10,7 @@ from logging import getLogger
 
 from django.apps import apps
 
+from enterprise.utils import parse_course_key
 from integrated_channels.integrated_channel.exporters.learner_data import LearnerExporter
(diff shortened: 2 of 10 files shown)


integrated_channels/degreed/exporters/learner_data.py
```diff
@@ -42,13 +42,15 @@ def get_learner_data_record(
...
             )
-            return DegreedLearnerDataTransmissionAudit(
-                enterprise_course_enrollment_id=enterprise_enrollment.id,
-                degreed_user_email=enterprise_enrollment.enterprise_customer_user.user_email,
-                course_id=enterprise_enrollment.course_id,
-                course_completed=completed_date is not None and is_passing,
-                completed_timestamp=completed_timestamp,
-            )
+            return [
+                DegreedLearnerDataTransmissionAudit(
```
mattdrayer: I'm trying to decide if we should do this for Degreed as well.  We're transmitting course-level content items to Degreed, so I'm assuming that we will also want to transmit a course-level completion item.

tests/test_integrated_channels/test_degreed/test_exporters/test_learner_data.py
```diff
@@ -73,11 +73,12 @@ def test_get_learner_data_record(self, completed_date, is_passing):
             course_id=self.course_id,
         )
         exporter = DegreedLearnerExporter('fake-user', self.config)
-        learner_data_record = exporter.get_learner_data_record(
+        learner_data_records = exporter.get_learner_data_records(
             enterprise_course_enrollment,
             completed_date=completed_date,
             is_passing=is_passing,
         )
+        learner_data_record = learner_data_records[0]
```
mattdrayer: Are we sure we're going to have at least one record here?  What if the array is empty for some reason?

integrated_channels/sap_success_factors/exporters/learner_data.py
```diff
@@ -42,14 +43,29 @@ def get_learner_data_record(self, enterprise_enrollment, completed_date=None, gr
...
             )
-            return SapSuccessFactorsLearnerDataTransmissionAudit(
-                enterprise_course_enrollment_id=enterprise_enrollment.id,
-                sapsf_user_id=sapsf_user_id,
-                course_id=enterprise_enrollment.course_id,
-                course_completed=course_completed,
-                completed_timestamp=completed_timestamp,
-                grade=grade,
-            )
+            # Since we have started sending courses to SuccessFactors instead of course runs,
```
mattdrayer: I like the amount of detail we've included in this comment, but it seems misplaced to me -- should it be moved to the transmitter, vs. the exporter?

mattdrayer: Also I'm probably just missing it, but I don't see the retry logic for when the first transmission fails?

brittneyexline: Yea, I guess this is ultimately depending on the logic in the transmitter, which is receiving an iterable of course transmissions to try.
In the case that the course key transmission succeeds, the course run id transmission will not get attempted because a transmission by that enrollment id has already succeeded. In the case that the course key transmission fails, there is no successful transmission for the enrollment id and so we try the course run id transmission.

integrated_channels/degreed/exporters/learner_data.py
brittneyexline: I don't really have enough understanding of how Degreed's api behaves to say if we should do this now or not... but I suppose it makes sense.

tests/test_integrated_channels/test_degreed/test_exporters/test_learner_data.py
brittneyexline: the method returns either None or an array with at least one item, changed from returning None or one item.

integrated_channels/sap_success_factors/exporters/learner_data.py
mattdrayer: Maybe we can add some comments explaining this?

tests/test_integrated_channels/test_degreed/test_exporters/test_learner_data.py
mattdrayer: makes sense to me

integrated_channels/degreed/exporters/learner_data.py
mattdrayer: Yeah the more I think about it, we should be transmitting the course-level completion item to Degreed vs. the run-level completion item.

PR #340, opened 2018-07-05
We don't need to tag this repo

nedbat:


nedbat:
@jmbowman this is what we decided, right?

PR #328, opened 2018-06-20
ENT-876: Implement reporting channel of course completion via xAPI

saleem-latif:
**Description:** 
implement XAPI data reporting via XAPI. Following are the two events that need to be tracked via XAPI.

1. Learner registered (enrolled) in a course.
2. Learner completed a course.

**JIRA:** [ENT-876](https://openedx.atlassian.net/browse/ENT-876)

**Installation instructions:**  Simply instal edx-enterprise using this branch.

**Merge checklist:**

- [x] Check that the versions of the requirements in the `platform-master.in` file match edx-platform.
- [x] New requirements are in the right place (`base.in` if only used in enterprise; in the correct `platform-****.in` files if they're hosted in edx-platform)
- [x] Regenerate requirements with `make upgrade && make requirements` (and make sure to fix any errors).
  **DO NOT** just add dependencies to `requirements/*.txt` files.
- [x] Called `make static` for webpack bundling if any static content was updated.
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Documentation updated (not only docstrings)
- [x] Commits are (reasonably) squashed
- [x] Translations are updated
- [x] PR author is listed in AUTHORS

**Post merge:**
- [ ] Create a tag
- [ ] Check new version is pushed to PyPi after tag-triggered build is finished.
- [ ] Delete working branch (if not needed anymore)
- [ ] edx-platform PR (be sure to include edx-platform requirements upgrades that were present in this PR)

PR #341, opened 2018-07-06
[ENT-1083]: Write Serializer to Collect Learner and Enrollment data

saleem-latif:
**Description:** 
Write utility functions/serializers to collect learner and enrollment data that would be sent as part of the xAPI statement.

**JIRA:** [ENT-1083](https://openedx.atlassian.net/browse/ENT-1083)

**Dependencies:**  https://github.com/edx/edx-enterprise/pull/328

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/integrated_channels/xapi/serializers.py
+++ b/integrated_channels/xapi/serializers.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+
+"""
+Serializers for xAPI data.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from rest_framework import serializers
+
+from enterprise.api.v1.serializers import ImmutableStateSerializer
+from enterprise.models import EnterpriseCustomerUser
+
+
+class LearnerInfoSerializer(ImmutableStateSerializer):
+    """
+    Serializer for Learner info.
+
+    Model: User
+    """
+    lms_user_id = serializers.IntegerField(source='id')
+    user_username = serializers.CharField(source='username')
+    user_email = serializers.EmailField(source='email')
+    user_country_code = serializers.CharField(source='profile.country.code')
+    user_account_creation_date = serializers.DateTimeField(source='date_joined')
+
+    enterprise_user_id = serializers.SerializerMethodField()
+    enterprise_sso_uid = serializers.SerializerMethodField()
+
+    def get_enterprise_user_id(self, obj):
+        """
+        Get enterprise user id from user object.
+
+        Arguments:
+            obj (User): Django User object
+
+        Returns:
+            (int): Primary Key identifier for enterprise user object.
+        """
+        try:
+            enterprise_learner = EnterpriseCustomerUser.objects.get(user_id=obj.id)
+        except EnterpriseCustomerUser.DoesNotExist:
+            return
+
+        return enterprise_learner.id
+
+    def get_enterprise_sso_uid(self, obj):
+        """
+        Get enterprise SSO UID.
+
+        Arguments:
+            obj (User): Django User object
+
+        Returns:
+            (str): string containing UUID for enterprise customer's Identity Provider.
+        """
+        try:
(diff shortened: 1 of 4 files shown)


integrated_channels/xapi/serializers.py
```diff
@@ -4,3 +4,56 @@
...
+class CourseInfoSerializer(serializers.Serializer):
+    """
+    Serializer for course info.
+
+    Model: CourseOverview
+    """
+    course_id = serializers.CharField(source='id')
+    course_title = serializers.CharField(source='display_name')
+    course_description = serializers.CharField(source='short_description')
+    course_details_url = serializers.CharField(source='marketing_url')
```
brittneyexline: would we ever want to send the marketing url, since that will point to the B2C landing pages?

saleem-latif: According to my understanding, the course-details page is the same for both B2B and B2C scenarios. This data will be sent as an extension to X-API object field that contains course details and will only be used by enterprise customer trying to get more info about a course. I think marketing URL is better than enterprise landing page. Let know your thoughts?

saleem-latif:
@brittneyexline , @georgebabey Please take another look

integrated_channels/xapi/serializers.py
georgebabey: I think the marketing page for now makes the most sense. Maybe someday we'll have b2b marketing pages in the portal, but for now this is the best place to get course details.

integrated_channels/xapi/serializers.py
```diff
@@ -0,0 +1,87 @@
...
+    course_effore = serializers.CharField(source='effort')
```
georgebabey: typo: `effort`

tests/test_integrated_channels/test_xapi/test_serializers.py
```diff
@@ -0,0 +1,105 @@
...
+            start=now,
+            end=now + timedelta(weeks=3, days=4),
+        )
+        self.course_overview = mock.Mock(**self.course_overview_mock_data)
+
+        self.expected_data = {
+            'course_description': self.course_overview_mock_data['short_description'],
+            'course_title': self.course_overview_mock_data['display_name'],
+            'course_duration': self.course_overview_mock_data['end'] - self.course_overview_mock_data['start'],
+            'course_effore': self.course_overview_mock_data['effort'],
```
georgebabey: same as above, typo: `effort`

integrated_channels/xapi/serializers.py
```diff
@@ -0,0 +1,87 @@
...
+
+        Returns:
+            (str): string containing UUID for enterprise customer's Identity Provider.
+        """
+        try:
+            enterprise_learner = EnterpriseCustomerUser.objects.get(user_id=obj.id)
+        except EnterpriseCustomerUser.DoesNotExist:
+            return
+
+        return enterprise_learner.enterprise_customer.identity_provider
```
brittneyexline: is this meant to get the learner's id passed through SSO from the IDP, or the slug of the IDP? the field name suggests the former, but the implementation and the comment suggest the latter.

georgebabey: That list comes from the enrollment data reports and the `enterprise_sso_uid` has always referred to the user id. See https://github.com/edx/edx-documentation/blob/dccdfb3ff9053967415c00c855091fd6b026d65b/en_us/enterprise_api/source/api_reference/reference.rst

Brittney is correct in that this shouldn't be pointing to the slug, but instead `enterprise_learner.get_remote_id()`

saleem-latif:
@georgebabey @brittneyexline I have incorporated your feedback, please take another look.

PR #346, opened 2018-07-13
Fix content filter default value not displaying in EnterpriseCustomer…

mushtaqak:
Fix content filter default value not displaying in EnterpriseCustomer admin form.

This PR adds the capability to display content filter default value to inline enterprise customer catalog admin form. 

[ENT-1099](https://openedx.atlassian.net/browse/ENT-1099)


mushtaqak:
Verified on business sandbox: https://business.sandbox.edx.org/admin/enterprise/enterprisecustomer/b295df22-11c9-4786-9ba5-b76f876b77d2/change/

PR #351, opened 2018-07-25
ENT-1125 Adjustments/additions to enterprise api around django group access

brittneyexline:
**Description:**

**JIRA:** https://openedx.atlassian.net/browse/ENT-1125

**Dependencies:** n/a

**Merge deadline:** ASAP

**Testing instructions:** 
Deployed on the business sandbox.

Using a staff user, hit the following endpoint: https://business.sandbox.edx.org/enterprise/api/v1/enterprise-customer/with_access_to?permissions=enterprise_enrollment_api_access
A list of all the enterprises should get returned.

Using a non-staff, enterprise "admin" user that is part of the enterprise_enrollment_api_access group, hit the same endpoint. You should only get only the enterprise that is linked to the user back.

If you use a non-staff user who is not linked to any enterprise but happens to have group membership, the result should be an empty list of enterprises.

If you use a non-staff user that is linked but is not part of the group, you should get an authorization related error message.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/api/v1/permissions.py
+++ b/enterprise/api/v1/permissions.py
@@ -18,10 +18,13 @@ class IsAdminUserOrInGroup(permissions.IsAdminUser):
     def has_permission(self, request, view):
         """
         Returns True if the requesting user is either 'staff' or belong to specific group, False otherwise.
+        It will also check for group membership against a list of groups in the permissions query parameter.
         """
-        return super(IsAdminUserOrInGroup, self).has_permission(request, view) or request.user.groups.filter(
-            name__in=self.ALLOWED_API_GROUPS
-        ).exists()
+        return (
+            super(IsAdminUserOrInGroup, self).has_permission(request, view) or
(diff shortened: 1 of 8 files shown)


enterprise/api/v1/permissions.py
```diff
@@ -19,9 +19,11 @@ def has_permission(self, request, view):
         """
         Returns True if the requesting user is either 'staff' or belong to specific group, False otherwise.
         """
-        return super(IsAdminUserOrInGroup, self).has_permission(request, view) or request.user.groups.filter(
-            name__in=self.ALLOWED_API_GROUPS
-        ).exists()
+        return (
```
georgebabey: Probably worth updating the docstring to indicate the group check for values in the request query_params.permissions

PR #354, opened 2018-08-07
ENT-1135 Move fields from DegreedGlobalConfiguration to DegreedEnterpriseCustomerConfig

brittneyexline:
**Description:** Some of the fields we placed on the global configuration can change from customer to customer, so they need to be moved to the customer specific configuration.

**JIRA:** https://openedx.atlassian.net/browse/ENT-1135

**Dependencies:** n/a

**Merge deadline:** ASAP

**Testing instructions:**
On the sandbox, change the catalog query of the Degreed catalog, so that a different set of courses are in the catalog. Check the integrated_channel_contentmetadataitemtransmission table to see what content has been transmitted already to Degreed.

Run the transmit course metadata command:
`$ ./manage.py lms --settings=aws transmit_content_metadata --catalog_user=staff`

Inspect the integrated_channel_contentmetadataitemtransmission table to see that courses have been added or removed as expected.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/integrated_channels/degreed/models.py
+++ b/integrated_channels/degreed/models.py
@@ -29,12 +29,6 @@ class DegreedGlobalConfiguration(ConfigurationModel):
     The global configuration for integrating with Degreed.
     """
 
-    degreed_base_url = models.CharField(
-        max_length=255,
-        verbose_name="Degreed Base URL",
-        help_text="The base URL used for API requests to Degreed, i.e. https://degreed.com."
-    )
-
     completion_status_api_path = models.CharField(
         max_length=255,
         verbose_name="Completion Status API Path",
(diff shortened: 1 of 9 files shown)


integrated_channels/degreed/models.py
```diff
@@ -132,6 +99,40 @@ class DegreedEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfigurati
...
+        verbose_name="Degreed User Password",
+        help_text=(
+            "The Degreed User Password provided to the content provider by Degreed. "
+            "It is required for getting the OAuth access token."
+        )
+    )
+
+    provider_id = models.CharField(
+        max_length=100,
+        default='EDX',
```
mattdrayer: Not sure we should have a default here, from an Open edX perspective -- it probably should just be a field that accepts null values.

PR #361, opened 2018-08-30
ENT-769 Remove SailThru Flags

asadiqbal08:
**Description:** Remove the SailThru Flags for enterprise learner. 

**JIRA:** [ENT-769](https://openedx.atlassian.net/browse/ENT-769)

**Testing Instructions:**

1. Go to login page and signin with third party auth e.g. TestShib 
2. Go to enterprise customer -> [Manage learner screen](https://business.sandbox.edx.org/admin/enterprise/enterprisecustomer/2f4235e7-c3e9-40ec-8d91-a7af697e4ea8/manage_learners) in Django admin. You see the user (that linked above) is linked with the Enterprise customer. 
3. After signing with identity provider, Go to user linked account screen and you can see user is linked with the provider account. 
4. In that screen, click on the "Unlink This Account" button and visit the enterprise customer -> Manage learner screen in Django admin again. You will see the user is no more linked with that Enterprise Customer. We are following the same approach when user [click on the "unlink" button in Django admin manage user screen.](https://business.sandbox.edx.org/admin/enterprise/enterprisecustomer/2f4235e7-c3e9-40ec-8d91-a7af697e4ea8/manage_learners)
5. Also verify that, SailThru flags should be removed for that learner. 
##### NOTE: SailThru Flags: 

- is_enterprise_learner should be False
- EnterpriseCustomer should be None for that learner.

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -55,6 +55,11 @@ from enterprise.utils import (
 )
 from enterprise.validators import validate_image_extension, validate_image_size
 
+try:
+    from lms.djangoapps.email_marketing.tasks import update_user
+except ImportError:
+    update_user = None
+
 LOGGER = getLogger(__name__)
 
 mark_safe_lazy = lazy(mark_safe, six.text_type)  # pylint: disable=invalid-name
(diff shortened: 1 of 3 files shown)


enterprise/models.py
```diff
@@ -473,6 +478,17 @@ def unlink_user(self, enterprise_customer, user_email):
             # not capturing DoesNotExist intentionally to signal to view that link does not exist
             link_record = self.get(enterprise_customer=enterprise_customer, user_id=existing_user.id)
             link_record.delete()
+
+            # Remove the SailThru flags for enterprise learner.
+            if update_user:
```
saleem-latif: I think it would be better to add a check for `update_user` and raise `NotConnectedToOpenEdX` instead of this condition.

```
if update_user is None:
      raise NotConnectedToOpenEdX("This package must be installed in an OpenEdX environment.")
```

PR #362, opened 2018-09-10
ENT-1211: Fixed 500 error on enterprise customer screen.

saleem-latif:
**Description:** 
This PR fixes the 500 error on enterprise customer admin screen.

I have moved the imports from top to inside method definition because otherwise `ImportError` is raised and that causes the issues we are seeing.

**JIRA:** [ENT-1211](https://openedx.atlassian.net/browse/ENT-1211)

**Installation instructions:** Simply install edx-enterprise through this branch.

**Testing instructions:**

1. Open enterprise customer details page.
2. Try to update "ENTERPRISE CUSTOMER IDENTITY PROVIDERS"
3. Make sure dropdown menu is displayed instead of the text field
4. Make sure adding/updating identity provider is not throwing errors.

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

**Author concerns:** This is a quick fix to the CAT-2 bug, We need to make sure why are the imports failing on the top.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[0.73.2] - 2018-09-11
+---------------------
+
+* Fixed 500 error on enterprise customer admin screen.
+
 [0.73.1] - 2018-08-30
 ---------------------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "0.73.1"
+__version__ = "0.73.2"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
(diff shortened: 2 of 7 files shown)


saleem-latif:
@mattdrayer adding logging when import fails, please take another look.

mattdrayer:
Thanks 👍 -- maybe we should keep the module-level import at the top and similarly log the exception if/when it fails, and also add the method-level imports + logging that you've introduced in this PR.  That way we may gain some insight into the original problem as well as get around the error condition for the short-term.

saleem-latif:
@mattdrayer Good Idea, I have added the top import along with the logging if there is import error.
@georgebabey, @zubair-arbi Can you please review this PR. Since this is an urgent bug I think we should get it merged if it looks good?

saleem-latif:
Removed tests were testing a method that calls LMS method from `Registry` and `Registry` was mocked. Now, with imports moved inside the method, there was no way to mock out the `Registry`, only another way was to mock the method which would have muted the purpose of the test. So, I thought the removing tests was the only viable option.

PR #365, opened 2018-09-12
Added the ability to unenroll in the enrollment API

davestgermain:
**Description:** Adds an optional field, ``is_active`` to support unenrollment in the course-enrollments endpoint.

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -558,6 +558,7 @@ class EnterpriseCustomerCourseEnrollmentsSerializer(serializers.Serializer):
     )
     email_students = serializers.BooleanField(default=False, required=False, write_only=True)
     detail = serializers.CharField(read_only=True)
+    is_active = serializers.BooleanField(required=False, default=True, write_only=True)
 
     def create(self, validated_data):
         """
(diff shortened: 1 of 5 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -570,18 +571,23 @@ def create(self, validated_data):
...
+                    enterprise_customer_user.enroll(course_run_id, course_mode)
+                else:
+                    enterprise_customer_user.unenroll(course_run_id)
             except (utils.CourseEnrollmentDowngradeError, HttpClientError) as exc:
                 validated_data['detail'] = str(exc)
                 return validated_data
 
-            track_enrollment('enterprise-customer-enrollment-api', enterprise_customer_user.user_id, course_run_id)
+            if is_active:
+                track_enrollment('enterprise-customer-enrollment-api', enterprise_customer_user.user_id, course_run_id)
```
brittneyexline: it might be good to track unenrollments as well

enterprise/api_client/lms.py
```diff
@@ -236,6 +236,26 @@ def enroll_user_in_course(self, username, course_id, mode):
...
+    def unenroll_user_from_course(self, username, course_id):
+        """
+        Call the enrollment API to unenroll the user in the course specified by course_id.
+        Args:
+            username (str): The username by which the user goes on the OpenEdx platform
+            course_id (str): The string value of the course's unique identifier
+        Returns:
+            bool: Whether the unenrollment succeeded
+        """
+        enrollment = self.get_course_enrollment(username, course_id)
```
brittneyexline: is there a reason we need to call the api to check if the enrollment exists before unenrolling? how does the enrollment api handle a call to unenroll someone from a course they're not enrolled in? i think it would be better to make one call instead of two.

enterprise/models.py
```diff
@@ -689,6 +689,16 @@ def enroll(self, course_run_id, mode):
                 )
             )
 
+    def unenroll(self, course_run_id):
+        """
+        Unenroll a user from a course track.
+        """
+        enrollment_api_client = EnrollmentApiClient()
+        if enrollment_api_client.unenroll_user_from_course(self.username, course_run_id):
+            EnterpriseCourseEnrollment.objects.filter(enterprise_customer_user=self, course_id=course_run_id).delete()
```
brittneyexline: my only concern with this is if for some reason we had an enterprise course enrollment record but no enrollment record. that state shouldn't be possible to get into, so this is probably fine.

PR #368, opened 2018-10-02
ENT-1199: SuccessFactors: Submit batch/chunk of OCN items to tenants until error status

saleem-latif:
**Description:** 
Currently, the edX system breaks up the set of OCN items into the specified batch/chunk size defined in a customer's SAPSuccessFactorsEnterpriseCustomerConfiguration record. The system then submits "total records/batch size" number of transactions to the SuccessFactors Learning API, one after the other. For example, given a total set of 100 records, and a batch size of 20, the edX system would submit 5 requests to the API.
Unfortunately, the SuccessFactors Learning API has a throttle constraint which currently allows for only one API request in a given cycle. So in the example above, the first request would be accepted, and the remaining four would be denied.
We keep trying to submit until we get a rate error exceeded or any error status and then will stop

**JIRA:** [ENT-1199](https://openedx.atlassian.net/browse/ENT-1199)

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/integrated_channels/sap_success_factors/transmitters/content_metadata.py
+++ b/integrated_channels/sap_success_factors/transmitters/content_metadata.py
@@ -40,31 +40,55 @@ class SapSuccessFactorsContentMetadataTransmitter(ContentMetadataTransmitter):
         prepared_items.update(items_to_update)
         prepared_items.update(items_to_delete)
 
+        skip_metadata_transmission = False
+
         for chunk in chunks(prepared_items, self.enterprise_configuration.transmission_chunk_size):
             chunked_items = list(chunk.values())
-            try:
-                self.client.update_content_metadata(self._serialize_items(chunked_items))
-            except ClientError as exc:
-                LOGGER.error(
-                    'Failed to update [%s] content metadata items for integrated channel [%s] [%s]',
-                    len(chunked_items),
-                    self.enterprise_configuration.enterprise_customer.name,
-                    self.enterprise_configuration.channel_code,
-                )
-                LOGGER.error(exc)
-
+            if skip_metadata_transmission:
                 # Remove the failed items from the create/update/delete dictionaries,
                 # so ContentMetadataItemTransmission objects are not synchronized for
                 # these items below.
-                for item in chunked_items:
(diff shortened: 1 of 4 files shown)


integrated_channels/sap_success_factors/transmitters/content_metadata.py
```diff
@@ -62,6 +62,10 @@ def transmit(self, payload, **kwargs):
                     items_to_update.pop(content_metadata_id, None)
                     items_to_delete.pop(content_metadata_id, None)
 
+                # SAP servers throttle incoming traffic, If a request fails than the subsequent would fail too,
+                # So, no need to keep trying and failing. We should stop here and retry later.
+                break
```
brittneyexline: glad that this is simple! however, we need to update items_to_create, items_to_update, and items_to_delete to reflect what items have actually been sent to the api, because below we commit them to the db. with this, the calls that get skipped after an error would be recorded as having been sent.

saleem-latif:
@brittneyexline I haved update the PR according to your feedback, please take another look.

tests/test_integrated_channels/test_sap_success_factors/test_transmitters/test_content_metadata.py
```diff
@@ -73,6 +78,10 @@ def test_transmit_create_failure(self, update_content_metadata_mock):
                 content_id: ContentMetadataItemExport(
                     {'key': content_id, 'content_type': 'course'},
                     {'courseID': content_id, 'update': True}
+                ),
+                content_id_2: ContentMetadataItemExport(
+                    {'key': content_id_2, 'content_type': 'course'},
+                    {'courseID': content_id_2, 'update': True}
```
brittneyexline: can you add a line to this test below that also asserts that there is no transmission object for content_id_2?

saleem-latif: Done

PR #369, opened 2018-10-02
WIP --- ENT-1200

muhammad-ammar:
**Description:** Support large catalog queries

**JIRA:** https://openedx.atlassian.net/browse/ENT-1200

**Dependencies:** https://github.com/edx/course-discovery/pull/1575 

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


muhammad-ammar:
@georgebabey This is how I am planning to do the things. Here is related [course-discovery PR](https://github.com/edx/course-discovery/pull/1575).

muhammad-ammar:
@georgebabey @zubair-arbi @mushtaqak This is ready for review.

enterprise/api_client/discovery.py
```diff
@@ -121,6 +121,29 @@ def get_search_results(self, querystring=None, traverse_pagination=True):
...
+
+        Returns:
+            dict: The paginated response.
+        """
+        response = {}
+
+        try:
+            endpoint = getattr(self.client, self.SEARCH_ALL_ENDPOINT)
+            response = endpoint().post(query)
+        except Exception as ex:  # pylint: disable=broad-except
```
zubair-arbi: Instead of a broad exception lets short list it, e.g.`except (SlumberBaseException, ConnectionError, Timeout) as ex`

enterprise/api_client/discovery.py
```diff
@@ -121,6 +121,29 @@ def get_search_results(self, querystring=None, traverse_pagination=True):
             many=False,
         )
 
+    def get_catalog_results(self, query):
```
zubair-arbi: Instead of using a different method name how about something like `get_search_results_with_post_method` (like method above `get_search_results`)?

enterprise/api_client/discovery.py
```diff
@@ -121,6 +121,29 @@ def get_search_results(self, querystring=None, traverse_pagination=True):
...
+        """
+        Return results from the discovery service's search/all endpoint.
```
zubair-arbi: Return results from the discovery service's search/all endpoint using POST method.

enterprise/api_client/discovery.py
muhammad-ammar: We are already doing the same thing in edx-platform [here](https://github.com/edx/edx-platform/blob/master/openedx/core/lib/edx_api_utils.py#L74)

enterprise/api_client/discovery.py
muhammad-ammar: Getting data using HTTP POST is the internal detail, I would prefer to name the method as it is.

enterprise/api_client/discovery.py
```diff
@@ -121,6 +121,29 @@ def get_search_results(self, querystring=None, traverse_pagination=True):
...
+            query (dict): query parameters used to filter search results.
+
+        Returns:
+            dict: The paginated response.
+        """
+        response = {}
+
+        try:
+            endpoint = getattr(self.client, self.SEARCH_ALL_ENDPOINT)
+            response = endpoint().post(query)
```
brittneyexline: I know I'm jumping in a little randomly here, but I discovered a bug on the business sandbox where our enterprise catalog api is not properly handling pagination - it always returns the first page of results no matter what page you pass in. After some digging, I ended up here. We need to be able to pass pagination params to the discovery service in order to support pagination for the enterprise api - I believe the function above (`get_search_results`) does this, so that needs to be replicated here somehow. Let me know if you have any questions!

muhammad-ammar: @brittneyexline Thanks a lot for reporting the bug and sorry for any inconvenience caused by this. I will fix the bug.

muhammad-ammar:
@brittneyexline @georgebabey This is ready for review.

@brittneyexline The 500 was coming due to the original issue as I reverted my changes when you found the pagination issue. Now it should be fixed.

PR #372, opened 2018-10-15
Prevent access to auto-cohorting API unless the feature is enabled on the customer level

davestgermain:
**Description:** 

We want to enable auto-cohorting in the enrollment API only for certain organizations. This adds a new field on EnterpriseCustomer to support toggling the feature.

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


enterprise/migrations/0028_link_enterprise_to_enrollment_template.py
```diff
@@ -15,7 +15,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='enrollmentnotificationemailtemplate',
             name='enterprise_customer',
-            field=models.OneToOneField(related_name='enterprise_enrollment_template', default=None, to='enterprise.EnterpriseCustomer'),
+            field=models.OneToOneField(related_name='enterprise_enrollment_template', default=None, to='enterprise.EnterpriseCustomer', on_delete=models.CASCADE),
```
georgebabey: Were these old migrations intentionally updated?

davestgermain: Yes, Django was throwing deprecation warnings for each migration, creating a lot of noise.

georgebabey: Thanks for eliminating the noise. I'm not sure we want to change these to cascading though. It looks like all of our deletes are set to RESTRICT, so I think `models.PROTECT` is what we want although I'm no expert here...
```
mysql> SELECT DELETE_RULE, TABLE_NAME, REFERENCED_TABLE_NAME from information_schema.REFERENTIAL_CONSTRAINTS where TABLE_NAME='enterprise_enrollmentnotificationemailtemplate';
+-------------+------------------------------------------------+-------------------------------+
| DELETE_RULE | TABLE_NAME                                     | REFERENCED_TABLE_NAME         |
+-------------+------------------------------------------------+-------------------------------+
| RESTRICT    | enterprise_enrollmentnotificationemailtemplate | enterprise_enterprisecustomer |
+-------------+------------------------------------------------+-------------------------------+
1 row in set (0.01 sec)
```

I vaguely remember someone saying edX shys away from cascading deletes, maybe for unexpected performance implications but I don't entirely remember the rationale. @feanil or @nasthagiri might know?

mattdrayer: Is it because we like to have lots of orphaned child records with no clear parent record?  😉

enterprise/models.py
```diff
@@ -186,6 +187,13 @@ class Meta:
         )
     )
 
+    enable_autocohorting = models.BooleanField(
+        default=False,
+        help_text=_(
+            "Specifies whether the customer is able to assign learners to cohorts upon enrollment."
+        )
+    )
```
nasthagiri: It seems odd to me that we are putting this configuration setting into our data model.  Do we plan to have a new column for each new type of access that we want to toggle for each enterprise customer?

Other alternatives to consider:

- **Django waffle flag**  We already have a [CourseWaffleFlag](https://github.com/edx/edx-platform/blob/0c738ee43cba6a35cd699cb6cec761a5432a5db5/openedx/core/djangoapps/waffle_utils/__init__.py#L343) that allows toggling on a per-course basis.  We can add an `OrganizationWaffleFlag` for this and other related needs.
- [ConfigurationModel](https://github.com/edx/django-config-models) - @cpennington has talked about extending this to support Organization-level configurations since the Revenue squad has a need for it.
- OAuth Scopes - If we view this as a permissioning/access thing rather than a configuration setting, then we should use Scopes.  Scopes can be granular if needed.

georgebabey: I read this more as a configuration on the enterprise that allows for auto-cohoriting. I agree the help_text makes it sound more like a permission check.

Some thoughts on the alternatives, and why we have an enterprise configuration model in the first place:
- Waffle flags: Theres are fine for boolean on/off switches, complex configuration like relating two models seems best fit in django admin. Also because our internal CS team maintains these configurations, having the option to override the django admin panels is useful from a usability standpoint.
- Organization ConfigurationModel - Organization != Enterprise. Maybe it should, but it doesn't today. 
- OAuth Scopes - if this is an access-level check than scopes may make sense but my understanding is this is a per-enterprise configuration that will be checked not for access, but during the registration flow to make sure they end up in a proper cohort.

@nasthagiri Can you speak at all to the original question around cascading deletes?

nasthagiri: My question here stems from my understanding of the business need for this toggle.  In my conversation with @scottrish, he mentions that they are "restricting access to an API for a capability that is in development.
So incrementally adding to a v2, but keeping v2 in an unpublished state and available only to ‘beta’ clients... until such time that we build out ‘enterprise cohorts’ we need to restrict access to this capability."

In my mind, that translates to a need for incrementally rolling out and beta-testing a feature.  This differs from a persistent configuration setting applied to individual enterprises.

I'll respond to the cascading deletes on that thread.

enterprise/migrations/0028_link_enterprise_to_enrollment_template.py
nasthagiri: I'm not aware of our standpoint on cascading deletes.  In addition to devops' concerns, there may be considerations from the data-analytics perspective as well.

Regardless of that best practice (which I agree we should define and clarify), I caution us from updating past migrations, especially in this way, that we have already run on our servers.  Since we will not run these old migrations on our servers and on pre-existing data, it may cause issues on other open edX instances (or even our own servers if we ever need to migrate back).

The Django-upgrade team may shed further light on this if we really need to address this noise at this time.

enterprise/models.py
davestgermain: @nasthagiri Are you suggesting to increment the version on whole enterprise API to v2, and gating v2 to certain enterprises (in which case, you still need a place to keep track of that setting)? That sounds very involved.

davestgermain: @nasthagiri That sounds ok, except for the fact that edx-enterprise does not depend on edx-platform, so it technically wouldn't have access to ``WaffleFlagNamespace`` (for tests).

nasthagiri: @davestgermain Two things to note:
* waffle_utils needs to be moved out of edx-platform, as we need a consistent way to toggle our features across our IDAs, per [OEP-17](https://open-edx-proposals.readthedocs.io/en/latest/oep-0017-bp-feature-toggles.html).  We've already created an [edx-toggles](https://github.com/edx/edx-toggles) repo that is intended for this code.  We just need a team to invest in that transition.
* As of today, edx-enterprise _does_ depend on edx-platform.  If you search for "from openedx.core" in the repo, you'll see occurrences of such imports ([example](https://github.com/edx/edx-enterprise/blob/8cf47ff167e3eb549a040b56d0f3f54e05d42baa/enterprise/api_client/lms.py#L30)).

davestgermain: As far as I know, edx-platform is not in the requirements files for edx-enterprise. Code which imports openedx is not exercised during tests. I could make a conditional import of waffle_utils, but it won't be testable...

nasthagiri: They are tested in end-to-end tests in edx-platform.  The enterprise team will know more.  Maybe in [enterprise_support](https://github.com/edx/edx-platform/tree/master/openedx/features/enterprise_support).

nasthagiri: @davestgermain @georgebabey Following up on this thread...
I spoke to @scottrish yesterday and although he would prefer a solution for temporary configuration/rollout that does _not_ require updating this enterprise table (he brought up this concern in the 1st place :D), other viable options set forth thus far jeopardize the team hitting their end-of-year deadline.  He plans to have the team revisit the entire cohorting solution for schools/enterprises post-deadline.  By then, I am also hoping that our recent initiative around Enterprise product vision will result in us architecting more intentional data models for "organizations" in our system.

Given that, you may want to update a docstring/comment to clarify the nature and temporality of the new setting and may-the-force-be-with-you!

davestgermain: @nasthagiri @georgebabey I've added the comment and also reverted all of the old migration changes. I haven't gotten a definitive answer about the intended default behavior of ``on_delete``, so let's just defer that to another time.

PR #376, opened 2018-10-16
WIP |Bexline/all courses for enterprise

christopappas:

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


enterprise/api/v2/views.py
```diff
@@ -0,0 +1,97 @@
...
+            update_content_filters(combined_content_filter, catalog.content_filter)
+
+        ######################### CONSTRUCTION ZONE #########################
+        # - Now that we have combined_content_filter, do something with it here
+        # - Call the discovery service using it
+        # - Get the results and serialize it
+        # (Looks like you can use something like get_catalog_courses, except while
+        #  accepting a query string. Need to figure out how to take this dict we created
+        #  and turn it into the proper datatype for the `querystring` variable)
+        catalog_api = CourseCatalogApiClient(request.user, enterprise_customer.site)
```
brittneyexline: so my original thought here was to use the `EnterpriseCustomerCatalogDetailSerializer` to both make the call and serialize the results by instantiating an EnterpriseCustomerCatalog with the new filter but not committing it to the database. I don't know 100% if that will work, but it would be preferable. in any case, the goal would be to reuse as much logic as possible for fetching and serializing responses from the discovery service search endpoint.

enterprise/api/v2/views.py
```diff
@@ -0,0 +1,97 @@
...
+def update_content_filters(combined_content_filter, new_content_filter):
+    """
+    Helper method for combining 2 content filter dicts
+    """
+    for filter_key, filter_value in new_content_filter.items():
+        if filter_key in combined_content_filter:
+            old_value = combined_content_filter[filter_key]
+            if isinstance(filter_value, list):
+                combined_content_filter[filter_key] = list(set(old_value) | set(filter_value))
+            elif filter_value != old_value:
```
brittneyexline: you probably already are planning to do this, but there are some keys that creating a list won't work for, in particular, `date_lte` sort of fields. just a note to account for that 😄 
i also think this should function should live in enterprise/api/utils.py

PR #377, opened 2018-10-29
ENT-1294:  Fix translations for enterprise pages.

saleem-latif:
**Description:** Enterprise Pages are not translating text according to user preferences. This PR fixes that.

**JIRA:** [ENT-1294](https://openedx.atlassian.net/browse/ENT-1294)

**Installation instructions:** Simply install this version of edx-enterprise into your LMS.

**Testing Instructions:**

- Go to enterprise landing page [here](https://business.sandbox.edx.org/enterprise/d583b097-7bd9-4a7e-a98d-c15dea220354/course/course-v1%3AedX%2BDemoX%2BDemo_Course/enroll/?new_enterprise_login=yes&tpa_hint=saml-bestrun)
- Login with the following credentials `edx_learner497:test123test`
- Make sure available translations are appearing
- If translations are not appearing make sure `Spanish` is selected as language in the [account settings page](https://business.sandbox.edx.org/account/settings)

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -161,68 +161,6 @@ class GrantDataSharingPermissions(View):
     consent to be provided, and a POST view that consumes said form.
     """
 
-    page_title = _('Data sharing consent required')
-    consent_message_header = _('Consent to share your data')
-    requested_permissions_header = _(
-        'Per the {start_link}Data Sharing Policy{end_link}, '
-        '{bold_start}{enterprise_customer_name}{bold_end} would like to know about:'
-    )
-    agreement_text = _(
-        'I agree to allow {platform_name} to share data about my enrollment, completion and performance '
-        'in all {platform_name} courses and programs where my enrollment is sponsored by {enterprise_customer_name}.'
-    )
-    continue_text = _('Yes, continue')
-    abort_text = _('No, take me back.')
-    policy_dropdown_header = _('Data Sharing Policy')
-    sharable_items_header = _(
-        'Enrollment, completion, and performance data that may be shared with {enterprise_customer_name} '
-        '(or its designee) for these courses and programs are limited to the following:'
-    )
-    sharable_items = [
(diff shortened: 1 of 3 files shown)


enterprise/views.py
```diff
@@ -238,48 +176,97 @@ def get_default_context(self, enterprise_customer, platform_name):
...
                     platform_name=platform_name
-                ) for item in self.sharable_items
+                ),
+                _(
+                    'My {platform_name} ID, and if I log in via single sign-on, '
+                    'my {enterprise_customer_name} SSO user-ID'
+                ).format(
+                    platform_name=platform_name,
+                    enterprise_customer_name=enterprise_customer.name,
+
```
brittneyexline: nit: remove this extra new line

enterprise/views.py
```diff
@@ -1265,21 +1201,37 @@ def get_enterprise_program_enrollment_page(self, request, enterprise_customer, p
         elif discount_data.get('total_incl_tax_excl_discounts') is None:
             messages.add_missing_price_information_message(request, program_title)
 
-        # Update our context with the above calculated details and more.
-        context_data = self.context_data.copy()
```
brittneyexline: did you intend to remove this line?

saleem-latif: yes, this is intentional, translations (in order to work) needed to be moved into the method instead of being defined as the class attribute.

saleem-latif:
I have incorporated your feedback, please take another look.

If translations are defined as class attributes then the call to `ugettext` will not be processed for each request rather it will be processed when class object is created. 
In future, we should make sure that translation function gets called each time the request is being processed for the translation to work correctly. i.e. avoid defining translation as class attributes etc.

PR #378, opened 2018-11-01
ENT-1169 Request codes api

irfanuddinahmad:
**Description:** An API to “Request More Codes”. This will trigger a notification to customersuccess@edx.org

**JIRA:** https://openedx.atlassian.net/browse/ENT-1169

**Testing instructions:**

1. Use any REST client to POST to ... enterprise/api/v1/request_codes
2. Test with and without the following json object
    {
      "email": "bob@alice.com",
      "enterprise_name": "IBM",
      "number_of_codes": "50"
    }
3. Expect a HTTP 200 response for correct input and email to customersuccess@edx.org
4. Expect a HTTP 400 response for incorrect input.
5. Expect a HTTP 500 response for an email send issue.

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -5,6 +5,7 @@ Views for enterprise api version 1 endpoint.
 from __future__ import absolute_import, unicode_literals
 
 from logging import getLogger
+from smtplib import SMTPException
 
 from edx_rest_framework_extensions.auth.bearer.authentication import BearerAuthentication
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
@@ -14,13 +15,17 @@ from rest_framework.decorators import detail_route, list_route
 from rest_framework.exceptions import NotFound
 from rest_framework.renderers import JSONRenderer
 from rest_framework.response import Response
-from rest_framework.status import HTTP_200_OK, HTTP_400_BAD_REQUEST
+from rest_framework.status import HTTP_200_OK, HTTP_400_BAD_REQUEST, HTTP_500_INTERNAL_SERVER_ERROR
+from rest_framework.views import APIView
 from rest_framework_xml.renderers import XMLRenderer
 from six.moves.urllib.parse import quote_plus, unquote  # pylint: disable=import-error,ungrouped-imports
 
+from django.apps import apps
 from django.conf import settings
+from django.core import mail
 from django.http import Http404
 from django.utils.decorators import method_decorator
+from django.utils.translation import ugettext as _
 
 from enterprise import models
 from enterprise.api.filters import EnterpriseCustomerUserFilterBackend, UserFilterBackend
@@ -28,9 +33,15 @@ from enterprise.api.pagination import get_paginated_response
 from enterprise.api.throttles import ServiceUserThrottle
 from enterprise.api.v1 import serializers
 from enterprise.api.v1.decorators import enterprise_customer_required, require_at_least_one_query_parameter
-from enterprise.api.v1.permissions import HasEnterpriseEnrollmentAPIAccess, IsInEnterpriseGroup
+from enterprise.api.v1.permissions import (
+    HasEnterpriseDataAPIAccess,
+    HasEnterpriseEnrollmentAPIAccess,
+    IsInEnterpriseGroup,
+)
 from enterprise.api_client.discovery import CourseCatalogApiClient
 from enterprise.constants import COURSE_KEY_URL_PATTERN
+from enterprise.errors import CodesAPIRequestError
+from enterprise.utils import get_request_value
 
 LOGGER = getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


enterprise/api/v1/views.py
```diff
@@ -510,3 +515,94 @@ class EnterpriseCustomerReportingConfigurationViewSet(EnterpriseReadOnlyModelVie
...
+        :param request: The request to this endpoint.
+        :return: The ``email``, ``enterprise_name``, and ``number_of_codes`` from the request.
+        """
+        email = get_request_value(request, self.REQUIRED_PARAM_EMAIL, '')
+        enterprise_name = get_request_value(request, self.REQUIRED_PARAM_ENTERPRISE_NAME, '')
+        number_of_codes = get_request_value(request, self.OPTIONAL_PARAM_NUMBER_OF_CODES, '')
+        if not (email and enterprise_name):
+            raise CodesAPIRequestError(
+                self.get_missing_params_message([
+                    ("'email'", bool(email)),
```
georgebabey: Rather than the hardcoded string I'd recommend using REQUIRED_PARAM_EMAIL, REQUIRED_PARAM_ENTERPRISE_NAME you set above

enterprise/api/v1/views.py
```diff
@@ -510,3 +515,94 @@ class EnterpriseCustomerReportingConfigurationViewSet(EnterpriseReadOnlyModelVie
...
+        """
+        try:
+            email, enterprise_name, number_of_codes = self.get_required_query_params(request)
+        except CodesAPIRequestError as invalid_request:
+            return Response({'error': str(invalid_request)}, status=HTTP_400_BAD_REQUEST)
+
+        subject_line = _('Code Management - Request for more Codes by {token_enterprise_name}').format(
+            token_enterprise_name=enterprise_name
+        )
+        plain_msg = _('{token_email} from {token_enterprise_name} has requested for more {token_number_codes} codes.'
```
adamstankiewicz: The number of codes is an optional parameter so, we should only conditionally include it in the email that gets sent.

enterprise/api/v1/views.py
```diff
@@ -510,3 +515,94 @@ class EnterpriseCustomerReportingConfigurationViewSet(EnterpriseReadOnlyModelVie
     )
     filter_fields = FIELDS
     ordering_fields = FIELDS
+
+
+class CouponCodesView(APIView):
+    """
+    API to request coupon codes.
+    """
+    permission_classes = (permissions.IsAuthenticated,)
```
georgebabey: I think we need an extra permission class here to check for the enterprise_api_data_access group to restrict it to only users that have portal access.

I wouldn't want any user to be able to spam CS by hitting this endpoint.

enterprise/api/v1/views.py
```diff
@@ -510,3 +515,94 @@ class EnterpriseCustomerReportingConfigurationViewSet(EnterpriseReadOnlyModelVie
...
+        subject_line = _('Code Management - Request for more Codes by {token_enterprise_name}').format(
+            token_enterprise_name=enterprise_name
+        )
+        plain_msg = _('{token_email} from {token_enterprise_name} has requested for more {token_number_codes} codes.'
+                      'Please reach out to them.').format(
+                          token_email=email,
+                          token_enterprise_name=enterprise_name,
+                          token_number_codes=number_of_codes
+                      )
+        from_email_address = None
```
adamstankiewicz: Should the `from_email_address` be the email address argument passed to the endpoint? That way, the CS team can more easily reply to the email (@georgebabey thoughts? Maybe that would be a question for CS team).

enterprise/api/v1/views.py
```diff
@@ -510,3 +515,94 @@ class EnterpriseCustomerReportingConfigurationViewSet(EnterpriseReadOnlyModelVie
...
+        cs_email = _('customersuccess@edx.org')
```
georgebabey: Can we make this a configuration option - we may want to separate this based on environment (stage, edx, prod)

enterprise/api/v1/views.py
georgebabey: nit: `has requested for more {token_number_codes} codes` => `has requested for {token_number_codes} more codes`

enterprise/api/v1/views.py
```diff
@@ -510,3 +515,94 @@ class EnterpriseCustomerReportingConfigurationViewSet(EnterpriseReadOnlyModelVie
...
+        data = {
+            self.REQUIRED_PARAM_EMAIL: email,
+            self.REQUIRED_PARAM_ENTERPRISE_NAME: enterprise_name,
+            self.OPTIONAL_PARAM_NUMBER_OF_CODES: number_of_codes,
+        }
+        if mail.send_mail(
```
georgebabey: I think we need to wrap this in a try/except, set fail_silently to false, and handle SMTPException so we can capture the email/enterprise name in the error message.  I believe as it is this would swallow the errors.

tests/test_enterprise/api/test_views.py
```diff
@@ -2401,3 +2401,77 @@ def test_enterprise_customer_catalogs_response_formats(self):
 
         response_xml = self.client.get('/enterprise/api/v1/enterprise_catalogs.xml')
         self.assertTrue(response_xml['content-type'] == 'application/xml; charset=utf-8')
+
+    @mock.patch('django.core.mail.send_mail')
+    @ddt.data(
```
adamstankiewicz: Should we also include a use case for the excluding the optional `number_of_codes` parameter?

enterprise/api/v1/views.py
```diff
@@ -510,3 +515,94 @@ class EnterpriseCustomerReportingConfigurationViewSet(EnterpriseReadOnlyModelVie
...
+            The name of the enterprise requesting more codes.
+        *number_of_codes*
+            The number of codes requested.
+        """
+        try:
+            email, enterprise_name, number_of_codes = self.get_required_query_params(request)
+        except CodesAPIRequestError as invalid_request:
+            return Response({'error': str(invalid_request)}, status=HTTP_400_BAD_REQUEST)
+
+        subject_line = _('Code Management - Request for more Codes by {token_enterprise_name}').format(
```
adamstankiewicz: nit: The word "more" is unnecessary in the subject line: "`Code Management - Request for Codes by {token_enterprise_name}`"

tests/test_enterprise/api/test_views.py
irfanuddinahmad: Test added

enterprise/api/v1/views.py
irfanuddinahmad: Done

enterprise/api/v1/views.py
irfanuddinahmad: @georgebabey @jshin-edx ... the from_email_address needs to be set to some specific address?

enterprise/api/v1/views.py
irfanuddinahmad: https://github.com/edx/configuration/pull/4865

enterprise/api/v1/views.py
```diff
@@ -510,3 +521,111 @@ class EnterpriseCustomerReportingConfigurationViewSet(EnterpriseReadOnlyModelVie
...
+        """
+        try:
+            email, enterprise_name, number_of_codes = self.get_required_query_params(request)
+        except CodesAPIRequestError as invalid_request:
+            return Response({'error': str(invalid_request)}, status=HTTP_400_BAD_REQUEST)
+
+        subject_line = _('Code Management - Request for Codes by {token_enterprise_name}').format(
+            token_enterprise_name=enterprise_name
+        )
+        msg_with_codes = _('{token_email} from {token_enterprise_name} has requested for {token_number_codes} codes.'
```
adamstankiewicz: nit: Small grammar fix. 
```python
'{token_email} from {token_enterprise_name} has requested {token_number_codes} additional codes. Please reach out to them.'
```

Same fix for the `msg_without_codes` string below: 
```python
'{token_email} from {token_enterprise_name} has requested additional codes. Please reach out to them.'
```

enterprise/api/v1/views.py
georgebabey: In this case `from_email_address` can come from CS as well (customersuccess@edx.org). This way we don't need to authorize another email address with SES

adamstankiewicz: Makes sense to come from CS as well. Either way, the `from_email_address` should not be `None` as it's a required field for Django's `send_mail` method: https://docs.djangoproject.com/en/2.1/topics/email/#send-mail

enterprise/api/v1/views.py
irfanuddinahmad: fixed

enterprise/api/v1/views.py
irfanuddinahmad: from_email_address added

PR #413, opened 2018-12-10
ENT-1324 Add enterprise customer types

zubair-arbi:

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


zubair-arbi:
@mattdrayer Please let me know if we are ok with this change.

enterprise/models.py
```diff
@@ -197,6 +197,24 @@ class Meta:
         )
     )
 
+    TYPE_ENTERPRISE = 'enterprise'
+    TYPE_PARTNER = 'partner'
+    TYPE_NON_PROD = 'non_prod'
+    CUSTOMER_TYPE_CHOICES = (
+        (TYPE_ENTERPRISE, 'Enterprise Customer'),
+        (TYPE_PARTNER, 'Partner Customer'),
+        (TYPE_NON_PROD, 'Non Production Customer'),
```
asadiqbal08: The label `Customer type` is indicating the type of customers.  I think, We should eliminate the word `Customer` from the options. 

Customer Types: 
1. Enterprise
2. Partners
3. Non Production

enterprise/models.py
```diff
@@ -197,6 +197,24 @@ class Meta:
...
+    )
+    customer_type = models.CharField(
+        max_length=25,
+        blank=False,
+        choices=CUSTOMER_TYPE_CHOICES,
+        default=TYPE_ENTERPRISE,
+        help_text=_(
+            'Specifies enterprise customer type according to the Partner.'
```
asadiqbal08: please mark the strings where need to be marked for translations.

zubair-arbi: Already did.

enterprise/models.py
```diff
@@ -197,6 +197,24 @@ class Meta:
         )
     )
 
+    TYPE_ENTERPRISE = 'enterprise'
+    TYPE_PARTNER = 'partner'
+    TYPE_NON_PROD = 'non_prod'
```
asadiqbal08: A suggestion for better readability: 
```
Class CustomerType:
          """"Possible types of customers """
          ENTERPRISE = 'enterprise'
          PARTNER = 'partner'
          NON_PROD = 'non_prod'

```
You can refer anywhere in code something like that --> CustomerType.ENTERPRISE , CustomerType.PARTNER ...

zubair-arbi: I have add new choices according to current implementation and what django docs suggest: https://docs.djangoproject.com/en/2.0/ref/models/fields/#choices. Also we still should be able to access the choices from `EnterpriseCustomer` model, e.g. `EnterpriseCustomer.TYPE_ENTERPRISE`.

If we want to define the choices as a class then we have to use Enum which is not needed for our current simple usecase in my opinion:
* http://anthonyfox.io/2017/02/choices-for-choices-in-django-charfields/
* https://hackernoon.com/using-enum-as-model-field-choice-in-django-92d8b97aaa63

zubair-arbi:
@asadiqbal08 I have addressed your feedback. Please have another look.

enterprise/models.py
asadiqbal08: I suggest, using a Enum or Class would be a good approach instead of adding a pre-word `TYPE` for every choices.

asadiqbal08:
overall lgtm

enterprise/models.py
asadiqbal08: I was referring to dropdown list names --- 'enterprise' etc ...

zubair-arbi:
@mattdrayer I have address your concerns by introducing a new model for storing customer types. Please have another look so that we can merge the changes.

enterprise/models.py
zubair-arbi: We have achieved this through introducing a new model `EnterpriseCustomerType`.

mattdrayer:
Overall the implementation LGTM, but @georgebabey should have the final say on whether the model-based solution is the right approach versus straight configuration, imho.

enterprise/migrations/0058_auto_20181211_0658.py
```diff
@@ -0,0 +1,45 @@
...
+        migrations.CreateModel(
+            name='EnterpriseCustomerType',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created', model_utils.fields.AutoCreatedField(default=django.utils.timezone.now, editable=False, verbose_name='created')),
+                ('modified', model_utils.fields.AutoLastModifiedField(default=django.utils.timezone.now, editable=False, verbose_name='modified')),
+                ('name', models.CharField(help_text='Specifies enterprise customer type.', max_length=25)),
+            ],
+            options={
+                'ordering': ['created'],
```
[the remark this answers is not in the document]
zubair-arbi: Ok, I will update it.

enterprise/models.py
```diff
@@ -81,6 +81,49 @@ def get_queryset(self):
         return super(EnterpriseCustomerManager, self).get_queryset().filter(active=True)
 
 
+@python_2_unicode_compatible
+class EnterpriseCustomerType(TimeStampedModel):
+    """
+    Enterprise Customer Types are used to differentiate Enterprise learners from Partners.
```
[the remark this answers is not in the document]
zubair-arbi: Fine, I will make the change.

zubair-arbi:
@georgebabey please let us know if you are ok with the implementation so that we can merge this PR.

tests/test_admin/test_forms.py
```diff
@@ -543,9 +543,11 @@ def test_empty_catalog_value(self):
                 'active': customer.active,
                 'slug': customer.slug,
                 'country': customer.country,
+                'customer_type': customer.customer_type_id,
             },
             instance=customer,
         )
+        # import pytest; pytest.set_trace()
```
georgebabey: remove commented code

PR #415, opened 2018-12-14
ENT-1173 Add option on EnterpriseCustomer for displaying code management in portal

brittneyexline:
**Description:** Add option on EnterpriseCustomer for displaying code management in portal

**JIRA:** https://openedx.atlassian.net/browse/ENT-1173

**Dependencies:** n/a

**Merge deadline:** ASAP

**Testing instructions:**

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


enterprise/migrations/0059_add_code_management_portal_config.py
```diff
@@ -0,0 +1,25 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='enterprisecustomer',
+            name='enable_portal_code_management_screen',
+            field=models.BooleanField(default=False, help_text='Specifies whether to allow access to the code management screen in the admin portal.'),
+        ),
+        migrations.AddField(
+            model_name='historicalenterprisecustomer',
```
adamstankiewicz: Just for context and my general understanding, what is the `historialenterprisecustomer` model for?

brittneyexline: it was auto generated by makemigrations because of this line: https://github.com/edx/edx-enterprise/blob/41580978a5ce1e30896e08c70399fa1e8a55a7d3/enterprise/models.py#L176
we have it presumably to keep the history of changes on instances of this model, but i'm not sure if we really use it or care about it.

PR #419, opened 2019-01-11
Remove a GET side-effect from consent flow

tuchfarber:
**Description:** First part of removing side effects from GET calls. Previously the logic for handling the consent was in a separate view that we redirected to. However 302 redirects (even from POSTs) become GETs, so we had a GET view that had side effects. By removing this view/URL and instead handling the logic in a function inside the post call we are better protected. This was the only instance of that view being called.

**JIRA:** http://openedx.atlassian.net/browse/ENT-1307

**Dependencies:** None

**Merge deadline:** N/A

**Installation instructions:** N/A

**Testing instructions:**

1. Enroll a enterprise user in a course where the enterprise has data sharing consent enabled
2. Verify things work as before.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


tuchfarber:
@edx/business-enterprise-team please review :)

tests/test_enterprise/views/test_handle_consent_enrollment.py
```diff
@@ -1,310 +0,0 @@
-# -*- coding: utf-8 -*-
-"""
-Tests for the ``HandleConsentEnrollment`` view of the Enterprise app.
```
christopappas: I think it might be worth adapting these tests from testing the endpoint to testing the function

tuchfarber:
@christopappas @brittneyexline Tests added and rejiggered for the new view. Ready for re-review whenever you get a chance.

PR #421, opened 2019-01-17
ENT-143:Remove Errors On Transmitting Learner Data

bessiesteinberg:
**Description:** 
- We were getting errors when running the transmit_learner_data command
  if there were users that had an enterprise enrollment but no
  enrollment for the actual course yet
- This change removes that error, instead an INFO message is printed
  with the format:
   ```User [HannahAbbott] not enrolled in course [course-v1:TestX+TEST101+2018], enterprise enrollment [2]``

**JIRA:** https://openedx.atlassian.net/browse/ENT-1439

**Dependencies:** None

**Merge deadline:** None

**Testing instructions:**
1. Create enroll enterprise customer enrollment for a course, but don't actually enroll the user in the course
2. Run the following management command (replace arguments as appropriate): ```./manage.py lms --settings=devstack_docker transmit_learner_data --api_user=edx --enterprise_customer=53ad158d73b346289a5533427c6885e2```

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

### DIFF
--- a/integrated_channels/integrated_channel/exporters/learner_data.py
+++ b/integrated_channels/integrated_channel/exporters/learner_data.py
@@ -9,6 +9,7 @@ enterprise customer.
 
 from __future__ import absolute_import, unicode_literals
 
+import json
 from logging import getLogger
 
 from slumber.exceptions import HttpNotFoundError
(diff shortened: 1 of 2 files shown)


integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -238,10 +239,22 @@ def _collect_grades_data(self, enterprise_enrollment, course_details):
...
             grades_data = self.grades_api.get_course_grade(course_id, username)
 
-        except HttpNotFoundError:
+        except HttpNotFoundError as error:
             # Grade not found, so we have nothing to report.
-            LOGGER.error("No grades data found for [%d]: [%s], [%s]", enterprise_enrollment.pk, course_id, username)
-            return None, None, None
+            try:
+                response_content = json.loads(error.content)
+                if response_content.get('error_code', '') == 'user_not_enrolled':
```
georgebabey: What about the non-"user_not_enrolled" http errors - I think we should still log  those errors in an else clause - I think as written this will swallow those.

georgebabey: It will, but now it will flood the logs again, I think we just need the `LOGGER.error` in an else clause so we *only* log if its not the `user_not_enrolled` error

BessieSteinberg: Sounds good, I'll take out the logger.info

PR #424, opened 2019-01-22
Pwnage101/add pii annotations

pwnage101:

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


.annotation_safe_list.yml
```diff
@@ -0,0 +1,26 @@
+# This is a Code Annotations automatically-generated Django model safelist file.
+# These models must be annotated as follows in order to be counted in the coverage report.
+# See https://code-annotations.readthedocs.io/en/latest/safelist.html for more information.
+#
+# fake_app_1.FakeModelName:
+#    ".. no_pii::": "This model has no PII"
+# fake_app_2.FakeModel2:
+#    ".. choice_annotation::": foo, bar, bazadmin.LogEntry: {}
```
bmedx: This fell afoul of the same bug with safelist generation as Julia's. I'll work on a fix for that tomorrow, but it will need a newline added.

openedx.yaml
```diff
@@ -5,4 +5,6 @@ owner: edx/platform-team
 nick: edx-enterprise
```
doctoryes: Nit: The owner above doesn't seem correct?

tox.ini
```diff
@@ -102,3 +102,12 @@ deps =
...
 commands =
     jasmine-ci
+
+[testenv:pii-annotations]
+setenv =
+    DJANGO_SETTINGS_MODULE = enterprise.settings.test
+deps =
+    -r{toxinidir}/requirements/test-master.txt
+commands =
+    code_annotations django_find_annotations --config_file .pii_annotations.yml --report_path pii_report/ \
```
doctoryes: Maybe leave out `--report_path`? Should be set to same path via the config file.

integrated_channels/degreed/models.py
```diff
@@ -188,6 +192,10 @@ def get_content_metadata_exporter(self, user):
 class DegreedLearnerDataTransmissionAudit(models.Model):
     """
     The payload we sent to Degreed at a given point in time for an enterprise course enrollment.
+
+    .. pii:: The degreed_user_email model field contains PII.
+    .. pii_types:: email_address
+    .. pii_retirement:: consumer_api
```
doctoryes: Honestly? I don't understand the difference between `local_api` and `consumer_api`. If the PII is stored in the edx-enterprise models, wouldn't it all be `local_api`? It all needs to be deleted locally, correct?

pwnage101: I thought local_api meant that in this codebase there is code to retire the PII.  In this case, the code to retire the PII is in the edx-platform codebase.  Isn't edx-platform the "consumer"?

openedx.yaml
pwnage101: it's so wrong.

integrated_channels/degreed/models.py
doctoryes: Ah - that makes sense now. The OEP-30 language could use some detail on this point.

enterprise/models.py
```diff
@@ -803,6 +808,9 @@ class PendingEnterpriseCustomerUser(TimeStampedModel):
         enterprise_customer (ForeignKey[:class:`.EnterpriseCustomer`]): enterprise customer
         user_email (:class:`django.db.models.EmailField`): user email
 
+    .. pii:: The user_email field contains PII, but locally deleted via enterprise.signals.handle_user_post_save.
```
bmedx: I think we talked about this model. I wonder if it makes sense to clarify that if the user never converts to a full user that their email address will persist. We should at least clarify in that there is an that deletes this data in edx-platform and add `consumer_api` to `pii_retirement`.

.annotation_safe_list.yml
pwnage101: I'm actually going to delete admin.LogEntry: {} because it is a 3rd party model which doesn't need to be annotated in this non-IDA repository.

pwnage101:
@georgebabey can you take a look?  edx-enterprise is one of our guinea pig repos for integrating the OEP-30 PII annotations work.

pwnage101:
Reviewers, I had to fix a bunch of pycodestyle violations in the commit called "Fix many pycodestyle errors surfaced by fixing pycodestyle config", so I highly recommend that you review this PR one commit at a time.

PR #426, opened 2019-01-24
ENT-1386 | Better logging and new handing of errors in views

christopappas:
… for program enrollments, course enrollments, and consenting data sharing

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Prove that a user can follow through normal consent workflow successfully
2. Prove that when an error occurs during consent workflow, that the proper new error page is displayed.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


BessieSteinberg:
I think error codes are a great idea, but make sure to include a readme or something like that that has a list of error codes so that developers don't have to regex through the code base when they want to create a new code :)

mattdrayer:
Agreed -- ref: https://github.com/edx/open-edx-proposals/blob/master/oeps/oep-0019-bp-developer-documentation.rst

enterprise/views.py
```diff
@@ -375,32 +465,71 @@ def get(self, request):
...
         program_uuid = request.GET.get('program_uuid', '')
         self.preview_mode = bool(request.GET.get('preview_mode', False))
-        if not (enterprise_customer_uuid and success_url and failure_url):
-            raise Http404
 
-        if not self.preview_mode:
-            if not self.course_or_program_exist(course_id, program_uuid):
-                raise Http404
+        errored = False
+        enterprise_customer = None
```
christopappas: this can go

enterprise/views.py
```diff
@@ -375,32 +465,71 @@ def get(self, request):
...
+            LOGGER.error(
+                'CourseCatalogApiServiceClient is improperly configured.'
+                'Returned error code {error_code} to user'.format(
+                    error_code=error_code,
+                )
+            )
+            messages.add_generic_error_message_with_code(request, error_code)
+            errored = True
+
+        if errored:
```
christopappas: i thought it was more concise to just render the error page with a few different error messages via 1 check later in the view if there were any errors... that technically does change the behavior of the view slightly though... i went with this solution because when using the helper methods that could choke in a few different ways, it was simpler just to keep tabs on whether or not anything went wrong throughout the process, and throw an error page at the end. questioning that now

enterprise/views.py
```diff
@@ -375,32 +465,71 @@ def get(self, request):
...
+                'failure_url: {failure_url}. '
+                'The following error code was reported to the user: {error_code}'.format(
+                    enterprise_customer_uuid=enterprise_customer_uuid,
+                    success_url=success_url,
+                    failure_url=failure_url,
+                    error_code=error_code,
+                )
+            )
+            messages.add_generic_error_message_with_code(request, error_code)
+            errored = True
```
christopappas: the reason this was the exception to the rule of rendering 404 immediately after an error was because i did not have the context_data yet, which requires an enterprise_customer to get the global context, so I deferred to waiting a little further down the code

enterprise/views.py
```diff
@@ -1124,16 +1342,46 @@ def get_program_details(self, request, program_uuid, enterprise_customer):
...
+                'A user of enterprise customer {enterprise_customer} encountered an error. '
+                'program_type is None for program_details of program_uuid {program_uuid}. '
+                'Returned error code {error_code} to user'.format(
+                    enterprise_customer=enterprise_customer.uuid,
+                    program_uuid=program_uuid,
+                    error_code=error_code,
+                )
+            )
+            messages.add_generic_error_message_with_code(request, error_code)
+            return ({}, error_code)
```
christopappas: because there are a few embedded methods here that normally return dictionaries, we want to pass up a blank dictionary and the error_code so we can bubble up the errors and catch them in the respective GET/POST

enterprise/views.py
christopappas: if the enterprise_customer_uuid is not valid here though, later on down the road to get the global context, things will choke (because getting the global context requires an enterprise customer). consider adding another render here, but with a more sparse context

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -606,28 +627,36 @@ def test_get_program_consent_missing_parameter(self, missing_parameter):
         valid_get_params = self.valid_get_params.copy()
         valid_get_params.pop(missing_parameter)
         self._login()
-        response = self.client.get(self.url, valid_get_params)
-        assert response.status_code == 404
+        with mock.patch('enterprise.views.render') as mock_render:
```
brittneyexline: looks like we have this pattern repeating in many tests: mock render and its return value, call some url, assert that the response code is 404. can this be consolidated in one method that each test that needs it calls?

enterprise/views.py
```diff
@@ -375,32 +465,70 @@ def get(self, request):
...
-        )
+        try:
+            updated_context_dict = self.get_course_or_program_context(
+                enterprise_customer,
+                course_id=course_id,
+                program_uuid=program_uuid
+            )
+            context_data.update(updated_context_dict)
+        except Http404:
+            error_code = 'ENTGDS000'
```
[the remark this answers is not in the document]
christopappas: yeesh. good catch. likely a copy paste error. the code is wrong here. updated.

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
christopappas: yeah can be done. in looking at this, i found a few copy paste errors (not just my own!) that needed to be fixed in this test file, so pushing that up with shortly.

enterprise/views.py
```diff
@@ -375,32 +390,122 @@ def get(self, request):
...
+                    )
+                )
+                return render_page_with_error_code_message(request, context_data, error_code, log_message)
+            else:
+                enterprise_customer = consent_record.enterprise_customer
+        elif not request.user.is_staff:
+            raise PermissionDenied()
 
+        # Retrieve context data again now that enterprise_customer logic has been run
         context_data = get_global_context(request, enterprise_customer)
```
brittneyexline: it's weird to me that we have to call this again, i doubt we would end up with a different enterprise customer object from the consent object since both lookups are by the uuid.

enterprise/views.py
```diff
@@ -453,13 +558,45 @@ def post(self, request):
...
-        if not (enterprise_uuid and success_url and failure_url):
-            raise Http404
-
         course_id = request.POST.get('course_id', '')
         program_uuid = request.POST.get('program_uuid', '')
+
+        enterprise_customer = get_enterprise_customer_or_404(enterprise_uuid)
+        context_data = get_global_context(request, enterprise_customer)
+
+        if not (enterprise_uuid and success_url and failure_url):
```
brittneyexline: again, at this point if enterprise_uuid was empty we would have already 404'd

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -347,19 +350,22 @@ def test_get_course_specific_consent_bad_api_response(
...
-        response = self.client.get(
-            self.url + '?course_id=course-v1%3AedX%2BDemoX%2BDemo_Course&next=https%3A%2F%2Fgoogle.com'
-        )
-        assert response.status_code == 404
+        with mock.patch('enterprise.views.render') as mock_render:
+            mock_render.return_value = HttpResponse()
+            self.client.get(
+                self.url + '?course_id=course-v1%3AedX%2BDemoX%2BDemo_Course&next=https%3A%2F%2Fgoogle.com',
+                data={
+                    'enterprise_customer_uuid': enterprise_customer.uuid,
```
brittneyexline: seems like we are testing something slightly different if we are sending in the enterprise_customer_uuid here when we didn't before.... i suspect this has to do with the ordering of your checks related to whether this param exists and how we get the enterprise customer object (see above)

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -388,10 +394,16 @@ def test_get_course_specific_consent_not_needed(
...
-            self.url + '?course_id=course-v1%3AedX%2BDemoX%2BDemo_Course&next=https%3A%2F%2Fgoogle.com'
-        )
-        assert response.status_code == 404
+
+        with mock.patch('enterprise.views.render') as mock_render:
+            mock_render.return_value = HttpResponse()
+            self.client.get(
+                self.url + '?course_id=course-v1%3AedX%2BDemoX%2BDemo_Course&next=https%3A%2F%2Fgoogle.com',
+                data={
+                    'enterprise_customer_uuid': enterprise_customer.uuid,
```
brittneyexline: same here - why do we need this now? i worry that it won't reflect the flow users actually go through; we might not be sending this parameter to the view in these cases in the wild.

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -640,28 +681,33 @@ def test_post_program_consent_missing_parameter(self, missing_parameter):
...
+        self._assert_post_returns_404_with_mock(self.url, self.valid_post_params)
 
     def test_post_program_consent_not_required(self):
         self.get_data_sharing_consent.return_value.consent_required.return_value = False
         self.course_catalog_api_client.return_value.program_exists.return_value = True
         self._login()
-        response = self.client.get(self.url, self.valid_post_params)
-        assert response.status_code == 404
+        response = self.client.post(self.url, self.valid_post_params)
+        assert response.status_code == 302
```
brittneyexline: why is this a 302 now?

christopappas: the test name was test_post_program_consent_not_required but the call in the test was a get. i changed it to a post, but a post with the same parameters results in a redirect, and not a 404, so i updated the result as well

enterprise/views.py
```diff
@@ -375,32 +390,122 @@ def get(self, request):
         course_id = request.GET.get('course_id', '')
         program_uuid = request.GET.get('program_uuid', '')
         self.preview_mode = bool(request.GET.get('preview_mode', False))
-        if not (enterprise_customer_uuid and success_url and failure_url):
-            raise Http404
+
+        # Get enterprise_customer to start in case we need to render a custom 404 page
+        # Then go through other business logic to determine (and potentially overwrite) the enterprise customer
+        enterprise_customer = get_enterprise_customer_or_404(enterprise_customer_uuid)
```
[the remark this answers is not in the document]
christopappas: should i be catching the http404 everywhere we call get_enterprise_customer_or_404?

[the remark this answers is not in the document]
christopappas: part of the reason why i left the old http404 from get_enterprise_customer_or_404 is because, if we don't have the enterprise customer, then we don't have what we need to get the proper context to render the page, and therefore are unable to properly render the custom error page

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
christopappas: okay so what happened here is that get_enterprise_customer_or_404 was throwing a value error on a blank string being handed to it throughout my work, so i ended up tweaking this and the other test you commented on. i didn't assume the test suite was correct.

that said, ultimately, i ended up changing utils.py to catch the value error and to return a http404, so I will change this test back

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
christopappas: per above comment, i'll change this back as well

PR #437, opened 2019-03-13
Assign enterprise roles to existing enterprise users

mushtaqak:
**Description:**

This PR creates a management command that assigns enterprise roles to existing enterprise users.

1. This will  find existing users with `enterprise_data_api_access` django group and assigns an `enterprise_admin` role to them.
2. This also assigns existing `EnterpriseCustomerUsers` an `enterprise_learner` role.

**JIRA:** 
https://openedx.atlassian.net/browse/ENT-1685
Main ticket: https://openedx.atlassian.net/browse/ENT-1554

### DIFF
--- a/enterprise/management/commands/assign_enterprise_user_roles.py
+++ b/enterprise/management/commands/assign_enterprise_user_roles.py
@@ -1,0 +1,134 @@
+"""
+Management command for assigning enterprise roles to existing enterprise users.
+"""
+from __future__ import absolute_import, unicode_literals
+
+import logging
+from time import sleep
+
+from django.contrib.auth.models import User
+from django.core.management.base import BaseCommand, CommandError
+
+from enterprise.constants import ENTERPRISE_ADMIN_ROLE, ENTERPRISE_DATA_API_ACCESS_GROUP, ENTERPRISE_LEARNER_ROLE
+from enterprise.models import EnterpriseCustomerUser, SystemWideEnterpriseRole, SystemWideEnterpriseUserRoleAssignment
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Management command for assigning enterprise roles to existing enterprise users.
+
+    Example usage:
+        $ ./manage.py assign_enterprise_user_roles --role=enterprise_admin
+        $ ./manage.py assign_enterprise_user_roles --role=enterprise_learner
+    """
+    help = 'Assigns enterprise roles to existing enterprise users.'
+
+    def add_arguments(self, parser):
+        """
+        Entry point for subclassed commands to add custom arguments.
+        """
+        parser.add_argument(
+            '--role',
+            action='store',
+            dest='role',
+            default=None,
+            help='Role to assign users role assignments.'
+        )
+        parser.add_argument(
+            '--batch-limit',
+            action='store',
+            dest='batch_limit',
+            default=100,
+            help='Number of users in each batch of conditional offer migration.',
+            type=int,
+        )
+
+        parser.add_argument(
+            '--batch-offset',
+            action='store',
+            dest='batch_offset',
+            default=0,
+            help='Which index to start batching from.',
+            type=int,
+        )
+
+        parser.add_argument(
+            '--batch-sleep',
+            action='store',
+            dest='batch_sleep',
+            default=10,
+            help='How long to sleep between batches.',
+            type=int,
+        )
+
+    def _get_enterprise_admin_users_batch(self, start, end):
+        """
+        Returns a batched queryset of User objects.
+        """
+        LOGGER.info('Fetching new batch of enterprise admin users from indexes: %s to %s', start, end)
+        return User.objects.filter(groups__name=ENTERPRISE_DATA_API_ACCESS_GROUP, is_staff=False)[start:end]
+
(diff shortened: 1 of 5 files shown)


enterprise/management/commands/assign_enterprise_user_roles.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def handle(self, *args, **options):
+        try:
+            enterprise_admin_role = SystemWideEnterpriseRole.objects.get(name=ENTERPRISE_ADMIN_ROLE)
+            enterprise_learner_role = SystemWideEnterpriseRole.objects.get(name=ENTERPRISE_LEARNER_ROLE)
+        except SystemWideEnterpriseRole.DoesNotExist:
+            raise CommandError('The configured SystemWideEnterpriseRole role does not exist.')
+
+        # Assign admin role to users with enterprise data api access .
+        users_with_enterprise_access = User.objects.filter(groups__name=ENTERPRISE_DATA_API_ACCESS_GROUP)
```
mushtaqak: Open question: Should we also filter users for `enterprise_enrollment_api_access` group?

enterprise/management/commands/assign_enterprise_user_roles.py
```diff
@@ -0,0 +1,38 @@
...
+class Command(BaseCommand):
+    """ management command class """
+    help = 'Assigns existing enterprise users to their specific roles.'
+
+    def handle(self, *args, **options):
+        try:
+            enterprise_admin_role = SystemWideEnterpriseRole.objects.get(name=ENTERPRISE_ADMIN_ROLE)
+            enterprise_learner_role = SystemWideEnterpriseRole.objects.get(name=ENTERPRISE_LEARNER_ROLE)
+        except SystemWideEnterpriseRole.DoesNotExist:
+            raise CommandError('The configured SystemWideEnterpriseRole role does not exist.')
```
muhammad-ammar: I don't think we need `try...except` here because `SystemWideEnterpriseRole.objects.get(name=ENTERPRISE_ADMIN_ROLE)` should always return a result. 
Do we expect that the above system wide role will not be present?

enterprise/management/commands/assign_enterprise_user_roles.py
mushtaqak: @brittneyexline ^^

mushtaqak:
@muhammad-ammar Addressed your feedback.

enterprise/management/commands/assign_enterprise_user_roles.py
brittneyexline: for now no - the enrollment api access is more or less separate from how we "grant admin access" at this point.

enterprise/management/commands/assign_enterprise_user_roles.py
```diff
@@ -0,0 +1,35 @@
...
+    """ management command class """
+    help = 'Assigns existing enterprise users to their specific roles.'
+
+    def handle(self, *args, **options):
+        enterprise_admin_role = SystemWideEnterpriseRole.objects.get(name=ENTERPRISE_ADMIN_ROLE)
+        enterprise_learner_role = SystemWideEnterpriseRole.objects.get(name=ENTERPRISE_LEARNER_ROLE)
+
+        # Assign admin role to users with enterprise data api access.
+        users_with_enterprise_access = User.objects.filter(groups__name=ENTERPRISE_DATA_API_ACCESS_GROUP)
+        for user_with_enterprise_access in users_with_enterprise_access:
```
brittneyexline: we should only be doing this for users who are not staff users. i am thinking that we should introduce an edx_enterprise_operator system wide role to apply to the staff users who need it, but you don't need to worry about that right now in the scope of this PR.

enterprise/management/commands/assign_enterprise_user_roles.py
```diff
@@ -0,0 +1,35 @@
...
+            SystemWideEnterpriseUserRoleAssignment.objects.update_or_create(
+                user=user_with_enterprise_access,
+                role=enterprise_admin_role
+            )
+
+        # Assign enterprise learner role to enterprise customer users.
+        enterprise_learners = EnterpriseCustomerUser.objects.all()
```
brittneyexline: we might want to separate this from the logic above as there are about 55 thousand of these records compared to those in the group (not more than 100). because of the size, we also should structure the command for enterprise learners so that we are batching these updates. see https://github.com/edx/edx-platform/blob/master/lms/djangoapps/grades/management/commands/compute_grades.py#L57

enterprise/management/commands/assign_enterprise_user_roles.py
```diff
@@ -0,0 +1,145 @@
...
+        current_batch_index = batch_offset
+        logger.info('Starting assigning enterprise roles to users!')
+
+        # Assign admin role to non-staff users with enterprise data api access.
+        if role == ENTERPRISE_ADMIN_ROLE:
+            enterprise_admin_users = self._get_enterprise_admin_users_batch(
+                batch_offset,
+                batch_offset + batch_limit
+            )
+            while len(enterprise_admin_users) > 0:
```
brittneyexline: seems like you could DRY this up and consolidate with some of the logic below

mushtaqak: @brittneyexline addressed :)

PR #441, opened 2019-03-25
ENT-1729 Introducing feature specific roles for enterprise

brittneyexline:
**Description:** This PR adds the feature specific role models, data, and django admin for enterprise.

**JIRA:** https://openedx.atlassian.net/browse/ENT-1729

**Testing instructions:**

Migrations run successfully when installed into edx-platform. The django admin allows successful creation, update, and deletion of the new role assignment model.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


enterprise/models.py
```diff
@@ -1823,27 +1855,49 @@ def __repr__(self):
...
+
+    def __repr__(self):
+        """
+        Return uniquely identifying string representation.
+        """
+        return self.__str__()
+
+
+@python_2_unicode_compatible
+class EnterpriseFeatureUserRoleAssignment(UserRoleAssignment, EnterpriseRoleAssignmentContextMixin):
```
muhammad-ammar: @brittneyexline  One more thing. `EnterpriseRoleAssignmentContextMixin` should come before `UserRoleAssignment`, otherwise according to MRO `UserRoleAssignment.get_context` will be called every time instead of `EnterpriseRoleAssignmentContextMixin.get_context`

PR #443, opened 2019-03-27
Assign edx enterprise operator role

mushtaqak:
The purpose of this PR is to add the capability to assign enterprise operator role to staff users with `enterprise_data_api_access` group in `assign_enterprise_user_roles` management command.

[ENT-1736](https://openedx.atlassian.net/browse/ENT-1736)

### DIFF
--- a/enterprise/management/commands/assign_enterprise_user_roles.py
+++ b/enterprise/management/commands/assign_enterprise_user_roles.py
@@ -9,7 +9,12 @@ from time import sleep
 from django.contrib.auth.models import User
 from django.core.management.base import BaseCommand, CommandError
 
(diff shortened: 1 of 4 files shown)


enterprise/management/commands/assign_enterprise_user_roles.py
```diff
@@ -122,6 +134,9 @@ def handle(self, *args, **options):
         if role == ENTERPRISE_ADMIN_ROLE:
             # Assign admin role to non-staff users with enterprise data api access.
             self._assign_enterprise_role_to_users(self._get_enterprise_admin_users_batch, options)
+        if role == ENTERPRISE_OPERATOR_ROLE:
```
brittneyexline: nit: this should probably be elif

PR #449, opened 2019-04-04
WL-1906 | Added an extra check of user's enrollment in Enterprise API.

hasnain-naveed:
**Description:** Added an extra check of user's enrollment in `RouterView`

**JIRA:** [WL-1906](https://openedx.atlassian.net/browse/WL-1906)

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -749,35 +749,63 @@ def get_course_run_start(course_run, default=None):
     return parse_datetime_handle_invalid(course_run.get('start')) or default
 
 
-def get_current_course_run(course):
+def get_closest_course_run(course_runs):
     """
-    Return the current course run.
+    Return course run with start date closest to now.
+    """
+    if len(course_runs) == 1:
+        return course_runs[0]
 
-    Current is defined as the following:
+    now = datetime.datetime.now(pytz.UTC)
+    # course runs with no start date should be considered last.
+    never = now - datetime.timedelta(days=3650)
+    return min(course_runs, key=lambda x: abs(get_course_run_start(x, never) - now))
+
+
+def get_active_course_runs(course, users_all_enrolled_courses):
+    """
+    Return active course runs (user is enrolled in) of the given course.
+
+    This function will return the course_runs of 'course' which have
+    active enrollment by looking into 'users_all_enrolled_courses'
+    """
+    # User's all course_run ids in which he has enrolled.
+    enrolled_course_run_ids = [
+        enrolled_course_run['course_details']['course_id'] for enrolled_course_run in users_all_enrolled_courses
+        if enrolled_course_run['is_active'] and enrolled_course_run.get('course_details')
+    ]
+    return [course_run for course_run in course['course_runs'] if course_run['key'] in enrolled_course_run_ids]
+
+
+def get_current_course_run(course, users_active_course_runs):
+    """
+    Return the current course run on the following conditions.
+
+    - If user has active course runs (already enrolled) then return course run with closest start date
+    Otherwise it will check the following logic:
     - Course run is enrollable (see is_course_run_enrollable)
     - Course run has a verified seat and the upgrade deadline has not expired.
     - Course run start date is closer to now than any other enrollable/upgradeable course runs.
     - If no enrollable/upgradeable course runs, return course run with most recent start date.
     """
-    now = datetime.datetime.now(pytz.UTC)
-    never = now - datetime.timedelta(days=3650)
-    all_course_runs = course['course_runs']
+    current_course_run = None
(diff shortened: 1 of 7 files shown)


enterprise/views.py
```diff
@@ -1659,7 +1660,11 @@ def get_course_run_id(enterprise_customer, course_key):
         except ImproperlyConfigured:
             raise Http404
 
-        course_run = get_current_course_run(course)
+        users_active_course_runs = [
+            course_run for course_run in course['course_runs']
+            if EnrollmentApiClient().is_enrolled(user.username, course_run['key'])
```
georgebabey: I don't love making this API call inside a loop, we run the risk of rate limiting ourselves if a course has many runs. Can we update this to either:
- make a single call with the course and have the enrollment API endpoint do the lookup of course_runs
- have the endpoint take the full list of runs to validate enrollment
- change it to call the enrollment code directly instead of via API

I don't have a strong preference on which of those 3 approaches (there may already be API endpoints for one of the first 2 options, I haven't looked) but I think we should update this before merging

enterprise/utils.py
```diff
@@ -749,35 +749,48 @@ def get_course_run_start(course_run, default=None):
...
+            if is_course_run_enrollable(course_run) and is_course_run_upgradeable(course_run):
+                filtered_course_runs.append(course_run)
 
-    if filtered_course_runs:
-        # Return course run with start date closest to now.
-        # Course runs with a null start date should be considered last.
-        return min(filtered_course_runs, key=lambda x: abs(get_course_run_start(x, never) - now))
+        if not filtered_course_runs:
+            # Consider all runs if there were not any enrollable/upgradeable ones.
+            filtered_course_runs = all_course_runs
```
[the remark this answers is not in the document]
douglashall: My guess is that this was the business logic asked for by product when this was originally written.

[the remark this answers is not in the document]
georgebabey: Thanks @douglashall. Seems odd but tangental to this PR and that logic was already there so lets keep it for now. Filing mental note if a ticket comes in around displaying all runs somewhere ;-)

enterprise/views.py
hasnain-naveed: @georgebabey I have updated the code. I noticed that there is an endpoint which returns the all active enrollment of users. I have introduced the new function in utils to filter. Let me know if you have still concerns.

enterprise/views.py
```diff
@@ -1659,7 +1661,10 @@ def get_course_run_id(enterprise_customer, course_key):
         except ImproperlyConfigured:
             raise Http404
 
-        course_run = get_current_course_run(course)
+        users_all_enrolled_courses = EnrollmentApiClient().get_enrolled_courses(user.username)
+        users_active_course_runs = get_active_course_runs(course, users_all_enrolled_courses) \
```
tuchfarber: nit: Mutliline with parens instead of backslash per pep8.

enterprise/utils.py
```diff
@@ -749,35 +749,63 @@ def get_course_run_start(course_run, default=None):
     return parse_datetime_handle_invalid(course_run.get('start')) or default
 
 
-def get_current_course_run(course):
+def get_closest_course_run(course_runs):
     """
-    Return the current course run.
+    Return course run with start date closest to now.
+    """
+    if len(course_runs) == 1:
```
mattdrayer: Question about this check -- it seems to assume that a learner is only enrolled in one run of a course, and if so it will short-circuit and return that run back to the caller.  However, the scenario that was presented which triggered this PR was that learners were enrolled in multiple runs of a course.  In this case, will the short-circuit be bypassed because there are two active runs for the course?  Which could potentially lead to a third run being selected as "current" at some point in the future, and then possibly a fourth, etc.?

hasnain-naveed: So basically there are two use cases when a user request for course

1. A user is coming for the first time for that course and a user has no clue for which course run to go in that case user will go to the closest course run. It was previously implemented and working the same.

2. User is already enrolled in one or more course runs. In that case, user will always land on the already enrolled course run, if user is enrolled in one course run then the user will land to that course run this short circuiting will be responsible for that. But if the user is enrolled more than 1 course run then we will find the closest one from those enrolled course runs. **The egde case was user will never be go to the next course run automtically.** User will be enrolled to the next course run manually by admin/enterprise admin.
FYI, @ziafazal @georgebabey

PR #453, opened 2019-04-11
Find users for feature specific enterprise roles

mushtaqak:
**Description:** 

This PR adds two feature specific role assignment functionality to the `assign_enterprise_user_roles` management command.

> For two of the feature specific roles, enterprise admins do not get mapped to them automatically.
We need to find the users who currently hit the enterprise catalog endpoints and grant them the feature specific catalog role. The best proxy would be any user who is linked to an enterprise but has also been granted client credentials api keys.
We also need to find the users who are in the enrollment_api_access_group and grant them the feature specific enrollment api admin role.

**JIRA:**
https://openedx.atlassian.net/browse/ENT-1785

** Testing Instructions **

a) _enrollment api admin role_
1. Create user and assign `enterprise_enrollment_api_access` group to him.
2. Run `/manage.py lms assign_enterprise_user_roles --role=enrollment_api_admin` in LMS.
3. Observe that `EnterpriseFeatureUserRoleAssignment` record has been created for the user.

b) _catalog admin_
1. Create user object.
2. Create a recod for the user in `oauth_provider_application` admin table for the user.
3. Create an enterprise customer user record for the user.
4. Run `/manage.py lms assign_enterprise_user_roles --role=catalog_admin` in LMS.
5. Observe that `EnterpriseFeatureUserRoleAssignment` record has been created for the user.

### DIFF
--- a/enterprise/management/commands/assign_enterprise_user_roles.py
+++ b/enterprise/management/commands/assign_enterprise_user_roles.py
@@ -6,16 +6,28 @@ from __future__ import absolute_import, unicode_literals
 import logging
 from time import sleep
 
+from django.apps import apps
 from django.contrib.auth.models import User
 from django.core.management.base import BaseCommand, CommandError
 
 from enterprise.constants import (
+    EDX_ORG_NAME,
     ENTERPRISE_ADMIN_ROLE,
+    ENTERPRISE_CATALOG_ADMIN_ROLE,
     ENTERPRISE_DATA_API_ACCESS_GROUP,
+    ENTERPRISE_ENROLLMENT_API_ACCESS_GROUP,
+    ENTERPRISE_ENROLLMENT_API_ADMIN_ROLE,
     ENTERPRISE_LEARNER_ROLE,
     ENTERPRISE_OPERATOR_ROLE,
+    OAUTH2_PROVIDER_APPLICATION_MODEL,
 )
-from enterprise.models import EnterpriseCustomerUser, SystemWideEnterpriseRole, SystemWideEnterpriseUserRoleAssignment
+from enterprise.models import (
+    EnterpriseCustomerUser,
+    EnterpriseFeatureRole,
+    EnterpriseFeatureUserRoleAssignment,
+    SystemWideEnterpriseRole,
+    SystemWideEnterpriseUserRoleAssignment,
+)
 
 LOGGER = logging.getLogger(__name__)
 
@@ -27,6 +39,9 @@ class Command(BaseCommand):
     Example usage:
         $ ./manage.py assign_enterprise_user_roles --role=enterprise_admin
         $ ./manage.py assign_enterprise_user_roles --role=enterprise_learner
+        $ ./manage.py assign_enterprise_user_roles --role=enterprise_openedx_operator
+        $ ./manage.py assign_enterprise_user_roles --role=catalog_admin
+        $ ./manage.py assign_enterprise_user_roles --role=enrollment_api_admin
     """
     help = 'Assigns enterprise roles to existing enterprise users.'
 
@@ -68,6 +83,12 @@ class Command(BaseCommand):
             type=int,
         )
 
+    def _get_enterprise_customer_user_ids(self):
+        """
+        Returns a queryset containing user ids.
+        """
+        return EnterpriseCustomerUser.objects.values('user_id')
+
     def _get_enterprise_admin_users_batch(self, start, end):
         """
         Returns a batched queryset of User objects.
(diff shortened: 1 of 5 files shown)


enterprise/management/commands/assign_enterprise_user_roles.py
```diff
@@ -87,9 +105,27 @@ def _get_enterprise_customer_users_batch(self, start, end):
         Returns a batched queryset of EnterpriseCustomerUser objects.
         """
         LOGGER.info('Fetching new batch of enterprise customer users from indexes: %s to %s', start, end)
-        return User.objects.filter(pk__in=EnterpriseCustomerUser.objects.values('user_id'))[start:end]
+        return User.objects.filter(pk__in=self._get_enterprise_customer_user_ids())[start:end]
```
saleem-latif: nit: I think the previous one is more descriptive.

enterprise/management/commands/assign_enterprise_user_roles.py
```diff
@@ -68,6 +80,12 @@ def add_arguments(self, parser):
             type=int,
         )
 
+    def _get_enterprise_customer_user_ids(self):
```
saleem-latif: This seems to be a static method.

saleem-latif: It would be better to make it static or go with a more descriptive approach as was being used previously.

enterprise/management/commands/assign_enterprise_user_roles.py
```diff
@@ -87,9 +108,27 @@ def _get_enterprise_customer_users_batch(self, start, end):
         Returns a batched queryset of EnterpriseCustomerUser objects.
         """
         LOGGER.info('Fetching new batch of enterprise customer users from indexes: %s to %s', start, end)
-        return User.objects.filter(pk__in=EnterpriseCustomerUser.objects.values('user_id'))[start:end]
+        return User.objects.filter(pk__in=self._get_enterprise_customer_user_ids())[start:end]
 
-    def _assign_enterprise_role_to_users(self, _get_batch_method, options):
+    def _get_enterprise_enrollment_api_admin_users_batch(self, start, end):     # pylint: disable=invalid-name
```
saleem-latif: This also should be converted to a static method.

enterprise/management/commands/assign_enterprise_user_roles.py
```diff
@@ -87,9 +108,27 @@ def _get_enterprise_customer_users_batch(self, start, end):
...
+        """
+        Returns a batched queryset of User objects.
+        """
+        LOGGER.info('Fetching new batch of enterprise enrollment admin users from indexes: %s to %s', start, end)
```
saleem-latif: Why do we need this log message here?

enterprise/management/commands/assign_enterprise_user_roles.py
```diff
@@ -87,9 +108,27 @@ def _get_enterprise_customer_users_batch(self, start, end):
...
+        return User.objects.filter(groups__name=ENTERPRISE_ENROLLMENT_API_ACCESS_GROUP, is_staff=False)[start:end]
+
+    def _get_enterprise_catalog_admin_users_batch(self, start, end):
```
saleem-latif: This also seems to be a static method.

enterprise/management/commands/assign_enterprise_user_roles.py
```diff
@@ -87,9 +108,27 @@ def _get_enterprise_customer_users_batch(self, start, end):
...
+        """
+        Returns a batched queryset of User objects.
+        """
+        Application = apps.get_model(OAUTH2_PROVIDER_APPLICATION_MODEL)     # pylint: disable=invalid-name
+        LOGGER.info('Fetching new batch of enterprise catalog admin users from indexes: %s to %s', start, end)
```
saleem-latif: Why do we need this log message here?

enterprise/management/commands/assign_enterprise_user_roles.py
mushtaqak: yeah, it's used on a couple places so moved to a method

enterprise/management/commands/assign_enterprise_user_roles.py
mushtaqak: I guess this isn't called outside this management command, and I don't see any other code calling this in near future and also may be this command would be deleted ... so I am leaving it as is.

enterprise/management/commands/assign_enterprise_user_roles.py
mushtaqak: To log the information specific to this user role

PR #458, opened 2019-04-17
WL-1911 We are masking discovery call failures from the client, and we should stop.

brittneyexline:
**Description:** In trying to get better debugging information for WL-1911, it seems that this endpoint might be behaving in unexpected ways. In particular, I noticed that it returns false for the contains_content_items call when an error occurs in calling the discovery service, and thus the error is masked from the client. We need this to be surfaced so that we can understand the failures better.

I also removed the new style permission decorator for the endpoint because it is an endpoint called by internal services, and I'd like to derisk it from any further issues.

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -385,9 +385,6 @@ class EnterpriseCustomerCatalogViewSet(EnterpriseReadOnlyModelViewSet):
 
     @method_decorator(require_at_least_one_query_parameter('course_run_ids', 'program_uuids'))
     @detail_route()
-    @permission_required(
-        'enterprise.can_view_catalog',
-        fn=lambda request, pk, course_run_ids, program_uuids: get_enterprise_customer_from_catalog_id(pk))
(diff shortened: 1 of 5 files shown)


enterprise/api/v1/views.py
```diff
@@ -385,9 +385,6 @@ def get_serializer_class(self):
 
     @method_decorator(require_at_least_one_query_parameter('course_run_ids', 'program_uuids'))
     @detail_route()
-    @permission_required(
```
mushtaqak: @brittneyexline So we will be add this permission check back on before rolling out Authz changes ?

mushtaqak: NVM, just read the description part regarding this.

PR #462, opened 2019-04-26
ENT-1054: Fix enrollment page course key issues

BessieSteinberg:
The enterprise enrollment page (```enterprise/<enterprise-customer-id>/course/<course-id>/enroll/```) is failing in certain edge cases where the ```course_run_id``` and the ```course_id``` don't follow the expected format.

*Some back story to clarify the problem*:  Traditionally ```course_id```s have been substrings of ```course_run_id```s.  For example, for the course ```edX+DemoX``` has a course run ```course-v1:edX+DemoX+Demonstration_Course```.  However, ```course_id```s do not have to be substrings of their ```course_run_id```s.  You may have a course ```edX+DemoX``` with a course run ```course-v1:edX+DemoY+run_Y```.  Unfortunately, the ```edx-enterprise``` code assumes that ```course_id```s are always substrings of ```course_run_id```s.  Let's say a user tries to enroll in the course run ```course-v1:edX+DemoY+run_Y``` through the enterprise enrollment URL.  In order to populate that page we need to get data on the course associated with that course run.  The code tries to find that course by calling ```parse_course_key()``` and gets ```edX+DemoY```.  That course does not exist and causes problems.

*The solution*:  The discovery service understands the relationship between courses and course runs.  To fix this issue, instead of calling ```parse_course_key()``` we need to makea  call to the discovery service.  Specifically, the ```course_runs``` endpoint which takes a ```course_run_id``` and in the JSON blob it returns the ```course_id```.

Disclaimer: I know that ```course_id``` and ```course_run_id``` are not the way we reference those concepts in other parts of the code, but it is the way it's referenced in most of ```edx-enterprsie``` repo and I didn't want to change it

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/consent/helpers.py
+++ b/consent/helpers.py
@@ -8,7 +8,7 @@ from __future__ import absolute_import, unicode_literals
 from django.apps import apps
 
 from consent.models import ProxyDataSharingConsent
-from enterprise.api_client.discovery import CourseCatalogApiServiceClient
+from enterprise.api_client.discovery import get_course_catalog_api_service_client
 from enterprise.utils import get_enterprise_customer
 
 
@@ -59,7 +59,7 @@ def get_program_data_sharing_consent(username, program_uuid, enterprise_customer
     :return: The data sharing consent object
     """
     enterprise_customer = get_enterprise_customer(enterprise_customer_uuid)
-    discovery_client = CourseCatalogApiServiceClient(enterprise_customer.site)
+    discovery_client = get_course_catalog_api_service_client(enterprise_customer.site)
     course_ids = discovery_client.get_program_course_keys(program_uuid)
     child_consents = (
         get_data_sharing_consent(username, enterprise_customer_uuid, course_id=individual_course_id)
--- a/consent/models.py
+++ b/consent/models.py
@@ -5,11 +5,14 @@ Models for edX Enterprise's Consent application.
 
 from __future__ import absolute_import, unicode_literals
 
+import logging
+
 import six
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 from simple_history.models import HistoricalRecords
 
+from django.core.exceptions import ImproperlyConfigured
 from django.db import models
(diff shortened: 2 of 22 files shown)


BessieSteinberg:
I think the issue is that the only way to know if it has failed to find the right course_id is by calling the discovery api to make sure it's a valid course_id otherwise it will fail silently.  We do cache the response so it shouldn't hit discovery too much

staubina:
@brittneyexline The issue is that you can never really rely on parsing the Course Run Key for the Course Key.  It is user generated and prone to errors and changes that were unintended.

brittneyexline:
@staubina @BessieSteinberg understood - i was just hoping there was more we could do to reduce API calls. it is something we'll have to revisit anyway, as they are scattered throughout the enrollment workflow and DSC.

tuchfarber:
If we're worried about API calls, can we cache these with a really long cache valid time since they really shouldn't change? Since it's just a short string for the key and value, it shouldn't take too much space, regardless of how many courses we're caching.

brittneyexline:
@tuchfarber i think that's a good idea. it looks like `get_edx_api_data` in edx-platform, which is what gets used in the client here to make api calls, has a param called `long_term_cache` that will use a longer cache ttl: https://github.com/edx/edx-platform/blob/master/openedx/core/lib/edx_api_utils.py#L22

i checked the regular ttl vs. the long ttl, it's 1 hr vs. 24 hrs, so maybe we could thread that param through for the calls we need here?

BessieSteinberg:
@brittneyexline That makes sense - I'll make it use the long cache

staubina:
We gave the calls from the LMS something like a 24-hour cache.  It sounds very reasonable to cache this for a long time. This data should not change that frequently.

PR #465, opened 2019-05-08
ENT-1783: Remove usages of get_decoded_jwt_from_request from rbac in favor of get_decoded_jwt from edx-drf-extensions

saleem-latif:
**Description:** 
Remove usages of get_decoded_jwt_from_request from rbac in favor of get_decoded_jwt from edx-drf-extensions

**JIRA:** [ENT-1783](https://openedx.atlassian.net/browse/ENT-1783)

**Installation instructions:**  Simply install `edx-enterprise` using this branch.

**Testing instructions:**
This PR is code refactoring, so nothing specifically to test. RBAC stuff works as expected.

**Merge checklist:**

- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added

**Post merge:**
- [ ] Create a tag


enterprise/rules.py
```diff
@@ -30,7 +26,7 @@ def has_implicit_access_to_dashboard(user, obj):  # pylint: disable=unused-argum
         boolean: whether the request user has access or not
     """
     request = get_request_or_stub()
```
brittneyexline: looking back at the ticket, i see this line: 
> I don’t think `get_request_or_stub` should be a function in `edx-rbac`.  Most likely the client can just use  `crum.get_current_request()` directly.

probably makes sense to take care of that in this PR too.

saleem-latif: I have added the same function in `enterprise.utils` and updated the import. I do not think we can use `crum.get_current_request` directly, since `edx_rbac.utils. get_request_or_stub` adds some additional functionality. Can you take another look at the changes.

enterprise/utils.py
```diff
@@ -852,3 +854,32 @@ def get_default_catalog_content_filter():
     Return default enterprise customer catalog content filter.
     """
     return settings.ENTERPRISE_CUSTOMER_CATALOG_DEFAULT_CONTENT_FILTER or DEFAULT_CATALOG_CONTENT_FILTER
+
+
+def get_request_or_stub():
```
mushtaqak: I think we should mention taken from edx-platform ?

enterprise/rules.py
brittneyexline: i don't think we should conflate testing needs with this functionality - i would prefer we update our tests appropriately instead of baking it into a function we call for production code. i would encourage you to go down the route of calling `crum.get_current_request()` directly here and changing test logic to get the test request object you need.

enterprise/utils.py
brittneyexline: as discussed, this function should be retired if possible.

PR #466, opened 2019-05-08
ARCH-727 Adding sync_learner_profile_data flag to data returned by enterprise-learner endpoint.

douglashall:
The new account settings MFE needs access to the sync_learner_profile_data flag in order to determine which fields should be editable/visible on the account settings form.

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -267,6 +267,21 @@ class EnterpriseCustomer(TimeStampedModel):
         except ObjectDoesNotExist:
             return None
 
+    @property
+    def sync_learner_profile_data(self):
+        """
+        Return the sync_learner_profile data flag for the identity provider associated with this enterprise customer.
+
+        Returns False if enterprise customer does not have any identity provider.
+        """
+        try:
+            return (
+                self.enterprise_customer_identity_provider and
+                self.enterprise_customer_identity_provider.sync_learner_profile_data
+            )
+        except ObjectDoesNotExist:
+            return False
+
     def __str__(self):
         """
         Return human-readable string representation.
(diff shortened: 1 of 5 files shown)


enterprise/models.py
```diff
@@ -1015,14 +1030,30 @@ def __repr__(self):
...
+        identity_provider = self.identity_provider
         return identity_provider and identity_provider.name
 
+    @property
+    def sync_learner_profile_data(self):
+        """
+        Readable name for the identity provider.
+        """
+        identity_provider = self.identity_provider
+        return identity_provider is not None and identity_provider.sync_learner_profile_data
```
nasthagiri: Q: Is there a test that tests the 2nd half of the `and` clause?

douglashall: Nope, do we need one?

enterprise/models.py
```diff
@@ -1015,14 +1030,30 @@ def __repr__(self):
...
         Readable name for the identity provider.
         """
-        identity_provider = utils.get_identity_provider(self.provider_id)
+        identity_provider = self.identity_provider
         return identity_provider and identity_provider.name
 
+    @property
+    def sync_learner_profile_data(self):
+        """
+        Readable name for the identity provider.
```
christopappas: This reads the same as `provider_name`, guessing a copy-paste error

enterprise/models.py
```diff
@@ -1015,14 +1030,30 @@ def __repr__(self):
...
         Readable name for the identity provider.
         """
-        identity_provider = utils.get_identity_provider(self.provider_id)
+        identity_provider = self.identity_provider
         return identity_provider and identity_provider.name
 
+    @property
+    def sync_learner_profile_data(self):
+        """
+        Should learner profile data received from the identity provider be synced to the edX profile.
```
christopappas: sorry to nitpick again, but, should this say `Should return...`? it sounds like a question to me when I read it

christopappas: ooooh i see what it's saying. hmm

christopappas: My personal recommendation would be to replace "Should" with "Return bool indicating if" (or something like that)

douglashall:
@christopappas Is this looking good now?

PR #468, opened 2019-05-10
clean up rbac authorization related waffle switches and logic

muhammad-ammar:
**Description:**  Removed all the role based accessed control switch OFF logic as switch is now permanently enabled.

**JIRA:** [ENT-1885](https://openedx.atlassian.net/browse/ENT-1885)


enterprise/api/v1/permissions.py
```diff
@@ -30,38 +24,6 @@ def has_permission(self, request, view):
...
-class HasEnterpriseEnrollmentAPIAccess(IsAdminUserOrInGroup):
-    """
-    Find the requesting user has access to the Enterprise Enrollment API feature set.
-    """
-    def __init__(self):
-        """ Initialize the class with a API_ALLOWED_GROUPS """
-        self.ALLOWED_API_GROUPS = [u'enterprise_enrollment_api_access', ]  # pylint: disable=invalid-name
-
-
 class HasEnterpriseDataAPIAccess(IsInEnterpriseGroup):
```
brittneyexline: can this class be deleted too?

enterprise/api/v1/views.py
```diff
@@ -214,10 +207,7 @@ def course_enrollments(self, request, pk):
         return Response(serializer.errors, status=HTTP_400_BAD_REQUEST)
 
     @method_decorator(require_at_least_one_query_parameter('permissions'))
-    @list_route(permission_classes=[
-        permissions.IsAuthenticated,
-        IsInEnterpriseGroup,
-    ])
+    @list_route(permission_classes=[permissions.IsAuthenticated, IsInEnterpriseGroup])
     def with_access_to(self, request, *args, **kwargs):  # pylint: disable=invalid-name,unused-argument
```
brittneyexline: i would say that we should delete this because we don't need it anymore, but i think the right approach would be to introduce a v2 enterprise api that doesn't include this endpoint.

tests/test_enterprise/api/test_views.py
```diff
@@ -663,6 +669,7 @@ def test_enterprise_customer_courses(
         """
         Make sure the enterprise courses view returns correct data.
         """
+        self.set_jwt_cookie(ENTERPRISE_ADMIN_ROLE, enterprise_uuid)
```
brittneyexline: why not just globally set the role to the operator in setup so we don't have to manually do this in each test? would that work?

enterprise/api/v1/views.py
muhammad-ammar: I am not sure what to do here. I think creating v2 is out of scope and also I am not sure what exactly we need to add into V2? Should we copy the existing V1 code to V2 and remove the `with_access_to` route?

brittneyexline: sorry, i should have clarified. this was meant to be informational/a note for the future. i don't know when the right time to create a v2 api is, but i just want us to start thinking about it.

muhammad-ammar: Thanks a lot for the clarification. So should I remove the endpoint and permission classes or leave them?

muhammad-ammar:
@brittneyexline This needs a final review.

enterprise/api/v1/permissions.py
brittneyexline: following up - is there a reason we are keeping this class around?

enterprise/api/v1/views.py
brittneyexline: leave this for now.

enterprise/api/v1/permissions.py
muhammad-ammar: This class is used by tests [here](https://github.com/edx/edx-enterprise/blob/master/tests/test_enterprise/api/test_permissions.py#L88-L95)

muhammad-ammar: Removed the class and made equivalent changes in tests.

PR #470, opened 2019-05-13
WL-1927: course-list API which lists courses in all catalogs of active enterprise customer

ziafazal:
**Description:** In order to display course metadata in cornerstone LMS we should provide a RESTful API which is going to be consumed by CSOD on daily basis to sync course catalogs of each customer(enterprise). This PR has changes to provide RESTful API which lists courses in all catalogs of active enterprise customer.

**JIRA:** https://openedx.atlassian.net/browse/WL-1927

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


integrated_channels/cornerstone/exporters/content_metadata.py
```diff
@@ -0,0 +1,108 @@
...
+                minutes = "{:02d}".format(int(60 * fraction))
+                duration = "{hours}:{minutes}:00".format(hours=hours, minutes=minutes)
+
+        return duration
+
+    def transform_languages(self, content_metadata_item):
+        """
+        Return the languages supported by course or `English` as default if no languages found.
+        """
+        return content_metadata_item.get('languages', ['English'])
```
georgebabey: should this be `content_language`? I see we use that in the SAPSF exporter...

integrated_channels/cornerstone/models.py
```diff
@@ -0,0 +1,203 @@
...
+
+LOGGER = getLogger(__name__)
+
+
+@python_2_unicode_compatible
+class CornerstoneGlobalConfiguration(ConfigurationModel):
+    """
+    The global configuration for integrating with Cornerstone.
+
+    .. no_pii:
```
georgebabey: 👍

integrated_channels/cornerstone/exporters/content_metadata.py
ziafazal: `/api/v1/search/all/` API in discovery does not return `content_language` field. Wonder why we use it in SAPSF exporter.

PR #474, opened 2019-05-22
api-gateway: remove old-style catalog endpoints

muhammad-ammar:
**Description:** Remove old-style enterprise catalogs endpoints from api-gateway

**JIRA:** [ENT-1947](https://openedx.atlassian.net/browse/ENT-1947)

**Note to Reviewers:** How to test and deploy changes in this PR?

### DIFF
--- a/api-compact.yaml
+++ b/api-compact.yaml
@@ -7,9 +7,6 @@
 #
 #
 # Available service endpoints -- note that alternate endpoints may be presented at the API Gateway tier
-#   /enterprise/api/v1/catalogs/
-#   /enterprise/api/v1/catalogs/{id}/
-#   /enterprise/api/v1/catalogs/{id}/courses/
 #   /enterprise/api/v1/enterprise_catalogs/
 #   /enterprise/api/v1/enterprise_catalogs/{uuid}/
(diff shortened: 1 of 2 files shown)


api-compact.yaml
```diff
@@ -191,18 +188,6 @@ auth_header: &auth_header
   required: true
   type: "string"
 
-limit_qs_parameter: &limit_qs_parameter
```
brittneyexline: are we removing these query params because they only apply to the catalog endpoints? seems like they could apply to any endpoint... and in fact these are the preferred query params to use for pagination.

api.yaml
```diff
@@ -681,169 +457,159 @@ endpoints:
...
               statusCode: "400"
           type: http
           uri: "https://${stageVariables.enterprise_host}/enterprise/api/v1/enterprise-customer/{uuid}/course_enrollments/"
-id_parameter: 
+id_parameter:
   in: path
   name: id
   required: true
   type: number
-limit_qs_parameter:
```
brittneyexline: same, should we delete these parameters?

muhammad-ammar:
@brittneyexline  I have added the delete query params back.

PR #490, opened 2019-05-31
ENT-1054: Fix enrollment page course key issues

BessieSteinberg:
The enterprise enrollment page (```enterprise/<enterprise-customer-id>/course/<course-id>/enroll/```) is failing in certain edge cases where the ```course_run_id``` and the ```course_id``` don't follow the expected format.

*Some back story to clarify the problem*:  Traditionally ```course_id```s have been substrings of ```course_run_id```s.  For example, for the course ```edX+DemoX``` has a course run ```course-v1:edX+DemoX+Demonstration_Course```.  However, ```course_id```s do not have to be substrings of their ```course_run_id```s.  You may have a course ```edX+DemoX``` with a course run ```course-v1:edX+DemoY+run_Y```.  Unfortunately, the ```edx-enterprise``` code assumes that ```course_id```s are always substrings of ```course_run_id```s.  Let's say a user tries to enroll in the course run ```course-v1:edX+DemoY+run_Y``` through the enterprise enrollment URL.  In order to populate that page we need to get data on the course associated with that course run.  The code tries to find that course by calling ```parse_course_key()``` and gets ```edX+DemoY```.  That course does not exist and causes problems.

*The solution*:  The discovery service understands the relationship between courses and course runs.  To fix this issue, instead of calling ```parse_course_key()``` we need to makea  call to the discovery service.  Specifically, the ```course_runs``` endpoint which takes a ```course_run_id``` and in the JSON blob it returns the ```course_id```.

Disclaimer: I know that ```course_id``` and ```course_run_id``` are not the way we reference those concepts in other parts of the code, but it is the way it's referenced in most of ```edx-enterprsie``` repo and I didn't want to change it

*Additionally*: When this change was originially rolled out to prod it there was a bug in the enterprise-catalogs api. This happened because I didn't not initially handle every case that ```parse_course_key()``` handled correctly.  Specifically, when ```parse_course_key()``` gets a course id as an argument it just returns it.  In my initial implementation, ```parse_course_key()``` returned ```None``` when given a course id.  This has now been fixed.

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/consent/models.py
+++ b/consent/models.py
@@ -5,11 +5,14 @@ Models for edX Enterprise's Consent application.
 
 from __future__ import absolute_import, unicode_literals
 
+import logging
+
 import six
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 from simple_history.models import HistoricalRecords
 
+from django.core.exceptions import ImproperlyConfigured
 from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.functional import lazy
@@ -20,10 +23,11 @@ from model_utils.models import TimeStampedModel
 
 from consent.errors import InvalidProxyConsent
 from consent.mixins import ConsentModelMixin
+from enterprise.api_client.discovery import get_course_catalog_api_service_client
 from enterprise.models import EnterpriseCustomer
-from enterprise.utils import parse_course_key
 
 mark_safe_lazy = lazy(mark_safe, six.text_type)  # pylint: disable=invalid-name
+LOGGER = logging.getLogger(__name__)
 
 
 class DataSharingConsentQuerySet(models.query.QuerySet):
@@ -58,7 +62,18 @@ class DataSharingConsentQuerySet(models.query.QuerySet):
                 except DataSharingConsent.DoesNotExist:
                     # A record for the course run didn't exist, so modify the query
                     # parameters to look for just a course record on the second pass.
-                    kwargs['course_id'] = parse_course_key(course_run_key)
+
+                    site = None
+                    if 'enterprise_customer' in kwargs:
+                        site = kwargs['enterprise_customer'].site
+
+                    try:
+                        course_id = get_course_catalog_api_service_client(site=site).get_course_id(
+                            course_identifier=course_run_key
+                        )
+                        kwargs['course_id'] = course_id
+                    except ImproperlyConfigured:
+                        LOGGER.warning('CourseCatalogApiServiceClient is improperly configured.')
 
         try:
             return self.get(*args, **kwargs)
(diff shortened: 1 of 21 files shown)


consent/models.py
```diff
@@ -58,7 +62,18 @@ def proxied_get(self, *args, **kwargs):
                 except DataSharingConsent.DoesNotExist:
                     # A record for the course run didn't exist, so modify the query
                     # parameters to look for just a course record on the second pass.
-                    kwargs['course_id'] = parse_course_key(course_run_key)
+
```
tuchfarber: Since your `.get_course_id()` already accepts a course_id or a course_run_id, we can ditch a lot of the duplicated logic here by just squishing away the first checks and relying on the `get_course_id`. I don't think it's worth a ticket, but maybe just a `NOTE` or `TODO` about cleaning this method up. 

(I don't think the work should be part of this PR, so we can get this over the line)

enterprise/api_client/discovery.py
```diff
@@ -155,6 +152,38 @@ def get_catalog_results(self, content_filter_query, query_params=None, traverse_
...
+
+        # If here, `course_identifier` must be a course run id.
+        # We cannot use `CourseKey.from_string` to find the course id because that method assumes the course id is
+        # always a substring of the course run id and this is not always the case.  The only reliable way to determine
+        # which courses are associated with a given course run id is by by calling the discovery service.
+        course_run_data = self.get_course_run(course_identifier)
+        if 'course' in course_run_data:
+            return course_run_data['course']
+
+        return None
```
tuchfarber: `parse_course_key` only ever returned a string. So even though `None` makes sense in the abstract, does this bite us in any of the downstream calls of `get_course_id()`, since it's passing a false-y value instead of simply an incorrect string? I tried to follow some of the code paths and it seemed okay.. Just wanted to call it out since `get_course_id` isn't just a corrected version of `parse_course_key`.

enterprise/views.py
```diff
@@ -380,6 +383,60 @@ def get_course_or_program_context(self, enterprise_customer, course_id=None, pro
             })
         return context_data
 
+    def get_page_language_context_data(
```
brittneyexline: i don't understand how these changes relate to the ticket?

BessieSteinberg: annoyingly with my additions the `get()` method failed quality tests because it was too long so I had to pull out some of the statements into another function

enterprise/api_client/discovery.py
BessieSteinberg: yeah that is valid.  I just wasn't sure what to do in the case that course_run_data did not have 'course' defined.  I chose `None` over an empty string because i figured returning `None` would fail louder and be easier to track down those bugs.  But I am open to other suggestions

consent/models.py
BessieSteinberg: sounds good i'll make a ticket and add a `TODO`

BessieSteinberg:
@brittneyexline I'm spinning up a sandbox now so I can test the bug there!

PR #499, opened 2019-06-12
Upgrade django-config-models to unblock DRF upgrades

nedbat:
**Description:** 

We are trying to upgrade Django REST Framework in edx-platform.  This is needed to make it possible.

I also made "make upgrade" more flexible, and cleaned up the README a tad.
**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag


nedbat:
I'm about to make another PR, so I'll wait to bump the version.

PR #503, opened 2019-06-17
Fix Issue: Calling Search API with ```None```

BessieSteinberg:
The ```edx-enterprise``` code could get into a state where it called the ```search/all``` endpoint in ```discovery``` with the query params ```key: [None]```.
This happened because the method ```get_course_id``` returns ```None`` when passed a ```course_identifier``` that has no valid ```course_key```.

The fix:
- Log when ```get_course_id``` cannot find a ```course_key```
- Filter out the course key ```None``` before calling discovery's search endpoint.  If ```None``` is the only course_key in the list, do not call discovery.

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -1322,9 +1322,12 @@ class EnterpriseCustomerCatalog(TimeStampedModel):
         catalog_client = get_course_catalog_api_service_client(self.enterprise_customer.site)
         course_keys = {catalog_client.get_course_id(k) for k in content_ids}
 
+        # Remove `None` from set of course_keys if present
+        course_keys = course_keys.difference({None})
+
         content_ids_in_catalog = self.content_filter_ids
         if not content_ids_in_catalog:
-            content_ids_in_catalog = self._filter_members('key', list(course_keys))
+            content_ids_in_catalog = self._filter_members('key', list(course_keys)) if course_keys else set()
 
         return set(content_ids).issubset(content_ids_in_catalog) or course_keys.issubset(content_ids_in_catalog)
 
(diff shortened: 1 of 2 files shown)


enterprise/models.py
```diff
@@ -1322,9 +1322,12 @@ def contains_courses(self, content_ids):
         catalog_client = get_course_catalog_api_service_client(self.enterprise_customer.site)
         course_keys = {catalog_client.get_course_id(k) for k in content_ids}
 
+        # Remove `None` from set of course_keys if present
+        course_keys = course_keys.difference({None})
+
         content_ids_in_catalog = self.content_filter_ids
         if not content_ids_in_catalog:
-            content_ids_in_catalog = self._filter_members('key', list(course_keys))
+            content_ids_in_catalog = self._filter_members('key', list(course_keys)) if course_keys else []
```
tuchfarber: `_filter_members()` returns a set not a list, and we use set methods on it below, so we need `else set()` instead of `else []`

BessieSteinberg: good catch! thanks!

georgebabey:
This looks good but I think we should have some tests for `contains_courses` around the removal of `None`  Both to catch something like Matt mentioned but also make sure future refactorings don't unintentionally break that.

BessieSteinberg:
good point @georgebabey I'll add some tests

PR #507, opened 2019-06-19
Warn if our pins get out of sync with edx-platform

nedbat:
**Description:** This is a replacement for #506 

That PR was getting complicated. This is a simpler approach to keeping the pinned versions here in sync with edx-platform.  Now during `make upgrade` (or as a separate make target) the versions from edx-platform are compared to the local pins in `test-master.in`.  Differences are warned.  No other action is taken.

**JIRA:** None

**Dependencies:** None

**Merge deadline:** Soon?

**Installation instructions:** Nothing unusual

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.


requirements/check-pins.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python
+#
+# Compare two pip pin files.  Everything in the first one should match one
+# in the second one.
+
+import re
+import sys
+
+def pins(lines):
+    for line in lines:
```
jmbowman: pylint probably isn't running on this file, may want to update its arguments in `tox.ini` to fix that.

requirements/edx-platform-constraints.txt
```diff
@@ -0,0 +1,238 @@
+### DON'T edit this file, it's copied from edx-platform. See make upgrade
```
jmbowman: Do we want to commit this, or should it be listed in `.gitignore`?  Seems like it's not strictly needed except when generated during `make upgrade` execution.

requirements/check-pins.py
nedbat: I've added pylint.

requirements/edx-platform-constraints.txt
nedbat: I figured checking it in would make it easier for named releases, or tracking down problems.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -14,12 +14,13 @@ instructions.
...
 3. Expect C to happen
-4. If D happend instead - check failed.
+4. If D happened instead - check failed.
 
 **Merge checklist:**
 
-- [ ] Check that the versions of the requirements in the `platform-master.in` file match edx-platform.
-- [ ] New requirements are in the right place (`base.in` if only used in enterprise; in the correct `platform-****.in` files if they're hosted in edx-platform)
+- [ ] New requirements are in the right place
+    - `base.in` if needed in production, but edx-platform doesn't install it.
```
christopappas: The `but edx-platform doesn't install it.` bit here... I thought we decided that requirements should be in base.in if they are used at all in edx-enterprise (regardless if they are also installed in edx-platform)

nedbat: This has been confusing because I started over with a simpler approach.  In this PR, `test-master.in` is the set of packages that are shared with edx-platform.  `base.in` is for packages only used by edx-enterprise.

PR #509, opened 2019-06-20
tests for get_course_id

BessieSteinberg:

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -1014,6 +1014,56 @@ class TestEnterpriseCustomerCatalog(unittest.TestCase):
         enterprise_customer_catalog = factories.EnterpriseCustomerCatalogFactory()
         assert enterprise_customer_catalog.get_course_and_course_run('fake-course-run-id') == (None, None)
 
+    @mock.patch('enterprise.api_client.discovery.CourseCatalogApiServiceClient')
+    def test_contains_courses(self, course_catalog_client_mock):
+        content_ids = [
+            "course-v1:JediAcademy+AppliedTelekinesis+T1",
+            "course-v1:TrantorAcademy+Psychohistory101+T1",
+            "course-v1:StarfleetAcademy+WarpspeedNavigation+T2337",
+            "course-v1:SinhonCompanionAcademy+Calligraphy+TermUnknown",
+            "course-v1:CampArthurCurrie+HeavyWeapons+T2245_5",
+        ]
+        course_catalog_client_mock.return_value.get_course_id.side_effect = content_ids
+        course_catalog_client_mock.return_value.get_catalog_results.return_value = {
+            'results': [{'key': content_id} for content_id in content_ids]
+        }
+
+        enterprise_customer_catalog = EnterpriseCustomerCatalog(
+            uuid=self.faker.uuid4(),  # pylint: disable=no-member
+            enterprise_customer=factories.EnterpriseCustomerFactory()
+        )
+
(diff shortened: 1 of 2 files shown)


tests/test_models.py
```diff
@@ -1014,6 +1014,56 @@ def test_get_course_and_course_run_no_content_items(self, contains_courses_mock)
...
+            "course-v1:StarfleetAcademy+WarpspeedNavigation+T2337",
+            "course-v1:SinhonCompanionAcademy+Calligraphy+TermUnknown",
+            "course-v1:CampArthurCurrie+HeavyWeapons+T2245_5",
+        ]
+        course_catalog_client_mock.return_value.get_course_id.return_value = None
+        course_catalog_client_mock.return_value.get_catalog_results.return_value = {
+            'results': [{'key': content_id} for content_id in content_ids]
+        }
+
+        enterprise_customer_catalog = EnterpriseCustomerCatalog(
```
tuchfarber: Can use `EnterpriseCustomerCatalogFactory` which does this for you.

tests/test_models.py
```diff
@@ -1014,6 +1014,56 @@ def test_get_course_and_course_run_no_content_items(self, contains_courses_mock)
...
+
+        contains_courses = enterprise_customer_catalog.contains_courses(content_ids)
+
+        assert contains_courses is True
+
+    @mock.patch('enterprise.api_client.discovery.CourseCatalogApiServiceClient')
+    def test_contains_courses_invalid_content_ids(self, course_catalog_client_mock):
+        """
+        Assert that contains courses returns False if content_ids are invalid.  In this case 'invalid' means that the
+        discovery service cannot find a course for the given course run id.
```
tuchfarber: I'd add a note for future us that this can happen when calling a discovery site that doesn't match the current site.

tests/test_models.py
```diff
@@ -1014,6 +1014,56 @@ def test_get_course_and_course_run_no_content_items(self, contains_courses_mock)
...
+        }
+
+        enterprise_customer_catalog = EnterpriseCustomerCatalog(
+            uuid=self.faker.uuid4(),  # pylint: disable=no-member
+            enterprise_customer=factories.EnterpriseCustomerFactory()
+        )
+
+        contains_courses = enterprise_customer_catalog.contains_courses(content_ids)
+
+        assert contains_courses is True
```
tuchfarber: This seems to be a testing an odd use case where:
1. Discovery correctly returns a catalog (not cross-site issues)
2. The course_run_ids fail because the course_ids don't exist
3. We're still going to show them DSC because the empty set of courses is in the non-empty set of catalog courses. 

This makes we wonder if we should just short-circuit and return `False` earlier if the `course_keys` is an empty set. I can't see a situation where we wouldn't be able to get the `course_id` from a `course_run_id` and would still want to say it's in the catalog. Perhaps @georgebabey can chime in. 

Once that's sorted, an additional test where the course *isn't* in the catalog would be helpful as well since both of these are `True`.

tuchfarber:
Why was this merged?

BessieSteinberg:
Really sorry about that - It was a mistake here is the revert PR: https://github.com/edx/edx-enterprise/pull/513

PR #510, opened 2019-06-24
ENT-2009: Changed page size of inactive user results from 500 to 1000

ziafazal:
Changed page size of inactive user results from 500 to 1000 to avoid timeouts for those customers who has large number of in-active users.

**JIRA:** https://openedx.atlassian.net/browse/ENT-2009

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[1.6.18] - 2019-06-24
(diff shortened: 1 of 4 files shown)


hasnain-naveed:
👍

PR #511, opened 2019-06-24
ENT-1598 | Added course duration info in degreed course description.

hasnain-naveed:
**Description:** Added course duration info in degreed course description.

**JIRA:** ENT-1598

**Dependencies:** This [PR](https://github.com/edx/course-discovery/pull/1970)

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/integrated_channels/degreed/exporters/content_metadata.py
+++ b/integrated_channels/degreed/exporters/content_metadata.py
@@ -7,6 +7,7 @@ from __future__ import absolute_import, unicode_literals
 
 from logging import getLogger
 
+from enterprise.utils import get_closest_course_run, get_course_run_duration_info
 from integrated_channels.integrated_channel.exporters.content_metadata import ContentMetadataExporter
 
 LOGGER = getLogger(__name__)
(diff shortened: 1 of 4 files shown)


integrated_channels/degreed/exporters/content_metadata.py
```diff
@@ -37,10 +49,24 @@ def transform_description(self, content_metadata_item):
...
         """
-        full_description = content_metadata_item.get('full_description') or ''
-        if 0 < len(full_description) <= self.LONG_STRING_LIMIT:  # pylint: disable=len-as-condition
-            return full_description
-        return content_metadata_item.get('short_description') or content_metadata_item.get('title') or ''
+        course_runs = content_metadata_item.get('course_runs')
+        course_run_duration_info = get_course_run_duration_info(
+            get_closest_course_run(course_runs)
+        ) if course_runs else ''
+        description = add_duration_info(
```
ziafazal: I thin we should not call `add_duration_info ` twice here. I think we can simplify logic here something like this
```
if 0 < len(full_description + course_run_duration_info) <= self.LONG_STRING_LIMIT:
    course_description = full_description
else:
   course_description = content_metadata_item.get('short_description') or content_metadata_item.get('title') or ''
return "{duration_info}{description}".format(duration_info=duration_info, description=course_description)
```
we probably don't need `add_duration_info` method

PR #521, opened 2019-07-12
use edx-drf-extensions jwt utils

muhammad-ammar:
**Description:** This upgrades the edx-rbac to latest version and as a result of this upgrade we are using jwt utils from edx-drf-extensions package because those jwt utils are now removed from edx-rbac.
 
**JIRA:** [ENT-2002](https://openedx.atlassian.net/browse/ENT-2002)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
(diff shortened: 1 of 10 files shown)


muhammad-ammar:
@brittneyexline Can we merge this now? `edx-enterprise-data` and `ecommerce` changes are already merged and live.

PR #523, opened 2019-07-15
Don't use pinned requirements

davestgermain:
The strict requirements in base.txt are causing edx-platform's `make upgrade` to fail. This changes the requirements to use base.in, instead.

### DIFF
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "1.6.23"
+__version__ = "1.6.24"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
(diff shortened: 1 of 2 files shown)


davestgermain:
@brittneyexline probably the best thing would be to run `make upgrade` in edx-platform and track down the failures. It seems that people have been bumping the edx-enterprise version by changing the requirements files rather than running `make upgrade`, otherwise they would've seen the problem. 

I think Enterprise should untangle this dependency mess.

cpennington:
`edx-enterprise` is a library? Why is it pinning depencies for install? If it needs to pin the testing, at should specifies ranges for use by tox

brittneyexline:
@davestgermain i'm going to take a stab at the dependency issues.
@cpennington there's not really a single reason or decision i could point to. we thought for a while when this was first built that it might be split out to its own service, but kept it as a library specifically meant for edx-platform as a first pass. over time, more has been added and dependencies were pinned in (poor) attempts to stay in sync with edx-platform, or to satisfy other feature requirements. no one is really happy about the current situation which has had increased complications, but i'm open to feedback on how to better manage it. my personal opinion is that it's time for edx-enterprise to graduate to its own service, but we're not going to get there overnight.

brittneyexline:
Follow up pr can be found here: https://github.com/edx/edx-enterprise/pull/525
This is meant to unblock, but the underlying issues will need to be addressed separately. I'll follow up with a ticket.

PR #525, opened 2019-07-15
Resolving conflicts in make upgrade

brittneyexline:
**Description:**
tl;dr, we should find a way to stop pinning as many dependencies as we can to avoid update conflicts with edx-platform.

Longer:
I started by running make upgrade in edx-platform against this branch, and adding things that failed as a constraint in requirements/test-master.in, and then running make upgrade in edx-enterprise, then repeating the process. There were several issues going on, but the main problem is that we are trying to stay in sync with edx-platform, but we pin explicitly when a package is upgraded, and it may be constrained for some reason in edx-platform. It took a while to come to this, and there is a corresponding change necessary in edx-platform because of transifex-client, which we would have had to resolve anyway in edx-platform.

Here is the corresponding edx-platform branch I used to test make upgrade: https://github.com/edx/edx-platform/pull/21132

**Merge deadline:** ASAP - we are blocking make upgrade from being run without errors in edx-platform.

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[1.7.0] - 2019-07-15
+--------------------
+
+* Pin certain constraints from edx-platform so that edx-enterprise will install properly there.
+
 [1.6.23] - 2019-07-15
 ---------------------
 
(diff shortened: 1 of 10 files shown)


brittneyexline:
@muhammad-ammar @mushtaqak @irfanuddinahmad - please feel free to merge and try upgrading in edx-platform when you come online, hopefully everything should work 🤞

PR #529, opened 2019-07-17
Enterprise catalog queries

mushtaqak:
[ENT-2066](https://openedx.atlassian.net/browse/ENT-2066)

Sandbox: https://mammar.sandbox.edx.org/admin/enterprise/enterprisecustomercatalog/


mushtaqak:
@brianhw Can you have a look at this? Thanks

enterprise/static/enterprise/admin/enterprise_customer_catalog.js
```diff
@@ -0,0 +1,28 @@
...
+
+        $("[name='enterprise_catalog_query']").on("change", function(event){
+            catalogQueryId = $(event.target).val();
+            if (!ignoreUpdate && catalogQueryId) {
+                // Extract content_filter for the selected catalogQueryId.
+                $.ajax({
+                    url: catalogQueryApiUrl + catalogQueryId,
+                    method: "get"
+                }).success(function (catalogQueryResponse) {
+                    $contentFilterEl.val(JSON.stringify(catalogQueryResponse));
```
muhammad-ammar: It would be great if we can indent the content filter? 
https://stackoverflow.com/questions/3515523/javascript-how-to-generate-formatted-easy-to-read-json-straight-from-an-object

enterprise/models.py
```diff
@@ -1169,6 +1169,45 @@ def __repr__(self):
         return self.__str__()
 
 
+@python_2_unicode_compatible
+class EnterpriseCatalogQuery(TimeStampedModel):
+    """
+    Store catalog query.
```
muhammad-ammar: We need to add more detail to clearly explain the purpose of this model.

enterprise/models.py
```diff
@@ -1169,6 +1169,45 @@ def __repr__(self):
...
+        null=False
+    )
+    content_filter = JSONField(
+        default={},
+        blank=True,
+        null=True,
+        load_kwargs={'object_pairs_hook': collections.OrderedDict},
+        help_text=_(
+            "Query parameters which will be used to filter the discovery service's search/all endpoint results, "
+            "specified as a Json object. An empty Json object means that all available content items will be "
```
muhammad-ammar: `Json` -> `JSON`

enterprise/api/v1/views.py
```diff
@@ -456,6 +456,25 @@ class EnterpriseCustomerReportingConfigurationViewSet(EnterpriseReadOnlyModelVie
     ordering_fields = FIELDS
 
 
+class CatalogQueryView(APIView):
+    """
+    View for fetching enterprise catalog query.
+    """
+    authentication_classes = [SessionAuthentication]
```
muhammad-ammar: Usually we use `(JwtAuthentication, BearerAuthentication, SessionAuthentication,)` as `authentication_classes`. Don't we need to support other two?

mushtaqak: We don't really need to support other two since this will only be called via djanog admin tool

mushtaqak:
>The url is https://discovery-mammar.sandbox.edx.org/api/v1search/all/? which is incorrect. / is missing between v1 and search. I think url should be https://discovery-mammar.sandbox.edx.org/api/v1/search/all/?

@muhammad-ammar I just re-used the existing code that is running in production, also pointed out this in our yesterdays' discussion.

enterprise/api/v1/views.py
```diff
@@ -456,6 +456,27 @@ class EnterpriseCustomerReportingConfigurationViewSet(EnterpriseReadOnlyModelVie
...
+    http_method_names = ['get']
+
+    def get(self, request, catalog_query_id):
+        """
+        API endpoint for fetching a enterprise catalog query.
+        """
+        try:
+            catalog_query = models.EnterpriseCatalogQuery.objects.get(pk=catalog_query_id)
+        except models.EnterpriseCatalogQuery.DoesNotExist:
+            return Response({"detail": "Could not find enterprise catalog query."}, status=HTTP_404_NOT_FOUND)
```
irfanuddinahmad: Do we need any logging here?

enterprise/models.py
```diff
@@ -1169,6 +1169,48 @@ def __repr__(self):
         return self.__str__()
 
 
+@python_2_unicode_compatible
+class EnterpriseCatalogQuery(TimeStampedModel):
+    """
+    Stores a re-usable catalog query.
+
+    We will use this saved catalog query in `EnterpriseCustomerCatalog` objects to build catalog's content_filter field.
+    This is basically a saved instance of `content_filter` that can be re-used accross different catalogs.
```
irfanuddinahmad: nit: remove 'basically'
reword: This stored catalog query is used in ...

enterprise/api/v1/views.py
mushtaqak: I don't think this is critical place to monitor plus we will hardly get this error.

enterprise/static/enterprise/admin/enterprise_customer_catalog.js
mushtaqak: I am a bit reluctant in indenting this since it may add up spaces which might bring up issue of max_length for content_filter field when saving.

PR #537, opened 2019-07-24
improve monitoring for DSC flow

muhammad-ammar:
**Description:** Add/Improve logging to monitoring the DSC flow.

**JIRA:** https://openedx.atlassian.net/browse/ENT-1953

### DIFF
--- a/consent/api/v1/views.py
+++ b/consent/api/v1/views.py
@@ -5,6 +5,8 @@ A generic API for edX Enterprise's Consent application.
 
 from __future__ import absolute_import, unicode_literals
 
+from logging import getLogger
+
 from edx_rest_framework_extensions.auth.bearer.authentication import BearerAuthentication
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
 from rest_framework.authentication import SessionAuthentication
@@ -17,6 +19,8 @@ from consent.errors import ConsentAPIRequestError
 from consent.helpers import get_data_sharing_consent
 from enterprise.api.throttles import ServiceUserThrottle
 from enterprise.utils import get_request_value
+
+LOGGER = getLogger(__name__)
 
 
 class DataSharingConsentView(APIView):
@@ -85,12 +89,29 @@ class DataSharingConsentView(APIView):
         Get the consent record relevant to the request at hand.
         """
         username, course_id, program_uuid, enterprise_customer_uuid = self.get_required_query_params(request)
-        return get_data_sharing_consent(
+        dsc = get_data_sharing_consent(
             username,
             enterprise_customer_uuid,
             course_id=course_id,
             program_uuid=program_uuid
         )
+        if not dsc:
+            log_message = (
+                '[Enterprise Consent API] The code was unable to get consent data for the user. '
+                'Course: {course_id}, '
+                'Program: {program_uuid}, '
+                'EnterpriseCustomer: {enterprise_customer_uuid}, '
+                'User: {username}, '
+                'ErrorCode: {error_code}'.format(
+                    course_id=course_id,
+                    program_uuid=program_uuid,
+                    enterprise_customer_uuid=enterprise_customer_uuid,
+                    username=username,
+                    error_code='ENTGDS001',
+                )
+            )
+            LOGGER.error(log_message)
+        return dsc
 
     def get_required_query_params(self, request):
         """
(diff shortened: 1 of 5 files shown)


consent/api/v1/views.py
```diff
@@ -85,12 +89,27 @@ def get_consent_record(self, request):
...
-        return get_data_sharing_consent(
+        dsc = get_data_sharing_consent(
             username,
             enterprise_customer_uuid,
             course_id=course_id,
             program_uuid=program_uuid
         )
+        if not dsc:
+            log_message = (
+                '[Enterprise Consent API Flow] Unable to get data sharing consent. '
```
mushtaqak: nit: `[Enterprise Consent API]`

enterprise/views.py
```diff
@@ -189,7 +189,21 @@ def course_or_program_exist(self, course_id, program_uuid):
             course_exists = course_id and CourseApiClient().get_course_details(course_id)
             program_exists = program_uuid and get_course_catalog_api_service_client().program_exists(program_uuid)
             return course_exists or program_exists
-        except ImproperlyConfigured:
+        except ImproperlyConfigured as error:
+            error_code = 'ENTGDS004'
+            log_message = (
+                '[Enterprise DSC Flow] CourseCatalogApiServiceClient is improperly configured. '
```
mushtaqak: `[Enterprise DSC API]`

enterprise/views.py
```diff
@@ -189,7 +189,21 @@ def course_or_program_exist(self, course_id, program_uuid):
...
+                'CourseId: {course_id}, '
+                'ProgramId: {program_uuid}, '
+                'ErrorCode: {error_code}, '
```
mushtaqak: is this `ErrorCode` important to know in monitoring stuff?

enterprise/views.py
```diff
@@ -360,7 +374,23 @@ def get_course_or_program_context(self, enterprise_customer, course_id=None, pro
             if not self.preview_mode:
                 try:
                     catalog_api_client = get_course_catalog_api_service_client(enterprise_customer.site)
-                except ImproperlyConfigured:
+                except ImproperlyConfigured as error:
+                    error_code = 'ENTGDS004'
+                    log_message = (
+                        '[Enterprise DSC Flow] CourseCatalogApiServiceClient is improperly configured. '
```
mushtaqak: `[Enterprise DSC API]`

enterprise/views.py
mushtaqak: Also can we attach `User` here? from request?

enterprise/views.py
```diff
@@ -452,22 +482,44 @@ def get(self, request):
 
         # Get enterprise_customer to start in case we need to render a custom 404 page
         # Then go through other business logic to determine (and potentially overwrite) the enterprise customer
-        enterprise_customer = get_enterprise_customer_or_404(enterprise_customer_uuid)
+        try:
+            enterprise_customer = get_enterprise_customer_or_404(enterprise_customer_uuid)
+        except Http404:
+            error_code = 'ENTGDS008'
+            log_message = (
+                '[Enterprise DSC Flow] Received incorrect value for enterprise customer in request. '
```
mushtaqak: `[Enterprise DSC API]

mushtaqak: Should we just say `Could not find enterprise customer with provided data` or something like that?

enterprise/views.py
```diff
@@ -452,22 +482,44 @@ def get(self, request):
...
+                    'ProgramId: {program_uuid}, '
+                    'EnterpriseCustomer: {enterprise_customer_uuid}, '
+                    'User: {user_id}, '
+                    'ErrorCode: {error_code} '.format(
                         course_id=course_id,
                         program_uuid=program_uuid,
-                        error_code=error_code,
-                        userid=request.user.id,
                         enterprise_customer_uuid=enterprise_customer_uuid,
+                        user_id=request.user.id,
```
mushtaqak: In above log messages we were using username for `User:` here the id, we should be consistent

enterprise/views.py
```diff
@@ -566,17 +636,22 @@ def get(self, request):
...
+                'EnterpriseCustomer: {enterprise_customer_uuid}, '
+                'User: {user_id}, '
+                'ErrorCode: {error_code}, '
+                'Message: {message}'.format(
                     course_id=course_id,
+                    program_uuid=program_uuid,
+                    enterprise_customer_uuid=enterprise_customer_uuid,
+                    user_id=request.user.id,
+                    error_code=error_code,
+                    message=error,
```
mushtaqak: Should we also add `context_data` here as well?

enterprise/views.py
```diff
@@ -603,39 +678,66 @@ def post(self, request):
...
+                'EnterpriseCustomer: {enterprise_customer_uuid}, '
+                'User: {user_id}, '
+                'ErrorCode: {error_code} '.format(
+                    course_id=course_id,
+                    program_uuid=program_uuid,
+                    enterprise_customer_uuid=enterprise_uuid,
+                    user_id=request.user.id,
+                    error_code=error_code,
+                )
+            )
```
mushtaqak: Let's also add exception Message here.

enterprise/views.py
```diff
@@ -649,16 +751,17 @@ def post(self, request):
...
+                'CourseId: {course_id}, '
+                'ProgramId: {program_uuid}, '
+                'EnterpriseCustomer: {enterprise_customer_uuid}, '
+                'User: {user_id}, '
+                'ErrorCode: {error_code} '.format(
                     course_id=course_id,
+                    program_uuid=program_uuid,
+                    enterprise_customer_uuid=enterprise_uuid,
+                    user_id=request.user.id,
+                    error_code=error_code,
```
mushtaqak: Should we also add `context_data` here as well?

enterprise/views_error_codes.txt
```diff
@@ -10,6 +10,7 @@ ENTGDS004 - Course catalog api configuration error
 ENTGDS005 - Required request values missing for action to be carried out
 ENTGDS006 - The course or program with a given id does not exist
 ENTGDS007 - There is no consent record, or consent is not required
```
muhammad-ammar: @christopappas Any idea why we have duplicate messages with different codes? 
ENTGDS000 and ENTGDS006 are same
ENTGDS002 and ENTGDS007 are same
ENTGDS003 and ENTGDS005 are same

christopappas: The codes themselves are more for indicating the flow in which the error occurred, and not the message that was returned to the user. It happens that some of the messages are the same, because we do similar checks in a few places, but we needed unique codes cause they're in different code paths

PR #544, opened 2019-08-07
ENT-2117 | Adding enterprise_learner_portal djangoapp to house endpoi…

christopappas:
…nts needed for enterprise learner portal frontend

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise_learner_portal/api/v1/serializers.py
+++ b/enterprise_learner_portal/api/v1/serializers.py
@@ -1,0 +1,73 @@
+# -*- coding: utf-8 -*-
+"""
+enterprise_learner_portal serializer
+"""
+from __future__ import absolute_import, unicode_literals
+
+from django.utils.translation import ugettext as _
+
+from rest_framework import serializers
+
+try:
+    from lms.djangoapps.certificates.api import get_certificate_for_user
+    from lms.djangoapps.program_enrollments.api.api import (
+        get_due_dates,
+        get_course_run_url,
+        get_emails_enabled,
+    )
+except ImportError:
+    get_certificate_for_user = None
+    get_due_dates = None
+    get_course_run_url = None
+    get_emails_enabled = None
+
+from enterprise.utils import NotConnectedToOpenEdX
+from enterprise_learner_portal.utils import get_course_run_status
+
+
+class EnterpriseCourseEnrollmentSerializer(serializers.Serializer):
+    """
+    A serializer for course enrollment information for a given course
+    and enterprise customer user.
+    """
+
+    def __init__(self, *args, **kwargs):
+        if get_certificate_for_user is None:
+            raise NotConnectedToOpenEdX(
+                _('To use this EnterpriseCourseEnrollmentSerializer, this package must be '
+                  'installed in an Open edX environment.')
+            )
+        super(EnterpriseCourseEnrollmentSerializer, self).__init__(*args, **kwargs)
+
+    def to_representation(self, instance):
+        representation = super(EnterpriseCourseEnrollmentSerializer, self).to_representation(instance)
+
+        request = self.context['request']
+        user = request.user
+
+        # Certificate
+        certificate_info = get_certificate_for_user(
+            user.username,
+            instance['id']
+        )
+        representation['certificate_download_url'] = certificate_info.get('download_url')
+
+        # Email enabled
+        emails_enabled = get_emails_enabled(user, instance['id'])
+        if emails_enabled is not None:
+            representation['emails_enabled'] = emails_enabled
+
+        representation['course_run_id'] = instance['id']
+        representation['course_run_status'] = get_course_run_status(
+            instance,
+            certificate_info,
+        )
+        representation['start_date'] = instance['start']
+        representation['end_date'] = instance['end']
+        representation['display_name'] = instance['display_name_with_default']
+        representation['course_run_url'] = get_course_run_url(request, instance['id'])
+        representation['due_dates'] = get_due_dates(request, instance['id'], user)
+        representation['pacing'] = instance['pacing']
+        representation['org_name'] = instance['display_org_with_default']
+
+        return representation
--- a/enterprise_learner_portal/api/v1/views.py
+++ b/enterprise_learner_portal/api/v1/views.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+
+from django.contrib.auth import get_user_model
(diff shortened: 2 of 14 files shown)


enterprise_learner_portal/api/v1/serializers.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def to_representation(self, instance):
+        representation = super(EnterpriseCourseEnrollmentSerializer, self).to_representation(instance)
+
+        request = self.context['request']
+        user = self.context['user']
+
+
+
+# course_run_status
```
christopappas: need to write the logic for this still (porting it over from program_enrollment code in the platform)

enterprise_learner_portal/api/v1/views.py
```diff
@@ -0,0 +1,60 @@
...
+    authentication_classes = (JwtAuthentication, BearerAuthentication, SessionAuthentication,)
+
+    def get(self, request, user_email):
+        if get_course_overviews is None:
+            raise NotConnectedToOpenEdX(
+                _('To use this endpoint, this package must be '
+                  'installed in an Open edX environment.')
+            )
+
+        user = get_object_or_404(User, email=user_email)
```
douglashall: I think we want this endpoint to return the enrollments for the authenticated user. The client should not be able to request a specific user with a query param.

enterprise_learner_portal/api/v1/serializers.py
```diff
@@ -0,0 +1,81 @@
...
+from __future__ import absolute_import, unicode_literals
+
+from rest_framework import serializers
+
+try:
+    from lms.djangoapps.certificates.api import get_certificate_for_user
+    from lms.djangoapps.program_enrollments.api.api import (
+        get_due_dates,
+        get_course_run_url,
+        get_emails_enabled,
```
douglashall: How many of these helper functions make DB calls or external API calls? I'm a bit worried about performance, but perhaps those concerns are premature at this point.

christopappas: Some, but not all. https://github.com/edx/edx-platform/pull/21258/files in api/api.py. get_due_dates looks to be one that does

enterprise_learner_portal/api/v1/views.py
christopappas: Is your concern that someone viewing a page on the frontend could get information about other learners?

douglashall: Yes, that would be a concern. We'd need to add an is_staff check in here to allow someone to request another user's enrollments. In general, though, I think this endpoint is designed to service the learner portal and in that context you would never need to request another user's enrollments.

christopappas: Okay. I can definitely make sure `is_staff` is a requirement.

As for the resource the endpoint uses to return data, the endpoint itself is more flexible if we don't limit it to the authenticated user. You could imagine a staff user having the option of checking data for number of users other than one's own. Especially where this is a new endpoint, I don't think we need to add this restriction just yet

enterprise_learner_portal/api/v1/views.py
```diff
@@ -0,0 +1,63 @@
...
+
+User = get_user_model()
+
+
+class EnterpriseCourseEnrollmentView(APIView):
+    permission_classes = (permissions.IsAuthenticated,)
+    authentication_classes = (JwtAuthentication, BearerAuthentication, SessionAuthentication,)
+
+    def get(self, request, user_email):
+        if not request.user.is_staff:
```
douglashall: This will prevent non-staff users from using this endpoint. Keep it simple for now, just return enrollments for the request user. We can make this endpoint more flexible in the future when/if we need to.

christopappas: Can you help me to understand the value of changing what I've already written here? This seems to be more of a matter of preference

christopappas: Actually @brittneyexline , do you have any sense of future product requirements that could guide this decision?

If we don't think we'll ever benefit from the way I've written this endpoint (specifically: being able to provide the email for any user), then I can change it

douglashall: The way you have written it here will prevent this endpoint from being useable by the learner portal. Most enterprise learners will not be staff so when the frontend app makes a request to this endpoint they will get a 403 response.

enterprise_learner_portal/api/v1/views.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def get(self, request):
+        if get_course_overviews is None:
+            raise NotConnectedToOpenEdX(
+                _('To use this endpoint, this package must be '
+                  'installed in an Open edX environment.')
+            )
+
+        user = request.user
+        enterprise_customer_user = get_object_or_404(
```
brittneyexline: i'm wondering if we should be mindful of having multiple enterprise's linked to the learner here. probably it would be best to follow up with separate work if we are trying to support this case... i will bring it up with jennifer to see her thoughts.

enterprise_learner_portal/api/v1/views.py
```diff
@@ -0,0 +1,61 @@
...
+            enterprise_customer_user=enterprise_customer_user
+        ).values_list('course_id', flat=True)
+
+        overviews = get_course_overviews(course_ids_for_ent_enrollments)
+
+        data = EnterpriseCourseEnrollmentSerializer(
+            overviews,
+            many=True,
+            context={
+                'user': enterprise_customer_user.user,
```
brittneyexline: why add user here separately when the user is in the request?

christopappas: good catch, i think this is a vestige of the previous implementation (where we would want to serialize the data of the user whose email was handed to us)

enterprise_learner_portal/api/v1/views.py
brittneyexline: we would like to have whatever we build support users linked to multiple enterprises, though we are not going to change any existing code unaffected by the scope of our project. i would recommend adding an enterprise id as a parameter passed into the path so that the endpoint will only return enrollments scoped to the enterprise. i leave it up to you to file a new story or complete the work with this story... i just don't want to forget about it  😄

PR #545, opened 2019-08-09
ENT-2168 | setting cache for get_catalog_results from discovery

HammadAhmadWaqas:
**Description:** setting cache for get_catalog_results from discovery

**JIRA:** https://openedx.atlassian.net/browse/ENT-2168

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise/api_client/discovery.py
+++ b/enterprise/api_client/discovery.py
@@ -13,9 +13,11 @@ from opaque_keys.edx.keys import CourseKey
 from requests.exceptions import ConnectionError, Timeout  # pylint: disable=redefined-builtin
 
 from django.conf import settings
+from django.core.cache import cache
 from django.core.exceptions import ImproperlyConfigured, ObjectDoesNotExist
 from django.utils.translation import ugettext_lazy as _
 
+from enterprise import utils
 from enterprise.utils import MultipleProgramMatchError, NotConnectedToOpenEdX, get_configuration_value_for_site
 
 try:
@@ -116,10 +118,20 @@ class CourseCatalogApiClient(object):
 
         return results
 
+    def get_catalog_results_from_discovery(self, content_filter_query, query_params=None, traverse_pagination=False):
+        """
+            Return results from the discovery service's search/all endpoint."""
+
+        endpoint = getattr(self.client, self.SEARCH_ALL_ENDPOINT)
+        response = endpoint().post(data=content_filter_query, **query_params)
+        if traverse_pagination:
+            response['results'] = self.traverse_pagination(response, endpoint, content_filter_query, query_params)
+            response['next'] = response['previous'] = None
+        return response
+
     def get_catalog_results(self, content_filter_query, query_params=None, traverse_pagination=False):
         """
-        Return results from the discovery service's search/all endpoint.
-
+            Return results from the cache or discovery service's search/all endpoint.
         Arguments:
             content_filter_query (dict): query parameters used to filter catalog results.
             query_params (dict): query parameters used to paginate results.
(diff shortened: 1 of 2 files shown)


enterprise/api_client/discovery.py
```diff
@@ -132,11 +142,22 @@ def get_catalog_results(self, content_filter_query, query_params=None, traverse_
         query_params = query_params or {}
 
         try:
-            endpoint = getattr(self.client, self.SEARCH_ALL_ENDPOINT)
-            response = endpoint().post(data=content_filter_query, **query_params)
-            if traverse_pagination:
-                response['results'] = self.traverse_pagination(response, endpoint, content_filter_query, query_params)
-                response['next'] = response['previous'] = None
+            cache_key = utils.get_cache_key(
+                service='discovery',
```
ziafazal: nit: no need to pass `service` I think `endpoint` should be sufficient

tests/test_enterprise/api_client/test_discovery.py
```diff
@@ -387,6 +387,25 @@ def test_get_catalog_results(self):
...
+        """
+        Verify `get_catalog_results` of CourseCatalogApiClient works as expected.
+        """
+        content_filter_query = {'content_type': 'course', 'aggregation_key': ['course:edX+DemoX']}
+        self.api.get_catalog_results(content_filter_query=content_filter_query)
+        assert mocked_get_catalog_results_from_discovery.call_count == 1
+
+        # searching same query should not hit discovery service again
+        self.api.get_catalog_results(content_filter_query=content_filter_query)
+        assert mocked_get_catalog_results_from_discovery.call_count == 1
```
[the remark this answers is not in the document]
HammadAhmadWaqas: we are calling `get_catalog_results` in line 401, So It should be one.

enterprise/api_client/discovery.py
HammadAhmadWaqas: Actually, I added `service key` to make sure cache_key is unique, and will not conflict with any other cache_key.
because `endpoint` contains only `search/all/` that is a common endpoint name. so I was afraid It might conflict with other platform services, maybe in the future?

enterprise/api_client/discovery.py
```diff
@@ -132,11 +144,22 @@ def get_catalog_results(self, content_filter_query, query_params=None, traverse_
...
+            )
+            response = cache.get(cache_key)
+            if not response:
+                # Response is not cached, so make a call.
+                response = self._get_catalog_results_from_discovery(
+                    content_filter_query,
+                    query_params,
+                    traverse_pagination
+                )
+                cache.set(cache_key, response, settings.DISCOVERY_GET_CATALOG_RESULT_CACHE_TIMEOUT)
```
ziafazal: instead of creating `DISCOVERY_GET_CATALOG_RESULT_CACHE_TIMEOUT ` in this we should use `ENTERPRISE_API_CACHE_TIMEOUT`.

enterprise/api_client/discovery.py
```diff
@@ -116,10 +118,20 @@ def traverse_pagination(response, endpoint, content_filter_query, query_params):
 
         return results
 
-    def get_catalog_results(self, content_filter_query, query_params=None, traverse_pagination=False):
+    def _get_catalog_results_from_discovery(self, content_filter_query, query_params=None, traverse_pagination=False):
```
ziafazal: Since we are using `_get_catalog_results_from_discovery` in tests can we remove underscore prefix from its name?

enterprise/api_client/discovery.py
HammadAhmadWaqas: I think ENTERPRISE_API_CACHE_TIMEOUT is used for API that belongs to Enterprise. That is why we are using it in EnterpriseApiClient.
But now we want some TimeOut period for Discovery to use in CourseCatalogApiClient.

ziafazal: That time out `ENTERPRISE_API_CACHE_TIMEOUT ` is used in cache results of discovery call. since that enterprise API is calling discovery.

HammadAhmadWaqas:
@schenedx  it appears `_load_data` calls `get_edx_api_data` function which has built-in cache support.

PR #549, opened 2019-08-16
Revert "ENT-2117 | Adding enterprise_learner_portal djangoapp to house endpoi…"

tuchfarber:
Reverts edx/edx-enterprise#544


tuchfarber:
Closing in favor of pinning a lower version so we can fix forward.

PR #550, opened 2019-08-19
create ecommerce order data upon successful manual learner course enrollment

muhammad-ammar:
**Description:** Create ecommerce order record for successful enrollment of a learner through the `manage_learners` admin page.

**JIRA:** https://openedx.atlassian.net/browse/ENT-2081

**Dependencies:** https://github.com/edx/ecommerce/pull/2520

**Sandbox:**
- [Manage Learners Admin Page](https://mammar.sandbox.edx.org/admin/enterprise/enterprisecustomer/cc8a0260-36d0-4520-8e1a-154d0431aa9c/change/)
- [Ecommerce Order Page](https://ecommerce-mammar.sandbox.edx.org/dashboard/orders/)

**Testing:**
1. Go to manage learners page
2. Enroll single or multiple learners. Enrollment should be successful.
3. Go to ecommerce orders page.
4. You should see an order with correct data for the learner enrolled.

**Single Enrollment Order Creation Failure Message**
Screenshot 2019-08-21 at 4 45 08 PM

**Bulk Enrollment Order Creation Failure Message**
Screenshot 2019-08-21 at 4 48 17 PM

### DIFF
--- a/enterprise/admin/views.py
+++ b/enterprise/admin/views.py
@@ -39,7 +39,7 @@ from enterprise.admin.utils import (
     split_usernames_and_emails,
     validate_email_to_link,
 )
-from enterprise.api_client.ecommerce import EcommerceApiClient
+from enterprise.api_client.ecommerce import get_ecommerce_api_client
 from enterprise.api_client.lms import EnrollmentApiClient
 from enterprise.constants import PAGE_SIZE
 from enterprise.models import (
@@ -49,13 +49,13 @@ from enterprise.models import (
     EnterpriseCustomerUser,
     PendingEnterpriseCustomerUser,
 )
-from enterprise.utils import get_configuration_value_for_site, send_email_notification_message, track_enrollment
-
-try:
-    from student.models import CourseEnrollment, CourseEnrollmentAttribute
-except ImportError:
-    CourseEnrollment = None
-    CourseEnrollmentAttribute = None
+from enterprise.utils import (
+    CourseEnrollmentOrderCreationError,
+    create_order_data_for_learner_enrollment,
+    get_configuration_value_for_site,
+    send_email_notification_message,
+    track_enrollment,
+)
 
 
 class TemplatePreviewView(View):
@@ -574,15 +574,16 @@ class EnterpriseCustomerManageLearnersView(View):
         failures = []
         order_failures = []
 
-        ecommerce_client = EcommerceApiClient()
+        ecommerce_client = get_ecommerce_api_client()
 
         for user in existing_users:
             succeeded = cls.enroll_user(enterprise_customer, user, course_mode, course_id)
 
             if succeeded:
-                # if learner enrollment is successfull then create an order record in ecommerce
-                order_created = cls.create_order_data_for_learner_enrollment(ecommerce_client, user, course_id)
-                if not order_created:
+                try:
+                    # if learner enrollment is successfull then create an order record in ecommerce
+                    create_order_data_for_learner_enrollment(ecommerce_client, user, course_id)
+                except CourseEnrollmentOrderCreationError:
                     order_failures.append(user)
 
                 successes.append(user)
@@ -598,42 +599,6 @@ class EnterpriseCustomerManageLearnersView(View):
             pending.append(pending_user)
 
         return successes, pending, failures, order_failures
-
-    @classmethod
-    def create_order_data_for_learner_enrollment(cls, ecommerce_client, user, course_id):
-        """
-        Create ecommerce order data for learner and also create `CourseEnrollmentAttribute` for the learner enrollment.
-        """
-        response = ecommerce_client.create_ecommerce_order_for_manual_course_enrollment(
-            user.id,
-            user.username,
-            user.email,
-            course_id,
-        )
-
-        if response and 'order_number' in response:
-            enrollment = CourseEnrollment.get_enrollment(user, course_id)
-            CourseEnrollmentAttribute.add_enrollment_attr(
-                enrollment=enrollment,
-                data_list=[{
-                    'namespace': 'order',
-                    'name': 'order_number',
-                    'value': response['order_number']
-                }]
-            )
-            return True
-
-        logging.warning(
-            '[Enterprise Manual Course Enrollment] Failed to create ecommerce order. LMSUserId: %s, User: %s, '
-            'Email: %s, Course: %s, Response: %s',
-            user.id,
-            user.username,
-            user.email,
-            course_id,
-            response
-        )
-
-        return False
 
     @classmethod
     def send_messages(cls, http_request, message_requests):
--- a/enterprise/admin/views.py
+++ b/enterprise/admin/views.py
@@ -51,7 +51,9 @@ from enterprise.models import (
 )
 from enterprise.utils import (
     CourseEnrollmentOrderCreationError,
+    CourseEnrollmentOrderUpdateError,
     create_order_data_for_learner_enrollment,
+    fail_order_data_for_learner_enrollment,
     get_configuration_value_for_site,
     send_email_notification_message,
     track_enrollment,
@@ -414,7 +416,7 @@ class EnterpriseCustomerManageLearnersView(View):
         return all_processable_emails
 
     @classmethod
-    def enroll_user(cls, enterprise_customer, user, course_mode, *course_ids):
+    def enroll_user(cls, enterprise_customer, user, course_mode, *course_ids, enrollment_attrs=None):
         """
         Enroll a single user in any number of courses using a particular course mode.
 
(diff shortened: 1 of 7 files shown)


enterprise/admin/views.py
```diff
@@ -565,9 +572,19 @@ def enroll_users_in_course(cls, enterprise_customer, course_id, course_mode, ema
...
+        order_failures = []
+
+        ecommerce_client = EcommerceApiClient()
 
         for user in existing_users:
             succeeded = cls.enroll_user(enterprise_customer, user, course_mode, course_id)
+
+            # if learner enrollment is successfull then create an order record in ecommerce
+            if succeeded:
+                order_created = cls.create_order_data_for_learner_enrollment(ecommerce_client, user, course_id)
```
mushtaqak: we can move this code in the if block below.

enterprise/admin/views.py
```diff
@@ -581,7 +598,43 @@ def enroll_users_in_course(cls, enterprise_customer, course_id, course_mode, ema
...
+            enrollment = CourseEnrollment.get_enrollment(user, course_id)
+            CourseEnrollmentAttribute.add_enrollment_attr(
+                enrollment=enrollment,
+                data_list=[{
+                    'namespace': 'order',
+                    'name': 'order_number',
+                    'value': response['order_number']
+                }]
+            )
+            return True
```
mushtaqak: Should we return `True` `False` or something better can be returned here?

enterprise/admin/views.py
```diff
@@ -723,6 +776,33 @@ def get_pending_enrollment_message(cls, pending_users, enrolled_in):
...
+    def get_failed_ecommerce_order_message(cls, users, course_id):
+        """
+        Create message for the users who were enrolled in a course but ecommerce order was not created successfully.
+
+        Args:
+            users: An iterable of users who were not successfully enrolled
+            course_id (str): A string identifier for the course
+
+        Returns:
+        tuple: A 2-tuple containing a message type and message text
```
mushtaqak: nit: indent one level

enterprise/admin/views.py
```diff
@@ -723,6 +776,33 @@ def get_pending_enrollment_message(cls, pending_users, enrolled_in):
...
+        """
+        failed_emails = [user.email for user in users]
+        count = len(failed_emails)
+
+        return (
+            'error',
+            ungettext(
+                'Learner {emails} is enrolled in {course_id} but ecommerce order creation failed',
+                'The following learners are enrolled in {course_id} but ecommerce order creation failed : {emails}',
+                count
```
mushtaqak: We can directly call `len(failed_emails)` ... we are not re-using it for now at other place.

tests/test_enterprise/api_client/test_ecommerce.py
```diff
@@ -66,3 +69,29 @@ def test_get_course_final_price(self, ecommerce_api_client_mock):
...
+        """
+        Test that `create_ecommerce_order_for_manual_course_enrollment` returns expected response in case of exception.
+        """
+        self._setup_ecommerce_api_client(
+            client_mock=ecommerce_api_client_mock,
+            method_name='manual_course_enrollment_order.post',
+            return_value={},
+            side_effect=Timeout()
+        )
+        assert EcommerceApiClient().create_ecommerce_order_for_manual_course_enrollment(1, '', '', '') is None
```
mushtaqak: We can check log message as well if it is easier ?

tests/test_admin/test_view.py
```diff
@@ -1479,6 +1608,44 @@ def test_post_link_and_enroll_into_program(
 
         self._assert_django_messages(response, expected_messages)
 
+    @mock.patch("enterprise.models.CourseCatalogApiClient")
+    @mock.patch("enterprise.api_client.ecommerce.ecommerce_api_client")
+    @mock.patch("enterprise.admin.views.EnrollmentApiClient")
+    @mock.patch("enterprise.admin.forms.EnrollmentApiClient")
+    def test_order_failure_message(
```
mushtaqak: May be rename this test to `test_order_failure_message_bulk` or something. Also if we can move it below the `test_order_failure_message`, would be great because both are covering same cases.

enterprise/api_client/ecommerce.py
```diff
@@ -27,7 +29,7 @@ class EcommerceApiClient(object):
     Object builds an API client to make calls to the E-Commerce API.
     """
 
-    def __init__(self, user):
+    def __init__(self, user=None):
```
brittneyexline: i'd rather have a static method that generates a client with the ecommerce service user than have that be the default behavior of initialization

enterprise/api_client/ecommerce.py
```diff
@@ -65,3 +70,31 @@ def get_course_final_price(self, mode, currency='$', enterprise_catalog_uuid=Non
         if price != mode['min_price']:
             return format_price(price, currency)
         return mode['original_price']
+
+    def create_ecommerce_order_for_manual_course_enrollment(  # pylint: disable=invalid-name
```
brittneyexline: nit: ecommerce can be removed from this function name; it's redundant since this is the ecommerce api client, and it will make the function name slightly more concise

enterprise/api_client/ecommerce.py
```diff
@@ -65,3 +70,31 @@ def get_course_final_price(self, mode, currency='$', enterprise_catalog_uuid=Non
...
+        except (SlumberBaseException, ConnectionError, Timeout) as exc:
+            LOGGER.exception(
+                '[Enterprise Manual Course Enrollment] Failed to create ecommerce order. LMSUserId: %s, User: %s, '
+                'Email: %s, Course: %s, Message: %s',
+                enrolled_learner_lms_user_id,
+                enrolled_learner_username,
+                enrolled_learner_email,
+                enrolled_course_run_key,
+                exc
+            )
```
brittneyexline: do you intend to swallow this exception or have it bubbled up to the caller? if it's the latter (which i would recommend), then you will have to add `raise exc` here

muhammad-ammar: If we do `raise exc` then we have to catch all the exceptions in the caller that we are already handling here. And in the caller we would catch the exceptions and return a `False` in the `except` block. I am not sure what benefit bubbling up the exception will give us in this case?

muhammad-ammar: On second thought, bubbling up exception will help in other cases, for example, when we do enrollment using enrollment api. So, I have created a custom exception and raised that exception.

brittneyexline: generally speaking, i think it is wrong to catch exceptions like this in the api client code because it typically masks problems when they occur. we should only "swallow" exceptions when the consequence can be easily handled for the business logic, ie trying to return a specific field and finding a `KeyError` or the like. only the caller can decide the consequence of getting an exception like this, and if it doesn't catch it then at least our logs and new relic will be loud about something being wrong.

muhammad-ammar: hmm...so in the client method(`create_order_for_manual_course_enrollment`), just re-raise exception and then let the caller of client method decide what to do with raised exceptions, right?

enterprise/utils.py
```diff
@@ -1010,3 +1022,26 @@ def discovery_query_url(content_filter):
...
+    Create ecommerce order data for learner and also create `CourseEnrollmentAttribute` for the learner enrollment.
+    """
+    response = ecommerce_client.create_order_for_manual_course_enrollment(
+        user.id,
+        user.username,
+        user.email,
+        course_id,
+    )
+
+    enrollment = CourseEnrollment.get_enrollment(user, course_id)
```
muhammad-ammar: @brittneyexline This line of code is giving hard time on devstack. After the successful enrollment, `CourseEnrollment.get_enrollment` always return `None`. I have tried all the possible ways to get the enrollment object but no luck. I If I put a breakpoint in the above code and check the mysql table(`student_courseenrollment`) then I can see the newly create enrollment there. I have also tried `CourseEnrollment.objects.get/filter` but no luck. I have also debugged the lms enrollment endpoint and it is working as expected and creates the enrollment but the above line fails to get that enrollment. Also if I use `EnrollmentApiClient` to get the latest enrollment than I get the enrollment but I don't want to call the api.

The weird thing is that, the above code works perfectly fine on sandbox. I have double checked the code on sandbox and it works as expected.

Any idea what is happening here? Any tip to fix the issue?

muhammad-ammar: Though I didn't figured out the issue but I think instead of directly importing the `CourseEnrollment` and `CourseEnrollmentAttribute`, It would be better to update the [enrollment api post endpoint](https://github.com/edx/edx-platform/blob/233bf245ec29f8c02cb9cf2998b84653a047a654/openedx/core/djangoapps/enrollments/views.py#L664) to handle `enrollment_attrs` updates. 
Thoughts? Any concern on this?

PR #563, opened 2019-09-03
ENT-1952 Updated logs for splunk monitoring

irfanuddinahmad:
**Description:** This PR improves the Enterprise Enrollment Workflow logs for efficient monitoring in Splunk

**JIRA:** https://openedx.atlassian.net/browse/ENT-1952

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -848,10 +848,11 @@ class HandleConsentEnrollment(View):
             context_data = get_global_context(request, enterprise_customer)
             error_code = 'ENTHCE000'
             log_message = (
-                'No course_modes for course_id {course_id} for enterprise_catalog_uuid '
-                '{enterprise_catalog_uuid}.'
-                'The following error was presented to '
-                'user {userid}: {error_code}'.format(
+                '[Enterprise Enrollment] Course_modes for course not found. '
+                'Course: {course_id}, '
+                'EnterpriseCatalog: {enterprise_catalog_uuid}, '
+                'ErrorCode: {error_code}, '
+                'User: {userid}'.format(
                     userid=request.user.id,
                     enterprise_catalog_uuid=enterprise_catalog_uuid,
                     course_id=course_id,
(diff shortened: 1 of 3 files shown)


enterprise/views.py
```diff
@@ -947,9 +948,8 @@ def get_available_course_modes(self, request, course_run_id, enterprise_catalog)
         """
         modes = EnrollmentApiClient().get_course_modes(course_run_id)
         if not modes:
-            LOGGER.warning('Unable to get course modes for course run id {course_run_id}.'.format(
-                course_run_id=course_run_id
-            ))
+            LOGGER.warning('[Enterprise Enrollment] Unable to get course modes. '
+                           'CourseRun: {course_run_id}'.format(course_run_id=course_run_id))
```
mushtaqak: Should we use `Course` here instead of `CourseRun` since course run id is a course id actually. This might help filter logs a bit easier ?

enterprise/views.py
```diff
@@ -958,8 +958,9 @@ def get_available_course_modes(self, request, course_run_id, enterprise_catalog)
             modes.sort(key=lambda course_mode: enterprise_catalog.enabled_course_modes.index(course_mode['slug']))
             if not modes:
                 LOGGER.info(
-                    'No matching course modes found for course run {course_run_id} in '
-                    'EnterpriseCustomerCatalog [{enterprise_catalog_uuid}]'.format(
+                    '[Enterprise Enrollment] Matching course modes were not found in EnterpriseCustomerCatalog. '
+                    'CourseRun: {course_run_id}, '
```
mushtaqak: Same here about course run id

enterprise/views.py
irfanuddinahmad: Course and CourseRun are logged separately elsewhere in the file also. Hence following the same convention.

enterprise/views.py
irfanuddinahmad: Course and CourseRun are logged separately elsewhere in the file also. Hence following the same convention.

PR #565, opened 2019-09-05
ENT-2230 Add patch method on learner portal enrollment endpoint

brittneyexline:

**JIRA:** Link to JIRA ticket

instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen

**Merge checklist:**

- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added

**Post merge:**
- [ ] Create a tag

migrations, etc.

### DIFF
--- a/enterprise_learner_portal/api/v1/views.py
+++ b/enterprise_learner_portal/api/v1/views.py
@@ -55,16 +55,59 @@ class EnterpriseCourseEnrollmentView(APIView):
             user_id=user.id,
             enterprise_customer__uuid=enterprise_customer_id,
         )
-        course_ids_for_ent_enrollments = EnterpriseCourseEnrollment.objects.filter(
+        enterprise_enrollments = EnterpriseCourseEnrollment.objects.filter(
             enterprise_customer_user=enterprise_customer_user
-        ).values_list('course_id', flat=True)
+        )
 
-        overviews = get_course_overviews(course_ids_for_ent_enrollments)
+        course_overviews = get_course_overviews(enterprise_enrollments.values_list('course_id', flat=True))
 
         data = EnterpriseCourseEnrollmentSerializer(
-            overviews,
+            enterprise_enrollments,
             many=True,
-            context={'request': request},
+            context={'request': request, 'course_overviews': course_overviews},
+        ).data
+
+        return Response(data)
+
+    def patch(self, request):
+        """
+        Patch method for the view.
+        """
+        if get_course_overviews is None:
+            raise NotConnectedToOpenEdX(
+                _('To use this endpoint, this package must be '
+                  'installed in an Open edX environment.')
+            )
+
+        user = request.user
+        enterprise_customer_id = request.query_params.get('enterprise_id', None)
+        course_id = request.query_params.get('course_id', None)
+        marked_done = request.query_params.get('marked_done', None)
+        if not enterprise_customer_id or not course_id or marked_done is None:
+            return Response(
+                {'error': 'enterprise_id, course_id, and marked_done must be provided as query parameters'},
+                status=HTTP_400_BAD_REQUEST
+            )
+
+        enterprise_customer_user = get_object_or_404(
+            EnterpriseCustomerUser,
+            user_id=user.id,
+            enterprise_customer__uuid=enterprise_customer_id,
+        )
+
+        enterprise_enrollment = get_object_or_404(
+            EnterpriseCourseEnrollment,
+            enterprise_customer_user=enterprise_customer_user,
+            course_id=course_id
+        )
+
+        enterprise_enrollment.marked_done = marked_done
+        enterprise_enrollment.save()
+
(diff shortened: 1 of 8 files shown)


enterprise_learner_portal/api/v1/views.py
```diff
@@ -55,16 +42,50 @@ def get(self, request):
...
+        enterprise_enrollments = EnterpriseCourseEnrollment.objects.filter(
             enterprise_customer_user=enterprise_customer_user
-        ).values_list('course_id', flat=True)
-
-        overviews = get_course_overviews(course_ids_for_ent_enrollments)
+        )
 
         data = EnterpriseCourseEnrollmentSerializer(
-            overviews,
+            enterprise_enrollments,
```
douglashall: Isn't this a deoptimization? Now we will make a query to the course overview table for every enterprise enrollment instead of batching them like the previous code was doing.

enterprise_learner_portal/api/v1/views.py
```diff
@@ -55,16 +42,50 @@ def get(self, request):
             user_id=user.id,
             enterprise_customer__uuid=enterprise_customer_id,
         )
-        course_ids_for_ent_enrollments = EnterpriseCourseEnrollment.objects.filter(
+        enterprise_enrollments = EnterpriseCourseEnrollment.objects.filter(
             enterprise_customer_user=enterprise_customer_user
-        ).values_list('course_id', flat=True)
-
-        overviews = get_course_overviews(course_ids_for_ent_enrollments)
```
douglashall: Maybe just move the `get_course_overviews` call to the serializer, but keep the ability to pass multiple enterprise enrollments to the serializer.

enterprise_learner_portal/api/v1/views.py
```diff
@@ -55,16 +42,50 @@ def get(self, request):
...
+
+    def patch(self, request):
+        """
+        Patch method for the view.
+        """
+        user = request.user
+        enterprise_customer_id = request.query_params.get('enterprise_id', None)
+        course_id = request.query_params.get('course_id', None)
+        marked_done = request.query_params.get('marked_done', None)
+        if not (enterprise_customer_id and course_id) or marked_done is None:
```
christopappas: Personal preference, but I think I prefer not mixing ANDs and ORs if it's possible. Something like:

```
if not enterprise_customer_id or not course_id or marked_done is None:
```

enterprise_learner_portal/api/v1/views.py
christopappas: By default the serializer will only ever work on 1 object at a time though, so even if we hand multiple course enrollments to the serializer, still would be doing a `get_course_overviews` call for each item in that list/queryset we hand it

A suggestion: in the view, make 1 call to get all the course overviews that we will ever need to look at, and then pass that to the serializer's context. So instead of calling `get_course_overviews` in the serializer,  you could just lookup the course_overview info in the dict that exists in the context

enterprise_learner_portal/api/v1/views.py
brittneyexline: the motivation for this was so that it was easier to serialize one enrollment's data below in what the patch function returned, and to also have access to the enterprise enrollment object itself in the serializer for the status determination. i considered passing the enterprise_enrollments as context, but it seemed awkward to do that considering this serializer is supposed to serialize "enrollments". i'm not sure what the performance implications are though... i would guess the table is indexed by course id?

enterprise_learner_portal/api/v1/views.py
```diff
@@ -55,16 +42,50 @@ def get(self, request):
...
+
+        enterprise_enrollment.marked_done = marked_done
+        enterprise_enrollment.save()
+
+        data = EnterpriseCourseEnrollmentSerializer(
+            enterprise_enrollment,
+            context={'request': request},
+        ).data
+
+        return Response(data)
```
adamstankiewicz: Having the response as the newly updated course enrollment serialized by `EnterpriseCourseEnrollmentSerializer` matches what the frontend will be looking for :)

tests/test_enterprise_learner_portal/api/test_views.py
```diff
@@ -134,3 +133,121 @@ def test_view_requires_openedx_installation(self):
...
+        View should update the enrollment's marked_done field and return serialized data from
+         EnterpriseCourseEnrollmentSerializer for the enrollment (which we mock in this case)
+        """
+        mock_get_overviews.return_value = {'overview_info': 'this would be a larger dict'}
+        mock_serializer.return_value = self.MockSerializer()
+        query_params = {
+            'enterprise_id': str(self.enterprise_customer.uuid),
+            'course_id': self.course_run_id,
+            'marked_done': True,
+        }
```
christopappas: Just a suggestion, but as a sanity check, maybe assert that `marked_done` is False before you hit the patch endpoint

PR #572, opened 2019-09-17
ENT-2017 Document for multiple enterprises decision

irfanuddinahmad:
JIRA: https://openedx.atlassian.net/browse/ENT-2016
          https://openedx.atlassian.net/browse/ENT-2017

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/docs/decisions/0001-multiple-enterprise-support.rst
+++ b/docs/decisions/0001-multiple-enterprise-support.rst
@@ -1,0 +1,51 @@
+Multiple Enterprise Support
+---------------------------
+
+Status
+======
+
+Accepted
+
+Context
+=======
+
+There is a business need to be able to support a learner being linked to multiple enterprises. Today, edX workflows only support a 1-to-1 linkage and our system does not prevent linking multiple Enterprises to a single learner. Hence not all workflows work as expected.
+
+Some examples of these errant workflows are:
+
+#. When a learner is linked with an enterprise and tries to link with another enterprise(through SSO), the learner is redirected to the login page instead of the registration page for the second enterprise. This is because the system detects that these credentials are already in the system.
+
+#. Enrollment, Progression, Completion, Consent records become ambiguous if a learner completes a course from enterprise A and then attempts the same course through enterprise B.
+
+#. Incorrect discounts are applied when an employee linked with multiple enterprises redeems the benefit applicable only to one of the enterprises.
+
+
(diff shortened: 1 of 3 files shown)


docs/decisions/0001-multiple-enterprise-support.rst
```diff
@@ -0,0 +1,47 @@
...
+#. When a learner is linked with an enterprise and tries to link with another enterprise(through SSO), the learner is redirected to the login page instead of the registration page for the second enterprise. This is because the system detects that these credentials are already in the system.
+#. Enrollment, Progression, Completion, Consent records become ambiguous if a learner completes a course from enterprise A and then attempts the same course through enterprise B.
+#. Incorrect discounts are applied when an employee linked with multiple enterprises redeems the benefit applicable only to one of the enterprises.
+
+
+Decisions
+=========
+
+Short term:
+#. As part of the login flow the user would select one of the linked enterprises.
```
muhammad-ammar: Can we some information regarding `What steps a logged in learner has to do to switch to another enterprise?`

irfanuddinahmad: Some information is added in the "Long term" section. More detail would be added as part of the detailed solution.

irfanuddinahmad:
> It would be great if we can add information regarding
> 
> 1. What will happen in the case of API or server-to-server calls enterprise, How we will decide what is the current enterprise context?
> 2. What to do in a case when we need context independent response? For example, Return all the enterprises of a learner.
> 3. How the current enterprise context is stored? Cache/Session/DB?

The current enterprise context could be part of Session, JWT etc. The answers to the above queries would be dependent on the detailed solution.

brittneyexline:
is there an ETA on when this is getting merged? i ask because there is another ADR that is in progress. thanks!

irfanuddinahmad:
> is there an ETA on when this is getting merged? i ask because there is another ADR that is in progress. thanks!

@brittneyexline I think this can be merged if I receive another approval. Can I request you for a review?

irfanuddinahmad:
> i don't have any particular feedback about the content of this, however when i viewed this in rich text, the formatting appeared to be messed up. i would recommend double checking that and fixing appropriately before merging.

@brittneyexline The formatting seems to be fine in gedit. Is rich text a requirement for ADRs?

PR #573, opened 2019-09-17
Added passwords for files to the API for reporting config

staubina:
[ENT-2321]

**Merge checklist:**
- [x] Translations updated

**Post merge:**
- [x] Tag pushed and a new [version](https://github.com/edx/edx-enterprise/releases) released
- [x] Version pushed to [PyPI](https://pypi.org/project/edx-enterprise/)
- [x] PR created in [edx-platform](https://github.com/edx/edx-platform) to upgrade dependencies (including edx-enterprise)

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -1814,6 +1814,9 @@ class EnterpriseCustomerReportingConfiguration(TimeStampedModel):
     def encrypted_password(self):
         """
         Return encrypted password as a string.
+
+        The data is encrypted in the DB at rest, but is unencrypted in the app when retrieved through the
+        decrypted_password field. This method will encrypt the password again before sending.
         """
         if self.decrypted_password:
             return force_text(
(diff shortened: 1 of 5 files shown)


enterprise/models.py
```diff
@@ -1823,6 +1823,13 @@ def encrypted_password(self):
             )
         return self.decrypted_password
 
+    @encrypted_password.setter
+    def encrypted_password(self, value):
```
tuchfarber: As we discussed offline, the confusion around the encrypted vs decrypted is a bit confusing. Please put some comments around either the fields or the properties related to those fields describing how the data is only encrypted at rest (i.e. in the DB) and is *always* unencrypted in code (unless we manually encrypt it).

tests/test_enterprise/api/test_views.py
```diff
@@ -2416,6 +2416,8 @@ def test_reporting_config_post_permissions(self, has_feature_role, expected_stat
         if has_feature_role:
             response_content = self.load_json(response.content)
             assert response_content['enterprise_customer']['uuid'] == str(enterprise_customer.uuid)
+            assert response_content['encrypted_password'] is not None
```
tuchfarber: I'm understanding this as `encrypted_password` sent to the API is unencrypted "password" and the `encrypted_password` in the response should be encrypted `q1w2e3r4`. If true, I think it's at least worth verifying that the response value is not equal to the submitted value.

You could go farther by encrypted the submitted value manually and comparing, but that seems like testing the framework, which isn't necessary.

tuchfarber: I also just realized that this is really a permissions test. I'd understand if this isn't the place to put that test.

PR #577, opened 2019-09-19
Ttracy/ent 2305 reporting feature toggle

Tj-Tracy:
Adds a toggle to the EnterpriseCustomer that allows turning on/off the new reporting config page for a slow rollout.

**Merge checklist:**
- [ ] Translations updated

**Post merge:**


enterprise/migrations/0076_auto_20190918_2037.py
```diff
@@ -0,0 +1,35 @@
...
+            field=models.BooleanField(default=False, help_text='Specifies whether to allow access to the reporting configurations screen in the admin portal.'),
+        ),
+        migrations.AddField(
+            model_name='historicalenterprisecustomer',
+            name='enable_portal_reporting_config_screen',
+            field=models.BooleanField(default=False, help_text='Specifies whether to allow access to the reporting configurations screen in the admin portal.'),
+        ),
+        migrations.AlterField(
+            model_name='enterprisecustomer',
+            name='learner_portal_hostname',
```
tuchfarber: Why are we altering this field?

Tj-Tracy: not sure (this was auto generated) could this have happened because I didn't put the new field at the end of the model?

tuchfarber: Looks like Doug might have ran the migration, then changed the help text in this commit (https://github.com/edx/edx-enterprise/commit/0166b59b121bf1ae5e958c583ba5f3b93a2fbf84), so this is fine to ignore.

PR #582, opened 2019-09-25
ENT-2122 | Added feature to transmit specific catalogs of Enterprise …

hasnain-naveed:
…Customer.

**Merge checklist:**
- [ ] Translations updated

**Post merge:**


enterprise/admin/forms.py
```diff
@@ -333,7 +338,44 @@ class Meta:
...
+            # "catalog_uuids",
+        )
+
+    def __init__(self, *args, **kwargs):
+        super(EnterpriseCustomerAdminForm, self).__init__(*args, **kwargs)
+        enterprise_customer = kwargs.get('instance', None)
+        self.fields["enterprise_customer_catalogs"].queryset = EnterpriseCustomerCatalog.objects.filter(
+            enterprise_customer=enterprise_customer
         )
+        self.fields["enterprise_customer_catalogs"].initial = enterprise_customer.customer_catalogs
```
ziafazal: We should not set and initial value for this field it should be empty which means content of all catalogs needs to be transmitted.

enterprise/admin/forms.py
```diff
@@ -333,7 +338,44 @@ class Meta:
             "enable_learner_portal",
             "learner_portal_hostname",
             "enable_portal_reporting_config_screen",
+            # "catalog_uuids",
+        )
+
+    def __init__(self, *args, **kwargs):
+        super(EnterpriseCustomerAdminForm, self).__init__(*args, **kwargs)
+        enterprise_customer = kwargs.get('instance', None)
+        self.fields["enterprise_customer_catalogs"].queryset = EnterpriseCustomerCatalog.objects.filter(
```
ziafazal: nit: we can use get catalogs by related name here
```
self.fields["enterprise_customer_catalogs"].queryset = enterprise_customer.enterprise_customer_catalogs.all()
```

enterprise/admin/forms.py
```diff
@@ -333,7 +338,44 @@ class Meta:
...
+        self.fields["enterprise_customer_catalogs"].initial = enterprise_customer.customer_catalogs
+
+    def save(self, commit=True):    # pylint: disable=unused-argument
+        instance = super(EnterpriseCustomerAdminForm, self).save(commit=False)
+        enterprise_customer_catalogs = self.cleaned_data.get('enterprise_customer_catalogs')
+        instance.catalog_uuids = [str(catalog.uuid) for catalog in enterprise_customer_catalogs]
+        instance.save()
+        return instance
+
+    # def clean_catalog_uuids(self):
```
ziafazal: why we have to comment this out?

enterprise/models.py
```diff
@@ -261,6 +262,13 @@ class Meta:
         default='',
         help_text=_("Hostname of the enterprise learner portal, e.g. bestrun.edx.org.")
     )
+    catalog_uuids = models.TextField(
+        blank=True,
+        null=True,
+        default=[],
```
ziafazal: default set to `[]` for Textfield. we probably don't need a default since it is nullable field.

enterprise/models.py
```diff
@@ -261,6 +262,13 @@ class Meta:
         default='',
         help_text=_("Hostname of the enterprise learner portal, e.g. bestrun.edx.org.")
     )
+    catalog_uuids = models.TextField(
```
ziafazal: This field should in `EnterpriseCustomerPluginConfiguration` model and its name should be `catalogs_to_transmit`

enterprise/api_client/enterprise.py
```diff
@@ -30,19 +30,20 @@ class EnterpriseApiClient(JwtLmsApiClient):
...
             enterprise_customer (EnterpriseCustomer): The EnterpriseCustomer to return content metadata for.
+            catalogs_to_transmit (EnterpriseCustomerCatalog): Optional list of EnterpriseCustomerCatalog objects
 
         Returns:
             list: List of dicts containing content metadata.
         """
         content_metadata = OrderedDict()
-
-        for enterprise_customer_catalog in enterprise_customer.enterprise_customer_catalogs.all():
+        enterprise_customer_catalogs = catalogs_to_transmit or enterprise_customer.enterprise_customer_catalogs.all()
```
ziafazal: we should pull catalogs which needs to be transmitted from enterprise->enterprise channel config->catalogs relationship instead of passing it in `catalogs_to_transmit` param.

integrated_channels/integrated_channel/models.py
```diff
@@ -73,6 +80,28 @@ def channel_worker_user(self):
...
+        """
+        Return the list of EnterpriseCustomerCatalog objects.
+        """
+        enterprise_customer_catalogs = []
+        catalogs_uuids = ast.literal_eval(self.catalogs_to_transmit or str(list()))
+        for uuid in catalogs_uuids:
+            try:
+                enterprise_customer_catalogs.append(EnterpriseCustomerCatalog.objects.get(uuid=uuid))
+            except ValidationError as error:
+                LOGGER.error(
```
ziafazal: It should be `LOGGER.warning` instead of `LOGGER.error`.

integrated_channels/integrated_channel/models.py
```diff
@@ -73,6 +80,28 @@ def channel_worker_user(self):
...
+                    "Customer '{enterprise_customer}' in '{class_name}' class contains invalid catalog uuid '{uuid}'. "
+                    "It can not be transmitted and it fails with {message}".format(
+                        enterprise_customer=self.enterprise_customer.name,
+                        class_name=type(self).__name__,
```
ziafazal: I don't adding `class_name` get us any benefit us. since logger has these.

integrated_channels/integrated_channel/models.py
```diff
@@ -73,6 +96,18 @@ def channel_worker_user(self):
...
 
+    @property
+    def customer_catalogs_to_transmit(self):
+        """
+        Return the list of EnterpriseCustomerCatalog objects.
+        """
+        catalogs_list = []
+        if self.catalogs_to_transmit:
+            catalogs_list = EnterpriseCustomerCatalog.objects.filter(
+                uuid__in=list(set([uuid.strip() for uuid in self.catalogs_to_transmit.split(",") if uuid.strip()]))
```
ziafazal: We should avoid this loop and set and then list conversion at multiple places instead make a method and put these operations inside it.

integrated_channels/integrated_channel/models.py
```diff
@@ -61,10 +62,32 @@ class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
...
+        null=True,
+        help_text=_("A comma-separated list of catalog UUIDs to transmit."),
+    )
 
     class Meta:
         abstract = True
 
+    def clean(self):
+        invalid_uuids = []
+        uuids = list(set([uuid.strip() for uuid in self.catalogs_to_transmit.split(",") if uuid.strip()]))
```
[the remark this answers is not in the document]
hasnain-naveed: to remove the leading and trailing spaces.

enterprise/api_client/enterprise.py
```diff
@@ -30,19 +30,20 @@ class EnterpriseApiClient(JwtLmsApiClient):
 
     DEFAULT_VALUE_SAFEGUARD = object()
 
-    def get_content_metadata(self, enterprise_customer):
+    def get_content_metadata(self, enterprise_customer, catalogs_to_transmit=None):
```
brittneyexline: nit: if we're adding this argument here, it would be better to name it simply `enterprise_catalogs` because this code is not specific to transmitting catalog data to a third party.

PR #586, opened 2019-09-27
ENT-2354: Added ability to enable/disable real time learner data transmit

ziafazal:
Added ability to enable/disable real time learner data transmit per enterprise.
ENT-2354

**Merge checklist:**
- [ ] Translations updated

**Post merge:**


georgebabey:
@ziafazal Thank you for creating this PR but as we discussed I don't think this is needed. We should just disable the integrations that are not pointing to valid tenants

georgebabey:
Closing this as its no longer needed.

PR #589, opened 2019-09-27
ENT-2310 | Adding a post_save receiver to listen to CourseEnrollments…

christopappas:
… table

~Before I merged, I definitely want to make sure I can verify that the async element of this will work on Production... will want to mess around a bit more with Devstack to that end.~

This works in devstack with `CELERY_ALWAYS_EAGER=True`, meaning that tasks run synchronously. Production and staging environment configs are different enough from devstack that I'm not 100% convinced testing asynchronous functionality means this code will or will not work in production, though I will still test setting up a broker and worker locally to prove the concept.

Barring discovering anything that drastically changes my approach, this should be ready for review!

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/tasks.py
+++ b/enterprise/tasks.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+"""
+Django tasks.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from logging import getLogger
+
+from celery import shared_task
+
+from enterprise.models import EnterpriseCourseEnrollment
+
+LOGGER = getLogger(__name__)
+
+
+@shared_task
+def create_enterprise_enrollment(course_id, enterprise_customer_user):
+    """
+    Create enterprise enrollment for user if course_id part of catalog for the ENT customer.
+    """
+    # Prevent duplicate records from being created if possible
+    # before we need to make a call to discovery
+    if EnterpriseCourseEnrollment.objects.filter(
+            enterprise_customer_user=enterprise_customer_user,
+            course_id=course_id,
+    ).exists():
+        LOGGER.info((
+            "EnterpriseCourseEnrollment record exists for user %s "
+            "on course %s. Exiting task."
+        ), enterprise_customer_user.user_id, course_id)
+        return
+
+    enterprise_customer = enterprise_customer_user.enterprise_customer
+    if enterprise_customer.catalog_contains_course(course_id):
+        LOGGER.info((
+            "Creating EnterpriseCourseEnrollment for user %s "
+            "on course %s for enterprise_customer %s"
+        ), enterprise_customer_user.user_id, course_id, enterprise_customer)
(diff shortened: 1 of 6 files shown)


enterprise/tasks.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    Create enterprise enrollment for user if course_id part of catalog for the ENT customer.
+    """
+    enterprise_customer = enterprise_customer_user.enterprise_customer
+    if enterprise_customer.catalog_contains_course(course_id):
+        LOGGER.info((
+            "Creating EnterpriseCourseEnrollment for user %s "
+            "on course %s for enterprise_customer %s"
+        ), enterprise_customer_user.user_id, course_id, enterprise_customer)
+        EnterpriseCourseEnrollment.objects.create(
```
christopappas: maybe use a `get_or_create` here just to be safe? (so we don't create duplicates)

christopappas: clicking enroll in devstack did NOT create 2 enterprise course enrollments, so i do not think this is necessary

adamstankiewicz: Hmm, good question. Even though in testing with devstack, it did not create 2 enterprise course enrollments... I feel that it might be good to be safe here just in case with `get_or_create`. If there are any edge cases where an enrollment might already exist in stage/production data, might be good to proactively handle that.

tests/test_enterprise/test_tasks.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    Tests tasks associated with Enterprise.
+    """
+    def setUp(self):
+        """
+        Setup for `TestEnterpriseTasks` test.
+        """
+        self.user = UserFactory(id=2, email='user@example.com')
+        self.enterprise_customer = EnterpriseCustomerFactory(
+            name='Team Titans',
```
adamstankiewicz: nice enterprise customer name ;)

christopappas:
The task will not be kicked off if `created` is not True in the signal though. So it's not just a save that has to happen for this task to be run, but a create on the CourseEnrollment table

Would there ever be a case where there would be more than 1 course enrollment created that maps to a single enterpriseCourseEnrollment? If not, I think this would be fine, right?

christopappas:
@brittneyexline 🔼

brittneyexline:
i'm saying that if the EnterpriseCourseEnrollment record had already been created at the point that the task has been kicked off, something that is definitely possible for the current enterprise enrollment workflow (see https://github.com/edx/edx-enterprise/blob/master/enterprise/models.py#L760), then we could bypass doing the work of in the task in this case too.

christopappas:
Ah, I see. Update coming

christopappas:
@brittneyexline addressed your recommendation. updated to make sure we check if the EnterpriseCourseEnrollment exists before making the call to discovery (and added a test)

PR #590, opened 2019-10-01
ENT-1289 | Removing enterprise entitlement logic from codebase

christopappas:
This needs to be merged and deployed AFTER the corresponding ecommerce  PR here: https://github.com/edx/ecommerce/pull/2612

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[2.0.0] - 2019-10-02
+---------------------
+
+* Removing EnterpriseCustomerEntitlement code
+
 [1.11.0] - 2019-10-02
 ---------------------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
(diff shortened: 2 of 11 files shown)


enterprise/api_client/ecommerce.py
```diff
@@ -43,25 +40,3 @@ def __init__(self, user):
 
         self.user = user
         self.client = ecommerce_api_client(user)
-
-    def get_course_final_price(self, mode, currency='$', enterprise_catalog_uuid=None):
```
brittneyexline: although the comment implies this is specific to entitlements, i think it's lying. it seems like this is generally trying to get discounted prices from ecommerce, and this is still relevant to the enterprise enrollment workflow. i would keep this in.

enterprise/views.py
```diff
@@ -922,22 +921,6 @@ class CourseEnrollmentView(NonAtomicView):
         'self_paced': _('Self-Paced')
     }
 
-    def set_final_prices(self, modes, request):
```
brittneyexline: again, i would keep all of this in.

christopappas:
@brittneyexline tagged you for rereview -- added back in the things you called out in your comments

PR #594, opened 2019-10-07
Added a notes field so that the requestor can write notes for the codes.

zamanafzal:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**


enterprise/api/v1/views.py
```diff
@@ -516,28 +516,30 @@ class CouponCodesView(APIView):
...
     REQUIRED_PARAM_ENTERPRISE_NAME = 'enterprise_name'
     OPTIONAL_PARAM_NUMBER_OF_CODES = 'number_of_codes'
+    OPTIONAL_PARAM_NOTES = 'notes'
 
     MISSING_REQUIRED_PARAMS_MSG = "Some required parameter(s) missing: {}"
 
     def get_required_query_params(self, request):
         """
-        Gets ``email``, ``enterprise_name``, and ``number_of_codes``,
+        Gets ``email``, ``enterprise_name``, ``number_of_codes``, and ``notes``,
```
irfanuddinahmad: message_notes may be more descriptive ...

enterprise/api/v1/views.py
```diff
@@ -565,36 +568,54 @@ def post(self, request):
...
-        msg_with_codes = _('{token_email} from {token_enterprise_name} has requested {token_number_codes} additional '
-                           'codes. Please reach out to them.').format(
-                               token_email=email,
-                               token_enterprise_name=enterprise_name,
-                               token_number_codes=number_of_codes)
-        msg_without_codes = _('{token_email} from {token_enterprise_name} has requested additional codes.'
-                              ' Please reach out to them.').format(
-                                  token_email=email,
-                                  token_enterprise_name=enterprise_name)
+        if number_of_codes and notes:
```
irfanuddinahmad: Can we extract out this code to a create_message_body function?

zamanafzal: I have added a new function named create_message_body. please take a look.

PR #598, opened 2019-10-10
ENT-1827 | Updated the 'contains_courses' function.

hasnain-naveed:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**


hasnain-naveed:
"contains_courses" function works fine if you provide the valid key. Then it will return True if that course/course_run exists in that catalog otherwise False. But there is an edge case where if you provide an invalid key that doesn't exist in our system then the "contains_courses" function will return True. So, I have catered this edge case.

PR #600, opened 2019-10-11
ENT-2372: Re-enabled the API calls to update completed_date and is_passing

irfanuddinahmad:
**Description**: Enterprise customers are seeing a course 'Pass' status for learners with 'Audit' enrollments. The transmit/export process, which does not differentiate between normal and audit learners, was recently changed to handle single grade transmissions instead of batch transmissions. This PR removes [this](https://github.com/edx/edx-enterprise/pull/574/files#diff-3051e117b960d55eba74f530e5050c0bR150) check in the recent code update which will enable correct population of completed_date and is_passing variables in the export and transmission log.

**JIRA**: https://openedx.atlassian.net/browse/ENT-2372

**Reference**: The [transmit/export process](https://github.com/edx/edx-enterprise/pull/574/files#diff-c36e78c76477eda2c1045fae45db0602R104) is assigning a grade based on the COURSE_GRADE_NOW_PASSED [signal](https://github.com/edx/edx-platform/blob/8c9b13e6fc8d5b1faa3767d625f20044910265a8/openedx/features/enterprise_support/signals.py#L72)

**Performance**: This PR enables the API calls to the grades api to fetch the current 'passed' status from the database instead of adding 'Pass' status in the transmitter/exporter. The extra API calls will ensure that data is fetched from the Single Source of Truth for grades!

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,11 @@ Change Log
 
 Unreleased
 ----------
+[2.0.7] - 2019-10-21
+---------------------
+
+* Added certificate and grades api calls for transmitting learner export to integrated channels
+
 [2.0.6] - 2019-10-18
 ---------------------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "2.0.6"
+__version__ = "2.0.7"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/integrated_channels/integrated_channel/exporters/learner_data.py
+++ b/integrated_channels/integrated_channel/exporters/learner_data.py
@@ -147,14 +147,13 @@ class LearnerExporter(Exporter):
             if not consent.granted or enterprise_enrollment.audit_reporting_disabled:
                 continue
 
-            if not is_passing and not completed_date:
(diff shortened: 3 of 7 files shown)


irfanuddinahmad:
@ziafazal Please see my response inline below:
> @irfanuddinahmad it appear these changes neither omit pass status of audit enrollments being sent to integrated channels nor stop audit enrollments transmission. We should investigate 2 issues
_The JIRA[ issue](https://openedx.atlassian.net/browse/ENT-2372) refers to incorrect Pass/Fail status of audit enrollments being sent via the integrated channels. Stopping audit type enrollment data is not a requirement of the client_
> 
> 1. how a user's grade is set to `Pass` while transmitting learner's data to SAP SF whereas `grades_persistentcoursegrade` table does not have any record
_I have commented on the[ issue](https://openedx.atlassian.net/browse/ENT-2372) with data from the transmission table. The data shows that this inconsistent state between the transmission record and the grades record started to appear only after the date when recent changes were made to the exporter code_
> 2. Do we checks in place to omit grade of audit enrollments while transmitting them if yes why it is not working and if there are no checks we should place them.
_As mentioned above, stopping audit type transmission is not a requirement of the client_

ziafazal:
@irfanuddinahmad  @brittneyexline If course passed signal has issues then we might need to have an additional check. However, I strongly agree we should add more logging around results of both grades and certificate APIs to probably get more concrete answers to about underlying issue.

PR #603, opened 2019-10-24
ENT-2428: Updated image url field for cornerstone and degreed

irfanuddinahmad:
JIRA: https://openedx.atlassian.net/browse/ENT-2428

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/integrated_channels/cornerstone/exporters/content_metadata.py
+++ b/integrated_channels/cornerstone/exporters/content_metadata.py
@@ -15,6 +15,7 @@ from django.apps import apps
 
 from enterprise.utils import get_closest_course_run, get_language_code
 from integrated_channels.integrated_channel.exporters.content_metadata import ContentMetadataExporter
+from integrated_channels.utils import get_image_url
 
 LOGGER = getLogger(__name__)
 
@@ -33,7 +34,7 @@ class CornerstoneContentMetadataExporter(ContentMetadataExporter):  # pylint: di
         'ID': 'key',
         'Title': 'title',
         'Description': 'description',
-        'Thumbnail': 'card_image_url',
+        'Thumbnail': 'image',
         'URL': 'enrollment_url',
         'IsActive': 'is_active',
         'LastModifiedUTC': 'modified',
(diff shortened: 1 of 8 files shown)


integrated_channels/cornerstone/exporters/content_metadata.py
```diff
@@ -98,6 +98,18 @@ def transform_estimated_hours(self, content_metadata_item):
 
         return duration
 
+    def transform_image(self, content_metadata_item):
```
georgebabey: Now that we have this same method in 3 different files, I think we should either move this to a utils helper or the base class.

irfanuddinahmad: Added to utils

integrated_channels/utils.py
```diff
@@ -136,3 +136,16 @@ def strfdelta(tdelta, fmt='{D:02}d {H:02}h {M:02}m {S:02}s', input_type='timedel
...
     return f.format(fmt, **values)
+
+
+def get_image_url(content_metadata_item):
+    """
+    Return the image URI of the content item.
+    """
+    image_url = ''
+    if content_metadata_item['content_type'] == 'program':
+        image_url = content_metadata_item.get('card_image_url')
```
christopappas: probably not a big deal, but figured i would ask: is it possible (and do we want it to be) that this function can return None and not empty string if the `card_image_url` or `image_url` fields don't exist on the dict for some reason?

integrated_channels/degreed/exporters/content_metadata.py
```diff
@@ -78,13 +78,7 @@ def transform_image(self, content_metadata_item):
         """
         Return the image URI of the content item.
         """
-        image_url = ''
```
ziafazal: @irfanuddinahmad I'm wondering why we previously has different transformation logic for `degreed` and `SAP SF`

integrated_channels/utils.py
irfanuddinahmad: @christopappas there is some logic around None ... 
https://github.com/edx/edx-enterprise/blob/a1974db6be53da3e4a11265de24d3a6f0b1e6824/integrated_channels/integrated_channel/exporters/content_metadata.py#L138

ziafazal: @irfanuddinahmad cornerstone does not accept values with `None` so we are skipping those items where we have a None value.

irfanuddinahmad: Thanks @ziafazal

PR #604, opened 2019-10-28
Add Source to the Enterprise Course Enrollments and PendingEnrollment

staubina:
models

[ENT-2082]

**Merge checklist:**
- [x] Translations updated

**Post merge:**

[ENT-2082]: https://openedx.atlassian.net/browse/ENT-2082


justinhynes:
I think I'm following what's going on here but I did have a question. Is the addition of 'source' to the PendingEnrollment and EnterpriseCourseEnrollment models just making it _possible_ to set an EnterpriseEnrollmentSource. 

Is there work required outside the scope of this story that will actually populate the data inside the field? Or am I missing some sort of magic that would set the source at object creation time?

staubina:
Yes, as I mentioned in Standup, this is just to add the columns I am still working on the code to populate the column on enrollment.

justinhynes:
Thanks for the confirmation and sorry for missing that detail.

enterprise/models.py
```diff
@@ -807,6 +807,30 @@ def __repr__(self):
...
+    """
+    Define a Name and Source for all Enterprise Enrollment Sources.
+
+    .. no_pii:
+    """
+
+    MANUAL = 'manual'
+    API = 'enterprise_api'
+    ENROLLMENT_URL = 'enrollment_url'
+    CODE_REDEMPTION = 'code_redemption'
```
brittneyexline: nit: i believe both offers and codes will end up getting mapped to this condition, so this might be better named as offer_redemption, as codes are really just a mechanism to apply offers.

enterprise/models.py
```diff
@@ -807,6 +807,30 @@ def __repr__(self):
...
+    b2c_site = 'b2c_site'
```
brittneyexline: why is b2c_site const variable lowercased?

PR #606, opened 2019-10-28
Add manual enrollment order API method

tuchfarber:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -4,6 +4,6 @@ Your project description goes here.
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "2.0.9"
+__version__ = "2.0.10"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/enterprise/api_client/ecommerce.py
+++ b/enterprise/api_client/ecommerce.py
@@ -65,3 +65,43 @@ class EcommerceApiClient(object):
         if price != mode['min_price']:
             return format_price(price, currency)
         return mode['original_price']
+
+    def create_manual_enrollment_orders(self, enrollments):
+        """
+        Calls ecommerce to create orders for the manual enrollments passed in.
+
+        `enrollments` should be a list of enrollments with the following format:
+        {
+            "lms_user_id": <int>,
+            "username": <str>,
+            "email": <str>,
+            "course_run_key": <str>
+        }
+
+        Since `student.CourseEnrollment` lives in LMS, we're just passing around dicts of the relevant information.
+        """
+        try:
+            order_repsonse = self.client.manual_course_enrollment_order.post(
+                {
+                    "enrollments": enrollments
+                }
+            )
+        except (SlumberBaseException, ConnectionError, Timeout) as exc:
+            LOGGER.exception(
+                "Failed to create order for manual enrollments for the following enrollments: %s. Reason: %s",
+                enrollments,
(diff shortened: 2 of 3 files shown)


enterprise/api_client/ecommerce.py
```diff
@@ -65,3 +65,41 @@ def get_course_final_price(self, mode, currency='$', enterprise_catalog_uuid=Non
...
+                str(exc)
+            )
+        order_creations = order_repsonse["orders"]
+        successful_creations = [order for order in order_creations if order["status"] == "success"]
+        failed_creations = [order for order in order_creations if order["status"] == "failure"]
+        LOGGER.info(
+            "Successfully created orders for the following manual enrollments. %s",
+            successful_creations
+        )
+        LOGGER.error(
```
tuchfarber: Is logging here enough to raise awareness if orders haven't been created? I want to make sure we notice if there's failures.

enterprise/__init__.py
```diff
@@ -4,6 +4,6 @@
 
 from __future__ import absolute_import, unicode_literals
 
-__version__ = "2.0.9"
+__version__ = "2.0.10"
```
crice100: Minor: also update the comment describing this version

enterprise/api_client/ecommerce.py
```diff
@@ -65,3 +65,41 @@ def get_course_final_price(self, mode, currency='$', enterprise_catalog_uuid=Non
...
+        except (SlumberBaseException, ConnectionError, Timeout) as exc:
+            LOGGER.exception(
+                "Failed to create order for manual enrollments for the following enrollments: %s. Reason: %s",
+                enrollments,
+                str(exc)
+            )
+        order_creations = order_repsonse["orders"]
+        successful_creations = [order for order in order_creations if order["status"] == "success"]
+        failed_creations = [order for order in order_creations if order["status"] == "failure"]
+        LOGGER.info(
```
crice100: Minor: maybe only log these messages if there are successful/failed creations

enterprise/api_client/ecommerce.py
crice100: I'm not sure. We could probably check opsgenie and make a new alert if we can't find any existing generic error checks

tuchfarber:
I was having testing issues and made a separate branch, then I ended up fixing the tincan and CHANGELOG stuff in the other branch 🤦‍. Carried over the changes from that branch and made the logging optional.

tuchfarber:
Re: testing, I think it's just a passthrough layer to another service so it's not worth testing since we would mostly just be testing the enterprise API client. We can add testing to the places we call it to make sure they're shuttling the data correctly.

enterprise/api_client/ecommerce.py
```diff
@@ -65,3 +65,43 @@ def get_course_final_price(self, mode, currency='$', enterprise_catalog_uuid=Non
...
+                    "enrollments": enrollments
+                }
+            )
+        except (SlumberBaseException, ConnectionError, Timeout) as exc:
+            LOGGER.exception(
+                "Failed to create order for manual enrollments for the following enrollments: %s. Reason: %s",
+                enrollments,
+                str(exc)
+            )
+        order_creations = order_repsonse["orders"]
```
HammadAhmadWaqas: @tuchfarber `order_repsonse` won't be declared if there will be an exception  in `manual_course_enrollment_order`
I think you want to return after logging exception?

PR #609, opened 2019-11-01
Enterprise selection page

muhammad-ammar:
**Description:** Create a page that will allow an enterprise learner to select one of many linked enterprises. This page will be shown to an enterprise learner right after login. 

**JIRA:** [ENT-2438](https://openedx.atlassian.net/browse/ENT-2438)

**NOTE:** Currently text on enterprise selection page is in progress as we are waiting on product feedback. Other than that functionality is completed I think.

**Testing:**  This PR is deployed on Sandbox

Sandbox: https://mammar.sandbox.edx.org/

1. Goto Enterprise Customers admin page and link a learner to multiple enterprises
2. Goto https://mammar.sandbox.edx.org/enterprise/select/active?success_url=/dashboard
3. Please note the `success_url` in the **step 2** is mandatory and it should be any valid edx-platform url
4. Dropdown on the enterprise selection page should show learner's all linked enterprises
5. Select an enterprise from the dropdown and press the submit button
6. You should be redirect to `success_url`
7. Currently there is no admin UI that shows which enterprise is active/inactive for learner, you need to verify that by ssh to sandbox and execute the below query `select * from enterprise_enterprisecustomeruser where user_id=<user_id>;` on mysql shell. In the results returned from the query you can verify that only the selected enterprise is active and all the other enterprises are inactive. 

**Merge checklist:**
- [ ] Translations updated

**Post merge:**


tests/test_enterprise/views/test_enterprise_selection.py
```diff
@@ -0,0 +1,165 @@
...
+                user_id=self.user.id,
+                enterprise_customer=enterprise_customer
+            )
+
+        enterprises = EnterpriseCustomerUser.objects.filter(
+            user_id=self.user.id
+        ).values_list(
+            'enterprise_customer__uuid', 'enterprise_customer__name'
+        )
+        self.enterprise_choices = [(str(k), v) for k, v in enterprises]
```
ziafazal: nit: replace `k,v`  with `uui, name`

tests/test_enterprise/views/test_enterprise_selection.py
```diff
@@ -0,0 +1,165 @@
...
+
+        # create a temporary template file
+        # rendering `enterprise/enterprise_customer_select_form.html` fails becuase of dependency on edx-platform
+        tpl = tempfile.NamedTemporaryFile(
```
ziafazal: Creating a temp file does not make sense. Can we mock the the template?

tests/test_enterprise/views/test_enterprise_selection.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def test_view_get(self):
+        """
+        Test that view HTTP GET works as expected.
+        """
+        self._login()
+        response = self.client.get(self.url + '?success_url={}'.format(self.success_url))
+        assert response.status_code == 200
+
+        assert response.context[u'select_enterprise_message_title'] == u'Select an organization'
```
ziafazal: unicode keys does not make sense.

enterprise/views.py
```diff
@@ -808,6 +817,68 @@ def post(self, request):
...
+class EnterpriseSelectionView(FormView):
+    """
+    Allow an enterprise learner to activate one of learner's linked enterprises.
+    """
+
+    form_class = EnterpriseSelectionForm
+    template_name = 'enterprise/enterprise_customer_select_form.html'
+
+    @method_decorator(login_required)
+    def dispatch(self, request, *args, **kwargs):
```
ziafazal: what is the reason for overriding this `dispatch`?

enterprise/forms.py
```diff
@@ -0,0 +1,63 @@
...
+            raise forms.ValidationError(_("Enterprise not found"))
+
+        # verify that learner is really a member of selected enterprise
+        if not EnterpriseCustomerUser.objects.filter(enterprise_customer=enterprise, user_id=self._user_id).exists():
+            raise forms.ValidationError(_("Wrong Enterprise"))
+
+        try:
+            # urlparse returns ParseResult(scheme='', netloc='', path=u'', params='', query='', fragment='')
+            # we are validating if path is a valid url
+            resolve(urlparse(cleaned_data['success_url'])[2])
```
ziafazal: I think `resolve(cleaned_data['success_url'])` should work we don't need to use `urlparse`

enterprise/views.py
muhammad-ammar: To ensure that only the logged in user can access to page.

tests/test_enterprise/views/test_enterprise_selection.py
muhammad-ammar: I tried but didn't found a way to mock the template. `ForView.template_name` must be a path to a template that is why we need a temporary file.

muhammad-ammar:
@ziafazal I am looking at https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/conventions/accessibility.html#test-your-code-for-accessibility to ensure a11y.

enterprise/forms.py
muhammad-ammar: @ziafazal I tried `resolve` on complete url and it gives `Resolver404` exception. Below is what I tried on sandbox django shell

```python
>>> from six.moves.urllib.parse import urlparse
>>> from django.urls import Resolver404, resolve
>>> url = 'https://mammar.sandbox.edx.org/enterprise/select/active'
>>> resolve(url)
  File "<console>", line 1, in <module>
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/urls/base.py", line 27, in resolve
    return get_resolver(urlconf).resolve(path)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/urls/resolvers.py", line 395, in resolve
    raise Resolver404({'path': path})
Resolver404: {u'path': u'https://mammar.sandbox.edx.org/enterprise/select/active'}
>>> urlparse(url)
ParseResult(scheme='https', netloc='mammar.sandbox.edx.org', path='/enterprise/select/active', params='', query='', fragment='')
>>> resolve(urlparse(url)[2])
ResolverMatch(func=enterprise.views.EnterpriseSelectionView, args=(), kwargs={}, url_name=enterprise_select_active, app_names=[], namespaces=[])
```

ziafazal: I thought we'd have relative url in `cleaned_data['success_url']`

enterprise/views.py
ziafazal: If that is the case then I think this is more clean way to enforce `login_required`
```
@method_decorator(login_required, name='dispatch')
class EnterpriseSelectionView(FormView):
```

enterprise/forms.py
muhammad-ammar: I have refactored the code and now success_url is a relative url.

muhammad-ammar:
@wittjeff I have refactored the code and now the `aria-live` element will always present on the page. I have tested with VoiceOver and it works perfectly. I have also addressed other a11y concerns as we discussed yesterday. Can you please review the page with new changes?

PR #610, opened 2019-11-01
Add order creation to the enterprise customer manual enroll tool

Tj-Tracy:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 ----------
 
+[2.0.13] - 2019-11-07
+---------------------
+
+* Add manual order creation to enterprise manual enrollment admin form
+
 [2.0.12] - 2019-11-06
 ---------------------
 
--- a/enterprise/admin/views.py
+++ b/enterprise/admin/views.py
@@ -39,6 +39,7 @@ from enterprise.admin.utils import (
     split_usernames_and_emails,
     validate_email_to_link,
 )
+from enterprise.api_client.ecommerce import EcommerceApiClient
 from enterprise.api_client.lms import EnrollmentApiClient
 from enterprise.constants import PAGE_SIZE
 from enterprise.models import (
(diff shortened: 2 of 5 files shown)


enterprise/admin/views.py
```diff
@@ -791,7 +792,15 @@ def _enroll_users(
...
             if pending:
                 pending_messages.append(cls.get_pending_enrollment_message(pending, program_identifier))
-
+        
+        enrollments = map(lambda success: {
+            "lms_user_id": success.id,
+            "email": success.email,
+            "username": success.username,
+            "course_run_key": course_id,
+        }, succeeded)
```
tuchfarber: While `map`s work and make sense from a functional point of view, a list comprehension here is a little more in line with the rest of our code base and they're generally more "pythonic". 

Fun fact: `map`, `reduce` and `filter` were all actually going to be removed from Python 3, but some people got real mad about that, so they stayed.

enterprise/admin/views.py
```diff
@@ -748,7 +749,7 @@ def _enroll_users(
             notify: Whether to notify (by email) the users that have been enrolled
         """
         pending_messages = []
-
+        ecommerceApi = EcommerceApiClient(request.user);
```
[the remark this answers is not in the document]
Tj-Tracy: Interesting - I was going to ask this question earlier today, but decided "I'll just write it like a JS developer and someone will correct me." 

I'll refactor it - good learning experience 😃

enterprise/admin/views.py
Tj-Tracy: Interesting - I was going to ask this question earlier today, but decided "I'll just write it like a JS developer and someone will correct me."

I'll refactor it - good learning experience 😃

CHANGELOG.rst
```diff
@@ -14,6 +14,10 @@ Change Log
 Unreleased
 ----------
 
+[2.0.11] - 2019-11-1
+---------------------
+* Added a Ecommerce API client call to Enterprise Customer manual enrollment form
```
crice100: Don't forget to bump the version

enterprise/admin/views.py
Tj-Tracy: Refactored!

CHANGELOG.rst
Tj-Tracy: version bumped.

PR #611, opened 2019-11-04
ENT-2419 | Create ecommerce orders during user reg. from pending enrollments

justinhynes:
[ENT-2419]
- Update 'EnterpriseCustomerUser' model. During the 'enroll' call we will use EcommerceApiClient to create an order for any enrollments pending for user.
- Added method to retrieve ecommerce worker in 'utils.py' used when creating an instance of the EcommerceApiClient. 

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

[ENT-2419]: https://openedx.atlassian.net/browse/ENT-2419

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -40,9 +40,16 @@ from model_utils.models import TimeStampedModel
 
 from enterprise import utils
 from enterprise.api_client.discovery import CourseCatalogApiClient, get_course_catalog_api_service_client
+from enterprise.api_client.ecommerce import EcommerceApiClient
 from enterprise.api_client.lms import EnrollmentApiClient, ThirdPartyAuthApiClient, parse_lms_api_datetime
 from enterprise.constants import ALL_ACCESS_CONTEXT, ENTERPRISE_OPERATOR_ROLE, json_serialized_course_modes
-from enterprise.utils import CourseEnrollmentDowngradeError, CourseEnrollmentPermissionError, get_configuration_value
+from enterprise.utils import (
+    CourseEnrollmentDowngradeError,
+    CourseEnrollmentPermissionError,
+    NotConnectedToOpenEdX,
+    get_configuration_value,
+    get_ecommerce_worker_user,
+)
 from enterprise.validators import validate_image_extension, validate_image_size
 
 try:
@@ -741,6 +748,8 @@ class EnterpriseCustomerUser(TimeStampedModel):
                 enterprise_customer_user=self,
                 course_id=course_run_id
             )
+            # create an ecommerce order for the course enrollment
+            self.create_order_for_enrollment(course_run_id)
         elif enrolled_in_course and course_enrollment.get('mode') in paid_modes and mode in audit_modes:
             # This enrollment is attempting to "downgrade" the user from a paid track they are already in.
             raise CourseEnrollmentDowngradeError(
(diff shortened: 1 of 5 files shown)


enterprise/models.py
```diff
@@ -763,9 +772,39 @@ def unenroll(self, course_run_id):
...
+                ecommerce_api_client = EcommerceApiClient(ecommerce_service_worker)
+                # pull data needed for creating the order
+                course_enrollments = [{
+                    'lms_user_id': self.user_id,
+                    'username': self.username,
+                    'email': self.user_email,
+                    'course_run_key': course_run_id,
+                }]
+                ecommerce_api_client.create_manual_enrollment_orders(course_enrollments)
+            except NotConnectedToOpenEdX:
```
justinhynes: According to the documentation, it doesn't seem like this exception is likely to be occur, but I added it since the init call _can_ raise this error. Let me know if I'm being too cautious.

enterprise/utils.py
```diff
@@ -982,8 +982,22 @@ def get_enterprise_worker_user():
...
+
+
+def get_ecommerce_worker_user():
+    """
+    Return the user object of ecommerce worker user.
+    """
+    return __get_service_worker(settings.ECOMMERCE_SERVICE_WORKER_USERNAME)
+
+
+def __get_service_worker(service_worker_username):
```
justinhynes: Still a Python newbie in general. I have read previously that starting a method with "_" or "__" is a way to denote a private method. The idea is that I wanted to deter people from calling "__get_service_worker" directly and rather use "get_ecommerce_worker_user" or the pre-existing "get_enterprise_worker_user". Let me know if I'm violating any specific convention here.

tuchfarber: I think this is a good way to do it, just change it from a `__` to a `_` prefix. Double-underscore ("dunder"), is usually for a class's operator overrides or constructor (e.g. `__init__`, `__gt__`, etc)

enterprise/models.py
```diff
@@ -763,9 +772,39 @@ def unenroll(self, course_run_id):
...
+        """
+        Update the session of a request for this learner.
+        """
         request.session['enterprise_customer'] = self.enterprise_customer.serialized
 
+    def create_order_for_enrollment(self, course_run_id):
+        """
+        Create an order on the Ecommerce side for tracking the course enrollment of a enterprise customer user.
+        """
+        LOGGER.info("Creating order for enterprise customer with id [%s] for enrollment in course with id: [%s]",
```
tuchfarber: Note on terminology:
- An "enterprise customer" is the actual business we're partnering with. 
- A "learner" is the edX-y way to say what most people would call a "user". You can also just use "user".
- An "enterprise customer user" is a learner that's attached to an enterprise customer. 

Also, confusingly, an `EnterpriseCustomerUser` has an `id` that's different from the `User`'s `id` (DBs and all that...).

So. This should probably say "...for enterprise learner with user id [%s] ...".

enterprise/models.py
tuchfarber: The caution is good. A couple python tips:
- You can grab a raised exception, stringify it, and log it directly. This lets us re-use some error code strings. So instead of 
    ```
    except NotConnectedToOpenEdX:
    ```
    you could set something like
    ```
    failure_log_msg = "Could not create order for user with id [%s] for enrollment in course with id [%s]. Reason: [%s]"
    ```
    above this block, then do
    ```
    except NotConnectedToOpenEdX as exc:
        LOGGER.exception(failure_log_msg, user_id, course_id, str(exc))
    ```
- The above would also allow you to do something like this at the bottom:
```
else:
    LOGGER.warning(failure_log_msg, user_id, course_id, "Failed to retrieve a valid ecommerce worker."
```

- You generally want to only do the 1 thing that might cause the exception inside the `try` block. Python's `try` system has a couple options of block you can use:
`try`: code to run that might contain exception
`except`: catch the exception if it happens
`else`: only run if `try` ran without an exception
`finally`: run always after the top blocks are done (used for cleanup in some projects, not heavily used here)
So this could could be transformed into:
```
try:
    # create client
except:
    # log error
else:
    # make api call with data
```

PR #619, opened 2019-11-11
ENT-2468 | Better handle when Integrated Channels return unexpected results

HammadAhmadWaqas:
whenever an integrated channel returns non-json type of content, raising
RequestException and handling it with other RequestException handlers.

ENT-2468

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/integrated_channels/sap_success_factors/client.py
+++ b/integrated_channels/sap_success_factors/client.py
@@ -70,10 +70,10 @@ class SAPSuccessFactorsAPIClient(IntegratedChannelApiClient):  # pylint: disable
         )
 
         response.raise_for_status()
-        data = response.json()
         try:
+            data = response.json()
             return data['access_token'], datetime.datetime.utcfromtimestamp(data['expires_in'] + int(time.time()))
-        except KeyError:
+        except (KeyError, ValueError):
             raise requests.RequestException(response=response)
 
     def __init__(self, enterprise_configuration):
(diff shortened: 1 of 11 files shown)


integrated_channels/sap_success_factors/client.py
```diff
@@ -86,28 +86,33 @@ def __init__(self, enterprise_configuration):
...
-        oauth_access_token, expires_at = SAPSuccessFactorsAPIClient.get_oauth_access_token(
-            self.enterprise_configuration.sapsf_base_url,
-            self.enterprise_configuration.key,
-            self.enterprise_configuration.secret,
-            self.enterprise_configuration.sapsf_company_id,
-            self.enterprise_configuration.sapsf_user_id,
-            self.enterprise_configuration.user_type
-        )
+        now = datetime.datetime.utcnow()
+        if self.session is None or self.expires_at is None or now >= self.expires_at:
```
ziafazal: this statement can be simplified like this
```
if self.expires_at is None or now >= self.expires_at:
```

integrated_channels/sap_success_factors/models.py
```diff
@@ -200,7 +201,15 @@ def unlink_inactive_learners(self):
...
-        sap_learner_manager.unlink_learners()
+        try:
+            sap_learner_manager.unlink_learners()
+        except ClientError as exc:
+            LOGGER.error(
+                'Failed to unlink learners for integrated channel [%s] [%s]',
+                self.enterprise_configuration.enterprise_customer.name,
+                self.enterprise_configuration.channel_code,
+            )
+            LOGGER.error(exc)
```
ziafazal: `exc` should be appended to previous `LOGGER.error` message.

tests/test_integrated_channels/test_sap_success_factors/test_client.py
```diff
@@ -242,21 +259,19 @@ def test_sap_api_application_error(self):
...
+           and make a call again and notice 2 OAuth calls in total are required.
+        """
         expired_token_response_body = {"expires_in": 0, "access_token": self.access_token}
         responses.add(
             responses.POST,
             self.url_base + self.oauth_api_path,
             json=expired_token_response_body,
             status=200
         )
-
```
ziafazal: why we need to remove this `responses` block.

integrated_channels/sap_success_factors/client.py
HammadAhmadWaqas: actually I added the same condition to make code consistent with another _create_session method that is in Degreed client.
https://github.com/edx/edx-enterprise/blob/master/integrated_channels/degreed/client.py#L190

integrated_channels/sap_success_factors/models.py
HammadAhmadWaqas: fixed

tests/test_integrated_channels/test_sap_success_factors/test_client.py
HammadAhmadWaqas: It was not mandatory to remove this block.
I removed this block as it was not necessary to add a new response as it doesn't matter if 2nd oauth values are going to expire instantly or not.

integrated_channels/sap_success_factors/models.py
```diff
@@ -200,7 +201,15 @@ def unlink_inactive_learners(self):
         Unlink inactive SAP learners form their related enterprises
         """
         sap_learner_manager = self.get_learner_manger()
-        sap_learner_manager.unlink_learners()
+        try:
+            sap_learner_manager.unlink_learners()
+        except ClientError as exc:
+            LOGGER.error(
```
brittneyexline: nit: use `LOGGER.exception(` here.
also, i am trying to understand the rationale of not raising the exception again here after logging it? is this point in the code the top of the call stack?

HammadAhmadWaqas: @brittneyexline 
this method is getting [called](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/tasks.py#L145) from a task `unlink_inactive_learners` which is actually on top of the call stack in the code. if we see in the code, called method is not supposed to return anything and the task itself is not doing much after calling this method.
So in the case of client error task will print error stack trace and exit.

PR #620, opened 2019-11-12
[MICROBA-56] Added Source to the EnterpriseEnrollments for Enterprise APIs

staubina:
[ENT-2082]

Now [MICROBA-56]

This PR adds the Source to Enterprise CourseEnrollments and Pending
Enrollments that are made through the EnterpriseCustomer
Enrollment API.

**Merge checklist:**
- [x] Translations updated

**Post merge:**

[MICROBA-56]: https://openedx.atlassian.net/browse/MICROBA-56

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -718,7 +718,7 @@ class EnterpriseCustomerUser(TimeStampedModel):
             return client.get_remote_id(self.enterprise_customer.identity_provider, user.username)
         return None
 
-    def enroll(self, course_run_id, mode, cohort=None):
+    def enroll(self, course_run_id, mode, cohort=None, source_slug=None):
         """
         Enroll a user into a course track, and register an enterprise course enrollment.
         """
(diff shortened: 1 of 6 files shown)


enterprise/models.py
```diff
@@ -746,9 +746,12 @@ def enroll(self, course_run_id, mode, cohort=None):
                 'is_upgrading': is_upgrading,
             })
             try:
-                EnterpriseCourseEnrollment.objects.get_or_create(
+                enrollment, created = EnterpriseCourseEnrollment.objects.update_or_create(
```
staubina: @brittneyexline Do you think this will be an issue? Changing this from a `get_or_create` to an `update_or_create`?

This change was triggered because it looks like the Enterprise Course Enrollment is already created by the time we get to this point while converting a PendingEnrollment to an EnterpriseCourseEnrollment.

The outcome from this would be an updated Source if the `EnterpriseCourseEnrollment` already existed.

tests/test_enterprise/api/test_views.py
```diff
@@ -1760,11 +1764,11 @@ def test_enterprise_customer_course_enrollments_detail_success(
                     payload.get('course_mode'),
                     cohort=payload.get('cohort'),
                 )
-        elif 'user_email' in post_data:
+        elif 'user_email' in payload and payload.get('is_active', True):
```
staubina: Note: The test code in this branch was not executing.  The changes here are my additions and changes needed to get this branch working in the test.

enterprise/models.py
crice100: If we're updating, could this overwrite the source?

staubina: Yes, an update will change the source to what is set here.  I am trying to find where it is being created first right now.

staubina: It looks like the issue is that we have a Task that is run on every save to the enrollment table that checks if the enrollment is an EnterpriseCourseEnrollment.  Then it creates the Enrollment.

In this case, we enroll the learner and then we create the EnterpriseCourseEnrollment in the task.  Then in the code, we attempt to `get_or_create` the EnterpriseCourseEnrollment with the Source and it already exists so the Source is not set.

PR #622, opened 2019-11-14
remove success url validation

muhammad-ammar:
**Description:** Initially we have added validation on `success_url` with the assumption that it will always be a platform url and also the url will not raise `Resolver404` if we apply `django.urls.resolve` function but it turns out this is not the case, for example, When we sign in through Google, the `success_url` is set to `/auth/complete/google-oauth2/?` and  `django.urls.resolve` function raises `Resolver404 ` on such a url due to `?` in the end. So we have decided to remove the validation on `success_url`.

**Testing:** Now both the below urls will work fine. 
1. https://mammar.sandbox.edx.org/enterprise/select/active?success_url=/dashboard/?
2. https://mammar.sandbox.edx.org/enterprise/select/active?success_url=/dashboard

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/spec/javascripts/enterprise_select_spec.js
+++ b/spec/javascripts/enterprise_select_spec.js
@@ -39,15 +39,12 @@ describe('Enterprise Selection Page', function () {
         });
 
         it('works expected on correct post data', function () {
-            var response = {
-                'success_url': '/dashboard'
-            };
             var redirectSpy = spyOn(window, 'redirectToURL');
 
             jasmine.Ajax
             .stubRequest('/enterprise/select/active')
             .andReturn({
-                responseText: JSON.stringify(response)
+                responseText: JSON.stringify({})
             });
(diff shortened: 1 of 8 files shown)


spec/javascripts/enterprise_select_spec.js
```diff
@@ -39,15 +39,12 @@ describe('Enterprise Selection Page', function () {
...
 
         it('works expected on correct post data', function () {
-            var response = {
-                'success_url': '/dashboard'
-            };
             var redirectSpy = spyOn(window, 'redirectToURL');
 
             jasmine.Ajax
             .stubRequest('/enterprise/select/active')
             .andReturn({
```
ziafazal: nit: we can remove `.andReturn` statement if we are not expecting response text

PR #623, opened 2019-11-15
ENT-2346 Overridden get in EnterpriseCustomerUser manager

irfanuddinahmad:
Description: This PR updates the Manager for EnterpriseCustomerUser model with an overridden get() method. This will enable a learner to be associated with multiple enterprises and the get() will pick the first entry from the queryset. As part of [ENT-2338](https://openedx.atlassian.net/browse/ENT-2338) it will be ensured that the first entry is always the one currently relevant for the learner.

JIRA: https://openedx.atlassian.net/browse/ENT-2346

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,10 @@ Change Log
 
 Unreleased
 ----------
+[2.0.22] - 2019-11-18
+---------------------
+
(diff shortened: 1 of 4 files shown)


irfanuddinahmad:
> @irfanuddinahmad need to add some unit tests to support your changes.

Added tests

PR #624, opened 2019-11-19
[ENT-2082] Added Source for Enrollments that come through an Enrollme…

staubina:
…nt URL

**Merge checklist:**
- [x] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -39,7 +39,12 @@ from enterprise.api_client.ecommerce import EcommerceApiClient
 from enterprise.api_client.lms import CourseApiClient, EmbargoApiClient, EnrollmentApiClient
 from enterprise.decorators import enterprise_login_required, force_fresh_session
 from enterprise.forms import ENTERPRISE_SELECT_SUBTITLE, EnterpriseSelectionForm
-from enterprise.models import EnterpriseCourseEnrollment, EnterpriseCustomerCatalog, EnterpriseCustomerUser
+from enterprise.models import (
+    EnterpriseCourseEnrollment,
+    EnterpriseCustomerCatalog,
+    EnterpriseCustomerUser,
+    EnterpriseEnrollmentSource,
+)
 from enterprise.utils import (
     CourseEnrollmentDowngradeError,
     CourseEnrollmentPermissionError,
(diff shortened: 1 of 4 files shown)


enterprise/views.py
```diff
@@ -160,6 +165,34 @@ def render_page_with_error_code_message(request, context_data, error_code, log_m
...
+):
+    """
+    Get or Create the Enterprise Course Enrollment.
+
+    This method will set the Source if the Enterprise Course Enrollment (ECE)
+    was already created. It is possible the ECE will have been created by the
+    background task at the time of enrollment so we need to update the source
+    in that case.
+    """
+    source = EnterpriseEnrollmentSource.get_source(EnterpriseEnrollmentSource.ENROLLMENT_URL)
```
crice100: I assume we know that the source should always be _ENROLLMENT_URL_, even for enrollments that already exist and that we're updating?

enterprise/views.py
```diff
@@ -160,6 +165,34 @@ def render_page_with_error_code_message(request, context_data, error_code, log_m
...
+    # Create the Enterprise backend database records for this course
+    # enrollment
+    enterprise_course_enrollment, created = EnterpriseCourseEnrollment.objects.get_or_create(
```
crice100: Just want to make sure I'm reading this correctly: if this call does create the enrollment, it will also set its source to _ENROLLMENT_URL_?

enterprise/views.py
staubina: When the views that use this method are called we want it to be `ENROLLMENT_URL` if it has not already been set.

There are cases where we create the CourseEnrollment before getting here that do not know what the Source is, so we update the Source and set it in this code if it is Null.

enterprise/views.py
staubina: Yes. If it `GETs` the Enrollment it will not change the Source.

enterprise/views.py
```diff
@@ -485,9 +518,9 @@ def create_enterprise_course_enrollment(request, consent_record, course_id):
...
         enterprise_customer_user.update_session(request)
-        __, created = EnterpriseCourseEnrollment.objects.get_or_create(
-            enterprise_customer_user=enterprise_customer_user,
-            course_id=course_id,
+        __, created = get_create_ent_enrollment(
+            course_id,
+            enterprise_customer_user
         )
         if created:
             track_enrollment('data-consent-page-enrollment', request.user.id, course_id, request.path)
```
brittneyexline: based on this, i don't think enterprise enrollments created through this route are guaranteed to come from enrollment urls, though i'm not sure what would result in an enterprise enrollment getting created here and not by something else...

enterprise/views.py
```diff
@@ -160,6 +165,34 @@ def render_page_with_error_code_message(request, context_data, error_code, log_m
...
+    # enrollment
+    enterprise_course_enrollment, created = EnterpriseCourseEnrollment.objects.get_or_create(
+        enterprise_customer_user=enterprise_customer_user,
+        course_id=course_id,
+        defaults={
+            'source': source
+        }
+    )
+    if not created and enterprise_course_enrollment.source is None:
+        enterprise_course_enrollment.source = source
```
brittneyexline: i'm not sure if this should be set if it's empty; technically we don't know where it came from and i'm not sure if we should assume it is from an enrollment url.

PR #625, opened 2019-11-19
Separate enterprise enrollments from all other enrollments in django admin

adamstankiewicz:
**Merge checklist:**
- [x] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/admin/__init__.py
+++ b/enterprise/admin/__init__.py
@@ -324,7 +324,7 @@ class EnterpriseCustomerUserAdmin(admin.ModelAdmin):
 
     def enterprise_enrollments(self, enterprise_customer_user):
         """
-        Return a string representing the courses a given EnterpriseCustomerUser is enrolled in
+        Return a string representing a given EnterpriseCustomerUser's enterprise course enrollments
 
         Args:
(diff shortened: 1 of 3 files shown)


enterprise/admin/__init__.py
```diff
@@ -320,17 +322,57 @@ def get_search_results(self, request, queryset, search_term):
...
+        Return a string representing a given EnterpriseCustomerUser's non-enterprise course enrollments
+
+        Args:
+            enterprise_customer_user: The instance of EnterpriseCustomerUser
+                being rendered with this admin form.
+        """
+        all_course_ids = self._get_all_enrollments(enterprise_customer_user)
+        enterprise_course_ids = self._get_enterprise_course_enrollments(enterprise_customer_user)
+        # remove overlapping enterprise enrollments from all enrollments
+        course_ids = set(all_course_ids) - set(enterprise_course_ids)
```
adamstankiewicz: Not entirely sure if this (i.e., removing the enterprise enrollments from all enrollments) is the best way to find/retrieve non-enterprise course enrollments. Definitely open to feedback here 😃

christopappas: Makes sense to me

PR #629, opened 2019-11-21
ENT-2346 Set the active enterprise for xapi

irfanuddinahmad:
JIRA: https://openedx.atlassian.net/browse/ENT-2346

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/integrated_channels/xapi/serializers.py
+++ b/integrated_channels/xapi/serializers.py
@@ -33,40 +33,22 @@ class LearnerInfoSerializer(ImmutableStateSerializer):
     user_country_code = serializers.CharField(source='profile.country.code')
     user_account_creation_date = serializers.DateTimeField(source='date_joined')
 
-    enterprise_user_id = serializers.SerializerMethodField()
-    enterprise_sso_uid = serializers.SerializerMethodField()
-
-    def get_enterprise_user_id(self, obj):
-        """
-        Get enterprise user id from user object.
-
-        Arguments:
-            obj (User): Django User object
-
-        Returns:
-            (int): Primary Key identifier for enterprise user object.
-        """
-        # An enterprise learner can not belong to multiple enterprise customer at the same time
-        # but if such scenario occurs we will pick the first.
-        enterprise_learner = EnterpriseCustomerUser.objects.filter(user_id=obj.id).first()
-
-        return enterprise_learner and enterprise_learner.id
-
-    def get_enterprise_sso_uid(self, obj):
-        """
-        Get enterprise SSO UID.
-
-        Arguments:
-            obj (User): Django User object
-
-        Returns:
-            (str): string containing UUID for enterprise customer's Identity Provider.
-        """
-        # An enterprise learner can not belong to multiple enterprise customer at the same time
-        # but if such scenario occurs we will pick the first.
-        enterprise_learner = EnterpriseCustomerUser.objects.filter(user_id=obj.id).first()
-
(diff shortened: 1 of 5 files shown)


integrated_channels/xapi/serializers.py
```diff
@@ -33,40 +33,24 @@ class LearnerInfoSerializer(ImmutableStateSerializer):
...
+            data = {
+                'enterprise_user_id': enterprise_learner.id,
+                'enterprise_sso_uid': enterprise_learner.get_remote_id(),
+            }
+        else:
+            data = {
+                'enterprise_user_id': None,
+                'enterprise_sso_uid': None,
+            }
+        return dict(representation, **data)
```
muhammad-ammar: A small improvement to remove the else part of if condition and removal of enterprise_customer variable.

```python
def to_representation(self, instance):
    representation = super(LearnerInfoSerializer, self).to_representation(instance)
    
    data = {
        'enterprise_user_id': None,
        'enterprise_sso_uid': None,
    }

    enterprise_learner = EnterpriseCustomerUser.objects.filter(
        user_id=instance.id,
        enterprise_customer=self.context.get('enterprise_customer')
    ).first()
    if enterprise_learner:
        data = {
            'enterprise_user_id': enterprise_learner.id,
            'enterprise_sso_uid': enterprise_learner.get_remote_id(),
        }
    
    return dict(representation, **data)
```

irfanuddinahmad: Done

integrated_channels/xapi/serializers.py
```diff
@@ -33,40 +33,22 @@ class LearnerInfoSerializer(ImmutableStateSerializer):
...
-
-        Returns:
-            (str): string containing UUID for enterprise customer's Identity Provider.
-        """
-        # An enterprise learner can not belong to multiple enterprise customer at the same time
-        # but if such scenario occurs we will pick the first.
-        enterprise_learner = EnterpriseCustomerUser.objects.filter(user_id=obj.id).first()
-
-        return enterprise_learner and enterprise_learner.get_remote_id()
+    def to_representation(self, instance):
```
ziafazal: I'm wondering what could be the reason for using `to_representation` method instead of field methods?.

tests/test_integrated_channels/test_xapi/test_serializers.py
```diff
@@ -57,7 +57,10 @@ def test_data(self, mock_third_party_api):
 
         with mock.patch.object(self.user, 'profile', create=True) as mock_user_profile:
             mock_user_profile.country.code = 'PK'
-            assert LearnerInfoSerializer(self.user).data == self.expected_data
+            assert LearnerInfoSerializer(
```
ziafazal: We should change the tests to reflect the actual use case. Which is a learner has multiple enterprises and when management command is run to transmit data of a particular enterprise only that enterprise is reflected in user's data.

integrated_channels/xapi/serializers.py
irfanuddinahmad: DRY: The query to the EnterpriseCustomerUser model is not duplicated.

tests/test_integrated_channels/test_xapi/test_serializers.py
irfanuddinahmad: Added another test to verify the expected output in case of user with multiple enterprises

PR #632, opened 2019-11-25
Replaced EDX_API_KEY with JWT in EnrollmentApiClient

sameenfatima78:
This PR is about replacing the `EDX-API-KEY` with `OAuth` authentication method using `JWT.` There are three clients that are relying on the `EDX-API-KEY` based authentication; `CourseApiClient,` `EnrollmentApiClient,` and `ThirdPartyAuthApiClient.` This PR only covers the `EnrollmentmentApiClient.`

### DIFF
--- a/enterprise/api_client/lms.py
+++ b/enterprise/api_client/lms.py
@@ -132,13 +132,14 @@ class EmbargoApiClient(object):
                 return redirect_url
 
 
-class EnrollmentApiClient(LmsApiClient):
+class EnrollmentApiClient(JwtLmsApiClient):
     """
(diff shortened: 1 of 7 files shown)


enterprise/api_client/lms.py
```diff
@@ -158,6 +159,7 @@ def get_course_details(self, course_id):
             )
             return {}
 
+    @JwtLmsApiClient.refresh_token
     def _sort_course_modes(self, modes):
```
irfanuddinahmad: Do we need the decorator on all the functions?

PR #634, opened 2019-11-25
[ENT-2082] Added the offer redemption and management command Sources to

staubina:
EnterpriseCourseEnrollments

**Merge checklist:**
- [x] Translations updated

**Post merge:**


CHANGELOG.rst
```diff
@@ -13,6 +13,12 @@ Change Log
 
 Unreleased
 ----------
+
+[2.0.27] - 2019-11-26
```
crice100: Don't forget to actually bump the version

enterprise/api/v1/serializers.py
```diff
@@ -184,6 +184,11 @@ def save(self):  # pylint: disable=arguments-differ
         __, created = models.EnterpriseCourseEnrollment.objects.get_or_create(
             enterprise_customer_user=self.enterprise_customer_user,
             course_id=course_id,
+            defaults={
+                'source': models.EnterpriseEnrollmentSource.get_source(
+                    models.EnterpriseEnrollmentSource.OFFER_REDEMPTION
```
crice100: Just curious - how did we pick this default?

staubina: This is less a default and more an indication that if we `create` we will set the `source`.

PR #635, opened 2019-11-26
Replaced EDX_API_KEY with JWT in CourseApiClient

sameenfatima78:
This PR is about replacing the `EDX-API-KEY` with `OAuth` authentication method using `JWT.` There are three clients that are relying on the `EDX-API-KEY` based authentication; `CourseApiClient,` `EnrollmentApiClient,` and `ThirdPartyAuthApiClient.` This PR only covers the `CourseApiClient.`

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -215,12 +215,12 @@ class GrantDataSharingPermissions(View):
 
     preview_mode = False
 
-    def course_or_program_exist(self, course_id, program_uuid):
+    def course_or_program_exist(self, request, course_id, program_uuid):
         """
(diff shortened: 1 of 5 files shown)


enterprise/views.py
```diff
@@ -215,12 +215,12 @@ class GrantDataSharingPermissions(View):
 
     preview_mode = False
 
-    def course_or_program_exist(self, course_id, program_uuid):
+    def course_or_program_exist(self, request, course_id, program_uuid):
```
irfanuddinahmad: Maybe just passing the request.user would suffice

PR #637, opened 2019-11-26
Mark the current learner enterprise-customer active flag to True 

zamanafzal:
Mark the learner current enterprise-customer active flag to True and all other enterprise-customers active flag to false.

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/tpa_pipeline.py
+++ b/enterprise/tpa_pipeline.py
@@ -69,8 +69,13 @@ def handle_enterprise_logistration(backend, user, **kwargs):
     # proceed with the creation of a link between the user and the enterprise customer, then exit.
     enterprise_customer_user, _ = EnterpriseCustomerUser.objects.update_or_create(
         enterprise_customer=enterprise_customer,
+        active=True,
         user_id=user.id
     )
+    # if learner has activated enterprise we need to de-activate other enterprises learner is linked to
+    EnterpriseCustomerUser.objects.filter(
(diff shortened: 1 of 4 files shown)


enterprise/tpa_pipeline.py
```diff
@@ -69,8 +69,13 @@ def handle_enterprise_logistration(backend, user, **kwargs):
     # proceed with the creation of a link between the user and the enterprise customer, then exit.
     enterprise_customer_user, _ = EnterpriseCustomerUser.objects.update_or_create(
         enterprise_customer=enterprise_customer,
+        active=True,
         user_id=user.id
     )
+    # if learner has activated enterprise we need to de-activate other enterprises learner is linked to
+    EnterpriseCustomerUser.objects.filter(
+        user_id=user.id
+    ).exclude(enterprise_customer=enterprise_customer).update(active=False)
```
muhammad-ammar: what about 
```python
EnterpriseCustomerUser.objects.filter(
    user_id=user.id
).exclude(id=enterprise_customer_user.id).update(active=False)
```

zamanafzal: We have to mark the active status of learner all enterprise_customer to False except the enterprise_customer with which user is logged in. So, I am filtering all the enterprise_customer of the learner except the current one and marking them `active=False`.

PR #641, opened 2019-12-04
ENT-2492 | Added an extra field percentage discount in manage learner…

hasnain-naveed:
… form and passed it in ecommerce endpoint for order creation.

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/admin/forms.py
+++ b/enterprise/admin/forms.py
@@ -90,6 +90,13 @@ class ManageLearnersForm(forms.Form):
         ],
     )
     reason = forms.CharField(label=_("Reason for manual enrollment"), required=True)
+    discount = forms.DecimalField(
+        label=_("Discount percentage for manual enrollment"),
+        help_text=_("Discount percentage should be from 0 to 100"),
+        required=True,
+        decimal_places=5,
+        initial=0.0
+    )
 
     class NotificationTypes(object):
         """
@@ -130,6 +137,7 @@ class ManageLearnersForm(forms.Form):
         PROGRAM = "program"
         NOTIFY = "notify_on_enrollment"
         REASON = "reason"
+        DISCOUNT = "discount"
 
     class CsvColumns(object):
         """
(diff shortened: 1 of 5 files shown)


enterprise/admin/forms.py
```diff
@@ -90,6 +90,12 @@ class ManageLearnersForm(forms.Form):
         ],
     )
     reason = forms.CharField(label=_("Reason for manual enrollment"), required=True)
+    discount = forms.IntegerField(
+        label=_("Discount percentage for manual enrollment"),
+        help_text=_("Discount percentage should be between 0 and 100, inclusive 0 and 100."),
```
ziafazal: nit: Discount percentage should be from 0 to 100

enterprise/admin/forms.py
```diff
@@ -90,6 +90,12 @@ class ManageLearnersForm(forms.Form):
...
+        required=True,
+        initial=0
```
ziafazal: it should have default value of zero.

enterprise/admin/forms.py
```diff
@@ -185,6 +192,16 @@ def clean_email_or_username(self):
 
         return email
 
+    def clean_discount(self):
+        """
+        Verify that discount value should be between 0 and 100, inclusive 0 and 100.
+        """
+        discount = self.cleaned_data[self.Fields.DISCOUNT]
```
ziafazal: nit: check `int` type

enterprise/admin/utils.py
```diff
@@ -102,6 +102,9 @@ class ValidationMessages(object):
     INVALID_ENCODING = _(
         "Unable to parse CSV file. Please make sure it is a CSV 'utf-8' encoded file."
     )
+    INVALID_DISCOUNT = _(
+        'Discount percentage should be between 0 and 100, inclusive 0 and 100.'
```
ziafazal: same as above

enterprise/admin/forms.py
```diff
@@ -90,6 +90,12 @@ class ManageLearnersForm(forms.Form):
         ],
     )
     reason = forms.CharField(label=_("Reason for manual enrollment"), required=True)
+    discount = forms.IntegerField(
```
christopappas: On the other work we did, there is the requirement for the ability to store up to 5 decimal places. Because of this, I believe you will need to use a DecimalField here

PR #643, opened 2019-12-06
ENT-2033: Added ability to include/exclude date from report configuration file name

sameenfatima78:
Added "Include Date" field in Enterprise Customer Report Configuration Admin Form as mentioned in ENT-2033.


enterprise/migrations/0083_merge_20191206_1150.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
```
zamanafzal: I think there should be only one migration file.

PR #644, opened 2019-12-06
ENT-2501: Updated context for user with multiple linked enterprises

irfanuddinahmad:
**Description** This PR updates the EnterpriseRoleAssignmentContextMixin to incorporate the case where a user may have multiple contexts e.g. a user with multiple linked enterprises

**JIRA** https://openedx.atlassian.net/browse/ENT-2501

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -23,7 +23,7 @@ from django.conf import settings
 from django.contrib.auth.models import User
 from django.contrib.sites.models import Site
 from django.core import mail
-from django.core.exceptions import NON_FIELD_ERRORS, MultipleObjectsReturned, ObjectDoesNotExist, ValidationError
+from django.core.exceptions import NON_FIELD_ERRORS, ObjectDoesNotExist, ValidationError
 from django.core.files.storage import default_storage
 from django.core.urlresolvers import reverse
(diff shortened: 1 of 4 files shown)


enterprise/models.py
```diff
@@ -1961,30 +1961,25 @@ class EnterpriseRoleAssignmentContextMixin(object):
...
-        return str(enterprise_user.enterprise_customer.uuid)
+        return [str(enterprise_user.enterprise_customer.uuid) for enterprise_user in enterprise_users]
 
     def get_context(self):
         """
-        Return the context for this role assignment class.
+        Return the context for this role assignment class. A list is returned in case of user with multiple contexts.
         """
-        return self.enterprise_customer_uuid
+        return str(self.enterprise_customer_uuids[0]) if len(self.enterprise_customer_uuids) == 1 else \
```
brittneyexline: does calling `self.enterprise_customer_uuids` multiple times here mean that we are recalculating (and thus hitting the db) each time or is that somehow cached? why not always return `self.enterprise_customer_uuids`?

irfanuddinahmad: We dont want to change the get_context() return type in a non-multiple-context scenario

PR #645, opened 2019-12-06
ENT-2464:Aligned xAPI statment formats with TinCan/Rustici standards

ziafazal:
Align xAPI statment formats with TinCan/Rustici standards.
ENT-2484

**Merge checklist:**
- [ ] Translations updated

**Post merge:**


integrated_channels/xapi/statements/base.py
```diff
@@ -15,36 +15,27 @@ class EnterpriseStatement(Statement):
...
 
-    def get_actor(self, username, email):
+    def get_actor(self, user):
         """
         Get actor for the statement.
         """
         return Agent(
-            name=username,
-            mbox='mailto:{email}'.format(email=email),
+            name=u'{id}-{username}'.format(id=user.id, username=user.username),
```
mattdrayer: I am waiting for feedback from SumTotal/Rustici regarding this change -- it may be preferable to instead populate this field with the SSO UID value.

integrated_channels/xapi/statements/base.py
```diff
@@ -15,36 +15,30 @@ class EnterpriseStatement(Statement):
...
     """
 
-    def get_actor(self, username, email):
+    def get_actor(self, user, user_social_auth):
         """
         Get actor for the statement.
         """
+        social_auth_uid = user_social_auth.uid if user_social_auth else ''
+        sso_id = social_auth_uid.split(':')[-1]
+        name = sso_id if sso_id else user.username
```
mattdrayer: I think here we should return the email address of the user and not the edX LMS username.

integrated_channels/xapi/statements/learner_course_enrollment.py
```diff
@@ -16,21 +16,19 @@ class LearnerCourseEnrollmentStatement(EnterpriseStatement):
...
     """
 
-    def __init__(self, user, course_overview, user_details, course_details, *args, **kwargs):
+    def __init__(self, user, user_social_auth, course_overview, *args, **kwargs):
         """
         Initialize and populate statement with learner info and course info.
 
         Arguments:
             user (User): Auth User object containing information about the learner enrolling in the course.
+            user_social_auth (UserSocialAuth): UserSocialAuth object for learner
```
mattdrayer: Not sure how much it matters, but this comment is redundant and doesn't really help to explain what the argument actually represents, it just restates the name of the argument.

integrated_channels/xapi/utils.py
```diff
@@ -26,25 +26,19 @@ def send_course_enrollment_statement(lrs_configuration, course_enrollment):
...
-        course_enrollment.user,
-        context={'enterprise_customer': lrs_configuration.enterprise_customer}
-    )
-    course_details = CourseInfoSerializer(
-        course_enrollment.course,
-        context={'site': lrs_configuration.enterprise_customer.site}
-    )
+    user = course_enrollment.user
     LOGGER.info(
         'Sending course enrollment to xAPI for user: {username} for course: {course_key}'.format(
```
mattdrayer: Tangential -- do we capture if/when the send fails?

integrated_channels/xapi/utils.py
```diff
@@ -59,26 +53,18 @@ def send_course_completion_statement(lrs_configuration, user, course_overview, c
...
     """
-    user_details = LearnerInfoSerializer(
-        user,
-        context={'enterprise_customer': lrs_configuration.enterprise_customer}
-    )
-    course_details = CourseInfoSerializer(
-        course_overview,
-        context={'site': lrs_configuration.enterprise_customer.site}
-    )
     LOGGER.info(
```
mattdrayer: As with the above comment -- do we capture if/when the send fails?

integrated_channels/xapi/statements/base.py
ziafazal: @mattdrayer we are not returning email address here. We are actually returning SSO uid after striping IDP slug and if we don't SSO uid we are returning lms username. Email address of user is in mbox property below.

integrated_channels/xapi/utils.py
ziafazal: we are capturing it [here](https://github.com/edx/edx-enterprise/blob/877511923e5b2a309bc8cb7a374cdace1fee7073/integrated_channels/xapi/management/commands/send_course_enrollments.py#L127)

integrated_channels/xapi/utils.py
ziafazal: yeah capturing it [here](https://github.com/edx/edx-enterprise/blob/877511923e5b2a309bc8cb7a374cdace1fee7073/integrated_channels/xapi/management/commands/send_course_completions.py#L152)

integrated_channels/xapi/statements/base.py
mattdrayer: Right, I just don't think it is going to be very useful to provide a third party LRS with the edX LMS username for the user -- as a matter of fact I don't think we should be sharing this information externally at all.  It would be better to fall back to the email address of the learner as the actor identifier if there is no sso_id value available.

ziafazal: @mattdrayer changed code to return email address instead of LMS username.

PR #646, opened 2019-12-09
ENT-2042 Replaced Edx-Api-Key in Enrollment Api Client (remaining endpoints)

sameenfatima78:
This PR only covers the remaining endpoints of EnrollmentApiClient by cloning the original `EnrollmentApiClient` and using that one in all the remaining endpoints.

### DIFF
--- a/enterprise/admin/__init__.py
+++ b/enterprise/admin/__init__.py
@@ -38,7 +38,7 @@ from enterprise.admin.views import (
     EnterpriseCustomerTransmitCoursesView,
     TemplatePreviewView,
 )
-from enterprise.api_client.lms import CourseApiClient, EnrollmentApiClient
+from enterprise.api_client.lms import CourseApiClient, EnrollmentApiClientJwt
 from enterprise.models import (
     EnrollmentNotificationEmailTemplate,
     EnterpriseCatalogQuery,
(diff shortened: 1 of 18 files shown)


enterprise/admin/__init__.py
```diff
@@ -369,7 +369,7 @@ def _get_all_enrollments(self, enterprise_customer_user):
             enterprise_customer_user: The instance of EnterpriseCustomerUser
                 being rendered with this admin form.
         """
-        enrollment_client = EnrollmentApiClient()
+        enrollment_client = EnrollmentApiClientJwt(enterprise_customer_user.user)
```
brittneyexline: is this the only caller of this client? any reason not to update original client (and all callers) and avoid having two copies of the same functionality hanging around?

sameenfatima78: Yes, the goal is to gradually replace the original client in all endpoints one by one where it is being called. The reason is that some of the endpoints cannot be tested locally so we are picking one endpoint at a time and pushing it to merge. This way, if any issue comes up, it will be easier to track and revert the changes. Also, this would make the PRs concise. There are around 15-16 endpoints in total so, covering all of them in one PR can be a little risky (given that some of those endpoints cannot even be tested locally). I have got an approval on this from @georgebabey too. Initially, I'm pushing only those endpoints that have been tested manually and are working fine. If no issue comes up in any of them after merge, we can start pushing the endpoints that can't be tested. Once the client is replaced completely, I'll remove the copy.

PR #648, opened 2019-12-09
ENT-2477 Enterprise Catalog IDA ADR

brittneyexline:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**


docs/decisions/0002-enterprise-catalog-ida.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Draft
+
+Context
+=======
+We are interested in improving the reliability and performance of our system at large, and we have identified that the enterprise catalog api and related functionality are a central dependency that often causes issues, such as users experiencing random failures during enrollment or other workflows. Almost every feature that our users interact with depends on enterprise catalogs: Enrollment, Code Redemption, Data Sharing Consent, Data transmission to external LMS systems, and API access given to select customers.
+
+Our main goals are as follows:
+
+1. Get the response time of the enterprise customer catalog endpoints to be below 1 second for 99% of calls.
```
bbaker6225: I don't think it necessarily needs to go in the ADR, but for my own curiosity do we have current benchmarks on these endpoints?

docs/decisions/0002-enterprise-catalog-ida.rst
```diff
@@ -0,0 +1,64 @@
...
+
+What is Enterprise Catalog?
+___________________________
+
+Enterprise catalogs are essentially saved elasticsearch queries attached to an Enterprise Customer. There can be many catalogs configured for a given customer, which enables different sets of learners from the same Enterprise to be given access to different sets of courses, and allows different discounts to be configured for different sets of courses.
+
+We expose two apis that are frequently used by various enterprise features: one to retrieve all of the content metadata that would be returned by the configured catalog query, and one to determine if a given course id or program id is part of the set of content returned by the catalog query. These apis exist both as python apis and REST apis that can be hit from within our edX ecosystem and externally from customers. They are wrappers around the search/all endpoint in the discovery service. Although there is caching in place, the calls to discovery are implemented synchronously, and we've found that half the time requests do not hit the cache, and thus the most of the time is spent waiting on the discovery service to return results.
```
bbaker6225: Not a comment on the ADR itself, but I'm wondering how easily fixable the cache misses to the `/search/all` endpoint are. The discovery theme dealt with a similar issue for the `/courses`, `/programs`, and `/course_runs` endpoints so maybe there's more improvement there?

docs/decisions/0002-enterprise-catalog-ida.rst
```diff
@@ -0,0 +1,64 @@
...
+
+.. image:: ../images/enterprise-catalog-api-calls.png
+
+Decisions
+=========
+
+In order to address these performance issues and to more effectively scale and build new features related to enterprise catalogs in the future, we will create a new IDA (Independently Deployable Application) to replace the existing enteprise catalog functionality.
+This IDA will support the existing API contracts established in the edx-enterprise API, but will be a new implementation separate from what we do today.
+
+The microservice will be supported by the following relational data model:
```
bbaker6225: Content catalog and catalog content seem a little confusingly named

docs/decisions/0002-enterprise-catalog-ida.rst
```diff
@@ -0,0 +1,64 @@
...
+
+.. image:: ../images/enterprise-catalog-ida-uml-diagram.png
+
+The microservice will expose the following APIs for use:
+
+* CRUD Operations for Enterprise Catalog
+* is content id in catalog: Given a course, course run, program, or other content id, and an enterprise catalog, returns true if the content id is contained within the catalog's current results.
+* get_catalog_metadata: Given an enterprise catalog, return the full content metadata results for content currently in the catalog.
+* refresh_catalog_metadata: Given an enterprise catalog, kick off a task to get the freshest content metadata for that catalog.
```
bbaker6225: This might just be me carrying baggage from discovery land, but this is a little scary. Data freshness isn't a bad thing, but we've dealt with lots of pain from `refresh_course_metadata` and I'm worried about this becoming something like that.
At the most trivial level, I'd suggest distancing ourselves from the RCM name

docs/decisions/0002-enterprise-catalog-ida.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Context
+=======
+We are interested in improving the reliability and performance of our system at large, and we have identified that the enterprise catalog api and related functionality are a central dependency that often causes issues, such as users experiencing random failures during enrollment or other workflows. Almost every feature that our users interact with depends on enterprise catalogs: Enrollment, Code Redemption, Data Sharing Consent, Data transmission to external LMS systems, and API access given to select customers.
+
+Our main goals are as follows:
+
+1. Get the response time of the enterprise customer catalog endpoints to be below 1 second for 99% of calls.
+
+2. Since many features do not hit the enterprise catalog api endpoints but perform the same calculations directly in code, we would also like to improve the performance of those code paths and drastically reduce the calls to the discovery service.
```
adamstankiewicz: Would it be helpful if an example code path was provided here?

docs/decisions/0002-enterprise-catalog-ida.rst
brittneyexline: there are some performance details in this document: https://docs.google.com/document/d/1zfuLsERLPuJM3MMc1GqN5JLqXyD-3qdRruBzTn_99EM/edit

docs/decisions/0002-enterprise-catalog-ida.rst
brittneyexline: sure, i can list some.

docs/decisions/0002-enterprise-catalog-ida.rst
brittneyexline: it is possible, but for various reasons, we don't think the current architecture is going to scale in the long term. i am hoping that if we have this in a separate service, we will:
* prove that we can iterate faster for development/bug fixes when we own our own service, instead of depending on edx-enterprise being installed in edx-platform.
* be able to build more robust/innovative functionality around catalog curation and enterprise enrollment lifecycles, depending on product needs. currently both are pain points in the system/product
* have more visibility into a key part of the enterprise system and be able to proactively monitor and alert

docs/decisions/0002-enterprise-catalog-ida.rst
brittneyexline: agreed, i will probably update the name. nothing is set in stone here so it's not too late to rename!

docs/decisions/0002-enterprise-catalog-ida.rst
brittneyexline: happy to change names. i think a lot of the issues with discovery (from my understanding) stem from the fact that it is managing bi-directional reads and writes. the discovery service should never be aware of this service, compared to discovery needing to be concerned with lms/ecommerce/drupal. also, though we may have to support search functionality in the future, for now we are not dealing with elasticsearch, and when we do hopefully we can use the latest version. my understanding is that a lot of the time in the search/all endpoint is spent with elasticsearch. 
i think your concerns are legitimate though. ultimately, i'm betting on us being able to have more control over what happens, because this is owned by just us, and not another team. having that control will enable us to change direction much more easily if something isn't working the way we want.

bbaker6225: 👍 I'm on board with that. As long as we're aware of that risk and have a plan of avoiding it going forward (which I think we do), then things should be fine

docs/decisions/0002-enterprise-catalog-ida.rst
bbaker6225: Agree that I don't think current discovery will scale and it makes to account for all of those things

PR #649, opened 2019-12-10
ENT-2042 Replaced Edx-Api-Key in CourseApiClient in GrantDataSharing endpoint

sameenfatima78:
This PR is about replacing the `EDX-API-KEY` with `OAuth` authentication method using `JWT.` There are three clients that are relying on the `EDX-API-KEY` based authentication; `CourseApiClient,` `EnrollmentApiClient,` and `ThirdPartyAuthApiClient.` This PR only covers the `enterprise/views` endpoint by cloning the original `CourseApiClient` and using that one in the mentioned endpoint.

### DIFF
--- a/enterprise/api_client/lms.py
+++ b/enterprise/api_client/lms.py
@@ -341,6 +341,32 @@ class CourseApiClient(LmsApiClient):
             return None
 
 
+class CourseApiClientJwt(JwtLmsApiClient):
+    """
+    Object builds an API client to make calls to the Course API.
+    """
+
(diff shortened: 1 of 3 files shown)


enterprise/api_client/lms.py
```diff
@@ -341,6 +341,32 @@ def get_course_details(self, course_id):
             return None
 
 
+class CourseApiClientJwt(JwtLmsApiClient):
```
christopappas: Again, a nitpick but my preference would be to end the class name with "client", as in `CourseApiJwtClient`

PR #654, opened 2019-12-16
ENT-2042 Replaced Edx-Api-Key in Enrollment Api Client in CourseEnrollmentView-endpoint

sameenfatima78:
This PR is about replacing the `EDX-API-KEY` with `OAuth` authentication method using `JWT.` There are three clients that are relying on the `EDX-API-KEY` based authentication; `CourseApiClient,` `EnrollmentApiClient,` and `ThirdPartyAuthApiClient.` This PR only covers the `enterprise/views.py/CourseEnrollmentView` endpoint by cloning the original `EnrollmentApiClient` and using that one in the mentioned endpoint.

### DIFF
--- a/enterprise/api_client/lms.py
+++ b/enterprise/api_client/lms.py
@@ -316,6 +316,198 @@ class EnrollmentApiClient(LmsApiClient):
         return self.client.enrollment.get(user=username)
 
 
+class EnrollmentApiClientJwt(JwtLmsApiClient):
+    """
+    Object builds an API client to make calls to the Enrollment API.
+    """
+
+    API_BASE_URL = settings.ENTERPRISE_ENROLLMENT_API_URL
+
+    @JwtLmsApiClient.refresh_token
+    def get_course_details(self, course_id):
+        """
+        Query the Enrollment API for the course details of the given course_id.
+
+        Args:
+            course_id (str): The string value of the course's unique identifier
+
+        Returns:
+            dict: A dictionary containing details about the course, in an enrollment context (allowed modes, etc.)
+        """
+        try:
+            return self.client.course(course_id).get()
+        except (SlumberBaseException, ConnectionError, Timeout) as exc:
+            LOGGER.exception(
+                'Failed to retrieve course enrollment details for course [%s] due to: [%s]',
+                course_id, str(exc)
+            )
+            return {}
+
+    def _sort_course_modes(self, modes):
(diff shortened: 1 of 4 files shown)


enterprise/api_client/lms.py
```diff
@@ -316,6 +316,198 @@ def get_enrolled_courses(self, username):
         return self.client.enrollment.get(user=username)
 
 
+class EnrollmentApiClientJwt(JwtLmsApiClient):
```
christopappas: Bit of a nitpick, but I think this class name would be clearer as `EnrollmentApiJwtClient`

christopappas: Only because I believe most of our client class names end in "client"

sameenfatima78: Thanks Chris for reviewing my PR. You're right but this class name is only set for a temporary period. We are eventually going to remove the original client and then rename the new client as the old one. I'm being a bit reluctant in changing the name because there are around 15-16 endpoints like this one, means 15 PRs would require this change. So, I guess we can go with this temporary name if it's eventually going to be replaced soon with the original name, right?

christopappas: I'm not sure I quite understand why there would need to be so many PRs. EnrollmentApiClientJwt is a new class name added in this PR, so I was thinking you could just change it here (other PRs off of this branch could pull the changes).

Anyway, it is possible I'm missing some context though, so please don't let me stop you. Ultimately I just thought a classname ending with `Jwt` sounded like a type of jwt object, and not a client object :)

sameenfatima78: Yes, noted. I'll change the classname :) 
**Context**: The goal is to gradually replace the original client in all endpoints one by one where it is being called. The reason is that some of the endpoints cannot be tested locally so we are picking one endpoint at a time and pushing it to merge. This way, if any issue comes up, it will be easier to track and revert the changes. Also, this would make the PRs concise. There are around 15-16 endpoints in total so, covering all of them in one PR can be a little risky (given that some of those endpoints cannot even be tested locally). I have got an approval on this from George too. Initially, I'm pushing only those endpoints that have been tested manually and are working fine. If no issue comes up in any of them after merge, we can start pushing the endpoints that can't be tested. Once the client is replaced completely, I'll remove the copy. Hope this answers your concerns and thanks for approving my PR. :)

PR #657, opened 2019-12-17
ENT-2042 Replaced Edx-Api-Key in remaining endpoints of CourseApiClient

sameenfatima78:
One endpoint of CourseApiClient has already been replaced with the new client. This PR covers the remaining endpoints of CourseApiClient.


enterprise/api_client/lms.py
```diff
@@ -341,6 +341,32 @@ def get_course_details(self, course_id):
             return None
 
 
+class CourseApiClientJwt(JwtLmsApiClient):
+    """
+    Object builds an API client to make calls to the Course API.
+    """
+
+    API_BASE_URL = settings.LMS_INTERNAL_ROOT_URL + '/api/courses/v1/'
```
HammadAhmadWaqas: nit: can we use `urljoin` here?

enterprise/admin/__init__.py
```diff
@@ -382,11 +382,11 @@ def get_readonly_fields(self, request, obj=None):
...
-    def get_enrolled_course_string(self, course_ids):
+    def get_enrolled_course_string(self, course_ids, enterprise_customer_user):
         """
         Get an HTML string representing the courses the user is enrolled in.
         """
-        courses_client = CourseApiClient()
+        courses_client = CourseApiClientJwt(enterprise_customer_user.user)
         course_details = []
         for course_id in course_ids:
             name = courses_client.get_course_details(course_id)['name']
```
irfanuddinahmad: get_course_details can return None also ... we should add a relevant test case and handle that scenario

hasnain-naveed: @sameenfatima78 `courses_client.get_course_details(course_id)['name']` will raise an exception if `get_course_details` function returns None. Is this intended behaviour?

sameenfatima78: Yes, in that case it doesn't throw any exception, it just converts `None` into string and shows that against the course details. However, I have handled this case now. Instead of None, it will now show "No course details found for {course_id}".

sameenfatima78: Yes, in that case it doesn't throw any exception, it just converts `None` into string and shows that against the course details. However, I have handled this case now. Instead of `None,` it will now show "No course details found for {course_id}".

PR #659, opened 2019-12-17
ENT-2042 Replaced Edx-Api-Key in EnterpriseCustomerCourseEnrollmentsSerializer

sameenfatima78:
This PR is about replacing the `EDX-API-KEY` with `OAuth` authentication method using `JWT.` There are three clients that are relying on the `EDX-API-KEY` based authentication; `CourseApiClient,` `EnrollmentApiClient,` and `ThirdPartyAuthApiClient.` This PR only covers the `enterprise/api/v1/serializers.py/EnterpriseCustomerCourseEnrollmentsSerializer` endpoint by cloning the original `ThirdPartyAuthApiClient` and using that one in the mentioned endpoint. It also covers all the test cases of the endpoint.

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -646,9 +646,14 @@ class EnterpriseCustomerCourseEnrollmentsSerializer(serializers.Serializer):
 
         It first uses the third party auth api to find the associated username to do the lookup.
         """
-        enterprise_customer = self.context.get('enterprise_customer')
-        request_user = self.context.get('request_user')
-
+        try:
+            enterprise_customer = self.context.get('enterprise_customer')
+            request_user = self.context.get('request_user')
+        except KeyError:
+            LOGGER.error(
+                'EnterpriseCustomer or Request User is not in the context of Serializer class'
+            )
+            return None
         try:
             tpa_client = ThirdPartyAuthApiClientJwt(request_user)
             username = tpa_client.get_username_from_remote_id(
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -16,7 +16,7 @@ from django.contrib.sites.models import Site
 from django.utils.translation import ugettext_lazy as _
 
 from enterprise import models, utils
-from enterprise.api_client.lms import ThirdPartyAuthApiClient
+from enterprise.api_client.lms import ThirdPartyAuthApiClientJwt
 from enterprise.constants import ENTERPRISE_PERMISSION_GROUPS
 from enterprise.utils import CourseEnrollmentDowngradeError, CourseEnrollmentPermissionError, track_enrollment
 
@@ -647,9 +647,10 @@ class EnterpriseCustomerCourseEnrollmentsSerializer(serializers.Serializer):
         It first uses the third party auth api to find the associated username to do the lookup.
         """
         enterprise_customer = self.context.get('enterprise_customer')
+        request_user = self.context.get('request_user')
 
         try:
-            tpa_client = ThirdPartyAuthApiClient()
+            tpa_client = ThirdPartyAuthApiClientJwt(request_user)
             username = tpa_client.get_username_from_remote_id(
                 enterprise_customer.identity_provider, value
             )
--- a/enterprise/api_client/lms.py
+++ b/enterprise/api_client/lms.py
@@ -14,7 +14,7 @@ from opaque_keys.edx.keys import CourseKey
 from requests import Session
 from requests.exceptions import ConnectionError, Timeout  # pylint: disable=redefined-builtin
 from slumber.exceptions import HttpNotFoundError, SlumberBaseException
-from six.moves.urllib.parse import urljoin
+from requests.compat import urljoin
 
 from django.conf import settings
 from django.utils import timezone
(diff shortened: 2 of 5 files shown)


enterprise/api_client/lms.py
```diff
@@ -398,6 +398,65 @@ def _get_results(self, identity_provider, param_name, param_value, result_field_
         return None
 
 
+class ThirdPartyAuthApiClientJwt(JwtLmsApiClient):
+    """
+    Object builds an API client to make calls to the Third Party Auth API.
+    """
+
+    API_BASE_URL = settings.LMS_INTERNAL_ROOT_URL + '/api/third_party_auth/v0/'
```
HammadAhmadWaqas: nit: better to use `urljoin`

enterprise/api_client/lms.py
```diff
@@ -398,6 +398,65 @@ def _get_results(self, identity_provider, param_name, param_value, result_field_
...
+        Retrieve the remote identifier for the given username.
+
+        Args:
+        * ``identity_provider`` (str): identifier slug for the third-party authentication service used during SSO.
+        * ``username`` (str): The username ID identifying the user for which to retrieve the remote name.
+
+        Returns:
+            string or None: the remote name of the given user.  None if not found.
+        """
+        return self._get_results(identity_provider, 'username', username, 'remote_id')
```
HammadAhmadWaqas: to make it more readable can we call like this
`self._get_results(identity_provider, 'remote_id', query_parms={'username': username} )`

enterprise/api_client/lms.py
```diff
@@ -398,6 +398,65 @@ def _get_results(self, identity_provider, param_name, param_value, result_field_
...
+        Args:
+        * ``identity_provider`` (str): identifier slug for the third-party authentication service used during SSO.
+        * ``remote_id`` (str): The remote id identifying the user for which to retrieve the usernamename.
+
+        Returns:
+            string or None: the username of the given user.  None if not found.
+        """
+        return self._get_results(identity_provider, 'remote_id', remote_id, 'username')
+
+    def _get_results(self, identity_provider, param_name, param_value, result_field_name):
```
HammadAhmadWaqas: accordingly 
`def _get_results(self, identity_provider, result_field_name, query_parms)`

enterprise/api_client/lms.py
```diff
@@ -398,6 +398,65 @@ def _get_results(self, identity_provider, param_name, param_value, result_field_
...
+        """
+        Calls the third party auth api endpoint to get the mapping between usernames and remote ids.
+        """
+        try:
+            kwargs = {param_name: param_value}
+            returned = self.client.providers(identity_provider).users.get(**kwargs)
```
HammadAhmadWaqas: `response = self.client.providers(identity_provider).users.get(**query_parms)`

enterprise/api/v1/serializers.py
```diff
@@ -647,9 +647,10 @@ def validate_tpa_user_id(self, value):
         It first uses the third party auth api to find the associated username to do the lookup.
         """
         enterprise_customer = self.context.get('enterprise_customer')
+        request_user = self.context.get('request_user')
 
         try:
-            tpa_client = ThirdPartyAuthApiClient()
+            tpa_client = ThirdPartyAuthApiClientJwt(request_user)
```
HammadAhmadWaqas: what will happen if we do not have `request_user` in context for EnterpriseCustomerCourseEnrollmentsSerializer?

sameenfatima78: I cross-checked it. The view that is using `EnterpriseCustomerCourseEnrollmentsSerializer` already passes `request_user` in context. See this for reference: https://github.com/edx/edx-enterprise/blob/master/enterprise/api/v1/views.py#L178

HammadAhmadWaqas: yes, We are passing for this instance, but what if in future someone instantiates `EnterpriseCustomerCourseEnrollmentsSerializer` object and do not pass `request_user`

sameenfatima78: I can pass `None` as a default value and then make the changes accordingly. Would that solve the problem? If you have any other solutions that you can propose, please do.

enterprise/api_client/lms.py
sameenfatima78: This seems to be a reasonable change but the `_get_results` method is being called at two different places, one passes 'remote_id' and its value as `param_name` and `param_value` and the other passes `username` and its value in the same arguments. We'll eventually have to declare these two variables(param_name+param_value) anyway inside `_get_results` to be used here:

https://github.com/edx/edx-enterprise/blob/master/enterprise/api_client/lms.py#L389
https://github.com/edx/edx-enterprise/blob/master/enterprise/api_client/lms.py#L396

HammadAhmadWaqas: yes, then you can go with your solution.

enterprise/api/v1/serializers.py
HammadAhmadWaqas: yes,
we can put these lines in try-catch.
```
enterprise_customer = self.context['enterprise_customer']
request_user = self.context['request_user']
```
and catch a KeyError expection then after logging error we can return None.

PR #664, opened 2019-12-23
ENT-2513: Disable manual enrollment orders for audit enterprise learners

irfanuddinahmad:
Description: Audit type enrollment transactions would no longer create $0 orders for financial reporting.

JIRA: https://openedx.atlassian.net/browse/ENT-2513

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/admin/views.py
+++ b/enterprise/admin/views.py
@@ -828,6 +828,7 @@ class EnterpriseCustomerManageLearnersView(View):
             notify: Whether to notify (by email) the users that have been enrolled
         """
         pending_messages = []
+        paid_modes = ['verified', 'professional']
 
         if course_id:
             succeeded, pending, failed = cls.enroll_users_in_course(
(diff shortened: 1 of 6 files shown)


enterprise/admin/views.py
```diff
@@ -828,6 +828,7 @@ def _enroll_users(
             notify: Whether to notify (by email) the users that have been enrolled
         """
         pending_messages = []
+        paid_modes = ['verified', 'professional']
```
muhammad-ammar: Can we use [paid_modes_for_course](https://github.com/edx/edx-platform/blob/08877db85836d9630ef39483a5bc8e294375af16/common/djangoapps/course_modes/models.py#L313) instead of hardcoded values here?

tests/test_models.py
```diff
@@ -426,7 +426,22 @@ def test_get_remote_id(self, provider_id, expected_value, called, mock_third_par
...
+        enrollment_api_client_mock.return_value.get_course_enrollment.return_value = None
+        enterprise_customer_user.enroll('course-v1:edX+DemoX+Demo_Course', 'verified')
+        enrollment_api_client_mock.return_value.enroll_user_in_course.assert_called_once()
+        analytics_mock.track.assert_called_once()
+        enrollment_order_mock.assert_called_once()
+
+    @mock.patch('enterprise.utils.segment')
+    @mock.patch('enterprise.models.EnrollmentApiClient')
+    @mock.patch('enterprise.models.EnterpriseCustomerUser.create_order_for_enrollment')
+    def test_audit_enroll_learner(self, enrollment_order_mock, enrollment_api_client_mock, analytics_mock):
```
muhammad-ammar: I think we can merge this and the above test into a single test because they are same.

enterprise/admin/views.py
irfanuddinahmad: Let's try not to create this coupling with platform.

tests/test_models.py
irfanuddinahmad: Done

PR #669, opened 2019-12-30
ENT-2475 Create Management command to create enterprise catalogs in new service

brittneyexline:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**


enterprise/management/commands/migrate_enterprise_catalogs.py
```diff
@@ -0,0 +1,56 @@
...
+    def handle(self, *args, **options):
+        api_username = options['api_user']
+        try:
+            user = User.objects.get(username=api_username)
+        except User.DoesNotExist:
+            raise CommandError(_('A user with the username {username} was not found.').format(username=api_username))
+
+        client = EnterpriseCatalogApiClient(user=user)
+        for enterprise_catalog in EnterpriseCustomerCatalog.objects.all():
+            LOGGER.info('Migrating Enterprise Catalog {}'.format(enterprise_catalog.uuid))
```
christopappas: nitpick about wording, but i would say "create" here too instead of "migrate" (which feels like this record would somehow be gone after the task is complete)

enterprise/settings/test.py
```diff
@@ -130,6 +130,7 @@ def root(*args):
 LMS_INTERNAL_ROOT_URL = "http://localhost:8000"
 LMS_ENROLLMENT_API_PATH = "/api/enrollment/v1/"
 ECOMMERCE_PUBLIC_URL_ROOT = "http://localhost:18130"
+ENTERPRISE_CATALOG_INTERNAL_ROOT_URL = "http://localhost:8000"
```
christopappas: does this not collide with LMS_INTERNAL_ROOT_URL?

tests/test_enterprise/api_client/test_enterprise_catalog.py
```diff
@@ -0,0 +1,122 @@
+# -*- coding: utf-8 -*-
+"""
+Tests for enterprise.api_client.lms.py
```
christopappas: doc string needs to be tweaked i believe

tests/test_enterprise/api_client/test_enterprise_catalog.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    Build a URL for the relevant API named by base_name.
+
+    Args:
+        base_name (str): A name to identify the root URL by
+        path (str): The URL suffix to append to the base path
+    """
+    return requests.compat.urljoin(enterprise_catalog.EnterpriseCatalogApiClient.API_BASE_URL, path)
+
+
```
christopappas: i'm curious about why we dont use the standard test class format here

brittneyexline: i copied this from another api_client test file, and i'm not sure why it was implemented this way before... 🤷‍♀

PR #671, opened 2020-01-02
added logging of discovery request size

chavesj:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/api_client/discovery.py
+++ b/enterprise/api_client/discovery.py
@@ -165,6 +165,10 @@ class CourseCatalogApiClient(object):
                     query_params,
                     traverse_pagination
                 )
+                LOGGER.info(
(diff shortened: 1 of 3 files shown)


enterprise/api_client/discovery.py
```diff
@@ -165,6 +165,10 @@ def get_catalog_results(self, content_filter_query, query_params=None, traverse_
                     query_params,
                     traverse_pagination
                 )
+                LOGGER.info(
+                    'ENT-2489 | Response has size %d',
```
brittneyexline: it might be good for us to see which `content_filter_query` is associated with the response, could you add it to this message?

chavesj: 👌

PR #674, opened 2020-01-07
WIP --- soft delete enterprise customer user to avoid deletion of related enterprise course enrollments

muhammad-ammar:
**Description:** ...
**JIRA:** [ENT-2112](https://openedx.atlassian.net/browse/ENT-2112)

**Merge checklist:**
- [ ] Translations updated

**Post merge:**


muhammad-ammar:
@ormsbee Needs your help to solve a problem. I have added the details below and made this PR to show what I am thinking. Can you please give it a look?

**Terminology**
`EnterpriseCustomer`: An Enterprise Customer
`EnterpriseCustomerUser`: Users belongs to an Enterprise -- Enterprise Learners
`EnterpriseCourseEnrollment`: Course Enrollments created for an EnterpriseCustomerUser --- these are separate records from LMS course enrollments.

**Problem**
Currently our enterprise system required that an `EnterpriseCustomerUser` can be linked to a single `EnterpriseCustomer` at any time. Sometimes due to how a particular `EnterpriseCustomer` is configured or for some other reason, it is desired that an `EnterpriseCustomerUser` should be linked to multiple enterprises but it is not possible due to limitation mentioned above . A workaround to this issue is to unlink an `EnterpriseCustomerUser` from an `EnterpriseCustomer` and `link` to some other `EnterpriseCustomer` and then `re-linked` back to original `EnterpriseCustomer`. The problem is that this `un-link` hard deletes the actual `EnterpriseCustomerUser` records and due to `CASCADE` the related `EnterpriseCourseEnrollment` also gets deleted. We don't want this. 

**Solution**
One solution to above problem is **soft delete** the `EnterpriseCustomerUser` records upon `unlink`. This can be done by a BooleanField named as `linked` and then `set/unset` it upon `link/unlink`.

**Implementation**
Custom model manager for `EnterpriseCustomerUser`. This manager will filter  the initial queryset and return `linked`  or `all` records based on how the manager is created.

On `EnterpriseCustomerUser` model, create 2 instances of manager, one for `linked=True` and second for `linked=False` records like below

    objects = EnterpriseCustomerUserManager()
    all_objects = EnterpriseCustomerUserManager(linked_only=False)

For create calls like  `get_or_create/update_or_create/create`calls use the `all_objects` manger and for others purposes use `objects` manager like below

    EnterpriseCustomerUser.all_objects.update_or_create(
        user_id=user_id, 
        defaults={linked=True}
    )

and 

    EnterpriseCustomerUser.objects.get(
        user_id=user_id
    )

**Concerns**
 - This above approach needs that we update all the code places where `EnterpriseCustomerUser` model is used
 - In future how a developer who is going to make changes will know what to use when? Which manager (`all_objects` or `objects`) one should use and what is appropriate way to use one?  

**Questions**
 - Can we improve this approach in such a way that we don't need to touch every code place and move the complexity and main logic inside Mode/Manager/Queryset?
 - Any other approach that can help solve the Problem?

ormsbee:
@muhammad-ammar: I'm not sure if you meant to tag me or if you were thinking of someone else–I've never really done any work on Enterprise, and I have a poor understanding of your business requirements. I'm going to be asking some dumb questions to get my bearings...

> Currently our enterprise system required that an EnterpriseCustomerUser can be linked to a single EnterpriseCustomer at any time. Sometimes due to how a particular EnterpriseCustomer is configured or for some other reason, it is desired that an EnterpriseCustomerUser should be linked to multiple enterprises but it is not possible due to limitation mentioned above.

* What are those situations? 
* Why is there a 1:1 mapping today?
* Do we really want to think of the user as having entirely switched EnterpriseCustomers, or is there another kind of M:M relationship model that we're missing?
* Is there a valid use case for a user to actively belong to multiple EnterpriseCustomers at once?

> A workaround to this issue is to unlink an EnterpriseCustomerUser from an EnterpriseCustomer and link to some other EnterpriseCustomer and then re-linked back to original EnterpriseCustomer. The problem is that this un-link hard deletes the actual EnterpriseCustomerUser records and due to CASCADE the related EnterpriseCourseEnrollment also gets deleted. We don't want this.

I suspect that we want some kind of soft delete functionality anyway, even for simple deactivation/reactivation, irrespective of whether we're switching EnterpriseCustomer values. But I'm concerned if we're using a soft-delete as a way to stash away changes in order to temporarily put the user in another EnterpriseCustomer for the purposes of some operation, only to switch back later. I feel like that could introduce a lot of confusion to the data model and be vulnerable to getting into broken states–e.g. if the code assumes that there will only be one active EnterpriseCustomerUser entry per user but that's not enforced via database layer constraint, and a race condition somehow gets you into a situation with two active entries.

Anyhow, those are my main concerns. I don't think I know enough about the use cases you're solving to make any strong recommendations.

Thank you.

muhammad-ammar:
@ormsbee Thanks a lot for the detailed reply. There is already work in progress to handle the multiple enterprises functionality and that will handle all the cases. For now I need your help only regarding the implementation approach I outlined above.

PR #678, opened 2020-01-15
ENT-2464: Added management command to reset SAPSF completion data

ziafazal:
Added management command to reset SAPSF completion data.
It would allow us resend those completion enteries which are reset.
ENT-2464

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/management/commands/reset_sapsf_learner_transmissions.py
+++ b/integrated_channels/integrated_channel/management/commands/reset_sapsf_learner_transmissions.py
@@ -1,0 +1,71 @@
+# -*- coding: utf-8 -*-
+"""
+Reset SAPSF learner transmissions between two dates.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from django.apps import apps
+from django.core.management.base import BaseCommand
+from django.utils.dateparse import parse_datetime
+from django.utils.translation import ugettext as _
+
+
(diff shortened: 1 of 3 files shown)


integrated_channels/integrated_channel/management/commands/reset_sapsf_learner_transmissions.py
```diff
@@ -0,0 +1,70 @@
...
+
+        SapSuccessFactorsLearnerDataTransmissionAudit = apps.get_model(  # pylint: disable=invalid-name
+            'sap_success_factors',
+            'SapSuccessFactorsLearnerDataTransmissionAudit'
+        )
+        enrollment_ids = SapSuccessFactorsLearnerDataTransmissionAudit.objects.filter(
+            created__gte=start_datetime, created__lte=end_datetime
+        ).values_list('enterprise_course_enrollment_id', flat=True)
+
+        for enrollment_id in enrollment_ids:
```
brittneyexline: curious how many of these records we expect to update, if there are more than a few thousand then we may want to chunk this.

ziafazal: This command is suppose to update around 13K records in single table. It is going to pull enrollment_ids of records which needs to be updated and then updates the record one by one.

PR #679, opened 2020-01-15
Sync catalog admin to enterprise-catalog ida

bbaker6225:
**Merge checklist:**
- [x] Translations updated

https://openedx.atlassian.net/browse/ENT-2476

**Post merge:**

### DIFF
--- a/enterprise/admin/__init__.py
+++ b/enterprise/admin/__init__.py
@@ -38,6 +38,7 @@ from enterprise.admin.views import (
     EnterpriseCustomerTransmitCoursesView,
     TemplatePreviewView,
 )
+from enterprise.api_client.enterprise_catalog import EnterpriseCatalogApiClient
 from enterprise.api_client.lms import CourseApiClient, EnrollmentApiClient
 from enterprise.models import (
     EnrollmentNotificationEmailTemplate,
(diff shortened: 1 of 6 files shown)


enterprise/admin/__init__.py
```diff
@@ -633,6 +638,67 @@ def get_form(self, request, obj=None, **kwargs):
...
+        """
+        On save, creates or updates the corresponding catalog in the enterprise-catalog IDA.
+
+        `change` indicates whether or not the object is being updated (as opposed to created).
+        """
+        # Save the changes no matter what as we can copy them over later if we need to
+        super(EnterpriseCustomerCatalogAdmin, self).save_model(request, obj, form, change)
+
+        catalog_uuid = obj.uuid
+        catalog_client = EnterpriseCatalogApiClient()
```
bbaker6225: Do we want the user for the client to be `request.user` in this case?

tox.ini
```diff
@@ -70,7 +70,7 @@ commands =
 deps =
     -r{toxinidir}/requirements/dev.txt
 commands =
-    isort --check-only --recursive tests test_utils enterprise enterprise_learner_portal consent integrated_channels manage.py setup.py
+    isort --check-only --diff --recursive tests test_utils enterprise enterprise_learner_portal consent integrated_channels manage.py setup.py
```
bbaker6225: Just a little convenience so you know _why_ your isort failed

enterprise/api_client/enterprise_catalog.py
```diff
@@ -43,22 +46,57 @@ def create_enterprise_catalog(
...
         }
-        return endpoint.post(post_data)
+        try:
+            return endpoint.post(post_data)
+        except (SlumberBaseException, ConnectionError, Timeout) as exc:
+            LOGGER.exception(
+                'Failed to create EnterpriseCustomer Catalog [%s] in enterprise-catalog due to: [%s]',
+                catalog_uuid, str(exc)
+            )
+            return False
```
brittneyexline: i guess my personal preference is to raise here anyway to force callers to deal with error scenarios, but i don't have a strong opinion on this.

enterprise/admin/__init__.py
```diff
@@ -633,6 +634,59 @@ def get_form(self, request, obj=None, **kwargs):
...
+        catalog_uuid = obj.uuid
+        catalog_client = EnterpriseCatalogApiClient(user=request.user)
+
+        if change:
+            update_fields = {
+                'enterprise_customer': str(obj.enterprise_customer.uuid),
+                'title': obj.title,
+                'content_filter': obj.content_filter,
+                'enabled_course_modes': obj.enabled_course_modes,
+                'publish_audit_enrollment_urls': obj.publish_audit_enrollment_urls,
```
brittneyexline: something i changed because it appeared to break things and seemed to work was to convert this from a boolean to a string, i got a bad request exception when just passing it directly. if you haven't run into that in your testing, then that's cool, just calling it out.

bbaker6225:
I had a hard time finding a good pattern for unit testing this sort of thing in the admin, but I'll look some more and see if I can come up with something

enterprise/admin/__init__.py
bbaker6225: Huh interesting. I haven't seen that, but I'll see if I can replicate

enterprise/api_client/enterprise_catalog.py
bbaker6225: Yeah I don't love just silently failing and only having the logs, but it also felt weird to me for them to hit save, get an error, and then have their changes actually take effect. Open to others' opinions on this

enterprise/admin/__init__.py
```diff
@@ -633,6 +634,59 @@ def get_form(self, request, obj=None, **kwargs):
...
+            )
+
+    def delete_model(self, request, obj):
+        """
+        Deletes the corresponding catalog in the enterprise-catalog IDA
+        """
+        catalog_uuid = obj.uuid
+        catalog_client = EnterpriseCatalogApiClient(user=request.user)
+        response = catalog_client.delete_enterprise_catalog(catalog_uuid)
+        # Early exit so we don't delete the catalog if the DELETE fails in enterprise-catalog
```
bbaker6225: A warning I just saw in the django docs says: 

> When overriding ModelAdmin.save_model() and ModelAdmin.delete_model(), your code must save/delete the object. They aren’t meant for veto purposes, rather they allow you to perform extra operations.

so it looks like we can't really guarantee the delete will be prevented. I don't know how often people actually delete enterprise catalogs to know if we're particularly concerned or if we're ok just checking the logs periodically and cleaning up any catalogs that don't get cleared out of the new service

bbaker6225: Based off Rob's answer on how often enterprise catalogs are deleted: 

> I know there have been times when they’ve been deleted, but as a team, we’ve made the determination that it is better not to delete catalogs

I feel comfortable allowing the deletion and then just handling any clean up afterwards since it should rarely (if ever) happen. @brittneyexline let me know if you disagree

PR #683, opened 2020-01-20
soft delete enterprisecustomeruser records

muhammad-ammar:
**Description:** Currently whenever we unlink an Enterprise Learner from an Enterprise, all the Enterprise Course Enrollments for the learner gets deleted due to `CASCADING`. To avoid this we have added soft deletion for `EnterpriseCustomerUser` records. The behaviour will remain same and get queryset methods will not return a soft deleted `EnterpriseCustomerUser` record and its related `EnterpriseCourseEnrollment` records but all the creation methods like `get_or_create/update_or_create/create` will always return an `EnterpriseCustomerUser` record either by creating a new object in database or by updating an existing soft-deleted object.

**JIRA:** https://openedx.atlassian.net/browse/ENT-2538

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -538,13 +538,32 @@ class EnterpriseCustomerUserManager(models.Manager):
     This class should contain methods that create, modify or query :class:`.EnterpriseCustomerUser` entities.
     """
 
+    def __init__(self, *args, **kwargs):
+        """
+        Initialize custom manager.
+
+        kwargs:
+            linked_only (Bool): create a manager with linked learners only if True else all(linked and unlinked) records
+        """
+        self.linked_only = kwargs.pop('linked_only', True)
+        super(EnterpriseCustomerUserManager, self).__init__(*args, **kwargs)
+
+    def get_queryset(self):
+        """
+        Return linked or unlinked learners based on how the manager is created.
+        """
+        if self.linked_only:
+            return super(EnterpriseCustomerUserManager, self).get_queryset().filter(linked=True)
+
+        return super(EnterpriseCustomerUserManager, self).get_queryset()
+
     def get(self, **kwargs):
         """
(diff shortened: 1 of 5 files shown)


muhammad-ammar:
@brittneyexline Can you please review the changes?

enterprise/models.py
```diff
@@ -532,13 +532,29 @@ class EnterpriseCustomerUserManager(models.Manager):
     This class should contain methods that create, modify or query :class:`.EnterpriseCustomerUser` entities.
     """
 
+    def __init__(self, *args, **kwargs):
+        """
+        Initialize custom manager.
+        """
+        self.linked_only = kwargs.pop('linked_only', True)
```
georgebabey: nit: Might be helpful to add the kwargs to the docstring

enterprise/models.py
```diff
@@ -654,6 +672,33 @@ class Meta(object):
         unique_together = (("enterprise_customer", "user_id"),)
         ordering = ['-active', '-modified']
 
+    def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
+        """
+        Override to handle creation of EnterpriseCustomerUser records.
+
+        This is needed becuase of soft deletion of EnterpriseCustomerUser records.
```
georgebabey: typo: `because`

enterprise/models.py
```diff
@@ -654,6 +672,33 @@ class Meta(object):
...
+        This will handle all of get_or_create/update_or_create/create methods.
+        """
+        if self.pk is None:
+            # We are trying to create a new object but it may be possible that an existing unlinked
+            # record exists, so if an existing record exists then just update that with linked=True
+            try:
+                existing = EnterpriseCustomerUser.all_objects.get(
```
georgebabey: I like this approach so to updating a previously soft-deleted record 👍

PR #684, opened 2020-01-21
ENT-2506: Added ability to send learner completion if grade is changed

ziafazal:
This PR has following changes

- Added ability to send learner completion if grade is changed
- Added index on `user_id` field of `EnterpriseCustomerUser` model
- Single learner completion should be sent for only current active enterprise of learner

**Merge checklist:**
- [ ] Translations updated

**Post merge:**


enterprise/models.py
```diff
@@ -641,7 +641,7 @@ class EnterpriseCustomerUser(TimeStampedModel):
         related_name='enterprise_customer_users',
         on_delete=models.deletion.CASCADE
     )
-    user_id = models.PositiveIntegerField(null=False, blank=False)
+    user_id = models.PositiveIntegerField(null=False, blank=False, db_index=True)
```
brittneyexline: didn't realize this wasn't indexed 😬

PR #685, opened 2020-01-21
initialize enrollmentapi client with service user

muhammad-ammar:
**Description:** For enroll/unenroll, initialize the `EnrollmentApiClient` with servier user. 

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/tests/test_admin/test_view.py
+++ b/tests/test_admin/test_view.py
@@ -157,7 +157,7 @@ class BaseTestEnterpriseCustomerManageLearnersView(TestCase):
         """
         super(BaseTestEnterpriseCustomerManageLearnersView, self).setUp()
         self.user = UserFactory.create(is_staff=True, is_active=True, id=1)
-        self.worker_user = UserFactory.create(
+        self.worker_user = UserFactory(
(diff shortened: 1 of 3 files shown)


tests/test_admin/test_view.py
```diff
@@ -158,7 +158,7 @@ def setUp(self):
         super(BaseTestEnterpriseCustomerManageLearnersView, self).setUp()
         self.user = UserFactory.create(is_staff=True, is_active=True, id=1)
         self.worker_user = UserFactory.create(
-            username=settings.ENTERPRISE_SERVICE_WORKER_USERNAME, is_staff=True, is_active=True
+            username=settings.ENTERPRISE_SERVICE_WORKER_USERNAME, is_staff=True, is_active=True, id=1000
```
ziafazal: why we need id?

muhammad-ammar: Tests are failing randomly in `test_view.py` due to `IntegrityErrors`, As per my understanding this is happening due to manually specifying `id` in lot `UserFactory` creation instances.

ziafazal: yeah but if we have 999 users already?

PR #686, opened 2020-01-21
ENT-871 Added total hours field for SuccessFactors export

irfanuddinahmad:
JIRA: https://openedx.atlassian.net/browse/ENT-871

**Merge checklist:**
- [ ] Translations updated

**Post merge:**


integrated_channels/sap_success_factors/exporters/content_metadata.py
```diff
@@ -109,6 +111,24 @@ def transform_description(self, content_metadata_item):
 
         return description_with_locales
 
+    def transform_estimated_hours(self, content_metadata_item):
```
ziafazal: I think we should move this method in the parent class and inherit it from there for SAP and Cornerstone.

tests/test_integrated_channels/test_sap_success_factors/test_exporters/test_content_metadata.py
```diff
@@ -423,3 +425,106 @@ def test_transform_course_description(self, course, expected_description):
                 'value': expected_description
             }
         ]
+
+    def _merge_dicts(self, dict1, dict2):
```
ziafazal: this can be move in `test_utils.py` and used it from there here in SAPSF and Cornerstone.

integrated_channels/sap_success_factors/exporters/content_metadata.py
brittneyexline: looks like it is already in the parent class? some of the functionality is duplicated so we could probably call the parent class to do the calculation bit that is in common but keep this one here since part of it seems specific to SAP

tests/test_integrated_channels/test_sap_success_factors/test_exporters/test_content_metadata.py
irfanuddinahmad: Done

integrated_channels/sap_success_factors/exporters/content_metadata.py
irfanuddinahmad: Agree with @brittneyexline . I have moved out the duplicated code for transforming the course duration.

PR #691, opened 2020-01-24
ENT-2475 | Updating enterprise catalog migration command to accept in…

christopappas:
…dividual catalog uuids

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/management/commands/migrate_enterprise_catalogs.py
+++ b/enterprise/management/commands/migrate_enterprise_catalogs.py
@@ -31,10 +31,10 @@ class Command(BaseCommand):
             help=_('Username of a user authorized to access the Enterprise Catalog API.'),
         )
         parser.add_argument(
-            '--catalog_uuid',
-            dest='catalog_uuid',
-            metavar='ENT_CATALOG_UUID',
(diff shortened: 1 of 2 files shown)


enterprise/management/commands/migrate_enterprise_catalogs.py
```diff
@@ -40,7 +46,14 @@ def handle(self, *args, **options):
             raise CommandError(_('A user with the username {username} was not found.').format(username=api_username))
 
         client = EnterpriseCatalogApiClient(user=user)
-        for enterprise_catalog in EnterpriseCustomerCatalog.objects.all():
+
+        catalog_uuid = options.get('catalog_uuid')
+        if catalog_uuid:
+            queryset = EnterpriseCustomerCatalog.objects.filter(uuid=catalog_uuid)
```
brittneyexline: should this support one uuid or potentially a list?

christopappas: I assumed if they were running it on the shell they could do the for-loop in bash, but happy to update this. should be easy

PR #696, opened 2020-01-28
ENT-2552 | Adding branding fields to models to support enterprise lea…

christopappas:
…rner portal

**Merge checklist:**
- [ ] Translations updated

**Post merge:**


enterprise/migrations/0086_auto_20200128_1347.py
```diff
@@ -0,0 +1,36 @@
...
+
+    dependencies = [
+        ('enterprise', '0085_enterprisecustomeruser_linked'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='enterprisecustomer',
+            name='contact_email',
+            field=models.EmailField(blank=True, help_text='Email for intended use on learner portal Need Help section.', max_length=254, null=True),
```
adamstankiewicz: I'd recommend changing the help text to be more generic and not tied to a specific UI feature. Perhaps, "Email for the Learning Coordinator that learners can reach out to with questions." That way, if/when the use of this field in the learner portal changes, we won't need to update the help text.

enterprise/models.py
```diff
@@ -1018,6 +1024,18 @@ class EnterpriseCustomerBrandingConfiguration(TimeStampedModel):
         null=True, blank=True, max_length=255,
         validators=[validate_image_extension, validate_image_size]
     )
+    banner_border_color = models.CharField(
+        null=True,
+        blank=True,
+        max_length=7,
+        validators=[validate_hex_color],
```
adamstankiewicz: These model fields in `portal-designer` specify a default for the colors (i.e., #FFFFFF). However, I think it makes more sense to not have default colors and leave it up to the frontend to handle the empty case 👍

enterprise/models.py
```diff
@@ -270,6 +270,12 @@ class Meta:
         help_text=_("Hostname of the enterprise learner portal, e.g. bestrun.edx.org.")
     )
 
+    contact_email = models.EmailField(
+        null=True,
+        blank=True,
+        help_text=_("Email for intended use on learner portal Need Help section.")
```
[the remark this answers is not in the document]
christopappas: fair enough, will do

PR #698, opened 2020-01-30
ENT-2482: Restrict PendingEnterpriseCustomerUser for multiple EnterpriseCustomers

sameenfatima78:
This PR restricts a PendingEnterpriseCustomerUser to be linked with multiple enterprises at the same time. The user will be able to link with another enterprise once he/she registers or the other enterprise unlinks that user. 

Ticket: https://openedx.atlassian.net/browse/ENT-2482

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/admin/views.py
+++ b/enterprise/admin/views.py
@@ -305,7 +305,7 @@ class EnterpriseCustomerManageLearnersView(View):
         return queryset
 
     @classmethod
-    def _handle_singular(cls, enterprise_customer, manage_learners_form):
+    def _handle_singular(cls, request, enterprise_customer, manage_learners_form):
         """
         Link single user by email or username.
 
@@ -316,10 +316,21 @@ class EnterpriseCustomerManageLearnersView(View):
         form_field_value = manage_learners_form.cleaned_data[ManageLearnersForm.Fields.EMAIL_OR_USERNAME]
         email = email_or_username__to__email(form_field_value)
         try:
-            validate_email_to_link(email, form_field_value, ValidationMessages.INVALID_EMAIL_OR_USERNAME, True)
+            existing_record = validate_email_to_link(email, form_field_value, ValidationMessages.
+                                                     INVALID_EMAIL_OR_USERNAME, True)
         except ValidationError as exc:
             manage_learners_form.add_error(ManageLearnersForm.Fields.EMAIL_OR_USERNAME, exc)
         else:
+            if isinstance(existing_record, PendingEnterpriseCustomerUser) and existing_record.enterprise_customer \
+                    != enterprise_customer:
+                messages.warning(
+                    request,
+                    ValidationMessages.PENDING_USER_ALREADY_LINKED.format(
+                        user_email=email,
+                        ec_name=existing_record.enterprise_customer.name
+                    )
+                )
+                return None
             EnterpriseCustomerUser.objects.link_user(enterprise_customer, email)
             return [email]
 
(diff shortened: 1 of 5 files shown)


enterprise/admin/views.py
```diff
@@ -315,13 +315,29 @@ def _handle_singular(cls, enterprise_customer, manage_learners_form):
...
         try:
-            validate_email_to_link(email, form_field_value, ValidationMessages.INVALID_EMAIL_OR_USERNAME, True)
+            existing_record = validate_email_to_link(email, form_field_value, ValidationMessages.
+                                                     INVALID_EMAIL_OR_USERNAME, True)
         except ValidationError as exc:
             manage_learners_form.add_error(ManageLearnersForm.Fields.EMAIL_OR_USERNAME, exc)
-        else:
-            EnterpriseCustomerUser.objects.link_user(enterprise_customer, email)
-            return [email]
+        if existing_record:
```
moconnell1453: The else: portion of the try-except should not be removed. Otherwise, if a ValidationError is raised, we will still try to call link_user() with the invalid email address.

enterprise/admin/views.py
```diff
@@ -315,13 +315,29 @@ def _handle_singular(cls, enterprise_customer, manage_learners_form):
...
+            if isinstance(existing_record, PendingEnterpriseCustomerUser) and existing_record.enterprise_customer != \
+                    enterprise_customer:
+                messages.warning(
```
moconnell1453: This text should be stored as a constant, allowing it to be reused and localized as necessary.
Preferably, this could be added to the list of ValidationMessages in utils.py. We can move the enterprise customer check into validate_email_to_link and raise a ValidationError if necessary. This will also help in the _handle_bulk_upload() method, since it will also need to validate the list of email addresses for all users.

sameenfatima78: Hey Mike, thank you for reviewing my code. Doing so would affect the functionality of bulk upload. Currently, `_handle_bulk_upload()` already takes care of the case of users trying to register with multiple enterprises. (See [this](https://github.com/edx/edx-enterprise/blob/master/enterprise/admin/views.py#L387) for reference).  So if we raise an exception from `validate_email_to_link,` it wouldn't allow the form to process and go any further. This would keep the other users from enrolling as well. According to the current functionality of bulk upload, it goes through all the emails, enrolls the valid emails and for the emails that could not be added, it shows individual errors at the top of the page.  Another slight issue that might occur with this approach is that by raising the exception, we will have to add all the errors to the form and they will be shown at the top of the form (not at the top of the page) so in case if the list of linked learners is long, the admin will have to scroll down to the form to see the error.  Also, the utility `validate_email_to_link` gets called during form validation as well. So I'm trying to keep it generic and simple for other callers.

I have, however, incorporated the rest of your feedback. Let me know if you have any questions.

enterprise/admin/views.py
sameenfatima78: Good catch. Thank you for pointing this out!

enterprise/admin/views.py
sameenfatima78: I also did some testing and found out that the `ignore_existing` flag is always set to `True` wherever `validate_email_to_link` is being called... so an exception on form is only raised if we provide an invalid email.

moconnell1453: Sameen, thank you for explaining more about how the bulk upload process works. As long as it is handling pending learners already, then we can skip the changes I suggested.

PR #712, opened 2020-02-18
Removing use of BearerAuthentication + clean up

jinder1s:
According to our investigation using new relic custom metrics, BearerAuthentication class is not used and is safe for removal.

[PR that adds metric](https://github.com/edx/edx-drf-extensions/pull/105)

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -159,7 +159,8 @@ def get_oauth2authentication_class():
     try:
         from openedx.core.lib.api.authentication import OAuth2AuthenticationAllowInactiveUser as OAuth2Authentication
     except ImportError:
-        from edx_rest_framework_extensions.auth.bearer.authentication import BearerAuthentication \
+        #Temporary until OAuth2Authentication has been moved out of edx-platform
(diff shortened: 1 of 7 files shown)


enterprise/utils.py
```diff
@@ -177,7 +177,8 @@ def get_oauth2authentication_class():
     try:
         from openedx.core.lib.api.authentication import OAuth2AuthenticationAllowInactiveUser as OAuth2Authentication
     except ImportError:
-        from edx_rest_framework_extensions.auth.bearer.authentication import BearerAuthentication \
+        # Temporary until OAuth2Authentication has been moved out of edx-platform
+        from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication \
```
jinder1s: For reviewer, this seemed like the easiest short term solution to removing BearerAuthentication in this case.

feanil: Within edx-enterprise this is only called once for cornerstone it looks like.  Are there other libraries that are also calling this function?

jinder1s:
@feanil do you think this is good to go?

PR #716, opened 2020-02-26
Make the enrollment reason field optional

moconnell1453:
The enrollment reason field should not be required if just creating a new
learner without enrolling them. However, if a course or program is specified,
the enrollment reason will still be required.

ENT-2568


enterprise/admin/forms.py
```diff
@@ -246,6 +246,12 @@ def clean_program(self):
 
         return program
 
+    def clean_reason(self):
```
ziafazal: nit: we can simplify this method
```
def clean_reason(self):
    return self.cleaned_data.get(self.Fields.REASON).strip()
```

enterprise/admin/forms.py
```diff
@@ -331,6 +338,19 @@ def _validate_program(self):
                 mode=course_mode, program_title=program.get("title"), modes=", ".join(available_modes)
             ))
 
+    def _validate_reason(self):
+        """
+        Verify that the reason field is populated if the new learner(s) is being enrolled
+        in a course or program.
+        """
+        program = self.cleaned_data.get(self.Fields.PROGRAM)
```
hasnain-naveed: NIT: Can you please remove the program field check from here? My current PR will remove the Program UUID field from the form.

tests/test_admin/test_forms.py
```diff
@@ -420,6 +420,29 @@ def test_clean_discount(self, discount, is_valid, is_decimal_error):
...
+        assert cleaned_data[ManageLearnersForm.Fields.REASON] == expected_reason
+
+    def test_validate_reason(self):
+        course_id = "course-v1:edX+DemoX+Demo_Course"
+        reason = ""
+        form = self._make_bound_form("irrelevant@example.com", course=course_id, reason=reason)
+        assert not form.is_valid()
+        assert form.errors == {
+            "__all__": [ValidationMessages.MISSING_REASON]
+        }
```
HammadAhmadWaqas: @moconnell1453 
I think this test will not pass due to two reasons.
1. We need to mock `EnrollmentApiClient` as `clean_course` method is using this client to get details.
2. We also need to pass `course_mode` otherwise the form will raise validation error for course_mode too.
so this test will look like this.
```
    @mock.patch("enterprise.admin.forms.EnrollmentApiClient")
    def test_validate_reason(self, enrollment_client):
        instance = enrollment_client.return_value
        instance.get_course_details.side_effect = fake_enrollment_api.get_course_details
        course_id = "course-v1:edX+DemoX+Demo_Course"
        reason = ""
        form = self._make_bound_form("irrelevant@example.com", course=course_id, reason=reason, course_mode="audit")
        assert not form.is_valid()
        assert form.errors == {
            "__all__": [ValidationMessages.MISSING_REASON]
        }
```

moconnell1453: Ahmad, thank you for the feedback. I have been having some problems running unit tests locally. I will need to resolve those issues before I can effectively write and correct unit tests for edx-enterprise.

PR #718, opened 2020-02-26
Fixing tests

christopappas:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**


christopappas:
Turned out this was a mocking issue. From the docs: 
```
When you nest patch decorators the mocks are passed in to the decorated function
in the same order they applied (the normal Python order that decorators are applied).
This means from the bottom up, so in [our case the mock for transmit_content_metadata]
is passed in first.
```
So in our case:

```
          ...
patch2 -> @mock.patch('integrated_channels.sap_success_factors.client.SAPSuccessFactorsAPIClient.update_content_metadata')
patch1 -> @mock.patch('integrated_channels.integrated_channel.management.commands.transmit_content_metadata.transmit_content_metadata.delay')
          def test_transmit_content_metadata_task_with_error(
              self,
mockobj1 ->   transmit_content_metadata_mock,
mockobj2 ->   sapsf_update_content_metadata_mock,
              ...
```

PR #720, opened 2020-02-28
ENT-2658 | Added the salesforce opportunity_id in manage learner djan…

hasnain-naveed:
…go admin.

**Merge checklist:**
- [ ] Translations updated

**Post merge:**


enterprise/admin/views.py
```diff
@@ -828,10 +834,8 @@ def post(self, request, customer_uuid):
             notification_type = manage_learners_form.cleaned_data.get(ManageLearnersForm.Fields.NOTIFY)
             notify = notification_type == ManageLearnersForm.NotificationTypes.BY_EMAIL
             discount = manage_learners_form.cleaned_data.get(ManageLearnersForm.Fields.DISCOUNT)
-
-            course_id = None
-            if course_details:
-                course_id = course_details['course_id']
+            sales_force_id = manage_learners_form.cleaned_data.get(ManageLearnersForm.Fields.SALES_FORCE_ID)
+            course_id = course_details['course_id'] if course_details else None
```
moconnell1453: Good optimization.

enterprise/migrations/0091_auto_20200311_0808.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
```
moconnell1453: Should we rename this migration file to reflect its purpose, rather than use the auto-generated name?

enterprise/models.py
```diff
@@ -905,6 +907,7 @@ def create_order_for_enrollment(self, course_run_id, discount_percentage):
                     "enterprise_customer_name": self.enterprise_customer.name,
                     "enterprise_customer_uuid": str(self.enterprise_customer.uuid),
                     "discount_percentage": float(discount_percentage),
+                    "sales_force_id": sales_force_id,
```
moconnell1453: We're adding sales_force_id to the enrollment data, but will that field be persisted? I see that we have a discussion about this in the ticket. Do we need to have a separate ticket to update ecommerce to persist the sales_force_id when creating the manual order?

PR #723, opened 2020-03-10
Fix JSON field formatting in Django Admin for ECS

bbaker6225:
ENT-2677

**Merge checklist:**
- [x] Translations updated

**Post merge:**


enterprise/models.py
```diff
@@ -1352,7 +1353,7 @@ class EnterpriseCatalogQuery(TimeStampedModel):
         blank=True,
         null=True,
         load_kwargs={'object_pairs_hook': collections.OrderedDict},
-        dump_kwargs={'indent': 4},
+        dump_kwargs={'indent': 4, 'cls': JSONEncoder, 'separators': (',', ':')},
```
bbaker6225: Added the `cls` and `separators` that were part of the default `dump_kwargs` included in prior migration files

PR #725, opened 2020-03-12
ENT-2622: Reset user authentication cookies

ziafazal:
Changes to reset user authentication cookies after successfull selection of enterprise by learner. This should update JWT cookies which would help us accurately map learner's enterprise where we fetch learner's enterprise from JWT cookie.

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -74,6 +74,12 @@ try:
 except ImportError:
     ProgramDataExtender = None
 
+try:
+    from openedx.core.djangoapps.user_authn import cookies as user_authn_cookies
+except ImportError:
+    user_authn_cookies = None
+
+
 LOGGER = getLogger(__name__)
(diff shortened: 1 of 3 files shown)


enterprise/views.py
```diff
@@ -909,7 +915,11 @@ def form_valid(self, form):
                 self.request.user.username,
                 enterprise_customer,
             )
-            return JsonResponse({})
+            response = JsonResponse({})
+            return user_authn_cookies.set_logged_in_cookies(
```
irfanuddinahmad: … I think we also need to handle the cases where the multiple enterprise selection page is not displayed … e.g. SAML based implicit selection of enterprise

ziafazal: in case of SSO flow it is already handled [here](https://github.com/edx/edx-platform/blob/26891841e4063e7878311536c25182d0f705eb7c/common/djangoapps/third_party_auth/pipeline.py#L695)

PR #730, opened 2020-03-24
ENT-2572: WaffleFlag/Config for transitioning enterprise endpoints to use enterprise_catalog

chavesj:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**


enterprise/api_client/enterprise.py
```diff
@@ -44,12 +47,16 @@ def get_content_metadata(self, enterprise_customer, enterprise_catalogs=None):
...
         for enterprise_customer_catalog in enterprise_customer_catalogs:
-            response = self._load_data(
-                self.ENTERPRISE_CUSTOMER_CATALOGS_ENDPOINT,
-                resource_id=str(enterprise_customer_catalog.uuid),
-                traverse_pagination=True,
-                querystring={'page_size': 1000},
-            )
+            if switch_is_active('ENTERPRISE_CATALOG_API_ENDPOINTS_ENABLED'):
+                response = \
+                    EnterpriseCatalogApiClient.enterprise_catalog_get_content_metadata(enterprise_customer_catalog.uuid)
```
muhammad-ammar: I think we need create `EnterpriseCatalogApiClient` object like `EnterpriseCatalogApiClient(user=self.user)`

enterprise/api_client/enterprise_catalog.py
```diff
@@ -99,3 +100,16 @@ def delete_enterprise_catalog(self, catalog_uuid):
                 catalog_uuid, str(exc)
             )
             return {}
+
+    @JwtLmsApiClient.refresh_token
+    def enterprise_catalog_get_content_metadata(self, catalog_uuid):
+        """Returns all the content linked to the specified catalog, ordered by content key."""
+        endpoint = getattr(self.client, self.GET_CONTENT_METADATA_ENDPOINT)(catalog_uuid)
```
muhammad-ammar: `self.GET_CONTENT_METADATA_ENDPOINT` is missing a format string.

muhammad-ammar:
@chavesj To test the changes in client, there are 2 ways
1. Execute the [management command](https://github.com/edx/edx-enterprise/blob/69ccd4007bf68f4dd9e2a325c548084d4ac439ae/integrated_channels/integrated_channel/management/commands/transmit_content_metadata.py#L19) directly, or
2. In django admin, go to [enterprise customer list page](http://localhost:18000/admin/enterprise/enterprisecustomer/) , click on any enterprise customer and you will see `Transmit Courses Metadata` button at top right, click on that button and it will ask for user, use `enterprise_work`. This method uses the management command mentioned above.

**Integrated Channel Configuration:** A channel configuration must exist for both of the above methods. Create any of the SAP SuccessFactors/Degreed/Cornerstone configuration from django admin. You can see a related configuration on stage.  

**Switch:** Enable the `ENTERPRISE_CATALOG_API_ENDPOINTS_ENABLED` switch using django admin.

**Management command**
```bash
python manage.py lms transmit_content_metadata --catalog_user=enterprise_worker --enterprise_customer=a3432c7d-91df-4e98-952e-999f62356ea5
```

**Note:** To make the above things work, You need to install the edx-enterprise as an editable pkg with in LMS from inside of you docker devstack if this is not already configured. See the below steps.

```bash
pip uninstall edx-enterprise   # uninstall the existing edx-enterprise
pip install -e /edx/src/edx-enterprise  # edx-enterprise repo should be cloned in this path
```

I have followed the above steps on my docker devstack and can see the control reaches the changes you made in `enterprise/api_client/enterprise.py::EnterpriseApiClient::get_content_metadata`

enterprise/migrations/0093_enable_enterprise_catalog_api_endpoints.py
```diff
@@ -0,0 +1,26 @@
...
+def create_switch(apps, schema_editor):
+    """Create and activate the ENTERPRISE_CATALOG_API_ENDPOINTS_ENABLED switch if it does not already exist."""
+    Switch = apps.get_model('waffle', 'Switch')
+    Switch.objects.get_or_create(name='ENTERPRISE_CATALOG_API_ENDPOINTS_ENABLED', defaults={'active': False})
+
+
+def delete_switch(apps, schema_editor):
+    """Delete the ENTERPRISE_CATALOG_API_ENDPOINTS_ENABLED switch."""
+    Switch = apps.get_model('waffle', 'Switch')
+    Switch.objects.filter(name='ENTERPRISE_CATALOG_API_ENDPOINTS_ENABLED').delete()
```
muhammad-ammar: I think we usually use lower case for switch names.

enterprise/api_client/enterprise.py
brittneyexline: agreed - ideally the flag would be at the points where we instantiate this client or the other and make the call to it.

brittneyexline:
thank you @muhammad-ammar for reviewing and giving us testing instructions! much appreciated :-)

enterprise/api_client/enterprise.py
```diff
@@ -44,12 +46,16 @@ def get_content_metadata(self, enterprise_customer, enterprise_catalogs=None):
         content_metadata = OrderedDict()
         enterprise_customer_catalogs = enterprise_catalogs or enterprise_customer.enterprise_customer_catalogs.all()
         for enterprise_customer_catalog in enterprise_customer_catalogs:
-            response = self._load_data(
-                self.ENTERPRISE_CUSTOMER_CATALOGS_ENDPOINT,
-                resource_id=str(enterprise_customer_catalog.uuid),
-                traverse_pagination=True,
-                querystring={'page_size': 1000},
-            )
+            if should_use_enterprise_catalog_api():
```
brittneyexline: i don't think the switch logic should live here, but wherever we are calling this function. this client should have no knowledge of any other client.

enterprise/toggles.py
```diff
@@ -0,0 +1,46 @@
...
+# .. toggle_use_cases: incremental_release
+# .. toggle_creation_date: tbd
+# .. toggle_expiration_date: tbd
+# .. toggle_warnings:
+# .. toggle_tickets:
+# .. toggle_status:
+ENTERPRISE_CATALOG_API_ENABLED = WaffleFlag(WAFFLE_FLAG_NAMESPACE, 'enterprise_catalog_api_enabled')
+
+
+def should_use_enterprise_catalog_api():
```
brittneyexline: i imagine this shouldn't be necessary if we're using WaffleFlag, correct? we should be able to do the waffle flag check directly wherever it's needed; this abstraction layer seems unnecessary unless i'm missing something.

chavesj: This is how it's used in other places within openEdx (i.e. [here](https://github.com/edx/edx-platform/blob/61e1eda20df2825a409db3e2d36c69d7c36d3e2d/openedx/features/learner_profile/toggles.py) ). If you search for WaffleSwitch in the edx org you'll find other examples that use it in this way

enterprise/api_client/enterprise.py
chavesj: 👍

enterprise/toggles.py
brittneyexline: ok. i am not sure if this will work based on what i see in Brandon's PR: https://github.com/edx/ecommerce/pull/2862/files#diff-93a1afea8e970e53e92ccb0030b68fb0R139
looks like the request object needs to be passed along for flags where it is not necessary for switches, and the way this is getting loaded ahead of time without the request context means you'd have to move the call within the function anyway. i prefer Brandon's approach because it is simpler and clearer to understand the code. remember that we will be removing this once we have performed the cutover.

fwiw, i haven't followed this paradigm before when i've used waffle in the past, so if you find a way to make it compatible with this pattern that is great, but i personally don't think it is necessary.

chavesj: Ah I see your concern. Although at the top level I wouldn't be passing the current request to the WaffleFlag is_enabled() function, it is obtained using crum [here](https://github.com/edx/edx-platform/blob/d8774e4f8dd20eece13af0fe05d9acb56a53ef10/openedx/core/djangoapps/waffle_utils/__init__.py#L281).

That being said if I use django-waffle flags out of the box I'll need to disable linting on that line since the Travis linting fails with an error saying to use the open edX WaffleFlag. In my opinion the use of WaffleFlag is cleaner and I've already come this far so this is the approach I'm going to stick with unless you are adamant that I use django-waffle.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -70,16 +75,32 @@ def __init__(self, user, enterprise_configuration):
...
+            )
+        else:
+            LOGGER.info(
+                'ENT-2572 [2] - Getting catalog metadata for enterprise [%s] from LMS Enterprise.',
+                self.enterprise_customer.name
+            )
+            content_metadata_items = self.enterprise_api.get_content_metadata(
+                self.enterprise_customer,
+                enterprise_catalogs=self.enterprise_configuration.customer_catalogs_to_transmit
+            )
```
iloveagent57, line 103: Total nit, but you could DRY this up a little bit:
```python
if request and waffle.flag_is_active(request, USE_ENTERPRISE_CATALOG):
    logger_index, metadata_source = (1, "Enterprise Catalog Service")
    metadata_function = self.enterprise_catalog_api.get_content_metadata
else:
    logger_index, metadata_source = (2, "LMS Enterprise")
    metadata_function = self.enterprise_api.get_content_metadata
            
LOGGER.info(
    'ENT-2572 [%s] - Getting catalog metadata for enterprise [%s] from %s.',
    logger_index,
    self.enterprise_customer.name,
    metadata_source
)
content_metadata_items = metadata_function(
    self.enterprise_customer,
    enterprise_catalogs=self.enterprise_configuration.customer_catalogs_to_transmit
)
```

...having typed that up, it doesn't really save many lines of code, so feel free to completely disregard :)

chavesj: This should be cleaned up once we fully transition to using enterprise-catalog endpoints and remove the use of the waffle flag so I think I'm gonna leave as is for now

iloveagent57: 👍 Make sense.

enterprise/migrations/0093_enable_enterprise_catalog_api_endpoints.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
```
adamstankiewicz, line 1: @bbaker6225 merged in his PR that adds this same migration and `USE_ENTERPRISE_CATALOG` constant. If you rebase this PR's branch on `master`, you'll no longer need this migration in the PR.

chavesj: 👍

enterprise/api_client/enterprise_catalog.py
```diff
@@ -25,6 +27,7 @@ class EnterpriseCatalogApiClient(JwtLmsApiClient):
 
     API_BASE_URL = settings.ENTERPRISE_CATALOG_INTERNAL_ROOT_URL + '/api/v1/'
     ENTERPRISE_CATALOG_ENDPOINT = 'enterprise-catalogs'
+    GET_CONTENT_METADATA_ENDPOINT = ENTERPRISE_CATALOG_ENDPOINT + '/{}/get_content_metadata.json'
```
adamstankiewicz, line 30: Is the `.json` extension necessary?

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -70,16 +75,32 @@ def __init__(self, user, enterprise_configuration):
...
-            self.enterprise_customer,
-            enterprise_catalogs=self.enterprise_configuration.customer_catalogs_to_transmit
-        )
+        request = crum.get_current_request()
+        if request and waffle.flag_is_active(request, USE_ENTERPRISE_CATALOG):
+            LOGGER.info(
+                'ENT-2572 [1] - Getting catalog metadata for enterprise [%s] from Enterprise Catalog Service.',
+                self.enterprise_customer.name
+            )
+            content_metadata_items = self.enterprise_catalog_api.get_content_metadata(
```
brittneyexline, line 91: i think it will be important to log the results of what we're getting back from this call for testing purposes, as well as which catalogs we are requesting data for.

enterprise/api_client/enterprise_catalog.py
chavesj: I don't believe so. Will remove

enterprise/api_client/enterprise_catalog.py
```diff
@@ -100,6 +103,39 @@ def delete_enterprise_catalog(self, catalog_uuid):
...
+            list: List of dicts containing content metadata.
+        """
+        content_metadata = OrderedDict()
+        enterprise_customer_catalogs = enterprise_catalogs or enterprise_customer.enterprise_customer_catalogs.all()
+
+        for enterprise_customer_catalog in enterprise_customer_catalogs:
+            catalog_uuid = enterprise_customer_catalog.uuid
+            endpoint = getattr(self.client, self.GET_CONTENT_METADATA_ENDPOINT.format(catalog_uuid))
+            try:
+                response = endpoint.get()
```
brittneyexline, line 125: this might come later, but one outstanding item is that these endpoints should be paginating results, but it hasn't happened yet so this is fine for now. once it does, we may need to add a parameter to make sure we're getting all the results at once.

chavesj: 👍

enterprise/api_client/enterprise_catalog.py
```diff
@@ -53,7 +56,7 @@ def create_enterprise_catalog(
                 'Failed to create EnterpriseCustomer Catalog [%s] in enterprise-catalog due to: [%s]',
                 catalog_uuid, str(exc)
             )
-            return {}
+            raise Exception
```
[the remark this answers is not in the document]
chavesj: 👍

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -70,16 +75,32 @@ def __init__(self, user, enterprise_configuration):
...
+                self.enterprise_customer,
+                enterprise_catalogs=self.enterprise_configuration.customer_catalogs_to_transmit
+            )
+        else:
+            LOGGER.info(
```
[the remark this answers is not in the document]
chavesj: removed ✅

integrated_channels/integrated_channel/exporters/content_metadata.py
chavesj: Logging string updated ✅

PR #734, opened 2020-03-30
ENT-2672 | Removed the 'EDX_API_KEY' from CourseApiClient and directl…

hasnain-naveed:
…y called the EdxRestApiClient.

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/api_client/lms.py
+++ b/enterprise/api_client/lms.py
@@ -58,6 +58,23 @@ class LmsApiClient:
         )
 
 
+class NoAuthenticationLmsApiClient:
+    """
+    Object builds an API client to make calls to the edxapp LMS API.
+
+    Authentication is not required.
+    """
+
+    API_BASE_URL = settings.LMS_INTERNAL_ROOT_URL + '/api/'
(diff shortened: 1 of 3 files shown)


enterprise/api_client/lms.py
```diff
@@ -317,14 +317,22 @@ def get_enrolled_courses(self, username):
         return self.client.enrollment.get(user=username)
 
 
-class CourseApiClient(LmsApiClient):
```
moconnell1453: We're removing LmsApiClient, which uses the header, but we aren't replacing it with JwtLmsApiClient. Does the courses API not require any authentication? If that's the case, can we create a new base class for these API clients to derive from? The base class would create the client rather than relying on each API client to create it.

hasnain-naveed: That seems like a good plan. I was thinking the same thing first. But this is the single and odd case as far as I know. If in the future we encounter this kind of use case again then we can create a parent class and can inherit all the classes from that.

PR #736, opened 2020-03-31
ENT-2483 | Return boolean value of whether or not the course is active

HammadAhmadWaqas:
added `active` key in enterprise catalog
https://openedx.atlassian.net/browse/ENT-2483
**Merge checklist:**

- [ ] Translations updated

**Post merge:**

### DIFF
--- a/integrated_channels/sap_success_factors/exporters/utils.py
+++ b/integrated_channels/sap_success_factors/exporters/utils.py
@@ -7,23 +7,9 @@ from __future__ import absolute_import, unicode_literals
 
 from logging import getLogger
 
-from enterprise.utils import is_course_run_enrollable
 from integrated_channels.sap_success_factors.constants import SUCCESSFACTORS_OCN_LANGUAGE_CODES
 
 LOGGER = getLogger(__name__)
-
-
-def course_available_for_enrollment(course_run):
-    """
-    Check if a course run is available for enrollment.
-    """
-    if course_run['availability'] not in ['Current', 'Starting Soon', 'Upcoming']:
-        # course is archived so not available for enrollment
-        return False
-
-    # now check if the course run is enrollable on the basis of enrollment
-    # start and end date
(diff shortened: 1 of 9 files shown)


integrated_channels/sap_success_factors/exporters/utils.py
```diff
@@ -7,25 +7,11 @@
...
 from logging import getLogger
 
-from enterprise.utils import is_course_run_enrollable
 from integrated_channels.sap_success_factors.constants import SUCCESSFACTORS_OCN_LANGUAGE_CODES
 
 LOGGER = getLogger(__name__)
 
 
-def course_available_for_enrollment(course_run):
-    """
```
HammadAhmadWaqas: moved this method to enterprise/utils with name `is_course_run_available_for_enrollment ` to reuse

test_utils/fake_catalog_api.py
```diff
@@ -910,6 +910,7 @@
     "short_description": None,
     "subject_uuids": [],
     "transcript_languages": [],
+    "course_runs": [],
```
HammadAhmadWaqas: I think `discovery` will never miss this `course_runs` key.  that's why added all places where it was missing.

test_utils/__init__.py
```diff
@@ -135,6 +140,9 @@ def update_course_with_enterprise_context(course, add_utm_info=True, enterprise_
     )
 
     course_runs = course.get('course_runs', [])
+
+    if add_active_info:
+        course['active'] = utils.has_course_run_available_for_enrollment(course_runs)
```
HammadAhmadWaqas: @ziafazal yes we are adding `active` value here to **assert** that if we are getting it from API or not.

HammadAhmadWaqas: [this](https://github.com/edx/edx-enterprise/blob/master/tests/test_enterprise/api/test_views.py#L837) unit test will assert that.

PR #737, opened 2020-04-01
Transition contains_courses behind waffle flag

bbaker6225:
Move the calls on EnterpriseCustomerCatalog::contains_courses to use
the new enterprise catalog service endpoints behind a waffle flag. The
flag will allow us to incrementally switch over time.

ENT-2575

**Merge checklist:**
- [ ] Translations updated

**Post merge:**


enterprise/api/v1/views.py
```diff
@@ -151,6 +151,7 @@ def contains_content_items(self, request, pk, course_run_ids, program_uuids):
 
         contains_content_items = False
         for catalog in enterprise_customer.enterprise_customer_catalogs.all():
+            # Note that these checks are not transitioned over
             contains_course_runs = not course_run_ids or catalog.contains_courses(course_run_ids)
```
bbaker6225, line 155: These calls are not transitioned over so that `edx-enterprise.contains_content_items` is not calling `enterprise-catalog.contains_content_items`

enterprise/api/v1/views.py
```diff
@@ -331,6 +332,7 @@ def contains_content_items(self, request, pk, course_run_ids, program_uuids):
 
         contains_content_items = True
         if course_run_ids:
+            # Note that these checks are not transitioned over
             contains_content_items = enterprise_customer_catalog.contains_courses(course_run_ids)
```
bbaker6225, line 336: Same as above

enterprise/models.py
```diff
@@ -1571,41 +1578,55 @@ def contains_programs(self, program_uuids):
 
         return bool(program_uuids and set(program_uuids).issubset(content_ids_in_catalog))
 
-    def get_course(self, course_key):
+    def get_course(self, course_key, request=None):
```
adamstankiewicz, line 1581: A small suggestion might be to use `crum` to get the current request, if it exists to avoid needing to pass `request` through to `get_course`, `get_course_run`, `get_course_and_course_run `, and `get_program`. 

```python
request = crum.get_current_request()
```

tests/test_models.py
```diff
@@ -1103,11 +1107,73 @@ def test_audit_param_in_program_enrollment_url(self, config_mock):
     def test_get_course_and_course_run_no_content_items(self, contains_courses_mock):
         """
         The ``get_course_and_course_run`` method returns a tuple (None, None) when no content items exist.
+        TODO: Replicate with new flag
```
adamstankiewicz, line 1110: Is this still a TODO?

enterprise/models.py
bbaker6225: Good tip, I didn't know about that!

tests/test_models.py
bbaker6225: Whoops, nope I'll remove this

enterprise/models.py
```diff
@@ -1567,12 +1575,20 @@ def get_course(self, course_key):
 
         Arguments:
             course_key (str): The course key which identifies the course.
+            request (request): Optional argument for detecting if USE_ENTERPRISE_CATALOG flag is active
```
adamstankiewicz, line 1578: Should we remove the docstring comments about the `request` arguments now that it's no longer passed into these methods?

enterprise/models.py
```diff
@@ -1621,8 +1648,15 @@ def get_program(self, program_uuid):
...
-            return None
+        request = crum.get_current_request()
+        # Temporarily gate enterprise catalog api usage behind waffle flag
+        if request and waffle.flag_is_active(request, USE_ENTERPRISE_CATALOG):
+            if not EnterpriseCatalogApiClient(user=request.user).contains_content_items(self.uuid, [program_uuid]):
+                return None
+        else:
+            if not self.contains_programs([program_uuid]):
+                return None
+
```
iloveagent57, line 1659: Nit: the same basic chunk of code is repeated 4 times, maybe we can have a little helper method like
```python
def _check_content_exists(self, request, existence_checker, content_ids, null_value):
    # ....
```
...and you can invoke it like `return self._check_content_exists(request, self.contains_programs, [program_uuid], None)`, or `return self._check_content_exists(request, self.contains_courses, [course_run_id], (None, None))`.

tests/test_models.py
```diff
@@ -1108,6 +1112,82 @@ def test_get_course_and_course_run_no_content_items(self, contains_courses_mock)
...
+    @mock.patch('enterprise.models.EnterpriseCatalogApiClient')
+    @mock.patch('enterprise.api_client.lms.JwtBuilder', mock.Mock())
+    def test_get_course_and_course_run_no_content_items_waffle_flag(self, api_client_mock, get_current_request_mock):
+        """
+        Verify the `get_course_and_course_run` method returns the same tuple with the enterprise catalog flag active.
+        """
+        with override_flag(USE_ENTERPRISE_CATALOG, active=True):
+            # Set up the API client return values
+            api_client_mock.return_value = mock.MagicMock()
+            api_client_mock.return_value.contains_content_items = mock.MagicMock()
```
iloveagent57, line 1125: Nit: I think you can save a few LOC by doing 
```python
@mock.patch('enterprise.models.EnterpriseCatalogApiClient', return_value=mock.MagicMock())
```
and
```python
api_client_mock.return_value.contains_content_items.return_value=False
```

bbaker6225: Good idea, will do

enterprise/models.py
bbaker6225: You're right it's definitely a bit repetitive. Given that we should be removing some of these code paths fairly soon with the rest of the catalog transition, I'm fine with leaving the duplication as it was before for the time being. But I also don't mind updating it if you'd rather

iloveagent57: I'm fine with leaving it as is, too, especially since this code is meant to be transient.

PR #738, opened 2020-04-02
ENT-2675 | Converted EnrollmentApiClient to JWT client.

hasnain-naveed:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/api_client/lms.py
+++ b/enterprise/api_client/lms.py
@@ -11,7 +11,6 @@ from time import time
 
 from edx_rest_api_client.client import EdxRestApiClient
 from opaque_keys.edx.keys import CourseKey
-from requests import Session
 from requests.exceptions import ConnectionError, Timeout  # pylint: disable=redefined-builtin
 from slumber.exceptions import HttpNotFoundError, SlumberBaseException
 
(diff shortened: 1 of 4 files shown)


enterprise/api_client/lms.py
```diff
@@ -150,13 +150,18 @@ def redirect_if_blocked(course_run_ids, user=None, ip_address=None, url=None):
         return None
 
 
-class EnrollmentApiClient(LmsApiClient):
+class EnrollmentApiClient(JwtLmsApiClient):
```
moconnell1453, line 153: Have we confirmed that no one else is using the enrollment client with the API key? If so, this change is fine. However, we may want to leave the LmsApiClient as a base class, and potentially modify the class to log any use of the API key so we can trace who is still using it.

hasnain-naveed: This definition of EnrollmentApiClient is only being used in edx-enterprise. We have changed it from EDX_API_KEY based client to JWT based client and there is no other usage of it.
All three clients which were inheriting `LmsApiClient` are no longer inheriting it we need to remove `LmsApiClient` once removal of EDX_API_KEY work is shipped.

PR #740, opened 2020-04-07
ENT-2755 | Tests are failing due to UNIQUE constraint failed

HammadAhmadWaqas:
Removed all auto-generated fields (e.g: id's) from factories to stop initializing them using `fake.random_int()`

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -1101,16 +1101,6 @@ class EnterpriseCustomerBrandingConfiguration(TimeStampedModel):
         verbose_name_plural = _("Branding Configurations")
         ordering = ['created']
 
-    def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
-        """Save the enterprise customer branding config."""
-        if self.pk is None:
-            logo_image = self.logo
-            self.logo = None
-            super(EnterpriseCustomerBrandingConfiguration, self).save(*args, **kwargs)
-            self.logo = logo_image
-
-        super(EnterpriseCustomerBrandingConfiguration, self).save(*args, **kwargs)
-
     def __str__(self):
         """
         Return human-readable string representation.
--- a/enterprise/settings/test.py
+++ b/enterprise/settings/test.py
@@ -148,8 +148,8 @@ EMAIL_BACKEND = 'django.core.mail.backends.locmem.EmailBackend'
 
 DEFAULT_FROM_EMAIL = 'course_staff@example.com'
 
-USER_THROTTLE_RATE = '90/minute'
-SERVICE_USER_THROTTLE_RATE = '100/minute'
+USER_THROTTLE_RATE = '190/minute'
+SERVICE_USER_THROTTLE_RATE = '200/minute'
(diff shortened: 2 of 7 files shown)


tests/test_enterprise/api/test_views.py
```diff
@@ -1482,7 +1482,7 @@ def create_course_enrollments_context(
             [{
                 'course_mode': 'audit',
                 'course_run_id': 'course-v1:edX+DemoX+Demo_Course',
-                'lms_user_id': 1,
+                'lms_user_id': 10,
```
HammadAhmadWaqas, line 1485: `1` ID is being used for 1st user record that we are creating in setup file in the test class.

enterprise/models.py
```diff
@@ -1101,16 +1101,6 @@ class Meta:
...
-    def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
-        """Save the enterprise customer branding config."""
-        if self.pk is None:
-            logo_image = self.logo
-            self.logo = None
-            super(EnterpriseCustomerBrandingConfiguration, self).save(*args, **kwargs)
-            self.logo = logo_image
-
-        super(EnterpriseCustomerBrandingConfiguration, self).save(*args, **kwargs)
-
```
HammadAhmadWaqas, line 1113: this override method calls save() method twice in case we do not pass it while using its Factory.
and causing an Integrity Error.
And I also can't find any good use of this override method so I removed it and it didn't fail any model test too.

enterprise/settings/test.py
```diff
@@ -148,8 +148,8 @@ def root(*args):
 
 DEFAULT_FROM_EMAIL = 'course_staff@example.com'
 
-USER_THROTTLE_RATE = '90/minute'
-SERVICE_USER_THROTTLE_RATE = '100/minute'
+USER_THROTTLE_RATE = '190/minute'
```
HammadAhmadWaqas, line 151: Now, `test` user (used for the client) has the same `id` every time so 90 throttle rate is not enough for all tests.

tests/test_enterprise/api/test_throttles.py
```diff
@@ -49,6 +49,8 @@ def test_user_throttle(self):
         """
         Make sure throttling works as expected for regular users.
         """
+        self.create_user('test_user', 'QWERTY')
+        self.client.login(username='test_user', password='QWERTY')
```
HammadAhmadWaqas, line 53: needed to change normal client `user` so that other tests do not get affected by consumption of throttle limit.
it was **not** failing before because we used to have different client IDs for every test class.

enterprise/models.py
moconnell1453: This is a bizarre piece of code that certainly warranted a comment explaining its existence. If the id isn't set, then we clear the logo field, save, then reset the logo field and save again? There must have been a reason for this. I agree with removing it, but we should pay close attention to this model after merging.

PR #742, opened 2020-04-09
Use waffle sample instead of waffle flag for using enterprise-catalog service

adamstankiewicz:
**Description**

This PR primarily swaps over the waffle flag to be a waffle sample. The reason for this is because waffle flags have a dependency on the `request` object to determine if the flag is active for a particular user/request. However, 2 of the places that we are migrating to the enterprise-catalog service are triggered by a management command, which inherently does not have any knowledge of a `request` object. As such, the conditional for `waffle.flag_is_active` will always be false in those cases.

By switching to a waffle sample, we will still be able to incrementally increase the percentage of traffic that hits the enterprise-catalog service, without needing to manage both a flag and sample.

https://waffle.readthedocs.io/en/stable/types/sample.html

**Summary**
1. Adds a migration to create the waffle sample object. It also removes the waffle flag that was previously created.
2. Swaps over existing uses of `waffle.flag_is_active` to use `waffle.switch_is_active`.
3. Updates the `EnterpriseCatalogApiClient` to make the `user` argument optional. If the `user` argument is not provided, it will instead get the "enterprise_worker" user instead.
4. No longer passes `user` to the `EnterpriseCatalogApiClient` during initialization in places where a request and/or user object doesn't already exist (particularly in "enterprise/models.py").

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/api_client/enterprise_catalog.py
+++ b/enterprise/api_client/enterprise_catalog.py
@@ -30,6 +30,10 @@ class EnterpriseCatalogApiClient(JwtLmsApiClient):
     GET_CONTENT_METADATA_ENDPOINT = ENTERPRISE_CATALOG_ENDPOINT + '/{}/get_content_metadata'
     APPEND_SLASH = True
 
+    def __init__(self, user=None):
+        user = user if user else utils.get_enterprise_worker_user()
+        super(EnterpriseCatalogApiClient, self).__init__(user)
+
     @JwtLmsApiClient.refresh_token
     def create_enterprise_catalog(
             self,
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -11,7 +11,6 @@ from decimal import Decimal
 from logging import getLogger
 from uuid import uuid4
 
-import crum
 import six
 import waffle
 from django_countries.fields import CountryField
@@ -1579,10 +1578,9 @@ class EnterpriseCustomerCatalog(TimeStampedModel):
         Return:
             dict: The course metadata.
         """
-        request = crum.get_current_request()
-        # Temporarily gate enterprise catalog api usage behind waffle flag
-        if request and waffle.flag_is_active(request, USE_ENTERPRISE_CATALOG):
-            if not EnterpriseCatalogApiClient(user=request.user).contains_content_items(self.uuid, [course_key]):
+        # Temporarily gate enterprise catalog api usage behind waffle sample
+        if waffle.sample_is_active(USE_ENTERPRISE_CATALOG):
+            if not EnterpriseCatalogApiClient().contains_content_items(self.uuid, [course_key]):
                 return None
         else:
             if not self.contains_courses([course_key]):
@@ -1600,10 +1598,9 @@ class EnterpriseCustomerCatalog(TimeStampedModel):
         Return:
             dict: The course run metadata.
         """
-        request = crum.get_current_request()
-        # Temporarily gate enterprise catalog api usage behind waffle flag
-        if request and waffle.flag_is_active(request, USE_ENTERPRISE_CATALOG):
-            if not EnterpriseCatalogApiClient(user=request.user).contains_content_items(self.uuid, [course_run_id]):
+        # Temporarily gate enterprise catalog api usage behind waffle sample
+        if waffle.sample_is_active(USE_ENTERPRISE_CATALOG):
+            if not EnterpriseCatalogApiClient().contains_content_items(self.uuid, [course_run_id]):
                 return None
         else:
             if not self.contains_courses([course_run_id]):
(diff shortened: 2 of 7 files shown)


enterprise/api_client/enterprise_catalog.py
```diff
@@ -30,6 +30,10 @@ class EnterpriseCatalogApiClient(JwtLmsApiClient):
     GET_CONTENT_METADATA_ENDPOINT = ENTERPRISE_CATALOG_ENDPOINT + '/{}/get_content_metadata'
     APPEND_SLASH = True
 
+    def __init__(self, user=None):
+        user = user if user else utils.get_enterprise_worker_user()
```
bbaker6225, line 34: I think it's safer to explicitly pass the user in all the time, so that we don't accidentally have places where we're using the permissions of the enterprise worker instead of a user's (potentially) lower permissions

bbaker6225: From the comment in standup, can we not use `utils.get_enterprise_worker_user()` in the management command?
I get that this is a previous pattern, but it feels less than ideal from a security perspective to me

enterprise/models.py
```diff
@@ -1579,10 +1578,9 @@ def get_course(self, course_key):
         Return:
             dict: The course metadata.
         """
-        request = crum.get_current_request()
-        # Temporarily gate enterprise catalog api usage behind waffle flag
-        if request and waffle.flag_is_active(request, USE_ENTERPRISE_CATALOG):
-            if not EnterpriseCatalogApiClient(user=request.user).contains_content_items(self.uuid, [course_key]):
+        # Temporarily gate enterprise catalog api usage behind waffle sample
+        if waffle.sample_is_active(USE_ENTERPRISE_CATALOG):
+            if not EnterpriseCatalogApiClient().contains_content_items(self.uuid, [course_key]):
```
iloveagent57, line 1583: Suggestion for a small engineering backlog ticket: a pattern I like to use for feature-gating is to create a `config.py` file that contains functions that map directly to whether a feature should be active.  So in this example, it'd be something like `def use_enterprise_catalog(*args, **kwargs): ...`.  Then all of your models, views, etc. can rely on that `config` module as a single interface for determining feature state, without worrying about the underlying switch/flag/settings that actually implement it (because we tend to change how features are toggled over the course of the release process).  I'll throw something in our tech-debt backlog if you think that makes sense.

chavesj: +1. This seems to be in line with the way the open edX WaffleFlag / WaffleSwitch are used in platform

enterprise/api_client/enterprise_catalog.py
adamstankiewicz: This is following the same pattern as the `EnrollmentApiClient` with using the "enterprise_worker" user if no `user` is supplied: https://github.com/edx/edx-enterprise/blob/1c7eba91041b2291dd14dca9b0f39b335d1ae542/enterprise/api_client/lms.py#L138

The `EcommerceApiClient` also uses the "ecommerce_worker" user when initializing the client within models.py, though it passes `request.user` when the client is initialized within a view.

From a security perspective, I think there's a distinction between a request made directly by a user and a service-to-service request. Taking the `get_course` method as an example, [that method is called from within a view](https://github.com/edx/edx-enterprise/blob/5ff8103fe8010bce4f19aeede64763aea4f7b7e0/enterprise/api/v1/views.py#L355), but the user's permissions have already been checked/validated by the `@permission_required` decorator on the `course_detail` endpoint. In this case, it's a service-to-service request that would be safe to use the "enterprise_worker" service user , since the requesting user's permissions have already been checked at that point.

Generally, my understanding is that Django models should typically not be `request`-aware as methods on a model are not always called from within a view associated with a HTTP request. If we want to keep passing in `request.user` to the initialization of `EnterpriseCatalogApiClient` within models.py, we are assuming a request object will exist, when it may not if the method isn't called from a view (e.g., from a management command).

If we begin to initialize `EnterpriseCatalogApiClient` directly within a view (i.e., not from a method on a model), that's when we would pass in the `request.user`.

Definitely feel free to correct me if my above understanding is incorrect, in which case, yes... we could probably use `utils.get_enterprise_worker_user()` within the management command.

bbaker6225: You're correct about the service to service request. I'm just thinking of if we need to consider the case where a user might have the correct permissions to the endpoint they're calling correctly, but not the endpoint that is called by the API client. That's probably not the case, and so everything is fine, but just wanted to explicitly call it out

enterprise/models.py
bbaker6225: I think it's a good practice in general. In my mind this is a simple enough scenario that the config is over-engineering in my mind, but it doesn't matter much

adamstankiewicz: I think this makes sense to throw something onto our tech-debt backlog. Perhaps we could get around to it the next time we need to implement feature-toggling.

PR #743, opened 2020-04-09
Update catalog_contains_course to use enterprise-catalog service if waffle sample is active

adamstankiewicz:
**Description**

This PR adds a conditional check to the `catalog_contains_course` method on the `EnterpriseCustomer` model to call the enterprise-catalog service when the `USE_ENTERPRISE_CATALOG` sample is active. If the sample is not active, the method uses the existing implementation.

Ticket: https://openedx.atlassian.net/browse/ENT-2574

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/api_client/enterprise_catalog.py
+++ b/enterprise/api_client/enterprise_catalog.py
@@ -28,6 +28,7 @@ class EnterpriseCatalogApiClient(JwtLmsApiClient):
     API_BASE_URL = settings.ENTERPRISE_CATALOG_INTERNAL_ROOT_URL + '/api/v1/'
     ENTERPRISE_CATALOG_ENDPOINT = 'enterprise-catalogs'
     GET_CONTENT_METADATA_ENDPOINT = ENTERPRISE_CATALOG_ENDPOINT + '/{}/get_content_metadata'
+    ENTERPRISE_CUSTOMER_ENDPOINT = 'enterprise-customer'
     APPEND_SLASH = True
 
     def __init__(self, user=None):
@@ -150,4 +151,16 @@ class EnterpriseCatalogApiClient(JwtLmsApiClient):
         """
         query_params = {'course_run_ids': content_ids}
         endpoint = getattr(self.client, self.ENTERPRISE_CATALOG_ENDPOINT)(catalog_uuid)
+        return endpoint.contains_content_items.get(**query_params)['contains_content_items']
+
+    @JwtLmsApiClient.refresh_token
+    def enterprise_customer_contains_content_items(self, enterprise_uuid, content_ids):
+        """
+        Checks whether an enterprise customer has any catalogs that contain the provided content ids.
+
+        The endpoint does not differentiate between course_run_ids and program_uuids so they can be used
+        interchangeably. The two query parameters are left in for backwards compatability with edx-enterprise.
+        """
+        query_params = {'course_run_ids': content_ids}
+        endpoint = getattr(self.client, self.ENTERPRISE_CUSTOMER_ENDPOINT)(enterprise_uuid)
         return endpoint.contains_content_items.get(**query_params)['contains_content_items']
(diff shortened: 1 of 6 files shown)


tests/test_enterprise/api_client/test_enterprise_catalog.py
```diff
@@ -140,3 +140,21 @@ def test_contains_content_items():
     client = enterprise_catalog.EnterpriseCatalogApiClient('staff-user-goes-here')
     actual_response = client.contains_content_items(TEST_ENTERPRISE_CATALOG_UUID, ['demoX'])
     assert actual_response == expected_response['contains_content_items']
+
+
+@responses.activate
+@mock.patch('enterprise.api_client.lms.JwtBuilder', mock.Mock())
+def test_enterprise_contains_content_items():
+    url = _url("enterprise-customer/{enterprise_uuid}/contains_content_items/?course_run_ids=demoX".format(
+        enterprise_uuid=TEST_ENTERPRISE_ID))
```
iloveagent57, line 149: Tiniest possible thing I could complain about, feel free to ignore: give the closing parens their own line.

enterprise/api_client/enterprise_catalog.py
```diff
@@ -151,3 +152,15 @@ def contains_content_items(self, catalog_uuid, content_ids):
...
         endpoint = getattr(self.client, self.ENTERPRISE_CATALOG_ENDPOINT)(catalog_uuid)
         return endpoint.contains_content_items.get(**query_params)['contains_content_items']
+
+    @JwtLmsApiClient.refresh_token
+    def enterprise_contains_content_items(self, enterprise_uuid, content_ids):
+        """
+        Checks whether an enterprise customer has any catalogs that contain the provided content ids.
+
+        The endpoint does not differentiate between course_run_ids and program_uuids so they can be used
+        interchangeably. The two query parameters are left in for backwards compatability with edx-enterprise.
```
iloveagent57, line 162: Nit: there's only one query parameter, right?

enterprise/api_client/enterprise_catalog.py
```diff
@@ -151,3 +152,15 @@ def contains_content_items(self, catalog_uuid, content_ids):
         query_params = {'course_run_ids': content_ids}
         endpoint = getattr(self.client, self.ENTERPRISE_CATALOG_ENDPOINT)(catalog_uuid)
         return endpoint.contains_content_items.get(**query_params)['contains_content_items']
+
+    @JwtLmsApiClient.refresh_token
+    def enterprise_contains_content_items(self, enterprise_uuid, content_ids):
```
chavesj, line 157: I'm not seeing any difference between this function and contains_content_items() above

adamstankiewicz: The `contains_content_items` method above is for the catalog-specific endpoint - does this particular catalog contain the content items?

The `enterprise_contains_content_items` method here is for determining if the content items exist in any of the catalogs associated with a particular enterprise customer.

The above function uses ENTERPRISE_CATALOG_ENDPOINT whereas this one uses ENTERPRISE_CUSTOMER_ENDPOINT.

enterprise/api_client/enterprise_catalog.py
adamstankiewicz: Both of the `contains_content_items`-related endpoints accept 2 query params (`course_run_ids` and `program_uuids`) for backwards compatibility with the existing endpoints in edx-enterprise. However, the endpoints in the enterprise-catalog service essentially [merges those 2 query params together](https://github.com/edx/enterprise-catalog/blob/9f6f27ae1689bd9db5d35057d2587e3e862088ea/enterprise_catalog/apps/api/v1/views.py#L106).

This comment and the one for the above `contains_content_items` method are basically just noting that while we're only sending the `course_run_ids` query parameter here, it's OK if the `content_ids` argument contains program ids since they get merged in together with the `course_run_ids` anyways.

enterprise/api_client/enterprise_catalog.py
chavesj: Okay cool I overlooked that. Maybe this is something that could be DRY'd up since the only difference is the endpoint being used - I'll leave it up to you 🙂

tests/test_enterprise/api_client/test_enterprise_catalog.py
chavesj: Agreed, makes me wonder if there's a way the linter could catch this

PR #747, opened 2020-04-10
fixing traversal of results

chavesj:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/api_client/enterprise_catalog.py
+++ b/enterprise/api_client/enterprise_catalog.py
@@ -135,7 +135,7 @@ class EnterpriseCatalogApiClient(JwtLmsApiClient):
                 )
                 response = {}
 
-            for item in response:
+            for item in response['results']:
                 content_id = utils.get_content_metadata_item_id(item)
                 content_metadata[content_id] = item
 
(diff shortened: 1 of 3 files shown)


enterprise/api_client/enterprise_catalog.py
```diff
@@ -135,7 +135,7 @@ def get_content_metadata(self, enterprise_customer, enterprise_catalogs=None):
                 )
                 response = {}
```
adamstankiewicz, line 136: The "empty" response object, in the case of an exception, should include the `results` key as an empty list if we're always going to try iterating through the results in the `for...in` loop below.

Alternatively, perhaps we only try to iterate through the results if an exception does not occur, by moving the `for...in` loop within the try block above.

enterprise/api_client/enterprise_catalog.py
```diff
@@ -135,7 +135,7 @@ def get_content_metadata(self, enterprise_customer, enterprise_catalogs=None):
...
 
-            for item in response:
+            for item in response['results']:
```
adamstankiewicz: The "empty" response object, in the case of an exception, should include the `results` key as an empty list if we're always going to try iterating through the results in the `for...in` loop below, since I believe trying to access the `results` key will throw an error if it doesn't exist.

Alternatively, perhaps we only try to iterate through the `results` if an exception does not occur, by moving the `for...in` loop within the try block above.

PR #751, opened 2020-04-14
ENT-2721 | Added extra exception handling for discovery call and Ente…

hasnain-naveed:
…rpriseCustomerUser model.

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/management/commands/create_enterprise_course_enrollments.py
+++ b/enterprise/management/commands/create_enterprise_course_enrollments.py
@@ -6,6 +6,9 @@ from __future__ import absolute_import, unicode_literals
 
 import logging
 
+from requests.exceptions import ConnectionError, Timeout  # pylint: disable=redefined-builtin
+from slumber.exceptions import SlumberBaseException
+
 from django.core.management import BaseCommand
 from django.db import connection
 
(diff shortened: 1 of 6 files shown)


enterprise/management/commands/create_enterprise_course_enrollments.py
```diff
@@ -39,50 +42,81 @@ def add_arguments(self, parser):
...
+            enterprise_customer_uuid = item['enterprise_customer_uuid']
 
             # pylint: disable=no-member
             enterprise_customer = EnterpriseCustomer.objects.get(uuid=enterprise_customer_uuid)
-            if enterprise_customer.catalog_contains_course(course_run_id):
-                enterprise_customer_user = EnterpriseCustomerUser.objects.get(
+
+            try:
+                does_course_exist_in_catalog = enterprise_customer.catalog_contains_course(course_run_id)
+            except (SlumberBaseException, ConnectionError, Timeout) as exc:
```
ziafazal, line 61: This exception handling needs to done [here](https://github.com/edx/edx-enterprise/blob/dc274195f43fb118c1b7482acc94e5b6fe4c9020/enterprise/models.py#L1536) to cover larger use cases

enterprise/management/commands/create_enterprise_course_enrollments.py
```diff
@@ -108,7 +142,8 @@ def _fetch_course_enrollment_data(self, enterprise_customer_uuid):
                 ON ece.enterprise_customer_user_id = ecu.id
                 AND ece.course_id = sce.course_id
             WHERE
-                ece.id IS NULL
+                ecu.linked = true
```
ziafazal, line 145: why we need this extra condition `linked = true`?

enterprise/models.py
```diff
@@ -932,6 +932,15 @@ def create_order_for_enrollment(self, course_run_id, discount_percentage, sales_
         else:
             LOGGER.warning(failure_log_msg, self.user_id, course_run_id, 'Failed to retrieve a valid ecommerce worker.')
 
+    @classmethod
+    def inactive_users_enterprise_customers(cls, user_id, enterprise_customer):
```
ziafazal, line 936: can we rename this method to `inactivate_other_customers`?

enterprise/management/commands/create_enterprise_course_enrollments.py
hasnain-naveed: Now, we don't delete the record of the model but we set linked=false. And we need only those records which are linked to enterprise_customers.

PR #758, opened 2020-05-04
Pass enterprise name to enterprise-catalog service

adamstankiewicz:
Ticket: https://openedx.atlassian.net/browse/ENT-2770

To add/update the `enrollment_url` and `marketing_url` with the appropriate UTM query parameters, the enterprise-catalog service needs to be aware of the enterprise customer's name associated with the enterprise catalog. Currently the enterprise-catalog service is only aware of the enterprise customer's uuid.

This PR passes the enterprise customer's name to the enterprise-catalog service via the `migrate_enterprise_catalogs` management command and the post-save hook on the edx-enterprise catalogs.

**Merge checklist:**
- [ ] Translations updated

**Post merge:**


enterprise/management/commands/migrate_enterprise_catalogs.py
```diff
@@ -60,6 +60,7 @@ def handle(self, *args, **options):
                 client.create_enterprise_catalog(
                     str(enterprise_catalog.uuid),
                     str(enterprise_catalog.enterprise_customer.uuid),
+                    enterprise_catalog.enterprise_customer.name,
```
bbaker6225, line 63: I think we'll need to update this management command to update enterprise catalogs as well. That way we can make sure that they all have an enterprise customer name attached to them

adamstankiewicz: True. The management command, as is, only creates new catalogs but doesn't update existing ones. Will work that into this PR as well.

adamstankiewicz: I updated the management command to update with a PUT request if a corresponding catalog already exists in the service. If the related catalog does not yet exist, it will be created.

PR #759, opened 2020-05-04
ENT-2803: Adding the use of traverse_pagination in get_content_metadata within the enterprise_catalog API Client

chavesj:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**


chavesj:
This change will need to be tested on stage due to the lack of a working configuration for integrated channels in devstack. To test I will be running the stage-edx-transmit_content_metadata Jenkins job with the 'enterprise_worker' user as the argument. This job triggers the transmit_content_metadata management command which uses the get_content_metadata function in which my changes were made.

chavesj:
[Jira Ticket](https://openedx.atlassian.net/browse/ENT-2803)

enterprise/api_client/enterprise_catalog.py
```diff
@@ -128,6 +128,13 @@ def get_content_metadata(self, enterprise_customer, enterprise_catalogs=None):
             endpoint = getattr(self.client, self.GET_CONTENT_METADATA_ENDPOINT.format(catalog_uuid))
             try:
                 response = endpoint.get()
+                results = utils.traverse_pagination(response, endpoint)
```
bbaker6225, line 131: You likely already know this, but just a reminder that this won't work until the pagination for `get_content_metadata` in enterprise catalog is actually implemented

enterprise/api_client/enterprise_catalog.py
```diff
@@ -128,6 +128,13 @@ def get_content_metadata(self, enterprise_customer, enterprise_catalogs=None):
...
+                response = {
+                    'count': len(results),
+                    'next': 'None',
+                    'previous': 'None',
+                    'results': results,
+                }
                 for item in response['results']:
```
iloveagent57, line 138: Do we actually need `response` here?  Can this be shortened to
```python
response = endpoint.get()
for item in utils.traverse_pagination(response, endpoint):
    ...
```

chavesj: yeah good point, will update

PR #764, opened 2020-05-05
ENT-2858 | Created an endpoint to link a user to an enterprise customer.

hasnain-naveed:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -24,6 +24,8 @@ from six.moves.urllib.parse import quote_plus, unquote  # pylint: disable=import
 from django.apps import apps
 from django.conf import settings
 from django.core import mail
+from django.core.exceptions import ValidationError
+from django.core.validators import validate_email
 from django.http import Http404
 from django.utils.decorators import method_decorator
 from django.utils.translation import ugettext as _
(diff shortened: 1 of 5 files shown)


enterprise/api/v1/views.py
```diff
@@ -619,3 +620,57 @@ def post(self, request):
...
+
+        enterprise_customer = get_enterprise_customer(enterprise_customer_uuid)
+
+        if enterprise_customer is None:
+            raise LinkUserRequestError(
+                "There is no enterprise customer with the following UUID [{enterprise_customer_uuid}]".format(
+                    enterprise_customer_uuid=enterprise_customer_uuid
+                )
+            )
+        if email:
```
irfanuddinahmad, line 651: Check for None and validate.

hasnain-naveed: Already added. TYPO

enterprise/api/v1/views.py
```diff
@@ -619,3 +621,57 @@ def post(self, request):
                 {'error': str('Request codes email could not be sent')},
                 status=HTTP_500_INTERNAL_SERVER_ERROR
             )
+
+
+class LinkUserView(APIView):
+    """
+    API to link the user to an enterprise customer.
+    """
+    permission_classes = (permissions.IsAuthenticated,)
```
muhammad-ammar, line 630: @hasnain-naveed Can we make this endpoint only available to service users?

muhammad-ammar: would be great if we can add a test too.

hasnain-naveed:
No need for this PR. We already have an endpoint for this use case.

PR #772, opened 2020-05-11
ENT-2788 Change where we sync metadata to post save hook on the Enterprise Catalog model

brittneyexline:
The previous spot for these updates was not catching django admin changes from the in-line form on the Enterprise Customer django admin screen. This will also future proof against any changes that may occur to these models that need to be synced that could happen outside of django admin.

**Merge checklist:**
- [ ] Translations updated

**Post merge:**


brittneyexline:
marked as WIP because I still need to update unit tests and make sure things look good locally

enterprise/signals.py
```diff
@@ -135,6 +136,57 @@ def delete_enterprise_learner_role_assignment(sender, instance, **kwargs):     #
...
+            # service, so we should update the existing catalog
+            update_fields = {
+                'enterprise_customer': str(instance.enterprise_customer.uuid),
+                'enterprise_customer_name': instance.enterprise_customer.name,
+                'title': instance.title,
+                'content_filter': instance.content_filter,
+                'enabled_course_modes': instance.enabled_course_modes,
+                'publish_audit_enrollment_urls': instance.publish_audit_enrollment_urls,
+            }
+            catalog_client.update_enterprise_catalog(catalog_uuid, **update_fields)
```
iloveagent57, line 174: Out of curiosity, why don't we cast `catalog_uuid` to a string here, like we're doing in the `create_enterprise_catalog()` call?

enterprise/signals.py
```diff
@@ -135,6 +136,57 @@ def delete_enterprise_learner_role_assignment(sender, instance, **kwargs):     #
...
+    Send data changes to Enterprise Catalogs to the Enterprise Catalog Service.
+    """
+    catalog_uuid = instance.uuid
+    try:
+        catalog_client = EnterpriseCatalogApiClient()
+        response = catalog_client.get_enterprise_catalog(catalog_uuid)
+    except NotConnectedToOpenEdX as exc:
+        logger.exception('Unable to update Enterprise Catalog {}'.format(str(catalog_uuid)), exc)
+    else:
+        if kwargs['created'] or not response:
```
iloveagent57, line 151: I know it shouldn't happen, but what if `kwargs['created']` is true _and_ we get a response?

tests/test_enterprise/test_signals.py
```diff
@@ -532,3 +533,87 @@ def test_receiver_does_not_call_task_if_ecu_not_exists(self, mock_task):
...
+    """
+
+    @mock.patch('enterprise.signals.EnterpriseCatalogApiClient')
+    def test_delete_catalog(self, api_client_mock):
+        enterprise_catalog = EnterpriseCustomerCatalogFactory()
+        enterprise_catalog_uuid = enterprise_catalog.uuid
+        api_client_mock.return_value = mock.MagicMock()
+        api_client_mock.return_value.get_enterprise_catalog = mock.MagicMock()
+        api_client_mock.return_value.get_enterprise_catalog.return_value = True
+        api_client_mock.return_value.delete_enterprise_catalog = mock.MagicMock()
```
iloveagent57, line 551: Mocked objects will return mocks by default, so I _think_ you can condense all of this into
```python
api_client_mock.return_value.get_enterprise_catalog.return_value = True
```

enterprise/signals.py
brittneyexline: i'm not sure why exactly the client ended up being inconsistent in this way, but after taking a quick glance at the client it appears to be because for non-POST calls, the uuid is part of the path and thus constructed through slumber vs. being part of the post parameter data where everything needs to be a string.

PR #773, opened 2020-05-13
ENT-2885: adding can_use_enterprise_catalog utility function to exclude enterpr…

chavesj:
…ises from transition to enterprise-catalog

**Merge checklist:**
- [ ] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files)
    - `base.in` if needed in production but edx-platform doesn't install it
    - `test-master.in` if edx-platform pins it, with a matching version
    - `make upgrade && make requirements` have been run to regenerate requirements
- [ ] `make static` has been run to update webpack bundling if any static content was updated
- [x] [Version](https://github.com/edx/edx-enterprise/blob/master/enterprise/__init__.py) bumped
- [x] [Changelog](https://github.com/edx/edx-enterprise/blob/master/CHANGELOG.rst) record added
- [ ] Translations updated

**Post merge:**
- [ ] Tag pushed and a new [version](https://github.com/edx/edx-enterprise/releases) released
- [ ] Version pushed to [PyPI](https://pypi.org/project/edx-enterprise/)
- [ ] PR created in [edx-platform](https://github.com/edx/edx-platform) to upgrade dependencies (including edx-enterprise)

PR #775, opened 2020-05-19
Chavesj/ent 2885

chavesj:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**


CHANGELOG.rst
```diff
@@ -14,6 +14,18 @@ Change Log
 Unreleased
 --------------------
 
+[3.2.13] - 2020-05-19
+---------------------
+
+* Converting UUID fields to string for use in can_use_enterprise_catalog
```
brittneyexline, line 20: should these be combined? they have the same release version, and neither matches what's below...

adamstankiewicz: @brittneyexline Spoke with @chavesj about this. I believe he just needs to rebase this branch with `master`. Most of the changes in this PR were actually merged from #773, but uses the same branch of `chavesj/ENT-2885`. After rebasing, this PR could be significantly smaller.

adamstankiewicz: Disregard me, I saw you mentioned rebasing in the overall review comment :)

brittneyexline: i didn't realize that entirely but it makes sense. i felt like i had pr review dejavu...

enterprise/models.py
```diff
@@ -432,7 +432,7 @@ def catalog_contains_course(self, course_run_id):
             bool: Whether the enterprise catalog includes the given course run.
         """
         # Temporarily gate enterprise catalog api usage behind waffle sample
-        if utils.can_use_enterprise_catalog(self.uuid):
+        if utils.can_use_enterprise_catalog(str(self.uuid)):
```
adamstankiewicz, line 435: Would it make sense to cast the `uuid` as a string within `can_use_enterprise_catalog` so we don't have to do it in multiple places?

PR #776, opened 2020-05-20
update the catalog preview url to use the catalog service

kellmarie:
ENT-2577

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -25,7 +25,6 @@ from django.http import Http404
 from django.template.loader import render_to_string
 from django.urls import reverse
 from django.utils.dateparse import parse_datetime
-from django.utils.html import format_html
 from django.utils.text import slugify
 from django.utils.translation import ugettext as _
 from django.utils.translation import ungettext
@@ -41,11 +40,6 @@ try:
     from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
 except ImportError:
     configuration_helpers = None
-
-try:
-    from openedx.core.djangoapps.catalog.models import CatalogIntegration
-except ImportError:
-    CatalogIntegration = None
 
 try:
     from lms.djangoapps.branding.api import get_url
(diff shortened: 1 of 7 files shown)


enterprise/api_client/lms.py
```diff
@@ -18,7 +18,11 @@
 from django.utils import timezone
 
 from enterprise.constants import COURSE_MODE_SORT_ORDER, EXCLUDED_COURSE_MODES
-from enterprise.utils import NotConnectedToOpenEdX, get_enterprise_worker_user
+try:
+    from enterprise.utils import NotConnectedToOpenEdX, get_enterprise_worker_user
+except ImportError:
+    NotConnectedToOpenEdX = None
+    get_enterprise_worker_user = None
```
kellmarie, line 25: So adding a `try` around these import statements fixed the initialization error that I was getting when importing EnterpriseCatalogApiClient in `enterprise.utils.py`

But then running `test_forms.py` has errors with this when initializing the mock catalog info.

brittneyexline: could you remove this change as well as the installed apps added in `enterprise/settings/test.py`?  i would like to see what test failures you got before these changes... i suspect we should be able to resolve by mocking something in a test

kellmarie: sure I'll do that

kellmarie: Removed those changes

enterprise/utils.py
```diff
@@ -42,10 +42,7 @@
 except ImportError:
     configuration_helpers = None
 
-try:
-    from openedx.core.djangoapps.catalog.models import CatalogIntegration
-except ImportError:
-    CatalogIntegration = None
+from enterprise.api_client.enterprise_catalog import EnterpriseCatalogApiClient
```
brittneyexline, line 45: so the issue seems to be that importing this class has a cascade effect of importing `JwtLmsApiClient` which is what causes the import error. you will want to copy/paste the try/except ImportError logic that was used for CatalogIntegration above.

enterprise/utils.py
```diff
@@ -1000,29 +997,17 @@ def _get_service_worker(service_worker_username):
...
-                'installed in an Open edX environment.'
-            )
-        )
-    discovery_root_url = CatalogIntegration.current().get_internal_api_url()
-    disc_url = '{discovery_root_url}{search_all_endpoint}?{query_string}'.format(
-        discovery_root_url=discovery_root_url,
-        search_all_endpoint='search/all/',
-        query_string=urlencode(content_filter, doseq=True)
-    )
+    preview_url = EnterpriseCatalogApiClient.API_BASE_URL + EnterpriseCatalogApiClient.GET_CONTENT_METADATA_ENDPOINT.format(uuid)
```
brittneyexline, line 1004: then here, i would recommend creating a static function in `EnterpriseCatalogApiClient` that returns the preview url so you can mock it in this method and test that function separately. it brings up the question of whether this function is necessary; why not call the function on the client directly?  that's an option too; but at the end of the day the caller will need to make sure `EnterpriseCatalogApiClient` is mocked out in the test, and the function returning the url is tested separately.

kellmarie: Sounds good, I'll do that.

PR #777, opened 2020-05-21
ENT-2819 | Enterprise's slug login backend changes.

hasnain-naveed:
**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/forms.py
+++ b/enterprise/forms.py
@@ -8,10 +8,18 @@ from django import forms
 from django.utils.translation import ugettext as _
 
 from enterprise.models import EnterpriseCustomer, EnterpriseCustomerUser
+from enterprise.utils import get_enterprise_customer_idp
 
 ENTERPRISE_SELECT_SUBTITLE = _(
     u'You have access to multiple organizations. Select the organization that you will use '
     'to sign up for courses. If you want to change organizations, sign out and sign back in.'
+)
+ENTERPRISE_LOGIN_TITLE = _(u'Enter the organization name')
+ENTERPRISE_LOGIN_SUBTITLE = _(
+    u'Have an account through your company, school, or organization? Enter your organization’s name below to sign in.'
+)
+ERROR_MESSAGE_FOR_SLUG_LOGIN = _(
+    "The attempt to login with this organization name was not successful. Please try again, or contact our support."
 )
 
 
@@ -49,4 +57,35 @@ class EnterpriseSelectionForm(forms.Form):
         if not EnterpriseCustomerUser.objects.filter(enterprise_customer=enterprise, user_id=self._user_id).exists():
             raise forms.ValidationError(_("Wrong Enterprise"))
 
+        return cleaned_data
+
+
+class EnterpriseLoginForm(forms.Form):
+    """
+    Enterprise Slug Login Form.
+    """
+
+    enterprise_slug = forms.CharField(label='Company Name')
+
+    def clean(self):
+        """
+        Validate POST data.
+        """
+        cleaned_data = super(EnterpriseLoginForm, self).clean()
+        enterprise_slug = cleaned_data['enterprise_slug']
+
+        # verify that given slug has any associated enterprise customer.
+        try:
+            enterprise_customer = EnterpriseCustomer.objects.get(slug=enterprise_slug)
+        except EnterpriseCustomer.DoesNotExist:
+            raise forms.ValidationError(ERROR_MESSAGE_FOR_SLUG_LOGIN)
+
+        # verify that enterprise customer has enabled the slug login feature.
+        if not enterprise_customer.enable_slug_login:
+            raise forms.ValidationError(ERROR_MESSAGE_FOR_SLUG_LOGIN)
+
+        # verify that a valid idp is linked to the enterprise customer.
+        enterprise_customer_idp = get_enterprise_customer_idp(enterprise_slug)
+        if enterprise_customer_idp is None or enterprise_customer_idp.identity_provider is None:
+            raise forms.ValidationError(ERROR_MESSAGE_FOR_SLUG_LOGIN)
         return cleaned_data
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -417,6 +417,23 @@ def get_enterprise_customer_user(user_id, enterprise_uuid):
         return None
 
 
+def get_enterprise_customer_idp(enterprise_customer_slug):
+    """
+    Return the identity provider for the given enterprise customer's slug if exists otherwise None.
+
+    Arguments:
+        enterprise_customer_slug (str): enterprise customer's slug.
+
+    Returns:
+        (EnterpriseCustomerIdentityProvider): enterprise customer identity provider record.
+    """
+    EnterpriseCustomerIdentityProvider = apps.get_model(    # pylint: disable=invalid-name
+        'enterprise',
+        'EnterpriseCustomerIdentityProvider'
+    )
+    return EnterpriseCustomerIdentityProvider.objects.filter(enterprise_customer__slug=enterprise_customer_slug).first()
+
+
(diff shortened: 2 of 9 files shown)


hasnain-naveed:
This is the subtask of ENT-2816. These are the backend changes but to run it, you should have backend + frontend changes. There is a branch `hasnain-naveed/ENT-2818` which have backend + frontend changes.
@moconnell1453 @saleem-latif @HammadAhmadWaqas @irfanuddinahmad

test_utils/__init__.py
```diff
@@ -358,3 +360,37 @@ def reset(self):
...
+        """
+        Mocked the rendering the template file.
+        """
+        super(EnterpriseFormViewTestCase, self).setUp()
+        # create a temporary template file
+        # rendering View's template fails becuase of dependency on edx-platform
+        tpl = tempfile.NamedTemporaryFile(
+            prefix='test_template.',
+            suffix=".html",
+            dir=settings.REPO_ROOT + '/templates/enterprise/',
```
HammadAhmadWaqas, line 385: nit. can we use os.path.join here ?

tests/test_enterprise/views/test_enterprise_login_view.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        Test that view HTTP GET works as expected.
+        """
+        response = self.client.get(self.url)
+        assert response.status_code == 200
+        assert response.context['enterprise_login_title_message'] == ENTERPRISE_LOGIN_TITLE
+        assert response.context['enterprise_login_subtitle_message'] == ENTERPRISE_LOGIN_SUBTITLE
+        assert list(response.context['form'].fields.keys())[0] == 'enterprise_slug'
+
+    def _assert_post_request(self, enterprise_slug, status_code, expected_response):
```
HammadAhmadWaqas, line 51: can we rename `status_code` to `expected_status_code` for better clearity

enterprise/utils.py
```diff
@@ -143,6 +143,34 @@ def get_idp_choices():
     return None
 
 
+def get_provider_login_url(request, provider_id):
+    """
+    Return the given provider's login URL.
+    """
+    try:
+        from third_party_auth import pipeline
+    except ImportError as exception:
```
HammadAhmadWaqas, line 152: In what circumcetances we will be getting this Import Error?

enterprise/utils.py
```diff
@@ -143,6 +143,34 @@ def get_idp_choices():
...
+        LOGGER.warning("Could not import Pipeline from third_party_auth")
+        LOGGER.warning(exception)
```
HammadAhmadWaqas, line 154: it is not recommended to use same logger in adjacent lines, we can use single logger to print a complete message.
If you are intended to print stack_trace than you should use either `exception` logger or pass param `exc_info=True`

enterprise/utils.py
```diff
@@ -143,6 +143,34 @@ def get_idp_choices():
...
+        get_next_url_for_login_page = None  # pylint: disable=redefined-outer-name
+
+    third_auth_login_page = ""
+    if pipeline and get_next_url_for_login_page:
+        third_auth_login_page = pipeline.get_login_url(
+            provider_id,
+            pipeline.AUTH_ENTRY_LOGIN,
+            redirect_url=get_next_url_for_login_page(request)
+        )
+    return third_auth_login_page
```
HammadAhmadWaqas, line 171: what behavior user will face in case this method will return an empty URL?
nit: can we rename `third_auth_login_page` to something line `third_auth_login_page_url` or provider_login_url for clearity.

enterprise/utils.py
hasnain-naveed: We are importing this from edx-platform. As enterprise got installed in edx-platform and we are practicing this from start.

enterprise/utils.py
hasnain-naveed: Users will see an error now. I have changed the logic.

enterprise/utils.py
```diff
@@ -417,6 +445,26 @@ def get_enterprise_customer_user(user_id, enterprise_uuid):
...
+
+    Returns:
+        (EnterpriseCustomerIdentityProvider): enterprise customer identity provider record.
+    """
+    EnterpriseCustomerIdentityProvider = apps.get_model(    # pylint: disable=invalid-name
+        'enterprise',
+        'EnterpriseCustomerIdentityProvider'
+    )
+    try:
+        return EnterpriseCustomerIdentityProvider.objects.get(enterprise_customer__slug=enterprise_customer_slug)
```
saleem-latif, line 463: you can simplify this by using the following.

```
return EnterpriseCustomerIdentityProvider.objects.filter(enterprise_customer__slug=enterprise_customer_slug).first()
```

enterprise/utils.py
irfanuddinahmad: Would it be a better design to keep the backend in edx-platform and only call the enterprise to get the IDP url? ... bringing the pipeline into enterprise is creating a lot of unnecessary coupling ...

enterprise/utils.py
```diff
@@ -143,6 +143,34 @@ def get_idp_choices():
     return None
 
 
+def get_provider_login_url(request, provider_id):
```
zamanafzal, line 146: I think we should move this method in platform as it will reduce complexity and unnecessary imports.

enterprise/forms.py
```diff
@@ -8,11 +8,20 @@
...
     'to sign up for courses. If you want to change organizations, sign out and sign back in.'
 )
+ENTERPRISE_LOGIN_TITLE = _(u'Enter the organization name')
+ENTERPRISE_LOGIN_SUBTITLE = _(
+    u'Have an account through your company, school, or organization? Enter your organization’s name below to sign in.'
+)
+ERROR_MESSAGE_FOR_WRONG_ENTERPRISE_CUSTOMER = _(
+    "Either you have entered the wrong name or your company hasn't enabled the login."
+)
+ERROR_MESSAGE_FOR_WRONG_IDP = _("Error occurred to your company/organization login. Kindly contact admin.")
```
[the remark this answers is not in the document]
hasnain-naveed: Already asked on the story.

enterprise/utils.py
hasnain-naveed: @irfanuddinahmad excellent suggestion. FYI we are already doing this kind of [thing](https://github.com/edx/edx-enterprise/blob/e656a52bc58cff879fcebc301bac25a601c639e0/enterprise/tpa_pipeline.py#L23) already. But I have incorporated your feedback. I have created the same method in `edx-platform/openedx/features/enterprise_support/utils.py`.

enterprise/utils.py
hasnain-naveed: Already done.

enterprise/views.py
```diff
@@ -847,6 +859,47 @@ def post(self, request):
...
+        """
+        # flatten the list of lists
+        errors = [item for sublist in form.errors.values() for item in sublist]
+        return JsonResponse({'errors': errors}, status=400)
+
+    def form_valid(self, form):
+        """
+        If the form is valid, redirect to the third party auth login page.
+        """
+        provider_login_url = get_provider_login_url(
```
HammadAhmadWaqas, line 892: `get_provider_login_url ` will be None in case of import error.
I guess you want to write something like
```
# This case will only happend when we try to run the edx-enterprise independently.
if not get_provider_login_url:
      return JsonResponse({'errors': [ERROR_MESSAGE_FOR_SLUG_LOGIN]}, status=400)
provider_login_url = get_provider_login_url(
       self.request,
       get_enterprise_customer_idp(form.cleaned_data['enterprise_slug']).provider_id
)
return JsonResponse({"url": provider_login_url})

```

hasnain-naveed: Typo. you are right.

PR #785, opened 2020-05-30
reduce log exceptions that aren't really errors

adamstankiewicz:
There is a signal for `post_save` on the `EnterpriseCustomerCatalog` model. It is triggered when a catalog gets created or updated (e.g., by ECS). In the signal, we attempt to get a catalog with no assumption that it actually exists yet in enterprise-catalog. If it 404s, we know it doesn't exist and we should create a new catalog in the service.

Since we expect to get a 404 when a catalog is first created in the LMS, we should no longer log it as an exception to avoid seeing false positives for the error in the email alerts.

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,13 @@ Change Log
 Unreleased
 --------------------
 
+[3.2.20] - 2020-06-01
+---------------------
+
+* Suppress the 404 exception in get_enterprise_catalog when we expect it
+* Add enterprise_customer_uuid to an error message to be more informative
+
+
 [3.2.19] - 2020-06-01
 ---------------------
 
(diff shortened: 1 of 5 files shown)


adamstankiewicz:
> it's a little odd to be passing a param to control logging, but i'm ok with it because i feel like we need to think about how to improve this data syncing situation in general.

Since we use `get_enterprise_catalog` to determine when we need to create a new catalog in enterprise-catalog (i.e., if it 404s, we know we need to create vs. update), there's no way to avoid an exception being thrown unless we suppress it or move the exception logging outside of `get_enterprise_catalog` to after it was called. That felt even more odd since that would remove the exception from occurring in places we do want it to happen.

PR #787, opened 2020-06-01
update seed_enterprise_devstack_data command to include name on profiles

adamstankiewicz:
Update seed_enterprise_devstack_data command to include name on user profiles when creating enterprise users since it's technically a required field (without name, things seem to break).

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/management/commands/seed_enterprise_devstack_data.py
+++ b/enterprise/management/commands/seed_enterprise_devstack_data.py
@@ -138,11 +138,10 @@ class Command(BaseCommand):
         if not UserProfile:
             LOGGER.error('UserProfile module does not exist.')
             return
-
-        user_profile = UserProfile.objects.get_or_create(user=user)
-        if user_profile:
-            user_profile.name = 'Test Enterprise User'
-            user_profile.save()
+        UserProfile.objects.update_or_create(
+            user=user,
+            defaults={'name': 'Test Enterprise User'},
+        )
 
     def _add_user_to_groups(self, user, role):
         """ Adds a user with a given role to the appropriate groups """
--- a/enterprise/management/commands/seed_enterprise_devstack_data.py
+++ b/enterprise/management/commands/seed_enterprise_devstack_data.py
@@ -139,7 +139,7 @@ class Command(BaseCommand):
             LOGGER.error('UserProfile module does not exist.')
             return
 
-        user_profile = UserProfile.objects.get(user=user)
+        user_profile = UserProfile.objects.get_or_create(user=user)
         if user_profile:
             user_profile.name = 'Test Enterprise User'
             user_profile.save()
--- a/enterprise/management/commands/seed_enterprise_devstack_data.py
+++ b/enterprise/management/commands/seed_enterprise_devstack_data.py
@@ -3,6 +3,7 @@ Management command for assigning enterprise roles to existing enterprise users.
 """
 from __future__ import absolute_import, unicode_literals
 
+import json
 import logging
 import textwrap
 
@@ -32,6 +33,12 @@ from enterprise.models import (
     SystemWideEnterpriseRole,
     SystemWideEnterpriseUserRoleAssignment,
 )
+
+try:
+    from student.models import UserProfile
+except ImportError:
+    UserProfile = None
+
 
 LOGGER = logging.getLogger(__name__)
 
@@ -114,6 +121,7 @@ class Command(BaseCommand):
                 # slightly different parameters, we will attempt to use the existing user.
                 user = User.objects.get(username=username)
             if user:
+                self._add_name_to_user_profile(user)
                 self._add_user_to_groups(user=user, role=role)
                 self._create_system_wide_role_assignment(user=user, role=role)
                 self._create_feature_role_assignments(user=user, role=role)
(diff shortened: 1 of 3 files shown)


binodpant:
sorry got tied up in calls will review this afternoon and reach back soon

enterprise/management/commands/seed_enterprise_devstack_data.py
```diff
@@ -125,6 +133,17 @@ def _create_enterprise_user(self, username, role):
...
 
+    def _add_name_to_user_profile(self, user):
+        """ Adds a name to a user's profile. """
+        if not UserProfile:
+            LOGGER.error('UserProfile module does not exist.')
+            return
+
+        user_profile = UserProfile.objects.get(user=user)
+        if user_profile:
+            user_profile.name = 'Test Enterprise User'
```
binodpant, line 144: does this fill in the Full Name field as per what we had tried manually?

binodpant:
Any testing notes on how I can verify locally? do I repeat the devstack instructions, or just some part of it to verify this locally?

enterprise/management/commands/seed_enterprise_devstack_data.py
binodpant: I was going to run ```./manage.py lms seed_enterprise_devstack_data``` after syncing this branch on edx-enterprise, if that is not correct let me know

binodpant:
FWIW I did the following (I had already run this before.. so had existing users)

```bash
  cd edx-enterprise
  git checkout astankiewicz/add-full-name-to-profile
  make lms-shell
  ./manage.py lms seed_enterprise_devstack_data
```

I then logged in as a few other users listed at http://localhost:18000/admin/enterprise/enterprisecustomeruser/

into 

http://localhost:18000/account/settings

But I see the full name field is empty,


Perhaps I did not do this correctly?

enterprise/management/commands/seed_enterprise_devstack_data.py
adamstankiewicz: @binodpant Yes, that is correct. The "Full Name" field maps to the `name` attribute on the `UserProfile` model.

Also, that is correct you should be able to checkout this edx-enterprise branch and re-run the management command. edx-enterprise is a django app that gets installed into edx-platform. So, to test local changes to edx-enterprise (e.g., checking out this branch), there's an extra step.

You should have a `src` directory as a sibling to your `devstack` directory. Our docker containers can access things that live within that `src` directory as it gets mounted to the container.

To install a local copy of edx-enterprise into edx-platform for testing:
1. Move/copy your edx-enterprise folder into the `src` directory.
2. Run `make lms-shell` from your devstack directory.
3. Run `pip install -e /edx/src/edx-enterprise` to uninstall the published edx-enterprise package and instead install your local copy from `src`.

binodpant: Thanks, after above and doing make lms-restart I ran again the manage.py command and I got this 

```
Creating enterprise users and assigning roles...
  File "./manage.py", line 123, in <module>
    execute_from_command_line([sys.argv[0]] + django_args)
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.5/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.5/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.5/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.5/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/edx/src/edx-enterprise/enterprise/management/commands/seed_enterprise_devstack_data.py", line 241, in handle
    role=ENTERPRISE_LEARNER_ROLE,
  File "/edx/src/edx-enterprise/enterprise/management/commands/seed_enterprise_devstack_data.py", line 124, in _create_enterprise_user
    self._add_name_to_user_profile(user)
  File "/edx/src/edx-enterprise/enterprise/management/commands/seed_enterprise_devstack_data.py", line 142, in _add_name_to_user_profile
    user_profile = UserProfile.objects.get(user=user)
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.5/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.5/site-packages/django/db/models/query.py", line 408, in get
    self.model._meta.object_name
student.models.DoesNotExist: UserProfile matching query does not exist.
```

Do I also need to run any migrations etc?

adamstankiewicz: Nope, no migrations are needed. I pushed up a change where we will now `get_or_create` a UserProfile instead of just trying to get it. In the case a `UserProfile` doesn't exist for a user, it will be created and returned. If a `UserProfile` does already exist for a user, it will be returned.

PR #795, opened 2020-06-11
Consider course run status when determining availability and enrollability

mattdrayer:
There are situations where a course run is considered "Current" and included in the API response, however it is flagged as "unpublished".  These runs should not be included in the API response.

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -741,6 +741,10 @@ def is_course_run_enrollable(course_run):
     2. enrollment_start is less than now, or undefined
     3. enrollment_end is greater than now, or undefined
     """
+    # Check if the course run is unpublished (sometimes these sneak through)
+    if not is_course_run_published(course_run):
+        return False
+
     now = datetime.datetime.now(pytz.UTC)
     reasonable_enrollment_window = now + datetime.timedelta(days=1)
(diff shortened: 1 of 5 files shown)


mattdrayer:
Tagging @HammadAhmadWaqas since he was the last person in this area, and @adamstankiewicz because he recently worked on some related logic in the enterprise-catalog repository.

enterprise/utils.py
```diff
@@ -755,6 +759,10 @@ def is_course_run_available_for_enrollment(course_run):
     """
     Check if a course run is available for enrollment.
     """
+    # Check if the course run is unpublished (sometimes these sneak through)
+    if not is_course_run_published(course_run):
+        return False
+
```
HammadAhmadWaqas, line 765: do we really need to add these lines here as well? as are already checking for `published` status in the `is_course_run_published` which we are calling from this method in the last line.

mattdrayer: Good catch -- I reviewed all of the dependencies on these methods and it looks like we can indeed remove this.  I've refactored the method to make it a bit more clear what is going on since I originally missed the call to `is_course_run_enrollable` in the return statement.  Thanks!

PR #798, opened 2020-06-12
ENT-2876: Update GrantDataSharingPermissionView to accept both, course_run_id as well as course_key in the course_id param

sameenfatima78:
JIRA: https://openedx.atlassian.net/browse/ENT-2876

**Merge checklist:**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -13,6 +13,8 @@ import pytz
 import waffle
 from dateutil.parser import parse
 from ipware.ip import get_ip
+from opaque_keys import InvalidKeyError
+from opaque_keys.edx.keys import CourseKey
 from six.moves.urllib.parse import urlencode, urljoin  # pylint: disable=import-error
 
 from django.conf import settings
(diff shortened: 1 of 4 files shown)


enterprise/views.py
```diff
@@ -833,7 +861,15 @@ def post(self, request):
         defer_creation = request.POST.get('defer_creation')
         consent_provided = bool(request.POST.get('data_sharing_consent', False))
         if defer_creation is None and consent_record.consent_required():
-            if course_id:
+
+            # Skip creating EnterpriseCourseEnrollment record if we found course_key instead of course_run_id
```
HammadAhmadWaqas, line 865: this comment is sounded a little confusing at first. can we change starting like that?
```
Creating EnterpriseCourseEnrollment record if we found course_run_id instead of course_key in the course_id param
```
something like that.

enterprise/views.py
```diff
@@ -416,6 +418,26 @@ def get_context_from_db(self, consent_page, platform_name, item, context):
...
+        Returns:
+            (Boolean): True or False
+        """
+        try:
+            # Check if we have a course ID or a course run ID
+            CourseKey.from_string(course_id)
+        except InvalidKeyError:
+            # The ID we have is for a course instead of a course run
+            return False
+        else:
```
HammadAhmadWaqas, line 437: no need to put else after the return statement.
```
except InvalidKeyError:
            # The ID we have is for a course instead of a course run
            return False
# If here, the course_id is a course_run_id
return True
```

PR #801, opened 2020-06-15
ENH: Add validation for content filter

taliaedX:
This adds validation for the content_filter field's subfields (filter keys) in both EnterpriseCatalogQuery and EnterpriseCustomerCatalog.

**WIP** - Tests and the rest of this checklist forthcoming.

This is for ticket [ENT-2828](https://openedx.atlassian.net/browse/ENT-2828) 

**Merge checklist:**
- [X] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files) **_N/A_**
- [X] `make static` has been run to update webpack bundling if any static content was updated **_N/A_**
- [ ] Translations updated

**Post merge:**

### DIFF
--- a/enterprise/validators.py
+++ b/enterprise/validators.py
@@ -13,6 +13,7 @@ from django.utils.translation import ugettext_lazy as _
 
 from enterprise.constants import CONTENT_FILTER_FIELD_TYPES
 
+
 def get_app_config():
     """
     :return Application configuration.
@@ -52,6 +53,9 @@ def validate_image_size(image):
 
 
 def validate_content_filter_fields(content_filter):
+    """
+    Validate particular fields (if present) passed in through content_filter are certain types.
+    """
     for key in CONTENT_FILTER_FIELD_TYPES:
         if key in content_filter.keys():
             if not isinstance(content_filter[key], CONTENT_FILTER_FIELD_TYPES[key]['type']):
--- a/enterprise/validators.py
+++ b/enterprise/validators.py
@@ -11,6 +11,7 @@ from django.apps import apps
 from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext_lazy as _
 
+from enterprise.constants import CONTENT_FILTER_FIELD_TYPES
 
 def get_app_config():
     """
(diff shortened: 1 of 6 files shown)


enterprise/validators.py
```diff
@@ -48,3 +50,22 @@ def validate_image_size(image):
...
             _("The logo image file size must be less than or equal to %s KB.") % config.valid_max_image_size)
+
+
+def validate_content_filter_fields(content_filter):
+    """
+    Validate particular fields (if present) passed in through content_filter are certain types.
+    """
+    for key in CONTENT_FILTER_FIELD_TYPES:
+        if key in content_filter.keys():
+            if not isinstance(content_filter[key], CONTENT_FILTER_FIELD_TYPES[key]['type']):
```
binodpant, line 61: minor point: could we store `CONTENT_FILTER_FIELD_TYPES[key]` as a constant in this scope (or if block) given it's used a few times ?

binodpant:
no flags from me. Are there any manual testing steps here to verify that this change will be safe to production ship? I am completely uneducated in this side of things, so just curious .. the automated test looks solid ... other than that this is LGTM! I would get some other more experience reviews too, but otherwise I am good! 

I presume we can check off the 'Translation updated' piece now?

enterprise/validators.py
taliaedX: oh you mean to just to shorten it? like `cf_key = CONTENT_FILTER_FIELD_TYPES[key]` ?

taliaedX:
> no flags from me. Are there any manual testing steps here to verify that this change will be safe to production ship? I am completely uneducated in this side of things, so just curious .. the automated test looks solid ... other than that this is LGTM! I would get some other more experience reviews too, but otherwise I am good!
> 
> I presume we can check off the 'Translation updated' piece now?

@binodpant I don't know if manual testing is "necessary", but I was planning on going into the admin on staging (`courses-internal.stage.edx.org/admin`) and modifying a course catalog's content_filter field as well as an EnterpriseCatalogQuery.

binodpant:
also @taliaedX sorry about this timing I just merged in a change that changes changelog as well as the version 
https://github.com/edx/edx-enterprise/pull/803/files#diff-de0a9b67ffe2874a076b5fdb15bad484R17

so you may have to merge/rebase and update

enterprise/validators.py
taliaedX: I went with shortening the import. The extra assignment isn't really costly, but might as well just make the whole constant as we use it shorter.

enterprise/validators.py
```diff
@@ -48,3 +50,22 @@ def validate_image_size(image):
...
+    Validate particular fields (if present) passed in through content_filter are certain types.
+    """
+    for key in cftypes:
+        if key in content_filter.keys():
+            if not isinstance(content_filter[key], cftypes[key]['type']):
+                raise ValidationError(
+                    "Content filter '%s' must be of type %s" % (key, cftypes[key]['type'])
+                )
+            if cftypes[key]['type'] == list:
+                if not all(isinstance(x, str) for x in content_filter[key]):
```
iloveagent57, line 66: I think you want to use that `subtype` value here, instead of always insisting that it's a string (not that there's any functional difference right now).

taliaedX: ah good point! Will update that.

PR #803, opened 2020-06-17
Enrollments endpoint updates to optimize client side frontend learner portal

binodpant:
This change allows the client learner portal to simply use a boolean field true instead of a string based field, it does come at the cost of an extra string compare in backend but it used to be implicit now I explicitly check for case insensitive 'true' so works for both current and new usages 

## Ephemeral change: Process marked_done as a string 'true' or 'false' which matches the boolean type to be used in client side code.
This change will allow the client side code to be optimized to use boolean for marked_done rather than a string.
This is incremental and in near future we will update it to use patch body and process this field as boolean. right now it gets a string 'true' or 'false' due to string based payload. See https://github.com/edx/frontend-app-learner-portal-enterprise/pull/27

Fixes: ENT-3047

## Blocks
https://github.com/edx/frontend-app-learner-portal-enterprise/pull/27

**Merge checklist:**
- [x] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files) : N/A
- [X] `make static` has been run to update webpack bundling if any static content was updated N/A
- [X] Translations updated N/A

**Post merge:**

### DIFF
--- a/enterprise_learner_portal/api/v1/views.py
+++ b/enterprise_learner_portal/api/v1/views.py
@@ -100,7 +100,7 @@ class EnterpriseCourseEnrollmentView(APIView):
             course_id=course_id
         )
 
-        # TODO: For now, this makes the change backward compatible, we will change this to true boolean support 
+        # TODO: For now, this makes the change backward compatible, we will change this to true boolean support
         enterprise_enrollment.marked_done = marked_done.lower() == 'true'
         enterprise_enrollment.save()
 
--- a/enterprise_learner_portal/api/v1/views.py
+++ b/enterprise_learner_portal/api/v1/views.py
@@ -100,7 +100,8 @@ class EnterpriseCourseEnrollmentView(APIView):
             course_id=course_id
         )
 
-        enterprise_enrollment.marked_done = marked_done
+        # TODO: change client to use patch body and then process this as proper boolean
+        enterprise_enrollment.marked_done = marked_done == 'true'
         enterprise_enrollment.save()
 
         course_overviews = get_course_overviews([course_id])
--- a/enterprise_learner_portal/api/v1/views.py
+++ b/enterprise_learner_portal/api/v1/views.py
@@ -100,8 +100,8 @@ class EnterpriseCourseEnrollmentView(APIView):
             course_id=course_id
         )
 
-        # TODO: change client to use patch body and then process this as proper boolean
-        enterprise_enrollment.marked_done = marked_done == 'true'
+        # TODO: For now, this makes the change backward compatible, we will change this to true boolean support 
+        enterprise_enrollment.marked_done = marked_done.lower() == 'true'
         enterprise_enrollment.save()
 
         course_overviews = get_course_overviews([course_id])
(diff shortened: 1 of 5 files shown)


binodpant:
on second thought, I am pulling out the changing of course status piece here since that breaks other places in the code.. I want to discuss with Adam before that change is made, right now a simple hack on the ui addresses that ...

binodpant:
on second thought, I am pulling out the changing of course status piece here since that breaks other places in the code.. I want to discuss with Adam before that change is made, right now a simple hack on the ui addresses that ...

binodpant:
also made the change in views.py backward compatible just in case someone calls with True instead of true ...

tests/test_enterprise_learner_portal/api/test_views.py
```diff
@@ -190,7 +190,7 @@ def test_patch_missing_params(self, mock_get_overviews, mock_serializer):
         query_params_full = {
             'enterprise_id': str(self.enterprise_customer.uuid),
             'course_id': self.course_run_id,
-            'marked_done': True,
```
binodpant, line 193: did not change the test, but verified that both boolean True and string 'true' here work

enterprise_learner_portal/utils.py
```diff
@@ -43,7 +43,9 @@ def get_course_run_status(course_overview, certificate_info, enterprise_enrollme
     is_certificate_passing = certificate_info.get('is_passing', False)
     certificate_creation_date = certificate_info.get('created', datetime.max)
 
-    if enterprise_enrollment and enterprise_enrollment.marked_done:
+    if enterprise_enrollment and enterprise_enrollment.marked_done is False:
```
binodpant: I am going to pull this change out since it breaks other areas. I think we may need to rethink the course status returned to the portal rather than change this logic ...

enterprise_learner_portal/api/v1/views.py
```diff
@@ -100,7 +100,8 @@ def patch(self, request):
             course_id=course_id
         )
 
-        enterprise_enrollment.marked_done = marked_done
+        # TODO: For now, this makes the change backward compatible, we will change this to true boolean support
+        enterprise_enrollment.marked_done = marked_done.lower() == 'true'
```
jlajoie, line 104: For context the problem is this was never resolving to be a boolean, or was it just never a boolean in the case of learner-portal's requests to this endpoint.

For backwards compatibility, we'd probably want to do some typechecking here instead after we do the `.get()` for `marked_done`. If we get a boolean back, great, let's keep it, if we get a string back, do the `.lower() == 'true'` check.

binodpant: I am pretty sure we always got a string here, previously we got a 'True' or 'False' and now we get 'true' or 'false' because of a change inthe frontend to send boolean fields which get turned into strings 'true' or 'false' (since a text/plain is used here when sedning from client) ... but the field being assigned to was a BooleanField so I think the first case just worked while the latter needed explicit handling ... I am ~95% on that

binodpant: also, soon this will change to be all boolean on both client and server side if I do this right ...

binodpant: and just to be sure I tested by interfering with the payload by changing the options['marked_done'] to 'True' or 'False' instead of 'true' or 'false' and both seem to work correctly...

PR #804, opened 2020-06-18
ENT-2657: Added health checks for enterprise service.

saleem-latif:
**Merge checklist:**
- [x] Translations updated
- [x] https://github.com/edx/edx-platform/pull/24248 is merged

**Post merge:**

### DIFF
--- a/enterprise/api_client/discovery.py
+++ b/enterprise/api_client/discovery.py
@@ -439,3 +439,27 @@ class CourseCatalogApiServiceClient(CourseCatalogApiClient):
             return bool(cls().get_program_by_uuid(program_uuid))
         except ImproperlyConfigured:
             return False
+
+
+class NoAuthDiscoveryClient:
+    """
+    Class to build a course discovery client to make calls to the discovery service.
+    """
+
+    API_BASE_URL = settings.COURSE_CATALOG_URL_ROOT
+    APPEND_SLASH = False
+
+    def __init__(self):
+        """
+        Create a course discovery client.
+        """
+        self.client = EdxRestApiClient(self.API_BASE_URL, append_slash=self.APPEND_SLASH)
+
+    def get_health(self):
+        """
+        Retrieve health details for course discovery service.
+
+        Returns:
+            dict: Response containing course discovery service health.
+        """
+        return self.client.health.get()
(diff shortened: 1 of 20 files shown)


enterprise/api_client/discovery.py
```diff
@@ -418,3 +418,27 @@ def program_exists(cls, program_uuid):
...
+    """
+    Class to build a course discovery client to make calls to the discovery service.
+    """
+
+    API_BASE_URL = settings.COURSE_CATALOG_URL_ROOT
+    APPEND_SLASH = False
+
+    def __init__(self):
+        """
+        Create an E-Commerce client.
```
HammadAhmadWaqas, line 433: Do we need to update comment?

enterprise/api_client/lms.py
```diff
@@ -490,6 +490,24 @@ def get_course_certificate(self, course_id, username):
         return self.client.certificates(username).courses(course_id).get()
 
 
+class LMSClient(NoAuthenticationLmsApiClient):
```
HammadAhmadWaqas, line 493: I think name of this Client class should be LMSHealthClient or something like that, otherwise, this LMS client could be used mistakenly with other endpoints (That will be not a good thing as its a NO Authentication Api client)

enterprise/api_client/ecommerce.py
```diff
@@ -105,3 +108,27 @@ def create_manual_enrollment_orders(self, enrollments):
                 "Failed to created orders for the following manual enrollments. %s",
                 failed_creations
             )
+
+
+class EcommerceClient:
```
HammadAhmadWaqas, line 113: should we mention that is also no authentication client in the class name?

enterprise/api_client/enterprise_catalog.py
```diff
@@ -193,3 +194,27 @@ def enterprise_contains_content_items(self, enterprise_uuid, content_ids):
         query_params = {'course_run_ids': content_ids}
         endpoint = getattr(self.client, self.ENTERPRISE_CUSTOMER_ENDPOINT)(enterprise_uuid)
         return endpoint.contains_content_items.get(**query_params)['contains_content_items']
+
+
+class EnterpriseCatalogClient:
```
HammadAhmadWaqas, line 199: same as above?

enterprise/api_client/lms.py
saleem-latif: I have renamed all the client and added `NoAuth` in front of each client to clarify.

PR #808, opened 2020-06-19
BUG: Migration required for content_filter changes.

taliaedX:
**Merge checklist:**

**Post merge:**


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,6 +4,11 @@
     - `test-master.in` if edx-platform pins it, with a matching version
     - `make upgrade && make requirements` have been run to regenerate requirements
 - [ ] `make static` has been run to update webpack bundling if any static content was updated
+- [ ] `./manage.py makemigrations` has been run
```
georgebabey, line 7: do you want to include a link to https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations ?

That wiki page includes some tips, namely when removing models and how to do it in a 2-step process so you don't break during the blue/green releases

PR #813, opened 2020-06-26
Discovery API Client: Fix invalid key references

mattdrayer:
**Merge checklist:**

**Post merge:**


mattdrayer:
Had to fix the key references because they were invalid and are causing the xAPi transmission job to fail. I'm not sure why this wasn't picked up by the test framework in the original PR but I've added a set of specific tests to cover the new function/method so it should be okay going forward.

PR #814, opened 2020-06-29
Add saved_for_later field

raq929:
This will eventually replace the marked_done field (this is step 1 from https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations#EverythingAboutDatabaseMigrations-Howtorenameacolumn)

**Merge checklist:**
- [x] [Changelog](https://github.com/edx/edx-enterprise/blob/master/CHANGELOG.rst) record added _I'm wondering whether to update this now, or after the name change is complete_

**Post merge:**


binodpant:
Looking good to me 

few things to address

- [ ] lint fix:
```enterprise_learner_portal/api/v1/views.py:85:8: W0612: Unused variable 'saved_for_later' (unused-variable)
```
- [ ] Do you mind tasking the removal of the older field as a task in the backlog? We can always do it later just to ensure we keep track of that, thanks!

- [ ] I believe Changelog should be updated pre-merge as well as part of this PR

binodpant:
Also because I have not priorly reviewed a migration PR in this repo maybe also include someone from Titans? or ping them for anything I may have missed? It is looking fine to me though since we are adding a new field, not changing an existing one so cannot have any side effects that I can tell

raq929:
Agreed on the naming. I saw that most of the others were auto-named so I wasn't sure what our policy was, but I also like naming them so will do that in the future. Thanks!

PR #815, opened 2020-06-30
create LicensedEnterpriseCourseEnrollment model.

kellmarie:
ENT-3073

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -1365,6 +1365,50 @@ def __repr__(self):
...
+
+    history = HistoricalRecords()
+
+    # Does there need to be a source (?)
+    source = models.ForeignKey(
+        EnterpriseEnrollmentSource,
+        blank=False,
+        null=True,
+        on_delete=models.SET_NULL
+    )
```
iloveagent57, line 1396: Nope, you don't need this field on this model.

enterprise/models.py
```diff
@@ -1365,6 +1365,50 @@ def __repr__(self):
...
+    license_uuid = models.UUIDField(
+        primary_key=False,
+        editable=False,
+        null=False
+    )
+
+    enterprise_course_enrollment = models.OneToOneField(
+        EnterpriseCourseEnrollment,
+        blank=False,
+        null=False,
```
iloveagent57, line 1380: This should be `null=True` - not every `EnterpriseCourseEnrollment` will have a related `LicensedEnterpriseCourseEnrollment`.

enterprise/models.py
```diff
@@ -1365,6 +1365,50 @@ def __repr__(self):
...
+        on_delete=models.SET_NULL
+    )
+
+    # The ticket said this would be nice to have for returning an associated license if one exists
+    # In what cases will we have a licensed enrollment but there not being an assocaited license?
+    @property
+    def associated_license(self):
+        """
+        Returns the license associated with this enterprise course enrollment if one exists
+        """
```
iloveagent57, line 1404: Right, you actually want to add this on the `EnterpriseCourseEnrollment` model.  Because you're adding a `OneToOneField` from this model back to `EnterpriseCourseEnrollment`, `EnterpriseCourseEnrollment` will now have an attribute named whatever the `related_name` is of the `OneToOneField` (`licensed_with`, in this case).

There's an example in the docs: https://docs.djangoproject.com/en/2.2/topics/db/examples/one_to_one/ about why this property would be useful:
```
> from django.core.exceptions import ObjectDoesNotExist
>>> try:
>>>     p2.restaurant
>>> except ObjectDoesNotExist:
>>>     print("There is no restaurant here.")
There is no restaurant here.
```

enterprise/models.py
kellmarie: Oh I thought this would have done the inverse where every `LicensedEnterpriseCourseEnrollment` has n `EnterpriseCourseEnrollment`

enterprise/models.py
```diff
@@ -1366,6 +1377,34 @@ def __repr__(self):
         return self.__str__()
 
 
+@python_2_unicode_compatible
```
iloveagent57, line 1380: Good news!  We don't support python 2 anymore, so you don't need this decorator.

enterprise/models.py
```diff
@@ -1366,6 +1377,34 @@ def __repr__(self):
...
+    license_uuid = models.UUIDField(
+        primary_key=False,
+        editable=False,
+        null=False
+    )
+
+    enterprise_course_enrollment = models.OneToOneField(
+        EnterpriseCourseEnrollment,
+        blank=False,
+        null=True,
```
iloveagent57, line 1397: I think we want `null=False` here, too.  You shouldn't be able to create one of these without associating it with an existing `EnterpriseCourseEnrollment`.

enterprise/models.py
kellmarie: Oh nice, I wasn't sure if we were building things to be backwards compatible or not. Good to know we aren't :D

PR #818, opened 2020-07-02
Make marked_done nullable

raq929:
So tests in phase 2 of renaming the column can pass.

(I should have done this in my previous PR.)

**Merge checklist:**

**Post merge:**


enterprise/migrations/0099_auto_20200702_1537.py
```diff
@@ -0,0 +1,23 @@
...
+        ('enterprise', '0098_auto_20200629_1756'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='enterprisecourseenrollment',
+            name='marked_done',
+            field=models.NullBooleanField(default=False, help_text='Specifies whether a user marked this course as completed in the learner portal.'),
+        ),
+        migrations.AlterField(
```
binodpant, line 18: do we need a models.py change for this one too or is it already covered?

raq929: The `historicalenterprisecourseenrollement` change is automatically created when `enterprisecourseenrollement` changes, so no additional model was changed. :)

PR #822, opened 2020-07-08
Remove references to saved_for_later in the database

raq929:
Retains marked_done for communication with the front end which may still have the field

**Merge checklist:**

**Post merge:**


enterprise/migrations/0100_move_data_to_saved_for_later.py
```diff
@@ -0,0 +1,24 @@
...
+def move_data_from_marked_done_to_saved_for_later(apps, schema_editor):
+    EnterpriseCourseEnrollment = apps.get_model('enterprise', 'EnterpriseCourseEnrollment')
+    for course_enrollment in EnterpriseCourseEnrollment.objects.all():
+        course_enrollment.saved_for_later = course_enrollment.marked_done
+
+
+def move_data_from_saved_for_later_to_marked_done(apps, schema_editor):
+    EnterpriseCourseEnrollment = apps.get_model('enterprise', 'EnterpriseCourseEnrollment')
+    for course_enrollment in EnterpriseCourseEnrollment.objects.all():
+        course_enrollment.marked_done = course_enrollment.saved_for_later
```
bbaker6225, line 14: I think you'll need to save the course_enrollment models after you edit them (for efficiency probably best to use https://docs.djangoproject.com/en/2.2/ref/models/querysets/#bulk-update)

enterprise_learner_portal/api/v1/serializers.py
```diff
@@ -70,8 +70,8 @@ def to_representation(self, instance):
         representation['due_dates'] = get_due_dates(request, course_run_id, user)
         representation['pacing'] = course_overview['pacing']
         representation['org_name'] = course_overview['display_org_with_default']
-
-        representation['marked_done'] = instance.marked_done
+        representation['marked_done'] = instance.saved_for_later
```
bbaker6225, line 73: Do we need to keep this old field around? If it's still being used in the API might be worth adding a comment about it being deprecated

bbaker6225:
I think you'll also need to remove the old field from the model (but not remove it in a migration yet)

enterprise_learner_portal/api/v1/serializers.py
raq929: It's still used by the frontend. The plan is to have a brief downtime where we switch the frontend over, then remove this in step 3.

raq929:
Yes. I also have to update the migration; I rebased, and missed the fact that there are now two 0100 migrations. Fixing now.

enterprise/migrations/0101_move_data_to_saved_for_later.py
```diff
@@ -0,0 +1,30 @@
...
+    course_enrollments = EnterpriseCourseEnrollment.objects.all()
+    for course_enrollment in course_enrollments:
+        course_enrollment.saved_for_later = course_enrollment.marked_done
+    EnterpriseCourseEnrollment.objects.bulk_update(course_enrollments, ['saved_for_later'])
+
+
+def move_data_from_saved_for_later_to_marked_done(apps, schema_editor):
+    EnterpriseCourseEnrollment = apps.get_model('enterprise', 'EnterpriseCourseEnrollment')
+    course_enrollments = EnterpriseCourseEnrollment.objects.all()
+    for course_enrollment in EnterpriseCourseEnrollment.objects.all():
```
bbaker6225, line 18:
```python
for course_enrollment in course_enrollments
```
so you don't call the same thing twice?

raq929: yeah, good call, I did it in the function above.

PR #823, opened 2020-07-08
Admin dashboard rules predicates now pass an object into the edx-rbac utility functions

iloveagent57:
We have other mechanisms for restricting the set of objects visible to an enterprise-linked user; for example, on the enterprise catalog endpoint, there is a filter that only returns results of enterprises the user is linked to, and returns everything if they're staff.  This change should help us not have to worry about doing that kind of filtering in other views that require the `can_access_admin_dashboard` permission.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/rules.py
+++ b/enterprise/rules.py
@@ -22,18 +22,26 @@ def has_implicit_access_to_dashboard(user, obj):  # pylint: disable=unused-argum
     """
     Check that if request user has implicit access to `ENTERPRISE_DASHBOARD_ADMIN_ROLE` feature role.
 
+    Params:
+        user: An ``auth.User`` instance.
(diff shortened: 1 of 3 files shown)


enterprise/rules.py
```diff
@@ -22,26 +22,35 @@ def has_implicit_access_to_dashboard(user, obj):  # pylint: disable=unused-argum
     """
```
adamstankiewicz, line 22: nit: is the `# pylint: disable=unused-argument` comment relevant here now that we're using `obj`? Similar to the other functions, as well.

iloveagent57: haha, actually, `user` is still unused.  It does apply to the `...explicit_access...` one, though.

PR #824, opened 2020-07-09
ENH: Add portal saml config checkbox.

taliaedX:
Adds a new option to the EnterpriseCustomer object that will be used to enable the SAML Configuration in the admin portal front end.
This is needed/used by features in this PR: https://github.com/edx/frontend-app-admin-portal/pull/288

**Merge checklist:**
- [X] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files)
- [X] `make static` has been run to update webpack bundling if any static content was updated
- [X] `./manage.py makemigrations` has been run

**Post merge:**


mattdrayer:
It probably makes sense to, at some point in the near future, create a new enterprise admin portal configuration model that encompasses these flags, rather than tacking them on to the EnterpriseCustomer model.

georgebabey:
> It probably makes sense to, at some point in the near future, create a new enterprise admin portal configuration model that encompasses these flags, rather than tacking them on to the EnterpriseCustomer model.

I think at some point we should remove these flags entirely and just have features on for all when they roll out rather than having to manually configure each, unless there is some upsell opportunity we're chasing

mattdrayer:
> > It probably makes sense to, at some point in the near future, create a new enterprise admin portal configuration model that encompasses these flags, rather than tacking them on to the EnterpriseCustomer model.
> 
> I think at some point we should remove these flags entirely and just have features on for all when they roll out rather than having to manually configure each, unless there is some upsell opportunity we're chasing

If you give sales something to sell, they will sell it 😄

PR #825, opened 2020-07-10
ENT-3078: DSC enrollment with license_uuid

chavesj:
**Merge checklist:**

**Post merge:**


enterprise/api_client/ecommerce.py
```diff
@@ -81,30 +81,30 @@ def create_manual_enrollment_orders(self, enrollments):
         Since `student.CourseEnrollment` lives in LMS, we're just passing around dicts of the relevant information.
         """
         try:
-            order_repsonse = self.client.manual_course_enrollment_order.post(
+            order_response = self.client.manual_course_enrollment_order.post(
                 {
                     "enrollments": enrollments
                 }
             )
+            order_creations = order_response["orders"]
```
chavesj, line 89: Prevents a "Local variable referenced before assignment" error from being thrown

CHANGELOG.rst
```diff
@@ -14,6 +14,11 @@ Change Log
 Unreleased
 --------------------
 
+[3.3.23] 2020-07-14
```
iloveagent57, line 17: I think there's a newer commit on master already for 3.3.23, FYI.

enterprise/views.py
```diff
@@ -867,10 +890,21 @@ def post(self, request):
...
+                        enrollment_api_client = EnrollmentApiClient()
+                        # Ensure the 'verified' track exists for the given course
+                        # If not, enroll in the 'audit' track
+                        course_modes = enrollment_api_client.get_course_modes(course_id)
+                        course_mode = 'verified' if 'verified' in course_modes else 'audit'
+                        enrollment_api_client.enroll_user_in_course(
+                            request.user.username,
+                            course_id,
+                            course_mode
+                        )
```
iloveagent57, line 907: Nice

enterprise/api_client/ecommerce.py
bbaker6225: 🙏

enterprise/views.py
```diff
@@ -867,10 +890,21 @@ def post(self, request):
...
                 try:
-                    self.create_enterprise_course_enrollment(request, consent_record, course_id)
+                    self.create_enterprise_course_enrollment(request, consent_record, course_id, license_uuid)
+                    if license_uuid:
+                        # Use enrollment api client to create an lms course enrollment
+                        enrollment_api_client = EnrollmentApiClient()
+                        # Ensure the 'verified' track exists for the given course
+                        # If not, enroll in the 'audit' track
+                        course_modes = enrollment_api_client.get_course_modes(course_id)
+                        course_mode = 'verified' if 'verified' in course_modes else 'audit'
```
bbaker6225, line 902: Just checking my understanding here, this means https://github.com/edx/frontend-app-learner-portal-enterprise/pull/58 won't be needed right?

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -401,29 +416,42 @@ def test_post_course_specific_consent(
...
+        enterprise_enrollment = EnterpriseCourseEnrollment.objects.create(
             enterprise_customer_user=ecu,
             course_id=course_id,
         )
         dsc = DataSharingConsentFactory(
             username=self.user.username,
             course_id=course_id,
             enterprise_customer=enterprise_customer,
-            granted=consent_provided
+            granted=False
```
bbaker6225, line 440: Was this change intentional?

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -435,8 +463,22 @@ def test_post_course_specific_consent(
...
         if not defer_creation:
+            dsc.refresh_from_db()
             assert dsc.granted is consent_provided
 
+        # we'll only create an enrollment record if (1) creation is not deferred,
+        # (2) consent is provided, and (3) we provide a course _run_ id
+        if (not defer_creation) and consent_provided and course_id.endswith('Demo_Course'):
+            mock_licensed_ent_course_enrollment_create.assert_called_once_with(
+                enterprise_course_enrollment=enterprise_enrollment,
+                license_uuid=license_uuid
```
bbaker6225, line 474: This feels a little weird to me? I'd expect we'd just check that a `LicensedEnterpriseCourseEnrollment` exists in the db?
Something like 
```python
assert LicensedEnterpriseCourseEnrollment.objects.get(
    enterprise_course_enrollment=enterprise_enrollment,
    license_uuid=license_uuid,
).exists() is True
```
maybe?

enterprise/views.py
chavesj: correct!

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
chavesj: good catch, this was a leftover debug change

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
chavesj: agreed 👍

enterprise/views.py
```diff
@@ -861,16 +884,30 @@ def post(self, request):
         consent_provided = bool(request.POST.get('data_sharing_consent', False))
         if defer_creation is None and consent_record.consent_required():
 
+            # WARNING: If consent is required for this course and we’re not deferring creation, we’ll create an
+            # enrollment, even if the request specifies that data_sharing_consent is False
```
iloveagent57, line 888: bah, tiny nit: "assuming that the given course_id is a course _run_ id"

enterprise/views.py
```diff
@@ -861,16 +884,30 @@ def post(self, request):
...
-
             if course_id and self.is_course_run_id(course_id):
                 try:
-                    self.create_enterprise_course_enrollment(request, consent_record, course_id)
+                    self.create_enterprise_course_enrollment(request, consent_record, course_id, license_uuid)
+                    if license_uuid:
+                        # Use enrollment api client to create an lms course enrollment
+                        enrollment_api_client = EnrollmentApiClient()
+                        # Ensure the 'verified' track exists for the given course
+                        # If not, enroll in the 'audit' track
```
iloveagent57, line 903: I think the code below is so clear that you don't need this comment.

enterprise/views.py
chavesj: I'll move this comment under the `if course_id and self.is_course_run_id(course_id):` condition to give the context that this only applies when `course_id` is for a course run

chavesj: update - I ended up addressing this issue directly rather than including a comment to warn of this edge case

enterprise/api_client/ecommerce.py
brittneyexline: i'm assuming this change is unrelated functionally to the rest of the changes? just want to confirm :-)

enterprise/views.py
```diff
@@ -859,18 +882,25 @@ def post(self, request):
...
             # an EnterpriseCourseEnrollment record will also get created.
-
             if course_id and self.is_course_run_id(course_id):
                 try:
-                    self.create_enterprise_course_enrollment(request, consent_record, course_id)
+                    self.create_enterprise_course_enrollment(request, consent_record, course_id, license_uuid)
+                    if license_uuid:
+                        enrollment_api_client = EnrollmentApiClient()
+                        course_modes = enrollment_api_client.get_course_modes(course_id)
+                        course_mode = 'verified' if 'verified' in course_modes else 'audit'
```
brittneyexline, line 898: verified isn't the only paid course mode... we should probably check for professional as well.

enterprise/views.py
```diff
@@ -859,18 +882,25 @@ def post(self, request):
...
+                        enrollment_api_client.enroll_user_in_course(
```
brittneyexline, line 899: we might want to catch failures here, if the LMS enrollment fails perhaps we should be redirecting to the failure URL with a notification to try again? also is it possible that this enrollment call is made somewhere else in this flow where we're already handling potential error cases? if so, we may be able to consolidate or simplify.

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -429,14 +455,34 @@ def test_post_course_specific_consent(
...
 
         assert resp.url.endswith(expected_redirect_url)  # pylint: disable=no-member
         assert resp.status_code == 302
         if not defer_creation:
+            dsc.refresh_from_db()
             assert dsc.granted is consent_provided
 
+        # we'll only create an enrollment record if (1) creation is not deferred, (2) consent has not been provided,
+        # (3) we provide a license_uuid, and (4) we provide a course _run_ id
+        if (not defer_creation) and (not consent_provided) and license_uuid and course_id.endswith('Demo_Course'):
```
brittneyexline, line 471: style nit: do you need the parens around the `not` statements here?

enterprise/api_client/ecommerce.py
chavesj: yeah, just preventing an error I encountered in local testing

enterprise/views.py
iloveagent57: Can a course have both 'verified' and 'professional' modes?

brittneyexline: i think in practice no, but i doubt there is a formal constraint. if for some reason both exist, we should probably default to 'verified'.

chavesj: I also found the `no-id-professional` course mode. Any input on the `credit` and `honor` course modes?

brittneyexline: you shouldn't have to worry about either of those, honor is basically the same as audit and credit is complicated, so it's easiest to not deal with it for now.

iloveagent57: I would use these as the reference point for "modes that allow you to buy stuff": https://github.com/edx/edx-platform/blob/master/common/djangoapps/course_modes/models.py#L171-L172

enterprise/views.py
chavesj: it seems like for other use cases the LMS enrollment happens outside of the DSC flow, so I this should be fine living right here. however the LMS enrollment should happen before the creation of an EnterpriseCourseEnrollment so I'll be switching the order of operations here and redirecting to the failure_url in the case that the LMS enrollment fails

enterprise/views.py
chavesj: okay, thanks!

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -372,17 +383,19 @@ def test_get_course_specific_consent_not_needed(
...
-        (False, False, '/failure_url', 'edX+DemoX'),
+        (True, True, '/successful_enrollment', 'course-v1:edX+DemoX+Demo_Course', ''),
+        (False, True, '/successful_enrollment', 'course-v1:edX+DemoX+Demo_Course', str(uuid.uuid4())),
+        (False, True, '/successful_enrollment', 'course-v1:edX+DemoX+Demo_Course', ''),
+        (True, False, '/failure_url', 'course-v1:edX+DemoX+Demo_Course', ''),
+        (False, False, '/failure_url', 'course-v1:edX+DemoX+Demo_Course', ''),
+        (True, True, '/successful_enrollment', 'edX+DemoX', ''),
+        (False, True, '/successful_enrollment', 'edX+DemoX', ''),
+        (True, False, '/failure_url', 'edX+DemoX', ''),
+        (False, False, '/failure_url', 'edX+DemoX', ''),
```
iloveagent57, line 398: Total nit that you can ignore here but might find useful in the future - this style:
```python
{
    'defer_creation': True, 'consent_provided': False, 'expected_redirect_url': '/successful_enrollment', 
    'course_id': 'course-v1:edX+DemoX+Demo_Course', 'license_uuid': ''
},
...
```
...while more verbose, does make it easier to understand at a glance what each set of data is actually testing.  And it works with `ddt.unpack` the same as tuples, so you won't have to change the signature of this test function.

chavesj: Agreed, since these tests existed before this PR I'm going to leave as is but I'll definitely be sure to use this syntax in the future.

PR #832, opened 2020-07-15
ENT-3139 New viewset/endpoint to create PendingEnterpriseCustomerUser records

brittneyexline:
This viewset has some special logic to ensure that if the endpoint is called with data that already exists as either a PendingEnterpriseCustomerUser or EnterpriseCustomerUser, we return a 204 No Content. Without this, DRF would just return 400.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/tests/test_enterprise/api/test_views.py
+++ b/tests/test_enterprise/api/test_views.py
@@ -97,6 +97,7 @@ ENTERPRISE_CUSTOMER_COURSE_ENROLLMENTS_ENDPOINT = reverse('enterprise-customer-c
 ENTERPRISE_CUSTOMER_REPORTING_ENDPOINT = reverse('enterprise-customer-reporting-list')
 ENTERPRISE_LEARNER_LIST_ENDPOINT = reverse('enterprise-learner-list')
 ENTERPRISE_CUSTOMER_WITH_ACCESS_TO_ENDPOINT = reverse('enterprise-customer-with-access-to')
+PENDING_ENTERPRISE_LEARNER_LIST_ENDPOINT = reverse('pending-enterprise-learner-list')
 
 
 def side_effect(url, query_parameters):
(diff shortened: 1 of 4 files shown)


tests/test_enterprise/api/test_views.py
```diff
@@ -420,6 +421,68 @@ def test_post_enterprise_customer_user_logged_out(self):
         response = self.client.post(settings.TEST_SERVER + ENTERPRISE_LEARNER_LIST_ENDPOINT, data=data)
         assert response.status_code == 401
 
+    @ddt.data(
+        (True, True, True, False, 204),
+        (True, True, False, True, 204),
+        (True, False, False, True, 204),
+        (True, False, False, False, 201),
+        (True, True, False, False, 201),
+        (False, False, False, False, 403)
```
iloveagent57, line 430: Nit: you can write these as
```python
{'is_staff': True, 'user_exists': True, 'ecu_exists': True, 'pending_ecu_exists': False, 'status_code': 204},
...
```
and @ddt.unpack will still unpack everything into the function parameters you have.  Definitely more verbose, but way easier to understand at a glance.

tests/test_enterprise/api/test_views.py
```diff
@@ -420,6 +421,68 @@ def test_post_enterprise_customer_user_logged_out(self):
...
+            user = factories.UserFactory(email=new_user_email)
+            if ecu_exists:
+                factories.EnterpriseCustomerUserFactory(user_id=user.id, enterprise_customer=enterprise_customer)
+
+        if pending_ecu_exists:
+            factories.PendingEnterpriseCustomerUserFactory(user_email=new_user_email, enterprise_customer=enterprise_customer)
+
+        response = self.client.post(settings.TEST_SERVER + PENDING_ENTERPRISE_LEARNER_LIST_ENDPOINT, data=data)
+        assert response.status_code == status_code
+        #assert False
```
iloveagent57, line 464: Nit: remove

tests/test_enterprise/api/test_views.py
```diff
@@ -420,6 +422,69 @@ def test_post_enterprise_customer_user_logged_out(self):
...
+        client_username = 'client_username'
+        self.client.logout()
+        self.create_user(username=client_username, password=TEST_PASSWORD, is_staff=is_staff)
+        self.client.login(username=client_username, password=TEST_PASSWORD)
+
+        enterprise_customer = factories.EnterpriseCustomerFactory(uuid=FAKE_UUIDS[0])
+        new_user_email = 'newuser@example.com'
+        data = {
+            'enterprise_customer': FAKE_UUIDS[0],
+            'user_email': 'newuser@example.com',
```
bbaker6225, line 453: The tiniest of nits: just use `new_user_email` here?

PR #840, opened 2020-07-23
ENT-2894: Add query params to proxy login redirect for new welcome template to be rendered

chavesj:
**Merge checklist:**

**Post merge:**


chavesj:
Associated Platform PR: https://github.com/edx/edx-platform/pull/24587

enterprise/views.py
```diff
@@ -1023,6 +1023,10 @@ def get(self, request):
             # Add the tpa_hint to the redirect's query parameters
             tpa_hint = enterprise_customer.identity_provider
             query_dict['tpa_hint'] = tpa_hint
+        else:
+            # Add Enterprise customer UUID and proxy_login to the redirect's query parameters
+            query_dict['enterprise_customer'] = [str(enterprise_customer.uuid)]
+            query_dict['proxy_login'] = [True]
```
adamstankiewicz, line 1029: Out of curiosity, do the values (e.g., `str(enterprise_customer.uuid)`) need to wrapped in a list?

chavesj: this was a linter workaround I made because parse_qs (line 1012) returns a Dict[AnyStr, List[AnyStr]] . The linter didn't complain about assigning tpa_hint or next_param because they have the type Any

tests/test_enterprise/views/test_enterprise_proxy_login_view.py
```diff
@@ -63,7 +63,11 @@ def test_redirect_no_tpa(self):
         customer_without_tpa = EnterpriseCustomerFactory()
         url = self._get_url_with_params(enterprise_slug_override=customer_without_tpa.slug, use_next=False)
         response = self.client.get(url)
-        self.assertRedirects(response, LMS_LOGIN_URL, fetch_redirect_response=False)
+        query_params = QueryDict(mutable=True)
+        query_params['enterprise_customer'] = str(customer_without_tpa.uuid)
```
adamstankiewicz, line 67: would these query params also apply to the redirect for tpa below (line 72)?

enterprise/views.py
bbaker6225, line 1029: Don't we always want to add these query params?

chavesj: Yeah good catch. I assumed that for sso we'd just redirect to the tpa login screen but now that I've tested with the identity provider active I see it also goes to an edX login screen. I'll update 👍

tests/test_enterprise/views/test_enterprise_proxy_login_view.py
chavesj: Not for the way it's currently implemented but once I address Brandon's comment they will 👍

chavesj: these are only for edX logistration to render the new sidebar, not needed for sso

enterprise/views.py
chavesj: these are only for edX logistration to render the new sidebar, not needed for sso

bbaker6225: Talked with Jeff on slack and we won't need this here. Misunderstanding from a mistake in my initial proxy login PR

PR #841, opened 2020-07-27
Added flag to update catalog_filter

anaidu0:
Added a new field to the EnterpriseCustomerCatalog model to specify whether the linked EnterpriseCatalogQuery should automatically update. We take the content_filter from the newly updated EnterpriseCatalogQuery and overwrite the content_filter on the EnterpriseCustomerCatalog models that are associated with that query (done through the reverse relationship between the two models).

ENT-2993

**Merge checklist:**

**Post merge:**


enterprise/signals.py
```diff
@@ -150,6 +151,20 @@ def delete_enterprise_learner_role_assignment(sender, instance, **kwargs):     #
...
             pass
 
+@receiver(post_save, sender=EnterpriseCatalogQuery) 
+def update_enterprise_query(sender, instance, **kwargs):
+    """
+    Sync data changes from Enterprise Catalog Query to the Enterprise Customer Catalog.
+    """
+    updated_content_filter = instance.content_filter
+
+    catalogs = instance.enterprise_customer_catalogs.all()
```
anaidu0, line 161: Retrieving all of the associated catalogs through a reverse relationship and 'related_name' between an EnterpriseCustomerCatalog and EnterpriseCatalogQuery. However, this does not seem to be performing what is intended.

enterprise/__init__.py
```diff
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.4.26"
+__version__ = "3.4.29"
```
anaidu0, line 5: Will update the version numbers and changelog to what they should actually be. I had to do this to make sure that my local lms was using my updated code rather than using the cached, older version.

bbaker6225: I'd suggest going ahead and pulling in master and updating the version number from there

enterprise/migrations/0104_auto_20200723_1418.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.14 on 2020-07-23 19:18
```
bbaker6225, line 1: You should only need one migration for adding this field as we want it to default to `False` right? Also a nit that I usually like naming migration files to give a sense of what they do

tests/test_enterprise/test_signals.py
```diff
@@ -625,3 +626,61 @@ def test_update_catalog_with_existing_service_catalog(self, api_client_mock):
             enabled_course_modes=enterprise_catalog.enabled_course_modes,
             publish_audit_enrollment_urls=enterprise_catalog.publish_audit_enrollment_urls
         )
+
+    @mock.patch('enterprise.signals.EnterpriseCatalogApiClient')
+    def test_update_enterprise_query(self, api_client_mock):
```
bbaker6225, line 631: Might be worth adding a docstring to explain what this test does? Especially the part about also ensuring the changes are synced to the enterprise catalog service as that's easy to miss

enterprise/migrations/0105_auto_20200727_1433.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.14 on 2020-07-27 19:33
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('enterprise', '0104_auto_20200723_1418'),
+    ]
```
chavesj, line 10: not sure what the best practice is with migrations but given the similarity between 104 and 105 it might be cleaner to roll back 104 and regenerate the migration to reflect `default=False` for the fields rather than having two separate migrations for adding the fields and then altering them. Anyone else have opinions on this?

tests/test_enterprise/test_signals.py
```diff
@@ -625,3 +626,61 @@ def test_update_catalog_with_existing_service_catalog(self, api_client_mock):
...
+        sender = mock.Mock()  # this would be an EnterpriseCatalogQuery class
+
+        # kwargs to satisfy parameter
+        kwargs = {
+            'update_fields': None,
+            'raw': False,
+            'signal': '<django.db.models.signals.ModelSignal object at test>',
+            'using': 'default',
+            'created': False,
+        }
```
bbaker6225, line 649: nit: I'd suggest moving this to right before you call `update_enterprise_query`

tests/test_enterprise/test_signals.py
```diff
@@ -625,3 +626,61 @@ def test_update_catalog_with_existing_service_catalog(self, api_client_mock):
...
+        test_query.content_filter = content_filter_1
+        test_query.save()
+
+        enterprise_catalog_1 = EnterpriseCustomerCatalogFactory()
+        enterprise_catalog_2 = EnterpriseCustomerCatalogFactory()
+
+        enterprise_catalog_1.sync_enterprise_catalog_query = True
+        enterprise_catalog_2.sync_enterprise_catalog_query = True
+        enterprise_catalog_1.enterprise_catalog_query = test_query
+        enterprise_catalog_2.enterprise_catalog_query = test_query
```
bbaker6225, line 661: You can probably do these things in the factory calls?

tests/test_enterprise/test_signals.py
```diff
@@ -625,3 +626,61 @@ def test_update_catalog_with_existing_service_catalog(self, api_client_mock):
...
+        self.assertEqual(enterprise_catalog_1.content_filter, enterprise_catalog_2.content_filter, content_filter_2)
+
+        enterprise_catalog_1.save()
+        enterprise_catalog_2.save()
+
+        api_client_mock.return_value.get_enterprise_catalog.return_value = True
+
+        # verify that the mock api was called when saving the catalog after updating the query
+        # enterprise_catalog_2 was most recently modified so the last call should be for that
+        api_client_mock.return_value.update_enterprise_catalog.assert_called_with(
```
bbaker6225, line 678: Just a note that this is a good thing to test thoroughly on stage, as we definitely want to be sure the sync still happens

CHANGELOG.rst
```diff
@@ -14,6 +14,24 @@ Change Log
...
+-------------------
+
+* Add query params to proxy login redirect for new welcome template to be rendered.
+* Fixing proxy_login SSO redirect, adding default next param from proxy_login.
+
+[3.4.27] 2020-07-23
+-------------------
+
+* Adds hide_course_original_price field to the serializer for the EnterpriseCustomer endpoint.
+
```
bbaker6225, line 34: Is this not on master? You shouldn't need to re-add this

tests/test_enterprise/test_signals.py
```diff
@@ -625,3 +626,70 @@ def test_update_catalog_with_existing_service_catalog(self, api_client_mock):
...
+        sender = mock.Mock()  # this would be an EnterpriseCatalogQuery class
+
+        # represents instance of EnterpriseCatalogQuery
+        test_query = EnterpriseCatalogQueryFactory(
+            content_filter=content_filter_1
+        )
+        # represents instances of EnterpriseCustomerCatalog
+        enterprise_catalog_1 = EnterpriseCustomerCatalogFactory(
+            sync_enterprise_catalog_query=True,
+            enterprise_catalog_query = test_query
```
bbaker6225, line 658: I think you'll need to get rid of the spaces around the equal sign here

tests/test_enterprise/test_signals.py
```diff
@@ -625,3 +626,70 @@ def test_update_catalog_with_existing_service_catalog(self, api_client_mock):
...
+            'content_type': 'course',
+        }
+
+        content_filter_2 = {
+            'content_type': 'course2',
+        }
+
+        sender = mock.Mock()  # this would be an EnterpriseCatalogQuery class
+
+        # represents instance of EnterpriseCatalogQuery
```
bbaker6225, line 651: super tiny nit: I don't think these "instances of" comments add much

tests/test_enterprise/test_signals.py
```diff
@@ -625,3 +626,70 @@ def test_update_catalog_with_existing_service_catalog(self, api_client_mock):
...
+            'using': 'default',
+            'created': False,
+        }
+
+        update_enterprise_query(sender, test_query, **kwargs)  # calls post_save signal
+
+        self.assertEqual(enterprise_catalog_1.content_filter, enterprise_catalog_2.content_filter, content_filter_2)
+
+        enterprise_catalog_1.save()
+        enterprise_catalog_2.save()
```
bbaker6225, line 681: Hm this seems like the functionality might not work quite right. We want the updates to get synced to the enterprise catalog service without having to manually save all of them

enterprise/signals.py
```diff
@@ -151,6 +152,19 @@ def delete_enterprise_learner_role_assignment(sender, instance, **kwargs):     #
             pass
 
 
+@receiver(post_save, sender=EnterpriseCatalogQuery)
+def update_enterprise_query(sender, instance, **kwargs):     # pylint: disable=unused-argument
```
adamstankiewicz, line 156: nit: might recommend `update_enterprise_catalog_query` just to be more explicit

CHANGELOG.rst
bbaker6225: Looks good to me once this and `__init__.py` are updated

PR #842, opened 2020-07-28
ENT-3231 setup canvas integration channel

binodpant:
## What this PR does / Acceptance Criteria etc

This PR is not meant to be a valid / usable integrated_channel yet, but the basics of registering an integrated_channel called `canvas` with the package name `integrated_channel.canvas` should be there, a basic set of models and their migrations to get this started should be there. But the actual finalizing of models will require more work (upcoming subtask in this story)

* [x] edx-platform PR needs to be created to pull in this app and make it available in edx-platform : Created https://github.com/edx/edx-platform/pull/24632 (WIP)

Basically if you can sync, run migration and see the tables in the db, and the tests pass for this project in CI, we are there... I think

**Merge checklist:**

**Post merge:**


binodpant:
Test needs to be fixed, but right now trying to sort out the first time migration and add it to the pr

binodpant:
Note: 

This did not pick up the migrations for me

```
./manage.py lms makemigrations
```

I had to run

```
./manage.py lms makemigrations canvas
```

integrated_channels/canvas/client.py
```diff
@@ -0,0 +1,48 @@
...
+        Args:
+            enterprise_configuration (CanvasEnterpriseCustomerConfiguration): An enterprise customers's
+            configuration model for connecting with Canvas
+        """
+        super(CanvasAPIClient, self).__init__(enterprise_configuration)
+        self.global_canvas_config = apps.get_model('canvas', 'CanvasGlobalConfiguration').current()
+        self.session = None
+        self.expires_at = None
+
+    def create_course_completion(self, user_id, payload):  # pylint: disable=unused-argument
```
binodpant, line 32: I do not expect this Pr to implement the client, so just creating a blank impl enough to give us the skeleton to work off of

integrated_channels/canvas/models.py
```diff
@@ -0,0 +1,135 @@
...
+from django.utils.encoding import python_2_unicode_compatible
+
+from integrated_channels.integrated_channel.models import EnterpriseCustomerPluginConfiguration
+
+LOGGER = getLogger(__name__)
+
+
+# pylint: disable=feature-toggle-needs-doc
+@python_2_unicode_compatible
+class CanvasGlobalConfiguration(ConfigurationModel):
```
binodpant, line 21: These models are also not final in this PR, much of that work to figure out fields and complete these models will happen in upcoming subtask (s)

integrated_channels/canvas/models.py
```diff
@@ -0,0 +1,135 @@
...
+    )
+
+    provider_id = models.CharField(
+        max_length=100,
+        default='EDX',
+        verbose_name="Provider Code",
+        help_text="The provider code that Canvas gives to the content provider."
+    )
+
+    history = HistoricalRecords()
```
binodpant, line 111: I did not quite get why for Degreed this model has a historicalRecords added but not for the other models, I just blindly followed that example, but hopefully one of the reviewers can enlighten me here; if the extra complexity is not needed we should reduce it .. but if this is a need let's keep it

integrated_channels/canvas/models.py
```diff
@@ -0,0 +1,135 @@
...
+        Return uniquely identifying string representation.
+        """
+        return self.__str__()
+
+    @staticmethod
+    def channel_code():
+        """
+        Returns an capitalized identifier for this channel class, unique among subclasses.
+        """
+        return 'CANVAS'
```
binodpant, line 135: I have not added the transmissionAudit model here since I expect whoever works on transmitters to do that work, think it makes more sense to do it then rather than now

tests/test_integrated_channels/test_canvas/test_client.py
```diff
@@ -0,0 +1,59 @@
...
+from integrated_channels.canvas.client import CanvasAPIClient
+from test_utils import canvas_factories
+
+NOW = datetime.datetime(2017, 1, 2, 3, 4, 5, tzinfo=timezone.utc)
+NOW_TIMESTAMP_FORMATTED = NOW.strftime('%F')
+
+
+@freeze_time(NOW)
+@pytest.mark.django_db
+class TestCanvasApiClient(unittest.TestCase):
```
binodpant, line 24: This test is also just a 'shell' right now to ensure the client itself is correctly instantiated, but much of the testing here will be done in the subtask to implement this client fully, coming up

binodpant:
Pulling this out of draft status since I consider this code complete now but not yet able to figure out the test failures

binodpant:
Doing a `make upgrade` led to a major versio upgrade of pytest, so rolling that back, at least locally that fixes the test failures for me

binodpant:
I excluded any make upgrade changes here since one of those chnages will break this bild

But if reviewers rather I did that and constrained the deps that causes issues I can do that too

binodpant:
just resolved changelost.rst conflict and pushed no other changes

integrated_channels/canvas/models.py
binodpant: yeah I just see that if we don't use historicalrecords it seems that we get a new row on each save() call, and current() will return the latest record. I figure when you work on the models in the next ticket, we should identify if we need any unique key fields or primary key fields in these models and update accordingly

My understanding is that: There will only be one record for a CanvasGlobalConfiguration (not per enterprise) assuming all canvas installations work the same way ..?

But the per enterprise config is going to be .. ahem.. per enterprise .. and I presume will undergo more frequent changes, hence it is wired up to historical records? I am making up my own reasoning here but that is one reason I can see for the difference

binodpant: in any case I doubt this blocks this PR since the purpose here was to setup something we can work with, nothing works right now here anyway :)

PR #904, opened 2020-07-31
ENT-3284 postman collection and readme

binodpant:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/canvas/dev_docs/README.md
+++ b/integrated_channels/canvas/dev_docs/README.md
@@ -1,0 +1,18 @@
+# Postman collection for Canvas work
+
+Just to help local dev and testing with Canvas
+
+Firstt ensure you have canvas setup as per
+https://openedx.atlassian.net/wiki/spaces/SOL/pages/1644462995/Work+with+a+local+Canvas+Learning+Management+System+on+Mac+OSX
+
+Import this collection into postman
+
+
+## To use this collection
+
+Setup a postman environment and setup these vars
+
+* canvas_url: usually just http://localhost:3000
+* access_token: once you have it
+
+Most endpoints are wired up to use these vars, but if not, fix them!
--- a/integrated_channels/canvas/dev_docs/README.md
+++ b/integrated_channels/canvas/dev_docs/README.md
@@ -2,7 +2,7 @@
 
 Just to help local dev and testing with Canvas
 
-Firstt ensure you have canvas setup as per
+First ensure you have canvas setup as per
 https://openedx.atlassian.net/wiki/spaces/SOL/pages/1644462995/Work+with+a+local+Canvas+Learning+Management+System+on+Mac+OSX
 
 Import this collection into postman
@@ -13,6 +13,6 @@ Import this collection into postman
 Setup a postman environment and setup these vars
 
 * canvas_url: usually just http://localhost:3000
-* access_token: once you have it
+* access_token: once you have it (it will be plugged into requests as a bearer token)
 
 Most endpoints are wired up to use these vars, but if not, fix them!
--- a/integrated_channels/canvas/dev_docs/README.md
+++ b/integrated_channels/canvas/dev_docs/README.md
@@ -13,6 +13,8 @@ Import this collection into postman
 Setup a postman environment and setup these vars
 
 * canvas_url: usually just http://localhost:3000
-* access_token: once you have it (it will be plugged into requests as a bearer token)
+* client_id : for oauth requests only
+* client_secret: for oauth requests only
+* access_token: once you have it (it will be plugged into requests as a bearer token). only used for post/put requests like create course
 
 Most endpoints are wired up to use these vars, but if not, fix them!
--- a/integrated_channels/canvas/dev_docs/README.md
+++ b/integrated_channels/canvas/dev_docs/README.md
@@ -16,5 +16,7 @@ Setup a postman environment and setup these vars
 * client_id : for oauth requests only
 * client_secret: for oauth requests only
 * access_token: once you have it (it will be plugged into requests as a bearer token). only used for post/put requests like create course
+* code: the code returned by the oauth2 endpoints used to fetch access tokens (only applicable to
+the `login/oauth2/token` endpoint)
 
 Most endpoints are wired up to use these vars, but if not, fix them!
--- a/integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
+++ b/integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
@@ -420,7 +420,7 @@
 					"formdata": [
 						{
 							"key": "course[image_url]",
-							"value": "lt5H1qDZfNo9aglt7jb9fLCLdRb1cEZNMdD3VZjORE3o82jyV1LAxNVI0up5fnur",
+							"value": "http://some_image_url",
 							"type": "text"
 						}
 					]
--- a/integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
+++ b/integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
@@ -24,7 +24,7 @@
 					"query": [
 						{
 							"key": "client_id",
-							"value": "10000000000002"
+							"value": "{{client_id}}"
 						},
 						{
 							"key": "response_type",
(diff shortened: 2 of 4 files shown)


integrated_channels/canvas/dev_docs/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Import this collection into postman
+
+
+## To use this collection
+
+Setup a postman environment and setup these vars
+
+* canvas_url: usually just http://localhost:3000
+* access_token: once you have it
```
taliaedX, line 16: Is this the bearer token or something else?

integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
```diff
@@ -0,0 +1,445 @@
...
+							"type": "text"
+						},
+						{
+							"key": "client_id",
+							"value": "10000000000002",
+							"type": "text"
+						},
+						{
+							"key": "client_secret",
+							"value": "bSxFhGIBmqluIMe7fIc0S6fNU75WtD7sB7r6T8euVwoYJ6eeNZIU0cfBS4VwjHwJ",
```
taliaedX, line 66: Is there a way to actually make the template "blank"? ie it has the key client_secret, but the template is not storing its value.
I know these are just for development, but it still seems odd to commit these values.

integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
```diff
@@ -0,0 +1,445 @@
...
+					]
+				},
+				"method": "PUT",
+				"header": [],
+				"body": {
+					"mode": "formdata",
+					"formdata": [
+						{
+							"key": "course[image_url]",
+							"value": "lt5H1qDZfNo9aglt7jb9fLCLdRb1cEZNMdD3VZjORE3o82jyV1LAxNVI0up5fnur",
```
taliaedX, line 423: why is the image url the same as the token? lol.
Given you have access_token as a var, I'd replace all spots with the token's value with the variable `{{access_token}}`

mattdrayer:
Just a heads up that we have a private repository for Postman collections, if you want to use it: https://github.com/edx/edx-postman-config

integrated_channels/canvas/dev_docs/README.md
```diff
@@ -0,0 +1,18 @@
+# Postman collection for Canvas work
+
+Just to help local dev and testing with Canvas
+
+Firstt ensure you have canvas setup as per
```
georgebabey, line 5: nit: `First`

integrated_channels/canvas/dev_docs/README.md
binodpant: yup updated doc

integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
binodpant: yah I do not know of a good way to do that, I think I could set the environment to none or sth in postman then export these.. but for this round I can just clear these out of the json

binodpant: actually I can and should set these to the variable instead of the value

binodpant: updated to use vars that can be sett in postman, instead of raw values

integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
binodpant: haha fixed it

binodpant:
> Just a heads up that we have a private repository for Postman collections, if you want to use it: https://github.com/edx/edx-postman-config

Thanks Matt. I was hoping to keep this local to this code but I also see the value of the private repo. For this instance since we only committed non prod or non sensitive values, does this still work?

mattdrayer:
I think it's really up to you all -- see the benefit of keeping the collection close to the code, however there are other people at edX who use Postman to connect to these systems (think: ECS, Sales Engineers) and so it might be easier for them to find collections if they are all in one place.  Also it's very possible that over time collections can/will grow to include more than what is implemented in the code.

At the end of the day we're not really talking about a lot of people requiring access to Postman collections, and so with proper documentation/pointers we can probably avoid confusion about where things live.

binodpant:
believe I addressed the feedback, except we can reconsider if moving this content to the other repo in the future makes sense we can do that. otherwise all done; version, changelog also updated

integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
```diff
@@ -58,17 +58,17 @@
...
 						{
 							"key": "client_secret",
-							"value": "bSxFhGIBmqluIMe7fIc0S6fNU75WtD7sB7r6T8euVwoYJ6eeNZIU0cfBS4VwjHwJ",
+							"value": "{{client_secret}}",
 							"type": "text"
 						},
 						{
 							"key": "code",
-							"value": "b77ce731852e1691c69eac9ac5bf8d62bdf8ae5f59c521726e37e0aa87ab49a6561b615146ff918e0bd79556c055a81f9fce589e35c5bf98eaafa88f703939bf",
+							"value": "{{code}}",
```
taliaedX, line 71: Just a heads up that this var isn't mentioned in the readme.

binodpant: nice catch got it!

integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
```diff
@@ -0,0 +1,445 @@
...
+		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
+	},
+	"item": [
+		{
+			"name": "GET oauth code",
+			"request": {
+				"method": "GET",
+				"header": [],
+				"url": {
+					"raw": "{{canvas_url}}/login/oauth2/auth?client_id=10000000000002&response_type=code&state=edx-canvas-user-state&redirect_uri=https://example.com/oauth_complete",
```
taliaedX, line 15: shouldn't `client_id=10000000000002` be `client_id={{client_id}}` here?

integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
```diff
@@ -0,0 +1,445 @@
...
+			},
+			"response": []
+		},
+		{
+			"name": "GET course settings",
+			"request": {
+				"method": "GET",
+				"header": [],
+				"url": {
+					"raw": ""
```
taliaedX, line 377: Am I wrong or is this basically just a blank request template?

integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
```diff
@@ -0,0 +1,445 @@
...
+			},
+			"response": []
+		},
+		{
+			"name": "list courses",
+			"request": {
+				"method": "GET",
+				"header": [],
+				"url": {
+					"raw": "{{canvas_url}}/api/v1/courses",
```
taliaedX, line 96: so according to the doc this endpoint gets courses for the "current user". I'm not sure how well this will work if the given user is not actually enrolled in any of these courses. I wonder if we should go by account id here, ie `{{canvas_url}}/api/v1/accounts/:account_id/courses/` but what has been your experience with courses you aren't specifically enrolled in?

integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
binodpant: yeah we can use that var too, I had been plugging in different values when testing multiple keys but better as a var, changing

integrated_channels/canvas/dev_docs/postman_collections/Canvas.postman_collection.json
binodpant: hmm will try re exporting it is not empty in postmane for me

PR #907, opened 2020-08-03
ENT-2892 | WAIT BEFORE MERGING! Migration to copy old field values to new fields.

iloveagent57:
~**DO NOT MERGE** until an edx-platform prod release has occurred for edx-enterprise version 3.4.33.~
3.4.33 been merged and a deploy has started: https://github.com/edx/edx-platform/pull/24667

After this, the frontend can safely expect the same data it's been getting when querying the backend for branding config data.

https://openedx.atlassian.net/browse/ENT-2892

Front-end changes that depend on this PR: https://github.com/edx/frontend-app-learner-portal-enterprise/pull/79

**Merge checklist:**

**Post merge:**


enterprise/migrations/0106_move_branding_config_colors.py
```diff
@@ -0,0 +1,37 @@
...
+from django.db import migrations
+from django.db.models import Q
+
+
+def copy_primary_secondary_colors(apps, schema_editor):
+    EnterpriseCustomerBrandingConfiguration = apps.get_model('enterprise', 'EnterpriseCustomerBrandingConfiguration')
+    for branding_config in EnterpriseCustomerBrandingConfiguration.objects.filter(
+        Q(banner_background_color__isnull=False) | Q(banner_border_color__isnull=False)
+    ):
+        print(branding_config)
```
bbaker6225, line 12: Leftover?

enterprise/migrations/0106_move_branding_config_colors.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def clear_primary_secondary_colors(apps, schema_editor):
+    EnterpriseCustomerBrandingConfiguration = apps.get_model('enterprise', 'EnterpriseCustomerBrandingConfiguration')
+    for branding_config in EnterpriseCustomerBrandingConfiguration.objects.filter(
+        Q(primary_color__isnull=False) | Q(secondary_color__isnull=False)
+    ):
+        branding_config.primary_color = None
+        branding_config.secondary_color = None
+        branding_config.save()
```
bbaker6225, line 25: Do we even need to have a reverse here? I assume it's fine if the old data stayed in the former fields if someone wanted to run back these migrations

iloveagent57: It would be fine, but it was easy to write and I liked having a way to repeatedly test locally and see the same result.

PR #910, opened 2020-08-04
Removed can_use_enterprise_catalog flag

anaidu0:
Removed all references to can_use_enterprise_catalog and removed related test cases as well as customer exclusion tests.

**Merge checklist:**

**Post merge:**


enterprise/utils.py
```diff
@@ -1085,24 +1085,6 @@ def discovery_query_url(content_filter, html_format=True):
...
-    enterprise uuid is not in the ENTERPRISE_CUSTOMERS_EXCLUDED_FROM_CATALOG list.
-
-    Args:
-        enterprise_uuid: the unique identifier for an enterprise customer
-
-    Returns:
-        boolean: True if sample is active and enterprise is not excluded
-                 False if sample not active or enterprise is excluded
-    """
-    return (waffle.sample_is_active(USE_ENTERPRISE_CATALOG) and
```
chavesj, line 1102: don't forget to remove unused imports here

enterprise/migrations/0109_remove_use_enterprise_catalog_sample.py
```diff
@@ -0,0 +1,32 @@
...
+
+from enterprise.constants import USE_ENTERPRISE_CATALOG
+
+
+def create_sample(apps, schema_editor):
+    """Create the `use_enterprise_catalog` sample if it does not already exist."""
+    Sample = apps.get_model('waffle', 'Sample')
+    Sample.objects.get_or_create(
+        name=USE_ENTERPRISE_CATALOG,
+        defaults={'percent': 0},
```
bbaker6225, line 14: We probably want the percentage at whatever it's at currently (100 I'm pretty sure) if the migration is reversed, right?

CHANGELOG.rst
```diff
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 --------------------
 
+[3.5.5] 2020-08-12
+-------------------
+
+* ENT-2939: removing waffle flag and utility function used in enterprise-catalog transition
```
bbaker6225, line 20: This is probably at least a minor version bump I think

test_utils/fake_catalog_api.py
```diff
@@ -46,6 +47,7 @@
     'end': '3000-12-31T18:00:00Z',
     'enrollment_start': None,
     'enrollment_end': None,
+    'enrollment_url': 'https://fake.url',
```
bbaker6225, line 50: Nit: maybe add a variable for this and use that?

tests/test_consent/api/test_views.py
```diff
@@ -310,11 +311,21 @@ def _assert_expectations(self, response, expected_body, expected_status):
...
     @ddt.unpack
-    def test_consent_api_get_endpoint(self, factory, items, request_body, expected_response_body, expected_status_code):
+    def test_consent_api_get_endpoint(
+            self,
+            factory,
+            items,
+            request_body,
+            expected_response_body,
+            expected_status_code,
+            catalog_api_client_mock,
```
bbaker6225, line 321: The naming of this feels a little weird. I think you can set the return value in the mock decorator to simplify things a bit and not have to do the renaming

tests/test_enterprise/api/test_views.py
```diff
@@ -1171,6 +1172,7 @@ def test_enterprise_catalog_contains_content_items_no_query_params(self):
         assert 'course_run_ids' in message
         assert response.status_code == 400
 
+    @pytest.mark.skip(reason='Catalog course run detail endpoint now lives in enterprise-catalog service')
```
bbaker6225, line 1175: These detail endpoints don't live in enterprise catalog yet though

tests/test_consent/api/test_views.py
chavesj: Good call. I meant to come back to this for cleanup after encountering an example of what you described 👍

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -76,22 +76,16 @@ def export(self, **kwargs):
...
+            self.enterprise_customer,
+            enterprise_catalogs=self.enterprise_configuration.customer_catalogs_to_transmit
+        )
+        LOGGER.info(
+            'Getting metadata for Enterprise [%s], Catalogs [%s] from Enterprise Catalog Service. Results: [%s]',
+            self.enterprise_customer.name,
+            self.enterprise_configuration.customer_catalogs_to_transmit,
+            json.dumps(content_metadata_items)
+        )
         LOGGER.info('Retrieved content metadata for enterprise [%s]', self.enterprise_customer.name)
```
adamstankiewicz, line 89: Do we need both of these logs? We could probably remove the `LOGGER.info` on line 89.

tests/test_models.py
```diff
@@ -1185,89 +1136,45 @@ def test_audit_param_in_program_enrollment_url(self, config_mock):
...
-        contains_courses_mock.return_value = False
-        enterprise_customer_catalog = factories.EnterpriseCustomerCatalogFactory()
-        assert enterprise_customer_catalog.get_course_and_course_run('fake-course-run-id') == (None, None)
-
     @mock.patch('enterprise.models.EnterpriseCatalogApiClient', return_value=mock.MagicMock())
-    def test_get_course_and_course_run_no_content_items_waffle_sample(self, api_client_mock):
+    def test_get_course_and_course_run_no_content_items(self, api_client_mock):
         """
-        Verify the `get_course_and_course_run` method returns the same tuple with the enterprise catalog sample active.
+        Verify the `get_course_and_course_run` method returns the same tuple now that the waffle flag is removed.
```
adamstankiewicz, line 1142: nit: I feel it might be unnecessary to mention that a waffle flag/sample was removed here.

For example, "same tuple now that the waffle flag is removed" in this docstring (and others) could be replaced by the original docstring for the removed test, e.g. "The ``get_course_and_course_run`` method returns a tuple (None, None) when no content items exist."

tests/test_enterprise/api/test_views.py
chavesj: Damn.. okay that's my bad. Working on test fixes now

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -105,22 +106,27 @@ def test_get_course_specific_consent(
...
         }
-        course_catalog_api_client_mock.return_value.get_course_run.return_value = course_run_details
-        course_catalog_api_client_mock.return_value.get_course_details.return_value = {'title': 'Demo Course'}
+
+        enterprise_catalog_api_client_mock.return_value.enterprise_contains_content_items.return_value = True
+
+        mock_discovery_catalog_api_client = course_catalog_api_client_mock.return_value
+        mock_discovery_catalog_api_client.get_course_id.return_value = course_key
+        mock_discovery_catalog_api_client.get_course_run.return_value = course_run_details
+        mock_discovery_catalog_api_client.get_course_details.return_value = {'title': 'Demo Course'}
```
adamstankiewicz, line 128: super tiny nit: could use the `title` from `course_run_details` here.

PR #914, opened 2020-08-05
add api endpoint to update mode for licensed enrollments

adamstankiewicz:
This PR introduces a new endpoint that will be called during license revocation to update a user's licensed enterprise course enrollments to the "audit" mode.

Endpoint URL: `/enterprise/api/v1/licensed-enterprise-course-enrollment/license_revoke/`

Related frontend PR: https://github.com/edx/frontend-app-learner-portal-enterprise/pull/84

**Merge checklist:**

**Post merge:**


enterprise/api/v1/views.py
```diff
@@ -242,6 +253,128 @@ def get_serializer_class(self):
...
+            user_id=user_id,
+            enterprise_customer=enterprise_id,
+        )
+
+        licensed_enrollments = self.queryset.filter(
+            enterprise_course_enrollment__enterprise_customer_user=enterprise_customer_user
+        )
+        if not licensed_enrollments:
+            msg = 'User {} does not have any licensed enterprise course enrollments.'.format(user_id)
+            return Response(msg, status=status.HTTP_404_NOT_FOUND)
```
kellmarie, line 310: Do we want to consider revoking an enterprise license from someone who hasn't enrolled in anything yet as an error? I think it's a valid scenario for someone to have a license revoked before they enrolled in anything. I wish HTTP had a "we successfully didn't do anything but that's what was supposed to happen in this case" type response status.

enterprise_learner_portal/api/v1/serializers.py
```diff
@@ -68,7 +68,7 @@ def to_representation(self, instance):
         representation['due_dates'] = []
         representation['pacing'] = course_overview['pacing']
         representation['org_name'] = course_overview['display_org_with_default']
-        representation['saved_for_later'] = instance.saved_for_later
+        representation['is_revoked'] = instance.license.is_revoked if instance.license else False
```
kellmarie, line 71: do we want to be removing `saved_for_later` from this serializer?

adamstankiewicz: Now that `saved_for_later` is returned as its own course status, there's not really a reason for it to be in the serializer since it'd also be coming back with `course_run_status: "saved_for_later"`, and it felt odd to have both `course_run_status` and `saved_for_later` saying the same thing 🤷‍♂️ 

It was necessary before since `saved_for_later` courses were previously returning with `course_run_status: "completed"`, so `saved_for_later` in the response was the only way to differentiate.

enterprise/api/v1/views.py
adamstankiewicz: Yeah, I guess there's no harm is letting the rest of the method execute in this case, since `licensed_enrollments` would be an empty list with nothing to iterate through anyways.

I agree a revoke is possible before they'd actually enrolled in anything. I will update.

enterprise/api/v1/views.py
```diff
@@ -242,6 +253,124 @@ def get_serializer_class(self):
...
+            models.EnterpriseCustomerUser,
+            user_id=user_id,
+            enterprise_customer=enterprise_id,
+        )
+
+        licensed_enrollments = self.queryset.filter(
+            enterprise_course_enrollment__enterprise_customer_user=enterprise_customer_user
+        )
+        enterprise_enrollments = [enrollment.enterprise_course_enrollment for enrollment in licensed_enrollments]
+        course_run_ids = [enrollment.course_id for enrollment in enterprise_enrollments]
```
iloveagent57, line 309: I think this and some stuff below gets simpler if you use a dictionary:
```python
enrollments_by_course_id = {enrollment.course_id: enrollment.enterprise_course_enrollment for enrollment in licensed_enrollments}
```
Then you don't have to do that popping from a new list inside the for loop.

enterprise/api/v1/views.py
```diff
@@ -242,6 +253,124 @@ def get_serializer_class(self):
...
+                _('To use this endpoint, this package must be '
+                  'installed in an Open edX environment.')
+            )
+
+        request_data = request.data.copy()
+        self._validate_license_revoke_data(request_data)
+
+        user_id = request_data.get('user_id')
+        enterprise_id = request_data.get('enterprise_id')
+        mode = CourseModes.AUDIT
```
iloveagent57, line 297: nit: `audit_mode`?

enterprise_learner_portal/api/v1/serializers.py
kellmarie: gotcha

tests/test_enterprise/api/test_views.py
```diff
@@ -2283,6 +2286,77 @@ def test_post_request_codes_permissions(self, implicit_perm, explicit_perm, expe
 
         assert response.status_code == expected_status
 
+    @mock.patch('enterprise.api.v1.views.EnrollmentApiClient')
+    @mock.patch('enterprise.api.v1.views.get_certificate_for_user')
+    @mock.patch('enterprise.api.v1.views.get_course_overviews')
+    @ddt.data(
+        (True, False, True, status.HTTP_204_NO_CONTENT),
```
iloveagent57, line 2293: Nit: `ddt.unpack` will unpack a dict if you want to write these as
```python
{'has_permissions': True, 'is_courrse_enrollment_completed': False, 'is_licensed_enrollment_revoked': True, 'expected_status_code': status.HTTP_204_NO_CONTENT}
```
...which I personally find helpful, but is maybe a matter of taste.

PR #915, opened 2020-08-05
ENT-3287 Start from scratch, fresh migrations for canvas with corrected fieldnames

binodpant:
Mainly, 

client_id instead of 'key'
client_secret instead of 'secret'

rest is the same as initially planned

this work will need to be followed up by a pr in edx-platform which will include adding canvas to the list of apps so migrations will actually run

**Merge checklist:**

**Post merge:**


integrated_channels/canvas/models.py
```diff
@@ -89,13 +89,15 @@ class CanvasEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguratio
 
     canvas_company_id = models.CharField(
         max_length=255,
+        null=True,
         blank=True,
```
taliaedX, line 93: Should this allow blank though? When ECS fills this out I imagine this field would be required no? Again, having null here means database-wise there will be no issues with not having current values, so you don't also need blank in that regard. The question is whether you want to allow someone to fill out the info for a canvas integration and *not* provide a company id.

integrated_channels/canvas/models.py
```diff
@@ -89,13 +89,15 @@ class CanvasEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguratio
...
         verbose_name="Canvas Organization Code",
         help_text="The organization code provided to the enterprise customer by Canvas."
     )
 
     canvas_base_url = models.CharField(
```
taliaedX, line 98: This can be something we adjust later I guess, but do we plan on this being a per-customer value, because if not, we probably shouldn't have a database field for it? Like is it going to include their account number or something or is it just going to be basically `canvas.com` for everybody?

binodpant: I am pretty sure we will have a base_url per customer.. since it's their running instance of canvas (or a url they give us from a hosted provider of some kind)

integrated_channels/canvas/models.py
binodpant: yeah that is a good question. not sure either. I am not actually sure how we are going to use this field, if at all. @alex-sheehan-edx may have thoughts here based on what has come out of oauth? and if we need this field for oauth and other purposes, looks like we must not allow blank or null either

integrated_channels/canvas/models.py
binodpant: also if two customers happen to provide the same base url somehow, I guess we will still handle cleanly since we will use account_id or oauth tokens to do the rest ?

taliaedX: Ah I assumed it was just uniformly hosted (aka saas), but you're absolutely right they could have it hosted anywhere.

Re uniqueness of the url:
As long as the base_url is not expected to be unique nor is it being used as a primary key it should be fine. I don't see any field options being used here that would indicate it as a unique field, so that should be fine if they do have the same instance url.

integrated_channels/canvas/models.py
binodpant: or at least not allow blank

alex-sheehan-edx: Slightly confused as to what the company ID represents here.. For OAuth, as of right now, the values I need are `canvas_base_url`, `client_id`, `client_secret`, `(client_?)code`, `refresh_token` (retrieved when the user is prompted by Canvas to authorize and clicks the confirmation button), `access_token` (retrieved by the previous values when instantiating the client's session) and the constants `oauth_token_auth_path` ("login/oauth2/token") and `oauth_redirect_uri` (a to be added endpoint for after the Canvas prompted authorization flow)

integrated_channels/canvas/models.py
alex-sheehan-edx: Yeah, I think this is customer dependent but not necessarily unique.

integrated_channels/canvas/models.py
binodpant: yeah looks like may not need it here then. degreed was different, it uses it like this https://github.com/edx/edx-enterprise/blob/af5b13a109f611375c3fffbf97aa236cf0fd38c7/integrated_channels/degreed/transmitters/content_metadata.py#L27 but for canvas if this is never going to be used, then let's drop it?

binodpant: same with `provider_id`?

taliaedX: I think provider_id is safe to get rid of for now. I'd however change canvas_company_id to canvas_account_number. We need that for a bunch of URLs, so we do need to store it per customer

integrated_channels/canvas/migrations/0001_initial.py
```diff
@@ -44,8 +44,8 @@ class Migration(migrations.Migration):
             ],
             options={
                 'get_latest_by': 'history_date',
-                'ordering': ('-history_date', '-history_id'),
                 'verbose_name': 'historical canvas enterprise customer configuration',
+                'ordering': ('-history_date', '-history_id'),
```
alex-sheehan-edx, line 48: fairly inconsequential, but does the reordering matter here?

taliaedX:
@alex-sheehan-edx I kind of agree with removing oauth api path. I don't agree with adding fields to this one. I get it's the initial so why not make it all encompassing, but it does correspond to model work, so let's keep that separate. I know we kept rewriting every damn initial migration when we did Q data depot work, but it aggravated me to do that. We're building off of initial work, it's okay to show that. Let's not rewrite history more than we have to.

alex-sheehan-edx:
> @alex-sheehan-edx I kind of agree with removing oauth api path. I don't agree with adding fields to this one. I get it's the initial so why not make it all encompassing, but it does correspond to model work, so let's keep that separate. I know we kept rewriting every damn initial migration when we did Q data depot work, but it aggravated me to do that. We're building off of initial work, it's okay to show that. Let's not rewrite history more than we have to.

Good points, I guess old habits are rearing their ugly heads. I'm fine with not throwing everything in the initial, but also 👍 to removing oauth api path.

integrated_channels/canvas/migrations/0001_initial.py
binodpant: this was automatically generated.. no idea what this means .. let's see what the latest migrations say

binodpant:
> If you are renaming fields and removing unnecessary fields from the model, do you want to add my `refresh_token` and remove `oauth_api_path` from the global configuration model (work that my draft PR is doing right now) as well? Or save it for when we iterate on my client work?

okie, let's do it

taliaedX:
@alex-sheehan-edx yeah it's understandable to feel like if we're just messing with the initial migration anyways, why not just add everybody's changes, but I think because yours come with particular testing needs it should be kept separate.

binodpant:
> @alex-sheehan-edx yeah it's understandable to feel like if we're just messing with the initial migration anyways, why not just add everybody's changes, but I think because yours come with particular testing needs it should be kept separate.

so just to summarize I will just to these then:

* get rid of provider_id
* change canvas_company_id to canvas_account_number

and that's it

integrated_channels/canvas/migrations/0001_initial.py
taliaedX: I assumed it was an isort issue.

alex-sheehan-edx:
@binodpant may be obvious but if you remove the oauth api path, don't forget to also remove it from the `canvas_factories.py` otherwise tests will fail (had to figure that one at through experience)

integrated_channels/canvas/models.py
binodpant: is `canvas_account_number` the :account_id in urls? so then it should be an integer?

taliaedX:
`change canvas_company_id to canvas_account_number` account_id per our chat, but yes, that's perfect.

PR #916, opened 2020-08-05
ENT-3227 : Client Impl + Adding oauth access token retrieval

alex-sheehan-edx:
**Merge checklist:**

**Post merge:**

**What this PR does:**
- [x] Custom field `refresh_token` added to the CanvasEnterpriseCustomerConfiguration model
- [x] Field `oauth_api_path ` moved from model to app config
- [x] Client now has the ability to create a session 
      - [ ] Sessions, upon instantiation, retrieve and save the access token associated with the enterprise user.

**What work is left after this merges:**
- Related Issue [ENT-3303](https://openedx.atlassian.net/browse/ENT-3303)
- [ ] Add another endpoint that will be hit when the user authorizes us to access their Canvas account, to do this, they will need to add this new endpoint as the redirect path on their developer key. 
     - I see the flow for a user setting up the necessary credentials for us to integrate Canvas as such:
              - User's generate a Developer Key through Canvas that contains both a client ID and secret. 
              - User adds our (TODO) `/oauth_complete` endpoint in the setting's of their developer key `redirect_uri`
              - The user needs to hit the endpoint `<user's canvas_base_url>/login/oauth2/auth?client_id=<user's client ID>&response_type=code&redirect_uri=<(TODO) our /oauth_complete endpoint>`
                        - Right now, I can see this happening a couple of ways. We could either have ECS provide them with the link and ask them to just copy and paste it into their browser, or we could possibly build out a UI in the admin frontend portal where they provide the client ID and hit a button labeled `Generate Oauth code`. Also very open to other suggestions here. 

**Steps to test/use locally:**
- fetch and pull this branch
- run migrations so that you have the the new field `refresh_token` in the canvas model
- Have a local instance of Canvas running (https://github.com/instructure/canvas-lms/wiki) 
- Create a Canvas admin user with developer key creation capabilities
- Generate a developer key
   - go into the settings of the dev key and add a redirect uri (before the oauth/complete endpoint is complete, it doesn't matter what this uri is, as long as you remember it.
- using the secret and client id from the developer key visit:
   - <your local canvas base path>/login/oauth2/auth?client_id=<your client id>&response_type=code&redirect_uri=<your redirect uri>
- when prompted, confirm authorization
- after authorizing, you will be redirected to whatever redirect uri was provided
   - Save the code passed in at the end of the redirect uri, you will use this to get the access token and refresh token
- Make a post request to `http://localhost:3000/login/oauth2/token` with the following params
   -  `grant_type` : `authorization_code`
   -  `client_id` : < your client id >
   -  `client_secret` : < your client secret >
   -  `redirect_uri` : < the same redirect uri that you saved to your developer key >
   -  `code` : < the code you got after authorizing >
- Copy both the access token and the refresh token 
   - the access token is your auth token that will allow you to post/delete courses to Canvas
- Create a new Enterprise user in edx-enterprise and create a new CanvasEnterpriseCustomerConfiguration and CanvasGlobalConfiguration model for the user, using the client id, secret, refresh token, base url, account id, course api path
- you should now be able to generate a new client session that retrieves and stores the access token in order to post and delete to your canvas instance.


integrated_channels/canvas/client.py
```diff
@@ -44,7 +47,78 @@ def update_content_metadata(self, serialized_data):
...
+        auth_token_url = urljoin(
+            self.enterprise_configuration.canvas_base_url,
+            self.config.oauth_token_auth_path,
+        )
+        auth_token_params = {
+            'grant_type': 'refresh_token',
+            'client_id': client_id,
+            'client_secret': client_secret,
+            'state': self.enterprise_configuration.enterprise_customer.uuid,
+            'refresh_token': self.enterprise_configuration.refresh_token,
```
taliaedX, line 114: so how do you get the first `refresh_token` ? Is that via the oauth/redirect button stuff?

tests/test_integrated_channels/test_canvas/test_client.py
```diff
@@ -42,17 +43,33 @@ def setUp(self):
...
+                responses.POST,
+                self.oauth_url,
+                json={"access_token": self.access_token},
+                status=200
+            )
+
+            canvas_api_client = CanvasAPIClient(self.enterprise_config)
+            canvas_api_client._create_session()  # pylint: disable=protected-access
+
+            assert canvas_api_client.session.headers['Authorization'] == "Bearer " + self.access_token
```
taliaedX, line 71: This should just test the `_get_oauth_access_token` and title the method around it as such. You're not really doing anything with the actual client or session here since it's mocked. You're really only testing the functionality of this oauth token function, so just test that directly.

If you want to test the client and session (and you probably should), test what happens under non-ideal conditions. For instance, what happens if the token isn't in the config? Is your code actually handling different status codes correctly? etc. Ideal conditions are useful for verifying your function logic, but it's not very helpful for testing a client functioning as expected. So keep this test to its core functional testing and add some tests here that actually test various responses and what happens if information is missing, because that's extremely valuable in catching bugs before we release this.

integrated_channels/canvas/client.py
alex-sheehan-edx: At the point where we're creating a client session to update/post/delete a course to canvas, the user should have already gone through the redirect/get code process. The auth code is a single use so we can't reliably use it to generate access tokens, so we need to have the refresh token from a new `oauth/complete` endpoint.

tests/test_integrated_channels/test_canvas/test_client.py
alex-sheehan-edx: Fair points, 👍 on the focused testing. I'll add and adjust accordingly.

alex-sheehan-edx: > For instance, what happens if the token isn't in the config?

I think I may have to generalize this test a bit, because as I have it set up, the client doesn't care that the refresh token doesn't exist until it would receive a bad request 400 from Canvas. But also it feels weird because I would be mocking an error response then just testing that the error response was raised. Not sure if that's worth it's own test. Thoughts?

binodpant:
Let's add the ticket number on the pr title? like `ENT-zzz: something` generally works

integrated_channels/canvas/apps.py
```diff
@@ -12,3 +12,5 @@ class CanvasConfig(AppConfig):
     """
     name = 'integrated_channels.canvas'
     verbose_name = "Enterprise Canvas Integration"
+    oauth_token_auth_path = "login/oauth2/token"
+    oauth_redirect_uri = "urn:ietf:wg:oauth:2.0:oob"  # TODO- the uri needs to link to Django admin (?)
```
binodpant, line 16: so is the plan here to use this out of band uri for now and change it to our prod value (using env vars I presume for different envrions), in a future round of work after the endpoint is ready?

binodpant: Even if this is temporary maybe let's add a doc link somewhere about what this urn is and why we are using it here?

integrated_channels/canvas/client.py
```diff
@@ -44,7 +47,78 @@ def update_content_metadata(self, serialized_data):
     def delete_content_metadata(self, serialized_data):
         pass
 
-    def _get_oauth_access_token(self, client_id, client_secret):
+    def _post(self, url, data, scope):
```
binodpant, line 50: do we intend to use scope now? if not. we can add it later when needed

binodpant: unless the interface requires it somehow

integrated_channels/canvas/client.py
```diff
@@ -44,7 +47,78 @@ def update_content_metadata(self, serialized_data):
...
+        return response.status_code, response.text
+
+    def _create_session(self):
         """
+        Instantiate a new session object for use in connecting with Canvas. Each enterprise customer
+        connecting to Canvas should have a single client session.
+        """
+        if self.session is None:
+            # Create a new session with a valid token
+            if self.session:
```
binodpant, line 81: wait how can this be? the if above says is None ... ? 😕

integrated_channels/canvas/client.py
```diff
@@ -44,7 +47,78 @@ def update_content_metadata(self, serialized_data):
...
+                self.session.close()
+            oauth_access_token = self._get_oauth_access_token(
+                self.enterprise_configuration.client_id,
+                self.enterprise_configuration.secret,
```
binodpant, line 85: this will change to client_secret now? if you rebase to my latest changes

integrated_channels/canvas/client.py
```diff
@@ -44,7 +47,78 @@ def update_content_metadata(self, serialized_data):
...
+            )
+            session = requests.Session()
+            session.headers['Authorization'] = 'Bearer {}'.format(oauth_access_token)
+            session.headers['content-type'] = 'application/json'
```
binodpant, line 89: I was not aware that the headers can be case insensitive when used with requests. I have always been used to `Content-Type` ...

integrated_channels/canvas/client.py
binodpant: hmm, will be worth checking where what is the expiry timeframe of that code? if it expires this code needs to run within that timeframe right? if nothing, we can/should document that somewhere.

Also, if the code expires how are we going to inform the user reading the logs of this? We may need a very specific message/exception for that case?

binodpant, line 114: yeah it may be worth doc'ing somewhere in the class level doc maybe? that the value of refresh_token here should already be set in the enterpriseccustomerconfig, and if it isn't we should raise an exception? that is a different failure mode than, say, the code expiring

integrated_channels/canvas/client.py
alex-sheehan-edx: Whoops, forgot to take this part out when using degreed as a reference. I don't yet see how scope would fit into our canvas integration here, just with how we're having users provide us with their auth credentials. But perhaps it'll come up later and we can re-add.

integrated_channels/canvas/client.py
alex-sheehan-edx: :face_palm: good points

integrated_channels/canvas/client.py
alex-sheehan-edx: Yup! Going through the rebase process rn.

integrated_channels/canvas/client.py
alex-sheehan-edx: yeah I was caught off guard too, but both degreed had lowercase and I was able to post through postman with it lowercase. ¯\_(ツ)_/¯ I'm open to changing it to be more inline with what your standard has been but I also feel as though we should keep some consistency throughout integrated channels. Thoughts?

integrated_channels/canvas/client.py
alex-sheehan-edx: Docing the whole process I think will be beneficial for ECS as well as any other engineer who may stumble across oauth 👍 . 

I could check for the presence of the refresh token, but the canvas API has some pretty straightforward error responses if you try and request without one. `{"error":"invalid_request","error_description":"You must provide the refresh_token parameter when using the refresh_token grant type"}`. My intuition is that either approach is viable here, but can default to your preferences if you think there's a standard we should adhere to.

binodpant: yeah ok if this output will echo into the logs when we get this that's great!

This maybe leads into some testing thoughts:

I was mainly thinking about explicitly locking down the various outcomes that can happen, ideally in small unit test cases

for instance can we have test cases unit level where we can get:

1. expired code hence cannot get refresh_token?
2. client_id is incorrect or sth like that and we get an invalid or auth response code? not sure what happens in that case
3. anything else you can think of that can go wrong, there may be more that I cannot think of right now

A really good way to do such things is to have explicit exceptions named appropriately and then unit tests can just inject the fails and check for the right exception, just a thought if not too much work

We should basically assume things will go wrong occasionally in this process (such as a temporary glitch in the server, and a 500 maybe for example), so in the very least maybe test at least one or two of the salient negative cases would be great

binodpant:
Are there any testing notes on what one can do locally to test these changes? Did you run this against your local canvas instance and obtained a refresh token? would be interesting to doc that either here or even in the dev_docs area whatever you feel appropriate. My personal pref is dev_docs since we are going to want to try this locally (or against instructure instance) soon

binodpant:
For the section `What work is left after this merges:` in description I wuld link to that ticket instead maybe and even move that content into that ticket? or at least link to the ticket perhaps would be best?

integrated_channels/canvas/apps.py
alex-sheehan-edx: This is the out of band URI because I found that when using it, you didn't have to match the user's indicated redirect url, which was useful for testing. Until the `oauth/complete` endpoint work is done, I've found it useful just to use this uri. I'll add a doc snippet explaining why it's here!

integrated_channels/canvas/client.py
binodpant: chatted with Alex. Alex, you can safely ignore my notes on 'code' expiring since that code will be used to call back the callback_uri so there is no notino of them expiring on our side. but let's find out about refresh_token itself (if they expire and how deal with that). But yeah the general thoughts on error handling are still pertinent I hope

alex-sheehan-edx: > 1. expired code hence cannot get refresh_token?

Discussed offline, the client/client session shouldn't save any reference to the code and to my knowledge refresh tokens don't expire (currently a 3 day old token isn't expired and nothing in their docs that I've found indicates an expiration **on the refresh token**. If the refresh token is invalid at the time of the request, we get a similar response `{"error":"invalid_request","error_description":"refresh_token not found"}`

> 2. client_id is incorrect or sth like that and we get an invalid or auth response code? not sure what happens in  that case

In the same line of thinking, Canvas responds with `{"error":"invalid_client","error_description":"unknown client"}` if you have an invalid client id, which I think is sufficient to let any dev know what's going wrong in the logs. 

Because of these canvas exception responses, it's difficult to think of situations that are testable for the client. Open to suggestions but going to take a moment and try and think of potential edge cases.

integrated_channels/canvas/apps.py
alex-sheehan-edx: 🤦 updates, redirect uri isn't required any more because we switched to refresh tokens, so I'm removing it from the config.

integrated_channels/canvas/client.py
binodpant: no worries, we can test manually too when we are testing this. I am happy to lend hands in that regard too if we want to do that even after merge is okay just to see what things look like. what happens often is once we ship to prod we find it hard to diagnose without the right exceptions, let's just do the best we can at this stage of the work, we are doing great!

tests/test_integrated_channels/test_canvas/test_client.py
taliaedX: Per offline convo:
- I for some reason thought the token was used in local logic, not just directly passing to the canvas API. In which case, testing what happens when it's missing isn't of much use.
- I think there are parts in the code where we're expecting certain errors (there's a part in the client code `except (KeyError, ValueError)`) and I think I'm curious what our fallback is there. We probably do still want to raise on novel errors, because we need to know they exist to handle them better, but not sure it should just cause the whole app to fail, but also silent failures are the worssst.
- I don't think it requires a test necessarily, but just verifying what happens when you are missing parameters. Especially ones we expect to be manually entered by ECS/Customer or dev adds to a config.

integrated_channels/canvas/client.py
```diff
@@ -12,8 +14,21 @@ class CanvasAPIClient(IntegratedChannelApiClient):
...
+        endpoint after the user authorizes the use of their Canvas account.
+
+    Order of Operations:
+        Before the client can connect with an Enterprise user's Canvas account, the user will need to
+        follow these steps
+            - Provide the ECS team with their developer key's client ID and secret.
+            - ECS will return a url for the user to visit which will prompt authorization and redirect when
+            the user hits the `confirm` button.
+            - The redirect will hit the `oauth/complete` endpoint which will use the provided oauth code
+            to request the access token necessary to post/delete/update courses and save it to the model.
```
binodpant, line 31: I think there is another step here which is to get the refresh token and save it to the model?

integrated_channels/canvas/models.py
```diff
@@ -91,6 +91,14 @@ class CanvasEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguratio
         help_text="The base URL used for API requests to Canvas, i.e. https://instructure.com."
     )
 
+    refresh_token = models.CharField(
```
binodpant, line 94: About this field, we won't always have refresh token available when we create this model, it will be populated later right? so do we need to allow NULL in db?

binodpant:
So I tested the steps up until when I can save the config with the refresh token etc and validated it works!!

```

from integrated_channels.canvas.models import CanvasEnterpriseCustomerConfiguration

config = CanvasEnterpriseCustomerConfiguration(client_id='10000000000002',client_secret='bSxFhGIBmqluIMe7fIc0S6fNU75WtD7sB7r6T8euVwoYJ6eeNZIU0cfBS4VwjHwJ',refresh_token='hQZ5yGRnDHhhla8oBidomMztz1lGXJQLMEYd22FTKIBZue3Jei5vZrJmx2RfiplE',canvas_base_url='http://localhost:3000',canvas_account_id=2, enterprise_customer=customer)

config.save()
```

Next up we will need to test by trying to use this refresh token to get access tokens etc

integrated_channels/canvas/models.py
binodpant: nvm, I think empty string is better for db, but I would also get Talia's opinion on this one

binodpant:
Almost done testing this, I am able to talk to the local machine host using host.docker.internal as its hostname instead of localhost (by updating the canvas_base_url in the db)

so now I can get an access_token by doing these

```
from integrated_channels.canvas.models import CanvasEnterpriseCustomerConfiguration
from integrated_channels.canvas.client import CanvasAPIClient
from enterprise.models import EnterpriseCustomer

customer = EnterpriseCustomer.objects.all()[0]

config = CanvasEnterpriseCustomerConfiguration.objects.all()[0]
config
<CanvasEnterpriseCustomerConfiguration for Enterprise Test Enterprise>
client = CanvasAPIClient(enterprise_configuration=config)
access_token = client._get_oauth_access_token(client_id=client.enterprise_configuration.client_id,client_secret=client.enterprise_configuration.client_secret)
```


but this does not yet work I think because when I called this there was no refresh_token in the db config. I will re-start after filling that in

binodpant:
Alright, after a second round I got this to work

I did the following:

1.  Create a developer key in Canvas
2. Created an enterprise_config in the db and plugged in value of client_id, client_secret, redirect_uri to be the same values as in the developer key (see above notes)
3. Fetched the code as per Alex's instructions and saved it handy
4. Used that code to fetch token and refresh token and kept it handy
5. Saved the value of the refresh_token into the enterpriser_customer_config database row and saved it
6. Now made a requset to create course using this token in place of the 'access_token' in my postman config to create course and it worked!

```
curl --location --request POST 'http://localhost:3000/api/v1/accounts/self/courses' \
--header 'Authorization: Bearer 4yXHbrKKpCPLf4y20shreUCfZrdYu2rdAurtLOgoz7eNYM0QoZLtBGEPrvGT7Nn4' \
--form 'course[name]=Created using Alex'\''s client requested token!' \
--form 'course[default_view]=syllabus'
```

binodpant:
Screen Shot 2020-08-10 at 2 25 59 PM

binodpant:
## TESTING SUGGESTION

I manually validated just now that removing the refresh_token from the config and making it an empty string i nthe db ,raises a 400 Client Error when making the request `requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: http://host.docker.internal:3000/login/oauth2/token`

if you are adding unit tests, this may be a good case to cover, but at least we know manually this is how it behaves in this case
to the reader of the logs, a message like 'refresh_token was not set' is clearer than `400 Client Error`, in my opinion

binodpant:
But overall I am good with this PR now as long as you address Talia's notes to the extent possible, manual testing looks good

binodpant:
Also ensure to do the VERSION and CHANGELOG as per checklist above and fix the lint issue with isort

tests/test_integrated_channels/test_canvas/test_client.py
alex-sheehan-edx: > verifying what happens when you are missing parameters. Especially ones we expect to be manually entered by ECS/Customer or dev adds to a config.

Manually retested missing client params- it seems like Canvas handles the requests nicely for us. Leaving out either the client id or secret results in `client not found`, no refresh token results in a `refresh token required` error. Grant type is hard-coded so hard to think that it'd be left out, but also get a nicely formatted `grant type not supported` if it's left out.

Technically state is not required but has to match our refresh token request, so there is a possible hole (though I'd argue that if we don't have an enterprise id, we're in a much worse state than not getting the canvas integration). I can add an exception raise if there is no enterprise user associated with the canvas config.

If the canvas base URL isn't there then we won't be able to establish a connection and would result in not a great error, so let me add a sanity check for it before the request. The oauth login path is contained in the app config so I don't think it can go missing?

  > (there's a part in the client code except (KeyError, ValueError)) and I think I'm curious what our fallback is there.

I specified both of those errors because we are already raising for response errors `auth_response.raise_for_status()` which leaves the case of we get a response but it's not in the format that we expect- ie we don't have any json associated with the response or the access token isn't included in the response.

integrated_channels/canvas/client.py
alex-sheehan-edx: Haha whoops, I got mixed up and left out the step 🙃

alex-sheehan-edx:
> ### TESTING SUGGESTION
> I manually validated just now that removing the refresh_token from the config and making it an empty string i nthe db ,raises a 400 Client Error when making the request `requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: http://host.docker.internal:3000/login/oauth2/token`
> 
> if you are adding unit tests, this may be a good case to cover, but at least we know manually this is how it behaves in this case
> to the reader of the logs, a message like 'refresh_token was not set' is clearer than `400 Client Error`, in my opinion

Ack! Good catch! This is because of the `raise for status` catching Canvas' 400 response before my try/catch. I'll make the change and add a test.

PR #919, opened 2020-08-10
Canvas Metadata Exporter

taliaedX:
**Merge checklist:**
- [X] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files) **NA**
- [X] `make static` has been run to update webpack bundling if any static content was updated **NA**
- [X] `./manage.py makemigrations` has been run **NA**

**Post merge:**

### Testing Notes:
- This assumes you have Canvas enabled in edx-platform.
- This does not use/require the client.
- You will need Enterprise Catalogs if you wish to test the actual code locally. These are the notes I saved during setup:
```
- Setup Enterprise Catalog and have service running.
- Add an enterprise catalog to your enterprise customer (just do it via django admin. Use the default “all content” filter it generates if you don’t want a hassle)
- Save.
- Go to Enterprise Customer Catalogs in Django Admin > click on your catalog.
- Now just search for “Waffle” in Django Admin and you’ll find the Django-Waffle section. Click on Samples
- Click on “use_enterprise_catalog”.
- Set to 100% (this means how many users will use the catalog and for dev purposes let’s just make it easy and say EVERYONE!)
- Migrate enterprise catalog - IF - the signal `update_enterprise_catalog_data` did not run for you, which it didn’t for me because reasons. (I verified via the db)
    - In LMS shell, run: `./manage.py lms migrate_enterprise_catalogs --api_user edx`
- Create Content Metadata
    - Open Enterprise Catalog admin (`localhost:18160/admin`)
    - Click on “Content Metadata” under the Catalog header.
    - I don’t know if you need a real course here or not. However, for “content key” I used id in the `course_overviews_courseoverview` table in LMS/edxapp. I also used this as the key in the json metadata (see in a moment). I also used display_name as title in the json metadata. The rest I copied from staging catalog_content_metadata tables in the enterprise_catalog database, but only the fields I needed for my export:
        - title, number, availability, start, end, key
        - For content_type I made it a course run.
        - I also added some other things because I wasn’t sure if they are necessary. Anyways, this all goes in the json_metadata field
- Create a CanvasEnterpriseCustomerConfiguration
    - via lms shell (./manage.py lms shell)
    - from enterprise.models import EnterpriseCustomer
    - customer = EnterpriseCustomer.objects.all()[0] (unless you have more than one enterprise and want to use a specific one)
    - config = CanvasEnterpriseCustomerConfiguration(active=True, channel_worker_username=‘edx’, key=‘’, secret=‘’, canvas_company_id=‘<canvas account id>’, canvas_base_url=‘http://localhost:3000', enterprise_customer=customer, catalogs_to_transmit=‘catalog uuid’
```

### DIFF
--- a/integrated_channels/canvas/exporters/content_metadata.py
+++ b/integrated_channels/canvas/exporters/content_metadata.py
@@ -1,0 +1,56 @@
+"""
+Content metadata exporter for Canvas
+"""
+
+from logging import getLogger
+
+from integrated_channels.integrated_channel.exporters.content_metadata import ContentMetadataExporter
+
+LOGGER = getLogger(__name__)
+
+
+class CanvasContentMetadataExporter(ContentMetadataExporter):
+    """
+        Canvas implementation of ContentMetadataExporter.
+    """
+    DATA_TRANSFORM_MAPPING = {
+        'name': 'title',
+        'course_code': 'number',
+        'start_at': 'start',
+        'end_at': 'end',
+        'integration_id': 'key',
+        'syllabus_body': 'description',
+        'default_view': 'default_view',
+    }
+
+    LONG_STRING_LIMIT = 2000
+
+    def transform_description(self, content_metadata_item):
+        """
+        Return the course description and enrollment url as Canvas' syllabus body attribute.
+        This will display in the Syllabus tab in Canvas.
+        """
+        enrollment_url = content_metadata_item.get('enrollment_url', None)
+        base_description = "<a href={enrollment_url}>To edX Course Page</a><br />".format(
+            enrollment_url=enrollment_url)
+        full_description = content_metadata_item.get('full_description') or None
+        if full_description and len(full_description + enrollment_url) <= self.LONG_STRING_LIMIT:
+            description = "{base_description}{full_description}".format(
+                base_description=base_description,
+                full_description=full_description
+            )
+        else:
+            short_description = content_metadata_item.get(
+                'short_description', content_metadata_item.get('title', '')
+            )
+            description = "{base_description}{short_description}".format(
+                base_description=base_description, short_description=short_description
+            )
+
+        return description
+
+    def transform_default_view(self, content_metadata_item):  # pylint: disable=unused-argument
+        """
+        Sets the Home page view in Canvas. We're using Syllabus.
+        """
+        return 'syllabus'
--- a/integrated_channels/canvas/models.py
+++ b/integrated_channels/canvas/models.py
@@ -11,6 +11,7 @@ from simple_history.models import HistoricalRecords
 from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
 
+from integrated_channels.canvas.exporters.content_metadata import CanvasContentMetadataExporter
 from integrated_channels.integrated_channel.models import EnterpriseCustomerPluginConfiguration
 
 LOGGER = getLogger(__name__)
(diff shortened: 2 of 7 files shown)


test_utils/factories.py
```diff
@@ -610,3 +611,41 @@ class Meta:
...
+
+    class Meta:
+        """
+        Meta for ``CanvasGlobalConfiguration``.
+        """
+
+        model = CanvasGlobalConfiguration
+
+    oauth_api_path = factory.LazyAttribute(lambda x: FAKER.file_path())
+    course_api_path = factory.LazyAttribute(lambda x: FAKER.file_path())
```
binodpant, line 631: I would check with @alex-sheehan-edx if this field is needed still? I thought he moved it to appconfig

integrated_channels/canvas/exporters/content_metadata.py
```diff
@@ -0,0 +1,56 @@
...
+            short_description = content_metadata_item.get(
+                'short_description', content_metadata_item.get('title', '')
+            )
+            description = "{base_description}{short_description}".format(
+                base_description=base_description, short_description=short_description
+            )
+
+        return description
+
+    def transform_default_view(self, content_metadata_item):  # pylint: disable=unused-argument
```
binodpant, line 52: not sure how we will use this field, but does it need the `content_metadata_item` argument? it looks like it is not used right now

integrated_channels/canvas/exporters/content_metadata.py
```diff
@@ -0,0 +1,56 @@
...
+
+    LONG_STRING_LIMIT = 2000
+
+    def transform_description(self, content_metadata_item):
+        """
+        Return the course description and enrollment url as Canvas' syllabus body attribute.
+        This will display in the Syllabus tab in Canvas.
+        """
+        enrollment_url = content_metadata_item.get('enrollment_url', None)
+        base_description = "<a href={enrollment_url}>To edX Course Page</a><br />".format(
```
binodpant, line 34: I think this is okay for now, we should ask with Ken what exactly the text etc wants here ... I am good with this for now, we can also review this during team level demo if we do one of those

test_utils/factories.py
taliaedX: Yeah this will likely need rebasing off of master anyways. I just had it in there at the time because that PR had not merged.

integrated_channels/canvas/exporters/content_metadata.py
taliaedX: I think this is due to how the base class interprets these methods. The other exporters that make up fields only used in the target system (I think Success Factors was one I saw or Cornerstone...cannot remember) also have this unused argument.

> not sure how we will use this field

it's required by Canvas. This is the field that tells Canvas we want the "Home" page to be the syllabus. It needs to be sent along with the rest of the fields in the mapping.

integrated_channels/canvas/exporters/content_metadata.py
taliaedX: Yep. Absolutely did not write this on the basis this was the final language lol.

integrated_channels/canvas/models.py
```diff
@@ -133,3 +134,9 @@ def channel_code():
         Returns an capitalized identifier for this channel class, unique among subclasses.
         """
         return 'CANVAS'
+
+    def get_content_metadata_exporter(self, user):
```
binodpant, line 138: ah this reminds me I need to add a similar method on my branch to return the custom transmitter too!

taliaedX: Yeah I only realized it when my exporter was not being hit _at all_

test_utils/factories.py
taliaedX: yeah course_api_path is still there but this oauth_api_path is unnecessary

PR #923, opened 2020-08-17
ENT-3330 | Added logging in enrollment endpoint for test purposes.

hasnain-naveed:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -638,6 +638,18 @@ class EnterpriseCustomerCourseEnrollmentsSerializer(serializers.Serializer):
         cohort = validated_data.get('cohort')
         email_students = validated_data.get('email_students')
         is_active = validated_data.get('is_active')
+        LOGGER.info(
+            "[Enrollment-API] Received a call with the following parameters. lms_user: [{lms_user}], "
+            "tpa_user: [{tpa_user}], user_email: {user_email}, course_run_id: {course_run_id}, "
+            "course_mode: {course_mode}, is_active: {is_active}".format(
+                lms_user=lms_user.id if lms_user else None,
+                tpa_user=tpa_user.id if tpa_user else None,
+                user_email=user_email.id if isinstance(user_email, models.EnterpriseCustomerUser) else user_email,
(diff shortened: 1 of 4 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -638,20 +638,47 @@ def create(self, validated_data):
         cohort = validated_data.get('cohort')
         email_students = validated_data.get('email_students')
         is_active = validated_data.get('is_active')
+        LOGGER.info(
+            "[Enrollment-API] Received a call with the following parameters. lms_user: [{lms_user}], "
```
sameenfatima78, line 642: since we are passing ids in `lms_user` and `tpa_user`, can we add "id" next to them as well in the logs? so `lms_user` becomes `lms_user_id`

enterprise/api/v1/serializers.py
```diff
@@ -695,6 +736,18 @@ def create(self, validated_data):
             )
 
         validated_data['detail'] = 'success'
+        LOGGER.info(
+            "[Enrollment-API] Returning success for a call with the following parameters. lms_user: [{lms_user}], "
```
sameenfatima78, line 740: Same comment as above

enterprise/models.py
```diff
@@ -819,6 +819,25 @@ def enroll(self, course_run_id, mode, cohort=None, source_slug=None, discount_pe
         paid_modes = ['verified', 'professional']
         is_upgrading = mode in paid_modes and course_enrollment.get('mode') in audit_modes
 
+        if enrolled_in_course and is_upgrading:
+            LOGGER.info(
+                "[Enroll] Trying to upgrade a enterprise user [{learner_id}] in course [{course_run_id}] in "
```
sameenfatima78, line 824: "a" before enterprise should be removed

enterprise/api/v1/serializers.py
hasnain-naveed: It will be enterprisecutomeruser id. lms_user or tpa_user will notify from where we did get the EnterpriseCustomerUser id.

PR #917, opened 2020-08-07
ENT-3303- creation of canvas oauth-complete endpoint

alex-sheehan-edx:
**Merge checklist:**
- [x] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files)
    - `base.in` if needed in production but edx-platform doesn't install it
    - `test-master.in` if edx-platform pins it, with a matching version
    - `make upgrade && make requirements` have been run to regenerate requirements
- [x] `make static` has been run to update webpack bundling if any static content was updated
- [x] `./manage.py makemigrations` has been run
    - Checkout the [Database Migration](https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations) Confluence page for helpful tips on creating migrations.
    - *Note*: This **must** be run if you modified any models.
      - It may or may not make a migration depending on exactly what you modified, but it should still be run.
    - This should be run from either a venv with all the lms/edx-enterprise requirements installed or if you checked out edx-enterprise into the src directory used by lms, you can run this command through an lms shell.
        - It would be `./manage.py lms makemigrations` in the shell.
- [x] [Version](https://github.com/edx/edx-enterprise/blob/master/enterprise/__init__.py) bumped
- [x] [Changelog](https://github.com/edx/edx-enterprise/blob/master/CHANGELOG.rst) record added
- [ ] Translations updated (see docs/internationalization.rst but also this isn't blocking for merge atm)

**Post merge:**
- [ ] Tag pushed and a new [version](https://github.com/edx/edx-enterprise/releases) released
    - *Note*: Assets will be added automatically. You just need to provide a tag (should match your version number) and title and description.
- [ ] After versioned build finishes in [Travis](https://travis-ci.org/github/edx/edx-enterprise), verify version has been pushed to [PyPI](https://pypi.org/project/edx-enterprise/)
    - Each step in the release build has a condition flag that checks if the rest of the steps are done and if so will deploy to PyPi.
    (so basically once your build finishes, after maybe a minute you should see the new version in PyPi automatically (on refresh))
- [ ] PR created in [edx-platform](https://github.com/edx/edx-platform) to upgrade dependencies (including edx-enterprise)
    - This **must** be done after the version is visible in PyPi as `make upgrade` in edx-platform will look for the latest version in PyPi.
    - Note: the edx-enterprise constraint in edx-platform **must** also be bumped to the latest version in PyPi.

**What this PR does:**
[...]

PR #926, opened 2020-08-18
ENT-3283: Adding django admin forms for canvas integration config

alex-sheehan-edx:
**Merge checklist:**

**Post merge:**

[Jira Ticket](https://openedx.atlassian.net/secure/RapidBoard.jspa?rapidView=630&projectKey=ENT&modal=detail&selectedIssue=ENT-3283)

This commit adds a Django admin form for the Canvas integration config model. The refresh token is displayed as read only- as the user should not have to enter or alter the token in the oauth process.  

Was able to follow the steps listed at the bottom of https://github.com/edx/edx-enterprise/pull/917 and generate a refresh token associated with an account created from the admin form.


integrated_channels/canvas/admin/__init__.py
```diff
@@ -0,0 +1,41 @@
+"""
+Canvas admin integration for configuring Canvas app to communicate with Canvas systems.
```
binodpant, line 2: maybe clean up the 'Canvas' redundancy in wording here? I don't mind it as it either though

integrated_channels/canvas/admin/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+        "canvas_account_id",
+        "canvas_base_url",
+    )
+
+    readonly_fields = (
+        "enterprise_customer_name",
+        "refresh_token",
+    )
+
+    search_fields = ("enterprise_customer_name",)
```
binodpant, line 28: hmm, that last comma needed?

binodpant:
If this works, I am a go! Can't wait to have this, will make life and demos so much easier

integrated_channels/canvas/admin/__init__.py
taliaedX: Hmmm the first one is probably the best removed, but otherwise meh.

integrated_channels/canvas/admin/__init__.py
taliaedX: It's required for tuples yeah

integrated_channels/canvas/admin/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+class CanvasEnterpriseCustomerConfigurationAdmin(admin.ModelAdmin):
+    """
+    Django admin model for CanvasEnterpriseCustomerConfiguration.
+    """
+    list_display = (
+        "enterprise_customer_name",
+        "client_id",
+        "client_secret",
+        "canvas_account_id",
+        "canvas_base_url",
```
taliaedX, line 20: so this is globalconfig not enterprise customer, but how/where is the `course_api_path` field going to be set?

integrated_channels/canvas/admin/__init__.py
alex-sheehan-edx: `Admin integration for configuring Canvas app to communicate with Canvas systems.` ? There is technically the Canvas system and app...

integrated_channels/canvas/admin/__init__.py
alex-sheehan-edx: I was going off of other integrated channels were using.

integrated_channels/canvas/admin/__init__.py
alex-sheehan-edx: `canvas_base_url` is enterprise customer specific and stored in the canvas enterprise customer config. I can add another admin form for the `course_api_path` but I have more overarching questions as to why we're storing that field in the model instead of something like the app config. Does this path change? Is it user specific?

integrated_channels/canvas/admin/__init__.py
binodpant: tuples!!!! got me again

integrated_channels/canvas/admin/__init__.py
binodpant: I figure we can do without an admin for for that for now since we assume it will never change, but (everything changes in real life), so I would be down with the idea of handling this in the future (either by moving it to app config, or creating a django form)

binodpant: same applies to oauth_api_path etc that are not enterprise specific

alex-sheehan-edx: Decided that the global config that contains the course api path was entirely unnecessary so I've removed it.

PR #931, opened 2020-08-24
don't return enterprise customer if learner is not linked

muhammad-ammar:
**Description:** Currently an enterprise learner can still view the enterprise's Learner Portal if the learner was linked to the enterprise, and then unlinked. After these changes, learner will be able to view the learner portal.

**JIRA:** https://openedx.atlassian.net/browse/ENT-3251

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/filters.py
+++ b/enterprise/api/filters.py
@@ -54,3 +54,25 @@ class EnterpriseCustomerUserFilterBackend(filters.BaseFilterBackend):
             queryset = queryset.filter(user_id=request.user.id)
 
         return queryset
+
+
+class EnterpriseLinkedUserFilterBackend(filters.BaseFilterBackend):
+    """
+    Filter backend to return user's linked enterprises only
+
+    * Staff users will bypass this filter.
+    * Non-staff users will receive only their linked enterprises.
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        """
+        Filter out enterprise customer if learner is not linked
+        """
+        if not request.user.is_staff:
+            filter_kwargs = {
+                view.USER_ID_FILTER: request.user.id,
+                'enterprise_customer_users__linked': 1
+            }
+            queryset = queryset.filter(**filter_kwargs)
+
+        return queryset
--- a/tests/test_enterprise/api/test_filters.py
+++ b/tests/test_enterprise/api/test_filters.py
@@ -12,6 +12,8 @@ from django.conf import settings
 
 from test_utils import FAKE_UUIDS, TEST_EMAIL, TEST_USERNAME, APITest, factories
 
+ENTERPRISE_CUSTOMER_LIST_ENDPOINT = reverse('enterprise-customer-list')
+
 
 @ddt.ddt
 @mark.django_db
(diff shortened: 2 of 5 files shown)


tests/test_enterprise/api/test_views.py
```diff
@@ -2385,6 +2385,77 @@ def test_post_licensed_course_enrollments_license_revoke(
...
 
+    @ddt.data(
+        # API should return enterprise customer details
+        (False, True, True),
+        # API should not return enterprise customer details
+        (False, False, False),
+        # for staff, API should return enterprise customer details irrespective of linked status
+        (True, True, True),
+        # for staff, API should return enterprise customer details irrespective of linked status
+        (True, False, True),
```
bbaker6225, line 2396: Nit: It's more verbose, but I think ddt params as dictionaries (like here: https://github.com/edx/license-manager/blob/a27017cae0b357936b8a22f46802e7a0877008fe/license_manager/apps/api/v1/tests/test_views.py#L952-L990) makes it easier to understand what each test is checking

tests/test_enterprise/api/test_views.py
```diff
@@ -2385,6 +2385,77 @@ def test_post_licensed_course_enrollments_license_revoke(
...
+            'uuid': FAKE_UUIDS[0], 'name': 'Test Enterprise Customer', 'slug': TEST_SLUG,
+            'active': True, 'enable_data_sharing_consent': True,
+            'enforce_data_sharing_consent': 'at_enrollment', 'enable_portal_code_management_screen': True,
+            'enable_portal_reporting_config_screen': False,
+            'enable_portal_saml_configuration_screen': False,
+            'site__domain': 'example.com', 'site__name': 'example.com',
+            'enable_portal_subscription_management_screen': False,
+            'contact_email': 'fake@example.com',
+        }
+        enterprise_customer = factories.EnterpriseCustomerFactory(**enterprise_customer_data)
```
bbaker6225, line 2410: Not necessarily a you thing, but it would be nice to not have to specify so much explicitly

tests/test_enterprise/api/test_views.py
```diff
@@ -2385,6 +2385,77 @@ def test_post_licensed_course_enrollments_license_revoke(
...
+
+        user = factories.UserFactory(username='test_user', is_active=True, is_staff=is_staff)
```
bbaker6225, line 2412: Seems like another thing that might be good as a utility function?

bbaker6225: One small step could be making the username and password here constants in the test class data so you can just use the variable

tests/test_enterprise/api/test_views.py
```diff
@@ -2385,6 +2385,77 @@ def test_post_licensed_course_enrollments_license_revoke(
...
+
+        client = APIClient()
+        client.login(username='test_user', password='test_password')
+
+        query_params = {'slug': TEST_SLUG}
+        response = client.get(settings.TEST_SERVER + ENTERPRISE_CUSTOMER_LIST_ENDPOINT + '?' + urlencode(query_params))
+        response = self.load_json(response.content)
+
+        if has_access:
+            # import pdb ; pdb.set_trace()
```
bbaker6225, line 2430: Leftover line from testing?

tests/test_enterprise/api/test_views.py
```diff
@@ -2385,6 +2385,77 @@ def test_post_licensed_course_enrollments_license_revoke(
...
+            assert response['results'][0] == {
```
bbaker6225, line 2431: Another thing that seems like a helper would be nice for

muhammad-ammar:
@bbaker6225 Thanks for the review. I have addressed the feedback. Please have a look.

enterprise/api/filters.py
```diff
@@ -54,3 +54,24 @@ def filter_queryset(self, request, queryset, view):
...
+    """
+    Filter backend for `EnterpriseCustomerViewSet`
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        """
+        Filter out enterprise customer if learner is not linked
+        """
+        enterprise_slug = request.GET.get('slug')
+        if not request.user.is_staff and enterprise_slug:
```
bbaker6225, line 69: This seems reasonable to me, but just want to confirm that we're ok with not applying these filters to staff users

enterprise/api/filters.py
```diff
@@ -54,3 +54,24 @@ def filter_queryset(self, request, queryset, view):
             queryset = queryset.filter(user_id=request.user.id)
 
         return queryset
+
+
+class EnterpriseCustomerFilterBackend(filters.BaseFilterBackend):
```
adamstankiewicz, line 59: Would a more generic filter name make sense here? At the end of the day, this filter is for filtering out non-linked EnterpriseCustomerUsers, so perhaps the name should reflect that. That way, the filter could more easily be used for other viewsets as well where relevant.

Maybe `EnterpriseLinkedUserFilterBackend` or similar?

enterprise/api/filters.py
adamstankiewicz: The `UserFilterBackend` also only currently applies for non-staff users; I think it's fine to replicate that here.

tests/test_enterprise/api/test_views.py
```diff
@@ -2385,6 +2385,52 @@ def test_post_licensed_course_enrollments_license_revoke(
...
+        {'is_staff': False, 'is_linked_to_enterprise': True, 'has_access': True},
+        # API should not return enterprise customer details
+        {'is_staff': False, 'is_linked_to_enterprise': False, 'has_access': False},
+        # for staff, API should return enterprise customer details irrespective of linked status
+        {'is_staff': True, 'is_linked_to_enterprise': True, 'has_access': True},
+        # for staff, API should return enterprise customer details irrespective of linked status
+        {'is_staff': True, 'is_linked_to_enterprise': False, 'has_access': True},
+    )
+    @ddt.unpack
+    def test_enterprise_customer_filter(self, is_staff, is_linked_to_enterprise, has_access):
```
adamstankiewicz, line 2399: I think these tests would make sense in [test_filters.py](https://github.com/edx/edx-enterprise/blob/f872e39e18046e1de395fb1c29f1b7d38f0016c8/tests/test_enterprise/api/test_filters.py), where we have existing tests for filters (e.g., `UserFilterBackend`).

enterprise/api/filters.py
```diff
@@ -54,3 +54,24 @@ def filter_queryset(self, request, queryset, view):
...
+    """
+    Filter backend for `EnterpriseCustomerViewSet`
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        """
+        Filter out enterprise customer if learner is not linked
+        """
+        enterprise_slug = request.GET.get('slug')
```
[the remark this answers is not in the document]
muhammad-ammar: @adamstankiewicz I made this change intentionally. This endpoint is used at many places and to be on the safe side and to avoid issues I just want to do filtering on the bases of `slug` only.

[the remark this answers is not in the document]
adamstankiewicz: @muhammad-ammar I understand that rationale, but we're only fixing the issue partially if we do the filtering only when the `slug` query parameter is supplied. An unlinked learner will still have access to their previously linked EnterpriseCustomer's data through the endpoint, which is still the underlying issue here.

That said, I think it's worth fixing more broadly than only when `slug` is provided.

muhammad-ammar:
@adamstankiewicz Thanks for the review. I have addressed the feedback. Please have a look.

muhammad-ammar:
@adamstankiewicz Is it ok to merge this?

tests/test_enterprise/api/test_filters.py
```diff
@@ -118,3 +120,54 @@ def test_filter_for_list(self, username, email, is_staff, expected_data_length):
         assert response.status_code == status.HTTP_200_OK
         data = self.load_json(response.content)
         assert len(data['results']) == expected_data_length
+
+
+@ddt.ddt
+@mark.django_db
+class TestEnterpriseLinkedUserFilterBackendd(APITest):
```
adamstankiewicz, line 127: nit: typo (extra "d" at the end of `TestEnterpriseLinkedUserFilterBackendd`)

PR #918, opened 2020-08-07
Canvas transmitter

binodpant:
**Merge checklist:**
- [x] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files)
    - `base.in` if needed in production but edx-platform doesn't install it
    - `test-master.in` if edx-platform pins it, with a matching version
    - `make upgrade && make requirements` have been run to regenerate requirements
- [x] `make static` has been run to update webpack bundling if any static content was updated
- [x] `./manage.py makemigrations` has been run
    - Checkout the [Database Migration](https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations) Confluence page for helpful tips on creating migrations.
    - *Note*: This **must** be run if you modified any models.
      - It may or may not make a migration depending on exactly what you modified, but it should still be run.
    - This should be run from either a venv with all the lms/edx-enterprise requirements installed or if you checked out edx-enterprise into the src directory used by lms, you can run this command through an lms shell.
        - It would be `./manage.py lms makemigrations` in the shell.
- [x] [Version](https://github.com/edx/edx-enterprise/blob/master/enterprise/__init__.py) bumped
- [x] [Changelog](https://github.com/edx/edx-enterprise/blob/master/CHANGELOG.rst) record added
- [x] Translations updated (see docs/internationalization.rst but also this isn't blocking for merge atm)

**Post merge:**
- [ ] Tag pushed and a new [version](https://github.com/edx/edx-enterprise/releases) released
    - *Note*: Assets will be added automatically. You just need to provide a tag (should match your version number) and title and description.
- [ ] After versioned build finishes in [Travis](https://travis-ci.org/github/edx/edx-enterprise), verify version has been pushed to [PyPI](https://pypi.org/project/edx-enterprise/)
    - Each step in the release build has a condition flag that checks if the rest of the steps are done and if so will deploy to PyPi.
    (so basically once your build finishes, after maybe a minute you should see the new version in PyPi automatically (on refresh))
- [ ] PR created in [edx-platform](https://github.com/edx/edx-platform) to upgrade dependencies (including edx-enterprise)
    - This **must** be done after the version is visible in PyPi as `make upgrade` in edx-platform will look for the latest version in PyPi.
    - Note: the edx-enterprise constraint in edx-platform **must** also be bumped to the latest version in PyPi.

## Testing notes
Tested locally by

[...]

PR #932, opened 2020-08-24
ENH: Adding update and deleting capabilities to the canvas integrated…

alex-sheehan-edx:
… channels

**Merge checklist:**

**Post merge:**

First draft of using the "Fetch all courses for account" solution to implementing update and delete to the canvas integrated channel. WIP TESTING AND EDGE CASE HANDLING TO BE ADDED DO NOT MERGE YET. Some questions posed inline in comments, more to come.

To test locally:
- Follow steps to create a course through our canvas client (be able to run `./manage.py lms transmit_content_metadata --catalog_user enterprise_openedx_operator` from the lms shell and then see a course on the Canvas local instance) & (See Binod's notes?/other canvas PR's for setting up oauth and data locally)
   - This is a rather large step to simply say "do it" so if you have any questions, I would refer to https://github.com/edx/edx-enterprise/pull/918 (The PR that implements the base transmitter + post capabilities) or https://github.com/edx/edx-enterprise/pull/926 and https://github.com/edx/edx-enterprise/pull/917 (creating Canvas integration configs and setting up + completing the oauth process) 
- In the enterprise catalog admin, change the content metadata associated with the posted course.
   - to test an update, all you have to do is change something like the `full_description` within the `json metadata`
Screen Shot 2020-08-24 at 6 09 26 PM

   - to test a delete, all you have to do is change the `content key` and associated `key` field within the json metadata to be different. The way I do this is increment the int at the end of the key.
 
Screen Shot 2020-08-24 at 6 12 27 PM

Once either of these things are done and the settings are saved, you can run the above `./manage.py lms transmit_content_metadata --catalog_user enterprise_openedx_operator` command from the lms shell and it should trigger either an update or delete respectfully.

### DIFF
--- a/integrated_channels/canvas/client.py
+++ b/integrated_channels/canvas/client.py
@@ -2,12 +2,14 @@
 """
 Client for connecting to Canvas.
 """
+import json
+
 import requests
 from six.moves.urllib.parse import urljoin  # pylint: disable=import-error
 
 from django.apps import apps
 
-from integrated_channels.exceptions import ClientError
+from integrated_channels.exceptions import CanvasClientError, ClientError
 from integrated_channels.integrated_channel.client import IntegratedChannelApiClient
 
 
@@ -55,19 +57,39 @@ class CanvasAPIClient(IntegratedChannelApiClient):
         pass
 
     def create_content_metadata(self, serialized_data):
-        url = '{}/api/v1/accounts/{}/courses'.format(
+        self._create_session()
+
+        url = '{}/api/v1/accounts/{}/courses/'.format(
             self.enterprise_configuration.canvas_base_url,
             self.enterprise_configuration.canvas_account_id,
         )
-        self._post(url, serialized_data)
+        return self._post(url, serialized_data)
 
     def update_content_metadata(self, serialized_data):
-        # Cannot update yet since we don't have course id
-        pass
+        self._create_session()
+
+        integration_id = self._get_integration_id_from_transmition_data(serialized_data)
+        course_id = self._get_course_id_from_integration_id(integration_id)
+
+        url = '{}/api/v1/courses/{}'.format(
+            self.enterprise_configuration.canvas_base_url,
+            course_id,
+        )
+
+        return self._put(url, serialized_data)
 
     def delete_content_metadata(self, serialized_data):
-        # Cannot delete yet since we don't have course id
-        pass
+        self._create_session()
+
+        integration_id = self._get_integration_id_from_transmition_data(serialized_data)
+        course_id = self._get_course_id_from_integration_id(integration_id)
+
+        url = '{}/api/v1/courses/{}'.format(
+            self.enterprise_configuration.canvas_base_url,
+            course_id,
+        )
+
+        return self._delete(url)
 
     def _post(self, url, data):
         """
@@ -75,23 +97,82 @@ class CanvasAPIClient(IntegratedChannelApiClient):
 
         Args:
             url (str): The url to send a POST request to.
-            data (str): The json encoded payload to POST.
+            data (bytearray): The json encoded payload to POST.
         """
-        self._create_session()
-        response = self.session.post(url, data=data)
-        return response.status_code, response.text
+        post_response = self.session.post(url, data=data)
+        post_response.raise_for_status()
+        return post_response.status_code, post_response.text
 
-    def _delete(self, url, data):
+    def _put(self, url, data):
         """
-        Make a DELETE request using the session object to a Canvas endpoint.
+        Make a PUT request using the session object to the Canvas course update endpoint
 
         Args:
-            url (str): The url to send a DELETE request to.
-            data (str): The json encoded payload to DELETE.
+            url (str): The canvas url to send update requests to.
+            data (bytearray): The json encoded payload to UPDATE. This also contains the integration
+            ID used to match a course with a course ID.
         """
-        self._create_session()
-        response = self.session.delete(url, data=data)
-        return response.status_code, response.text
+        put_response = self.session.put(url, data=data)
+        put_response.raise_for_status()
+        return put_response.status_code, put_response.text
+
+    def _delete(self, url):
+        """
+        Make a DELETE request using the session object to the Canvas course delete endpoint.
+
+        Args:
+            url (str): The canvas url to send delete requests to.
+        """
+        delete_response = self.session.delete(url, data='{"event":"delete"}')
+        delete_response.raise_for_status()
+
+        return delete_response.status_code, delete_response.text
+
+    def _get_integration_id_from_transmition_data(self, data):
+        """
+        Retrieve the integration ID string from the encoded transmission data and apply appropriate
+        error handling.
+
+        Args:
+            data (bytearray): The json encoded payload intended for a Canvas endpoint.
+        """
+        if not data:
+            raise CanvasClientError("No data to transmit.")
+        try:
+            integration_id = json.loads(
+                data.decode("utf-8")
+            )['course']['integration_id']
+        except KeyError:
+            raise CanvasClientError("Could not transmit data, no integration ID present.")
+        except AttributeError:
+            raise CanvasClientError("Unable to decode data.")
+
+        return integration_id
+
+    def _get_course_id_from_integration_id(self, integration_id):
+        """
+        To obtain course ID we have to request all courses associated with the integrated
+        account and match the one with our integration ID.
+
+        Args:
+            integration_id (string): The ID retrieved from the transmission payload.
+        """
+        url = "{}/api/v1/accounts/{}/courses/".format(
+            self.enterprise_configuration.canvas_base_url,
+            self.enterprise_configuration.canvas_account_id
+        )
+        all_courses_response = self.session.get(url).json()
+        course_id = None
+        for course in all_courses_response:
+            if course['integration_id'] == integration_id:
+                course_id = course['id']
+                break
+
+        if not course_id:
+            raise CanvasClientError("No Canvas courses found with associated integration ID: {}.".format(
+                integration_id
+            ))
+        return course_id
 
     def _create_session(self):
         """
--- a/tests/test_integrated_channels/test_canvas/test_client.py
+++ b/tests/test_integrated_channels/test_canvas/test_client.py
@@ -4,6 +4,7 @@ Tests for clients in integrated_channels.
 """
 
 import datetime
+import json
 import unittest
 
 import pytest
@@ -14,7 +15,7 @@ from six.moves.urllib.parse import urljoin  # pylint: disable=import-error
 from django.utils import timezone
 
 from integrated_channels.canvas.client import CanvasAPIClient
-from integrated_channels.exceptions import ClientError
+from integrated_channels.exceptions import CanvasClientError, ClientError
 from test_utils import factories
 
 NOW = datetime.datetime(2017, 1, 2, 3, 4, 5, tzinfo=timezone.utc)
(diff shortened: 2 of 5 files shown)


integrated_channels/canvas/client.py
```diff
@@ -90,9 +109,36 @@ def _delete(self, url, data):
...
+        integration_id = json.loads(
+            data.decode("utf-8")
+        )['course']['integration_id']
+        course_id = self._get_course_id_from_integration_id(integration_id)
+
+        url = '{}/api/v1/courses/{}'.format(
+            self.enterprise_configuration.canvas_base_url,
+            course_id,
+        )
+        response = self.session.delete(url, data='{"event":"delete"}')
```
alex-sheehan-edx, line 122: Apparently it's not good enough to simply make a `DELETE` request to canvas, you also have to supply this data because they've also included `concluding` a course within their `DELETE` endpoint... Maybe this is standard practice and I'm biased against this API at this point 😒

binodpant: yup, you have to ask for 'delete' or 'conclude' event. I doubt we ever want a conclude, so delete should be out default mode https://community.canvaslms.com/t5/Instructor-Guide/How-do-I-delete-a-course-as-an-instructor/ta-p/1181 also apparently it is possible to recover 'deleted' courses, so canvas itself looks like it is doing a soft delete. I think so.

integrated_channels/canvas/client.py
```diff
@@ -46,6 +48,8 @@ def __init__(self, enterprise_configuration):
         super(CanvasAPIClient, self).__init__(enterprise_configuration)
         self.config = apps.get_app_config('canvas')
         self.session = None
+        # Wondering if we want to keep track of a timeout on the access token to recall "get auth token"
+        self.session_time_out = None
```
binodpant, line 52: not sure why this is here but perhaps we can investigate this separately? I think you meant what happens if the access token expires? I guess we will have to handle that somehow by refetching a new token, it would be cool for us to create a ticket for that if you are inclined.

integrated_channels/canvas/client.py
```diff
@@ -81,7 +83,24 @@ def _post(self, url, data):
         response = self.session.post(url, data=data)
         return response.status_code, response.text
```
binodpant, line 85: does pylint ask for two lines here or one? I forget. I see above have two lines ... be good to be consistent I guess

alex-sheehan-edx: Where are you seeing the double line? I thought pylint required it for when you define classes, class methods seem to be single lined unless I'm missing something?

integrated_channels/canvas/client.py
```diff
@@ -90,9 +109,36 @@ def _delete(self, url, data):
...
+            self.enterprise_configuration.canvas_base_url,
+            self.enterprise_configuration.canvas_account_id
+        )
+        all_courses_response = self.session.get(url).json()
+
+        for course in all_courses_response:
+            if course['integration_id'] == integration_id:
+                # raise if we find two matching?
+                # break if we find a matching integration id?
+                course_id = course['id']
```
binodpant, line 139: hmm, wouldn't we want to end the loop once we find the course_id? ... unless we want to find duplicates etc which I think is not needed since Canvas already controls uniqueness of integration_id, I am pretty sure. But please do confirm that .. if we do then we can early return rather than loop the rest of the items

integrated_channels/canvas/client.py
```diff
@@ -81,7 +83,24 @@ def _post(self, url, data):
...
-    def _delete(self, url, data):
+    def _put(self, data):
+        """
+        doc TBD
+        """
+        self._create_session()
+        integration_id = json.loads(
+            data.decode("utf-8")
+        )['course']['integration_id']
+        course_id = self._get_course_id_from_integration_id(integration_id)
```
binodpant, line 94: I think that in either of the following cases, we would like to raise a 400 Bad Request. A test can nicely cover these too

data is empty/None

data.decode fails and/or
  data.decode() does not contain 'course'->'integration_id'

course_id is None or cannot be extracted somehow

integrated_channels/canvas/client.py
binodpant: also check what the response of delete is in different situations... such as (1) course does not exist (2) course could not be deleted etc.

integrated_channels/canvas/client.py
binodpant: guess I was dreaming lol! sorry

integrated_channels/canvas/client.py
alex-sheehan-edx: Yup, I wasn't sure what the possibilities were in terms of if we could get dups, so initially I added the check. I am now more sure that it will never happen so I added a break 👍

tests/test_integrated_channels/test_canvas/test_client.py
```diff
@@ -86,3 +89,144 @@ def test_client_instantiation_fails_without_refresh_token(self):
...
+                                               'Could not transmit data, no integration ID present.'
+
+    def test_course_delete_fails_when_course_id_not_found(self):
+        self.update_fails_when_course_id_not_found("_delete")
+
+    def test_course_update_fails_when_course_id_not_found(self):
+        self.update_fails_when_course_id_not_found("_put")
+
+    def update_fails_when_course_id_not_found(self, request_type):
+        course_to_update = '{{"course": {{"integration_id": "{}", "name": "test_course"}}}}'.format(
```
binodpant, line 170: this is just me but is there a way to make this json formatting in the code clearner? no biggie but if you have a quick way to do so, multiline string?

tests/test_integrated_channels/test_canvas/test_client.py
```diff
@@ -86,3 +89,144 @@ def test_client_instantiation_fails_without_refresh_token(self):
             canvas_api_client = CanvasAPIClient(self.enterprise_config)
             canvas_api_client._create_session()  # pylint: disable=protected-access
         assert client_error.value.__str__() == "Failed to generate oauth access token: Refresh token required."
+
+    def test_course_delete_fails_with_empty_data(self):
```
binodpant, line 93: hmm, are the empty data cases actual failures? I think generally the transmitter should not call with empty data but if it does, is it just a 'meh, no data' vs an error ?

alex-sheehan-edx: https://github.com/edx/edx-enterprise/pull/932#discussion_r476435593

> data is empty/None

The edge case handling here was in response to this comment, which for the most part I get the sense is impossible/incredibly unlikely to happen but I thought it was a relatively small error check and why not add it to be careful  `¯\_(ツ)_/¯`

binodpant: maybe an error is fine, just wondering if this empty data case can ever even occur .. does not hurt to add this protection though, just wondering about error vs pass ... error maybe slightly better since we will get the logging in this case at least ... just want to hear your thoughts, but no issues here, resolving this convo ...

tests/test_integrated_channels/test_canvas/test_client.py
alex-sheehan-edx: I can clean it up a bit 👍  it's kinda messy

tests/test_integrated_channels/test_canvas/test_client.py
```diff
@@ -86,3 +89,144 @@ def test_client_instantiation_fails_without_refresh_token(self):
...
+                    json={'access_token': self.access_token},
+                    status=200
+                )
+                transmitter_method = getattr(canvas_api_client, request_type)
+                transmitter_method(course_to_update)
+
+        assert client_error.value.__str__() == 'Canvas Client Error: ' \
+                                               'No Canvas courses found with associated integration ID.'
+
+    def test_successful_client_update(self):
```
binodpant, line 198: I think it would be good to find a way to also call the actual public client level method (non private) in the tests, even though today we are getting full coverage. methods like `create_content_metadata`, `delete_content_metadata` etc (esp if they get changed in the future, then your tests will automatically cover them). just a thought. non blocking.

integrated_channels/canvas/client.py
```diff
@@ -62,36 +64,110 @@ def create_content_metadata(self, serialized_data):
...
+        )
+        all_courses_response = self.session.get(url).json()
+        course_id = None
+        for course in all_courses_response:
+            if course['integration_id'] == integration_id:
+                course_id = course['id']
+                break
+
+        if not course_id:
+            raise CanvasClientError("No Canvas courses found with associated integration ID.")
```
binodpant, line 169: nit: ah also to those reading these logs (or newrelic etc) it will be awesome to have the integration_id in the logs

integrated_channels/canvas/client.py
alex-sheehan-edx: Current iteration is raising Canvas' error response. However we may want to change this to be more clear in the future?

integrated_channels/canvas/client.py
alex-sheehan-edx: Absolutely!

tests/test_integrated_channels/test_canvas/test_client.py
alex-sheehan-edx: Since as of right now, all the delete and update content metadata methods are just passing the data to the respective delete and put methods, I think this should be possible.

alex-sheehan-edx: The only issue with this and the success case is that the method doesn't return any response or code on success so there's nothing to assert in the test. I've made the change, could you take one last look and make sure you agree with the process?

binodpant: superb!

integrated_channels/canvas/client.py
```diff
@@ -55,43 +57,119 @@ def delete_course_completion(self, user_id, payload):  # pylint: disable=unused-
...
+        url = '{}/api/v1/accounts/{}/courses/'.format(
             self.enterprise_configuration.canvas_base_url,
             self.enterprise_configuration.canvas_account_id,
         )
         self._post(url, serialized_data)
 
     def update_content_metadata(self, serialized_data):
-        # Cannot update yet since we don't have course id
-        pass
+        self._put(serialized_data)
```
taliaedX, line 67: what's the point of making this a separate function? I know `update_content_metadata` is one we are overwriting from the base class, but why overwrite with another defined method rather than just stick the logic here?

1. We aren't using `_put` outside of this one context anyways (same with delete and post)
2. The incredibly generic name makes it seem like it _could_ be usable for learner data but it can't because it does a bunch of course specific logic.

I'd recommend moving the logic to these main classes and maybe use put/post/delete to pass in the specific url and just have it do requests.post/put/delete ? Or just not have those quite frankly?

taliaedX: also this goes for _delete and _post as well

tests/test_integrated_channels/test_canvas/test_client.py
```diff
@@ -86,3 +90,154 @@ def test_client_instantiation_fails_without_refresh_token(self):
...
             canvas_api_client._create_session()  # pylint: disable=protected-access
         assert client_error.value.__str__() == "Failed to generate oauth access token: Refresh token required."
+
+    def test_course_delete_fails_with_empty_data(self):
+        self.transmission_with_empty_data("delete_content_metadata")
+
+    def test_course_update_fails_with_empty_data(self):
+        self.transmission_with_empty_data("update_content_metadata")
+
+    def transmission_with_empty_data(self, request_type):
```
taliaedX, line 100: I get why you structured the helpers in this way, but I think it'd be better if they were all together at the top of the file. Python allows for functions to be defined in any order, but it's still a bit weird to have the helpers after the functions calling them. It's also a little weird seeing them mixed in between tests.

alex-sheehan-edx: fair points, I can reorder.

integrated_channels/canvas/client.py
alex-sheehan-edx: I agree that `_put`, `_post` and `_delete` have grown out of their generic context. I'm going to revert all three to accepting a url and data and returning the response of their respective session calls. All other logic will go in the `update_content_metadata`, `delete_content_metadata` and `create_content_metadata`.

PR #937, opened 2020-08-26
Moodle Client Updates

jlajoie:
**Merge checklist:**

**Post merge:**


integrated_channels/moodle/client.py
```diff
@@ -5,6 +5,7 @@
 
 import json
 import requests
+from six.moves.urllib.parse import urljoin  # pylint: disable=import-error
```
taliaedX, line 8: hmmm why this specific import path instead of just `from urllib.parse`? Especially curious since there is an import from urllib.parse already in this file (I imported urlencode because I wanted the base functionality not Django's wrapped version)

integrated_channels/moodle/client.py
```diff
@@ -36,10 +51,13 @@ def __init__(self, enterprise_configuration):
         """
         super(MoodleAPIClient, self).__init__(enterprise_configuration)
         self.config = apps.get_app_config('moodle')
+        self.token = self._get_access_token()
```
taliaedX, line 54: Hmmm. Given the tokens seem to persist on the moodle side, _should_ we be getting a new token every time instead of persisting it? Would we not end up possibly hitting some data limit or just otherwise polluting the web service token screen with 8000 tokens? lol

taliaedX: secondly: mostly a note for myself, but if we go this route, I need to update my parameters as it's using `self.enterprise_configuration.api_token` in several spots.

integrated_channels/moodle/client.py
```diff
@@ -116,3 +136,31 @@ def delete_content_metadata(self, serialized_data):
             params=params
         )
         return response.status_code, response.text
+
+    def _get_access_token(self):
+        querystring = {
+            'service': self.enterprise_configuration.service_short_name
```
taliaedX, line 142: Is this only required for token requests? Mostly just curious since I have never included it, but since tokens seem to be assigned per webservice that might make sense.

integrated_channels/moodle/client.py
jlajoie: Bad merge and shamelessly stolen from elsewhere, I'll update and simplify

integrated_channels/moodle/client.py
jlajoie: So the way I've seen it is it doesn't actually create a new token every time, if you spam this auth endpoint as the same user you'll get the same token returned.

Let me hold off on merging this, I'll see what it's like with a TTL set on token expiry just to verify that's true in that scenario as well, good catch here.

integrated_channels/moodle/client.py
jlajoie: Yup! This'll give you the token for a service with specific functions/permissions. That token is then used in Moodle elsewhere (and I'm guessing they resolve the token to a service to get perms) so you don't need to specify that name in other requests.

jlajoie:
Also realizing we should include the categoryid in here as well, going to update migrations

integrated_channels/moodle/client.py
taliaedX: oh yeah I forgot it returns the same token. I had that happen too when we tried testing this a while back.

taliaedX:
>Also realizing we should include the categoryid in here as well, going to update migrations

Hmmm. Should we? Wouldn't the default be the same for every customer? Or did you check with Ken about custom categories?

integrated_channels/moodle/client.py
taliaedX: I also wonder if this wrapper could just pass token or username/password and the base url in the format of `base_params = {'moodlewsrestformat': 'json', 'wstoken': self.token, 'wsusername', [etc]}` since you might see I have to use those same values in every single request, but that's something I can also play around with outside of this particular pr.

integrated_channels/moodle/migrations/0001_initial.py
```diff
@@ -1,22 +1,20 @@
-# Generated by Django 2.2.15 on 2020-08-20 16:06
+# Generated by Django 2.2.14 on 2020-08-27 20:11
```
taliaedX, line 1: look at you using an old version of Django. :p

jlajoie: This is with a brand new install in my venv too!

jlajoie:
@taliaedX going to do some smoke testing with everything you have up then merge this into your branch if things look good

taliaedX:
I'm working on fixing tests, but @alex-sheehan-edx I'd love a review whenever you have time and @jlajoie I know it's your PR, but I'd appreciate you looking over the latest commit changes to double-check my changes.

taliaedX:
Couple side notes:
1. I'm already planning on cleaning up any commented out bits, unnecessary comments, print statements, etc. as part of style fixes.
2. None of the merge checklist is relevant here since it's being merged into my PR not master.

integrated_channels/moodle/client.py
```diff
@@ -36,31 +66,43 @@ def __init__(self, enterprise_configuration):
...
+            ),
             params=params
         )
+        parsed_response = json.loads(response.text)
+        if not parsed_response.get('courses'):
+            raise ClientError('MoodleAPIClient request failed: 404 Course key '
+                              '"{}" not found in Moodle.'.format(key))
 
-        return json.loads(response.text)['courses'][0]['id']
+        return parsed_response[0]['id']
```
jlajoie, line 103: Might want to add some handling around array length and key existence. You'll also not want the moodle_request_wrapper on this if you're not returning a raw response

jlajoie (commented):
Just the one comment but looks good

alex-sheehan-edx (approved):
Took a high level first pass. Looks pretty good to me, but talked with @taliaedX and will probably do a more in-depth review of the main PR for moodle work.

integrated_channels/moodle/client.py
```diff
@@ -74,24 +116,24 @@ def create_content_metadata(self, serialized_data):
           [...]
         }
         """
-
-        serialized_data['wstoken'] = self.enterprise_configuration.api_token
+        # http://localhost:80/webservice/rest/server.php?wsfunction=core_course_create_courses&moodlewsrestformat=json&courses[0][fullname]=YourCourseFullName&courses[0][shortname]=YourCourseShortName&courses[0][categoryid]=1
```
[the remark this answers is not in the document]
taliaedX: ah I missed that one. Good catch

integrated_channels/moodle/models.py
```diff
@@ -29,38 +29,57 @@ class MoodleEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguratio
...
+    category_id = models.IntegerField(
+        blank=True,
+        null=True,
+        verbose_name="Category ID",
+        help_text=_(
+            "The category ID for what edX courses should be associated with."
+        )
     )
 
-    wsusername = models.CharField(
```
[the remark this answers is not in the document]
taliaedX: I named them wsusername and wspassword since those are the actual parameter names in Moodle's api.
My guess is Jeff changed them since these field names would be visible in Django admin and might confuse support or something idk.

@jlajoie why you hatin' on my field names?

PR #938, opened 2020-08-27
ENT-3388: post course image to canvas once course created

binodpant:
Do an extra PUT to send course image once course is created (only done if image_url is present in content_metadata item)

**Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,7 @@ Change Log
 Unreleased
 --------------------
 
-[3.6.10] 2020-08-28
+[3.7.1] 2020-08-28
 -------------------
 
 * Also send course image_url to Canvas when creating course.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,7 +24,6 @@ Unreleased
 
 * Fixed Duplicate Calls to OCN API.
 
-
 [3.6.9] 2020-08-26
 -------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,10 +14,16 @@ Change Log
 Unreleased
 --------------------
 
+[3.6.10] 2020-08-28
+-------------------
+
+* Also send course image_url to Canvas when creating course.
+
 [3.7.0] 2020-08-27
 -------------------
 
 * Fixed Duplicate Calls to OCN API.
+
 
 [3.6.9] 2020-08-26
 -------------------
--- a/integrated_channels/canvas/client.py
+++ b/integrated_channels/canvas/client.py
@@ -100,7 +100,8 @@ class CanvasAPIClient(IntegratedChannelApiClient):
                 self._put(url, json.dumps({
                     'course' : {'image_url': content_metadata['image_url']}
                 }).encode('utf-8'))
-        except: # we do not want course image update to cause failures
+        except Exception:
+            # we do not want course image update to cause failures
             pass
 
         return status_code, response_text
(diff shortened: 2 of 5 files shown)


integrated_channels/canvas/client.py
```diff
@@ -59,19 +79,39 @@ def delete_course_completion(self, user_id, payload):  # pylint: disable=unused-
...
 
-        url = '{}/api/v1/accounts/{}/courses/'.format(
+        # step 1: create the course
+        url = CanvasAPIClient.course_create_endpoint(
             self.enterprise_configuration.canvas_base_url,
-            self.enterprise_configuration.canvas_account_id,
+            self.enterprise_configuration.canvas_account_id
         )
-        return self._post(url, serialized_data)
+        status_code, response_text = self._post(url, serialized_data)
```
taliaedX, line 87: what if the status code here isn't a 200? Ie 400/404/500, etc.

tests/test_integrated_channels/test_canvas/test_client.py
```diff
@@ -56,6 +56,7 @@ def setUp(self):
         )
         self.integration_id = 'course-v1:test+TEST101'
 
+
```
taliaedX, line 59: unnecessary line addition?

integrated_channels/canvas/client.py
binodpant: ah yeah, so we just `raise_for_status` already in the _post() method which is then handled up by transmitter layer in terms of reporting, but without causing an interruption of flow. will that work you think?

CHANGELOG.rst
```diff
@@ -14,11 +14,17 @@ Change Log
...
+-------------------
+
+* Also send course image_url to Canvas when creating course.
+
 [3.7.0] 2020-08-27
 -------------------
 
 * Fixed Duplicate Calls to OCN API.
 
+
```
alex-sheehan-edx, line 27: Any reason for this new line?

integrated_channels/canvas/client.py
```diff
@@ -59,19 +79,39 @@ def delete_course_completion(self, user_id, payload):  # pylint: disable=unused-
...
+
+        # step 2: upload image_url if present
+        try:
```
alex-sheehan-edx, line 90: I forget, does the image url get updated in the single update request put? Whereas here I see we're making the second call.

CHANGELOG.rst
binodpant: hah nope I am dumb here

integrated_channels/canvas/client.py
binodpant: line 100 does it

binodpant: the first post simply ignores even if the image_url is there

alex-sheehan-edx: sorry- was unclear. I was referring to when content metadata is updated, did we figure out that the it only requires the one request?

binodpant: oh yea the update endpoint does handle it, TBH I did not test update but I am certain the endpoint (since it's the same I used) will update image when a change is presnet

binodpant: the issue is only with the create endpoint

integrated_channels/canvas/client.py
taliaedX: I think my concern is if the course fails to be created, why are we even attempting to upload an image for it?

taliaedX: the id not existing would throw an exception when doing the `created_course_id` assignment, however, because that's in a try block already we'd get it picked up. I'm a little unsure of whether we're going to silently fail both of these tasks though since I'm not actually sure we log status_code and response_text anywhere.

PR #940, opened 2020-09-01
ENT-3250: add support for pending enterprise admin users

adamstankiewicz:
[ENT-3250](https://openedx.atlassian.net/browse/ENT-3250)

This PR adds a new model `PendingEnterpriseCustomerAdminUser` to allow Enterprise Admins to use an enterprise-specific registration link which will automatically link the user to their Enterprise and assign the "enterprise_admin" system-wide role. The "enterprise_admin" role is only granted for users who have a matching `PendingEnterpriseCustomerAdminUser` record based on the user's email.

Also, adds an inline form to the EnterpriseCustomer model for ease of access for ECS:
image

Related PRs:
* https://github.com/edx/frontend-app-admin-portal/pull/327
* https://github.com/edx/edx-internal/pull/2997

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -2304,3 +2302,81 @@ def __repr__(self):
         Return uniquely identifying string representation.
         """
         return self.__str__()
+
+
+class PendingEnterpriseCustomerAdminUser(TimeStampedModel):
+    # pylint: disable=line-too-long
```
bbaker6225, line 2308: nit: Do we need this? Can we do a line break for the long line?

enterprise/models.py
```diff
@@ -2304,3 +2302,81 @@ def __repr__(self):
...
+    .. pii_retirement: local_api, consumer_api
+    """  # pylint: enable=line-too-long
+
+    enterprise_customer = models.ForeignKey(
+        EnterpriseCustomer,
+        blank=False,
+        null=False,
+        on_delete=models.CASCADE,
+    )
+    user_email = models.EmailField(null=False, blank=False, unique=True)
```
bbaker6225, line 2323: Do you want to remove the `unique=True` from here since we're doing a unique_together on customer and email?

enterprise/models.py
```diff
@@ -2304,3 +2302,81 @@ def __repr__(self):
...
+        the PendingEnterpriseCustomerAdminUser record is removed.
+
+        Arguments:
+            user: a User instance
+            role: a SystemWideEnterpriseRole instance
+        """
+        try:
+            pending_admin_user = PendingEnterpriseCustomerAdminUser.objects.get(
+                user_email=user.email,
+            )
```
bbaker6225, line 2356: Couldn't there be multiple of these since user_email is unique with the customer? (This might be resolved by the question on that uniqueness)

enterprise/models.py
```diff
@@ -2304,3 +2302,81 @@ def __repr__(self):
...
+            self.enterprise_customer.slug
+        )
+        return registration_url
+
+    def activate_admin_permissions(self, user, role):
+        """
+        Activates admin permissions for an existing PendingEnterpriseCustomerAdminUser.
+
+        Specifically, the "enterprise_admin" system-wide role is assigned to the user and
+        the PendingEnterpriseCustomerAdminUser record is removed.
```
bbaker6225, line 2347: Feels a little weird to me to pass in a role as an argument when the comment says we are always assigning this particular role

enterprise/signals.py
```diff
@@ -106,6 +107,58 @@ def default_content_filter(sender, instance, **kwargs):     # pylint: disable=un
...
 
+@receiver(post_save, sender=EnterpriseCustomerUser)
+def assign_or_delete_enterprise_admin_role(sender, instance, **kwargs):     # pylint: disable=unused-argument
+    """
+    Assign or delete enterprise_admin role for EnterpriseCustomerUser when updated.
+
+    This only occurs if a PendingEnterpriseCustomerAdminUser record exists.
+    """
+    if instance.user:
+        enterprise_admin_role, __ = SystemWideEnterpriseRole.objects.get_or_create(name=ENTERPRISE_ADMIN_ROLE)
```
bbaker6225, line 118: Wouldn't this create an admin role for any `EnterpriseCustomerUser` that gets saved?

enterprise/signals.py
```diff
@@ -106,6 +107,58 @@ def default_content_filter(sender, instance, **kwargs):     # pylint: disable=un
...
+        try:
+            pending_enterprise_admin_user = PendingEnterpriseCustomerAdminUser.objects.get(
+                user_email=instance.user.email,
+            )
```
bbaker6225, line 122: Same uniqueness question

enterprise/signals.py
bbaker6225: Never mind, took me a second to realize that we weren't assigning it to the user yet

tests/test_enterprise/test_signals.py
```diff
@@ -246,11 +248,213 @@ def test_default_content_filter(self, default_content_filter, expected_content_f
...
+    Test signals associated with PendingEnterpriseCustomerAdminUser.
+    """
+
+    def setUp(self):
+        """
+        Setup for `TestPendingEnterpriseAdminUserSignals` test.
+        """
+        self.admin_user = UserFactory(id=2, email='user@example.com')
+        self.enterprise_customer = EnterpriseCustomerFactory()
+        super(TestPendingEnterpriseAdminUserSignals, self).setUp()
```
bbaker6225, line 263: This can probably all go in the class method `setUpTestData` to only be run once

tests/test_enterprise/test_signals.py
```diff
@@ -246,11 +248,213 @@ def test_default_content_filter(self, default_content_filter, expected_content_f
...
+
+    def test_create_pending_enterprise_admin_user(self):
```
bbaker6225, line 265: nit: small thing, but I feel like the test naming (or the addition of a docstring) could make it a little more explicit that we're testing that creating the `PendingEnterpriseCustomerAdminUser` results in a non-admin customer user also being created

tests/test_enterprise/test_signals.py
```diff
@@ -246,11 +248,213 @@ def test_default_content_filter(self, default_content_filter, expected_content_f
...
+        # associated PendingEnterpriseCustomerUser is also deleted.
+        PendingEnterpriseCustomerAdminUser.objects.filter(
+            user_email=self.admin_user.email,
+            enterprise_customer=self.enterprise_customer,
+        ).delete()
+        pending_ecus = PendingEnterpriseCustomerUser.objects.filter(
+            user_email=self.admin_user.email,
+            enterprise_customer=self.enterprise_customer,
+        )
+        self.assertFalse(pending_ecus.exists())
```
bbaker6225, line 310: nit: could potentially refactor this assert does exist/does not exist check into a helper to reduce some duplication

tests/test_enterprise/test_signals.py
```diff
@@ -246,11 +248,213 @@ def test_default_content_filter(self, default_content_filter, expected_content_f
...
+
+
+@mark.django_db
+@ddt.ddt
+class TestEnterpriseAdminRoleSignals(unittest.TestCase):
+    """
+    Test signals associated with EnterpriseCustomerUser and the enterprise_admin role.
+    """
+
+    def setUp(self):
```
bbaker6225, line 320: same thing about setUpTestData

tests/test_enterprise/test_signals.py
```diff
@@ -246,11 +248,213 @@ def test_default_content_filter(self, default_content_filter, expected_content_f
...
+
+        # verify that a new admin user role assignment is created when appropriate.
+        admin_role_assignment = SystemWideEnterpriseUserRoleAssignment.objects.filter(
+            user=self.admin_user,
+            role=self.enterprise_admin_role,
+        )
+        if has_pending_admin_user:
+            self.assertTrue(admin_role_assignment.exists())
+        else:
+            self.assertFalse(admin_role_assignment.exists())
```
bbaker6225, line 372: nit: Could do `assert admin_role_assignment.exists() == has_pending_admin_user`? No strong preference

tests/test_enterprise/test_signals.py
```diff
@@ -246,11 +248,213 @@ def test_default_content_filter(self, default_content_filter, expected_content_f
...
+
+        # update EnterpriseCustomerUser record.
+        enterprise_customer_user = EnterpriseCustomerUser.objects.get(
+            user_id=self.admin_user.id
+        )
+        if should_unlink_user:
+            enterprise_customer_user.linked = False
+        else:
+            enterprise_customer_user.active = False
+        enterprise_customer_user.save()
```
bbaker6225, line 402:
```
enterprise_customer_user.linked = should_unlink_user
enterprise_customer_user.active = should_unlink_user
enterprise_customer_user.save()
```
maybe?

enterprise/models.py
adamstankiewicz: I passed the role as an argument to avoid another DB call to retrieve the same role, but I get the confusion with the comment as written. I'm OK with either approach.

enterprise/models.py
adamstankiewicz: I wasn't sure if the PII annotations with a line break would causes problems on the PII annotation linting that happens, so I copied this pattern from an existing docstring with similar PII annotations. I'll give it a shot with a line break and see if CI fails, though.

enterprise/models.py
adamstankiewicz: Ah, yes. I meant to take that out when I added in the `unique_together` 😄 Good catch.

tests/test_enterprise/test_signals.py
adamstankiewicz: Yup, missed adding the docstring here 😄 Test name reflects the name of the signal its testing.

enterprise/models.py
bbaker6225: Ah ok makes sense. I'm fine with either as well

enterprise/models.py
```diff
@@ -2328,3 +2331,84 @@ def __repr__(self):
...
+    @cached_property
+    def admin_registration_url(self):
+        """
+        Returns a URL to be used by a pending enterprise admin user to register their account.
+        """
+        registration_url = '{}/{}/admin/register'.format(
+            settings.ENTERPRISE_ADMIN_PORTAL_BASE_URL,
+            self.enterprise_customer.slug
+        )
+        return registration_url
```
bbaker6225, line 2370: Would adding a uuid to the url route help with the identification/activation in a simpler way?

enterprise/models.py
```diff
@@ -2328,3 +2331,84 @@ def __repr__(self):
...
+
+    def activate_admin_permissions(self, user, enterprise_customer):
```
bbaker6225, line 2372: Seems like this should be a class method if we're not going to reference `self` but instead look up PendingEnterpriseCustomerAdminUsers with the given user and customer

enterprise/signals.py
```diff
@@ -106,6 +107,59 @@ def default_content_filter(sender, instance, **kwargs):     # pylint: disable=un
...
+            pending_enterprise_admin_user = None
+
+        if kwargs['created'] and pending_enterprise_admin_user:
+            # EnterpriseCustomerUser record was created and a pending admin user
+            # exists, so assign the enterprise_admin role.
+            pending_enterprise_admin_user.activate_admin_permissions(
+                user=instance.user,
+                enterprise_customer=instance.enterprise_customer,
+            )
+        elif not kwargs['created'] and not instance.linked:
```
bbaker6225, line 134: Should we have an else statement with logging for if anything ends up in a weird state that doesn't fit the current two branches?

enterprise/signals.py
```diff
@@ -106,6 +107,59 @@ def default_content_filter(sender, instance, **kwargs):     # pylint: disable=un
...
+    if instance.user:
+        enterprise_admin_role, __ = SystemWideEnterpriseRole.objects.get_or_create(name=ENTERPRISE_ADMIN_ROLE)
+        try:
+            SystemWideEnterpriseUserRoleAssignment.objects.get(
+                user=instance.user,
+                role=enterprise_admin_role
+            ).delete()
+        except SystemWideEnterpriseUserRoleAssignment.DoesNotExist:
+            # Do nothing if no role assignment is present for the enterprise customer user.
+            pass
```
bbaker6225, line 160: It shouldn't happen often, but you can technically have multiple role assignments for the same role and user so I'd suggest we use `filter` instead of `get`
  
image

tests/test_enterprise/test_signals.py
bbaker6225: You should be able to use `setUpTestData` here too I think?

tests/test_enterprise/test_signals.py
```diff
@@ -246,11 +248,213 @@ def test_default_content_filter(self, default_content_filter, expected_content_f
...
+            user=self.admin_user,
+            role=self.enterprise_admin_role,
+        )
+        self.assertTrue(admin_role_assignments.exists())
+
+        # delete EnterpriseCustomerUser record and verify that admin role assignment is deleted as well.
+        enterprise_customer_user = EnterpriseCustomerUser.objects.filter(
+            user_id=self.admin_user.id
+        )
+        enterprise_customer_user.delete()
```
bbaker6225, line 445: nit: you can just add the `delete()` to the end of the filter

bbaker6225:
Looking pretty good so far. I leave it to your confidence levels to decide, but I feel like another reviewer might be helpful as all of these different model names are blending together in my head and I could definitely have missed something

tests/test_enterprise/test_signals.py
adamstankiewicz: Using `setUpTestData` causes tests to fail due to `self.admin_user` not existing in the test case, so I'm opting to keep it as `setUp` for now.

bbaker6225: 👍

iloveagent57 (approved):
Looks good!

chavesj (approved):
lgtm

PR #945, opened 2020-09-03
ENT-3411: Canvas discovery improvements (final touches before MVP)

binodpant:
Ok, this PR will address the following changes:

- [x]  Visibility: public (canvas default is course)
- [x] ”Let students self-enroll by sharing with them a secret URL”: true (false by default)
- [x] Course start and end dates should be displayed in Canvas syllabus page (there is a section but it’s empty right now)
    - This one is interesting because course content_type does not have start field so we don't post it. But even if I send a PUT to update the course start date, it only shows on the settings page but not on the syllabus page. So this is a different issue going on. We do send start/end dates when the field is available (such as for course_runs content type)
- [x] Course code should be set to a reasonably unique default value (right now it seems canvas auto sets it to some value that usually does not make sense)
- [x] Change link text to ‘Go to edX Course’

It will not address the following changes since I cannot find a way to do that (will check with instructure tomorrow as well)

* “Include this course in the public course index”: true (is false by default)
* ”Add a "Join this Course" link to the course home page”: true (in the UI this only shows up once the prior setting is turned on)
* Publish course: do no think this can be done via our system (or should be done). It seems this would be an institution action. Publish is needed for course to show up in public index but is not needed if instructor wants to send enrollment url

Decided not to do this one since there is no compelling rasosn right now and can cause issues with update/delete

- [] Course integration_id should be {account_id}_{key} instead of just {key}. This is not part of visibility but helps with multi account setups

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/canvas/client.py
+++ b/integrated_channels/canvas/client.py
@@ -186,6 +186,10 @@ class CanvasAPIClient(IntegratedChannelApiClient):
         self.session = session
 
     def _update_course_details(self, course_id, serialized_data):
+        """
+        Update a course for image_url (and possibly other settings in future)
+        This is used only for settings that are not settable in the initial course creation
+        """
         try:
             # there is no way to do this in a single request during create
             # https://canvas.instructure.com/doc/api/all_resources.html#method.courses.update
--- a/integrated_channels/canvas/client.py
+++ b/integrated_channels/canvas/client.py
@@ -49,6 +49,10 @@ class CanvasAPIClient(IntegratedChannelApiClient):
         self.config = apps.get_app_config('canvas')
         self.session = None
         self.expires_at = None
+        self.course_create_url = CanvasAPIClient.course_create_endpoint(
+            self.enterprise_configuration.canvas_base_url,
+            self.enterprise_configuration.canvas_account_id
+        )
 
     @staticmethod
     def course_create_endpoint(canvas_base_url, canvas_account_id):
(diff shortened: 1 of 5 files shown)


integrated_channels/canvas/client.py
```diff
@@ -49,6 +49,10 @@ def __init__(self, enterprise_configuration):
         self.config = apps.get_app_config('canvas')
         self.session = None
         self.expires_at = None
+        self.course_create_url = CanvasAPIClient.course_create_endpoint(
```
alex-sheehan-edx, line 52: 👍  I was thinking about adding basically this, thanks for including it

integrated_channels/canvas/client.py
```diff
@@ -74,29 +78,11 @@ def create_content_metadata(self, serialized_data):
         self._create_session()
 
         # step 1: create the course
-        url = CanvasAPIClient.course_create_endpoint(
-            self.enterprise_configuration.canvas_base_url,
-            self.enterprise_configuration.canvas_account_id
-        )
-        status_code, response_text = self._post(url, serialized_data)
+        status_code, response_text = self._post(self.course_create_url, serialized_data)
+        created_course_id = json.loads(response_text)['id']
```
alex-sheehan-edx, line 82: What's the fail safe here? What happens if Canvas returns an error?

integrated_channels/canvas/client.py
```diff
@@ -74,29 +78,11 @@ def create_content_metadata(self, serialized_data):
...
-                    created_course_id,
-                )
-                self._put(url, json.dumps({
-                    'course': {'image_url': content_metadata['image_url']}
-                }).encode('utf-8'))
-        except Exception:  # pylint: disable=broad-except
-            # we do not want course image update to cause failures
-            pass
+        # step 2: upload image_url and any other details
+        self._update_course_details(created_course_id, serialized_data)
```
alex-sheehan-edx, line 85: 👍  liking the organization a lot. This file was getting hard to read.

integrated_channels/canvas/client.py
```diff
@@ -126,6 +112,62 @@ def delete_content_metadata(self, serialized_data):
...
+        }
+        update_grade_response = self.session.put(submission_url, json=submission_data)
+        update_grade_response.raise_for_status()
+        return update_grade_response.status_code, update_grade_response.text
+
+    def delete_course_completion(self, user_id, payload):  # pylint: disable=unused-argument
+        # Todo: There isn't a great way for users to delete course completion data
+        pass
+
+    # Private Methods
```
alex-sheehan-edx, line 169: 👍

alex-sheehan-edx (approved):
Took a first pass. Had one small question, but the code looks pretty good. Other than that I'm slightly confused- you mentioned that this involved a change to how we handle integration id, but it looks like it's still being parsed and used for filtering in the same way?

integrated_channels/canvas/client.py
binodpant: ya I figured there were a lot of methods, private vs public so good to separate them out

integrated_channels/canvas/client.py
binodpant: two cases

[1] if exception thrown here (due to raise_for_status or due to response_text not being valid, for example):
because the base transmitter class is calling this in a try.. except, I believe it will be handled in the 'normal' error handling way (so no transmission records will be created), will that be okay for this round of work? IO was not planning on changing anything there https://github.com/edx/edx-enterprise/blob/f40fe6749e57ef91e6b0f9904700b40b9bb6c651/integrated_channels/integrated_channel/transmitters/content_metadata.py#L127

[2] if no exception here then we just return status_code, response_text

binodpant: discussed offline with Alex, this is all set

PR #946, opened 2020-09-04
ENT-3397, ENT-3398. Add default contract discount for API enrollments

georgebabey:
Adds support to capture contract discounts from the Enrollment API by
- Creating a new field `default_contract_discount` on the Enterprise Customer model
- Sending the `default_contract_discount` if set to Ecommerce when an order is created for enrollments from the API
- Capturing the `default_contract_discount` for pending enrollments from the enrollment API


irfanuddinahmad (approved):
LGTM!

PR #947, opened 2020-09-04
Improve seed data command

raq929:
Allows creation of a multiple enterprises by allowing users to specify an enterprise name
Turns on the learner portal access by default
Prints out the usernames of users created
Works in the same way as before if an enterprise name is not specified

**Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,14 @@ Change Log
 
 Unreleased
 --------------------
+
+[3.7.7] 2020-09-04
+------------------
+
+* The ``seed_enterprise_devstack_data`` management command now accepts an enterprise name when creating an enterprise,
+  and the learner portal is activated by default.
+
+
 [3.7.6] 2020-09-09
 -------------------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.7.6"
+__version__ = "3.7.7"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/enterprise/management/commands/seed_enterprise_devstack_data.py
+++ b/enterprise/management/commands/seed_enterprise_devstack_data.py
@@ -57,20 +57,33 @@ class Command(BaseCommand):
     data in devstack for related Enterprise models.
     '''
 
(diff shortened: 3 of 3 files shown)


CHANGELOG.rst
```diff
@@ -14,6 +14,12 @@ Change Log
 Unreleased
 --------------------
 
+[3.7.3] 2020-09-04
```
iloveagent57, line 17: You should rebase from master, I think 3.7.3 was released several days ago.

enterprise/__init__.py
```diff
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.7.2"
+__version__ = "3.7.3"
```
iloveagent57, line 5: 3.7.4

enterprise/management/commands/seed_enterprise_devstack_data.py
```diff
@@ -57,20 +57,33 @@ class Command(BaseCommand):
     data in devstack for related Enterprise models.
     '''
 
+    def add_arguments(self, parser):
+        """
+        Entry point for subclassed commands to add custom arguments.
+        """
+        parser.add_argument(
+            '--name',
```
iloveagent57, line 65: Nit: `customer-name` to distinguish from usernames, etc.

enterprise/management/commands/seed_enterprise_devstack_data.py
```diff
@@ -57,20 +57,33 @@ class Command(BaseCommand):
...
         enterprise_customer, __ = EnterpriseCustomer.objects.get_or_create(  # pylint: disable=no-member
-            name=customer_name,
+            name=name,
             site_id=site.id,
-            slug=slugify(customer_name),
+            slug=slugify(name),
+            country='US',
             defaults={
                 'country': 'US',
+                'enable_learner_portal': True,
```
iloveagent57, line 86: Nice.

iloveagent57 (commented):
Awesome, great idea.  Couple minor tweaks suggested below.

enterprise/management/commands/seed_enterprise_devstack_data.py
raq929: Good call. I went with "enterprise-name" because we use that elsewhere in the code.

iloveagent57 (approved):

PR #952, opened 2020-09-09
ENT-3363: enrollment_url should point to Learner Portal course page

chavesj:
[ENT-3363](https://openedx.atlassian.net/browse/ENT-3363)

**Associated PRs**
https://github.com/edx/enterprise-catalog/pull/164
https://github.com/edx/frontend-app-learner-portal-enterprise/pull/91

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -415,14 +426,27 @@ def get_course_run_enrollment_url(self, course_run_key):
...
-        return utils.update_query_parameters(url, utils.get_enterprise_utm_context(self))
+        else:
+            url = urljoin(
+                get_configuration_value('LMS_ROOT_URL', settings.LMS_ROOT_URL),
+                reverse(
+                    'enterprise_course_run_enrollment_page',
+                    kwargs={'enterprise_uuid': self.uuid, 'course_id': course_run_key}
+                )
+            )
+        return utils.update_query_parameters(url, params)
```
iloveagent57, line 449: Nit: this chunk of code is almost identical to the above (save for `course_key` vs. `course_run_key`).  Maybe you want a couple of helper functions like `_learner_portal_enrollment_url(content_key)` and `_enterprise_enrollment_url(content_key)`.

enterprise/models.py
```diff
@@ -397,13 +398,23 @@ def get_course_enrollment_url(self, course_key):
...
-                kwargs={'enterprise_uuid': self.uuid, 'course_key': course_key}
+        if self.enable_learner_portal:
+            url = '{}/{}/course/{}'.format(
+                get_configuration_value(
+                    'ENTERPRISE_LEARNER_PORTAL_BASE_URL',
+                    settings.ENTERPRISE_LEARNER_PORTAL_BASE_URL
+                ),
+                self.slug,
+                course_key,
+            )
```
adamstankiewicz, line 409: I wonder if it would make sense to move these lines into a util function or otherwise since it's used a few times, both in the tests and the actual code? That way, `'{}/{}/course/{}` is only defined once.

enterprise/utils.py
```diff
@@ -1091,3 +1091,15 @@ def delete_data_sharing_consent(course_id, customer_uuid, user_email):
...
+def get_course_key_from_course_run_key(course_run_key):
+    """
+    Args:
+        course_run_key (str): The course run id for which we would like to know the parent course id.
+
+    Returns:
+        parent_course_key (str): The parent course id for the given course run.
+    """
+    split_key = course_run_key.split(':')[1].split('+')
+    return split_key[0] + '+' + split_key[1]
```
iloveagent57, line 1105: FYI, I'm trying to find out if there's a more canonical way to do this.

iloveagent57 (commented):
This looks good, just the one nit about code reusability.  I'll try to find out if there's a "better" way of extracting course identifiers from course run identifiers.

enterprise/utils.py
iloveagent57: OK, course-discovery does something similar: https://github.com/edx/course-discovery/blob/3d928790bb9b71f12d0ed1e06af57970aee7c954/course_discovery/apps/course_metadata/utils.py#L348-L359 We should probably include the logic around `if '/' in course_run_key else ...` to handle the odd "old-style" course run keys.

enterprise/constants.py
```diff
@@ -50,6 +50,17 @@ class CourseModes:
...
+    """
+    Class to distinguish between Course, Course Run, and Program content types.
+    """
+
+    COURSE = 'course'
+    COURSE_RUN = 'courserun'
+    METADATA_KEY = 'content_type'
+    PROGRAM = 'program'
+
+
```
iloveagent57, line 63: Good idea!

iloveagent57 (approved):
Thanks for going through and changing all those strings to constants.

PR #953, opened 2020-09-10
allowing for learner data transmission to throw client errors

alex-sheehan-edx:
This PR: 
- changes a model field w/ a migration (small mistake I found)
- removes canvas client errors and replaces with client error
- adds client errors to the learner data exception catch
**Merge checklist:**

**Post merge:**


integrated_channels/integrated_channel/transmitters/learner_data.py
```diff
@@ -84,7 +85,7 @@ def transmit(self, payload, **kwargs):
                         enterprise_enrollment_id,
                     )
                 )
-            except RequestException as request_exception:
+            except (RequestException, ClientError) as request_exception:
                 code = 500
```
taliaedX, line 89: Do we really want to handle every ClientError from learner data as a 500?

taliaedX: There's also a little bit of a counterintuitive design in the base class around the error handling here.
It's handling errors in 2 ways:
1. See the try block above this, we should be receiving a status code and a couple lines below this it's setting an error_message based on codes. Thus it seems to expect certain failures to NOT raise and instead just pass a status code (like 404, 401, etc.)
2. For anything that does catastrophically fail it just uses a 500.

Thus, I think the intention here is that for any errors we know about we should *not* raise them and instead set a status code and let the exception handling occur for catastrophic failures (ie ones we wouldn't have caught which likely would be at the Requests level and thus be a RequestException)

taliaedX: So ultimately my recommendation here is to not add ClientError here and to also remove any raises in the `create_course_completion` including the last `update_grade_response.raise_for_status()` since raising in any method here will actually prevent us passing in http codes properly. If any error is raised in this method because we aren't catching it it's likely going to be caught in the transmitter anyways, especially Response errors.

taliaedX (commented):
This diff view helps a lot with narrowing down the uses/issues here.
If we need the migration part out the door right away, I'd probably split that off that way you can take your time on this bit/think through error handling a bit without rushing, since we'll have the integration available and working either way.

alex-sheehan-edx:
PR fell out of date, so updated with no approach: https://github.com/edx/edx-enterprise/pull/973

PR #958, opened 2020-09-15
adding more informative logs to the integrated channels

alex-sheehan-edx:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/exporters/learner_data.py
+++ b/integrated_channels/integrated_channel/exporters/learner_data.py
@@ -123,6 +123,11 @@ class LearnerExporter(Exporter):
 
         # Fetch course details from the Course API, and cache between calls.
         course_details = None
+
+        LOGGER.info(
+            '[Integrated Channel] Beginning export of enrollments: {}'.format(list(enrollment_queryset.values()))
+        )
+
         for enterprise_enrollment in enrollment_queryset:
 
             if TransmissionAudit and skip_transmitted and \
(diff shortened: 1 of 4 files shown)


integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -141,6 +146,9 @@ def export(self, **kwargs):
                 if self.course_api is None:
                     self.course_api = CourseApiClient()
                 course_details = self.course_api.get_course_details(course_id)
+                LOGGER.info('[Integrated Channel] Successfully retrieved course details for course: {}'.format(
```
binodpant, line 149: maybe even log course_details if not None? We have been having trouble with course_details down the line, we can remove it in future if too much text

binodpant: we are in a loop here so prior value of course_details is also used if available, so we should probably log those

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -137,10 +142,21 @@ def export(self, **kwargs):
 
             # Fetch course details from Courses API
             # pylint: disable=unsubscriptable-object
+            if course_details:
+                LOGGER.info(
+                    '[Integrated Channels] Currently exporting for course: {curr_course}. '
+                    'Course details already found for course: {found_course}'.format(
+                        curr_course=course_id,
+                        found_course=course_details['course_id'])
```
binodpant, line 150: I have a feeling this line will fail the same way as what we saw (perhaps there is a way to log this without extracting the field course_id ?

binodpant (approved):
let's try this! hopefully it will reveal the issue with the course_details

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -137,10 +142,24 @@ def export(self, **kwargs):
 
             # Fetch course details from Courses API
             # pylint: disable=unsubscriptable-object
+            if course_details:
+                LOGGER.info(
+                    '[Integrated Channels] Currently exporting for course: {curr_course}. '
+                    'Course details already found for course: {found_course}'.format(
+                        curr_course=course_id,
+                        found_course=course_details.get(
```
binodpant, line 150: don't know what this will do for a byte case, but let's see

tests/test_management.py
```diff
@@ -687,6 +693,8 @@ def get_expected_output(cmd_kwargs, certificate, self_paced, passed, **expected_
             " <SAPSuccessFactorsEnterpriseCustomerConfiguration for Enterprise Spaghetti Enterprise>",
             "[Integrated Channel] Starting Export. CompletedDate: None, Course: None, Grade: None,"
             " IsPassing: False, User: None",
+            "[Integrated Channel] Beginning export of enrollemnts:",
```
binodpant, line 696: typo

alex-sheehan-edx: good catch

alex-sheehan-edx:
jenkins run python

PR #960, opened 2020-09-16
including enterprise customer within logs

alex-sheehan-edx:
**Merge checklist:**

**Post merge:**

Any other logs you'd like to see here?

### DIFF
--- a/integrated_channels/integrated_channel/exporters/learner_data.py
+++ b/integrated_channels/integrated_channel/exporters/learner_data.py
@@ -19,7 +19,7 @@ from consent.models import DataSharingConsent
 from enterprise.api_client.lms import CertificatesApiClient, CourseApiClient, GradesApiClient
 from enterprise.models import EnterpriseCourseEnrollment
 from integrated_channels.integrated_channel.exporters import Exporter
-from integrated_channels.utils import is_already_transmitted, parse_datetime_to_epoch_millis
+from integrated_channels.utils import generate_formatted_log, is_already_transmitted, parse_datetime_to_epoch_millis
 
 LOGGER = getLogger(__name__)
 
@@ -72,7 +72,7 @@ class LearnerExporter(Exporter):
         """
         return self.GRADE_INCOMPLETE
 
-    def export(self, **kwargs):
+    def export(self, **kwargs):  # pylint: disable=R0915
         """
         Collect learner data for the ``EnterpriseCustomer`` where data sharing consent is granted.
 
@@ -84,6 +84,7 @@ class LearnerExporter(Exporter):
           for self-paced courses, when the course end date is passed, or when the learner achieves a passing grade.
         * ``grade``: string grade recorded for the learner in the course.
         """
+        channel_name = kwargs.get('app_label')
         exporting_single_learner = False
         learner_to_transmit = kwargs.get('learner_to_transmit', None)
         course_run_id = kwargs.get('course_run_id', None)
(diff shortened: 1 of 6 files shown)


tests/test_management.py
```diff
@@ -1026,6 +1027,7 @@ def setUp(self):
          dict(completed='true', timestamp=PAST_TIMESTAMP, grade='Pass', total_hours=0.0)),
     )
     @ddt.unpack
+    @skip(("This test is hard coding log order and OC team needs more comprehensive logs for staging"))
```
binodpant, line 1030: shall we mention we plan to restore this test once the issues have been resolved?

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -125,16 +126,20 @@ def export(self, **kwargs):
         course_details = None
 
         LOGGER.info(
-            '[Integrated Channel] Beginning export of enrollments: {}'.format(list(enrollment_queryset.values()))
+            '[Integrated Channel] Beginning export of enrollments: {enrollments} for enterprise user: {user}'.format(
```
binodpant, line 129: being nitpicky but I was thinking we could standardize the way our logging looks a bit by having sth consistent looking like 

```
[Integrated Channel] [channel_name] [enterprise_customer_name_or_uuid]: <message>
```

of course it may depend on what you are logging. having a util function to log with such consistency may make it easier for us to search through and correlate log entries

if a piece of info is not available, you would simply drop it from the logging (to avoid failures)

feel free to only do enough as needed for this pr, just some thoughts on logging overall

binodpant (approved):
i am good, just made some overall comments and thoughts on logging overall

integrated_channels/integrated_channel/exporters/learner_data.py
taliaedX: Agreed on this front. I think a logging util is definitely a good way to go to enforce consistency (though it doesn't work for methods we override, but we're manually adding the channel name to those anyways and we can refactor that later)

binodpant: yup esp the channel, since now we have multiple integrated channels, and sometimes it take a while to even ralize it's not even canvas!

integrated_channels/utils.py
```diff
@@ -221,3 +224,17 @@ def get_subjects_from_content_metadata(content_metadata_item):
...
+    """
+    Formats and logs a message for the integrated channels.
+    """
+    if enterprise_customer_identifier:
+        enterprise_customer_identifier = '[ENT CUSTOMER: ' + enterprise_customer_identifier + '] '
+
+    if channel_name:
+        channel_name = '[' + channel_name + '] '
+
+    log_message = '[Integrated Channel] ' + enterprise_customer_identifier + channel_name + message
```
binodpant, line 239: need some delimiters here? esp before mesage

integrated_channels/utils.py
```diff
@@ -221,3 +224,17 @@ def get_subjects_from_content_metadata(content_metadata_item):
...
+    LOGGER.error(log_message) if is_error else LOGGER.info(log_message)  # pylint: disable=expression-not-assigned
```
binodpant, line 240: I think this is a great improvement indeed! have some other thoughts for later logging work but this should get us a great set of logs for now!

integrated_channels/utils.py
binodpant, line 239: better readability to use `format()` here or the dict style input to logger, but this is fine too!

alex-sheehan-edx: Good point, will definitely switch.

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -94,14 +95,18 @@ def export(self, **kwargs):
...
+        generate_formatted_log(
+            'Starting Export. CompletedDate: {completed_date}, Course: {course_run}, '
+            'Grade: {grade}, IsPassing: {is_passing}, User: {user_id}'.format(
+                completed_date=completed_date,
+                course_run=course_run_id,
+                grade=grade,
+                is_passing=is_passing,
+                user_id=learner_to_transmit.id if learner_to_transmit else None
+            ),
+            channel_name=channel_name,
```
binodpant, line 107: +1 for this!

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -114,57 +119,81 @@ def export(self, **kwargs):
...
-            '[Integrated Channel] Beginning export of enrollments: {}'.format(list(enrollment_queryset.values()))
+        generate_formatted_log(
+            'Beginning export of enrollments: {enrollments}.'.format(
+                enrollments=list(enrollment_queryset.values()),
+            ),
+            channel_name=channel_name,
+            enterprise_customer_identifier=self.enterprise_customer.name
         )
 
         for enterprise_enrollment in enrollment_queryset:
```
taliaedX, line 144: have you considered sticking a debug log here to basically say "are we even in this for loop lol"

taliaedX: Though I guess the "Beginning..." log should suffice.

alex-sheehan-edx: Not a bad idea, looking for places where I can add more clarity.

alex-sheehan-edx: The `Beginning` log was supposed to accomplish that, yes.

integrated_channels/utils.py
```diff
@@ -221,3 +224,21 @@ def get_subjects_from_content_metadata(content_metadata_item):
...
+    """
+    Formats and logs a message for the integrated channels.
+    """
+    if enterprise_customer_identifier:
+        enterprise_customer_identifier = ' [ENT CUSTOMER: ' + enterprise_customer_identifier + ']'
+
+    if channel_name:
+        channel_name = ' [' + channel_name + ']'
+
+    log_message = '[Integrated Channel]{enterprise_customer_identifier}{channel_name}: {message}'.format(
```
taliaedX, line 239: Chatted via DM, but just noting it here for posterity.
This would be better as `log_message = '[Integrated Channel][{enterprise_customer_identifier}][{channel_name}]: {message}' and removing the if statements above this as having `[None]` is better (IMO) than just `None`

PR #966, opened 2020-09-18
search course by integration_id urlencoded

binodpant:
Locally, was able to get name updated, via update method, by manually changing name in the content metadata and running task again. Verified in canvas the name was changed

Now this code will find the course by integration_id (it must be url encoded for it to work)

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/canvas/client.py
+++ b/integrated_channels/canvas/client.py
@@ -6,6 +6,7 @@ import json
 
 import requests
 from six.moves.urllib.parse import urljoin  # pylint: disable=import-error
+from requests.utils import quote
 
 from django.apps import apps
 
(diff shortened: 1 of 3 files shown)


binodpant:
Screen Shot 2020-09-18 at 4 03 47 PM

integrated_channels/canvas/client.py
```diff
@@ -272,9 +273,10 @@ def _get_course_id_from_integration_id(self, integration_id):
         Args:
             integration_id (string): The ID retrieved from the transmission payload.
         """
-        url = "{}/api/v1/accounts/{}/courses/".format(
+        url = "{}/api/v1/accounts/{}/courses/?search_term={}".format(
             self.enterprise_configuration.canvas_base_url,
-            self.enterprise_configuration.canvas_account_id
+            self.enterprise_configuration.canvas_account_id,
+            quote(integration_id),
```
taliaedX, line 279: I think through some Moodle work I've found that requests may already escape characters in query params, but I don't think it harms anything to use it here so no biggie.

Also it's part of urllib so you could use that since you're already importing it instead of requests. 🤷 
ie `from six.moves.urllib.parse import urljoin, quote`

taliaedX (approved):
LGTM.

alex-sheehan-edx (approved):
Nice addition! And relatively simple too! 🙂  lgtm (with Talia's comment)

integrated_channels/canvas/client.py
binodpant: ok thanks, will merge this but yeah will checkout the six version too, ty

PR #969, opened 2020-09-22
ENT-3381: Only create new session if token expired

binodpant:
Why?
To avoid creating a bazillion sessions and tokens
Right now every call to create/update/delete content_metadata will issue a new session and a new auth token. While it works, it's an optimization for us to keep reusing the session and the auth token while it is still valid

This PR creates a new session when _create_session() is called (and obtains a new auth token and jams it into the session), but only if one of these is true:

1. token expiry is detected based on a stored `expires_at` attribute
2. token has never been obtained (is None)
3. session has never been initialized (is None)

If 2 and 3 are false, it still handles the case of 1, token expiry

also moved around test code for better org (and one small refactor on token response method)

**Merge checklist:**

**Post merge:**


integrated_channels/canvas/client.py
```diff
@@ -424,6 +433,6 @@ def _get_oauth_access_token(self, client_id, client_secret):
         auth_response.raise_for_status()
         try:
             data = auth_response.json()
-            return data['access_token']
+            return data['access_token'], data["expires_in"]
```
alex-sheehan-edx, line 436: 👍  nice that it worked out this well

integrated_channels/canvas/client.py
```diff
@@ -173,18 +174,25 @@ def _create_session(self):
...
+                self.enterprise_configuration.client_id,
+                self.enterprise_configuration.client_secret,
+            )
+            session = requests.Session()
+            session.headers['Authorization'] = 'Bearer {}'.format(oauth_access_token)
+            session.headers['content-type'] = 'application/json'
+            self.session = session
+            # expiry expected after `expires_in` seconds
+            if expires_in is not None:
+                self.expires_at = datetime.utcnow() + timedelta(seconds=expires_in)
```
alex-sheehan-edx, line 195: I didn't look into the expiration handling from Canvas, is it always in seconds? What is the actual amount of time until we lose the token?

integrated_channels/canvas/client.py
```diff
@@ -394,6 +402,7 @@ def _get_oauth_access_token(self, client_id, client_secret):
 
         Returns:
             access_token (str): the OAuth access token to access the Canvas API as the user
+            expires_in (int): the number of seconds after which token will expire
```
alex-sheehan-edx, line 405: soooo nitpicky here, but `the token will expire`?

alex-sheehan-edx (approved):
A couple q's but otherwise this looks real good 👍

integrated_channels/canvas/client.py
binodpant: yup, always in seconds (1 hour by default)

binodpant: https://canvas.instructure.com/doc/api/file.oauth.html#using-refresh-tokens

integrated_channels/canvas/client.py
binodpant: :)

integrated_channels/canvas/client.py
binodpant: regardless, this will use the expires_in (Seconds) as a way to setup when we should expect expiry

PR #972, opened 2020-09-23
ENT-3471: Basic setup for Blackboard Integrated channel

binodpant:
Includes:
* basic wiring up of client, exporter, transmitter (For content metadata), one config model with minimal fields, unit test for client  (just sanity check level right now)

Does not yet include these (will be added in other subtasks):
* actual implementation of any of the pieces!
* admin page work
* views/endpoints if needed for oauth (not sure yet)

**Merge checklist:**

**Post merge:**


integrated_channels/blackboard/__init__.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+"""
+The Blackboard Integrated Channel package.
+"""
+
+__version__ = "0.0.1"
+
+default_app_config = ("integrated_channels.blackboard.apps."  # pylint: disable=invalid-name
```
raq929, line 8: So, tell me about why this name ends in a dot and why we're disabling invalid-name linting?

integrated_channels/blackboard/models.py
```diff
@@ -0,0 +1,89 @@
...
+from .transmitters.content_metadata import BlackboardContentMetadataTransmitter
+from .exporters.content_metadata import BlackboardContentMetadataExporter
+
+
+@python_2_unicode_compatible
+class BlackboardEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguration):
+    """
+    The Enterprise-specific configuration we need for integrating with Blackboard.
+
+    .. no_pii:
```
raq929, line 21: What is this line doing?

integrated_channels/blackboard/models.py
```diff
@@ -0,0 +1,89 @@
...
+        Return uniquely identifying string representation.
+        """
+        return self.__str__()
+
+    @staticmethod
+    def channel_code():
+        """
+        Returns an capitalized identifier for this channel class, unique among subclasses.
+        """
+        return 'BLACKBOARD'
```
raq929, line 83: This might be the kind of thing that I store in an enum somewhere along with the other channel codes, but, not super important atm.

tests/test_apps.py
```diff
@@ -135,6 +135,23 @@ def setUp(self):
...
+        """
+        Set up test environment
+        """
+        super(TestBlackboardConfig, self).setUp()
+        self.app_config = integrated_channels.blackboard.apps.BlackboardConfig(
+            'blackboard', integrated_channels.blackboard
+        )
+
+    def test_name(self):
+        assert self.app_config.name == 'blackboard'
```
raq929, line 154: Again, I'm tempted to avoid "magic strings" and have "blackboard" stored somewhere, but again, not critical.

tests/test_integrated_channels/test_blackboard/test_client.py
```diff
@@ -0,0 +1,30 @@
...
+        self.enterprise_config = BlackboardEnterpriseCustomerConfigurationFactory(
+            client_id="id",
+            client_secret="secret",
+            blackboard_base_url="https://base.url",
+            refresh_token="token",
+        )
+
+    def test_client_has_valid_app_config(self):
+        api_client = BlackboardAPIClient(self.enterprise_config)
+        assert api_client.config is not None
```
raq929, line 30: Should we test that it has some of the config you passed to it?

integrated_channels/blackboard/models.py
```diff
@@ -0,0 +1,89 @@
...
+        max_length=255,
+        null=True,
+        verbose_name="API Client ID",
+        help_text=(
+            "The API Client ID provided to edX by the enterprise customer to be used to make API "
+            "calls on behalf of the customer."
+        )
+    )
+
+    client_secret = models.CharField(
```
raq929, line 34: Something I always like to think about when storing secrets is how they're being stored, and how they're being displayed. I don't know what the considerations are here, but, food for thought.

raq929 (approved):
Some questions/comments, but nothing pressing!

raq929:
Oh, it looks like you didn't bump the edx-enterprise version or changelog. Is that intentional?

integrated_channels/blackboard/__init__.py
binodpant: hah! it seems like we are splitting on multiple lines (why???) copy paste I think, removed it into one line good catch!

integrated_channels/blackboard/models.py
binodpant: I believe it is related to this 
https://github.com/edx/edx-enterprise/blob/4a598e8b6a908f474b533a61cbba2ea2667e6d48/.annotation_safe_list.yml#L1

that file handles autogenerated models, and this here is for manually annotating them for whatever task is checking these ... I need to learn more but pretty sure we needed to add these ...

integrated_channels/blackboard/models.py
binodpant: Good thought ! we should probably talk about this in the team channel. Generally all channels seem to be storing this as plain text probably because you need this plus a bunch of other info to get a valid token (client_id, redirect_uri all of which have to match).

tests/test_apps.py
binodpant: yup I like the idea, perhaps when I next work on this blackboard area, I will take care of that ... all channels could use that improvement

integrated_channels/blackboard/models.py
raq929: Oh, interesting. :)

tests/test_integrated_channels/test_blackboard/test_client.py
binodpant: yup, the config part is tested in another test mostly but I added extra checks to be 100% rigid in the testing thanks

integrated_channels/blackboard/apps.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+"""
+Enterprise Integrated Channel Blackboard Django application initialization.
+"""
+
+from django.apps import AppConfig
+
+CHANNEL_NAME = 'integrated_channels.blackboard'
+VEROBSE_NAME = 'Enterprise Blackboard Integration'
```
taliaedX, line 9: nit: Spelling `verbose`

taliaedX (approved):
1 nitpick but otherwise looks fine to me.

PR #975, opened 2020-09-23
pull platform sync migration test to enterprise.

taliaedX:
Trying to prevent the dreaded successful merge to master > fail in edx-platform pipeline.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/tests/test_migrations.py
+++ b/tests/test_migrations.py
@@ -3,68 +3,19 @@
 Tests for migrations, especially potentially risky data migrations.
 """
 
-from django.db import connection
-from django.db.migrations.executor import MigrationExecutor
-from django.test.testcases import TransactionTestCase
+from django.core.management import call_command
+from django.test.testcases import TestCase
+from django.test.utils import override_settings
+from django.utils.six import StringIO
 
 
-class MigrationTestCase(TransactionTestCase):
+class MigrationTests(TestCase):
     """
-    A base test case class for migration test cases.
+    Runs migration tests using Django Command interface.
     """
-
-    migrate_origin = None
-    migrate_dest = None
-    model = None
-
-    def setUp(self):
-        super(MigrationTestCase, self).setUp()
-        self.executor = MigrationExecutor(connection)
-        self.executor.migrate(self.migrate_origin)
-
-    def migrate_to_origin(self):
-        """
-        Performs the migration to the designated origin.
-
-        This only really does anything if you have migrated forward
-        in some way or no migrations were performed at all.
-        """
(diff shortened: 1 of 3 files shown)


tests/test_migrations.py
```diff
@@ -3,9 +3,12 @@
...
+from django.core.management import call_command
 from django.db import connection
 from django.db.migrations.executor import MigrationExecutor
-from django.test.testcases import TransactionTestCase
+from django.test.testcases import TestCase, TransactionTestCase
+from django.test.utils import override_settings
+from django.utils.six import StringIO
 
 
 class MigrationTestCase(TransactionTestCase):
```
taliaedX, line 14: for real what is this even doing?

tests/test_migrations.py
```diff
@@ -68,3 +71,15 @@ def model_label(self):
...
+
+
+class MigrationTests(TestCase):
+    """
+    Runs migration tests using Django Command interface.
+    """
+    @override_settings(MIGRATION_MODULES={})
+    def test_migrations_are_in_sync(self):
+        out = StringIO()
+        call_command("makemigrations", dry_run=True, verbosity=3, stdout=out)
```
taliaedX, line 83: I noticed there is also a `test_management.py` file as well, so I could potentially just put this in that file instead since I am in fact using the management command version. However, that file seems to be dedicated to _custom_ management commands, not the base Django commands.

tests/test_migrations.py
binodpant: hmm is it jist a base testcase? is it even subclassed by anything?

taliaedX: Nope. Couldn't find this class referenced anywhere. 🤷

tests/test_migrations.py
alex-sheehan-edx: The management file is also where that test that asserts log order is... I'm ok with it living here!

alex-sheehan-edx (approved):
looks pretty good to me! Will this get automatically run before we merge?

tests/test_migrations.py
taliaedX: hmm yeah I could test that out. It looks like tests are already failing so I'll figure out what that issue is and then give removal here a shot.

taliaedX:
>Will this get automatically run before we merge?

The test? Yep!

tests/test_migrations.py
taliaedX: seems like the tests don't fail with it removed so I'll probably just leave it removed in this commit.

PR #977, opened 2020-09-25
ENT-3482:Add functionality to save logo in EnterpriseCustomerBrandingConfiguration at only one location

sameenfatima78:
JIRA: https://openedx.atlassian.net/browse/ENT-3482

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/signals.py
+++ b/enterprise/signals.py
@@ -3,12 +3,11 @@
 Django signal handlers.
 """
 
-import os
 from logging import getLogger
 
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import transaction
-from django.db.models.signals import post_delete, post_save
+from django.db.models.signals import post_delete, post_save, pre_save
 from django.dispatch import receiver
 
 from enterprise.api_client.enterprise_catalog import EnterpriseCatalogApiClient
@@ -33,6 +32,7 @@ except ImportError:
     CourseEnrollment = None
 
 logger = getLogger(__name__)  # pylint: disable=invalid-name
+_UNSAVED_FILEFIELD = 'unsaved_filefield'
 
 
 @disable_for_loaddata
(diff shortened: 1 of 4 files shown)


HammadAhmadWaqas:
@sameenfatima78 could we add a test to support our changes?
ideally, that test should fail without your changes and pass after implementing your changes.

HammadAhmadWaqas (commented):
@sameenfatima78 could we add a test to support our changes?
ideally, that test should fail without your changes and pass after implementing your changes.

saleem-latif (approved):
Just one question otherwise looks good 👍

enterprise/signals.py
```diff
@@ -99,17 +99,24 @@ def _complete_user_enrollment():
...
+def save_logo_file(sender, instance, **kwargs):            # pylint: disable=unused-argument
     """
-    Update logo name and path to replace None with instance id.
+    Now that the object is instantiated and instance.id exists, save the image at correct path and re-save the model.
     """
-    filename = instance.logo.name
-    if kwargs['created'] and 'None' in filename:
-        extension = os.path.splitext(filename)[1].lower()
-        instance_file = instance.logo.file
-        new_name = 'enterprise/branding/' + str(instance.id) + '/' + str(instance.id) + '_logo' + extension
```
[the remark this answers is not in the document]
sameenfatima78: @saleem-latif No, this code was actually added by me to fix ENT-3406. It was added to re-save the instance at the correct path once the model object was created. But now we are only saving the image once after post_save and the path to upload the image is already given [here](https://github.com/edx/edx-enterprise/blob/master/enterprise/models.py#L1129) already.

enterprise/signals.py
```diff
@@ -99,17 +99,24 @@ def _complete_user_enrollment():
...
-        instance.logo.save(new_name, instance_file)
+    if kwargs['created'] and hasattr(instance, _UNSAVED_FILEFIELD):
+        instance.logo = getattr(instance, _UNSAVED_FILEFIELD)
+        instance.__dict__.pop(_UNSAVED_FILEFIELD)
```
HammadAhmadWaqas, line 119: can we use `delattr` builtin function here?

sameenfatima78: Good suggestion. I have incorporated that.

tests/test_models.py
```diff
@@ -782,6 +782,23 @@ def test_logo_path(self, file_exists, delete_called):
...
+        branding_config = EnterpriseCustomerBrandingConfiguration(
+            enterprise_customer=factories.EnterpriseCustomerFactory(),
+            logo="logo.png"
+        )
+        storage_mock = mock.MagicMock(spec=Storage, name="StorageMock")
+        with mock.patch("django.core.files.storage.default_storage._wrapped", storage_mock):
+            branding_config.save()
+            config_id = branding_config.id
+            path = logo_path(branding_config, branding_config.logo.name)
+            expected_path = "enterprise/branding/"+str(config_id)+"/"+str(config_id)+"_logo.png"
```
HammadAhmadWaqas, line 799: I think we should assert the saved path of the image with the expected path from the logo_path.
```
saved_path = branding_config.logo.path
expected_path = logo_path(branding_config, branding_config.logo.name)
self.assertEqual(saved_path, expected_path)
```

HammadAhmadWaqas (commented):
minor suggestion, otherwise good.

tests/test_models.py
sameenfatima78: I can't check the exact path as I'm mocking Storage. All I need to test is that the path has the correct instance id in it instead of None.

HammadAhmadWaqas: I didn't get why we have to mock storage for this test?

sameenfatima78: Let's catch up on Slack

HammadAhmadWaqas (approved):

PR #979, opened 2020-09-28
ENT-3334 | SAP SF customer is seeing courses as not open for enrollment when they should be

HammadAhmadWaqas:
comparing `status` of the course_run with the `unpublished` instead of 'published' to handle cases where unpublished key is missing.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -769,7 +769,7 @@ def is_course_run_enrollable(course_run):
     3. enrollment_end is greater than now, or undefined
     """
     # Check if the course run is unpublished (sometimes these sneak through)
-    if not is_course_run_published(course_run):
+    if is_course_run_unpublished(course_run):
         return False
 
     now = datetime.datetime.now(pytz.UTC)
@@ -822,12 +822,12 @@ def is_course_run_upgradeable(course_run):
     return False
 
 
-def is_course_run_published(course_run):
+def is_course_run_unpublished(course_run):
     """
-    Return True if the course run's status value is "published".
+    Return True if the course run's status value is "unpublished".
     """
     if course_run:
-        if course_run.get('status') == 'published':
+        if course_run.get('status') == 'unpublished':
             return True
     return False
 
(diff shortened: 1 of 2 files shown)


enterprise/utils.py
```diff
@@ -822,12 +822,12 @@ def is_course_run_upgradeable(course_run):
...
 
-def is_course_run_published(course_run):
+def is_course_run_unpublished(course_run):
     """
-    Return True if the course run's status value is "published".
+    Return True if the course run's status value is "unpublished".
     """
     if course_run:
-        if course_run.get('status') == 'published':
+        if course_run.get('status') == 'unpublished':
```
hasnain-naveed, line 830: What are the other statuses?

saleem-latif, line 830: Why was this causing the issue? Do we not have the correct value of status for published courses?

HammadAhmadWaqas: @saleem-latif 
the main purpose of this method was to filter out all unpublished course_runs from the enrollable courses.
it was doing it successfully for all the course_runs where the status key was present, but the course_runs where the status key was missing(as it's not mandatory) this check was returning incorrect value. (None == published)
So just modified this method in a way to handle scenarios where the `status` key is not present to consider that course available for enrollment.
@saleem-latif is it helping to answer your question?

HammadAhmadWaqas: this was the PR when we wrote this method.
https://github.com/edx/edx-enterprise/pull/795

saleem-latif: So for missing status, the course will be considered published, right?

HammadAhmadWaqas:
Yes

saleem-latif (approved):

hasnain-naveed (approved):

HammadAhmadWaqas:
closing this PR, in the favor of other solution.
https://github.com/edx/course-discovery/pull/2758

mattdrayer:
Yay, I didn't break it after all :)

PR #985, opened 2020-09-30
Adding completion data reporting for moodle

alex-sheehan-edx:
**Merge checklist:**

**Post merge:**

V0 Draft:
What remains
- testing
- adding real doc strings
- decide on possible configurations?


integrated_channels/moodle/client.py
```diff
@@ -150,6 +151,100 @@ def _get_access_token(self):
...
             )
 
+    @moodle_request_wrapper
+    def _get_enrolled_users(self, course_id):
+        """
+        placeholder
+        """
+        params = {
+            'wstoken': self.token,
+            'wsfunction': 'core_enrol_get_enrolled_users',
```
taliaedX, line 161: these jokers really spell it `enrol` and `enrolment`, but then use `enrolled`. _flips table_

integrated_channels/moodle/client.py
```diff
@@ -178,9 +273,11 @@ def get_course_id(self, key):
         response = self._get_courses(key)
         parsed_response = json.loads(response.text)
         if not parsed_response.get('courses'):
-            raise ClientError('MoodleAPIClient request failed: 404 Course key '
-                              '"{}" not found in Moodle.'.format(key))
-
+            raise ClientError(
+                'MoodleAPIClient request failed: 404 Course key '
+                '"{}" not found in Moodle.'.format(key),
+                HTTPStatus.NOT_FOUND.value
```
taliaedX, line 279: what actually is the value of Not found here? I'm just not familiar with it.

integrated_channels/moodle/client.py
```diff
@@ -228,7 +325,27 @@ def delete_content_metadata(self, serialized_data):
...
+            'wsfunction': 'core_grades_update_grades',
+            'source': module_name,
+            'courseid': course_id,
+            'component': 'mod_assign',
+            'activityid': course_module_id,
+            'itemnumber': 0,
+            'grades[0][studentid]': moodle_user_id,
+            # The grade is exported as a decimal between [0-1]
+            # This could potentially be another point of configuration for the user
+            # as they could potentially supply us with how they want grades displayed (percent, point, letter)
```
taliaedX, line 344: I assume you mean the grade on edx's side is a decimal (TIL if so) and we can add logic to pass this to Moodle however the customer would like? Can you even do a letter (ie is the field on Moodle's side expecting an int? Or is it just a char field?

integrated_channels/moodle/exporters/learner_data.py
```diff
@@ -0,0 +1,73 @@
...
+            **kwargs
+    ):  # pylint: disable=arguments-differ,unused-argument
+        """
+        Return a MoodleLearnerDataTransmissionAudit with the given enrollment and course completion data.
+        If completed_date is None, then course completion has not been met.
+        If no remote ID can be found, return None.
+        """
+        completed_timestamp = None
+        if completed_date is not None:
+            completed_timestamp = parse_datetime_to_epoch_millis(completed_date)
```
taliaedX, line 36: I didn't even know about this util. Nice.

taliaedX (commented):
I have some questions; mostly nitpicks, but generally looks good.

integrated_channels/moodle/client.py
alex-sheehan-edx: 404!

taliaedX: oh it's just the status code. They don't use `.code`? Weird

integrated_channels/moodle/client.py
alex-sheehan-edx: ...As I'm sure you know already, that's not even the worst part! They have function names misspelled in their docs that don't have a list of parameters so when you grep through the source code nothing comes up 😞

integrated_channels/moodle/client.py
```diff
@@ -150,6 +151,100 @@ def _get_access_token(self):
...
+        return response
+
+    def get_course_final_grade_module(self, course_id):
+        """
+        placeholder
+        """
+        response = self._get_course_contents(course_id)
+
+        course_module_id = None
+        if isinstance(response.json()) is list:
```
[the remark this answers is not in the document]
alex-sheehan-edx: this is actually the incorrect way to use `isinstance` 🤦  will change

integrated_channels/moodle/client.py
taliaedX: oh yes like grades vs grade

integrated_channels/moodle/exporters/learner_data.py
alex-sheehan-edx: (tbh I learned about it while implementing Canvas and seeing some other channel do it)

integrated_channels/moodle/client.py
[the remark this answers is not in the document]
taliaedX: ah yes I missed that. Good follow up/catch!

integrated_channels/moodle/client.py
```diff
@@ -150,6 +151,100 @@ def _get_access_token(self):
...
+        )
+        return user_response
+
+    def get_creds_of_user_in_course(self, course_id, user_email):
+        """
+        placeholder
+        """
+        response = self._get_enrolled_users(course_id)
+        parsed_response = response.json()
+        user_id = None
```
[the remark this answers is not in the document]
alex-sheehan-edx: I don't believe that
```
if not user_id:
``` 
will appropriately handle the condition that we never find the user's email in the list of enrolled students without the assignment to None. However, please correct me if I'm wrong! Happy to remove it if it's unnecessary.

integrated_channels/moodle/client.py
```diff
@@ -150,6 +151,100 @@ def _get_access_token(self):
...
+
+        course_module_id = None
+        if isinstance(response.json()) is list:
+            for course in response.json():
+
+                # This is one point where I think we can make this configurable-
+                # A freshly made course will have "General" module that this code
+                # is assuming the `final grade` activity will live. We could pretty easily
+                # have the user give us the ID of the module where the activity lives and
+                # even default to the 'General'
```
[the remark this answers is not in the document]
alex-sheehan-edx: @georgebabey mentioned to me that we might have a contact within Moodle. I think being able to talk to someone who has some insider knowledge will be helpful in choosing the right course of action.

integrated_channels/moodle/client.py
alex-sheehan-edx: I just tried `.code` just to __make sure__ aannnnd got an attribute not found.

integrated_channels/moodle/client.py
alex-sheehan-edx: I added the percent grade to the exportation, and I believe (some investigation pending?) that things like letter grades should be accessible too. And yeah, the main point of the comment was to say that we could allow the customer to configure this and I'm wondering if we want to add the functionality to this PR

alex-sheehan-edx: To add more clarity- how I see this configuration (+ creating shell final grade assignment configurations?), we'd probably add another admin form/model for MoodleCompletionDataConfiguration or something similarly named.

integrated_channels/moodle/client.py
```diff
@@ -150,6 +151,100 @@ def _get_access_token(self):
                 response.status_code
             )
 
+    @moodle_request_wrapper
+    def _get_enrolled_users(self, course_id):
+        """
+        placeholder
```
binodpant, line 157: is this text intentional or actually means the field is a placeholder?

integrated_channels/moodle/client.py
```diff
@@ -150,6 +151,100 @@ def _get_access_token(self):
...
+                self.enterprise_configuration.moodle_base_url,
+                self.MOODLE_API_PATH,
+            ),
+            params=params
+        )
+        return user_response
+
+    def get_creds_of_user_in_course(self, course_id, user_email):
+        """
+        placeholder
```
binodpant, line 176: ??

integrated_channels/moodle/client.py
```diff
@@ -150,6 +151,100 @@ def _get_access_token(self):
...
+
+                # This is one point where I think we can make this configurable-
+                # A freshly made course will have "General" module that this code
+                # is assuming the `final grade` activity will live. We could pretty easily
+                # have the user give us the ID of the module where the activity lives and
+                # even default to the 'General'
+                if course.get('name') == 'General':
+                    modules = course.get('modules')
+                    for module in modules:
+                        if module.get('name') == '(edX integration) Final Grade':
```
binodpant, line 235: is this a constant somewhere, or should be if used outside of this code as well?

binodpant: ah yeah I would constantify it esp since it's used again later here too

integrated_channels/moodle/client.py
binodpant: from their doc:
```>>> HTTPStatus.OK.value
200
>>> HTTPStatus.OK.phrase
'OK'
```

integrated_channels/moodle/client.py
```diff
@@ -228,7 +325,27 @@ def delete_content_metadata(self, serialized_data):
...
+        moodle_user_id = self.get_creds_of_user_in_course(course_id, user_id)
+
+        params = {
+            'wsfunction': 'core_grades_update_grades',
+            'source': module_name,
+            'courseid': course_id,
+            'component': 'mod_assign',
+            'activityid': course_module_id,
+            'itemnumber': 0,
+            'grades[0][studentid]': moodle_user_id,
```
binodpant, line 341: I think the `grades[0][FIELD]` needs a brief comment here? it may not be obvious reading this code why grades[0] .. for example or maybe in function docstring is good too. 

also, will there ever be a grades[1] ?

integrated_channels/moodle/client.py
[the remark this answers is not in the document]
taliaedX: I've read that first sentence several times and I don't quite understand.
You have: 
```
if enrollment.get('email') == user_email:
    user_id = enrollment.get('id')
    break
```
so if we never get a user_id from this loop because there is never a match then the next logic:
`if not user_id` would still catch it because user_id is still None. It's never defined. This is what I mean by not needing to set it explicitly to None.

integrated_channels/moodle/exporters/learner_data.py
```diff
@@ -0,0 +1,73 @@
...
+
+        if enterprise_enrollment.enterprise_customer_user.user_email is None:
+            LOGGER.debug(
```
binodpant, line 39: small but may look cleaner? I feel like since there are so many uses of `enterprise_enrollment.enterprise_customer_user`, a cached variable would be nice ...

integrated_channels/moodle/models.py
```diff
@@ -106,8 +109,92 @@ def channel_code():
...
+    """
+    The payload we send to Moodle at a given point in time for an enterprise course enrollment.
+
+    """
+    moodle_user_email = models.CharField(max_length=255, blank=False, null=False)
+
+    enterprise_course_enrollment_id = models.PositiveIntegerField(blank=False, null=False, db_index=True)
+    course_id = models.CharField(max_length=255, blank=False, null=False)
+    course_completed = models.BooleanField(default=True)
+    grade = models.CharField(max_length=100, blank=False, null=False)
```
binodpant, line 136: hmm, grade max_length 100 chars? too long?

integrated_channels/moodle/client.py
taliaedX: Yeah but my question was does Moodle support a letter grade?
I ask because their date field (which is an int [commence groaning]) is strictly typed and I wasn't sure if this grade field is also strictly typed since you are currently passing an int. 

TL;DR if you set grade to "B" here would it fail to load into Moodle?

integrated_channels/moodle/models.py
```diff
@@ -106,8 +109,92 @@ def channel_code():
...
         return MoodleContentMetadataTransmitter(self)
+
+
+@python_2_unicode_compatible
+class MoodleLearnerDataTransmissionAudit(models.Model):
+    """
+    The payload we send to Moodle at a given point in time for an enterprise course enrollment.
+
+    """
+    moodle_user_email = models.CharField(max_length=255, blank=False, null=False)
```
binodpant, line 131: would EmailField be better here?

https://docs.djangoproject.com/en/3.1/ref/models/fields/#emailfield

integrated_channels/moodle/models.py
```diff
@@ -106,8 +109,92 @@ def channel_code():
...
+
+    enterprise_course_enrollment_id = models.PositiveIntegerField(blank=False, null=False, db_index=True)
+    course_id = models.CharField(max_length=255, blank=False, null=False)
+    course_completed = models.BooleanField(default=True)
+    grade = models.CharField(max_length=100, blank=False, null=False)
+    total_hours = models.FloatField(null=True, blank=True)
+    course_completed = models.BooleanField(
+        default=True,
```
binodpant, line 139: minor question I guess, is default True sensible because we are already transmitting grades? is default False any better/worse?

integrated_channels/moodle/transmitters/learner_data.py
```diff
@@ -0,0 +1,35 @@
...
+        By default, use the ``MoodleAPIClient`` for learner data transmission to Moodle.
+        """
+        super(MoodleLearnerTransmitter, self).__init__(
+            enterprise_configuration=enterprise_configuration,
+            client=client
+        )
+
+    def transmit(self, payload, **kwargs):
+        """
+        Send a completion status call to Canvas using the client.
```
binodpant, line 27: Canvas? :)

integrated_channels/moodle/client.py
taliaedX: oh this reminded me actually...there's a spot where I too had array indices to deal with and I did iterate through them in the exporter. That may be what we want here if we do have the potential for multiple grade items in one post.

However, I don't think the format needs a comment, because this kind of formatting nonsense is sprinkled throughout every Moodle client/exporter/transmitter implementation and amounts to "nonsensical and inconsistent PHP array stuff"

integrated_channels/moodle/client.py
alex-sheehan-edx: it means I was rushing to get the PR up and said to myself "oh I'll just add the doc strings soon" 😬

integrated_channels/moodle/client.py
```diff
@@ -150,6 +151,100 @@ def _get_access_token(self):
...
+        """
+        placeholder
+        """
+        params = {
+            'wstoken': self.token,
+            'wsfunction': 'core_enrol_get_enrolled_users',
+            'courseid': course_id,
+            'moodlewsrestformat': 'json'
+        }
+        user_response = requests.get(
```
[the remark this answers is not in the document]
alex-sheehan-edx: yeah, I'm actually confused as to why this is succeeding... it should be a post. Agreed on the usage of the generalized wrapper.

integrated_channels/moodle/client.py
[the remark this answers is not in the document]
alex-sheehan-edx: Resolved offline! I believe this comes down to preferences of implementation. There are ways around the predefining of `user_id` to None, but I think my solution is perfectly reasonable here.

integrated_channels/moodle/client.py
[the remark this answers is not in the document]
alex-sheehan-edx: Tested locally- Seems like making both a post and get request to Moodle succeed and return the expected list of user enrollments? Coooooool. In that case, since we already have the `_post` wrapper method, perhaps I should just use that?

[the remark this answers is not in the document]
alex-sheehan-edx: yeah, this entire method can be simplified down to 
```
    @moodle_request_wrapper
    def _get_enrolled_users(self, course_id):
        """
        placeholder
        """
        params = {
            'wstoken': self.token,
            'wsfunction': 'core_enrol_get_enrolled_users',
            'courseid': course_id,
            'moodlewsrestformat': 'json'
        }
        return self._post(params)
```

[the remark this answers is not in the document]
alex-sheehan-edx: (finally taking the placeholder out now)

integrated_channels/moodle/client.py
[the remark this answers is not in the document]
taliaedX: This was also my misunderstanding. I have used `if not var` in a lot of places where the variable was not defined in the given code block, but didn't think about the fact that function parameters still count as definition.

integrated_channels/moodle/client.py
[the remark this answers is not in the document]
alex-sheehan-edx: Removed the comment but made sure to bring up with the team the usefulness of talking with Moodle/customers.

integrated_channels/moodle/client.py
alex-sheehan-edx: Looking at the UI on Moodle's side, the activities (what the user has to create in order for us to send them grades) don't seem to support letter grades, though there is a _different_ grade "item" (the one I originally thought would be our solution to our implementation) in Moodle that supports a grade type of `text`. It feels like this is how you'd get a letter grade. However having investigated and now discussed with the team, I don't think that configurable options are the best play for this PR. We can communicate with customers going forward and if there's interest then I think we can refer back to this thread.

integrated_channels/moodle/client.py
alex-sheehan-edx: annnnd replaced!

integrated_channels/moodle/client.py
alex-sheehan-edx: Fear not! The placeholders are gone!

integrated_channels/moodle/client.py
alex-sheehan-edx: It is not a constant, that's a great point and one that I meant to do but forgot. Good catch, thanks!!

integrated_channels/moodle/client.py
alex-sheehan-edx: Uhg this is really because of Moodle [`tl:dr`](https://github.com/moodle/moodle/blob/99777d963f740ced7920137345a43fdec187da11/lib/classes/grades_external.php#L432) 
The parameters are expected to be an array of grades to update: ie
```
[
   { 
      'studentid': <student ID 1>,
      'grade': <grade to be assigned>
   }
   { 
      'studentid': <student ID 2>,
      'grade': <grade to be assigned>
   }
]
```
So by saying `grades[0][FIELD]` we're creating a single entry for one grade item to be updated. Talia is pretty much spot on, there's the possibility of us doing bulk completion reporting. However, since the client only deals in single user ID's/payloads of only one enrollment, I don't think we can/need to create the ability for the bulk update (without some sort of refactoring of how the client works). I'm happy to explain this as a tl:dr inline comment.

integrated_channels/moodle/client.py
alex-sheehan-edx: 😠  agreed

integrated_channels/moodle/models.py
alex-sheehan-edx: 🤦 haha yeah, that's probably supposed to be an int/decimal, not even a char

integrated_channels/moodle/client.py
taliaedX: Ah yeah in the case I was formatting during the exporter phase, so yeah since we're not formatting this parameter until the client it should be fine. I'm not a huge fan of formatting it in that way (ie I'd prefer we do format in the exporter and can just chunk and send without too much extra formatting), but at that point we would have to use the method I had in the content_metadata exporter.

taliaedX: also is there a fundamental reason we can't send a bunch of grades at once? I know we've queried around one at a time, but Moodle does support multiple objects/grades unlike Canvas, so we don't necessarily have to chunk at 1.

taliaedX: ie if you leave the chunk size as its default (which is 100 or 500 or something) and have multiple students and grade items, does this function as is?

integrated_channels/moodle/models.py
alex-sheehan-edx: Yeah, I have the default here because we have other checks in the transmission flow that make sure the course is completed. I don't think it's necessary either way, but willing to make a change if you think it's better practice to "assume it's not until we prove it is"

integrated_channels/moodle/transmitters/learner_data.py
alex-sheehan-edx: 😠  why is there always one place that I forgot

integrated_channels/moodle/client.py
binodpant: yeah I think we may want to doc this somewhere on what to expect on grading .. or sth like it

integrated_channels/moodle/models.py
binodpant: based on how your code is setup this change is not needed, but yeah as a general dataa modeling concern perhaps it makes sense to default to False? then someone can explicitly set to True if conditions are right. I think a False for this value will mean no grade transmission, so wil leave it up to you

integrated_channels/moodle/client.py
alex-sheehan-edx: Looking at the base transmitter, I think you're right.  it should be able to handle chunk sizes larger than 1, but we would need to make some modifications to the Moodle exporter and transmitter. 

  From the base transmitter-
```
                code, body = self.client.create_course_completion(
                    getattr(learner_data, kwargs.get('remote_user_id')),
                    serialized_payload
                )
```
We would have to make sure that `learner_data['remote_user_id']` is exported as an array of user emails rather than a single one. As you say, and did in with the metadata exporter, the payload formatting (ie `'grades[0][studentid]'`) could be done in the moodle exporter rather than client if we're inclined to go in that direction. Do we want this implemented in this iteration?

binodpant (approved):
lgtm! if we find any issues we can address them later, but this seems like it will handle one student one grade item just fine

integrated_channels/moodle/client.py
taliaedX: hmmm I think we'd want to support multiple grade items immediately. I wouldn't mind leaving this until later if we had other work we definitely need to do, but since we are _ideally_ at just bug fixes and small features from customer/Ken feedback, I think it'd be best if we support this now

binodpant: I like the idea of supporting this this, but also end of sprint so, probably a new PR?

taliaedX: My argument is that I don't know if this even currently works when multiple items are present and a chunk size is not 1.

If we can process them one at a time successfully, then sure we can make a separate pr to support multiple items at once.

I also feel like just because Canvas cannot support multiple things at once, doesn't mean that should necessarily be our standard to call something done but that's just me.

binodpant: Hmm ok, sorry I probably misunderstood then. If this does not work when chunk size is set to more than 1 (despite moodle supporting multiple items) then yes it's worth fixing. Canvas was different since its api does not support multiple items at once. if that's not true of moodle, agree a fix is worth doing now rather than later

taliaedX: I meant to update this and forgot - Alex verified it still sends 1 learner's data at a time even with a >1 chunk size.

taliaedX (approved):
after discussing this all LGTM!

PR #986, opened 2020-10-01
enroll learner in course when DSC is disabled

kellmarie:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -600,70 +600,62 @@ class GrantDataSharingPermissions(View):
         # EnterpriseCourseEnrollment will be created when the user will select a course run.
         # A CourseEnrollment record will be created and on the post signal of the CourseEnrollment,
         # an EnterpriseCourseEnrollment record will also get created.
-        try:
-            CourseKey.from_string(course_id)
-        except InvalidKeyError:
-            LOGGER.error(
-                'Course with id: {} is not a valid course run id'.format(course_id)
-            )
-            raise
-
-        enrollment_api_client = EnrollmentApiClient()
-        course_modes = [mode['slug'] for mode in enrollment_api_client.get_course_modes(course_id)]
-        LOGGER.info(
-            'Retrieved Course Modes for Course {course_id}: {course_modes}'.format(
-                course_id=course_id,
-                course_modes=course_modes
-            )
-        )
-        course_mode = 'verified' if 'verified' in course_modes \
-            else 'professional' if 'professional' in course_modes \
-            else 'no-id-professional' if 'no-id-professional' in course_modes \
-            else 'audit'
-        try:
-            enrollment_api_client.enroll_user_in_course(
-                request.user.username,
-                course_id,
-                course_mode
-            )
+        if course_id and self.is_course_run_id(course_id):
+            enrollment_api_client = EnrollmentApiClient()
+            course_modes = [mode['slug'] for mode in enrollment_api_client.get_course_modes(course_id)]
             LOGGER.info(
-                'Created LMS enrollment for User {user} in Course {course_id} '
-                'with License {license_uuid} in Course Mode {course_mode}.'.format(
-                    user=request.user.username,
+                'Retrieved Course Modes for Course {course_id}: {course_modes}'.format(
                     course_id=course_id,
-                    license_uuid=license_uuid,
-                    course_mode=course_mode
+                    course_modes=course_modes
                 )
             )
-        except Exception as exc:  # pylint: disable=broad-except
-            LOGGER.error(
-                'Unable to create an LMS enrollment from the DSC view: {exc}'.format(
-                    exc=exc
+            course_mode = 'verified' if 'verified' in course_modes \
+                else 'professional' if 'professional' in course_modes \
+                else 'no-id-professional' if 'no-id-professional' in course_modes \
+                else 'audit'
+            try:
+                enrollment_api_client.enroll_user_in_course(
+                    request.user.username,
+                    course_id,
+                    course_mode
                 )
-            )
-            raise
-        try:
-            self.create_enterprise_course_enrollment(request, enterprise_customer, course_id, license_uuid)
-        except IntegrityError:
-            error_code = 'ENTGDS009'
-            log_message = (
-                '[Enterprise DSC API] IntegrityError while creating EnterpriseCourseEnrollment.'
-                'Course: {course_id}, '
-                'Program: {program_uuid}, '
-                'EnterpriseCustomer: {enterprise_customer_uuid}, '
-                'User: {user_id}, '
-                'License UUID: {license_uuid}, '
-                'ErrorCode: {error_code}'.format(
-                    course_id=course_id,
-                    program_uuid=program_uuid,
-                    enterprise_customer_uuid=enterprise_customer.uuid,
-                    user_id=request.user.id,
-                    license_uuid=license_uuid,
-                    error_code=error_code,
+                LOGGER.info(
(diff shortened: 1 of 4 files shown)


iloveagent57 (commented):
Just a small suggestion about your helper function signature, looks like there's some tests to fix up in `tests/test_enterprise/views/test_grant_data_sharing_permissions.py`, but this looks like the right thing to do.

enterprise/views.py
```diff
@@ -585,6 +585,84 @@ def create_enterprise_course_enrollment(request, consent_record, course_id, lice
...
+                            course_mode=course_mode
+                        )
+                    )
+                except Exception as exc:  # pylint: disable=broad-except
+                    LOGGER.error(
+                        'Unable to create an LMS enrollment from the DSC view: {exc}'.format(
+                            exc=exc
+                        )
+                    )
+                    return redirect(failure_url)
```
chavesj, line 640: piggybacking on Alex's comment re: success_url and failure_url not being needed here - I'd avoid catching the exception in here with a redirect since redirecting to a failure URL is an over-opinionated way of handling an enrollment error. Keep this logging in place but re-throw the exception, catch it where the util method is used and avoid having to pass the URLs as args to the utility method

enterprise/views.py
```diff
@@ -585,6 +585,84 @@ def create_enterprise_course_enrollment(request, consent_record, course_id, lice
...
+                    'ErrorCode: {error_code}'.format(
+                        course_id=course_id,
+                        program_uuid=program_uuid,
+                        enterprise_customer_uuid=enterprise_customer.uuid,
+                        user_id=request.user.id,
+                        license_uuid=license_uuid,
+                        error_code=error_code,
+                    )
+                )
+                LOGGER.exception(log_message)
```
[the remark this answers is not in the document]
kellmarie: yeah, this was copy and pasted from the current post function so I might have missed a line. I'll return False here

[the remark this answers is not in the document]
chavesj: since this method will already be throwing an exception maybe we also re-throw this Integrity Error after logging

chavesj (commented):
I think the trickiest part of this PR will be resolving the broken tests (ensuring all other DSC use cases still work as expected), but for this specific use case it looks like you're doing the right thing in the right place. Also, to supplement my suggestion to re-throw exceptions from your utility method here's [a 2007 article](https://nedbatchelder.com/blog/200711/rethrowing_exceptions_in_python.html) from @nedbat I just came across.

enterprise/views.py
```diff
@@ -886,74 +973,18 @@ def post(self, request):
...
-                    )
-                    LOGGER.exception(log_message)
+            successful_enrollment = self._enroll_leaner_in_course(
+                request=request,
+                enterprise_customer=enterprise_customer,
+                course_id=course_id,
+                program_uuid=program_uuid,
+                license_uuid=license_uuid)
             consent_record.granted = consent_provided
             consent_record.save()
```
chavesj, line 985: should the consent record still be updated if the enrollment fails?

kellmarie: It looks like it was originally. I think we should maintain the logic how it was for now and investigate whether or not the consent record should be updated later.

chavesj: previously, the redirect (on failure) would happen before the consent record is updated so to keep with existing logic it shouldn't be updated after failure here

chavesj: another reason to wrap these calls in a try/except and rethrow exceptions from within the enrollment util method if you ask me 🤷‍♂️

enterprise/views.py
```diff
@@ -887,73 +976,19 @@ def post(self, request):
...
+                self._enroll_leaner_in_course(
+                    request=request,
+                    enterprise_customer=enterprise_customer,
+                    course_id=course_id,
+                    program_uuid=program_uuid,
+                    license_uuid=license_uuid)
+            except Exception:  # pylint: disable=broad-except
+                pass
             consent_record.granted = consent_provided
             consent_record.save()
```
chavesj, line 989: The consent record update can go within the try block and we should be redirecting to the failure_url when we catch the exception

enterprise/views.py
```diff
@@ -887,73 +976,19 @@ def post(self, request):
...
 
-        return redirect(success_url if consent_provided else failure_url)
+        return redirect(success_url) if consent_provided else redirect(failure_url)
```
chavesj, line 991: nit: I don't see a reason this line should change

enterprise/views.py
```diff
@@ -585,6 +585,82 @@ def create_enterprise_course_enrollment(request, consent_record, course_id, lice
...
+                        enterprise_customer_uuid=enterprise_customer.uuid,
+                        user_id=request.user.id,
+                        license_uuid=license_uuid,
+                        error_code=error_code,
+                    )
+                )
+                LOGGER.exception(log_message)
+                raise
+            return True
+        return False
```
chavesj, line 662: these booleans aren't being used since an exception is raised on failure, should be safe to remove

kellmarie: False gets returned if the course id that was provided isn't for a valid course

enterprise/views.py
kellmarie: because I had changed it to be something else, and then changed it back to consent_provided, and then didn't change it fully back. I'll do that.

chavesj (commented):
this is looking better now that the existing tests have been fixed but it still needs testing for the new logic introduced

enterprise/views.py
chavesj: ah, I was looking at the usage within the post() method where the return value doesn't get used. To rephrase: since we're raising an exception for other failures here we should also raise an exception in the case that an invalid course is provided to be consistent

kellmarie: Or wait it's something about course_runs I guess? Not exactly sure what that means

chavesj: The course_id needs to be for a course run not a course for that conditional to be true, but that doesn't need to change. What I was saying is that a return value shouldn't be needed from this method. 

After taking a closer look I also realized that the previous behavior didn't fail in any way if that line evaluated to false, it just skipped creating the enrollment. Knowing that, I suggest removing the return statements in `_enroll_leaner_in_course`, which should allow you to simply redirect to the `success_url` in the try block within the `get()` method

enterprise/views.py
kellmarie: Well all the current unit tests seem to document that that's not the expected behavior and I don't think we should be changing up the expected behavior in a section of the code not being actively worked on.

If you _really_ want me to change up the behavior then I'm going to need to fix a lot of unit tests and manually check those enrollment cases for regressions and given the DSC screen doesn't currently work on my local environment we'd probably need to repoint this ticket and throw on a few more points

chavesj: the point of the exceptions are to know when to redirect to the `failure_url` in the method you are calling `_enroll_leaner_in_course` from. as of right now you don't redirect on failure the same way as the original logic did

In the case of the `InvalidKeyError` and `IntegrityError` you should not be re-raising the exception because the original logic didn't redirect to `failure_url` in those cases. For the general `Exception` catch block in `_enroll_leaner_in_course`, where a redirect to the `failure_url` was originally triggered on failure, you should be re-raising the exception so that you know to redirect from the calling method

chavesj: I'm happy to hop on a call if this needs more explanation, I know communicating solely through text can be difficult

iloveagent57 (approved):
Nice job, I like the way this turned out.

chavesj (approved):
:shipit:

PR #988, opened 2020-10-05
Fix 414s from MoodleCloud (possibly other Moodle providers/hosts in future TBD)?

taliaedX:
### Description:
Most of the changes in this PR are just copypasta from the base transmitter class, there are only a couple of new lines in each transmit function and they are all the same:
1. chunk size parameter and variable definition
2. catching 414 specifically and doing a recursive.

**Merge checklist:**

**Post merge:**


integrated_channels/moodle/exporters/content_metadata.py
```diff
@@ -25,7 +25,7 @@ class MoodleContentMetadataExporter(ContentMetadataExporter):
         'categoryid': 'categoryid',
     }
 
-    LONG_STRING_LIMIT = 2000
+    LONG_STRING_LIMIT = 500
```
taliaedX, line 28: so for Moodlecloud, the 2000 length limit + 5 courses worked successfully, but I feel like some combination of lower description limit and item slicing could be better? Idk. Maybe having the full description is _that_ important.

integrated_channels/moodle/transmitters/content_metadata.py
```diff
@@ -2,9 +2,19 @@
...
+from itertools import islice
 
+from django.conf import settings
+
+from integrated_channels.exceptions import ClientError
 from integrated_channels.integrated_channel.transmitters.content_metadata import ContentMetadataTransmitter
 from integrated_channels.moodle.client import MoodleAPIClient
+from integrated_channels.utils import chunks
+
+LOW_CHUNK_SIZE = 5
```
taliaedX, line 15: A little arbitrary, but was able to transmit 5 courses with the full 2000 character limit long description successfully.

I thought of just doing a division each attempt, but that might be taxing performance-wise and I didn't feel like utilizing rounding to make sure I could divide by 2 (or whatever digit, but by half each time seemed reasonable)

integrated_channels/moodle/transmitters/content_metadata.py
```diff
@@ -21,6 +31,96 @@ def __init__(self, enterprise_configuration, client=MoodleAPIClient):
...
+        transmission_limit = settings.INTEGRATED_CHANNELS_API_CHUNK_TRANSMISSION_LIMIT.get(
+            self.enterprise_configuration.channel_code()
+        )
+        for chunk in islice(chunk_items, transmission_limit):
+            serialized_chunk = self._serialize_items(list(chunk.values()))
+            try:
+                self.client.create_content_metadata(serialized_chunk)
+            except ClientError as exc:
+                if exc.status_code == 414:
+                    self._transmit_create(channel_metadata_item_map, chunk_size=LOW_CHUNK_SIZE)
```
taliaedX, line 49: I realized a scenario:
1 chunk has very small descriptions/other data so it gets successfully transmitted at the current chunk size.
Then the next chunk has very wordy content_metadata and hits this error.

With the way this is constructed, I am not 100% sure if it would attempt to resubmit the successful items. I *imagine* it would not due to the _create_transmissions call, but this is an elaborate test case, so I haven't set it up.

integrated_channels/moodle/transmitters/content_metadata.py
```diff
@@ -31,7 +34,7 @@ def __init__(self, enterprise_configuration, client=MoodleAPIClient):
             client=client
         )
 
-    def _transmit_create(self, channel_metadata_item_map, chunk_size=None):  # pylint: disable=arguments-differ
+    def _transmit_create(self, channel_metadata_item_map, count=0):  # pylint: disable=arguments-differ
```
taliaedX, line 37: note: if this looks fine, I need to implement this for the rest of the `_transmit_*` classes

taliaedX: also I do still need chunk size damnit. I forgot why that was there.

integrated_channels/moodle/exporters/content_metadata.py
```diff
@@ -37,18 +37,22 @@ def transform_description(self, content_metadata_item):
...
             description = "{base_description}{full_description}".format(
                 base_description=base_description,
                 full_description=full_description
             )
-        else:
-            short_description = content_metadata_item.get(
-                'short_description', content_metadata_item.get('title', '')
-            )
+        elif short_description and len(short_description + enrollment_url) <= self.LONG_STRING_LIMIT:
+            short_description = content_metadata_item.get('short_description')
```
binodpant, line 47: think you can re use the short_description variable from line 40?

integrated_channels/moodle/models.py
```diff
@@ -83,6 +83,11 @@ class MoodleEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguratio
         )
     )
 
+    transmission_chunk_size = models.IntegerField(
+        default=4,
```
binodpant, line 87: I like this default value setup, perhaps we can just default to 1 in canvas (why did we not do that ?)

integrated_channels/moodle/transmitters/content_metadata.py
```diff
@@ -21,6 +34,105 @@ def __init__(self, enterprise_configuration, client=MoodleAPIClient):
...
+                    )
+                    LOGGER.exception(exc)
+            else:
+                self._update_transmissions(chunk, transmission_map)
+
+    def _transmit_delete(self, channel_metadata_item_map, chunk_size=None, count=0):  # pylint: disable=arguments-differ
+        """
+        Transmit content metadata deletion to integrated channel.
+        """
+        chunk_size = chunk_size if chunk_size else self.enterprise_configuration.transmission_chunk_size
```
binodpant, line 108: I feel like since this snippet is now used several places including base transmitter, does it makes sense now to make it a method in the base class and re use it here?

integrated_channels/moodle/transmitters/content_metadata.py
```diff
@@ -21,6 +34,105 @@ def __init__(self, enterprise_configuration, client=MoodleAPIClient):
...
+        chunk_items = chunks(channel_metadata_item_map, chunk_size)
+        transmission_limit = settings.INTEGRATED_CHANNELS_API_CHUNK_TRANSMISSION_LIMIT.get(
+            self.enterprise_configuration.channel_code()
+        )
+        for chunk in islice(chunk_items, transmission_limit):
+            serialized_chunk = self._serialize_items(list(chunk.values()))
+            try:
+                self.client.delete_content_metadata(serialized_chunk)
+            except ClientError as exc:
+                if exc.status_code == 414 and count <= 1:
```
binodpant, line 118: I am a bit unclear on this change and the description change combined. Even despite the description change you made, will we expect to hit 414s? I am wondering if we also considered the alternative which is to send truncated descriptions and avoid the problem (basically ensure max desc sizes are set by us). I think clicking on edX link can take us to the full description website anyway, is having the full description in moodle a must? may want to discuss this one in team channel... it would make things simpler for us (and thanks Moodle for GET apis!)

integrated_channels/moodle/transmitters/content_metadata.py
```diff
@@ -21,6 +34,105 @@ def __init__(self, enterprise_configuration, client=MoodleAPIClient):
...
+        )
+        for chunk in islice(chunk_items, transmission_limit):
+            serialized_chunk = self._serialize_items(list(chunk.values()))
+            try:
+                self.client.update_content_metadata(serialized_chunk)
+            except ClientError as exc:
+                if exc.status_code == 414 and count <= 1:
+                    count += 1
+                    sleep(5)
+                    self._transmit_update(channel_metadata_item_map, transmission_map, chunk_size=LOW_CHUNK_SIZE, count=count)
```
binodpant, line 89: just a thought out loud is that if we can find a way to determine the max chunk size we can use without running into the limit the we use it here in the first place, and never actually hit 414 (ideally), but if we still hit 414s then yeah ... looks like we need some defense too

binodpant (approved):
I am a go since it's a big improvement, I think we should think about an optimized way to guess the chunk size we can use to avoid 414s pre emptively though which may simplify our code, but I see that we may still need to handle 414s

also bonus thought, why not move the 414 handling to base transmitter! seems like any channel can hit this theoretically

integrated_channels/moodle/exporters/content_metadata.py
taliaedX: oh right. I added the short_description definition up top after having to define it so many times, but then apparently forgot to remove those definitions! 😅

integrated_channels/moodle/models.py
taliaedX: I was wondering that as well and actually went looking at Canvas' models expecting to see this override. I think we just limit it in the client or transmitter?

integrated_channels/moodle/transmitters/content_metadata.py
taliaedX: To summarize what we discussed via slack, there's a few things here:
1. I was trying to maximize the number of items we send against the risk of 414s, which is why I didn't pick just a chunk size of 1 (which is *the* safest bet)

2. The actual max URL size is not a set standard and will vary by Moodle host. For instance, I didn't even encounter this issue when testing locally. It was only when I tested against Moodlecloud that I had this issue. Jeff pointed out that nginx has a default url limit and since we both know nginx is used very widely, it seemed best to start calculating an average/safe default based on their default max URL length.
Here's some TL;DR on those calcs:
- Nginx's limit is 8k bytes.
- I used 3 bytes as the max to conservatively account for non-latin characters (such as Chinese, Korean, and other special characters). This means we have a 2666 (rounded down) character limit.
- I calculated the length of the parameter names which at worst look like `[course][fieldname][0]`
- I calculated the max date length (when in timestamp format) and used that (which equates to the year 3001, so again very conservative) and then doubled that since we have to deal with 2 dates (start & end)
- I calculated the average length from prod data and also used the max char length (255 for everything except the descriptions) of all the content metadata fields we use to come up with some conservative and more lenient basis for max description length.
- I added additional padding and rounded down a bit (50ish chars) to cover things I missed (like other weird array parameters, quotation marks, equal signs which I forgot about, the length of the hostname which I cannot determine).
- So yeah it's _very_ conservative. Even at 4 items, we should hit the 414 basically never. (I didn't hit it with my original guess of 5 either)

3. Another general note on strategies: Binod mentioned a webservice plugin which wraps Moodle's existing webservice functions in proper rest format. I think this is nice, but think the additional time to convert our existing code and added requirements/procedures on our customers makes it probably not ideal to use at this point in time.

integrated_channels/moodle/transmitters/content_metadata.py
taliaedX: Not sure I grasp. This chunk size definition is not used in the base class. I added the chunk_size parameter and thus needed to account for it here.

integrated_channels/moodle/transmitters/content_metadata.py
taliaedX: Also without limiting the description, even if we have a chunk size of 1, _if_ the description *does* contain non-latin characters, we can potentially hit the 414 on a single item.

integrated_channels/moodle/models.py
binodpant: for canvas we rely on the config setting it to 1 (not a good idea!) but the default value would be a good idea I think

integrated_channels/moodle/transmitters/content_metadata.py
binodpant: oh not sure if I misread I thought the few lines that produce the tuple values were the same in all the methods, but if the base does not also have the same code lines probably not worth it. was looking for avoidance of dup if possible, if anything you can make it a shared method in this very class instead too if you like

integrated_channels/moodle/transmitters/content_metadata.py
binodpant: aargh yeah that does limit us quite a bit then hah! damn api that uses query strings...

Since it looks like worst case we can still hit 414 even with 1 chunk size, I will make one last pitch to see if we can get away with truncating the long description until we do fall within limit. Basically we still send it but truncated. I think we can convince product folks that is okay to do. Basically [long text with ... at the end]. Since edX will show full description anyway.  

But having said that, your idea of using the short only description also is a variation of that anyway so I am also okay with this idea.

integrated_channels/moodle/transmitters/content_metadata.py
binodpant: yeah I think your idea given all the limitations is also fine. The only remaining alternative I had proposed was truncated full_description but always including it. If that is also a problematic, yeah I think it's best to go with your idea of sending limited description fields.

integrated_channels/moodle/transmitters/content_metadata.py
taliaedX: I think the base integrated channels has a util function `chunks` method we call later that also does islicing and I didn't dig into deeply to understand why we seem to duplicate that even in the base transmitter, but other than that no, no duplication really here except obviously defining this over and over for the 3 overrides.

yeah I had thought about just having an entire wrapper honestly, since ultimately all of these `_transmit_*` overrides are just: use chunk size, attempt a transmission, recurse on 414. But the wrapping inception and recursiveness seemed gross together and a nightmare to eventually test.

integrated_channels/moodle/transmitters/content_metadata.py
taliaedX: I mean that's why I'm using shorter values and that "use short_description if full description is too long" logic exists in other integrated_channels implementations, so it's not a unique hack here. I am just being safe around the short_description as well since it **technically** is a "longtext" field which can be up to 4gb. We don't have any data in our production database that goes anywhere remotely near this length, but it _is_ possible.

taliaedX: Based on this prior art, it seemed like the consensus was that a short description was better than arbitrarily cutting off the full description.

integrated_channels/moodle/transmitters/content_metadata.py
taliaedX: yeah as I mention above, prior art suggests a complete short description is better than a partial full description (though it is interesting that no one checked the length limits on short descriptions, before just auto-defaulting to them in place of the full description).

There is no technical limitation to just truncating the full description and always including it, but I'm not sure it is a better alternative.

PR #989, opened 2020-10-05
ENT-3472: blackboard exporter

binodpant:
- [X] still need to ensure discovery back from blackboard to edX but otherwise can create a course at least with this json output from exporter
     - Have a temporary solution to this by adding a link to enrollment_url in the description field but I am too psyched about it, will chat with Ken later today and explore ways to make this better. But this basically allows learners to find edX from the search page at least.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/blackboard/exporters/content_metadata.py
+++ b/integrated_channels/blackboard/exporters/content_metadata.py
@@ -28,8 +28,4 @@ class BlackboardContentMetadataExporter(ContentMetadataExporter):
         """
         enrollment_url = content_metadata_item.get('enrollment_url', None)
         url_link = self.DESCRIPTION_TEXT_TEMPLATE.format(enrollment_url=enrollment_url)
-        short_description = content_metadata_item.get(
-            'short_description', content_metadata_item.get('title', '')
-        )
-        description = "<p>{}</p><p>{}</p>".format(short_description, url_link)
-        return description
+        return url_link
--- a/integrated_channels/blackboard/exporters/content_metadata.py
+++ b/integrated_channels/blackboard/exporters/content_metadata.py
@@ -16,6 +16,20 @@ class BlackboardContentMetadataExporter(ContentMetadataExporter):
     DATA_TRANSFORM_MAPPING = {
         'name': 'title',
         'externalId': 'key',
-        'description': 'full_description',
+        'description': 'description',
         'courseId': 'key',
     }
+
+    DESCRIPTION_TEXT_TEMPLATE = "<a href={enrollment_url}>Go to edX course page</a><br />"
+
+    def transform_description(self, content_metadata_item):
+        """
+        This will show a link to edX course on blackboard course description
+        """
+        enrollment_url = content_metadata_item.get('enrollment_url', None)
+        url_link = self.DESCRIPTION_TEXT_TEMPLATE.format(enrollment_url=enrollment_url)
+        short_description = content_metadata_item.get(
+            'short_description', content_metadata_item.get('title', '')
+        )
+        description = "<p>{}</p><p>{}</p>".format(short_description, url_link)
+        return description
--- a/integrated_channels/blackboard/exporters/content_metadata.py
+++ b/integrated_channels/blackboard/exporters/content_metadata.py
@@ -12,17 +12,22 @@ LOGGER = getLogger(__name__)
 class BlackboardContentMetadataExporter(ContentMetadataExporter):
     """
         Blackboard implementation of ContentMetadataExporter.
+        The odd choice of making enrollment_url into description with a link is best explained as:
+            Blackboard does not show description by default on home page.
+            However, description shows on the course search page.
+            This seems to be the only place to visibly place a link for now for learners to
+            find and visit the edX course page. But this can change as we explore blackboard more.
     """
     DATA_TRANSFORM_MAPPING = {
         'name': 'title',
         'externalId': 'key',
-        'description': 'description',
+        'description': 'enrollment_url',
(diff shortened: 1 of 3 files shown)


binodpant:
So I came up with a small improvement where the description shown on the course search page now contains a link to the edX course. At least this lets the user search a course and click on the edX link. IT does not yet allow clicking to edX from the home page of the course.

also: If I try to use any extra formatting it does not display well. So I decided to just add a link here.

There may be a way for me to customize the way the home page looks but I have not yet figured out (editing the banner inserting a link there is an option but not the greatest)

integrated_channels/blackboard/exporters/content_metadata.py
```diff
@@ -13,3 +13,19 @@ class BlackboardContentMetadataExporter(ContentMetadataExporter):
...
+
+    DESCRIPTION_TEXT_TEMPLATE = "<a href={enrollment_url}>Go to edX course page</a><br />"
+
+    def transform_description(self, content_metadata_item):
+        """
+        This will show a link to edX course on blackboard course description
+        """
+        enrollment_url = content_metadata_item.get('enrollment_url', None)
+        url_link = self.DESCRIPTION_TEXT_TEMPLATE.format(enrollment_url=enrollment_url)
+        return url_link
```
taliaedX, line 31: wouldn't this transform the description into just a link?
Ie instead of "This course will teach you blah blah blah....[enroll here]" you get "[enroll here]"

binodpant:
So I am going to chat with Ken about the product view on this and how best we can represent the edX link in blackboard, and just generally talk about discovery experience in blackboard, at the end of the day today. In the meantime I will switch to other tasks.

Therefore, we can go either way on this, 
1. approve this PR and I can come back and update the exporter per chat with Ken, it's possible changes also needed in transmitter but hopefully not
2. hold on to this PR until Ken and I have resolved that question

tests/test_integrated_channels/test_blackboard/test_exporters/test_content_metadata.py
```diff
@@ -0,0 +1,74 @@
...
+    @mock.patch('enterprise.api_client.enterprise_catalog.EnterpriseCatalogApiClient.get_content_metadata')
+    def test_content_exporter_export(self, mock_get_content_metadata):
+        """
+        ``BlackboardContentMetadataExporter``'s ``export`` produces the expected export.
+        """
+        fake_content_metadata = OrderedDict()
+        fake_content_metadata[FAKE_COURSE_RUN['key']] = FAKE_COURSE_RUN
+        fake_content_metadata[FAKE_COURSE['key']] = FAKE_COURSE
+
+        mock_get_content_metadata.return_value = list(fake_content_metadata.values())
```
taliaedX, line 48: ah good choice here on this test. Nice and contained.

tests/test_integrated_channels/test_blackboard/test_exporters/test_content_metadata.py
```diff
@@ -0,0 +1,74 @@
...
+            )
+
+    def test_transform_description(self):
+        content_metadata_item = {
+            'enrollment_url': 'http://some/enrollment/url/',
+            'short_description': 'short desc',
+        }
+        exporter = BlackboardContentMetadataExporter('fake-user', self.config)
+        description = exporter.transform_description(content_metadata_item)
+        expected_description = exporter.DESCRIPTION_TEXT_TEMPLATE.format(
```
taliaedX, line 72: your expectation here is also just expecting a link rather than the description.

taliaedX (commented):
I have one issue with the description, but otherwise this looks fine.

integrated_channels/blackboard/exporters/content_metadata.py
binodpant: correct observation! Indeed not my fav solution as of now. The issue I found is that 'description' does not show up anywhere on course home page on blackboard. to make the course link findable in blackboard the only location I have found so far is the 'description' column shows in the search page. Hence this 'hack'

pro of approach: What this does do it nicely show a link when someone searches for a course. Hence helps with course discovery.
con: blackboard description is just a link

so perhaps I can come back and re-address this once I chat with Ken. For now, this seems to be the 'best' solution

taliaedX: Hmmm I guess I'm still confused why it's only a link instead of the description + link? Assuming there is nowhere else that we'd use the description, why not put all of it there?

Otherwise, if the plan is for this to just be a link (whether it be for this PR or permanently) then you don't need to map it to the short_description field. Just map it to the enrollment url and then modify it slightly in your "transform_description" method.

binodpant: yeah putting description there right now messes up how it shows up on the course search page, however we could put a shorter description plus link there if we want. But yes if I wanted just a link there, I should map to the enrollment_url field instead of description, yes. let me try that change for a incremental improvement

binodpant: so that would mean just changing transform_description to transform_enrollment_url right?

like was thinking

```
DATA_TRANSFORM_MAPPING = {
...
        'description': 'enrollment_url',
...
    }
```

and 

```
    def transform_enrollment_url(self, content_metadata_item):

```

binodpant: made this change. still not feeling awesome of a solution, I will expect to improve on this as we continue our work with blackboard

taliaedX (approved):
LGTM

PR #991, opened 2020-10-06
[BUG] Moodle client return error fix

alex-sheehan-edx:
Base transmitter expects a tuple of (code, body)

**Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,11 @@ Change Log
 
 Unreleased
 
+[3.8.30]
+--------
+
+* Moodle client bug fix
+
 [3.8.29]
 --------
 
@@ -30,12 +35,10 @@ Unreleased
 
 * Update ``get_service_usernames()`` to read from a list variable (that may not exist).
 
-
-[3.8.26] 2020-10-05
--------------------
+[3.8.26]
+--------
 
 * Moodle completion data implementation
-
 
 [3.8.25]
 --------
@@ -46,7 +49,6 @@ Unreleased
 -------------------
 
 * Allow learners to enroll with their license in courses when DSC is disabled.
-
 
 [3.8.23] 2020-10-01
 -------------------
--- a/integrated_channels/moodle/client.py
+++ b/integrated_channels/moodle/client.py
@@ -275,6 +275,31 @@ class MoodleAPIClient(IntegratedChannelApiClient):
         return parsed_response['courses'][0]['id']
 
     @moodle_request_wrapper
+    def _wrapped_create_course_completion(self, user_id, payload):
+        """
+        Wrapped method to request and use Moodle course and user information in order
+        to post a final course grade for the user.
+        """
+        completion_data = json.loads(payload)
+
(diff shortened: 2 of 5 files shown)


CHANGELOG.rst
```diff
@@ -25,13 +30,11 @@ Unreleased
 
 * Update ``get_service_usernames()`` to read from a list variable (that may not exist).
 
-
```
alex-sheehan-edx, line 28: Just want things to be consistent...

CHANGELOG.rst
```diff
@@ -15,6 +15,11 @@ Change Log
 
 Unreleased
 
+[3.8.26] 2020-10-06
```
binodpant, line 18: version is wrong

alex-sheehan-edx: 🤦

integrated_channels/moodle/client.py
```diff
@@ -337,7 +336,9 @@ def create_course_completion(self, user_id, payload):  # pylint: disable=unused-
             'grades[0][grade]': completion_data['grade'] * 100
         }
 
-        return self._post(params)
+        # The base learner data transmitter expects a (code, body) tuple to be returned
```
binodpant, line 339: it would be good to see if there is a test that captures this.. any potential to do that?

alex-sheehan-edx: Uhh, yeah! Can do

integrated_channels/moodle/client.py
```diff
@@ -317,7 +317,6 @@ def delete_content_metadata(self, serialized_data):
             querystring=urlencode(course_ids_to_delete))
         return self._post(params, url)
 
-    @moodle_request_wrapper
```
taliaedX, line 320: I think we need this to ensure a token though. I wonder if there is another way to use the wrapper here to get the right result.

taliaedX (commented):
I have concerns about removing the wrapper.

CHANGELOG.rst
```diff
@@ -15,6 +15,11 @@ Change Log
 
 Unreleased
 
+[3.8.29] 2020-10-06
```
taliaedX, line 18: _sigh_ so now we're being inconsistent with the dates (not your fault, there's a few examples here)
I know Brandon led the way in getting rid of them, but we seem to be inconsistently adding them back in.
Let's remove it, to try to solidify the trend of removing them :)

integrated_channels/moodle/client.py
alex-sheehan-edx: I think this is doable. One sec.

CHANGELOG.rst
alex-sheehan-edx: Mixed feelings, but meh- want to be consistent regardless

taliaedX (approved):
the new wrapped method and call works for me (I mean as long as it works locally in actual functionality lol)

binodpant (approved):
lgtm!

PR #992, opened 2020-10-06
Refactor license_revoke endpoint into bite-sized pieces.

iloveagent57:
Doing this as a first part of https://openedx.atlassian.net/browse/ENT-3414
Makes no functional changes, just refactors the revoke endpoint into smaller parts, so that implementing new logic is easier to manage.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -258,6 +258,13 @@ class EnterpriseCourseEnrollmentViewSet(EnterpriseReadWriteModelViewSet):
         return serializers.EnterpriseCourseEnrollmentWriteSerializer
 
 
+class EnrollmentModificationException(Exception):
+    """
+    An exception that represents an error when modifying the state
+    of an enrollment via the EnrollmentApiClient.
+    """
+
+
 class LicensedEnterpriseCourseEnrollmentViewSet(EnterpriseWrapperApiViewSet):
     """
     API views for the ``licensed-enterprise-course-enrollment`` API endpoint.
@@ -266,7 +273,8 @@ class LicensedEnterpriseCourseEnrollmentViewSet(EnterpriseWrapperApiViewSet):
     queryset = models.LicensedEnterpriseCourseEnrollment.objects.all()
     serializer_class = serializers.LicensedEnterpriseCourseEnrollmentReadOnlySerializer
 
-    def _validate_license_revoke_data(self, request_data):
+    @staticmethod
+    def _validate_license_revoke_data(request_data):
         """
         Ensures the request data contains the necessary information.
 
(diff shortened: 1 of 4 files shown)


enterprise/api/v1/views.py
```diff
@@ -282,6 +283,94 @@ def _validate_license_revoke_data(self, request_data):
...
+    def _enrollments_by_course_for_licensed_user(self, enterprise_customer_user):
+        """
+        Helper method to return a dictionary mapping course ids to EnterpriseCourseEnrollments
+        for each licensed enrollment associated with the given enterprise user.
+        """
+        licensed_enrollments = self.queryset.filter(
+            enterprise_course_enrollment__enterprise_customer_user=enterprise_customer_user
+        )
+        return {
+            enrollment.enterprise_course_enrollment.course_id: enrollment.enterprise_course_enrollment
```
bbaker6225, line 314: We probably want to select/prefetch related on this filter here since we then reference `course_id` on the related objects in this for loop

enterprise/api/v1/views.py
```diff
@@ -282,6 +283,94 @@ def _validate_license_revoke_data(self, request_data):
...
+        and saves the enterprise enrollment for later.
+        """
+        licensed_enrollment = enterprise_enrollment.license
+        licensed_enrollment.is_revoked = True
+        licensed_enrollment.save()
+
+        enterprise_enrollment.saved_for_later = True
+        enterprise_enrollment.save()
+
+    def _revoke_enrollment(self, enrollment_api_client, enterprise_enrollment, course_overview):
```
bbaker6225, line 331: nit: it might be just because I'm new to this area of the code, but I feel like having one of these methods detailing the type of the params (mainly `course_overview`) in the docstring could be helpful

bbaker6225 (commented):
Small nit but otherwise seems good to me (other than the comment for avoiding the N+1 queries)

enterprise/api/v1/views.py
bbaker6225: Looks like it should be select related since it's a OneToOne

iloveagent57: Good catch.

enterprise/models.py
```diff
@@ -1473,6 +1473,26 @@ class LicensedEnterpriseCourseEnrollment(TimeStampedModel):
 
     history = HistoricalRecords()
 
+    @classmethod
+    def enrollments_for_user(cls, enterprise_customer_user):
```
bbaker6225, line 1477: Amazing

bbaker6225 (approved):

bbaker6225:
Approval pending the version conflict in __init__.py and the changelog of course

chavesj (approved):
🚢

PR #994, opened 2020-10-08
During revocation, unenroll if audit track does not exist

iloveagent57:
https://openedx.atlassian.net/browse/ENT-3414

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -45,17 +45,19 @@ from enterprise.api.v1 import serializers
 from enterprise.api.v1.decorators import require_at_least_one_query_parameter
 from enterprise.api.v1.permissions import IsInEnterpriseGroup
 from enterprise.api_client.lms import EnrollmentApiClient
-from enterprise.constants import COURSE_KEY_URL_PATTERN, CourseModes
+from enterprise.constants import COURSE_KEY_URL_PATTERN
 from enterprise.errors import CodesAPIRequestError
 from enterprise.utils import NotConnectedToOpenEdX, get_request_value
 from enterprise_learner_portal.utils import CourseRunProgressStatuses, get_course_run_status
 
 try:
+    from course_modes.models import CourseMode
     from lms.djangoapps.certificates.api import get_certificate_for_user
     from openedx.core.djangoapps.content.course_overviews.api import get_course_overviews
 except ImportError:
     get_course_overviews = None
     get_certificate_for_user = None
+    CourseMode = None
 
 
 LOGGER = getLogger(__name__)
@@ -273,6 +275,15 @@ class LicensedEnterpriseCourseEnrollmentViewSet(EnterpriseWrapperApiViewSet):
     queryset = models.LicensedEnterpriseCourseEnrollment.objects.all()
     serializer_class = serializers.LicensedEnterpriseCourseEnrollmentReadOnlySerializer
 
+    class RevocationStatus:
+        """
+        Defines statuses related to enrollment states during the revocation process.
+        """
+        COURSE_COMPLETED = 'course already completed'
+        MOVED_TO_AUDIT = 'moved to audit'
+        UNENROLLED = 'unenrolled'
+        UNENROLL_FAILED = 'unenroll_user_from_course returned false.'
+
     @staticmethod
     def _validate_license_revoke_data(request_data):
         """
@@ -343,12 +354,9 @@ class LicensedEnterpriseCourseEnrollmentViewSet(EnterpriseWrapperApiViewSet):
             course_overview (CourseOverview): The course overview object associated with the enrollment. Used
                 to check for course completion.
         """
-        if self._has_user_completed_course_run(enterprise_enrollment, course_overview):
-            return
-
         course_run_id = course_overview.get('id')
         enterprise_customer_user = enterprise_enrollment.enterprise_customer_user
-        audit_mode = CourseModes.AUDIT
+        audit_mode = CourseMode.AUDIT
         enterprise_id = enterprise_customer_user.enterprise_customer.uuid
 
         log_message_kwargs = {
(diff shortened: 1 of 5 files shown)


enterprise/api/v1/views.py
```diff
@@ -343,12 +345,9 @@ def _revoke_enrollment(self, enrollment_api_client, enterprise_enrollment, cours
             course_overview (CourseOverview): The course overview object associated with the enrollment. Used
                 to check for course completion.
         """
-        if self._has_user_completed_course_run(enterprise_enrollment, course_overview):
-            return
-
         course_run_id = course_overview.get('id')
         enterprise_customer_user = enterprise_enrollment.enterprise_customer_user
-        audit_mode = CourseModes.AUDIT
+        audit_mode = CourseMode.AUDIT
```
adamstankiewicz, line 350: What if `CourseMode` is `None` due to the try/except with the import?

enterprise/api/v1/views.py
```diff
@@ -406,14 +441,21 @@ def license_revoke(self, request, *args, **kwargs):
         enrollments_by_course_id = self._enrollments_by_course_for_licensed_user(enterprise_customer_user)
 
         enrollment_api_client = EnrollmentApiClient()
+
+        revocation_results = {}
+        any_failures = False
         for course_overview in get_course_overviews(list(enrollments_by_course_id.keys())):
             enterprise_enrollment = enrollments_by_course_id.get(course_overview.get('id'))
+            course_id = str(course_overview.get('id'))
```
adamstankiewicz, line 449: nit: if you swap line 448 and 449, you could use `course_id` variable in `enrollments_by_course_id.get(course_overview.get('id'))`

enterprise/api/v1/views.py
```diff
@@ -406,14 +441,21 @@ def license_revoke(self, request, *args, **kwargs):
...
                 self._revoke_enrollment(enrollment_api_client, enterprise_enrollment, course_overview)
+                revocation_results[course_id] = 'success'
             except EnrollmentModificationException as exc:
-                return Response(str(exc), status=status.HTTP_500_INTERNAL_SERVER_ERROR)
+                revocation_results[course_id] = str(exc)
+                any_failures = True
 
-        return Response(status=status.HTTP_204_NO_CONTENT)
+        status_code = status.HTTP_200_OK if not any_failures else status.HTTP_422_UNPROCESSABLE_ENTITY
+        return Response(revocation_results, status=status_code)
```
adamstankiewicz, line 458: Overall, I think this approach makes sense and gives all the information about whether a course enrollment was properly modified. I wonder if it's worth surfacing whether a course enrollment was moved to audit vs. unenrolled in this response, beyond a blanket "success"?

enterprise/api/v1/views.py
```diff
@@ -357,43 +356,79 @@ def _revoke_enrollment(self, enrollment_api_client, enterprise_enrollment, cours
...
+            except Exception as exc:  # pylint: disable=broad-except
+                msg = (
+                    'License revocation: unable to update LMS enrollment for User {user} and Enterprise {enterprise} '
+                    'in Course {course_id} to Course Mode {mode} because: {reason}'.format(
+                        reason=str(exc),
+                        **log_message_kwargs
+                    )
+                )
+                LOGGER.error('{msg}: {exc}'.format(msg=msg, exc=exc))
+                raise EnrollmentModificationException(msg)
```
adamstankiewicz, line 387: Per the ticket AC:

> IF an operation that tries to drop learners to audit track of a course fails (under any circumstances), unenroll the learner from that course instead

If we fail on updating the enrollment to audit here for whatever reason, is the requested business logic that we should force unenroll at this point? Seems a bit aggressive, but figured I'd bring up this minor discrepancy with the ticket AC.

bbaker6225, line 387: I agree with Adam's comment here. To add on, I wonder if it's worth not unenrolling immediately on _any_ exception. It might be worth handling 429s more gracefully, and maybe some other things too?
Anyway, I don't know how nuanced we care to be with this one already somewhat aggressive end state, so feel free to ignore this

enterprise/api/v1/views.py
iloveagent57: This is only called from the `license_revoke` action method in this viewset.  The first thing it does is make sure that `get_course_overviews, get_certificate_for_user, CourseMode` are all truthy, and errors out otherwise.

enterprise/api/v1/views.py
iloveagent57: Ah, yeah, I did not strictly adhere to this AC, which I agree is aggressive. I'll raise it in group chat.

enterprise/api/v1/views.py
iloveagent57: Yeah, that's a good idea.  I'm trying that out now, so the response data might look like:
```
{'success': true, 'message': 'moved to audit'}
```
or
```
{'success': true, 'message': 'unenrolled'}
```
or
```
{'success': false, 'message': 'Could not update to audit track because: murder hornets.'}
```
What do you think?

adamstankiewicz: This response structure makes sense to me, and would be easily parseable by a client! If only "murder hornets" were a real reason for a failure here haha 😉

enterprise/api/v1/views.py
```diff
@@ -273,6 +275,15 @@ class LicensedEnterpriseCourseEnrollmentViewSet(EnterpriseWrapperApiViewSet):
     queryset = models.LicensedEnterpriseCourseEnrollment.objects.all()
     serializer_class = serializers.LicensedEnterpriseCourseEnrollmentReadOnlySerializer
 
+    class RevocationStatus:
```
kellmarie, line 278: Should this be in here? Would it make sense for it to have its own file/be in a constants file?

enterprise/api/v1/views.py
kellmarie: Did you raise it in group chat? I don't remember but it's also been a long weekend so last week was years ago.

chavesj (approved):
LGTM as long as Product is okay with not revoking on failure of `update_course_enrollment_mode_for_user` (this implementation seems right to me)

enterprise/models.py
```diff
@@ -1481,7 +1481,11 @@ def enrollments_for_user(cls, enterprise_customer_user):
         """
         return cls.objects.filter(
             enterprise_course_enrollment__enterprise_customer_user=enterprise_customer_user
-        ).select_related('enterprise_course_enrollment')
+        ).select_related(
+            'enterprise_course_enrollment',
+            'enterprise_course_enrollment__enterprise_customer_user',
+            'enterprise_course_enrollment__enterprise_customer_user__enterprise_customer',
```
kellmarie, line 1487: Update the doc comment to include these. Also do we need to return these in this function?

enterprise/api/v1/views.py
iloveagent57: This is a set of constants that's only relevant to this ViewSet, with a namespace.  I don't think it makes sense to extract it, given that no other ViewSet should use it.

enterprise/models.py
iloveagent57: Ah, yeah, we need those because this thing's caller looks at each of those associated model instances, so better to get it out of the way with a single query.

PR #995, opened 2020-10-08
ENT-3460: add default branding config object to the Customer record if null

chavesj:
[ENT-3460](https://openedx.atlassian.net/browse/ENT-3460)

**Description**
- Added `branding_configuration` property to `EnterpriseCustomer` to return default config object if none exists
- Added `logo` property to `EnterpriseCustomerBrandingConfiguration` to return the platform logo URL if none exists

**Preview**
image

**Testing Instructions**
- Pull branch, install enterprise into platform, run migrations, ensure the learner portal is running
- Create a new Enterprise Customer from the LMS django admin with no branding configuration
- Use the UUID for the new customer to hit the customer detail endpoint e.g. http://localhost:18000/enterprise/api/v1/enterprise-customer/[uuid]
- Observe that a branding configuration is returned in the customer record despite one not existing for the customer.
- Use the slug for the customer you created to attempt logging into the learner portal (you may need to clear your cookies for this to work properly). E.g. http://localhost:8734/[slug]
- Observe the branded login screen for your new customer and how the platform logo is displayed.
- Link a test learner with the new customer and login.
- Observe the logo successfully displayed in the learner portal header and the edX branding colors in the learner portal.
- Open an LMS shell and run `./manage.py lms shell` to initialize django and enter a python shell
- Create an instance of the `EnterpriseCustomer` model with no arguments to create a customer without a branding configuration.
- Run `print(customer.branding_configuration)` and observe the default branding configuration object printed rather than an error message for `RelatedObjectDoesNotExist`

**Merge checklist:**

**Post merge:**


enterprise/admin/__init__.py
```diff
@@ -164,7 +164,6 @@ class EnterpriseCustomerAdmin(DjangoObjectActions, SimpleHistoryAdmin):
         'site',
         'country',
         'active',
-        'has_logo',
```
chavesj, line 167: Removing has_logo since customer records will return the platform logo URL if no branding exists

enterprise/api/v1/serializers.py
```diff
@@ -94,8 +94,15 @@ class Meta:
             'tertiary_color',
         )
 
+    enterprise_customer = serializers.SerializerMethodField()
```
chavesj, line 97: This had to be added to prevent `TypeError: Object of type 'UUID' is not JSON serializable`

CHANGELOG.rst
```diff
@@ -15,6 +15,11 @@ Change Log
 
 Unreleased
 
+[3.8.33]
```
iloveagent57, line 18: Needs to be 3.8.34 now

enterprise/models.py
```diff
@@ -395,6 +401,26 @@ def serialized(self):
...
+        Return the customer branding configuration or a default object if none exists.
+        """
+        try:
+            return self._branding_configuration
+        except EnterpriseCustomerBrandingConfiguration.DoesNotExist:
+            default_branding = EnterpriseCustomerBrandingConfiguration(
+                enterprise_customer=self,
+                primary_color='#1a337b',
+                secondary_color='#d7e3fc',
+                tertiary_color='#007d88',
```
iloveagent57, line 416: Nit: maybe store these in variables and leave a comment about where you got them from?

iloveagent57 (commented):
One small nit and a version bump, but looks good!

test_utils/fake_enterprise_api.py
```diff
@@ -171,3 +171,17 @@ def build_fake_enterprise_catalog_detail(enterprise_catalog_uuid=FAKE_UUIDS[1],
...
+    """
+    Return a fake EnterpriseCustomerBrandingConfiguration object
+    """
+    return {
+        'enterprise_customer': customer_uuid,
+        'enterprise_slug': customer_slug,
+        'logo': 'http://fake.url',
+        'primary_color': '#1a337b',
+        'secondary_color': '#d7e3fc',
+        'tertiary_color': '#007d88',
```
adamstankiewicz, line 186: similar comment as Alex's above - should these move to constants?

iloveagent57 (approved):

enterprise/models.py
```diff
@@ -395,6 +401,26 @@ def serialized(self):
...
+                enterprise_customer=self,
+                primary_color=DefaultColors.PRIMARY,
+                secondary_color=DefaultColors.SECONDARY,
+                tertiary_color=DefaultColors.TERTIARY,
+            )
+            return default_branding
+
+    @branding_configuration.setter
+    def branding_configuration(self, obj):
+        self._branding_configuration = obj
```
adamstankiewicz, line 422: Do we need a `@branding_configuration.deleter` defined as well? Or, are we intentionally OK with not handling that here? I'm curious as using the `@property` decorator with `.setter` and `.deleter` is a new concept to me.

chavesj: I just tested this and deleting a customer with an existing branding config (verified in the DB) happens successfully as a cascading deletion. We don't delete the branding config using the customer model anywhere in the code so a `.deleter` shouldn't be needed

adamstankiewicz (approved):

PR #996, opened 2020-10-08
ENT-3473: Transmitter for blackboard, sends courseId as a hash of course Key

binodpant:
**Merge checklist:**

**Post merge:**


integrated_channels/blackboard/client.py
```diff
@@ -37,7 +41,34 @@ def __init__(self, enterprise_configuration):
         self.expires_at = None
 
     def create_content_metadata(self, serialized_data):
-        """TODO"""
+        """
+        Create a course from serialized course metadata
+        """
+        if not serialized_data:
+            return 204, "No courses created, empty content_metadata_items"
```
taliaedX, line 48: Wouldn't this normally be a noop anyways? IE you should never even call this method if you don't have data?

integrated_channels/blackboard/exporters/content_metadata.py
```diff
@@ -34,3 +34,10 @@ def transform_enrollment_url(self, content_metadata_item):
         enrollment_url = content_metadata_item.get('enrollment_url', None)
         url_link = self.DESCRIPTION_TEXT_TEMPLATE.format(enrollment_url=enrollment_url)
         return url_link
+
+    def transform_course_unique_key(self, content_metadata_item):
+        """
+        Blackboard courseId cannot contain some chars used in edX courseIds, so using hash
+        We can always look up courses by externalId whch is the same as edX courseId
+        """
+        return content_metadata_item.get('key').replace('+','').replace(':','').replace('/','')
```
taliaedX, line 43: I feel like there's gotta be a simpler way then 3 .replace

integrated_channels/blackboard/transmitters/content_metadata.py
```diff
@@ -19,3 +19,9 @@ def __init__(self, enterprise_configuration, client=BlackboardAPIClient):
             enterprise_configuration=enterprise_configuration,
             client=client
         )
+
+    def _prepare_items_for_transmission(self, channel_metadata_items):
+        # here is a hack right now to send only one item
+        # we have to investigate how to handle multiple
+        # metadata items since there is no batch course create endpoint in blackboard
+        return channel_metadata_items[0]
```
taliaedX, line 27: but what about chunk_size default of 1?

taliaedX (commented):
some questions

integrated_channels/blackboard/exporters/content_metadata.py
binodpant: replaced with client cdoe of hash instead...

integrated_channels/blackboard/transmitters/content_metadata.py
binodpant: ah yeah need to do that too thanks for the reminder...

integrated_channels/blackboard/client.py
binodpant: yeah I think we can scrap this one and assume we will get valid data otherwise we will fail later anyway

integrated_channels/blackboard/client.py
```diff
@@ -37,7 +42,43 @@ def __init__(self, enterprise_configuration):
...
+
+        serialized_channel_metadata = json.dumps(copy_of_channel_metadata)
+
+        LOGGER.info("Transmitting create_content_metadata with course_id: %s, %s",
+                    course_id_generated, serialized_channel_metadata)
+        self._create_session()
+        create_course_url = "{}/learn/api/public/v3/courses".format(
+            self.enterprise_configuration.blackboard_base_url
+        )
+        status_code, response_text = self._post(create_course_url, serialized_channel_metadata)
```
alex-sheehan-edx, line 68: A good practice that I like from Talia's work is wrapping the requests/error handling (checking status code). I added helper methods for posts, gets and patches in my PR if you want to use those!

integrated_channels/blackboard/client.py
```diff
@@ -86,14 +127,14 @@ def _get_oauth_access_token(self):
         if not self.enterprise_configuration.refresh_token:
             raise ClientError(
                 "Failed to generate oauth access token: Refresh token required.",
-                HTTPStatus.INTERNAL_SERVER_ERROR
+                HTTPStatus.INTERNAL_SERVER_ERROR.value
```
alex-sheehan-edx, line 130: 👍

integrated_channels/blackboard/exporters/content_metadata.py
```diff
@@ -12,17 +12,13 @@
 class BlackboardContentMetadataExporter(ContentMetadataExporter):
     """
         Blackboard implementation of ContentMetadataExporter.
-        The odd choice of making enrollment_url into description with a link is best explained as:
-            Blackboard does not show description by default on home page.
-            However, description shows on the course search page.
-            This seems to be the only place to visibly place a link for now for learners to
-            find and visit the edX course page. But this can change as we explore blackboard more.
+        Note: courseId is not being exported here (instead done in client), because it will be
```
alex-sheehan-edx, line 15: perhaps an explanation as to why the courseId is random/why it's ok that the course id is random?

integrated_channels/integrated_channel/management/commands/__init__.py
```diff
@@ -19,11 +20,12 @@
 INTEGRATED_CHANNEL_CHOICES = OrderedDict([
     (integrated_channel_class.channel_code(), integrated_channel_class)
     for integrated_channel_class in (
-        SAPSuccessFactorsEnterpriseCustomerConfiguration,
-        DegreedEnterpriseCustomerConfiguration,
-        CornerstoneEnterpriseCustomerConfiguration,
+        BlackboardEnterpriseCustomerConfiguration,
```
alex-sheehan-edx, line 23: 🎉 alphabetical 🎉

alex-sheehan-edx (commented):
Overall, looks pretty good! Just a couple q's and we should definitely work together while merging.

integrated_channels/blackboard/exporters/content_metadata.py
binodpant: yup I think the hash function works as designed, seeds at python subsystem startup time so we can't use it as is if we want consistent hash values ...but I can see this comment being a bit confusing here (why even mention that issue here, since there is no code here to generate courseId)

integrated_channels/blackboard/client.py
binodpant: hmm, not clear on how to do that, where are your helpers? I only have helpers here for oauth and auth header creation

binodpant: I think I see yours now https://github.com/edx/edx-enterprise/pull/998/files#diff-e8224b0f5aaaee88283c12d5d533f5700f33f2e63c7939e93f814ef94bf95716R122 yeah I guess we could just have error handling inside the _put/_post methods. Typically I don't consider it a great idea to have automatic error handling in lower level methods since 400 is an issue to be dealt with by whoever calls _put/_post etc while the lower level put just does the putting.

Also if we have error handling in _put/_post we don't have the ability to customize error messages  (unless we catch and rethrow?)

Somehow I will find a way to leverage your code once it merges while also being able to throw exceptions with custom messages

binodpant: also noting thatt your code does return the response_text along with the code so it's actually going to be quite good I thin even as is and I can try not even having any error handling in my code initially then see if I need to tune messages

binodpant:
Sorry for lots of small commits. Let's hold on reviewing this until I handle update and possibly delete

but otherwise other code/tests are in place

will hopefully get to all this by tomorrow

alex-sheehan-edx:
> Sorry for lots of small commits. Let's hold on reviewing this until I handle update and possibly delete
> 
> but otherwise other code/tests are in place
> 
> will hopefully get to all this by tomorrow

What you have now LGTM. Will review quickly if you push the changes for update and such.

integrated_channels/blackboard/client.py
```diff
@@ -40,16 +45,190 @@ def __init__(self, enterprise_configuration):
...
+        channel_metadata_item = json.loads(serialized_data.decode('utf-8'))
+        if 'externalId' not in channel_metadata_item:
+            raise ClientError("No externalId found in metadata, please check json data format", 400)
+
+        external_id = channel_metadata_item['externalId']
+
+        # blackboard does not support all characters in our courseIds so let's gen a hash instead
+        course_id_generated = abs(hash(external_id))
+
+        copy_of_channel_metadata = copy.deepcopy(channel_metadata_item)
```
taliaedX, line 65: why perform a copy if you're just appending a key that wasn't in there in the first place? (I assume 'courseId' is otherwise not populated)

binodpant: yeah I was just being a good citizen wrt not modifying input vars.. not sure if we utilize this pattern much around our codebase

taliaedX: Hmm yeah that's very important in React, but I am not so sure about Python. I'm not 100% up on the best practices though.

I will say that copying variables is definitely not the best practice when dealing with large amounts of data, so it's something to keep in mind. I don't think the data we're dealing with here is particularly large, but I also don't know where this runs precisely and how low of memory limits we may be dealing with.

For instance, in a previous project at a previous company, we were doing a lot of data transformations on csv files, some of which were several gbs in size. Originally because of the different transformations stacked on top of each other, we were copying the original data variable 4+ times. We crashed instances many, many times. These instances had 32+ gb of memory. 😅 

So TL;DR: I don't know if it is normally a best practice or not, but for performance reasons, I'm generally wary of it, but we also probably deal with so little data that it's not a big deal here.

binodpant: yeah totally great points, we have to be watchful of performance when doing this for sure. in this case yeah ti seem innocuous enough to try at least, thanks

integrated_channels/blackboard/client.py
```diff
@@ -40,15 +45,235 @@ def __init__(self, enterprise_configuration):
...
+        ---------------
+            '{
+                courseID: course-edx+555+3T2020,
+                score: 0.85,
+                completedTimestamp: 1602265162589,
+            }'
+
+        """
+        self._create_session()
+        learner_data = json.loads(payload)
```
binodpant, line 124: renamed a couple vars in this function, fyi, for clarity

tests/test_integrated_channels/test_blackboard/test_client.py
```diff
@@ -77,25 +83,93 @@ def test_course_completion_with_no_course(self):
...
+        client = self._create_new_mock_client()
 
         # Mock the course ID request
-        client._get_courses_by_internal_id = unittest.mock.MagicMock(  # pylint: disable=protected-access
-            name="_get_courses_by_internal_id",
-            return_value=[{
-                'externalId': self.course_id,
-                'id': self.blackboard_course_id
-            }]
+        client._resolve_blackboard_course_id = unittest.mock.MagicMock(  # pylint: disable=protected-access
```
binodpant, line 170: refactored client a bit so we enclose the fetching of blackboard course id into one function, hence this change

integrated_channels/blackboard/client.py
```diff
@@ -14,13 +15,17 @@
...
+
+LOGGER = logging.getLogger(__name__)
 
 # TODO: Refactor candidate (duplication with canvas client)
 GRADEBOOK_PATH = '/learn/api/public/v1/courses/{course_id}/gradebook/columns'
 ENROLLMENT_PATH = '/learn/api/public/v1/courses/{course_id}/users'
 COURSE_PATH = '/learn/api/public/v1/courses'
 POST_GRADE_COLUMN_PATH = '/learn/api/public/v2/courses/{course_id}/gradebook/columns'
 POST_GRADE_PATH = '/learn/api/public/v2/courses/{course_id}/gradebook/columns/{column_id}/users/{user_id}'
+COURSE_V3_PATH = '/learn/api/public/v3/courses/{course_id}'
```
taliaedX, line 28: I don't question the accuracy, this is just a question for my understanding: Why are we using v1-v3 versions? Do they really not have the functionality all in a "v3"?

integrated_channels/blackboard/client.py
```diff
@@ -40,15 +45,235 @@ def __init__(self, enterprise_configuration):
         self.config = apps.get_app_config('blackboard')
         self.session = None
         self.expires_at = None
+        self.create_course_url = "{}/learn/api/public/v3/courses".format(
```
taliaedX, line 48: why not use the `COURSE_PATH` var here (and admittedly update it to be v3 instead of v1)?

integrated_channels/blackboard/models.py
```diff
@@ -63,6 +63,15 @@ class BlackboardEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfigur
                   "used to re-request the access tokens over multiple client sessions."
     )
 
+    # overriding base model field, to use chunk size 1 default
+    transmission_chunk_size = models.IntegerField(
+        default=1,
```
taliaedX, line 68: 👍

taliaedX (approved):
Looking good to me! A question or 2 around API versions we're accessing, but really no issues here.

integrated_channels/blackboard/client.py
binodpant: ya they are transitioning slowly it seems

integrated_channels/blackboard/client.py
binodpant: and I have a bug here! should have tested locally, my url is not complete ... will fix and be back

binodpant: and I can't use the existing vars due to the way they are setup..

taliaedX: so they really don't have their get all courses endpoint on v3, but they do have their get/delete individual course on v3?

integrated_channels/blackboard/client.py
alex-sheehan-edx: docs are a mix of versions. Seems like certain areas are farther developed than others. Unfortunately, don't think we have much options here.

integrated_channels/blackboard/client.py
alex-sheehan-edx: why isn't this a constant with the other paths?

integrated_channels/blackboard/client.py
alex-sheehan-edx: that's fine with me :)

integrated_channels/blackboard/client.py
```diff
@@ -40,15 +45,235 @@ def __init__(self, enterprise_configuration):
...
+        )
 
     def create_content_metadata(self, serialized_data):
-        """TODO"""
+        """
+        Create a course from serialized course metadata
+        Returns: (int, str) Status code, Status message
+        """
+        channel_metadata_item = json.loads(serialized_data.decode('utf-8'))
+        if 'externalId' not in channel_metadata_item:
```
alex-sheehan-edx, line 58: Don't you have a static method for this?

integrated_channels/blackboard/client.py
```diff
@@ -40,15 +45,235 @@ def __init__(self, enterprise_configuration):
...
+    def _resolve_blackboard_course_id(self, external_id):
+        """
+        Extract course id from blackboard, given it's externalId
+        """
+        params = {'externalId': external_id}
+        courses_responses = self._get(self.generate_courses_url(), params).json()
+        course_response = courses_responses.get('results')
+
+        for course in course_response:
+            if course.get('externalId') == external_id:
```
alex-sheehan-edx, line 190: if you're searching by externalId on Blackboards side, will this ever not be true?

alex-sheehan-edx: Besides a 404 that is.

integrated_channels/blackboard/client.py
```diff
@@ -40,15 +45,235 @@ def __init__(self, enterprise_configuration):
...
+            self.enterprise_configuration.save()
+            return data['access_token'], data["expires_in"]
+        except (KeyError, ValueError):
+            raise ClientError(auth_response.text, auth_response.status_code)
+
+    def _create_auth_header(self):
+        """
+        auth header in oauth2 token format as required by blackboard doc
+        """
+        if not self.enterprise_configuration.client_id:
```
alex-sheehan-edx, line 261: I have mixed feelings about creating static methods to verify variable existence vs just checking inline, but I do feel like we should be consistent.

alex-sheehan-edx: How do you feel about either doing something like https://github.com/edx/edx-enterprise/pull/996/files#diff-e8224b0f5aaaee88283c12d5d533f5700f33f2e63c7939e93f814ef94bf95716R163? Or if not, then I don't see why we should have those static methods.

alex-sheehan-edx (commented):
A couple of questions and comments but should be good afterwards! 👍

integrated_channels/blackboard/client.py
binodpant: yeah looks like that one is either way for me, it's a style thing. they are more accurately validator methods I think. stuff like does externalId exist? does course_id valid? and since they are used in more than once place I figured it would be better to DRY it up

I thikn if we called these validate_xyz instead of raise_for_ then it starts to make a bit more sense, what'cha think?

binodpant: the reason I think a method makes sense btw, is that it's very possible we want to validate more things about the payload in the future, so just thought would be a nice convenient place to put them (also makes it easier for us to tune the exception message in the future or do logging in one place, if we want to do those things for those cases)

integrated_channels/blackboard/client.py
binodpant: good point, more precisely really we could even just get the first item that is returned and use it without checking. I actually need to test this manually once, I __think__ it will work but I really have to be sure.. be back in a bit with this answer

integrated_channels/blackboard/client.py
binodpant: yes thank you!

integrated_channels/blackboard/client.py
alex-sheehan-edx: My edge case questions would be:
- what's the response with no courses present (will we even have to worry about 404's with the error handling?)
- what's the  response with other courses but not the one we're looking for (will we even have to worry about 404's with the error handling?)
- what's the successful response body when we find just our course and just our course exists (I've tested this, it's not really an edge case)
- what's the successful response body when there are multiple courses, including our's?

If those are consistent, then I think we could just go without checking!

alex-sheehan-edx (approved):
Last bit of response on the comments, but otherwise LGTM once tests pass!

integrated_channels/blackboard/client.py
binodpant: yeah I would think that for if any reason we can't find a course_id here we can bail out safely

binodpant: and tested this code just now works!

PR #998, opened 2020-10-09
[FEAT] Adding learner data transmission capabilities to the Blackboard IC

alex-sheehan-edx:
Pulled from my work on Moodle and Canvas transmitter and exporters. 

What's necessary to get this to work:
   - Follow steps to generate an enterprise Bb config
   - Transmit the content metadata such that the course appears on Blackboard instance.
   - Enroll learner in the Blackboard course (with the same email as their edX enterprise learner account)
   - Complete the edX course such that a course grade is reported to the grade's api.
   - Run the management command `./manage.py lms transmit_learner_data`

Behavior decisions:
   - The flow for the IC BB client is as such:
      * Create the session and generate the necessary auth creds. 
      * Retrieve course ID from requesting all courses by filtering by `externalID`
      * Retrieve the learner's BB ID by requesting all enrollments 
      * Retrieve or create the BB edX integration grade column from the course
      * Using the exporter's payload, request a PATCH to update the grade column with the user's reported course grade

Design decisions:
   - The BB integration client's methods are ordered in the following format (in order):
      * Methods still do to/being worked on in parallel
      * url formatting helper methods
      * Request wrapper methods (`GET`, `POST`, `PATCH`)
      * Completion Reporting helper methods- intermittent steps of the completion reporting process
      * `create_course_completion()`
      * Methods relating to authorization and the client session

Still todo:
   - Unit tests

**Merge checklist:**

**Post merge:**


integrated_channels/blackboard/models.py
```diff
@@ -88,3 +94,93 @@ def get_content_metadata_exporter(self, user):
...
+class BlackboardLearnerDataTransmissionAudit(models.Model):
+    """
+    The payload we send to Blackboard at a given point in time for an enterprise course enrollment.
+
+    """
+    blackboard_user_email = models.EmailField(
+        max_length=255,
+        blank=False,
+        null=False,
+        help_text='The learner`s Blackboard email. This must match the email on edX'
```
binodpant, line 115: Love the comment! maybe add a suffix like ' must match the email on edX in order to ...' ? technically blackboard will take any email but the issue will be during SSO back to edX... not sure if you want to say all that here lol

integrated_channels/blackboard/models.py
```diff
@@ -88,3 +94,93 @@ def get_content_metadata_exporter(self, user):
...
+    )
+
+    enterprise_course_enrollment_id = models.PositiveIntegerField(blank=False, null=False, db_index=True)
```
binodpant, line 118: this is a total nit but just for easier field browsing, lay them out alphabetically here?

integrated_channels/blackboard/models.py
```diff
@@ -88,3 +94,93 @@ def get_content_metadata_exporter(self, user):
...
+        """
+        return self.__str__()
+
+    def serialize(self, *args, **kwargs):  # pylint: disable=unused-argument
+        """
+        Return a JSON-serialized representation.
+
+        Sort the keys so the result is consistent and testable.
+
+        # TODO: When we refactor to use a serialization flow consistent with how course metadata
```
binodpant, line 170: interesting, curious to learn more here but okay if this works

integrated_channels/blackboard/exporters/learner_data.py
```diff
@@ -0,0 +1,74 @@
...
+        )
+
+        course_catalog_client = get_course_catalog_api_service_client(
+            site=enterprise_customer.enterprise_customer.site
+        )
+
+        percent_grade = kwargs.get('grade_percent', None)
+
+        return [
+            BlackboardLearnerDataTransmissionAudit(
```
binodpant, line 58: is None for percent_grade acceptable here or an exceptional case? what happens if we apply the None value?

integrated_channels/blackboard/client.py
```diff
@@ -45,11 +51,232 @@ def update_content_metadata(self, serialized_data):
...
+        )
+
+    def _get_or_create_integrated_grade_column(self, course_id):
+        """
+        Helper method to search an edX integrated Blackboard course for the designated edX grade column.
+        If the column does not yet exist within the course, create it.
+
+        Parameters:
+        -----------
+            course_id (str): The Blackboard course ID in which to search for the edX final grade,
```
binodpant, line 174: since there may be both course_ids in play in our code, perhaps best to use prefix like bb_course_id or blackboard_course_id ?

integrated_channels/blackboard/client.py
```diff
@@ -35,6 +36,11 @@ def __init__(self, enterprise_configuration):
         self.config = apps.get_app_config('blackboard')
         self.session = None
         self.expires_at = None
+        self.gradebook_path = '/learn/api/public/v1/courses/{course_id}/gradebook/columns'
```
binodpant, line 39: since these are all constants, maybe make them defined constants up at the top in CAPS? upto you, more of a pref thing. It's possible it will make it easier to also use them in tests when we have to

integrated_channels/blackboard/exporters/learner_data.py
alex-sheehan-edx: It will actually just delete the user's entry in the edx final grade column.

integrated_channels/blackboard/client.py
binodpant: I had been thinking of doing this too

integrated_channels/blackboard/exporters/learner_data.py
alex-sheehan-edx: No exception raised. Therefore, if we get None here, I assume that the learner doesn't have a grade and we _should_ be removing any existing from Blackboard (None to None also is a noop)

integrated_channels/blackboard/client.py
```diff
@@ -45,11 +51,232 @@ def update_content_metadata(self, serialized_data):
...
+            base=self.enterprise_configuration.blackboard_base_url,
+            path=self.gradebook_path.format(course_id=course_id),
+        )
+
+    def generate_enrollment_url(self, course_id):
+        """
+        Blackboard API url helper method.
+        Path: Get course enrollments
+        """
+        # By including `expand=user` we get access to the user's contact info
```
binodpant, line 69: sweet! 🍯

integrated_channels/blackboard/client.py
alex-sheehan-edx: At the top or in a separate `integrated channels api paths` file?

integrated_channels/blackboard/models.py
alex-sheehan-edx: This was including from other channels before we started our integrations. Not sure if it's still in effect?

binodpant: yeah we will want to check this out

binodpant (approved):
ok to get in as long as tests pass and this builds okay, I will merge any client changes in my upcoming pr work for transmitter

binodpant:
and yeah reminder to do version and changelog

tests/test_integrated_channels/test_blackboard/test_client.py
```diff
@@ -20,16 +31,100 @@ class TestBlackboardApiClient(unittest.TestCase):
...
         self.enterprise_config = BlackboardEnterpriseCustomerConfigurationFactory(
             client_id="id",
             client_secret="secret",
             blackboard_base_url="https://base.url",
-            refresh_token="token",
+            refresh_token=self.token,
         )
+        self.user_email = 'testemail@example.com'
+        self.course_id = 'course-edx+{}'.format(str(random.randint(100, 999)))
+        self.black_board_course_id = random.randint(100, 999)
```
binodpant, line 43: black_board is weird,, otherwise lgtm! good tests

alex-sheehan-edx: oh true. Consistency is best!!

binodpant (approved):
les gow!

PR #1002, opened 2020-10-14
ENT-3522: Request for field to specify locale in Django enterprise accounts

saleem-latif:
**Jira Ticket:** [ENT-3522](https://openedx.atlassian.net/browse/ENT-3522)

__Description:__
1. Add a new column in enterprise customer config model, that will be used by the customers to change the learner's default language
2. Introduce a new middleware that will check If a linked enterprise learner’s enterprise has a non-null preferred language, then that learner will have their language cookie set to the preferred language (assuming they themselves have not set a language preference on the account settings page).
3.  There is already a middleware in `edx-platform/lang_pref` that can be used as a reference for the implementation https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/lang_pref/middleware.py#L52
4. All experiences on edX influenced by the edX language toggle  “English/Spanish” will be updated accordingly. 

**Merge checklist:**

**Post merge:**


enterprise/middleware.py
```diff
@@ -0,0 +1,84 @@
...
+        if hasattr(request, 'user'):
+            current_user = getattr(request.user, 'real_user', request.user)
+
+        if current_user and current_user.is_authenticated:
+            enterprise_customer = get_enterprise_customer_for_user(current_user)
+
+            if not (enterprise_customer and enterprise_customer.default_language):
+                # If user does not belong to an enterprise customer or the default language for the user's
+                # enterprise customer is not set, then no need to go any further.
+                return response
```
sameenfatima78, line 64: Are we handling the case when enterprise default language is not set but user's language preference exists. Wouldn't the expected behaviour in that case be to go with user's preference? I'm guessing this scenario is being handled in edx-platform's middleware that will run after this middleware, correct?

sameenfatima78 (approved):
Looks good... just one question

enterprise/middleware.py
saleem-latif: Yes, that is correct. That behavior is being handled by the middleware in the edx-platform.

enterprise/middleware.py
```diff
@@ -0,0 +1,84 @@
...
+            # Get the user's language preference
+            try:
+                user_pref = get_user_preference(current_user, LANGUAGE_KEY)
+            except (UserAPIRequestError, UserAPIInternalError):
+                # Ignore errors related to user preferences not found.
+                pass
+
+            # If user's language preference is not set and enterprise customer has a default language configured
+            # then set the default language as the learner's language
+            if not user_pref and not is_request_from_mobile_app(request) and enterprise_customer.default_language:
```
hasnain-naveed, line 75: why are we checking `enterprise_customer.default_language`  here again as you have a strict if condition above for this case? Am I missing something?

hasnain-naveed (commented):
One question why are we by-passing the mobile requests? I have seen the edx-platform middleware. Any idea just curious?

hasnain-naveed:
One more thing, according to your story, I guess you should only provide an English or Spanish option in the admin panel? This is more of a business decision. Just letting you know.

enterprise/middleware.py
saleem-latif: Done

hasnain-naveed (approved):

zamanafzal (approved):
💯

muhammad-ammar (approved):
LGTM 👍

PR #1005, opened 2020-10-15
fix saml strip to no longer strip extra characters.

taliaedX:
Using `strip(-saml)` did not work as intended. This would strip the prefix and part of the actual provider name/slug.
Since we always expect "-saml", I have just excluded the length of `saml-` (ie 5 characters). This will ensure it does not parse out valid parts of the slug as well.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/tpa_pipeline.py
+++ b/enterprise/tpa_pipeline.py
@@ -103,7 +103,10 @@ def get_user_from_social_auth(tpa_provider, user_id):
         user_id (str): User id of user in third party LMS
 
     """
-    provider_slug = tpa_provider.provider_id.strip('saml-')
+    # We attach the auth type to the slug at some point in this flow,
+    # so to match the original slug, we need to chop off that backend name.
+    # We only use saml here, so we are removing the first 5 characters, ie 'saml-'
+    provider_slug = tpa_provider.provider_id[5:]
     social_auth_uid = '{0}:{1}'.format(provider_slug, user_id)
     # we are filtering by both `provider` and `uid` to make use of provider,uid composite index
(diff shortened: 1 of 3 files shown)


enterprise/tpa_pipeline.py
```diff
@@ -103,7 +103,7 @@ def get_user_from_social_auth(tpa_provider, user_id):
         user_id (str): User id of user in third party LMS
 
     """
-    provider_slug = tpa_provider.provider_id.strip('saml-')
+    provider_slug = tpa_provider.provider_id[5:]
```
georgebabey, line 106: Given this makes the intent less obvious and feels like a magic number, can you put a comment indicating what the 5 characters we're chopping off are.

georgebabey (approved):
Looks good, small request for a comment and is there any quick test we could write that shows the breakage when using `.strip`

taliaedX:
>is there any quick test we could write that shows the breakage when using .strip

So yeah the current tests involving the get_user_from_auth method are all mocked/patches, thus changing this value doesn't seem to affect them at all.

I was hoping to write a test, but I'd like to have this do a real search as it were: ```    user_social_auth = UserSocialAuth.objects.select_related('user').filter(
        provider=tpa_provider.backend_name, uid=social_auth_uid
    ).first()``` however, that doesn't seem to be possible as in edx-enterprise, social_auth is not installed nor configured (we also do a try/except on the import, probably because this just lives in edx-platform).

Otherwise I'm not sure I can mock that object without linking it to the "original" UserSocialAuth object and without that I'd just be mocking the entire function.

PR #1009, opened 2020-10-19
ENT-3557: improve exception reporting for authfailure blackboard

binodpant:
WHY?

Since today's exception reporting actually fails during the raise line meaning we don't actually get to see what the `auth_response` details were. I am getting a failure in staging that does not reproduce locally for me and it's great to be able to see the status_text, response of the response object at least

also handles the two cases KeyError and ValueError differently since they are for different reasons and have different pieces of info available in except blocks

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/blackboard/views.py
+++ b/integrated_channels/blackboard/views.py
@@ -3,7 +3,6 @@ Views containing APIs for Blackboard integrated channel
 """
 
 import base64
-
 import requests
 from rest_framework import generics
 from rest_framework.exceptions import APIException, NotFound, ParseError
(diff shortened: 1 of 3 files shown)


integrated_channels/blackboard/views.py
```diff
@@ -109,8 +109,19 @@ def get(self, request, *args, **kwargs):  # pylint: disable=arguments-differ
...
             refresh_token = data['refresh_token']
-        except (KeyError, ValueError):
-            raise requests.RequestException(response=auth_response)
+        except KeyError as exception:
+            raise ParseError(
+                "BLACKBOARD: failed to find refresh_token in auth response. "
+                "Auth response text: {}, Response code: {}, JSON response: {}".format(
+                    auth_response.text,
+                    auth_response.status_code,
+                    data,
```
binodpant, line 118: just a note: even though this line will contain refresh token I think we are good security wise since that refresh token only is usable once and gets overwritten in future runs of client ...

integrated_channels/blackboard/views.py
```diff
@@ -109,8 +109,19 @@ def get(self, request, *args, **kwargs):  # pylint: disable=arguments-differ
...
+                )
+            ) from exception
+        except ValueError as exception:
+            raise ParseError(
+                "BLACKBOARD: auth response is invalid json. auth_response: {}".format(auth_response)
```
binodpant, line 123: ditto here on any security concerns like printing refresh token

alex-sheehan-edx (approved):
LGTM

integrated_channels/blackboard/views.py
binodpant: will remove logging of refresh token after investigation is over

PR #1010, opened 2020-10-20
add action to kick off refreshing catalogs

kellmarie:
**Merge checklist:**

**Post merge:**

All catalogs are refreshing:
Screen Shot 2020-10-20 at 11 43 09 AM

Catalogs that failed to refresh:
Screen Shot 2020-10-20 at 11 44 12 AM

Mixture of successfully refreshing and failing to refresh catalogs:
Screen Shot 2020-10-20 at 11 40 31 AM

### DIFF
--- a/enterprise/admin/actions.py
+++ b/enterprise/admin/actions.py
@@ -8,7 +8,11 @@ Custom `Django Admin actions`_ used in enterprise app.
 import unicodecsv
 from six import string_types
 
+from django.contrib import messages
 from django.http import HttpResponse
+from django.utils.safestring import mark_safe
+
+from enterprise.api_client.enterprise_catalog import EnterpriseCatalogApiClient
 
 
 def export_as_csv_action(description="Export selected objects as CSV file", fields=None, header=True):
@@ -59,3 +63,34 @@ def export_as_csv_action(description="Export selected objects as CSV file", fiel
 
     export_as_csv.short_description = description
     return export_as_csv
+
+
+def refresh_catalog(self, request, queryset):  # pylint: disable=unused-argument
+    """
+    Kicks off background running tasks for refreshing catalogs
+    """
+    catalog_client = EnterpriseCatalogApiClient(user=request.user)
+    refreshed_catalogs, failed_to_refresh_catalogs = catalog_client.refresh_catalogs(queryset)
+
+    # display catalog and task ids that were successfully started refreshing
+    updated_message = ''
+    if refreshed_catalogs.items():
+        updated_message = '<b>The following catalogs are being refreshed:</b><ul>'
+        for uuid, task_id in refreshed_catalogs.items():
+            updated_message = updated_message + '<li>' + str(uuid) + ' with task id: ' + str(task_id) + '</li>'
+        updated_message = updated_message + '</ul>'
+
+    # display catalog ids that failed to start refreshing
+    if failed_to_refresh_catalogs:
+        updated_message = updated_message + '<b>Failed to refresh catalogs with the following ids:</b><ul>'
+        for failed_catalog_uuid in failed_to_refresh_catalogs:
+            updated_message = updated_message + '<li class="error">' + str(failed_catalog_uuid) + '</li>'
+        updated_message = updated_message + '</ul>'
+
+    # Info level based on how successful refreshing catalogs was overall
+    level = messages.INFO
+    if failed_to_refresh_catalogs:
+        level = messages.ERROR
+        if refreshed_catalogs.items():
+            level = messages.WARNING
+    self.message_user(request, mark_safe(updated_message), level=level)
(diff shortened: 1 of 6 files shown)


enterprise/api_client/enterprise_catalog.py
```diff
@@ -169,6 +170,35 @@ def get_content_metadata(self, enterprise_customer, enterprise_catalogs=None):
 
         return list(content_metadata.values())
 
+    @JwtLmsApiClient.refresh_token
+    def refresh_catalogs(self, enterprise_catalogs):
```
bbaker6225, line 174: I don't think we should allow refreshing multiple catalogs in this way since it could unintentionally cause weird behavior if the catalogs share a catalog query. It's not the best UX, but I think it's reasonable to say that if you only want to update a few catalogs you just individually update them.
If we do want to support this however, I think we'd need to update the backend to take multiple catalogs and update the set of distinct catalog queries between them all

enterprise/admin/actions.py
```diff
@@ -59,3 +63,34 @@ def export_as_csv(modeladmin, request, queryset):  # pylint: disable=unused-argu
...
+
+def refresh_catalog(self, request, queryset):  # pylint: disable=unused-argument
+    """
+    Kicks off background running tasks for refreshing catalogs
+    """
+    catalog_client = EnterpriseCatalogApiClient(user=request.user)
+    refreshed_catalogs, failed_to_refresh_catalogs = catalog_client.refresh_catalogs(queryset)
+
+    # display catalog and task ids that were successfully started refreshing
+    updated_message = ''
```
bbaker6225, line 76: It probably simplifies a lot if we restrict the updates to a single catalog, but the string concatenation for updated_message is hard to follow and read for me. It's not perfectly DRY but might be cleaner to write out the admin message in larger chunks?

bbaker6225: I also think it could be helpful to throw the task id into the success message if we can. Doesn't matter for ECS but saves a little digging for any developer who winds up doing this and then wants to check flower/splunk to follow the progress

enterprise/api_client/enterprise_catalog.py
kellmarie: Well is there a way to get django actions to work so that what gets acted on isn't through a bulk selection? And the catalogs sharing a query isn't an issue for the daily jobs that run?

bbaker6225: Sure, just give an error message if `queryset.count() > 1`. And the daily job does the idea of distinct catalogs mentioned above: https://github.com/edx/enterprise-catalog/blob/4fe13a5b20912422b5463edc720bb78af32e9862/enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py#L56

enterprise/admin/actions.py
kellmarie: Task id is in the success message. I posted screenshots of what the messages look like in the PR descriptions.

bbaker6225: Ah nice, I missed it at the bottom. Ignore that comment then

enterprise/api_client/enterprise_catalog.py
kellmarie: Actually how does refreshing the multiple catalogs individually by selecting them one by one address the shared catalogs weirdness issue? If I were to individually click on catalog A and refresh it and then click on catalog B and refresh it one after the other would the window of time it would take me to do the clicking be large enough that we won't have to worry about the weirdness happening like we do for making multiple api calls in a loop?

bbaker6225: It would still happen. I don't like either way so I don't think it _really_ matters which way we go. It just feels weirder to me to allow multiple at once when we don't really support that, especially when this seems like it would only ever be used for a single catalog at a time by ECS/developers

kellmarie: I think it would be weirder to have a UI that is designed to be used for bulk selecting but we would need to somehow communicate with people to only select one catalog at a time. But even if we do communicate that we don't really have a way to say that what they should actually be doing is waiting for the previously refreshed catalog task to finish refreshing before selecting another. Because telling them to refresh them individually wouldn't actually solve the issue about worrying out data weirdness happening.

Maybe we should just change the refresh endpoint to allow for multiple catalogs so we can do what the daily job does and only refresh the distinct queries? That way we can try to avoid any data weirdness and also not have to try communicating on the front end to select things individually and then wait for the task to finish even though they might not actually know where to look to know if the task had finished.

iloveagent57: My two cents:
* It's an internal tool, and my _guess_ is that it will mostly be used to update only a single catalog at a time.  We can communicate to the users of this tool that they're probably better off only updating one at a time.
* I think it's reasonable for this client (or any other) to make a series of POSTs to `refresh_metadata` for a series of catalogs.
* We should modify the enterprise-catalog service to more gracefully deal with this type of situation, somehow ensuring that we don't bother updating the same piece of content metadata multiple times within a short time window.

iloveagent57 (approved):
Let's give it a whirl.

PR #1012, opened 2020-10-20
Adding blackboard course integration customization

alex-sheehan-edx:
**Merge checklist:**

**Post merge:**

This PR adds: 
- Refactored exported content metadata so that it contains both solo the enrollment url as well as the embedded html element. This is to ensure that the description in course discovery keeps it's `go to edX page` hyper link, as well as have the course content page title be the same embedded link. 
- The course content creation method now has the additional process of creating a content object associated with the newly made course, as well as posts the HTML body to the course content's child. 

Outstanding q's:
- How do we want the course content's child to be formatted? It's a simplified HTML markdown language so we are restricted but there is a reasonable amount of customization todo here.

- Is testing required?

### DIFF
--- a/integrated_channels/blackboard/client.py
+++ b/integrated_channels/blackboard/client.py
@@ -13,6 +13,7 @@ from six.moves.urllib.parse import urljoin  # pylint: disable=import-error
 
 from django.apps import apps
 
+from integrated_channels.blackboard.exporters.content_metadata import BLACKBOARD_COURSE_CONTENT_NAME
 from integrated_channels.exceptions import ClientError
 from integrated_channels.integrated_channel.client import IntegratedChannelApiClient
 from integrated_channels.utils import refresh_session_if_expired
@@ -27,6 +28,9 @@ POST_GRADE_COLUMN_PATH = '/learn/api/public/v2/courses/{course_id}/gradebook/col
 POST_GRADE_PATH = '/learn/api/public/v2/courses/{course_id}/gradebook/columns/{column_id}/users/{user_id}'
 COURSE_V3_PATH = '/learn/api/public/v3/courses/{course_id}'
 COURSES_V3_PATH = '/learn/api/public/v3/courses'
+COURSE_CONTENT_PATH = '/learn/api/public/v1/courses/{course_id}/contents'
+COURSE_CONTENT_CHILDREN_PATH = '/learn/api/public/v1/courses/{course_id}/contents/{content_id}/children'
+COURSE_CONTENT_DELETE_PATH = '/learn/api/public/v1/courses/{course_id}/contents/{content_id}'
 
 
 class BlackboardAPIClient(IntegratedChannelApiClient):
@@ -54,6 +58,7 @@ class BlackboardAPIClient(IntegratedChannelApiClient):
         """
         channel_metadata_item = json.loads(serialized_data.decode('utf-8'))
         BlackboardAPIClient._validate_channel_metadata(channel_metadata_item)
+        BlackboardAPIClient._validate_full_course_metadata(channel_metadata_item)
 
         external_id = channel_metadata_item.get('externalId')
 
@@ -61,28 +66,73 @@ class BlackboardAPIClient(IntegratedChannelApiClient):
         course_id_generated = self.generate_blackboard_course_id(external_id)
 
         copy_of_channel_metadata = copy.deepcopy(channel_metadata_item)
-        copy_of_channel_metadata['courseId'] = course_id_generated
+        copy_of_channel_metadata['course_metadata']['courseId'] = course_id_generated
 
         LOGGER.info("Creating course with courseId: %s", external_id)
         self._create_session()
         create_url = self.generate_course_create_url()
-        response = self._post(create_url, copy_of_channel_metadata)
-        return response.status_code, response.text
+        response = self._post(create_url, copy_of_channel_metadata['course_metadata'])
+
+        # We wrap error handling in the post, but sanity check for the ID
+        bb_course_id = response.json().get('id')
+        if not bb_course_id:
+            raise ClientError(
+                'Something went wrong while creating base course object on Blackboard. Could not retrieve course ID.',
+                HTTPStatus.NOT_FOUND.value
+            )
+
+        LOGGER.info("Creating content page for Blackboard course with course ID={}".format(external_id))
+        course_created_response = self.create_integration_content_for_course(bb_course_id, copy_of_channel_metadata)
+
+        success_body = 'Successfully created Blackboard integration course={bb_course_id} with integration ' \
+                       'content={bb_child_id}'.format(
+                           bb_course_id=bb_course_id,
+                           bb_child_id=course_created_response.json().get('id'),
+                       )
+
+        return course_created_response.status_code, success_body
 
     def update_content_metadata(self, serialized_data):
         """Apply changes to a course if applicable"""
         self._create_session()
-        channel_metadata_item = json.loads(serialized_data.decode("utf-8"))
+        channel_metadata_item = json.loads(serialized_data.decode('utf-8'))
 
         BlackboardAPIClient._validate_channel_metadata(channel_metadata_item)
         external_id = channel_metadata_item.get('externalId')
         course_id = self._resolve_blackboard_course_id(external_id)
         BlackboardAPIClient._validate_course_id(course_id, external_id)
 
-        LOGGER.info("Updating course with courseId: %s", course_id)
+        LOGGER.info('Updating course with courseId: {}'.format(course_id))
         update_url = self.generate_course_update_url(course_id)
-        response = self._patch(update_url, channel_metadata_item)
-        return response.status_code, response.text
+        response = self._patch(update_url, channel_metadata_item.get('course_metadata'))
+
+        bb_course_id = response.json().get('id')
+
+        if not bb_course_id:
+            raise ClientError(
+                'Unable to update course={} content on Blackboard. Failed to retrieve ID from course update response:'
+                '(status_code={}, body={})'.format(external_id, response.status_code, response.text)
+            )
+
+        course_contents_url = self.generate_create_course_content_url(bb_course_id)
+        course_contents_response = self._get(course_contents_url)
+
+        bb_contents = course_contents_response.json().get('results')
+        bb_content_id = None
+
+        for content in bb_contents:
+            if content.get('title') == BLACKBOARD_COURSE_CONTENT_NAME:
+                bb_content_id = content.get('id')
+
+        if not bb_content_id:
+            LOGGER.info('Blackboard integrated course content not found. Generating content.')
+        else:
+            course_content_delete_url = self.generate_course_content_delete_url(bb_course_id, bb_content_id)
+            self._delete(course_content_delete_url)
+
+        course_updated_response = self.create_integration_content_for_course(bb_course_id, channel_metadata_item)
+
+        return course_updated_response.status_code, course_updated_response.text
 
     def delete_content_metadata(self, serialized_data):
         """Delete course from blackboard (performs full delete as of now)"""
@@ -172,6 +222,21 @@ class BlackboardAPIClient(IntegratedChannelApiClient):
         if not course_id:
             raise ClientError(
                 'Could not find course:{} on Blackboard'.format(external_id),
+                HTTPStatus.NOT_FOUND.value
+            )
+
+    @staticmethod
+    def _validate_full_course_metadata(channel_metadata_items):
+        """
+        Raise error if course_metadata, course_content_metadata or course_child_content_metadata are invalid.
+        """
+        metadata_set = set(channel_metadata_items.keys())
+        course_metadata_groups = {'course_metadata', 'course_content_metadata', 'course_child_content_metadata'}
+        if not course_metadata_groups.issubset(metadata_set):
+            raise ClientError(
+                'Could not find course metadata group(s):{} necessary to create Blackboard integrated course.'.format(
+                    course_metadata_groups - metadata_set
+                ),
                 HTTPStatus.NOT_FOUND.value
             )
 
@@ -351,6 +416,36 @@ class BlackboardAPIClient(IntegratedChannelApiClient):
             ),
         )
 
+    def generate_create_course_content_url(self, course_id):
+        """
+        Blackboard API url helper method.
+        Path: Course's content objects
+        """
+        return '{base}{path}'.format(
+            base=self.enterprise_configuration.blackboard_base_url,
+            path=COURSE_CONTENT_PATH.format(course_id=course_id)
+        )
+
+    def generate_create_course_content_child_url(self, course_id, content_id):
+        """
+        Blackboard API url helper method.
+        Path: Course content's children objects
+        """
+        return '{base}{path}'.format(
+            base=self.enterprise_configuration.blackboard_base_url,
+            path=COURSE_CONTENT_CHILDREN_PATH.format(course_id=course_id, content_id=content_id)
+        )
+
+    def generate_course_content_delete_url(self, course_id, content_id):
+        """
+        Blackboard API url helper method.
+        Path: Course content deletion
+        """
+        return '{base}{path}'.format(
+            base=self.enterprise_configuration.blackboard_base_url,
+            path=COURSE_CONTENT_DELETE_PATH.format(course_id=course_id, content_id=content_id)
+        )
+
     def _get(self, url, data=None):
         """
         Returns request's get response and raises Client Errors if appropriate.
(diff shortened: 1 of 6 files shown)


integrated_channels/blackboard/client.py
```diff
@@ -351,6 +376,46 @@ def generate_post_users_grade_url(self, course_id, column_id, user_id):
...
+        Format the necessary json to create content data through Blackboard's API.
+        """
+        data = {
+            'title': title,
+            'availability': 'Yes',
+            'contentHandler': {
+                'id': 'resource/x-bb-externallink',
+                'url': link_to_enrollment
+            },
+            'body': '<div>{title}</div><div>{description}</div><img src={image_url} /></p>'.format(
```
binodpant, line 411: will this format properly in blackboard side with three divs next to each other? I figure they will handle all layout css etc?

integrated_channels/blackboard/client.py
```diff
@@ -464,3 +529,21 @@ def _get_or_create_integrated_grade_column(self, bb_course_id):
...
+            "contentHandler": {"id": "resource/x-bb-folder"}
+        }
+        content_url = self.generate_create_course_content_url(bb_course_id)
+
+        content_resp = self._post(content_url, course_content_data)
+        bb_content_id = content_resp.json().get('id')
+
+        content_child_url = self.generate_create_course_content_child_url(bb_course_id, bb_content_id)
+        self._post(content_child_url, course_content_child_data)
+        return bb_content_id,
```
binodpant, line 549: does the comma here indicate we want to return a tuple? is this necessary?

integrated_channels/blackboard/client.py
```diff
@@ -57,17 +59,40 @@ def create_content_metadata(self, serialized_data):
...
+        LOGGER.info("Creating content page for Blackboard course with course ID={}".format(external_id))
+        bb_content_id = self.create_integration_content_for_course(bb_course_id, course_content_child_data)
+
+        success_body = 'Successfully created Blackboard integration course={bb_course_id} with integration ' \
+                       'content={bb_content_id}'.format(
+                           bb_course_id=bb_course_id,
+                           bb_content_id=bb_content_id,
+                       )
+
+        return HTTPStatus.OK.value, success_body
```
taliaedX, line 95: why are you assuming success here? We aren't doing error handling here or in the create_integration_content_for_course method so right now if it gets a KeyError or ValueError or something because the value doesn't exist, it's just going to fail badly.

integrated_channels/blackboard/client.py
```diff
@@ -464,3 +529,21 @@ def _get_or_create_integrated_grade_column(self, bb_course_id):
...
+        """
+        course_content_data = {
+            'title': 'edX Integration',
+            'position': 0,
+            "contentHandler": {"id": "resource/x-bb-folder"}
+        }
+        content_url = self.generate_create_course_content_url(bb_course_id)
+
+        content_resp = self._post(content_url, course_content_data)
+        bb_content_id = content_resp.json().get('id')
```
taliaedX, line 545: This could fail with a KeyError if the id isn't present in the response.

integrated_channels/blackboard/client.py
binodpant: also isn't it nice to have the status code and text returned here as well maybe in addition to this info to keep our pattern of preserving that info through out the code?

integrated_channels/blackboard/client.py
binodpant, line 95: I think it may be better to return the actual response code here from the prior post? alternatively, you can just log that response code in the log above? just wondering if we can get codes that are not 200 but still success that will be good to know?

integrated_channels/blackboard/client.py
```diff
@@ -57,17 +59,40 @@ def create_content_metadata(self, serialized_data):
 
         external_id = channel_metadata_item.get('externalId')
 
+        # Retrieve and pop the data not needed for course creation and necessary for content customization
+        enrollment_link_metadata = channel_metadata_item.pop('enrollment_metadata')
```
binodpant, line 63: wrt error handling, do we still want to allow course creation if something happens in the process of adding this content (keyError or any other error?). Similarly, what what a failure once course is created, but the adding of this metadata/customizatoin fails? How do we want to handle that? I think since course was created, it should still be reported as successful (otherwise will integrated channels system try to create course again hi the future?)

let's maybe just think over this aspect a bit...

integrated_channels/blackboard/client.py
```diff
@@ -27,6 +27,8 @@
 POST_GRADE_PATH = '/learn/api/public/v2/courses/{course_id}/gradebook/columns/{column_id}/users/{user_id}'
 COURSE_V3_PATH = '/learn/api/public/v3/courses/{course_id}'
 COURSES_V3_PATH = '/learn/api/public/v3/courses'
+COURSE_CONTENT_PATH = '/learn/api/public/v1/courses/{course_id}/contents'
```
binodpant, line 30: not for this line per se, but assuming you plan to add tests for the new feature?

integrated_channels/blackboard/exporters/content_metadata.py
```diff
@@ -17,7 +17,9 @@ class BlackboardContentMetadataExporter(ContentMetadataExporter):
     DATA_TRANSFORM_MAPPING = {
         'name': 'title',
         'externalId': 'key',
-        'description': 'enrollment_url',
+        'enrollment_metadata': 'enrollment_url',
+        'syllabus_body': 'full_description',
+        'image_url': 'image_url',
```
taliaedX, line 22: Why not just have a `'content_body': 'content_body'` mapping and then write a `transform_content_body` method that contains all of the values you're using exclusively for the course content page/section here instead of doing a bunch of different popping and copying in the client?

ie the content transform would return: `{'course_content': {'element': 'url', 'link': 'other url', 'image_url': 'value', 'syllabus_body': 'description'}}` and then in the client you could just pop that whole dictionary off and use it.

Even further, you could just do everything in the `generate_integration_content_data` method here instead.

Idk how you guys feel about that, but since the exporter classes all seem to just be doing data transformations, I personally feel like we should just put as many data transformations as we can in the exporter, rather than in the client. There is some weird timing stuff (of the calls and transformations) to consider, but I think what you're adjusting should be fine to do at an exporter level.

taliaedX (commented):
I think there's some missing error handling and some stuff might be able to be consolidated a little bit/cleaned up, but once we have error handling I think this is good to go!

integrated_channels/blackboard/exporters/content_metadata.py
binodpant: Yeah, separation of concerns wise, I think it's best, when possible, to do all data work in exporter, so transmitter is just a dumb-ish sender. I found it not possible to avoid breaking this rule when I did the course_id work in Bb as you know, for reasons, but yeah in general I agree with this principle.

binodpant: also makes testing super easy, exporter tests cover all transformations, and transmitter tests don't need to be rigorous then... just a thought on that line too

taliaedX: Yeah I also had to do some transformations in Moodle because of its hideous array formatting. I don't expect there to be none there, but think we should _try_ to do them in the exporter and if we find we need something changed because the overall item format needs to be changed, then we can do that in the client.

integrated_channels/blackboard/client.py
alex-sheehan-edx: "Properly" is hard to define here for me. I've been trying to get Ken to help work up a design. This "displays" all the information, but in a way that I would argue isn't ideal. I'll try and improve upon the design and get Ken's feedback before this is merged.

integrated_channels/blackboard/client.py
alex-sheehan-edx: I thiiiiink I was originally returning status/body, thus the comma haha.

alex-sheehan-edx: There's a point here about logging, which is at the heart of why I made the switch originally, we're creating a content object and a child to that content object. Which ID do we log? Do we even need to log anything? I can follow our patterns and return the status code and body of the response right above, but then if we log the ID from the response, we're logging the content's child's ID, which could potentially lead to confusion while debugging. I'm flexible here, what do you think?

integrated_channels/blackboard/client.py
alex-sheehan-edx: Do you want to return the status code from the initial course creation? From the content creation? From the content child creation? I'm ok with taking one of those responses, but the reason why I did the generic `200` was because I felt like there was some ambiguity here. Are you ok with returning the success body?

integrated_channels/blackboard/client.py
alex-sheehan-edx: 😞  .... yes....

integrated_channels/blackboard/client.py
alex-sheehan-edx: Error handling is __mostly__ handled by the client's wrapped request methods (get, post, patch) which handle cases when BB returns anything other than a `200`. Since the majority of the calls here are posts (so 404's are _unlikely_), and I'm fairly certain BB doesn't pull a moodle and return 200's on failed requests, I think this is very unlikely to explode. Having said that, sanity checks don't hurt and who knows what edge cases may pop up in the future so I'm willing to add. 👍

integrated_channels/blackboard/client.py
alex-sheehan-edx: yes, but this is creating the content and if we don't get an ID back while also succeeding... I'd be really confused.

alex-sheehan-edx: However, happy to add the sanity check 👍

integrated_channels/blackboard/client.py
binodpant: got it, as long as it displays 'well' on Bb side this should be fine.. just wondering if they have any guidance on div classes etc .. but if no, then we are done for this round

integrated_channels/blackboard/client.py
binodpant: yeah as a simple rule if we keep that status text, code tuple to be always the return value, it will help with consistency as well as reporting. having said that, you can have more info returned than just code and body, like this case, so could be benefit from an output like (status_code, status_text, bb_content_id) ? as long as the method is clear about what it returns I thought having the two pieces of info always first, then followed by other needed info could be useful.

also, depending on how you want to handle failures for this response, it may make that work easier too if we keep track of the status and text from the post method above this line

integrated_channels/blackboard/client.py
binodpant: good question. I think that base IntegratedChannelApiClient makes _no_ comments or restrictions about what this method should return if anything, the output is never consumed anywhere. I think we introduced this notion in our code?

if we want the return values as is, I would say let's return the most recent call's status and body. If anything fails we return that status and body at the time of failure

however, we have the option to also not return anything from this method since no one seems to be consuming it and the base class does not require it

the only use of this method seems to be here

https://github.com/edx/edx-enterprise/blob/4b8faec47ae0aa74e5a3f941035e5f35e88a807e/integrated_channels/integrated_channel/transmitters/content_metadata.py#L127

I personally do not like method that return nothing. I usually at least return a boolean success (or in a fluent api return the object instance). But it is what it is right now

Does this work?

integrated_channels/blackboard/exporters/content_metadata.py
alex-sheehan-edx: Agreed with the exporter doing the bulk of the work instead of the client. It's difficult to do no work here because we're dealing with three separate posts to create objects on blackboard, all of which require their own formatting. Perhaps there's some consolidation that can be done here. Let me look.

integrated_channels/blackboard/client.py
taliaedX: Ah I think I missed the raise in the `_post` method somehow.

integrated_channels/blackboard/exporters/content_metadata.py
taliaedX: You may need that method `generate_integration_content_data` but from what I looked at (and maybe didn't look close enough), you could at the very least group all the data together like my first suggestion. We may not be able to consolidate much more than that and that's okay

alex-sheehan-edx: I've broken content metadata exporting for blackboard up into three sections: 
```
    DATA_TRANSFORM_MAPPING = {
        'course_metadata': 'course_metadata',
        'course_content_metadata': 'course_content_metadata',
        'course_child_content_metadata': 'course_child_content_metadata',
    }
```
Now, we perform 3 transformations that format each of these fields to the exact json that Blackboard is looking for each of these post operations. The only thing that needs to be formatted in the client now is the hashed unique BB id.

alex-sheehan-edx: That is, when making the posts to BB, all we have to do is `content_metadata.get('course_metadata')` or equivelent for each of the steps.

integrated_channels/blackboard/client.py
alex-sheehan-edx: > I think we introduced this notion in our code?

I think that I introduced this behavior with learner data because we record and store the returned body and status in order to figure out future transmission updates/deletes. Seems like it got transferred over into content metadata somewhere... 

I agree that returning nothing seems like a step in the wrong direction. Let me put up a change and we can see if that satisfies everyone 😄

integrated_channels/blackboard/exporters/content_metadata.py
```diff
@@ -15,17 +15,51 @@ class BlackboardContentMetadataExporter(ContentMetadataExporter):
...
+        """
+        title = content_metadata_item.get('title', None)
+        return {
+            'title': title,
+            'availability': 'Yes',
+            'contentHandler': {
+                'id': 'resource/x-bb-externallink',
+                'url': content_metadata_item.get('enrollment_url', None),
+            },
+            'body': '<div>{title}</div><div>{description}</div><img src={image_url} /></p>'.format(
```
binodpant, line 60: isn't the closing <p> tag superflous?

taliaedX (approved):
Oh this looks good to me. (assuming tests pass ofc)

integrated_channels/blackboard/client.py
```diff
@@ -175,6 +224,21 @@ def _validate_course_id(course_id, external_id):
                 HTTPStatus.NOT_FOUND.value
             )
 
+    @staticmethod
+    def _validate_full_course_metadata(channel_metadata_items):
```
binodpant, line 228: A nit but isn't this one item, `channel_metadata_item` ?

integrated_channels/blackboard/client.py
```diff
@@ -175,6 +224,21 @@ def _validate_course_id(course_id, external_id):
...
+        """
+        Raise error if course_metadata, course_content_metadata or course_child_content_metadata are invalid.
+        """
+        metadata_set = set(channel_metadata_items.keys())
+        course_metadata_groups = {'course_metadata', 'course_content_metadata', 'course_child_content_metadata'}
+        if not course_metadata_groups.issubset(metadata_set):
+            raise ClientError(
+                'Could not find course metadata group(s):{} necessary to create Blackboard integrated course.'.format(
+                    course_metadata_groups - metadata_set
```
binodpant, line 237: wait isn't metadata_set expected to be the larger set in general, is this subtraction reversed if the intent is to show the set difference? or it is better logging wise to show both sets? just curiuos what this output would look like

integrated_channels/blackboard/client.py
alex-sheehan-edx: yes, but the metadata is broken into three separate metadata objects

integrated_channels/blackboard/client.py
```diff
@@ -387,6 +481,25 @@ def _delete(self, url):
             raise ClientError(response.text, response.status_code)
         return response
 
+    def _handle_course_customization_error(self, bb_course_id, bb_content_id=None):
```
binodpant, line 484: I think this method isn't necessarily (on its own) an error handling method. I think a better name for it might something like delete_course_content_and_course (or sth like it)? The context in which it is used will I guess define (e.g. during error handling) but I can see it being used outside of that context too.. just a thought

integrated_channels/blackboard/client.py
alex-sheehan-edx: So this is checks for the presence of `course_metadata`, 'course_content_metadata`, and `course_child_content_metadata`

integrated_channels/blackboard/client.py
```diff
@@ -464,3 +577,50 @@ def _get_or_create_integrated_grade_column(self, bb_course_id):
                 )
 
         return grade_column_id
+
+    def create_integration_content_for_course(self, bb_course_id, content_metadata):
```
binodpant, line 581: can you double check but I thin kthe proper name for this var is channel_metadata_item? at least that is what we used in other methods. I think content_metadata may refer to the pre-exporter dataset while channel_metadata (and channel_metadata_item) to the post-exporter.. but I would check if this is 100% true. if so let's rename this var?

integrated_channels/blackboard/client.py
alex-sheehan-edx: this outputs any of three course metadata objects that are missing like:
`{'course_metadata', 'course_child_content_metadata`}

alex-sheehan-edx: so the log would show you anything that wasn't able to be exported

integrated_channels/blackboard/client.py
alex-sheehan-edx: Fair enough! I agree the method name should reflect the operations handled and not the context that the method was called within. I'm down to rename.

integrated_channels/blackboard/client.py
binodpant: that's cool, but sorry I was not clear. I thought that variable name shoiuld be `channel_metadata_item` since we work with one item at a time, no?

integrated_channels/blackboard/client.py
binodpant: interesting, I maybe was not clear on how the - operator works with these sets... but cool if you have verified it

integrated_channels/blackboard/client.py
alex-sheehan-edx: I used `content metadata` because our base methods for transmitting are named `update_content_metadata` and `create_content_metadata`. As for meaning, I don't see a major difference in the two labels. However, I like consistency and this is the obvious outlier, so willing to change!

integrated_channels/blackboard/client.py
alex-sheehan-edx: 🤦  You know, I think I originally read this parameter as `"content_metadata_items"` and as such thought "yes, there are mutliple content items here". Okay, I'm fine with making this singular (or changing to `content_metadata_items`

integrated_channels/blackboard/client.py
alex-sheehan-edx:
```
>>> set_a = {1, 2, 3}
>>> set_b = {2, 3, 8}
>>> set_c = set_a - set_b
>>> set_c
{1}
```
I have a set of the expected, `you should have these` content items. I subtract what the exporter supplies from that set so any missing should be the remainder.

binodpant (approved):
lgtm!

PR #1013, opened 2020-10-20
ENT-3116: Adds in API endpoint that mimics admin manager_learner functionality

jlajoie:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -15,6 +15,7 @@ from django.contrib.sites.models import Site
 from django.utils.translation import ugettext_lazy as _
 
 from enterprise import models, utils
+from enterprise.api.v1.fields import Base64EmailCSVField
 from enterprise.api_client.lms import ThirdPartyAuthApiClient
 from enterprise.constants import ENTERPRISE_PERMISSION_GROUPS, DefaultColors
 from enterprise.utils import (
@@ -913,4 +914,37 @@ class EnterpriseCustomerCourseEnrollmentsSerializer(serializers.Serializer):
                 'lms_user_id, tpa_user_id, user_email'
             )
 
+        return data
+
+
+# pylint: disable=abstract-method
+class EnterpriseCustomerBulkEnrollmentsSerializer(serializers.Serializer):
+    """
+    Serializes a email_csv or email field for bulk enrollment requests.
+    """
+    email = serializers.CharField(required=False)
+    email_csv = Base64EmailCSVField(required=False)
+    course_run_key = serializers.CharField(required=False)
+    course_mode = serializers.ChoiceField(
+        choices=[
+            ("audit", _("Audit")),
+            ("verified", _("Verified")),
+            ("professional", _("Professional Education")),
+            ("no-id-professional", _("Professional Education (no ID)")),
+            ("credit", _("Credit")),
+            ("honor", _("Honor")),
+        ],
+        required=False,
+    )
+    reason = serializers.CharField(required=False)
+    salesforce_id = serializers.CharField(required=False)
+    discount = serializers.DecimalField(None, 5, required=False)
+    notify = serializers.BooleanField(default=False)
+
+    def create(self, validated_data):
+        return validated_data
+
+    def validate(self, data):  # pylint: disable=arguments-differ
+        if not data.get('email') and not data.get('email_csv'):
+            raise serializers.ValidationError('Must include either email or email_csv in request.')
         return data
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -17,7 +17,13 @@ from rest_framework.exceptions import NotFound, ValidationError
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.renderers import JSONRenderer
 from rest_framework.response import Response
-from rest_framework.status import HTTP_200_OK, HTTP_400_BAD_REQUEST, HTTP_404_NOT_FOUND, HTTP_500_INTERNAL_SERVER_ERROR
+from rest_framework.status import (
+    HTTP_200_OK,
+    HTTP_202_ACCEPTED,
+    HTTP_400_BAD_REQUEST,
+    HTTP_404_NOT_FOUND,
+    HTTP_500_INTERNAL_SERVER_ERROR,
+)
 from rest_framework.views import APIView
 from rest_framework_xml.renderers import XMLRenderer
 from six.moves.urllib.parse import quote_plus, unquote  # pylint: disable=import-error,ungrouped-imports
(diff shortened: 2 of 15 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -914,3 +916,46 @@ def validate(self, data):  # pylint: disable=arguments-differ
...
+        ],
+        required=False,
+    )
+    reason = serializers.CharField(required=False)
+    salesforce_id = serializers.CharField(required=False)
+    discount = serializers.DecimalField(None, 5, required=False)
+    notify = serializers.BooleanField(default=False)
+
+    def create(self, validated_data):
+        return validated_data
```
raq929, line 961: Nit: blank line at end of file

enterprise/api/v1/views.py
```diff
@@ -200,6 +210,93 @@ def course_enrollments(self, request, pk):
...
+            except ValidationError as exc:
+                errors.append(exc)
+
+            if errors:
+                return Response(errors, status=HTTP_400_BAD_REQUEST)
+
+            for email in emails:
+                models.EnterpriseCustomerUser.objects.link_user(enterprise_customer, email)
+
+            course_run_key = serializer.data.get('course_run_key')
```
raq929, line 257: Should we not be getting this from `validated_data`?

enterprise/api/v1/views.py
```diff
@@ -200,6 +210,93 @@ def course_enrollments(self, request, pk):
...
+            mode = serializer.data.get('course_mode')
```
raq929, line 258: Same as above.

enterprise/api/v1/views.py
```diff
@@ -200,6 +210,93 @@ def course_enrollments(self, request, pk):
...
+            if course_run_key:
```
raq929, line 259: Right now it looks like if there's not a course run key, we just return 202? Is that what we want here, or should we return some sort of error response if there's no course run?

enterprise/api/v1/views.py
```diff
@@ -200,6 +210,93 @@ def course_enrollments(self, request, pk):
...
+                            "email": success.email,
+                            "username": success.username,
+                            "course_run_key": course_run_key,
+                            "discount_percentage": float(discount),
+                            "enterprise_customer_name": enterprise_customer.name,
+                            "enterprise_customer_uuid": str(enterprise_customer.uuid),
+                            "mode": mode,
+                            "sales_force_id": serializer.data.get('salesforce_id'),
+                        } for success in succeeded]
+                        EcommerceApiClient(get_ecommerce_worker_user()).create_manual_enrollment_orders(enrollments)
```
raq929, line 297: What happens if this fails? Do we return the appropriate error to our front end? I wonder if we want to have a celery task do this bit, because it might take some time...

enterprise/api/v1/views.py
```diff
@@ -200,6 +210,93 @@ def course_enrollments(self, request, pk):
...
+                models.EnterpriseCustomerUser.objects.link_user(enterprise_customer, email)
+
+            course_run_key = serializer.data.get('course_run_key')
+            mode = serializer.data.get('course_mode')
+            if course_run_key:
+                this_customer_linked_emails = [
+                    email for email, customer in already_linked_emails if customer == enterprise_customer
+                ]
+                linked_learners = list(emails) + this_customer_linked_emails
+                if linked_learners:
```
raq929, line 264: Same as above, if there are no linked learners, should we just return 202, or do something else?

enterprise/api/v1/views.py
```diff
@@ -200,6 +210,93 @@ def course_enrollments(self, request, pk):
 
         return Response(serializer.errors, status=HTTP_400_BAD_REQUEST)
 
+    @detail_route(methods=['post'], permission_classes=[permissions.IsAuthenticated])
+    @permission_required('enterprise.can_enroll_learners', fn=lambda request, pk: pk)
+    def enterprise_learners(self, request, pk):
```
raq929, line 215: I think we'll want to call this `enroll_enterprise_learners`? Since I suspect we'll eventually want to unenroll them, etc?

enterprise/api/v1/serializers.py
```diff
@@ -914,3 +916,46 @@ def validate(self, data):  # pylint: disable=arguments-differ
             )
 
         return data
+
+
+class Base64EmailCSVField(serializers.Field):
```
raq929, line 921: I'd put it in a separate fields file so people can find it for re-use, but that's a nit. How are we uploading other CSVs?

raq929 (commented):
Looking good in general! I've got some questions around validated data and what we return if values are missing.

enterprise/api/v1/views.py
```diff
@@ -200,6 +210,93 @@ def course_enrollments(self, request, pk):
...
+        enterprise_customer = self.get_object()
+        serializer = serializers.EnterpriseCustomerBulkEnrollmentsSerializer(
+            data=request.data,
+            context={
+                'enterprise_customer': enterprise_customer,
+                'request_user': request.user,
+            }
+        )
+        if serializer.is_valid():
+            singular_email = serializer.data.get('email')
```
raq929, line 228: I think you'll want to grab from `validated_data` instead?

enterprise/api/v1/serializers.py
```diff
@@ -914,3 +916,46 @@ def validate(self, data):  # pylint: disable=arguments-differ
...
+
+    def to_representation(self, value):
+        return value
+
+
+# pylint: disable=abstract-method
+class EnterpriseCustomerBulkEnrollmentsSerializer(serializers.Serializer):
+    email = serializers.CharField(required=False)
+    email_csv = Base64EmailCSVField(required=False)
+    course_run_key = serializers.CharField(required=False)
```
raq929, line 943: It seems like this is sort of required, why not make the serializer require it?

enterprise/api/v1/views.py
jlajoie: Yes we should actually, thank you 👍

enterprise/api/v1/views.py
jlajoie: If this fails we'll get a 500 error as a response. I agree though, ideally it'd be best to background this, I'm not sure if we do, but it doesn't look like we have celery set up for enterprise, but that might just be because it's intended as part of platform, I'll dig a little. If it's easy to do I'll add it, if it's difficult and expensive to set up a queue/workers I'll probably punt on this.

enterprise/api/v1/views.py
jlajoie: It still accepts the payload so I think a 202 is still appropriate here even if it ends up being a no-op, one of the downsides of taking a form and trying to pretend it's a REST API now.

enterprise/api/v1/views.py
jlajoie: Generally I want to avoid having verbs in endpoints, that's the point of the HTTP verbs. I don't see that as part of the ACs anywhere, or as part of the miro board spec, it's also not possible right now with the manage_leaners form, so not sure that's something we'll eventually want to even do, but if we decide to I think a DELETE call would be appropriate rather than changing the route name.

enterprise/api/v1/views.py
jlajoie: So this has parity with the original in that it's an optional param, if it's provided it'll enroll, if it's not, it'll just link the learners to the enterprise if that makes sense.

enterprise/api/v1/serializers.py
jlajoie: Yeah can move this. Other CSVs are uploaded as part of a multipart/form outside of DRF.

enterprise/api/v1/serializers.py
jlajoie: It's not required today which'll result in just linking the learners to an enterprise. Keeping parity with how the admin form works today here.
```    
        course = forms.CharField(
        label=_("Enroll these learners in this course"), required=False,
        help_text=_("To enroll learners in a course, enter a course ID."),
    )```

enterprise/api/v1/serializers.py
raq929: Oh, yeah, this approach is better. 🎉

raq929:
Don't forget to bump the version and changelog! ;)

raq929 (approved):
Good to go once checks are passing.

PR #1017, opened 2020-10-21
Fix data sharing consent test to actually test enrolling a learner wh…

kellmarie:
…o consented

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -585,7 +585,7 @@ class GrantDataSharingPermissions(View):
         if created:
             track_enrollment('data-consent-page-enrollment', request.user.id, course_id, request.path)
 
-    def _enroll_leaner_in_course(
+    def _enroll_learner_in_course(
             self,
             request,
             enterprise_customer,
@@ -636,27 +636,27 @@ class GrantDataSharingPermissions(View):
                         )
                     )
                     raise
-                try:
-                    self.create_enterprise_course_enrollment(request, enterprise_customer, course_id, license_uuid)
-                except IntegrityError:
-                    error_code = 'ENTGDS009'
-                    log_message = (
-                        '[Enterprise DSC API] IntegrityError while creating EnterpriseCourseEnrollment.'
-                        'Course: {course_id}, '
-                        'Program: {program_uuid}, '
-                        'EnterpriseCustomer: {enterprise_customer_uuid}, '
-                        'User: {user_id}, '
-                        'License UUID: {license_uuid}, '
-                        'ErrorCode: {error_code}'.format(
-                            course_id=course_id,
-                            program_uuid=program_uuid,
-                            enterprise_customer_uuid=enterprise_customer.uuid,
-                            user_id=request.user.id,
-                            license_uuid=license_uuid,
-                            error_code=error_code,
-                        )
+            try:
+                self.create_enterprise_course_enrollment(request, enterprise_customer, course_id, license_uuid)
+            except IntegrityError:
+                error_code = 'ENTGDS009'
+                log_message = (
+                    '[Enterprise DSC API] IntegrityError while creating EnterpriseCourseEnrollment.'
+                    'Course: {course_id}, '
+                    'Program: {program_uuid}, '
+                    'EnterpriseCustomer: {enterprise_customer_uuid}, '
+                    'User: {user_id}, '
+                    'License UUID: {license_uuid}, '
+                    'ErrorCode: {error_code}'.format(
+                        course_id=course_id,
+                        program_uuid=program_uuid,
+                        enterprise_customer_uuid=enterprise_customer.uuid,
+                        user_id=request.user.id,
+                        license_uuid=license_uuid,
+                        error_code=error_code,
                     )
-                    LOGGER.exception(log_message)
+                )
+                LOGGER.exception(log_message)
 
     @method_decorator(login_required)
     def get(self, request):  # pylint: disable=too-many-statements
@@ -718,7 +718,7 @@ class GrantDataSharingPermissions(View):
             # If DSC is entirely disabled proceed to enroll the learner in the course
             if not enterprise_customer.requests_data_sharing_consent:
                 try:
-                    self._enroll_leaner_in_course(
+                    self._enroll_learner_in_course(
                         request=request,
                         enterprise_customer=enterprise_customer,
                         course_id=course_id,
@@ -974,7 +974,7 @@ class GrantDataSharingPermissions(View):
         consent_provided = bool(request.POST.get('data_sharing_consent', False))
         if defer_creation is None and consent_record.consent_required() and consent_provided:
             try:
-                self._enroll_leaner_in_course(
+                self._enroll_learner_in_course(
                     request=request,
                     enterprise_customer=enterprise_customer,
                     course_id=course_id,
--- a/tests/test_enterprise/views/test_grant_data_sharing_permissions.py
+++ b/tests/test_enterprise/views/test_grant_data_sharing_permissions.py
@@ -438,6 +438,11 @@ class TestGrantDataSharingPermissions(MessagesMixin, TestCase):
         assert response.status_code == 302
         self.assertRedirects(response, 'https://google.com', fetch_redirect_response=False)
 
+        # Verify the enterprise course enrollment was made with and without a license
+        assert EnterpriseCourseEnrollment.objects.filter(
+            enterprise_customer_user=ecu,
+        ).exists() is True
+
         if license_uuid:
             assert LicensedEnterpriseCourseEnrollment.objects.filter(
                 license_uuid=license_uuid,
@@ -449,10 +454,6 @@ class TestGrantDataSharingPermissions(MessagesMixin, TestCase):
                 course_mode
             )
         else:
-            # Verify a user without a license has no course enrollments
-            assert EnterpriseCourseEnrollment.objects.filter(
-                enterprise_customer_user=ecu,
-            ).exists() is False
             assert not mock_enrollment_api_client.return_value.enroll_user_in_course.called
 
     @mock.patch('enterprise.views.render', side_effect=fake_render)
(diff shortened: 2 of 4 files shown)


tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -561,11 +561,13 @@ def test_post_course_specific_consent(
             user_id=self.user.id,
             enterprise_customer=enterprise_customer
         )
+        # Data sharing consent record needs to exist to check if they've already consented before
+        # Granted being false means they haven't enrolled in the course yet
         dsc = DataSharingConsentFactory(
             username=self.user.username,
             course_id=course_id,
             enterprise_customer=enterprise_customer,
-            granted=consent_provided
```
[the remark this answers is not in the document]
kellmarie: It shouldn't be preset on the DSC record though, it should only be getting passed in as a param in the POST request.

The DSC record gets the granted value updated to True after a learner enrolls in a course, it being set to True here is what causes the `consent_record.consent_required()` check to return False and is what made it so that this ddt test case: `(False, True, '/successful_enrollment', 'course-v1:edX+DemoX+Demo_Course', str(uuid.uuid4())),` never went through the assertions under line 602. Having DSC.granted and consent_provdided being the same value from initialization makes it impossible to hit those assertions.

DSC.granted basically means whether the learner has consented and successfully enrolled into the course before. `consent_provided` is whether or not the learner consented on this request to enroll.

If they're both False then the learner doesn't get enrolled because they didn't provide consent on this request. And if they're both True then the learner doesn't get enrolled because they had already consented and been enrolled previously.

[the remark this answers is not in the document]
kellmarie: https://github.com/edx/edx-enterprise/blob/master/consent/mixins.py#L42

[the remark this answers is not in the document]
chavesj: Okay now that I've looked at this in more depth I agree that `consent_provided` shouldn't be used for both "whether the user checked the box to provide DSC" and "user had previously provided consent" - which is the cause of confusion here. Passing `False` to the DSCFactory is the right thing to do since the user will [redirect](https://github.com/edx/edx-enterprise/blob/31183667f211582ffe872740c5bde141b40e720a/enterprise/views.py#L805) to `success_url` if they hit `get` in this view after previously providing consent, and the only way to hit `post` in this view is by hitting `get` first and submitting the form from the DSC page.

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -594,13 +596,10 @@ def test_post_course_specific_consent(
 
         assert resp.url.endswith(expected_redirect_url)  # pylint: disable=no-member
         assert resp.status_code == 302
-        if not defer_creation:
-            dsc.refresh_from_db()
-            assert dsc.granted is consent_provided
```
chavesj, line 599: Can you explain why this was removed?

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -594,13 +596,10 @@ def test_post_course_specific_consent(
...
 
-        # we'll only create an enrollment record if (1) creation is not deferred, (2) consent has not been provided,
+        # we'll only create an enrollment record if (1) creation is not deferred, (2) consent has been provided,
```
chavesj, line 600: Given our confusion between "consent was previously provided" and "consent is currently provided in the post request", it seems worth making the distinction in this comment

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
kellmarie: I moved it to be under line 602, we don't want dsc.granted to be consent_provided because of previous conversation, but we do want to verify that it's set to True after the learner enrolls in the course.

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -594,13 +596,10 @@ def test_post_course_specific_consent(
...
         # (3) we provide a license_uuid, and (4) we provide a course _run_ id
-        if not defer_creation and not consent_provided and license_uuid and course_id.endswith('Demo_Course'):
+        if not defer_creation and consent_provided and license_uuid and course_id.endswith('Demo_Course'):
```
chavesj, line 602: Nice, this now matches https://github.com/edx/edx-enterprise/blob/31183667f211582ffe872740c5bde141b40e720a/enterprise/views.py#L975

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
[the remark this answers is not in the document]
adamstankiewicz: It does seem there is a distinction between ddt's `consent_required` and the `granted` field on DSC, where `consent_required` probably shouldn't be used for the initial state of `granted` like Kellie mentioned, when `consent_required` is representing the expected param from the POST request, not necessarily the initial state of `granted`.

However, per Jeff's comment, if we are always creating the DSC record with `granted=False`, would we now be missing test case(s) for if/when there happens to be an existing DSC record with `granted=True` (say, for example, that field gets updated programmatically or through Django Admin somehow)?

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
chavesj: Yeah, that's not quite the same thing though. These lines tested whether the DSC record has been updated with `consent_provided` (given `defer_creation` is `False`), and ~~isn't~~ shouldn't be specific to `license_uuid` being passed. 

While digging to understand this I realized something potentially big - [these lines](https://github.com/edx/edx-enterprise/blob/31183667f211582ffe872740c5bde141b40e720a/enterprise/views.py#L639-L659) shouldn't be placed within the `if license_uuid:` block. See indentation from original code in [this commit diff](https://github.com/edx/edx-enterprise/commit/bf838e982476f615072fec8239ca2fc24e808463). If my assumptions are correct, enterprise course enrollments aren't possible without a `license_uuid` right now.

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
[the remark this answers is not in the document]
chavesj: We would still benefit from the following ddt test case: `(False, False, '/failure_url', 'course-v1:edX+DemoX+Demo_Course', str(uuid.uuid4()))`

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
kellmarie: Those lines aren't placed within the `if license_uuid:` block, that block is inside of `self._enroll_leaner_in_course(` which the update to the DSC record is outside of. Updating the DSC record is in line with this if statement though `if defer_creation is None and consent_record.consent_required() and consent_provided:` which is what it was originally in line with.

You had told me to put updating the granted value inside of the try block because it didn't actually matter in the original functionality failing to enroll the learner in a course would return a redirect to the failure url before it got updated anyway.

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
kellmarie: Reasonable, that comment was also just wrong about when we create the enrollment record. But I'll add a distinction for "when the user is giving consent without having previously consented"

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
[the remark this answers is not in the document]
kellmarie: I can add that ddt test case and I will also add a new set of tests for when `granted=True` as it's own set of ddt test cases so that we don't further complicate this set of tests.

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
chavesj: You don't seem to be talking about the lines I linked so take a look at this picture. Highlighted lines shouldn't be at that indentation level (within the `if license_uuid:` block). 

Screen Shot 2020-10-22 at 2 43 39 PM

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
chavesj: The comment was referring to how consent shouldn't have previously been provided, bringing us full circle to the initial point of confusion

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
chavesj: After talking to Brittney about this it isn't an urgent thing since, as far as we can tell, all enterprise enrollments are either with a `license_uuid` or `defer_creation`. Meaning, highlighted code doesn't even get hit anymore 😂 . Either way this indentation should be fixed to be consistent with previous implementation, we can look into cleaning DSC up later if Brittney and Joe so choose. 

Also, the function name `_enroll_leaner_in_course` could use some improvement - "learner" should be spelled right and since it requires `license_uuid` I propose it changes to `_enroll_licensed_learner_in_course`

kellmarie: Ahhh now I see what you mean. And oops on the typo

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
kellmarie: Well the original comment and the if statement in the test were both just wrong about the conditions for when the enrollment record got created in the view which is what I mean about the comment being wrong.

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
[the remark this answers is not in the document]
kellmarie: I added a lot of test cases might as well cover all of the bases and combinations

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
kellmarie: Well I unindented the thing, fixed the typo, changed the tests to verify that the enterprise course enrollment is created regardless of there being a license_uuid or not.

enterprise/views.py
```diff
@@ -636,27 +636,27 @@ def _enroll_leaner_in_course(
...
+                    'EnterpriseCustomer: {enterprise_customer_uuid}, '
+                    'User: {user_id}, '
+                    'License UUID: {license_uuid}, '
+                    'ErrorCode: {error_code}'.format(
+                        course_id=course_id,
+                        program_uuid=program_uuid,
+                        enterprise_customer_uuid=enterprise_customer.uuid,
+                        user_id=request.user.id,
+                        license_uuid=license_uuid,
+                        error_code=error_code,
```
iloveagent57, line 656: Does this mean the licensed enrollments are the _only_ enterprise enrollments getting created via this DSC POST view?

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
```diff
@@ -594,30 +604,120 @@ def test_post_course_specific_consent(
...
+        (True, True, '/successful_enrollment', 'course-v1:edX+DemoX+Demo_Course', ''),
+        (True, False, '/failure_url', 'course-v1:edX+DemoX+Demo_Course', ''),
+        (False, True, '/successful_enrollment', 'course-v1:edX+DemoX+Demo_Course', ''),
+        (False, False, '/failure_url', 'course-v1:edX+DemoX+Demo_Course', ''),
+        (True, True, '/successful_enrollment', 'edX+DemoX', ''),
+        (True, False, '/failure_url', 'edX+DemoX', ''),
+        (False, True, '/successful_enrollment', 'edX+DemoX', ''),
+        (False, False, '/failure_url', 'edX+DemoX', ''),
+    )
+    @ddt.unpack
```
iloveagent57, line 654: Not necessarily feedback for this PR, because I know you're following the pattern of existing tests, but I feel like this is _not_ a good use case for `ddt` - 
* it's really hard to read and understand which parameter lines up with which tuple element
* I'm generally not a fan of mixing input with expected output like this
* It complicates the logic of the test body.

I'd rather split up the primary test conditions (is creation deferred, was consent provided?) into smaller tests with some common helper functions to set up all the test data we need.  I feel like that'll help make any future modifications of code and associated tests much easier to deal with.

iloveagent57 (commented):
I'm interested in the ramifications of de-denting that chunk of code in `_enroll_learner_in_course()`.

enterprise/views.py
kellmarie: Yeah currently. That chunk of code that I'm de-denting wasn't supposed to have gotten indented when I put out the check for the license_uuid. According to another conversation with Jeff:

`After talking to Brittney about this it isn't an urgent thing since, as far as we can tell, all enterprise enrollments are either with a license_uuid or defer_creation. Meaning, highlighted code doesn't even get hit anymore 😂 .`

tests/test_enterprise/views/test_grant_data_sharing_permissions.py
chavesj: This test can be simplified to a single test case along the lines of "consent was previously provided, assert `_enroll_learner_in_course` was not called, assert the redirect was to `success_url`". We don't need all of these test cases for what is essentially a "take no action" scenario.

kellmarie: Alrighty I can reduce it down to just the one test case

chavesj:
LGTM once you pull from master and bump the version

enterprise/views.py
iloveagent57: OK. This whole thing feels really convoluted to me, but that's not your fault, so I'm happy to go along with this for now.

iloveagent57 (approved):

chavesj (approved):

PR #1022, opened 2020-10-27
ENT-3279: Upgrade to pytest 6+ (and factoryboy 3+)

binodpant:
Why? To bring enterprise in line with platform wrt pytest

Notes:

1: pytest 6 now fails on logging errors (instead of prior where used to go to stderr), courtesy Alex:
[#6433:](https://github.com/pytest-dev/pytest/pull/7231/files) If an error is encountered while formatting the message in a logging call, for example logging.warning("oh no!: %s: %s", "first") (a second argument is missing), pytest now propagates the error, likely causing the test to fail.
Previously, such a mistake would cause an error to be printed to stderr, which is not displayed by default for passing tests. This change makes the mistake visible during testing.
You may supress this behavior temporarily or permanently by setting logging.raiseExceptions = False.
2: here is a logging exception in the signal handler that most every test using the EnterpriseCustomerCatalogFactory() appears to hit!!! (which I do not like). I figure it's because one of the imports fails as we are not running inside the edx-platform environment with the needed libs.
3: This was behind almost all of the test failures with pytest6! The least obvious (still haven't fully grasped it) is a mock realted failure. My guess is we never hit the mock code correctly due to this error, but have not stepped through it.
4: The only real needed fix was to fix the logging.exception line

Q for reviewers: Is this the best way to apply the upgrade?

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/signals.py
+++ b/enterprise/signals.py
@@ -331,7 +331,9 @@ def update_enterprise_catalog_data(sender, instance, **kwargs):     # pylint: di
         else:
             response = catalog_client.get_enterprise_catalog(catalog_uuid=catalog_uuid)
     except NotConnectedToOpenEdX as exc:
-        logger.exception('Unable to update Enterprise Catalog {}'.format(str(catalog_uuid)))
+        logger.exception(
+            'Unable to update Enterprise Catalog {}'.format(str(catalog_uuid)), exc_info=exc
+        )
     else:
         if not response:
             # catalog with matching uuid does NOT exist in enterprise-catalog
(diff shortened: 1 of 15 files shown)


requirements/dev.txt
```diff
@@ -4,6 +4,10 @@
 #
 #    make upgrade
 #
+--index-url http://edx.devstack.devpi:3141/root/pypi/+simple/
```
binodpant, line 7: wtf is this? ok I can remove it manually eventually unless we want it for some reason

iloveagent57: That's pip saying that it's ok to search our local devpi servers for packages to install, it's configured in `/root/.pip/pip.conf` inside our containers.  However, it seems to never work as intended on travis.  I usually just strip it out manually, but I wish there was a better way to handle this.

enterprise/signals.py
```diff
@@ -331,7 +331,9 @@ def update_enterprise_catalog_data(sender, instance, **kwargs):     # pylint: di
         else:
             response = catalog_client.get_enterprise_catalog(catalog_uuid=catalog_uuid)
     except NotConnectedToOpenEdX as exc:
-        logger.exception('Unable to update Enterprise Catalog {}'.format(str(catalog_uuid)), exc)
+        logger.exception(
+            'Unable to update Enterprise Catalog {}'.format(str(catalog_uuid)), exc_info=exc
```
iloveagent57, line 335: Nit: I *think* adding `exc_info=exc` here is redundant, because we're calling `logger.exception()`, but it's also harmless, so no big deal either way.

iloveagent57 (approved):
Needs a new version and changelog entry, then I think it's good to go.

enterprise/signals.py
binodpant: ah I guess I am still getting used to this logging setup in python.. but yeah thanks for the note. I will have to delve into my logging training a bit then.

PR #1024, opened 2020-10-27
finalization on integrated Blackboard course content structure

alex-sheehan-edx:
**Merge checklist:**

**Post merge:**

Finalizes the integrated course content to be:
image

Also fixed error handling in course content creation process.

### DIFF
--- a/integrated_channels/blackboard/client.py
+++ b/integrated_channels/blackboard/client.py
@@ -130,7 +130,10 @@ class BlackboardAPIClient(IntegratedChannelApiClient):
             course_content_delete_url = self.generate_course_content_delete_url(bb_course_id, bb_content_id)
             self._delete(course_content_delete_url)
 
-        course_updated_response = self.create_integration_content_for_course(bb_course_id, channel_metadata_item)
+        course_updated_response = self.update_integration_content_for_course(
+            bb_course_id,
+            channel_metadata_item,
+        )
 
         return course_updated_response.status_code, course_updated_response.text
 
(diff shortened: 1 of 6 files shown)


integrated_channels/blackboard/client.py
```diff
@@ -579,23 +583,28 @@ def delete_course_from_blackboard(self, bb_course_id, bb_content_id=None):
         resp = self._delete(delete_course_path)
         return resp
 
-    def create_integration_content_for_course(self, bb_course_id, channel_metadata_item):
+    def create_integration_content_for_course(self, bb_course_id, channel_metadata_item, update=False):
```
binodpant, line 586: if it's not much of an ask I would like to recommend separation of methods here between update and create, as in 

`create_integration_content_for_course`

`update_integration_content_for_course`

mainly because we are doing two different things entirely, and also doing that will allow us to avoid the multiple boolean checks and massaging of the error messages too such as in  L605 (I presume the message for update case needs to say accordingly). It will make for easier reasoning of the code and errors too I think

what say you?

integrated_channels/blackboard/exporters/content_metadata.py
```diff
@@ -22,7 +22,27 @@ class BlackboardContentMetadataExporter(ContentMetadataExporter):
         'course_child_content_metadata': 'course_child_content_metadata',
     }
 
-    DESCRIPTION_TEXT_TEMPLATE = "<a href={enrollment_url}>Go to edX course page</a><br />"
+    DESCRIPTION_TEXT_TEMPLATE = "<a href={enrollment_url}>Go to edX course page</a><br/>"
```
binodpant, line 25: should these names match what are in the html template string more or less (e.g. go_to_link?)

integrated_channels/blackboard/client.py
```diff
@@ -579,23 +583,28 @@ def delete_course_from_blackboard(self, bb_course_id, bb_content_id=None):
         resp = self._delete(delete_course_path)
```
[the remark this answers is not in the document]
binodpant: this is done, hopefully will fix it

integrated_channels/blackboard/client.py
```diff
@@ -610,14 +619,18 @@ def create_integration_content_for_course(self, bb_course_id, channel_metadata_i
             content_child_id = course_fully_created_response.json().get('id')
         except ClientError as error:
             content_child_id = None
+            error_message = ' and received error response={}'.format(error.message)
+            is_error = True
 
         if not content_child_id:
-            error_handling_response = self.delete_course_from_blackboard(bb_course_id, bb_content_id)
+            error_message = error_message if is_error else ''
```
binodpant, line 626: if you decide to refactor as I had recommended, this comment will may be somewhat impacted but do existing tests cover this error handling change? may want to review/revisit if we want a more pointed test to test that we delete (or do not delete) in response to an error using mocks assert_called counts check maybe, but leave it upto you

integrated_channels/blackboard/client.py
alex-sheehan-edx: yeah, I'm happy to separate these out

integrated_channels/blackboard/client.py
```diff
@@ -560,42 +563,104 @@ def _get_or_create_integrated_grade_column(self, bb_course_id):
...
+                'content child ID and got error response={}. Deleted associated course content shell with response: '
+                '(status_code={}, body={})'.format(
+                    error_message,
+                    error_handling_response.status_code,
+                    error_handling_response.text
+                ),
+                HTTPStatus.NOT_FOUND.value
+            )
+
+        return course_fully_created_response
```
binodpant, line 642: should course_fully_created_response need to be explicitly initialized to a default value, in case we somehow fail to set in in the try block above? Figure that may not happen since _post does not raise, I think? Good to check though

binodpant (approved):
good to go, just a minor comment posted

integrated_channels/blackboard/client.py
alex-sheehan-edx: `_post` will raise if the response is >`400`. However, we catch that raise **and** if we can't retrieve an ID from that response, we raise a client error, so I don't think that a default value is needed.

PR #1037, opened 2020-11-10
ENT-3529: Added Spanish translations for data sharing consent page.

saleem-latif:
**Merge checklist:**

**Post merge:**


.tx/config
```diff
@@ -2,13 +2,13 @@
 host = https://www.transifex.com
 
 [edx-platform.edx-enterprise]
-file_filter = enterprise/conf/locale/<lang>/LC_MESSAGES/django.po
-source_file = enterprise/conf/locale/en/LC_MESSAGES/django.po
+file_filter = enterprise/locale/<lang>/LC_MESSAGES/django.po
```
zamanafzal, line 5: Is it intentional to change the translations file path?

enterprise/views.py
```diff
@@ -365,6 +365,11 @@ def get_default_context(self, enterprise_customer, platform_name):
                 _('If you decline to consent, that fact may be shared with {enterprise_customer_name}.').format(
                     enterprise_customer_name=enterprise_customer.name
                 ),
+                _(
```
zamanafzal, line 368: Please ignore if I am wrong. I think you are only adding the value without any key. For reference please check L369.

sameenfatima78 (approved):

.tx/config
saleem-latif: Yes, it is intentional. third party apps must have this path for translations otherwise translations do not work.

enterprise/views.py
saleem-latif: Actually, this is another item on the list. the main dict contains an item that is a list and this new message is part of that list.

zamanafzal (approved):
LGTM 👍

PR #1045, opened 2020-11-16
ENT-3614: Adds in API for CanvasEnterpriseCustomerConfiguration

jlajoie:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/api/v1/canvas/views.py
+++ b/integrated_channels/api/v1/canvas/views.py
@@ -1,0 +1,39 @@
+"""
+Viewsets for Canvas
+"""
+
+from edx_rbac.mixins import PermissionRequiredMixin
+from rest_framework import permissions, status, viewsets
+
+from django.core.exceptions import PermissionDenied
+from django.http import Http404
+
+from enterprise.models import EnterpriseCustomerUser
+from integrated_channels.canvas.models import CanvasEnterpriseCustomerConfiguration
+
+from .serializers import CanvasEnterpriseCustomerConfigurationSerializer
+
+
+class CanvasConfigurationViewSet(PermissionRequiredMixin, viewsets.ModelViewSet):
+    serializer_class = CanvasEnterpriseCustomerConfigurationSerializer
+    permission_classes = (permissions.IsAuthenticated,)
+    permission_required = 'enterprise.can_access_admin_dashboard'
+
+    def get_queryset(self):
+        enterprise_customer_users = EnterpriseCustomerUser.objects.filter(user_id=self.request.user.id)
+        if enterprise_customer_users:
+            enterprise_customers_uuids = [str(enterprise_customer_user.enterprise_customer.uuid) for
+                                          enterprise_customer_user in enterprise_customer_users]
+            enterprise_customer_uuid = self.request.query_params.get('enterprise_customer')
+
+            if enterprise_customer_uuid:
+                if enterprise_customer_uuid in enterprise_customers_uuids:
+                    return CanvasEnterpriseCustomerConfiguration.objects.filter(
+                        enterprise_customer__uuid=enterprise_customer_uuid
+                    )
+                else:
+                    raise PermissionDenied()
+            return CanvasEnterpriseCustomerConfiguration.objects.filter(
+                enterprise_customer__uuid__in=enterprise_customers_uuids
+            )
+        raise Http404
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,10 @@ Change Log
 Unreleased
 -----------
 
+[3.14.1]
+--------
+* Adds new API for Canvas LMS configurations.
+
 [3.14.0]
 --------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.14.0"
+__version__ = "3.14.1"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/enterprise/urls.py
+++ b/enterprise/urls.py
@@ -103,5 +103,5 @@ urlpatterns += [
     url(
         r'^integrated_channels/api/',
         include('integrated_channels.api.urls')
-    )
+    ),
 ]
--- a/integrated_channels/api/urls.py
+++ b/integrated_channels/api/urls.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 """
-URL definitions for enterprise API endpoint.
+URL definitions for integrated_channels API endpoint.
 """
 
 from django.conf.urls import include, url
--- a/integrated_channels/api/v1/canvas/serializers.py
+++ b/integrated_channels/api/v1/canvas/serializers.py
@@ -1,0 +1,12 @@
+"""
+Serializers for Canvas.
+"""
+from rest_framework import serializers
+
+from integrated_channels.canvas.models import CanvasEnterpriseCustomerConfiguration
+
+
+class CanvasEnterpriseCustomerConfigurationSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = CanvasEnterpriseCustomerConfiguration
+        fields = '__all__'
--- a/integrated_channels/api/v1/canvas/urls.py
+++ b/integrated_channels/api/v1/canvas/urls.py
@@ -1,0 +1,13 @@
+# -*- coding: utf-8 -*-
+"""
+URL definitions for integrated_channels/canvas API version 1.
+"""
+
+from rest_framework import routers
+
+from .views import CanvasConfigurationViewSet
+
+app_name = 'canvas'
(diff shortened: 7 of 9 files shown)


integrated_channels/api/v1/canvas/views.py
```diff
@@ -0,0 +1,53 @@
...
+from enterprise.models import EnterpriseCustomerUser
+
+
+from integrated_channels.canvas.models import CanvasEnterpriseCustomerConfiguration
+
+from .serializers import CanvasEnterpriseCustomerConfigurationSerializer
+
+
+class CanvasConfigurationViewSet(viewsets.ModelViewSet):
+    serializer_class = CanvasEnterpriseCustomerConfigurationSerializer
```
taliaedX, line 16: oh interesting you put this directly in the viewset instead of a mixin. I don't actually know what a mixin does, so seems fine?

integrated_channels/api/v1/canvas/views.py
```diff
@@ -0,0 +1,53 @@
...
+    permission_classes = (permissions.IsAuthenticated, permissions.DjangoModelPermissions,)
+
+    def get_queryset(self):
+        if self.request.user.is_staff:
+            return CanvasEnterpriseCustomerConfiguration.objects.all()
+        try:
+            enterprise_customer = EnterpriseCustomerUser.objects.get(user_id=self.request.user.id).enterprise_customer
```
taliaedX, line 23: do we typically send a user id or how might that look in the front end? I know we track enterprise customer id already which is why I just went with passing that along.

I think practically there is probably no one who belongs to multiple enterprises and also isn't a staff user, but that is a possible gap here as well.

integrated_channels/api/v1/canvas/views.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @permission_required('enterprise.can_access_admin_dashboard')
+    def list(self, request):
+        return super(CanvasConfigurationViewSet, self).list(request)
+
+    @permission_required('enterprise.can_access_admin_dashboard')
+    def create(self, request):
+        return super(CanvasConfigurationViewSet, self).create(request)
+
+    @permission_required('enterprise.can_access_admin_dashboard')
```
taliaedX, line 38: why not just define/wrap the permission_required at the class level?

integrated_channels/api/v1/canvas/views.py
```diff
@@ -0,0 +1,53 @@
...
+    def retrieve(self, request, pk=None):
+        return super(CanvasConfigurationViewSet, self).retrieve(request, pk=pk)
```
taliaedX, line 40: interesting. I haven't seen this many methods defined yet. How do the endpoints differentiate retrieve vs list? I get partial_update == PATCH vs update/create == POST. Does Django just know based on whether you pass a parameter or not?

integrated_channels/api/v1/canvas/views.py
taliaedX: Also, if we use the retrieve endpoint on the frontend, does retrieve expect an enterprise customer? Judging by "pk" I'm guessing it's looking for a pure id that we wouldn't have available on the front end (unless we just created the config), so the only endpoint usable would be "list" and the front ends for most (if not all) of the sidebar objects assumes/(passes to the api) 1 enterprise customer.

taliaedX (commented):
I'm not sure about the query set split (providing all configs if staff), but otherwise this looks fine.

integrated_channels/api/v1/canvas/views.py
jlajoie: Yeah a mixin just effectively is a superclass so if you have fields on a mixin it'll apply to the  subclasses as well. Since we weren't re-using it outside of this viewset didn't make sense to externalize.

integrated_channels/api/v1/canvas/views.py
jlajoie: I actually didn't know if that works but if it does HELL yes because this seemed dumb to do and didn't like that I typed it.

integrated_channels/api/v1/canvas/views.py
jlajoie: > do we typically send a user id or how might that look in the front end?

Should always have a User ID for an authenticated learner who's requesting data. Believe it gets unmarshalled out from the jwt. 

Good call on the multiple enterprises though, should handle this via a filter instead of get just in case multiple are returned.

Retrieve doesn't expect an enterprise customer and it shouldn't, though we should be able to support querystring args for querying.

Right now it expects the primary key for the Conf object. Since we're accessing the view directly PK made sense. For getting to a point of displaying which views are available, I figured the typical approach is "Hit the list endpoint to see what options are available, dive in via the retrieve endpoint".

integrated_channels/api/v1/canvas/views.py
jlajoie: Does it based on the endpoint requested and which HTTP verb you use. So:

* list -> `GET api/v1/resource/`
* retrieve -> `GET api/v1/resource/<id>`
* create -> `POST api/v1/resource/`
* update -> `PUT api/v1/resource/<id>`
* partial_update -> `PATCH api/v1/resource/<id>`
* destroy -> `DELETE api/v1/resource/<id>`

integrated_channels/api/v1/canvas/views.py
taliaedX: I think the list to specific flow makes sense generally, but I'm not sure it's really suitable here.
We shouldn't have multiple configs so we should be going to directly to a specific result.

We could still use a list and just assume the first item of the list is the result we want, but yeah I'm still iffy on parameters there because we're currently allowing multiple enterprises on the backend. That was why we went with a query string enterprise customer, so that even ECS/multi-enterprise users could limit down to one customer.

integrated_channels/api/v1/canvas/views.py
taliaedX: Well to clarify, I'm not sure if you can use that specific decorator at the class level (a brief look at the code suggests probably not). However, what we've been using is the PermissionRequiredMixin class (https://github.com/edx/edx-rbac/blob/a2af3aa4ab87e36bfbc7f5e1073f0f8b77802a54/edx_rbac/mixins.py#L16) and just defining `permission_required = 'enterprise.can_access_admin_dashboard'` in the ViewSet class and then this get_permission_object method. Not sure if you actually need it or not. The logic looks pretty close to the decorator, but the decorator sets object to a function passthrough whereas the class uses None....You can see this in use here: https://github.com/edx/edx-platform/blob/master/common/djangoapps/third_party_auth/samlproviderconfig/views.py#L80

integrated_channels/api/v1/canvas/views.py
jlajoie: Would it make more sense to just serialize this data as part of the EnterpriseCustomer endpoint and include all of configs as part of the response body?

```
{
	"uuid": "b0990312-c5a3-48c4-b0fe-b8d844249623",
	"name": "TestDemoEnterpriseTest",
	"slug": "testdemoenterprisetest",
	"active": true,
	"site": {
		"domain": "courses.stage.edx.org",
		"name": "stage"
	},
	"enable_data_sharing_consent": true,
	"enforce_data_sharing_consent": "at_enrollment",
	"branding_configuration": {
		"enterprise_customer": "b0990312-c5a3-48c4-b0fe-b8d844249623",
		"enterprise_slug": "testdemoenterprisetest",
		"logo": "https://edxuploads.s3.amazonaws.com/enterprise/branding/1/1_logo.png",
		"primary_color": "#1a337b",
		"secondary_color": "#d7e3fc",
		"tertiary_color": "#007d88"
	},
	"identity_provider": "saml-arbi-test-shib",
	"enable_audit_enrollment": true,
	"replace_sensitive_sso_username": false,
	"enable_portal_code_management_screen": false,
	"sync_learner_profile_data": true,
	"enable_audit_data_reporting": true,
	"enable_learner_portal": true,
	"enable_portal_reporting_config_screen": false,
	"enable_portal_saml_configuration_screen": false,
	"contact_email": null,
	"enable_portal_subscription_management_screen": false,
	"hide_course_original_price": false,
	"enable_analytics_screen": false,
	"enable_integrated_customer_learner_portal_search": true
}
```

and add in

```
moodle_lms_configuration: {},
canvas_lms_configuration: {},
```

What I want to avoid is us having something "smart" that'll just parse out and return whatever is on the backend like ```third_party_lms_config: {type: 'moodle'}```, where you're constantly trying to figure out what the type is on the frontend.

taliaedX: This seems like it just changes at which point we do the API calls without really changing the backend logic to me unless I'm missing something?

If you're proposing a foreign key relationship on the enterprise customer <> every configuration table that seems a bit heavy.
If you're proposing making the API call here during the EnterpriseCustomer view/api call then that's fine, but I am not sure that really changes the issue of returning all enterprise configurations vs only 1.

My only issue here was that we need to presume only 1 enterprise customer and the permission logic of checking if a user is staff and returning all configurations was counter to what we expect since it can result in many, many configurations.

taliaedX (approved):
Lookin good!

PR #1047, opened 2020-11-18
ENT-3700 Bug fix - raise instead of return empty

taliaedX:
**NOTE:** Most of the file changes here is just from me adding retry and running make upgrade. This is an optional change. Retry doesn't have any new dependencies so it seemed a simple lightweight wrapper to help with connection errors while I was dealing with them for other reasons.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api_client/enterprise_catalog.py
+++ b/enterprise/api_client/enterprise_catalog.py
@@ -167,6 +167,7 @@ class EnterpriseCatalogApiClient(JwtLmsApiClient):
                     'Failed to get content metadata for Catalog [%s] in enterprise-catalog due to: [%s]',
                     catalog_uuid, str(exc)
                 )
+                raise
 
         return list(content_metadata.values())
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,11 @@ Change Log
 Unreleased
 -----------
 
+[3.12.4]
+--------
+
+* Fix silent exception in catalog api call.
+
 [3.12.3]
 --------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.12.3"
+__version__ = "3.12.4"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name


enterprise/api_client/enterprise_catalog.py
```diff
@@ -139,6 +140,7 @@ def delete_enterprise_catalog(self, catalog_uuid):
             )
             return {}
 
+    @retry(exceptions=(ConnectionError, Timeout), tries=2, delay=4, logger=LOGGER)
```
taliaedX, line 143: this whole shebang is honestly optional. Just seemed useful if we have actual temporary errors when connecting to the catalog.

enterprise/api_client/enterprise_catalog.py
```diff
@@ -162,11 +164,12 @@ def get_content_metadata(self, enterprise_customer, enterprise_catalogs=None):
...
-                LOGGER.exception(
-                    'Failed to get content metadata for Catalog [%s] in enterprise-catalog due to: [%s]',
-                    catalog_uuid, str(exc)
-                )
+            except SlumberBaseException as exc:
+               LOGGER.exception(
+                   'Failed to get content metadata for Catalog [%s] in enterprise-catalog due to: [%s]',
+                   catalog_uuid, str(exc)
+               )
+               raise
```
taliaedX, line 172: I only updated get_content_metadata for now to just fix the integrated_channels issue. I don't think other calls returning empty dicts/lists/etc. is an issue atm. It's just integrated channels has to handle someone's catalog being removed/changed and thus equates empty items to items needing deletion thus the raise here.

This does end up causing a massive waterfalling exception traceback, but I really don't want to return something here.

taliaedX: I think the alternative would be to raise, catch, raise, etc. up the chain and only log the exception at the top level, but that's also the most useless part of the traceback.

binodpant: In general my thinking is that raising is almost always better than dropping exceptions. I like it, and if your local testing shows it works for the issue you tested, I am good with it. Interesting no test changes were needed..

binodpant (approved):

enterprise/api_client/enterprise_catalog.py
taliaedX: Yeah that also threw me off. Though I feel like if someone wrote a test like "exception raised from catalog and returns empty" they would have fixed it to not work like that lol.

enterprise/api_client/enterprise_catalog.py
alex-sheehan-edx: I get temporary/one time exceptions locally that this would be hugely helpful for. Do we see these in prod or staging ever? Would just make this all the more useful

enterprise/api_client/enterprise_catalog.py
alex-sheehan-edx: I think I'm with Binod on this one, I like raising and catching.

enterprise/api_client/enterprise_catalog.py
alex-sheehan-edx: Looked back at my logs, seems like `get_course_details` is the endpoint that sometimes intermittently fails for me. I wonder if something like this could be applied there?

alex-sheehan-edx (approved):
I'm curious about retry capabilities throughout the integrated channels. If tests are happy with the change, then I'm good with it! 👍

enterprise/api_client/enterprise_catalog.py
taliaedX: Yeah we have retry logic somewhere else as well (I was trying to find a pr Brandon B wrote up that added retries somewhere but couldn't find it). This library with minimal dependencies (only depends on decorator which we already use) was just a nice and clean wrapper, so I wanted to use it.

I definitely prefer having retry logic as good code practice. However, I think in the way that we use this method it doesn't really matter since we will just rerun the job at some point anyways? So it's sort of a toss up to me about whether I actually include this or not (also I'm not sure if I want to be responsible for all the other unnecessary package upgrades via `make upgrade` breaking prod... lol)

enterprise/api_client/enterprise_catalog.py
taliaedX: yeah there was always going to be raising here. It's just about whether we raise and catch up the chain to prevent printing every waterfall exception or not. I don't think it's a _huge_ deal though.

taliaedX: As far as testing I may add a test that actually checks that an exception is raised. I think that's probably what is missing here (or perhaps there is a test that verifies the log line from the exception and nothing else and since I kept the log line it doesn't affect the test - either way I'm gonna fix that)

PR #1048, opened 2020-11-18
Implementing base level course assessment level reporting 

alex-sheehan-edx:
**WIP**: 
- TODO's:
   - Logging

**_NOTE:_**
Canvas implementation of assessment level reporting is drafted and at a point in development where it's fully functional but not included in this PR as to keep base level work separate and finalized before confirming individual channels.

Having said that, this is what assessment level reporting is going to potentially manifest as:
image

image

Note that assessments that the user has not attempted will not be reported on.

**Merge checklist:**

**Post merge:**


integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -80,6 +83,133 @@ def grade_audit(self):
         """
         return self.GRADE_AUDIT
 
+    def fetch_course_details(self, course_details, enrollment_course_id):
+        """
+        Placeholder
+        """
+        # Check if first run or if the cached course ID doesn't match the enrollment course ID
+        if course_details is None or course_details['course_id'] != enrollment_course_id:
```
taliaedX, line 91: The fact that it looks like the EnterpriseCourseEnrollment table stores course_run_ids (which I see us querying off of in another file) and yet every part of the course api seems to say course_id even though it can use course_run....my brain. It's dying.

I was going to ask for consistently using `course_run_id` as we do later in this file, but since everything externally uses course_id as well it doesn't really help my confusion.

_this has been useless comments by Talia™️_

enterprise/api_client/lms.py
```diff
@@ -469,6 +469,17 @@ def get_course_grade(self, course_id, username):
 
         raise HttpNotFoundError('No grade record found for course={}, username={}'.format(course_id, username))
 
+    @JwtLmsApiClient.refresh_token
+    def get_course_assessment_grades(self, course_id, username):
+        """
+        Placeholder
```
binodpant, line 475: ??

integrated_channels/integrated_channel/transmitters/learner_data.py
```diff
@@ -32,6 +32,86 @@ def __init__(self, enterprise_configuration, client=IntegratedChannelApiClient):
...
+            app_label=kwargs.get('app_label', 'integrated_channel'),
+            model_name=kwargs.get('model_name', 'LearnerDataTransmissionAudit'),
+        )
+        kwargs.update(
+            TransmissionAudit=TransmissionAudit,
+        )
+
+        # Retrieve learner data for each existing enterprise enrollment under the enterprise customer
+        # and transmit the data according to the current enterprise configuration.
+        for learner_data in exporter.bulk_assessment_level_export(**kwargs):
```
taliaedX, line 80: this is my stupidity, but in the bulk_assessment_level_export method we're yielding records, which from my understanding suspends the function. At what point does it produce all results for this for loop? Is it basically streaming them (ie 'if record, yield record' goes to 'for learner_data in that record...' then go back to the suspended bulk function where we left off, yield another record, etc.?)

integrated_channels/utils.py
```diff
@@ -161,20 +161,26 @@ def get_image_url(content_metadata_item):
...
-def is_already_transmitted(transmission, enterprise_enrollment_id, grade):
+def is_already_transmitted(transmission, enterprise_enrollment_id, grade, subsection_id=None):
     """
     Returns: Boolean indicating if completion date for given enrollment is already sent of not.
 
     Args:
         transmission: TransmissionAudit model to search enrollment in
         enterprise_enrollment_id: enrollment id
         grade: 'Pass' or 'Fail' status
+        subsection_id (Optional): The id of the subsection, needed if there exist multiple audits per course
```
taliaedX, line 172: multiple audits per course? I get that means the audit table, but my brain was confused for a second there, can we just say for assessments?

enterprise/api_client/lms.py
alex-sheehan-edx: image

integrated_channels/utils.py
```diff
@@ -161,20 +161,26 @@ def get_image_url(content_metadata_item):
...
     try:
         already_transmitted = transmission.objects.filter(
             enterprise_course_enrollment_id=enterprise_enrollment_id,
             error_message='',
-        ).latest('id')
+        )
+        if subsection_id:
+            already_transmitted.filter(subsection_id=subsection_id)
+
+        already_transmitted = already_transmitted.latest('id')
```
taliaedX, line 182: So we see if a course exists in the transmission logs for grades and then if we're trying to actually check assessments, we then filter the already-transmitted course down to subsection, but you have `already_transmitted.filter(subsection_id=subsection_id)` without assignment so does it still save the filtered result set when you then call `already_transmitted.latest('id')

taliaedX (commented):
I haven't run this or anything (we probably should at some point before this becomes available, but we can do that during the Canvas review) but aside from some questions because my brain is smol, this looks fine to me.

enterprise/api_client/lms.py
```diff
@@ -469,6 +469,17 @@ def get_course_grade(self, course_id, username):
...
+        """
+        results = self.client.gradebook(course_id).get(user_contains=username).get('results')
```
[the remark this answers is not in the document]
alex-sheehan-edx: Writing up here because we did offline work- TLDR platform's gradebook API currently has a bug and `500's` when specifically querying with `username`. We were able to find out exactly why this happens, but lack the context within the gradebook API to suggest a fix at this point. I've notified `masters-grades` but their response time seems a tad slow so we're going to continue with this line until we can get some more info.

enterprise/api_client/lms.py
```diff
@@ -469,6 +469,17 @@ def get_course_grade(self, course_id, username):
...
+        for row in results:
+            if row.get('username') == username:
+                return row.get('section_breakdown')
+
+        raise HttpNotFoundError('No grade record found for course={}, username={}'.format(course_id, username))
```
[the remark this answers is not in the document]
alex-sheehan-edx: Lol thanks for pointed out, I'd have probably missed it 🙃

integrated_channels/integrated_channel/exporters/learner_data.py
alex-sheehan-edx: Off the top of my head, I've seen mixtures of `course_id`, `courseID`, and `course_run_id` all pretty much used for the same thing.. and yet not. It's my understanding that most of this confusion came from the fact that the original integrated channels didn't know whether or not it was dealing with a course or a course run, so they had to compensate for both. With assessment level, I don't believe we will ever get an assessment level grade from a course (please please correct me if I'm wrong because then I have some more work to do 🙃 ), so there is an argument to making a switch.

integrated_channels/integrated_channel/transmitters/learner_data.py
alex-sheehan-edx: Don't think it's stupidity at all, I think your description is fairly accurate. It's just a way to stream each learner data record to the transmitter.

integrated_channels/utils.py
alex-sheehan-edx: Sure! I think both are accurate to the situation, multiple assessments and each assessment has an audit. This is `**assessment level reporting**` so I don't think it's a stretch at all to say assessments here.

integrated_channels/utils.py
alex-sheehan-edx: I was working under the assumption that the `alread_transmitted` object would retain the filter. I believe we do something similar in the base course level exporter ie 
```
        enrollment_queryset = EnterpriseCourseEnrollment.objects.select_related(
            'enterprise_customer_user'
        ).filter(
            enterprise_customer_user__enterprise_customer=self.enterprise_customer,
            enterprise_customer_user__active=True,
        )
        if learner_to_transmit and course_run_id:
            enrollment_queryset = enrollment_queryset.filter(
                course_id=course_run_id,
                enterprise_customer_user__user_id=learner_to_transmit.id,
            )
            exporting_single_learner = True
            generate_formatted_log(
                'Exporting single learner. Course: {course_run}, User: {user_id}'.format(
                    course_run=course_run_id,
                    user_id=learner_to_transmit.id
                ),
                channel_name=channel_name,
                enterprise_customer_identifier=self.enterprise_customer.name
            )
        enrollment_queryset = enrollment_queryset.order_by('course_id')
```
But perhaps this is slightly different? I would assume that the `.latest('id)` is still necessary for the assessment level transmission because we may have multiple records for a single particular subsection and need to get the latest one. Please correct me if I'm wrong here!

integrated_channels/integrated_channel/exporters/learner_data.py
taliaedX: Yeah it's my understanding we wouldn't be able to have assessments without course runs. However, I think using course_run and having to say course_id==course_run in spots doesn't do much to alleviate the issue, which is why I said my comment was just useless lol.

I imagine we _could_ switch everything enrollment related, but not sure it's worth it at this point and not sure if we'll still hit spots where we equate a "course_id" with course run id anyways.

integrated_channels/utils.py
taliaedX: yeah I think basically swapping audit for subsection or assessment might be a little clearer. That or just `multiple assessment audits` but that implies multiple audits per assessment (which might also be possible idk)

integrated_channels/utils.py
taliaedX: yeah but where we're filtering in that example we're setting assignment:
```
enrollment_queryset = enrollment_queryset.filter(
                course_id=course_run_id,
                enterprise_customer_user__user_id=learner_to_transmit.id,
            )
```
but yours is just:
```
        if subsection_id:
            already_transmitted.filter(subsection_id=subsection_id)
```
That's what I'm confused by.

taliaedX: I'm a bit doubtful running a filter on an object inherently updates/saves the object with the filter without assignment? But I could be wrong.

integrated_channels/integrated_channel/exporters/learner_data.py
alex-sheehan-edx: in my eyes, defs not worth lol

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -80,6 +83,148 @@ def grade_audit(self):
...
+        * ``grade``: string grade recorded for the learner in the course.
+        """
+        enrollment_queryset = EnterpriseCourseEnrollment.objects.select_related(
+            'enterprise_customer_user'
+        ).filter(
+            enterprise_customer_user__enterprise_customer=self.enterprise_customer,
+            enterprise_customer_user__active=True,
+        ).order_by('course_id')
+
+        # Fetch course details from the Course API, and cache between calls.
```
alex-sheehan-edx, line 120: so I'm suddenly thinking we may not need course details in the bulk export

enterprise/api_client/lms.py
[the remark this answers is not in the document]
binodpant: is there a tracking ticket for this? if we need to make a change here after that it may be worth noting here or even in the code... otherwise okie

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -372,6 +532,58 @@ def _collect_certificate_data(self, enterprise_enrollment):
...
+        try:
+            assessment_grades_data = self.grades_api.get_course_assessment_grades(course_id, username)
+        except HttpNotFoundError as error:
+            if hasattr(error, 'response'):
+                response_content = error.response.json()  # pylint: disable=no-member
+                if response_content.get('error_code', '') == 'user_not_enrolled':
+                    return {}
+
+            return {}
+
```
binodpant, line 571: not clear why the error processing is here? it seems we always return {} regardless and there is no raise or anything

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_learner_data.py
```diff
@@ -378,6 +415,47 @@ def test_learner_data_self_paced_course(
...
+        mock_enrollment_api.return_value.get_course_enrollment.return_value = dict(
+            mode='verified',
+        )
+
+        # Mock grades data
+        assessment_grade_data = [dict(
+            attempted=True,
+            subsection_name='subsection_1',
+            category='subsection_category',
+            percent='1.0',
```
binodpant, line 443: do these come in as strings in real life?

integrated_channels/utils.py
binodpant: hmm did not think filter did any updates... https://docs.djangoproject.com/en/3.1/ref/models/querysets/#filter you may have to assign the output of filter back if further procesing

integrated_channels/integrated_channel/transmitters/learner_data.py
```diff
@@ -32,6 +32,100 @@ def __init__(self, enterprise_configuration, client=IntegratedChannelApiClient):
...
+                code = client_error.status_code
+                body = client_error.message
+                self.handle_transmission_error(learner_data, client_error)
+
+            learner_data.status = str(code)
+            learner_data.error_message = body if code >= 400 else ''
+
+            learner_data.save()
+
+    def assessment_level_transmit(self, exporter, **kwargs):
```
binodpant, line 72: hmm.. this model of passing the exporter itself is new? I thought we typically did `transmitter.transmit(exporter.export())` or something like it?

enterprise/api_client/lms.py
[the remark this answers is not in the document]
alex-sheehan-edx: So far no one has responded to my messages in `masters-grades`

integrated_channels/utils.py
alex-sheehan-edx: Talia and I labbed this and are (I hope) happy with how it's handled.

integrated_channels/integrated_channel/exporters/learner_data.py
alex-sheehan-edx: oh, you know- this existed in the course learner data exporter. I don't think it's really necessary  ¯\_(ツ)_/¯

alex-sheehan-edx: Ahh, I see. The course learner data transmitter uses the error response content for logging reasons then returns the same thing. So yup, not needed. Unless we wanted to add more logging here.

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_learner_data.py
alex-sheehan-edx: Oh, percent is a float. Everything else should be ok!

integrated_channels/integrated_channel/transmitters/learner_data.py
alex-sheehan-edx: learner data and content metadata are slightly different! 

**Course grade reporting**
```
    def transmit_learner_data(self, user):
        """
        Iterate over each learner data record and transmit it to the integrated channel.
        """
        exporter = self.get_learner_data_exporter(user)
        transmitter = self.get_learner_data_transmitter()
        transmitter.transmit(exporter)
```

**Assessment level grade reporting**
```
    def transmit_subsection_learner_data(self, user):
        """
        Iterate over each assessment learner data record and transmit them to the integrated channel.
        """
        exporter = self.get_learner_data_exporter(user)
        transmitter = self.get_learner_data_transmitter()
        transmitter.assessment_level_transmit(exporter)
```

**Content metadata transmission**
```
    def transmit_content_metadata(self, user):
        """
        Transmit content metadata to integrated channel.
        """
        exporter = self.get_content_metadata_exporter(user)
        transmitter = self.get_content_metadata_transmitter()
        transmitter.transmit(exporter.export())
```

alex-sheehan-edx: You can look at the base integrated channels models.py to see these.

integrated_channels/integrated_channel/exporters/learner_data.py
taliaedX: I'm mostly curious what we want to do in the case we don't hit that if condition (ie `if response_content.get('error_code', '') == 'user_not_enrolled':`). We seem to be saying that in this condition, we can just provide an empty dict, but otherwise.....something should happen. Either we just log and continue or we raise or something.

alex-sheehan-edx: I copied the behavior here from the similar `_collect_grades_data` method which does the following: 
```
        except HttpNotFoundError as error:
            if hasattr(error, 'response'):
                response_content = error.response.json()  # pylint: disable=no-member
                if response_content.get('error_code', '') == 'user_not_enrolled':
                    # This means the user has an enterprise enrollment record but is not enrolled in the course yet
                    LOGGER.info(
                        '[Integrated Channel] User is not enrolled in the course.'
                        ' Course: {course_id}, EnterpriseEnrollment: {enterprise_enrollment},'
                        ' Username: {username}'.format(
                            course_id=course_id,
                            username=username,
                            enterprise_enrollment=enterprise_enrollment.pk))
                    return None, None, None, None

            LOGGER.error('[Integrated Channel] Grades data not found.'
                         ' Course: {course_id}, EnterpriseEnrollment: {enterprise_enrollment},'
                         ' Username: {username}'.format(
                             course_id=course_id,
                             username=username,
                             enterprise_enrollment=enterprise_enrollment.pk))
            return None, None, None, None
```
The outcomes of both conditions yields the same return, but we log that there is an enterprise enrollment without a course enrollment. Since I was avoiding logging to make the PR more clear in its intent, I just removed the first condition. Having said that, I don't even think there's a reason for the first condition to return at all, except to not double log.  ¯\_(ツ)_/¯

taliaedX: Yeah it's just in copying the behaviour you made an if condition that doesn't do anything. At least in that condition above it logs based on the condition. 

It's fine if we feel logs here would be useless, but then since it's not raising an exception, I think we should just get rid of this entire `if hasattr...` block.

alex-sheehan-edx: https://github.com/edx/edx-enterprise/pull/1048/files#diff-f488bfc03a35af19cc710b17a06ab3014e240abef4ad52114822c8b0e081b5caR562-R565
I already did, which is why I originally marked this as `resolved`. Unless you think logs here are necessary, in which case I'm happy to add it back in.

enterprise/api_client/lms.py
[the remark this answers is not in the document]
alex-sheehan-edx: we got a response! And hopefully a fix soon 🤞

tests/test_integrated_channels/test_integrated_channel/test_transmitters/test_learner_data.py
```diff
@@ -0,0 +1,70 @@
...
+@mark.django_db
+@ddt.ddt
+class TestLearnerDataTransmitter(unittest.TestCase):
+    """
+    Tests for the class ``LearnerDataTransmitter``.
+    """
+    def setUp(self):
+        super(TestLearnerDataTransmitter, self).setUp()
+
+        enterprise_customer = factories.EnterpriseCustomerFactory(name='Starfleet Academy')
```
taliaedX, line 26: cute

taliaedX (approved):
I have a minor quibble but otherwise it looks good to me.

integrated_channels/degreed/client.py
```diff
@@ -40,6 +40,11 @@ def __init__(self, enterprise_configuration):
         self.session = None
         self.expires_at = None
 
+    def create_assessment_reporting(self, user_id, payload):
+        """
+        Not implemented yet.
```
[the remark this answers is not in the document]
alex-sheehan-edx: pylint was complaining because the base level client has the non-implemented method for `create_assessment_reporting`, following the pattern for all the `client` methods. So I believe all the other integrated channels need this as of right now?

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -80,6 +83,140 @@ def grade_audit(self):
...
+            return None
+
+        # No caching because we're only fetching one course detail
+        course_details = self.fetch_course_details(None, course_run_id)
+
+        if not course_details:
+            return None
+
+        if not LearnerExporter.has_data_sharing_consent(enterprise_enrollment):
+            return None
```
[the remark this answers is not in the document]
alex-sheehan-edx: Great suggestion! Idk why my brain just went `oh return None`.

integrated_channels/degreed/client.py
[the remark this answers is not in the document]
taliaedX: ahhh I didn't realize that was also missing in the base class. My b.

jmbowman:
It looks like this and https://github.com/edx/edx-enterprise/commit/ab2b1202eb238073e97cbab042f51ae98c24feaa#diff-75f77e06cece69709273d656d58a5d9cd8be7f2cb8bf9bf78bf764e93bd0b018 were added roughly in parallel, so the two new tasks added here didn't get the `@set_code_owner_attribute` decorator that was added to the rest of the tasks in the module.  Could you follow up and add that when you get a chance?

PR #1051, opened 2020-11-20
ENT-3723: Rebranding update: more robust get_platform_logo_url and Learner Portal default colors update

chavesj:
**ENT-3723:**

As part of the rebranding effort, `get_logo_url` is being updated. Since `get_logo_url` will always return an absolute URL, we can remove the use of `urljoin` in `get_platform_logo_url`

[Associated edx-platform PR](https://github.com/edx/edx-platform/pull/25654)

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -7,7 +7,6 @@ import datetime
 import json
 import logging
 import re
-from urllib.parse import urljoin
 from uuid import UUID
 
 import bleach
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -7,6 +7,7 @@ import datetime
 import json
 import logging
 import re
+from urllib.parse import urljoin
 from uuid import UUID
 
 import bleach
@@ -1455,22 +1456,23 @@ def enroll_users_in_course(
 
 
 def is_valid_url(url):
-  try:
-    result = urlparse(url)
-    return all([result.scheme, result.netloc])
-  except ValueError:
-    return False
+    """
+    Return where the specified URL is a valid absolute url.
+    """
+    try:
+        result = urlparse(url)
+        return all([result.scheme, result.netloc])
+    except ValueError:
+        return False
 
 
 def get_platform_logo_url():
     """
     Return an absolute URL of the platform logo using the branding api.
     """
-    # Return fake URL for tests rather than mock get_logo_url for every test using EnterpriseCustomer
-    fake_url = 'http://fake.url'
-    logo_url = get_logo_url() 
+    logo_url = get_logo_url() if get_logo_url else None
     if not logo_url:
-        return fake_url
+        return None
 
     is_abs_url = is_valid_url(logo_url)
     if is_abs_url:
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -1455,15 +1455,29 @@ def enroll_users_in_course(
     return successes, pending, failures
 
 
+def is_valid_url(url):
+  try:
+    result = urlparse(url)
+    return all([result.scheme, result.netloc])
+  except ValueError:
+    return False
+
+
 def get_platform_logo_url():
     """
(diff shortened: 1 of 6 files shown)


enterprise/utils.py
```diff
@@ -1446,10 +1445,7 @@ def get_platform_logo_url():
     Return an absolute URL of the platform logo using the branding api
     """
     # Return fake URL for tests rather than mock get_logo_url for every test using EnterpriseCustomer
-    return urljoin(
```
awaisdar001, line 1449: I would just want to have what was the reason it was added in the first place? Just wanted to make sure we are not missing an edge case.

iloveagent57 (commented):
Let's have a unit test for this?  At least in a follow-on PR, if you're trying to ship this quickly.

enterprise/utils.py
chavesj: I believe this was done in this way because `get_logo_url` returned a relative path to the platform logo URL at time of implementation and the domain of that URL was the LMS base URL.

awaisdar001: just wanted to throw this question, can you tell if openedX installations are using edx-enterprise? i think the answer is no, but just wanted to ask.

iloveagent57 (approved):

adamstankiewicz (approved):

enterprise/utils.py
chavesj: Here's the [devstack setting](https://github.com/edx/edx-platform/blob/master/lms/envs/devstack.py#L350) to enable/disable enterprise integration (use of edx-enterprise). Enterprise integration is currently enabled by default in devstack.

enterprise/utils.py
```diff
@@ -1446,10 +1445,7 @@ def get_platform_logo_url():
...
-        settings.LMS_ROOT_URL,
-        get_logo_url()
-    ) if get_logo_url else 'http://fake.url'
+    return get_logo_url() if get_logo_url else 'http://fake.url'
```
awaisdar001, line 1448: Can we make this method a little robust using `urlparse`?
here is what i mean. 

```
from urllib.parse import urlparse


def is_valid_url(url):
  try:
    result = urlparse(url)
    return all([result.scheme, result.netloc])
  except ValueError:
    return False

fake_url = 'http://fake.url'
logo_url = get_logo_url() 
if not logo_url:
    return fake_url

is_abs_url = is_valid_url(logo_url)
if is_abs_url:
    return logo_url

return urljoin(settings.LMS_ROOT_URL, logo_url)
```

the reason I say that there is a good chance that a logo can be overridden using site_configuration with either `logo_image_url` key or `university` key. so in that case, we can still get a relative URL (which would need urljoin).

awaisdar001: if we do this way, then this PR shouldn't be blocked by the edx-platform PR.

enterprise/utils.py
```diff
@@ -1451,15 +1450,29 @@ def enroll_users_in_course(
...
     """
     # Return fake URL for tests rather than mock get_logo_url for every test using EnterpriseCustomer
-    return urljoin(
-        settings.LMS_ROOT_URL,
-        get_logo_url()
-    ) if get_logo_url else 'http://fake.url'
+    fake_url = 'http://fake.url'
+    logo_url = get_logo_url()
+    if not logo_url:
+        return None
```
awaisdar001, line 1469: would like to return `fake_url` url here?

adamstankiewicz: Based on the comment "Return fake URL for tests rather than mock get_logo_url for every test using EnterpriseCustomer", I think the `fake_url` was only included here for tests? I think it would be odd to actually return `http://fake.url` for real as that's not a valid URL, so might as well just return `None`.

I will remove that comment now that the test has been updated.

awaisdar001: Ok got it.

PR #1053, opened 2020-11-24
Moodle Integration - Announcement Post/Course Details

taliaedX:
This creates an announcement post on course creation.
I also did some cleanup/fixing of the rest of the tasks as well.

**TODO:**
- fix/add tests
- Some missing doc strings
- Probably 800 other style fixes.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/moodle/client.py
+++ b/integrated_channels/moodle/client.py
@@ -39,7 +39,7 @@ def moodle_request_wrapper(method):
             # On course creation (and ONLY course creation) success,
             # Moodle returns a list of JSON objects, because of course it does.
             # Otherwise, it fails instantly and returns actual JSON.
-            return response
+            return body
         if isinstance(body, int):
             # This only happens for grades AFAICT. Zero also doesn't necessarily mean success,
             # but we have nothing else to go on
@@ -311,6 +311,52 @@ class MoodleAPIClient(IntegratedChannelApiClient):
         return self._post(params)
 
     @moodle_request_wrapper
+    def _wrapped_create_content_metadata(self, payload):
+        return self._post(payload)
+
+    @moodle_request_wrapper
+    def _get_course_forum(self, course_id):
+        """
+        Obtains a list of forums for a given course id.
+        """
+        params = {
+            'wsfunction': 'mod_forum_get_forums_by_courses',
+            'courseids[]': course_id,
+            'moodlewsrestformat': 'json',
+            'wstoken': self.token,
+        }
+        response = requests.get(
+            urljoin(
+                self.enterprise_configuration.moodle_base_url,
+                self.MOODLE_API_PATH,
+            ),
+            params=params
+        )
+        return response
+
+    def get_announcement_forum(self, course_id):
+        """
+        The response is a list type, because of course it is.
+        For new courses, there should only be 1 forum and it should be Announcements
+        """
+        response = self._get_course_forum(course_id)
+        return response[0]['id']
+
+    def _create_forum_post(self, course_id, announcement):
+        """
+        Creates a new Announcement post featuring edX course details and link.
+        """
+        forum_id = self.get_announcement_forum(course_id)
+        params = {
+            'wsfunction': 'mod_forum_add_discussion',
+            'forumid': forum_id,
+            'subject': 'Welcome! Click here for course details and edX link',
+            'options[0][name]': 'discussionpinned',
+            'options[0][value]': 'true',
+            'message': announcement
+        }
+        return self._post(params)
+
     def create_content_metadata(self, serialized_data):
         """
         The below assumes the data is dict/object.
@@ -325,8 +371,24 @@ class MoodleAPIClient(IntegratedChannelApiClient):
         }
         """
         serialized_data['wsfunction'] = 'core_course_create_courses'
-        response = self._post(serialized_data)
-        return response
+        announcements = serialized_data.pop('announcements')
+        response = self._wrapped_create_content_metadata(serialized_data)
+        # Response format should be [{"id":1, "shortname": "name"},{...}]
+        for course in response:
+            if course.get('id', None):
+                post = self._create_forum_post(
+                    course.get('id'),
+                    announcements[course.get('shortname')]
+                )
+                if post.json()['warnings']:
+                    raise ClientError(
+                        'Moodle Client failed to create post for course {}'.format(
+                            course.get('shortname')
+                        ),
+                        HTTPStatus.BAD_REQUEST.value
+                    )
+        return 200, ''
+
 
     @moodle_request_wrapper
     def update_content_metadata(self, serialized_data):
@@ -338,6 +400,12 @@ class MoodleAPIClient(IntegratedChannelApiClient):
     @moodle_request_wrapper
     def delete_content_metadata(self, serialized_data):
         moodle_course_id = self.get_course_id(serialized_data['courses[0][shortname]'])
+        import pdb; pdb.set_trace()
+        course_ids_to_delete = []
+        for key in list(serialized_data):
+            if 'shortname' in key:
+                moodle_course_id = self.get_course_id(serialized_data[key])
+                course_ids_to_delete.append(('courseids[]', moodle_course_id))
         params = {
             'wsfunction': 'core_course_delete_courses',
             'courseids[]': moodle_course_id
--- a/integrated_channels/moodle/client.py
+++ b/integrated_channels/moodle/client.py
@@ -15,6 +15,7 @@ from integrated_channels.exceptions import ClientError
 from integrated_channels.integrated_channel.client import IntegratedChannelApiClient
 
 MOODLE_FINAL_GRADE_ASSIGNMENT_NAME = '(edX integration) Final Grade'
+ANNOUNCEMENT_POST_SUBJECT = 'Welcome! Click here for course details and edX link'
 
 
 def moodle_request_wrapper(method):
@@ -315,6 +316,10 @@ class MoodleAPIClient(IntegratedChannelApiClient):
         return self._post(payload)
 
     @moodle_request_wrapper
+    def _wrapped_update_content_metadata(self, payload):
+        return self._post(payload)
+
+    @moodle_request_wrapper
     def _get_course_forum(self, course_id):
         """
         Obtains a list of forums for a given course id.
@@ -342,6 +347,42 @@ class MoodleAPIClient(IntegratedChannelApiClient):
         response = self._get_course_forum(course_id)
         return response[0]['id']
 
+    @moodle_request_wrapper
+    def _get_forum_discussions(self, forum_id):
+        """
+        Returns discussion posts found in the specified forum
+        """
+        params = {
+            'wsfunction': 'mod_forum_get_forum_discussions',
+            'forumid': forum_id,
+            'moodlewsrestformat': 'json',
+            'wstoken': self.token,
+        }
+        return self._post(params)
+
+    def get_announcement_post(self, forum_id):
+        """
+        Returns the id of the discussion post matching our announcement or returns nothing.
+        """
+        response = self._get_forum_discussions(forum_id)
+        for index, discussion in enumerate(response):
+            if discussion['subject'] == ANNOUNCEMENT_POST_SUBJECT:
+                return response[index]['id']
+        return ''
+
+    def _update_announcement_post(self, course_id, announcement):
+        """
+        Updates announcement post with announcement text composed in the exporter.
+        """
+        forum_id = self.get_announcement_forum(course_id)
+        post_id = self.get_announcement_post(forum_id)
+        params = {
+            'wsfunction': 'mod_forum_update_discussion_post',
+            'postid': post_id,
+            'message': announcement,
+        }
+        return self._post(params)
+
     def _create_forum_post(self, course_id, announcement):
         """
         Creates a new Announcement post featuring edX course details and link.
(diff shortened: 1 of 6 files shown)


integrated_channels/moodle/client.py
```diff
@@ -39,7 +40,7 @@ def inner(self, *args, **kwargs):
             # On course creation (and ONLY course creation) success,
             # Moodle returns a list of JSON objects, because of course it does.
             # Otherwise, it fails instantly and returns actual JSON.
-            return response
+            return body
```
taliaedX, line 43: ohhhh this is why my responses changed. I didn't even remember doing this. I should double check grades' responses.

integrated_channels/moodle/client.py
```diff
@@ -311,33 +312,175 @@ def _wrapped_create_course_completion(self, user_id, payload):
...
+            post = self._create_forum_post(
+                response[0].get('id'),
+                announcement
+            )
+            if post.json().get('warnings', None) or post.json().get('exception', None):
+                params = {
+                    'wsfunction': 'core_course_delete_courses',
+                    'courseids[]': response[0].get('id')
+                }
+                self._post(params)
```
taliaedX: Yes this is just getting called and hoping for the best.
This is because I don't know what to do if delete specifically fails. We need to delete the course to have any hope of getting back to a clean state, otherwise:
- If the course is in Moodle, but we don't register it, we're just going to hit errors constantly.
- If we do register it, but never made the announcement post successfully, then it's in an incomplete state on our side and nothing will ever trigger an update (even if the course itself changes, the updates for the announcement post check the existing post id, which won't exist).

So if delete fails, I really don't know what to do here. We can't do any clean up and the errors will just keep coming.

integrated_channels/moodle/client.py
```diff
@@ -311,38 +312,200 @@ def _wrapped_create_course_completion(self, user_id, payload):
...
+        response = self._wrapped_create_content_metadata(serialized_data)
+        # Response format should be [{"id":1, "shortname": "name"},{...}]
+        if response[0].get('id', None):
+            try:
+                post = self._create_forum_post(
+                    response[0].get('id'),
+                    announcement
+                )
+            except (TypeError, IndexError):
+                self._delete_content_metadata(response[0].get('id'))
```
taliaedX, line 434: I'm just calling this and hoping for the best, because if it fails to delete a course we're in a really bad state that we can't actually fix. :(

integrated_channels/moodle/client.py
binodpant: oh yes looks like body is what we want here since it's the json

integrated_channels/moodle/client.py
```diff
@@ -311,38 +312,191 @@ def _wrapped_create_course_completion(self, user_id, payload):
...
+                raise ClientError(
+                    'Moodle Client Content Metadata Creation failed to create post for course {} '
+                    'as forum could not be found'.format(
+                        response[0].get('shortname'),
+                    ),
+                    HTTPStatus.NOT_FOUND.value
+                )
+            except Exception as exc:
+                self._delete_content_metadata(response[0].get('id'))
```
binodpant, line 434: minor potential change: store `response[0].get('id')` in a var since it's used quite a few times

integrated_channels/moodle/exporters/content_metadata.py
```diff
@@ -112,3 +98,23 @@ def transform_end(self, content_metadata_item):
...
+
+    def transform_announcement(self, content_metadata_item):
+        """
+        Formats a post in a course's Announcement forum with
+        edX's course summary and enrollment link
+        """
+        full_description = content_metadata_item.get('full_description') or None
+        short_description = content_metadata_item.get('short_description') or None
+
+        if full_description and len(full_description) <= (self.LONG_STRING_LIMIT-400):
```
binodpant, line 110: interesting, perhaps worth a comment here on why the -400 ?

taliaedX: I'm actually just going to bake that into the long_string_limit var and explain the arbitrary buffer going on there.

integrated_channels/moodle/transmitters/content_metadata.py
```diff
@@ -36,8 +38,4 @@ def _prepare_items_for_transmission(self, channel_metadata_items):
         return items
 
     def _serialize_items(self, channel_metadata_items):
-        """
```
alex-sheehan-edx, line 39: What's the reason for removing this doc string?

integrated_channels/moodle/client.py
```diff
@@ -311,38 +312,191 @@ def _wrapped_create_course_completion(self, user_id, payload):
...
+        announcement = serialized_data.pop('courses[0][announcement]')
         moodle_course_id = self.get_course_id(serialized_data['courses[0][shortname]'])
         serialized_data['courses[0][id]'] = moodle_course_id
         serialized_data['wsfunction'] = 'core_course_update_courses'
-        return self._post(serialized_data)
 
-    @moodle_request_wrapper
+        response = self._wrapped_update_content_metadata(serialized_data)
+        response_text = json.loads(response.text)
+        if response_text.get('exception', None) or response_text.get('warnings', None):
```
alex-sheehan-edx, line 479: these Moodle responses pain me

integrated_channels/moodle/client.py
```diff
@@ -311,38 +312,191 @@ def _wrapped_create_course_completion(self, user_id, payload):
...
+            raise ClientError(
+                'Moodle Client Course Creation failed to create course {}'
+                .format(serialized_data['courses[0][shortname]']),
+                HTTPStatus.BAD_REQUEST.value
+            )
+
+        return 200, ''
+
+
     def update_content_metadata(self, serialized_data):
```
alex-sheehan-edx, line 471: a doc string here explaining the flow may be nice, though you can decide if necessary.

alex-sheehan-edx (commented):
A couple comments/questions but I think it's pretty good! 👍 Hope testing wraps up nicer than it started 😬

integrated_channels/moodle/client.py
```diff
@@ -311,38 +312,191 @@ def _wrapped_create_course_completion(self, user_id, payload):
...
+        """
+        params = {
+            'wsfunction': 'mod_forum_get_forum_discussions',
+            'forumid': forum_id,
+            'moodlewsrestformat': 'json',
+            'wstoken': self.token,
+        }
+        return self._post(params)
+
+    def get_announcement_post(self, forum_id):
```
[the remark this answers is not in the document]
binodpant: yeah I really hope they come up with a new api version that is actually higher up in the maturity model so we can actually read and use it

integrated_channels/moodle/client.py
```diff
@@ -311,38 +312,191 @@ def _wrapped_create_course_completion(self, user_id, payload):
...
+    def get_announcement_forum(self, course_id):
         """
-        serialized_data['wsfunction'] = 'core_course_create_courses'
-        response = self._post(serialized_data)
-        return response
+        The response is a list type, because of course it is.
+        For new courses, there should only be 1 forum and it should be Announcements
+        """
+        response = self._get_course_forum(course_id)
+        return response[0]['id']
```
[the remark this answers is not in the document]
taliaedX: Fun fact that I'll probably want to address with Ken: if you make the course format single activity, you cannot add other forums. So in our flow, we'd have 0-1 forums (and 0 throws an exception as I have unfortunately found out already, which is why the course format change was added later).

So I agree in general that we should conditionalize on the Announcements forum specifically, however with the state of things it's not necessary. Even if users change the format, unfortunately during updates it will be changed back (actually unsure how users messing with settings is going to mess everything up - another question for Ken!)

Still it shouldn't be particularly difficult to check the subject(? I forget what the response object here looks like) of the forum and isolate on that, so I will check into that.

integrated_channels/moodle/client.py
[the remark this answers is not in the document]
taliaedX: I mean moodle isn't the one using "announcement" anywhere. I specify that since there COULD (in a very theoretical sense that would likely break our integration) be other forums and definitely other posts.

When you limit to single activity, the one forum is added to the one activity which at creation time is the Announcements section. That's why I specified announcement in the names.

integrated_channels/moodle/transmitters/content_metadata.py
taliaedX: lol oops.

integrated_channels/moodle/client.py
taliaedX: Yeah I'm still not done with doc strings (part of style fixes), but will make sure this section's is more detailed.

integrated_channels/moodle/client.py
taliaedX: yeah. This _could_ be resolved in the `inner` wrapper we have as we already handle a lot of the weird types that get returned (such as list, str, int). However, I think because of some cases where we just pass `_post` all the way through, we need to have access to status codes and other response object attrs so that's why even when we go through the process of checking the body, we still return the whole response object and thus need to reparse this garbage. :/

taliaedX:
oh PS, thanks to @alex-sheehan-edx's comments making me double-check what it means to have a "single activity" format, I'm putting this on hold until we hear back from Louvain. Single activity seems to remove the ability to add other activities (activities can be a lot of different things according to the menu where you add them (lesson, survey, quiz, database(????), etc.) and thus could be extremely annoying to our customers if they actually create other activities and expect to do so for their edX courses.

(I don't know how to make that not read in a "frustrated" or "accusatory" manner, but I'm really grateful for the comment actually because it really did remind me of a potential huge issue! So thanks Alex!)

integrated_channels/moodle/client.py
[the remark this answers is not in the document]
taliaedX: After investigation, if we specify single activity or social formats, there can be only one ™️ 
But there can be other discussions and posts in said forum. What is the difference between a discussion and post? Discussion is like a reddit post and posts are exclusively replies... Hopefully.

integrated_channels/moodle/client.py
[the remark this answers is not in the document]
alex-sheehan-edx: Testing different settings seems like it's gonna be a time and a half 😬

integrated_channels/moodle/client.py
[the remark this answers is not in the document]
alex-sheehan-edx: >  Discussion is like a reddit post and posts are exclusively replies... Hopefully.

This actually helps a lot with my understanding 👍. If only **their** wording could be more clear...

integrated_channels/moodle/client.py
```diff
@@ -15,6 +15,7 @@
 from integrated_channels.integrated_channel.client import IntegratedChannelApiClient
 
 MOODLE_FINAL_GRADE_ASSIGNMENT_NAME = '(edX integration) Final Grade'
+ANNOUNCEMENT_POST_SUBJECT = 'Welcome! Click here for course details and edX link'
```
edx-abolger, line 18: How does this text get to translation team (eventually) if we are concerned about multi lingual use? Are there decorators or something elsewhere?

integrated_channels/moodle/exporters/content_metadata.py
```diff
@@ -112,3 +99,23 @@ def transform_end(self, content_metadata_item):
...
+        Formats a post in a course's Announcement forum with
+        edX's course summary and enrollment link
+        """
+        full_description = content_metadata_item.get('full_description') or None
+        short_description = content_metadata_item.get('short_description') or None
+
+        if full_description and len(full_description) <= self.LONG_STRING_LIMIT:
+            description = full_description
+        else:
+            description = short_description
```
edx-abolger, line 114: Is there always a short description to be relied on? Or does a truncation of the long description need to exist as a fallback?

integrated_channels/moodle/exporters/content_metadata.py
```diff
@@ -16,18 +16,27 @@ class MoodleContentMetadataExporter(ContentMetadataExporter):
...
+        'announcement': 'announcement',
     }
 
-    LONG_STRING_LIMIT = 1700  # Actual maximum value we can support for any individual course
+# default url length = 8000 bytes / max char size of 4 bytes = 2000 char limit
+# subtracted arbitrary char buffer for template HTML, token, and other query params we send.
+    LONG_STRING_LIMIT = 1500
     SKIP_KEY_IF_NONE = True
 
+    ANNOUNCEMENT_TEMPLATE = '<div><div style="display: inline-block">' \
```
edx-abolger, line 34: Is this mobile/responsive compatible? If not, does it need to be? If it is, what are the test criteria?

integrated_channels/moodle/client.py
taliaedX: Don't know. We haven't been utilizing any of the translation services. It's a fair question for any of the customized data we've been sending to the integrations though.

Also this PR was for the announcement post that we were unsure we were going to do, which is why it's been chilling since November. It wasn't waiting on review, but totally understand if this is for your own understanding.

integrated_channels/moodle/exporters/content_metadata.py
taliaedX: I believe in practice short_description is always present. In one of these I had done title as a fallback if none of these were present and valid, but found title to be less often present than even short_description.

taliaedX: Arbitrarily chopping up the full_description was determined to be a worse UI experience than using a short_description.

integrated_channels/moodle/exporters/content_metadata.py
taliaedX: Mobile compatibility was not part of requirements. I'm not sure how we'd test mobile specific display either tbh. Something I should probably take into account/learn in the future.

PR #1062, opened 2020-12-01
ENT-3591: update EnterpriseProxyLoginView

sameenfatima78:
JIRA: https://openedx.atlassian.net/browse/ENT-3591

**A.C:**

- enterprise/proxy-login should redirect to the IDP if there is only one IDP linked to the enterprise customer.
- if user has provided IDP in requested URL, user redirected to that IDP login Page.
- if there are multiple IDPs linked to the customer and tpa_hint is also not provided in query_string params then redirect to the normal login page.
- if there is not an IDP, then the user should be redirected to the normal login page.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -1066,15 +1066,14 @@ class EnterpriseProxyLoginView(View):
             )
             query_dict['next'] = learner_portal_base_url + '/' + enterprise_slug
 
-        tpa_hint = query_params.get('tpa_hint')
+        # Check if tpa_hint was passed as query param then redirect to that IDP login
+        # Otherwise check if enterprise is linked to one IDP, then page will be redirected there
+        tpa_hint = query_params.get('tpa_hint') or enterprise_customer.identity_providers.first().provider_id if \
+            enterprise_customer.identity_providers.count() == 1 else None
+
         if tpa_hint:
-            # If there is tpa_hint already in the query_string then no need to update it in query_dict
-            pass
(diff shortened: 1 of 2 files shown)


enterprise/views.py
```diff
@@ -1066,15 +1066,21 @@ def get(self, request):
             )
             query_dict['next'] = learner_portal_base_url + '/' + enterprise_slug
 
-        if enterprise_customer.identity_provider:
+        tpa_hint = query_params.get('tpa_hint')
+        if tpa_hint:
+            # If there is tpa_hint already in the query_string then no need to update it in query_dict
+            pass
+        elif enterprise_customer.identity_providers and enterprise_customer.identity_providers.count() == 1:
```
saleem-latif, line 1073: do we need non-null check on `enterprise_customer.identity_providers` ? shouldn't  `enterprise_customer.identity_providers.count() == 1` be enough?

saleem-latif (approved):
One question otherwise looks good 👍

enterprise/views.py
sameenfatima78: Wouldn't we need that check if `enterprise_customer.identity_providers` is `None`?

sameenfatima78: Or will this property return an empty queryset in that case?

saleem-latif: It will return an empty query set if there are no associated identity providers.

irfanuddinahmad (approved):

saleem-latif (approved):

PR #1063, opened 2020-12-01
Assessment level reporting for Canvas

alex-sheehan-edx:
Adds assessment level reporting for Canvas, which builds off of the base level [implementation](https://github.com/edx/edx-enterprise/pull/1048)

+ Refactors the canvas client's assignment creation process to not hard code the assignment's title, but instead takes an additional assignment name parameter.

Manually tested success case (see base level [PR](https://github.com/edx/edx-enterprise/pull/1048) screenshots) as well as a scenario in which partially transmitted assignments are interrupted and then attempted again (we create some but not all of the assignments on Canvas and then rerun the transmission)
**Merge checklist:**

**Post merge:**


CHANGELOG.rst
```diff
@@ -17,6 +17,7 @@ Unreleased
 -----------
 
 [3.13.1]
+--------
```
alex-sheehan-edx, line 20: lol whoops

integrated_channels/canvas/client.py
```diff
@@ -117,8 +117,70 @@ def delete_content_metadata(self, serialized_data):
...
-        Not implemented yet
+        Send assessment level learner data, retrieved by the integrated channels exporter, to Canvas in the form of
+        an assignment and submission.
         """
+        learner_data = json.loads(payload)
+        self._create_session()
+
+        # Retrieve the Canvas user ID from the user's edx email (it is assumed that the learner's Edx
+        # and Canvas emails will match).
+        canvas_user_id = self._search_for_canvas_user_by_email(user_id)
```
taliaedX, line 128: for clarity: I assume this means the edX `user_id` we're passing in here is actually an email address and not a UUID or numerical id?

integrated_channels/canvas/client.py
```diff
@@ -117,8 +117,70 @@ def delete_content_metadata(self, serialized_data):
...
+
+        # With the Canvas user ID, retrieve all courses for the user.
+        user_courses = self._get_canvas_user_courses_by_id(canvas_user_id)
+
+        # Find the course who's integration ID matches the learner data course ID
+        course_id = None
+        for course in user_courses:
+            integration_id = course['integration_id']
+            if integration_id == learner_data['courseID']:
```
taliaedX, line 137: for understanding:
I am guessing we (=the learner data export) only send 1 course at a time? (I ask because otherwise the structure here and particularly the break wouldn't work if we send more than one at a time in the payload)

integrated_channels/canvas/client.py
```diff
@@ -117,8 +117,70 @@ def delete_content_metadata(self, serialized_data):
...
+                course_id = course['id']
+                break
+
+        if not course_id:
+            raise ClientError(
```
taliaedX, line 142: Is the learner_data passed into this method limited to the catalog(s) the customer is sending to their integration?

Ie, could we trip this error way too much due to us passing a bunch of courses in that are not even in their LMS because it's in a different catalog?

integrated_channels/canvas/client.py
```diff
@@ -411,11 +477,11 @@ def _handle_canvas_assignment_retrieval(self, integration_id, course_id):
         if not assignment_id:
             assignment_creation_data = {
                 'assignment': {
-                    'name': '(Edx integration) Final Grade',
```
taliaedX, line 414: I know we used this name for final grades (and it is particularly relevant with Moodle), so:
How are we going to handle final grades in this scenario then? Are we just no longer doing a course-level grade if they use assignments?

integrated_channels/canvas/models.py
```diff
@@ -101,6 +101,9 @@ def channel_code():
     def get_learner_data_transmitter(self):
         return CanvasLearnerTransmitter(self)
 
+    def get_subsection_learner_data_exporter(self, user):
+        return CanvasLearnerExporter(user, self)
```
taliaedX, line 105: given this just returns the same thing as `get_learner_data_exporter` what is the point of defining this here? I don't see it specifically called (at least not elsewhere in this PR)

taliaedX (commented):
I have a few questions thus far. Will dive in a bit more maybe running locally. Have to get a grade up and running locally though which...eep.

integrated_channels/canvas/client.py
alex-sheehan-edx: Yeah, this is an email here. Through each of the specific integrated channels' exporters, we designate what we want the learner data's `remote user id` to be, which should the source of data that links an edx learner to an external learner. In this case, we use email. Part of the reason why the parameter is user_id is because of the `remote_user_id` label being passed to the client. But yeah.. both user ID and email.

integrated_channels/canvas/client.py
alex-sheehan-edx: yeah, the exports are handled by enrollment, so it's one at a time.

integrated_channels/canvas/client.py
alex-sheehan-edx: > limited to the catalog(s) the customer is sending to their integration

The ultimate truth/source of `assessment level course learner data` being passed into this method comes from an enterprise enrollment giving us the course run ID as well as user ID. We use that information to query the gradebook view in platform which returns all the necessary learner data related to each subsection within the class. We then export and pass each subsection worth of data to the client which posts it to Canvas, one at a time.

So, to your question of is the learner data limited to the catalogs that the customer has designated? ...I would say no? I'm hesitant because while the data being passed in does not consider customer catalogs when retrieving it, (only the gradebook view which retrieves the user's grade data from our system), I don't think an enterprise learner would be enrolled in courses not contained within the customer's catalog... This obviously differs from how we export and post `course metadata` as that comes from the catalogs. Currently I am looking into and am slightly uneasy about our logic on posting both courses and course runs as it affects the way we match a learner data `course ID` to the Canvas' integration ID (as the learner data course ID will almost certainly always be a course run ID, where as I'm not entirely sure if the Canvas integration ID's will be course run ID's, course ID's or a mix of both). 

I can see the following cases (with varying degrees of possibilities of them actually happening IRL) causing course_id to not be found here:
1) Canvas changes their API response such that their `integration id` field no longer matches our expected set ID that we created when posting the course or there is a bug introduced into our code such that integration ID's don't match the gradebook view's `course ID` 
2) The user is enrolled in a Canvas course that has an edX `course ID`, rather than an edX `course run ID` as the integration ID, and our learner data `course ID` is the edX `course run ID`.
3) Something is horribly wrong and there just isn't a Canvas course that correlates to the edX course linked to the exported learner data and the enterprise learner has still somehow enrolled, started and completed a subsection of the edX course.

I don't think any of these are likely.. except for the `course ID` vs `course run ID` situation, but that is also dependent on how we/the Canvas admins are handling courses on their end. (Do they want one Canvas course to represent all edX courses, or do they one a Canvas course per course run that they're enrolling students in). This is a talk that I am trying to schedule with George and Ken to get better context on.

- [where we break down the **learner data export** by enrollment](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/exporters/learner_data.py#L113-L121)
- [where we use the course ID and user ID to get the learner data from the gradebook API](https://github.com/edx/edx-enterprise/blob/2931d8eab3f70539139a676468843a89cf4d09c1/integrated_channels/integrated_channel/exporters/learner_data.py#L558)
- [where we pass individual subsections of data back to the transmitter, and then subsequently to the client](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/exporters/learner_data.py#L127-L139)

integrated_channels/canvas/client.py
alex-sheehan-edx: My hope was to create the assessment level reporting flow completely separated from course completion level reporting, while also leaving course level completely in tact. As it turns out, most of these hand shakes with Canvas are similar/identical whether doing assessment level or completion level (we create an assignment and post a submission to the assignment). So I tried and create common handlers that would take care of shared behaviors. [we are now passing the name of the "assignment"](https://github.com/edx/edx-enterprise/blob/771db5b8391249e57313966e3f0432d4a6e24fe0/integrated_channels/canvas/client.py#L418) (name of the edx subsection or the name "final grade") to the helper function. [Here is where we're passing the "final grade" name into the assignment creation process](https://github.com/edx/edx-enterprise/blob/771db5b8391249e57313966e3f0432d4a6e24fe0/integrated_channels/canvas/client.py#L208) so both should still work!

alex-sheehan-edx: This also has the added benefit of being more easily testable without having to stub out all previous steps in the process! We can just call the helper method and test that it behaves properly.

integrated_channels/canvas/models.py
alex-sheehan-edx: 😬 Good catch! This just straight up isn't needed. I had thought it would be needed early in development while I had it in my head that I would create a completely different assessment level exporter class rather than just add assessment level exporter capabilities to the existing learner data exporter. You can see now I'm just returning the learner data exporter [here](https://github.com/edx/edx-enterprise/blob/771db5b8391249e57313966e3f0432d4a6e24fe0/integrated_channels/integrated_channel/models.py#L178). Will remove 👍

integrated_channels/canvas/client.py
taliaedX: Ah okay. That answered my would-be follow up question of "why not label it `email`?" so thank you for that great explanation! :D

integrated_channels/canvas/client.py
taliaedX: I'm guessing this was the comment you were telling me about lol. It's greatly informative and does bring up valid concerns that we should think about and maybe address with product. I agree that courses being sent to our integrations when we only ever associate grades with course runs is a bit of a problem.

To clarify the point I was initially bringing up: `I don't think an enterprise learner would be enrolled in courses not contained within the customer's catalog`
The scenario I was thinking of was not that the catalog was missing, but rather the scenario where a customer has multiple catalogs, but has only designated a subset to be sent to their integrated LMS. This is a really doubtful scenario and with the way I've configured the front end, the customer wouldn't be able to set this up themselves (I removed the catalog field, so it just sends all of their catalogs), but I just wanted to raise this possible edge case. We can also ignore it, since it is very unlikely to be hit - that is a perfectly good option.

integrated_channels/canvas/client.py
taliaedX: Nice. I missed that addition.

binodpant:
is this still correct? 

> TODO
Add unit testing (in progress)

alex-sheehan-edx:
> is this still correct?
> 
> > TODO
> > Add unit testing (in progress)

ack, nope 😄 I did a bunch of work to add unit testing!

taliaedX (approved):
Oh I think I'm good on this. I'd love if there was a way to test it in stage or even prod prior to it being automatically triggered, but I don't know if that's possible with signals.

binodpant (approved):
discussed with Alex, looks like there is no alternative that helps re use existing methods right now for the learner data course fetch

PR #1067, opened 2020-12-02
Both list_route and detail_route removed in drf = 3.10

awais786:
edx-platform gives error while upgrading drf version to 3.10 due to enterprise.

https://openedx.atlassian.net/browse/BOM-2130

This issue is caught by `make docs` inside edx-platform. Tests failed to capture this issue.

https://www.django-rest-framework.org/community/3.8-announcement/#action-decorator-replaces-list_route-and-detail_route

Both list_route and detail_route are now pending deprecation. They will be deprecated in 3.9 and removed entirely in 3.10.

The new action decorator takes a boolean detail argument.

Replace detail_route uses with @action(detail=True).
Replace list_route uses with @action(detail=False).

### DIFF
--- a/tests/test_enterprise/api/test_views.py
+++ b/tests/test_enterprise/api/test_views.py
@@ -77,18 +77,15 @@ ENTERPRISE_CATALOGS_CONTAINS_CONTENT_ENDPOINT = reverse(
     kwargs={'pk': FAKE_UUIDS[1]}
 )
 ENTERPRISE_CATALOGS_COURSE_ENDPOINT = reverse(
-    # pylint: disable=anomalous-backslash-in-string
-    r'enterprise-catalogs-courses/(?P<course-key>[^/+]+(/|\+)[^/+]+)',
+    'enterprise-catalogs-course-detail',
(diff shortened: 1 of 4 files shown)


awais786:
@irfanuddinahmad Please review this PR. It would be great if some one from your team can hit these end points locally. Thanks !!!

saleem-latif (approved):
Looks Good 👍  
Please also add an entry in `CHANGELOG.rst`.

tests/test_enterprise/api/test_views.py
```diff
@@ -77,18 +77,15 @@
     kwargs={'pk': FAKE_UUIDS[1]}
 )
 ENTERPRISE_CATALOGS_COURSE_ENDPOINT = reverse(
-    # pylint: disable=anomalous-backslash-in-string
-    r'enterprise-catalogs-courses/(?P<course-key>[^/+]+(/|\+)[^/+]+)',
+    'enterprise-catalogs-course-detail',
```
awais786, line 80: I have to make this change otherwise tests fails to find-out the end point because endpoint is expecting `(?P<course-key>[^/+]+(/|\\+)[^/+]+)` note escape backslash in regex.

PR #1069, opened 2020-12-03
ENT-3781: Empty sequence bugfix

sameenfatima78:
This PR fixes the bug that came in catalog api `max() arg is an empty sequence`. The bug comes when all of the course runs end dates are `null`.

**JIRA**: https://openedx.atlassian.net/browse/ENT-3781

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -930,7 +930,11 @@ def get_last_course_run_end_date(course_runs):
     """
     latest_end_date = None
     if course_runs:
-        latest_end_date = max(course_run['end'] for course_run in course_runs if course_run['end'] is not None)
+        try:
+            latest_end_date = max(course_run.get('end') for course_run in course_runs if
+                                  parse_datetime_handle_invalid(course_run.get('end')) is not None)
+        except ValueError:
+            latest_end_date = None
     return latest_end_date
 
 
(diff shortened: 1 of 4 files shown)


moconnell1453 (approved):

enterprise/utils.py
```diff
@@ -930,7 +930,10 @@ def get_last_course_run_end_date(course_runs):
     """
     latest_end_date = None
     if course_runs:
-        latest_end_date = max(course_run['end'] for course_run in course_runs if course_run['end'] is not None)
+        try:
+            latest_end_date = max(course_run['end'] for course_run in course_runs if course_run['end'] is not None)
```
irfanuddinahmad, line 934: Do we need to check for isinstance(course_run['end'], datetime.date)?

irfanuddinahmad (approved):

enterprise/utils.py
sameenfatima78: I think all dates are saved in the same format but let me confirm that.

sameenfatima78: I see the discovery api is showing all dates in the same format.

moconnell1453 (approved):

PR #1072, opened 2020-12-04
adding blackboard support for assessment level reporting

alex-sheehan-edx:
Implementation of assessment level reporting. Largely based off of Canvas implementation. 

**Merge checklist:**

**Post merge:**


integrated_channels/blackboard/models.py
```diff
@@ -111,6 +111,112 @@ def get_learner_data_transmitter(self):
...
+    course_id = models.CharField(
+        max_length=255,
+        blank=False,
+        null=False,
+        help_text="The course run's key which is used to uniquely identify the course for blackboard."
+    )
+
+    subsection_id = models.CharField(
+        max_length=255,
+        blank=False,
```
taliaedX, line 143: are we *always* going to send subsections?

I ask because subsection_id and subsection_name both have `blank=False` which would fail validation if they aren't included, but based on the rest of the logic, we seem to allow for the prospect of not having a subsection grade all the time.

taliaedX (commented):
I have one question on database table structure, but otherwise LGTM.

integrated_channels/blackboard/models.py
alex-sheehan-edx: Yikes, yeah we're not always going to be sending subsections, this is shared between course completion and assessment level reporting. Idk what I was thinking...

alex-sheehan-edx: or.... are they shared? Hold on, I think I confused myself. Investigating.

alex-sheehan-edx: Ok sorry- my brain was thinking of the base `LearnerDataTransmissionAudit`. It being "shared" as in it's the base 🤦, and the base has the integration ID and name as optional.

alex-sheehan-edx: Having said that, both the subsection ID and subsection name are a part of the subsection grade data retrieved from platform's gradebook API and that is the only place where we create these audits. So I guess we _**should**_ be always sending those pieces of data with this particular audit?

taliaedX: ah yeah I didn't pay close attention to the class name (I missed "Assessment" in it). Yeah given the only reason this would be checked against/run is for assessment reporting then yes I think it's quite reasonable to expect subsection id and name.

taliaedX (approved):
Updating to ✅  (Approved)

integrated_channels/blackboard/client.py
```diff
@@ -248,20 +265,6 @@ def _validate_full_course_metadata(channel_metadata_items):
                 HTTPStatus.NOT_FOUND.value
             )
 
-    def _resolve_blackboard_course_id(self, external_id):
```
alex-sheehan-edx, line 251: This seemed more like a helper method rather than a validation method so I moved it closer to the other helper methods.

integrated_channels/blackboard/client.py
```diff
@@ -154,8 +154,37 @@ def delete_content_metadata(self, serialized_data):
 
     def create_assessment_reporting(self, user_id, payload):
         """
-        Not implemented yet
+        Post a learner's subsection assessment grade to the integrated Blackboard course.
+
+        Parameters:
+        -----------
+            user_id (str): The shared email between a user's edX account and Blackboard account
```
binodpant, line 161: interesting I thought we used the name user_email in another function.. is this better named user_email?

tests/test_integrated_channels/test_blackboard/test_client.py
```diff
@@ -42,9 +42,11 @@ def setUp(self):
         )
         self.user_email = 'testemail@example.com'
         self.course_id = 'course-edx+{}'.format(str(random.randint(100, 999)))
+        self.course_subsection_id = 'subsection_id:{}'.format(str(random.randint(100, 999)))
         self.blackboard_course_id = random.randint(100, 999)
         self.blackboard_user_id = random.randint(100, 999)
         self.blackboard_grade_column_id = random.randint(100, 999)
+        self.blackboard_grade_column_name = 'ayylmao'
```
binodpant, line 49: nice column name, lmao!

integrated_channels/blackboard/client.py
alex-sheehan-edx: https://github.com/edx/edx-enterprise/pull/1063#discussion_r533737416

Had this discussion with @taliaedX on the Canvas assessment work. The "ID" here will always be an email, but in the overall integrated channels, this is not email specific. We designate the `remote_user_id` to be whatever we want in the specific channel's transmitter. Because we pass `remote_user_id` into the client from the base transmitter, I decided to keep this as `user_id`

binodpant (approved):
only a minor comments, but looks good to me! Great job on the tests @alex-sheehan-edx

integrated_channels/blackboard/client.py
binodpant: makes sense, got it

PR #1081, opened 2020-12-11
Success Factors Config API endpoints

taliaedX:
Adds SAP Success Factors API endpoints to the new Integrated Channels API.
(This is going to need conflict fixes once Canvas is merged)

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/api/v1/sap_success_factors/urls.py
+++ b/integrated_channels/api/v1/sap_success_factors/urls.py
@@ -1,0 +1,12 @@
+"""
+    url mappings for integrated_channels/v1/sap_success_factors/
+"""
+
+from rest_framework import routers
+
+from .views import SAPSuccessFactorsConfigurationViewSet
+
+app_name = 'sap_success_factors'
+router = routers.DefaultRouter()
+router.register(r'configuration', SAPSuccessFactorsConfigurationViewSet, basename="configuration")
+urlpatterns = router.urls
--- a/integrated_channels/api/v1/sap_success_factors/views.py
+++ b/integrated_channels/api/v1/sap_success_factors/views.py
@@ -1,0 +1,40 @@
+"""
+    Viewset for integrated_channels/v1/sap_success_factors/
+"""
+from edx_rbac.mixins import PermissionRequiredMixin
+from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
+from rest_framework import permissions, viewsets
+from rest_framework.authentication import SessionAuthentication
+
+from django.core.exceptions import PermissionDenied
+from django.http import Http404
+
+from enterprise.models import EnterpriseCustomerUser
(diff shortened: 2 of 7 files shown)


integrated_channels/api/v1/sap_success_factors/urls.py
```diff
@@ -0,0 +1,12 @@
+"""
+    url mappings for integrated_channels/v1/sap_success_factors/
+"""
+
+from rest_framework import routers
+
+from .views import SAPSuccessFactorsConfigurationViewSet
+
+app_name = 'sap_success_factors'
+router = routers.DefaultRouter()  # pylint: disable=invalid-name
```
raq929, line 10: I'm sure it's necessary, but why are we having to disable the invalid-name rule?

integrated_channels/api/v1/sap_success_factors/views.py
```diff
@@ -0,0 +1,40 @@
...
+class SAPSuccessFactorsConfigurationViewSet(PermissionRequiredMixin, viewsets.ModelViewSet):
+    authentication_classes = (JwtAuthentication, SessionAuthentication,)
+    permission_classes = (permissions.IsAuthenticated,)
+    serializer_class = SAPSuccessFactorsConfigSerializer
+    permission_required = 'enterprise.can_access_admin_dashboard'
+
+    def get_queryset(self):
+        enterprise_customer_users = EnterpriseCustomerUser.objects.filter(user_id=self.request.user.id)
+        if enterprise_customer_users:
+            enterprise_customers_uuids = [str(enterprise_customer_user.enterprise_customer.uuid) for
```
raq929, line 27: Wait, I'm confused, how can an enterprise customer have multiple UUIDs?

raq929 (approved):
LGTM, just some questions for my edification

integrated_channels/api/v1/sap_success_factors/views.py
taliaedX: an enterprise customer (ie just an enterprise) would not have multiple uuids.
An enterprise customer _user_  can belong to multiple enterprises (see ECS, devs testing things, etc.). It would only affect those of us internal more than likely though.

This code is taking the user id and finding the enterprise UUIDs that the user is associated to, which can be plural.

integrated_channels/api/v1/sap_success_factors/urls.py
taliaedX: Upon running tox with this comment removed it doesn't actually appear to be necessary.
This was only just copypasta from other places.
It's likely either this rule applied to edx-platform but somehow not edx-enterprise OR whatever rule we used when this was first implemented from wherever I copied it is either explicitly ignored or no longer applies. I'll just try committing the removal.

taliaedX:
This is good to go except for adding a version/changelog and squashing. Probably waiting until after vacation to do that.

PR #1086, opened 2020-12-16
refactor revoking a course enrollment, create task

kellmarie:
**Overview**
As part of the work to expire a subscription plan we need to change active licensed course enrollments to be either in the audit track if available, or the learner gets unenrolled if it isn't.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -378,10 +378,11 @@ class LicensedEnterpriseCourseEnrollmentViewSet(EnterpriseWrapperApiViewSet):
 
     queryset = models.LicensedEnterpriseCourseEnrollment.objects.all()
     serializer_class = serializers.LicensedEnterpriseCourseEnrollmentReadOnlySerializer
+    REQ_EXP_LICENSE_UUIDS_PARAM = 'expired_license_uuids'
 
-    class RevocationStatus:
+    class EnrollmentTerminationStatus:
         """
-        Defines statuses related to enrollment states during the revocation process.
+        Defines statuses related to enrollment states during the course unenrollment process.
         """
         COURSE_COMPLETED = 'course already completed'
         MOVED_TO_AUDIT = 'moved to audit'
@@ -446,7 +447,7 @@ class LicensedEnterpriseCourseEnrollmentViewSet(EnterpriseWrapperApiViewSet):
             for enrollment in licensed_enrollments
         }
 
-    def _revoke_enrollment(self, enrollment_api_client, enterprise_enrollment, course_overview):
+    def _terminate_enrollment(self, enrollment_api_client, enterprise_enrollment, course_overview):
         """
         Helper method that switches the given enrollment to audit track, or, if
         no audit track exists for the given course, deletes the enrollment.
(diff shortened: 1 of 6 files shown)


enterprise/api/v1/views.py
```diff
@@ -577,17 +574,68 @@ def license_revoke(self, request, *args, **kwargs):
...
+    @permission_required('enterprise.can_enroll_learners')
+    def licensed_enrollments_expiration(self, request):
+        """
+        Changes the mode for licensed enterprise course enrollments to the "audit" course mode,
+        or unenroll the user if no audit mode exists for each expired license uuid
+        """
+        expired_license_uuids = get_request_value(request, self.REQ_EXP_LICENSE_UUIDS_PARAM, '')
+
+        if not expired_license_uuids:
+            return Response('expired license uuids must be provided', status=status.HTTP_400_BAD_REQUEST)
```
adamstankiewicz, line 601: nit: maybe include `self.REQ_EXP_LICENSE_UUIDS_PARAM` in the response so it's clear exactly which param is needed.

tests/test_enterprise/api/test_views.py
```diff
@@ -105,6 +105,9 @@
 LICENSED_ENTERPISE_COURSE_ENROLLMENTS_REVOKE_ENDPOINT = reverse(
     'licensed-enterprise-course-enrollment-license-revoke'
 )
+EXPIRED_LICENSED_ENTERPISE_COURSE_ENROLLMENTS_ENDPOINT = reverse(
```
adamstankiewicz, line 108: nit: `EXPIRED_LICENSED_ENTERPISE_COURSE_ENROLLMENTS_ENDPOINT` is missing the second "R" in `ENTERPISE`

adamstankiewicz (approved):
Two small nits but otherwise looks good to me!

enterprise/api/v1/views.py
```diff
@@ -577,17 +574,71 @@ def license_revoke(self, request, *args, **kwargs):
...
+        )
+        indexed_overviews = {overview.get('id'): overview for overview in course_overviews}
+
+        any_failures = False
+        for licensed_enrollment in licensed_enrollments:
+            enterprise_course_enrollment = licensed_enrollment.enterprise_course_enrollment
+            course_id = enterprise_course_enrollment.course_id
+            course_overview = indexed_overviews.get(course_id)
+            try:
+                termination_status = self._terminate_enrollment(
```
kellmarie, line 622: I think there's a limit to the number of API calls we can make at a time. If a subscription is going to have a max 5k learners and each learner could have multiple enrollments then it might make sense to set a delay for every few hundred enrollments

iloveagent57: I would just make this endpoint return an error if `expired_license_uuids` is longer than some constant, 100 is probably fine.

iloveagent57: I'm also interested to see how badly `_terminate_enrollment()` slows us down when we try to expire 100s of enrollments! (I'm not suggesting you change anything, here, just expressing genuine curiosity)

iloveagent57 (commented):
This is looking good, only the two suggestions.

PR #1091, opened 2020-12-16
ENT-3787: Swaps to GitHub actions

jlajoie:
**Merge checklist:**

**Post merge:**


UsamaSadiq:
Hi @jlajoie, for jasmine tests after updating the jasmine command the tests started failing instead of silently passing so now I've added the jasmine tests in allow_failures in Travis. You should keep the jasmine test in the GitHub actions and mark it as `continue-on-error=True`.

Makefile
```diff
@@ -97,7 +97,7 @@ upgrade: check_pins	## update the requirements/*.txt files with the latest packa
 	$(PIP_COMPILE) -o requirements/doc.txt requirements/doc.in
 	$(PIP_COMPILE) -o requirements/test.txt requirements/test.in
 	$(PIP_COMPILE) -o requirements/dev.txt requirements/dev.in
-	$(PIP_COMPILE) -o requirements/travis.txt requirements/travis.in
+	$(PIP_COMPILE) -o requirements/github.txt requirements/github.in
```
UsamaSadiq, line 100: Just a suggestion, In all other services we use `github.in` for github dependencies and have switched the name `travis.in` to `ci.in` & `ci.txt`.

taliaedX (approved):
I can't comment on whether the exact commands are the right syntax or not but the steps themselves and alterations from Travis to GitHub actions looks good to me.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,37 @@
...
+    branches: [master]
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.5']
+        django-version: ['django22']
+        toxenv: ['django22', 'quality', 'docs', 'pii-annotations']
```
UsamaSadiq, line 16: I think you should add jasmine tests too in the tox environments and mark them as allow_failures.

jlajoie: Added it as a ticket after talking with George about doing this as follow up work. Would rather not have things running that have no intention of passing. Can be tracked at https://openedx.atlassian.net/browse/ENT-3880

PR #1093, opened 2020-12-20
ENT-3884: refactor ``handle_user_post_save`` signal to account for admin permissions

adamstankiewicz:
https://openedx.atlassian.net/browse/ENT-3884

Related PR: https://github.com/edx/frontend-app-admin-portal/pull/421

Previously, admin permissions were granted to the user based on the creation/update of an `EnterpriseCustomerUser` record through a Django signal. However, if a user is already linked to an `EnterpriseCustomer`, there is no creation/update of an `EnterpriseCustomerUser` record so the admin permissions were never granted on their next login.

Instead, the granting of admin permissions was moved to coincide whenever a user logs in instead so that we will always grant the correct access roles on each login.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -1110,14 +1110,12 @@ class PendingEnterpriseCustomerUser(TimeStampedModel):
         app_label = 'enterprise'
         ordering = ['created']
 
-    @classmethod
-    def link_pending_enterprise_user(cls, pending_ecu, is_user_created, user):
+    def link_pending_enterprise_user(self, user, is_user_created):
         """
         Link a PendingEnterpriseCustomerUser to the appropriate EnterpriseCustomer by
         creating a EnterpriseCustomerUser record.
 
         Arguments:
-            pending_ecu: a PendingEnterpriseCustomerUser instance
             is_user_created: a boolean whether the User instance was created or updated
             user: a User instance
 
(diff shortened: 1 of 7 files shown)


kellmarie (approved):

enterprise/models.py
```diff
@@ -1107,6 +1110,74 @@ class Meta:
...
         ordering = ['created']
 
+    @classmethod
+    def link_pending_enterprise_user(cls, pending_ecu, is_user_created, user):
+        """
+        Link a PendingEnterpriseCustomerUser to the appropriate EnterpriseCustomer by
+        creating a EnterpriseCustomerUser record.
+
+        Arguments:
+            pending_ecu: a PendingEnterpriseCustomerUser instance
```
brittneyexline, line 1120: why pass in pending_ecu as an argument when this method is in its class? why not just call this method on the pending_ecu?

brittneyexline (approved):
i think everything here makes sense, but now i understand why this is somewhat risky. i would recommend outlining a few manual test cases that can be performed in stage/prod to smoke test with. not sure if there is any particular dashboard that would be worth monitoring to see that there is no impact? in any case, i trust your judgement if anything goes awry 😄

adamstankiewicz:
> i think everything here makes sense, but now i understand why this is somewhat risky. i would recommend outlining a few manual test cases that can be performed in stage/prod to smoke test with. not sure if there is any particular dashboard that would be worth monitoring to see that there is no impact? in any case, i trust your judgement if anything goes awry 😄

Most manual test cases are nicely outlined in the ticket: https://openedx.atlassian.net/browse/ENT-3884

I will document a couple other edge cases to verify in the PR description shortly 👍 Not sure if there is a dashboard worth monitoring here; I will try to see if there is one and monitor it once this is on stage/prod.

PR #1094, opened 2020-12-21
ENT-3764 | Modify the learner portal enterprise_course_enrollments en…

iloveagent57:
…dpoint to include an `is_enrollment_active` key that indicates the status of the enterprise enrollment's related student.CourseEnrollment.  Allow the endpoint to optionally accept an `?is_active` query param, so that clients may request only active (or only inactive) enrollments from it.

Partially addresses https://openedx.atlassian.net/browse/ENT-3764

So now, when requesting http://localhost:18000/enterprise_learner_portal/api/v1/enterprise_course_enrollments/?enterprise_id=378d5bf0-f67d-4bf7-8b2a-cbbc53d0f772 you'll get something like
```
[
    {
        "certificate_download_url": null,
        "course_run_id": "course-v1:edX+DemoX+Demo_Course",
        "course_run_status": "in_progress",
        "start_date": "2013-02-05T06:00:00Z",
        "end_date": null,
        "display_name": "edX Demonstration Course",
        "course_run_url": "http://localhost:18000/courses/course-v1:edX+DemoX+Demo_Course/course/",
        "due_dates": [],
        "pacing": "instructor",
        "org_name": "edX",
        "is_revoked": false,
        "is_enrollment_active": true
    }
]
```

...and adding `&is_active=false` to the above url will get you an empty list in response.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/tests/test_enterprise_learner_portal/api/test_views.py
+++ b/tests/test_enterprise_learner_portal/api/test_views.py
@@ -5,6 +5,7 @@ Tests for the EnterpriseCourseEnrollmentview of the enterprise_learner_portal ap
 
 import uuid
 
+import ddt
 import mock
 from pytest import mark
 from six.moves.urllib.parse import urlencode  # pylint: disable=import-error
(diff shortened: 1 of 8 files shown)


tests/test_enterprise_learner_portal/api/test_views.py
```diff
@@ -16,8 +17,14 @@
 from enterprise.utils import NotConnectedToOpenEdX
 from test_utils import factories
 
+FAKE_SERIALIZED_DATA = [
```
chavesj, line 20: DRY nit: each of these course enrollment data objects can be broken out into their own variables (e.g. `ACTIVE_ENROLLMENT`, `INACTIVE_ENROLLMENT`) that way if they need to be expanded upon it can be done in one place. Also nitty - I feel like this variable name can be more descriptive but I'm not going to die on that hill 😂 up to you

chavesj (commented):
One small nit, besides that looks great

chavesj (approved):
noice

kellmarie (approved):

PR #1096, opened 2020-12-23
fix: support pending learner/admin users tied to more than 1 Enterprise

adamstankiewicz:
https://openedx.atlassian.net/browse/ENT-3896

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -708,7 +708,8 @@ def get_link_by_email(self, user_email):
             pass
 
         try:
-            return PendingEnterpriseCustomerUser.objects.get(user_email=user_email)
+            # return the first element in case of admin/learner with multiple pending enterprise associations.
+            return PendingEnterpriseCustomerUser.objects.filter(user_email=user_email).first()
```
adamstankiewicz, line 712: This is OK to return just the first element if there are multiple. This function is used downstream to determine if the user has any existing pending links to any customer.

tests/test_enterprise/api/test_views.py
```diff
@@ -431,7 +431,7 @@ def test_post_enterprise_customer_user_logged_out(self):
 
     @ddt.data(
         {'is_staff': True, 'user_exists': True, 'ecu_exists': True, 'pending_ecu_exists': False, 'status_code': 204},
-        {'is_staff': True, 'user_exists': True, 'ecu_exists': False, 'pending_ecu_exists': True, 'status_code': 204},
+        {'is_staff': True, 'user_exists': True, 'ecu_exists': False, 'pending_ecu_exists': True, 'status_code': 201},
```
adamstankiewicz, line 434: As the view is implemented, if we post to the pending enterprise customer user API for a user that already exists and has a PendingEnterpriseCustomerUser record, we opt to link the user by creating an `EnterpriseCustomerUser` record for that user/customer.

As such, we would expect a 201 in this case, not a 204.

enterprise/models.py
edx-abolger: Then perhaps make it a boolean?

adamstankiewicz: We do expect it to the function to return either a `EnterpriseCustomerUser` or `PendingEnterpriseCustomerUser` object as we refer to its `enterprise_customer` relationship, so returning a boolean here wouldn't quite make sense.

enterprise/signals.py
```diff
@@ -63,27 +64,21 @@ def handle_user_post_save(sender, **kwargs):  # pylint: disable=unused-argument
     if user_instance is None:
         return  # should never happen, but better safe than 500 error
 
-    try:
-        pending_ecu = PendingEnterpriseCustomerUser.objects.get(user_email=user_instance.email)
-    except PendingEnterpriseCustomerUser.DoesNotExist:
-        pending_ecu = None
+    pending_ecus = PendingEnterpriseCustomerUser.objects.filter(user_email=user_instance.email)
```
bbaker6225, line 67: If we’re only looking up by email, no need to index together with customer. I misunderstood that we’d be using both in the lookup

bbaker6225 (approved):
One small comment, but otherwise lgtm

enterprise/signals.py
adamstankiewicz: Most other places where we call `PendingEnterpriseCustomerUser.objects.get` already supplied both a `user_email` and an `enterprise_customer`. Would it make sense to keep the index together with customer for those cases?

bbaker6225: Yep that makes sense. You can also add user email on it’s own for this case like in the docs example:

class Meta:
        indexes = [
            models.Index(fields=['last_name', 'first_name']),
            models.Index(fields=['first_name'], name='first_name_idx'),
        ]

adamstankiewicz: Ah, got it. Cool, will update!

PR #1101, opened 2021-01-05
ENT-3764 | Listen for student.CourseEnrollment unenrollment signal an…

iloveagent57:
…d delete associated EnterpriseCourseEnrollment record (we will have a historical record of the deletion).

https://openedx.atlassian.net/browse/ENT-3764

How to manually test:
* Be enrolled in a course as an enterprise learner.
* Make sure there's an `EnterpriseCourseEnrollment` associated with your course.  Find that enrollment in the Django admin e.g. http://localhost:18000/admin/enterprise/enterprisecourseenrollment/1/change/
* From the LMS dashboard, unenroll from the course (there's a little gear icon on the dashboard course card, under that is an unenroll option).
* Click the button that asks if you _really_ want to unenroll.
* Notice a message in the LMS logs saying the EnterpriseCourseEnrollment was deleted.
* Notice in the Django admin that the enterprise enrollment is gone.
* Confirm there's still a historical record of the enterprise enrollment in the database, e.g.`select * from enterprise_historicalenterprisecourseenrollment where id = 10\G`

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/signals.py
+++ b/enterprise/signals.py
@@ -15,6 +15,7 @@ from enterprise.decorators import disable_for_loaddata
 from enterprise.models import (
     EnterpriseAnalyticsUser,
     EnterpriseCatalogQuery,
+    EnterpriseCourseEnrollment,
     EnterpriseCustomer,
     EnterpriseCustomerBrandingConfiguration,
     EnterpriseCustomerCatalog,
@@ -35,8 +36,10 @@ from enterprise.utils import (
 
 try:
     from common.djangoapps.student.models import CourseEnrollment
+    from common.djangoapps.student.signals import UNENROLL_DONE
 except ImportError:
     CourseEnrollment = None
+    UNENROLL_DONE = None
 
 logger = getLogger(__name__)  # pylint: disable=invalid-name
 _UNSAVED_FILEFIELD = 'unsaved_filefield'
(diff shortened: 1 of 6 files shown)


edx-abolger (commented):
I'll leave formal approval for a more knowledgeable engineer but LGTM. (I mostly checked it out as a self contained signals PR)

enterprise/signals.py
```diff
@@ -347,6 +350,35 @@ def create_enterprise_enrollment_receiver(sender, instance, **kwargs):     # pyl
...
+
+    and deletes the ``EnterpriseCourseEnrollment`` associated with the modified
+    ``student.CourseEnrollment``.
+    """
+    course_enrollment = kwargs['course_enrollment']
+
+    try:
+        user_id = course_enrollment.user.id
+        course_id = course_enrollment.course_id
+        enterprise_customer_user = EnterpriseCustomerUser.objects.get(user_id=user_id)
```
adamstankiewicz, line 366: `EnterpriseCustomerUser.objects.get(user_id=user_id)` assumes a user will only be linked to a single enterprise. If a user is linked to more than 1 enterprise, there will be several `EnterpriseCustomerUser` records for the specified user id.

Could we retrieve the `EnterpriseCourseEnrollment` records without needing to rely on a specific `EnterpriseCustomerUser` record?

```python
enterprise_course_enrollments = EnterpriseCourseEnrollment.objects.filter(
  enterprise_customer_user__user_id=user_id,
  course_id=course_id,
)
```

iloveagent57: Mmmmm, good point.

iloveagent57: Hmm, but what if they're enrolled in the same course in multiple enterprises, but only want to unenroll in their currently selected enterprise?

adamstankiewicz: @iloveagent57 If they're enrolled in the same course through multiple enterprises, wouldn't the user still have a single non-Enterprise (aka "normal") course enrollment record? As in, if that user goes to the B2C dashboard, they wouldn't see the course listed twice, correct?

If that assumption is true and the user unenrolls from that course, I feel like I'd expect the unenrollment to apply to both enterprise course enrollments.

If we only unenrolled for their currently selected enterprise, would we get into a situation where the user lost their actual course enrollment but we'd still have a `EnterpriseCourseEnrollment` record for the now-unenrolled course?

adamstankiewicz (approved):
✅ Looks good! Though, I'm wondering if we might want to add an explicit test case for when there is more than one EnterpriseCourseEnrollment that gets deleted to catch future regressions if/when things change in the future.

PR #1111, opened 2021-01-11
fix: always return "completed" course run status if learner has a passing certificate

adamstankiewicz:
This PR modifies the `get_course_run_status` util function such that if a learner has a passing certificate, it will always return the course as "completed". This results in the following cases:

1. Enrollment saved for later --> `CourseRunProgressStatuses.SAVED_FOR_LATER`
1. Course has started but not ended --> `CourseRunProgressStatuses.IN_PROGRESS`
1. Course has ended --> `CourseRunProgressStatuses.COMPLETED`
1. Learner has passing certificate --> `CourseRunProgressStatuses.COMPLETED`
1. Course has not started --> `CourseRunProgressStatuses.UPCOMING`

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise_learner_portal/utils.py
+++ b/enterprise_learner_portal/utils.py
@@ -21,45 +21,32 @@ class CourseRunProgressStatuses:
 
 def get_course_run_status(course_overview, certificate_info, enterprise_enrollment):
     """
-    Get the progress status of a course run, given the state of a user's certificate in the course.
+    Get the status of a course run, given the state of a user's certificate in the course.
 
-    In the case of self-paced course runs, the run is considered completed when either the course run has ended
-    OR the user has earned a passing certificate 30 days ago or longer.
+    A run is considered "complete" when either the course run has ended OR the user has earned a
+    passing certificate.
 
     Arguments:
         course_overview (CourseOverview): the overview for the course run
-        certificate_info: A dict containing the following keys:
+        certificate_info: A dict containing the following key:
             ``is_passing``: whether the  user has a passing certificate in the course run
-            ``created``: the date the certificate was created
 
     Returns:
         status: one of (
+            CourseRunProgressStatuses.SAVED_FOR_LATER,
             CourseRunProgressStatuses.COMPLETE,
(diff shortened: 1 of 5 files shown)


enterprise_learner_portal/utils.py
```diff
@@ -21,45 +17,32 @@ class CourseRunProgressStatuses:
 
 def get_course_run_status(course_overview, certificate_info, enterprise_enrollment):
     """
-    Get the progress status of a course run, given the state of a user's certificate in the course.
+    Get the status of a course run, given the state of a user's certificate in the course.
 
-    In the case of self-paced course runs, the run is considered completed when either the course run has ended
-    OR the user has earned a passing certificate 30 days ago or longer.
```
iloveagent57, line 27: What was the original motivation for this 30-day wait period?

enterprise_learner_portal/utils.py
```diff
@@ -21,45 +17,32 @@ class CourseRunProgressStatuses:
...
-            ``created``: the date the certificate was created
 
     Returns:
         status: one of (
+            CourseRunProgressStatuses.SAVED_FOR_LATER,
             CourseRunProgressStatuses.COMPLETE,
             CourseRunProgressStatuses.IN_PROGRESS,
             CourseRunProgressStatuses.UPCOMING,
         )
         None if pacing type is not matched
```
iloveagent57, line 37: Nit - this will no longer return null.

iloveagent57 (approved):
LGTM, one nit to update the docstring.

enterprise_learner_portal/utils.py
adamstankiewicz: Good question. We tried to replicate some logic around course run status from edx-platform 17 months ago, which had that 30-day wait period built in.

This was for the `program_enrollments` endpoint. I checked with Joe about the 30-day wait period to see if there was a valid reason for it; he reached out to Marco who wasn't even aware of it haha.

So, no need for it here; it's not serving any purpose really other than causing UX issues :)

PR #1112, opened 2021-01-11
ADR for explicit customer role assignments.

iloveagent57:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
+++ b/docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
@@ -1,0 +1,74 @@
+Enterprise role assignments will record a customer UUID
+-------------------------------------------------------
+
+Status
+======
+
+Accepted - January 2021
+
+Context
+=======
+
+The ``SystemWideEnterpriseUserRoleAssignment`` model is used to grant role-based permissions
+to enterprise users.  These permissions generally control the users' access to data
+and ability to perform certain functions within an enterprise customer.  This assignment model does `not` currently
+refer to any specific customer - the "context" to which a role is applied is inferred from
+a user (an ``auth.User`` record) having been "linked" to a customer via the ``EnterpriseCustomerUser`` model.
+
+However, a single ``auth.User`` may be linked to multiple ``EnterpriseCustomers``, and if that user
+is assigned a role (such as ``enterprise_admin``), it is inferred that the role assignment applies
+to **every** customer to which the user is linked.  This is undesirable; we generally do not want an admin
+of `Customer A` to also have admin permissions for `Customer B` (to which the user may be linked as a mere ``learner``).
+Instead, the ``SystemWideEnterpriseUserRoleAssignment`` model should explicitly record *which* customer
+the assigned user has the given role in.
+
+Decisions (interspersed with Consequences)
+==========================================
+
+We'll add an ``enterprise_customer`` field to the assignment model
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* It will be an optional field.  It should be a foreign key on ``EnterpriseCustomer.uuid``.
+* Via an edx-rbac change, we will also add a boolean ``applies_to_all_contexts`` field,
+  which is a wildcard that, if true, indicates the user has the role assigned for every customer.
+  It should default to ``false``.
+* The semantics of ``applies_to_all_contexts`` imply that ``enterprise_customer`` will sometimes be null.
+  For example, a user assigned the ``enterprise_openedx_operator`` role will have ``applies_to_all_contexts``
+  set to ``true`` and an ``enterprise_customer`` as ``null``.
+* We will add a customer validator to enforce that the ``enterprise_customer`` field can be null
+  only when ``applies_to_all_contexts`` is true.
+
+We'll backfill the values of this field
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* We'll make the current, implicit assignments (that is, if a role is assigned to a user, that role
+  applies to the user in every customer to which the user is linked) explicit by filling
+  in the UUID value of the linked enterprise customer.
+* For ``enterprise_openedx_operator`` roles, the ``applies_to_all_contexts`` field will be marked true.
+
+We'll do a hand-audit of enterprise_admin role assignments
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* For every user who is effectively an admin of two different enterprise customers,
+  we'll have to manually determine (with help from our support team) which of the two customers
+  they are actually an admin of, and which they are simply a learner within (unless they're truly an admin of both).
+
+The context of a system-wide enterprise assignment is determined from ``enterprise_customer``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* This is true for every role.
+* If ``applies_to_all`` is true, ``get_context()`` should return the wildcard context token "*".
+* The code that populates the user JWTs with (role name, customer UUID) pairs shouldn't have to change -
+  that code should only be looking at the ``get_context()`` method of the role assignment class.
+
+Admin forms must record a value for ``enterprise_customer``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* It would be nice to "auto-populate" it somehow if the user is linked to only one enterprise customer.
+
+This field will be populated by signal handlers related to enterprise user CRUD operations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* Every user will get an ``enterprise_learner`` role assignment for the given enterprise.
+* Pending admin "conversion" will result in an ``enterprise_admin`` role assignment for the given enterprise.
+* We'll delete the role assignment when an ``EnterpriseCustomerUser`` is deactivated.


docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
```diff
@@ -0,0 +1,59 @@
...
+Add an ``enterprise_customer`` field to the assignment model
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* It will be an optional field (for now?)...will ``null`` be a special/wildcard value?.
+* Can we ever make this a required field?  Specifically, will edX/internal users be problematic?
+
+We'll backfill the values of this field
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* It's fine to have multiple assignments for the ``enterprise_learner`` role on a single user -
```
adamstankiewicz, line 37: Yes, it is also fine to have multiple assignments for the `enterprise_admin` role on a single user.

docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
```diff
@@ -0,0 +1,59 @@
...
+  what about ``enterprise_admin``?  Should we allow the backfilling process to create multiple user assignments
+  for this role and clean up the data afterwards?
```
adamstankiewicz, line 39: Can you say a bit more about the data cleanup?

docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
```diff
@@ -0,0 +1,59 @@
...
+Instead, the ``SystemWideEnterpriseUserRoleAssignment`` model should explicitly record *which* customer
+the assigned user has the given role in.
+
+Decisions
+=========
+
+Add an ``enterprise_customer`` field to the assignment model
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* It will be an optional field (for now?)...will ``null`` be a special/wildcard value?.
```
adamstankiewicz, line 31: Hm, if we allow the field to be optional, `null` is essentially a wildcard value. Are there situations in which we want/need to keep that wildcard-like behavior for our Enterprise roles? I feel like there may not be.

adamstankiewicz (commented):
First pass at some thoughts and words.

docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
iloveagent57: Yup, fine from a data perspective.  I really meant, how can we know the ground truth of whether an admin of Customer A should be an admin in Customer B, too, or just a learner?  And I think we can't without getting a list of every admin user who belongs to multiple enterprises and doing a hand-audit.

docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
iloveagent57: The `enterprise_openedx_operator` role will also be assigned in this same table, how do we assign that without giving it an explicit wildcard enterprise UUID (either `null` or `*`)?

docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
iloveagent57: The data cleanup is the hand-audit I mentioned in my reply above.  For every user who is effectively an admin of two different enterprise customers, we'll have to manually determine (with help from support) which of the two customers they are actually an admin of, and which they are simply a learner within (unless they're truly an admin of both).

docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
```diff
@@ -0,0 +1,73 @@
...
+  applies to the user in every customer to which the user is linked) explicit by filling
+  in the UUID value of the linked enterprise customer.
+* For ``enterprise_openedx_operator`` roles, the ``applies_to_all`` field will be marked true.
+
+Do a hand-audit of enterprise_admin role assignments
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* For every user who is effectively an admin of two different enterprise customers,
+  we'll have to manually determine (with help from our support team) which of the two customers
+  they are actually an admin of, and which they are simply a learner within (unless they're truly an admin of both).
```
taliaedX, line 52: Since we have a foreign key on the enterprise uuid, how will this work if they *are* an admin of multiple enterprises? Will there be a row for each enterprise rather than each role/user?

taliaedX: The applies_to_all flag just allows it to be null rather than contain multiple entries, so right now it looks like we allow you to be admin of 1 or all enterprises with no in between. That may be acceptable, just want to verify.

iloveagent57: Right, there would be one row for each (user, role, enterprise_customer) - except in the case of `applies_to_all=true`.  So if you were and enterprise of Pepsico and Coca-cola, there'd be rows like
```
('Talia', 'enterprise_admin', 'Pepsico'),
('Talia', 'enterprise_admin', 'Coca-cola'),
```

taliaedX: Okay good. That makes sense.

taliaedX: I guess another question is on a state change of going from "enterprise_admin" to the operator role, would the old rows be deleted or do we expect to have some logic that fetches by role and ignores the enterprise when looking at an operator role?

iloveagent57: Yeah, that logic already exists - if you have the operator role, the context/customer you have permissions for is `"*"`

docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
```diff
@@ -0,0 +1,73 @@
...
+
+Decisions
+=========
+
+Add an ``enterprise_customer`` field to the assignment model
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* It will be an optional field.  It should
+  also be a foreign key on ``EnterpriseCustomer.uuid``.
+* Can we ever make this a required field?  Specifically, will edX/internal users be problematic?
```
adamstankiewicz, line 33: Can you talk a bit more about this question? Is it an open question that needs more thought before finalizing this ADR?

docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
```diff
@@ -0,0 +1,73 @@
...
+  they are actually an admin of, and which they are simply a learner within (unless they're truly an admin of both).
+
+The context of a system-wide enterprise assignment is determined from this field
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* This is true for every role.
+* If ``applies_to_all`` is true, ``get_context()`` should return the wildcard context token "*".
+* The code that populates the user JWTs with (role name, customer UUID) pairs shouldn't have to change -
+  that code should only be looking at the ``get_context()`` method of the role assignment class.
+  This is an assumption that has to be checked before proceeding.
```
adamstankiewicz, line 61: > This is an assumption that has to be checked before proceeding.

Is there anything else we need to consider here? Otherwise, I think this statement can be nixed.

docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
```diff
@@ -0,0 +1,73 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* It would be nice to "auto-populate" it somehow if the user is linked to only one enterprise customer.
+
+This field will be populated by signal handlers related to enterprise user CRUD operations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* Every user will get an ``enterprise_learner`` role assignment for the given enterprise.
+* Pending admin "conversion" will result in an ``enterprise_admin`` role assignment for the given enterprise.
+* We'll delete the role assignment when an ``EnterpriseCustomerUser`` is deactivated.
```
adamstankiewicz, line 73: Nice, this largely covers my remaining question of whether we should be calling out more explicitly what will be affected by this update :)

docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
```diff
@@ -0,0 +1,74 @@
...
+to enterprise users.  These permissions generally control the users' access to data
+and ability to perform certain functions within an enterprise customer.  This assignment model does `not` currently
+refer to any specific customer - the "context" to which a role is applied is inferred from
+a user (an ``auth.User`` record) having been "linked" to a customer via the ``EnterpriseCustomerUser`` model.
+
+However, a single ``auth.User`` may be linked to multiple ``EnterpriseCustomers``, and if that user
+is assigned a role (such as ``enterprise_admin``), it is inferred that the role assignment applies
+to **every** customer to which the user is linked.  This is undesirable; we generally do not want an admin
+of `Customer A` to also have admin permissions for `Customer B` (to which the user may be linked as a mere ``learner``).
+Instead, the ``SystemWideEnterpriseUserRoleAssignment`` model should explicitly record *which* customer
```
binodpant, line 22: commentary only: TBH this name confuses me, I thought `system wide` meant actually you have access to all enterprises, but looks like I was inferring too much! But love this idea, it seems important to say the least.

docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
```diff
@@ -0,0 +1,74 @@
...
+* We will add a customer validator to enforce that the ``enterprise_customer`` field can be null
+  only when ``applies_to_all_contexts`` is true.
+
+We'll backfill the values of this field
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* We'll make the current, implicit assignments (that is, if a role is assigned to a user, that role
+  applies to the user in every customer to which the user is linked) explicit by filling
+  in the UUID value of the linked enterprise customer.
+* For ``enterprise_openedx_operator`` roles, the ``applies_to_all`` field will be marked true.
```
binodpant, line 47: this should be `applies_to_all_contexts`?

docs/decisions/0004-explicitly-declare-enterprise-for-role-assignments.rst
```diff
@@ -0,0 +1,74 @@
...
+
+We'll add an ``enterprise_customer`` field to the assignment model
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* It will be an optional field.  It should be a foreign key on ``EnterpriseCustomer.uuid``.
+* Via an edx-rbac change, we will also add a boolean ``applies_to_all_contexts`` field,
+  which is a wildcard that, if true, indicates the use has the role assigned for every customer.
+  It should default to ``false``.
+* The semantics of ``applies_to_all_contexts`` imply that ``enterprise_customer`` will sometimes be null.
+  For example, a user assigned the ``enterprise_openedx_operator`` role will have ``applies_to_all_contexts``
```
binodpant, line 36: If something sets the enterprise_customer field, will it also set the `applies_to_all_contexts` field to false?

iloveagent57: Good question!  I'm going to say no (for now), but maybe we could add some messaging in the Django Admin form around this.

binodpant: ok yeah that may be helpful, but yeah just wondering of a case if the enterprise_customer is set and the applies* field is also true .. guess the latter will win?

binodpant (approved):
LGTM!

PR #1114, opened 2021-01-12
ENT-3514 Refactor out duplication between blackboard and canvas Integrated channel codebase

binodpant:
History is that when we did blackboard we reused much of the code from canvas so things started to look very duplicatish, we already cleaned out some, this is just remaining vestiges now turned out not much to be done

Duplication being addressed:

* `client#_create_session` is identical in the two so can be easily refactored to use the existing function `refresh_session_if_expired`, blackboard already does that
* exporter for learner data looks almost identical between the two however refactor to DRY does not make a lot of sense right now so simply moved lines of code around so they both read similar and make easier if we want to DRY in the future

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/canvas/client.py
+++ b/integrated_channels/canvas/client.py
@@ -14,6 +14,7 @@ from django.apps import apps
 
 from integrated_channels.exceptions import ClientError
 from integrated_channels.integrated_channel.client import IntegratedChannelApiClient
+from integrated_channels.utils import refresh_session_if_expired
 
 
 class CanvasAPIClient(IntegratedChannelApiClient):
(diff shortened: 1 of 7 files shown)


integrated_channels/canvas/client.py
```diff
@@ -505,12 +481,20 @@ def _handle_get_user_canvas_course(self, canvas_user_id, learner_data_course_id)
...
-    def _get_oauth_access_token(self, client_id, client_secret):
-        """Uses the client id, secret and refresh token to request the user's auth token from Canvas.
+    def _create_session(self):
+        """
+        Instantiate a new session object for use in connecting with Canvas. Each enterprise customer
+        connecting to Canvas should have a single client session.
+        Will only create a new session if token expiry has been reached
+        """
+        self.session, self.expires_at = refresh_session_if_expired(
+            self._get_oauth_access_token,
```
alex-sheehan-edx, line 491: Oh I see, you're able to make this a util for both Canvas and blackboard by passing in their individual token retrieval methods 👍 . I'm very in favor of this!

integrated_channels/canvas/exporters/learner_data.py
```diff
@@ -32,32 +32,34 @@ def get_learner_data_records(
...
                 'No learner data was sent for user [%s] because a Canvas user ID could not be found.',
-                enterprise_enrollment.enterprise_customer_user.username
+                enterprise_customer_user.username
             )
             return None
         percent_grade = kwargs.get('grade_percent', None)
         completed_timestamp = completed_date.strftime("%F") if isinstance(completed_date, datetime) else None
 
+        course_catalog_client = get_course_catalog_api_service_client(
+            site=enterprise_customer_user.enterprise_customer.site
```
alex-sheehan-edx, line 46: 👍

alex-sheehan-edx (approved):
nice cleanup! 👍  lgtm

integrated_channels/canvas/client.py
binodpant: yeah I had forgotten we had this method then reviewing side by side just saw they were identical as long as I can pasa f unctoin to get oauth_access_token so works out!

PR #1115, opened 2021-01-12
Degreed API Endpoints

taliaedX:
This adds endpoints for the Degreed LMS to the Integrated Channels API. Tests are completed as well.
**NOTE**: Will do versioning post approval to avoid millions of rebases.
I am also likely aware of the many quality failures (not getting the right quality checks running locally which is annoying) so please don't point those out.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/api/v1/degreed/urls.py
+++ b/integrated_channels/api/v1/degreed/urls.py
@@ -1,0 +1,12 @@
+"""
+    url mappings for integrated_channels/api/v1/degreed/
+"""
+
+from rest_framework import routers
+
+from .views import DegreedConfigurationViewSet
+
+app_name = 'degreed'
+router = routers.DefaultRouter()  # pylint: disable=invalid-name
+router.register(r'configuration', DegreedConfigurationViewSet, basename="configuration")
+urlpatterns = router.urls
--- a/integrated_channels/api/v1/degreed/views.py
+++ b/integrated_channels/api/v1/degreed/views.py
@@ -1,0 +1,40 @@
+"""
+Viewsets for Degreed
+"""
+
+from edx_rbac.mixins import PermissionRequiredMixin
+from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
+from rest_framework import permissions, viewsets
+from rest_framework.authentication import SessionAuthentication
+
+from django.core.exceptions import PermissionDenied
+from django.http import Http404
+
+from enterprise.models import EnterpriseCustomerUser
+from integrated_channels.degreed.models import DegreedEnterpriseCustomerConfiguration
+
+from .serializers import DegreedConfigSerializer
+
+
+class DegreedConfigurationViewSet(PermissionRequiredMixin, viewsets.ModelViewSet):
+    serializer_class = DegreedConfigSerializer
+    permission_classes = (permissions.IsAuthenticated,)
+    permission_required = 'enterprise.can_access_admin_dashboard'
+
+    def get_queryset(self):
+        enterprise_customer_users = EnterpriseCustomerUser.objects.filter(user_id=self.request.user.id)
+        if enterprise_customer_users:
+            enterprise_customers_uuids = [str(enterprise_customer_user.enterprise_customer.uuid) for
+                                          enterprise_customer_user in enterprise_customer_users]
+            enterprise_customer_uuid = self.request.query_params.get('enterprise_customer')
+
+            if enterprise_customer_uuid:
+                if enterprise_customer_uuid in enterprise_customers_uuids:
+                    return DegreedEnterpriseCustomerConfiguration.objects.filter(
+                        enterprise_customer__uuid=enterprise_customer_uuid
+                    )
+                raise PermissionDenied()
+            return DegreedEnterpriseCustomerConfiguration.objects.filter(
+                enterprise_customer__uuid__in=enterprise_customers_uuids
+            )
+        raise Http404
--- a/tests/test_integrated_channels/test_api/test_degreed/test_views.py
+++ b/tests/test_integrated_channels/test_api/test_degreed/test_views.py
@@ -1,0 +1,106 @@
+"""
+Tests for the `integrated_channels` Degreed configuration api.
+"""
+import json
+from uuid import uuid4
+
+import mock
+
+from django.urls import reverse
+
+from enterprise.constants import ENTERPRISE_ADMIN_ROLE
+from integrated_channels.degreed.models import DegreedEnterpriseCustomerConfiguration
+from test_utils import APITest, factories
+
+ENTERPRISE_ID = str(uuid4())
+
+
+class DegreedConfigurationViewSetTests(APITest):
+    """
+    Tests for DegreedConfigurationViewSet REST endpoints
+    """
+    def setUp(self):
+        super().setUp()
+        self.set_jwt_cookie(self.client, [(ENTERPRISE_ADMIN_ROLE, ENTERPRISE_ID)])
+        self.client.force_authenticate(user=self.user)
+
+        self.enterprise_customer = factories.EnterpriseCustomerFactory(uuid=ENTERPRISE_ID)
+        self.enterprise_customer_user = factories.EnterpriseCustomerUserFactory(
+            enterprise_customer=self.enterprise_customer,
+            user_id=self.user.id,
+        )
+
+        self.degreed_config = factories.DegreedEnterpriseCustomerConfigurationFactory(
+            enterprise_customer=self.enterprise_customer
+        )
+
+    @mock.patch('enterprise.rules.crum.get_current_request')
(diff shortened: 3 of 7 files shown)


integrated_channels/api/v1/degreed/urls.py
```diff
@@ -0,0 +1,12 @@
+"""
+    url mappings for integrated_channels/v1/degreed/
```
binodpant, line 2: nit: just noted the path here `integrated_channels/v1/degreed/` probably needs to be `integrated_channels/api/v1/degreed/`? having said this all other channels seem to be doing this style! not too worried if we keep it the way it is or even remove the comment!

integrated_channels/api/v1/degreed/views.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def get_queryset(self):
+        enterprise_customer_users = EnterpriseCustomerUser.objects.filter(user_id=self.request.user.id)
+        if enterprise_customer_users:
+            enterprise_customers_uuids = [str(enterprise_customer_user.enterprise_customer.uuid) for
+                                          enterprise_customer_user in enterprise_customer_users]
+            enterprise_customer_uuid = self.request.query_params.get('enterprise_customer')
+
+            if enterprise_customer_uuid:
+                print(enterprise_customer_uuid)
```
binodpant, line 33: should we remove the print?

tests/test_integrated_channels/test_api/test_degreed/test_views.py
```diff
@@ -0,0 +1,106 @@
...
+    @mock.patch('enterprise.rules.crum.get_current_request')
+    def test_delete(self, mock_current_request):
+        mock_current_request.return_value = self.get_request_with_jwt_cookie(
+            system_wide_role=ENTERPRISE_ADMIN_ROLE,
+        )
+        url = reverse('api:v1:degreed:configuration-detail', args=[self.degreed_config.id])
+        response = self.client.delete(url)
+        configs = DegreedEnterpriseCustomerConfiguration.objects.filter()
+        self.assertEqual(response.status_code, 204)
+        self.assertEqual(len(configs), 0)
```
binodpant, line 106: ideally should there be a check for this length before the delete is issue to be further strict?

taliaedX: ? I don't think I quite follow. The url passed to the delete function includes an id. If we had 0 configs to start with, then that id would be invalid and throw an error.

integrated_channels/api/v1/degreed/views.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class DegreedConfigurationViewSet(PermissionRequiredMixin, viewsets.ModelViewSet):
+    authentication_classes = (JwtAuthentication, SessionAuthentication,)
+    serializer_class = DegreedConfigSerializer
+    permission_classes = (permissions.IsAuthenticated,)
+    permission_required = 'enterprise.can_access_admin_dashboard'
+
+    def get_queryset(self):
+        enterprise_customer_users = EnterpriseCustomerUser.objects.filter(user_id=self.request.user.id)
```
edx-abolger, line 26: I assume you dont have to check for user existence here before asking for user.id because this will never allow a null request.user because of the super classes?

edx-abolger: Since at the end of this whole thing you know you only want 1 or 0 enterprise customer UUID,  there's a util here you could use that only does 1 db hit, I think. See utils.get_enterprise_customer_user()

edx-abolger (commented):
No changes for me, will leave to Binod to approve at this time.

tests/test_integrated_channels/test_api/test_degreed/test_views.py
alex-sheehan-edx: I think he means check the length of `configs` before you do the delete as you do afterwards. We could also add another so that we have two and go to one, either way.

alex-sheehan-edx (approved):
A question or two but other than those this LGTM 👍

tests/test_integrated_channels/test_api/test_degreed/test_views.py
binodpant: ah maybe I am mixed up then, I assumed there would be a len(configs) of 1 before doing the delete request

binodpant: but given that you had indicated if the record does not exist it wil fail anyway I am not too worried just that if it's easy that kind of test is a bit more explicit.. but maybe not worth for this small bit

binodpant (approved):
I am good now, made some minor comments and test notes but none I feel are blocking

integrated_channels/api/v1/degreed/views.py
taliaedX: yeah they'd get a 403 long before we get here if there was no user.

integrated_channels/api/v1/degreed/views.py
```diff
@@ -0,0 +1,41 @@
...
+from django.http import Http404
+
+from enterprise.models import EnterpriseCustomerUser
+from integrated_channels.degreed.models import DegreedEnterpriseCustomerConfiguration
+
+from .serializers import DegreedConfigSerializer
+
+
+class DegreedConfigurationViewSet(PermissionRequiredMixin, viewsets.ModelViewSet):
+    authentication_classes = (JwtAuthentication, SessionAuthentication,)
```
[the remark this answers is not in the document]
taliaedX: So I ended up adding these because I couldn't actually test the API directly via postman without it. Basically by using postman I am bypassing LMS/Platform's authentication pathway and needed a way to actually do so which is what this allows.

Most of our APIs actually have these authentication classes (enterprise itself is one example). However, the actual implementation details I have not investigated so can't provide more detail atm about the exact mechanisms.

tests/test_integrated_channels/test_api/test_degreed/test_views.py
taliaedX: Gotcha. Yeah specifying 1 is definitely more explicit, but yeah like I said we wouldn't even have an id to delete and it'd throw an error.

integrated_channels/api/v1/degreed/views.py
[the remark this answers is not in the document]
alex-sheehan-edx: I believe that is exactly what Postman Interceptor is for- In terms of bypassing LMS' authentication pathway. Let me know if you want any info on it, @edx-abolger is also a good reference as she showed me 🙂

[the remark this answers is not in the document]
taliaedX: I think the use of these classes is more something we probably want to discuss with Titans since I don't know if that is the *only* use of these authentication classes given how expansively they are used in this repo. As far as the utility, yeah if I don't use these classes that will likely come in handy.

For now, leaving it or not is pretty harmless (I mean if there was a security concern then I'd be eyeing like 90% of our API endpoints in edx-enterprise, so uhhh I hope not). It has a use, but if we wanted to adopt this interceptor workflow then that use would be unnecessary.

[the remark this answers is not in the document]
taliaedX: My uninformed guess is that they come in handy when hitting the endpoints outside of an authentication process (ie backend to backend) which is likely to be forever irrelevant for these particular APIs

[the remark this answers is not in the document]
binodpant: fwiw @alex-sheehan-edx there is an unused import of jwtauthentication in blackboard (sorry not related to this pr so we can take offline) https://github.com/edx/edx-enterprise/blob/de43ebeb0f147e67b8d5fbdbac19579b89c7ab8c/integrated_channels/api/v1/blackboard/views.py#L5 I cannot find usage of jwtauthetnication in any of the new channels we worked on (but they exist in other channels). we should find out more.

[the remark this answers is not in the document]
binodpant: my understanding is that the `PermissionRequiredMixin` handles the RBAC by only allowing users with the right roles, but it was not clear to me if it depended on the jwtAuthentication being present.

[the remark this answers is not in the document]
binodpant: yeah and it does not depend on there being a jwt auth of course, e.g., https://github.com/edx/edx-enterprise/blob/320ea9842175b23d40f4f32a80806e094cd85421/integrated_channels/api/v1/canvas/views.py#L18 so my guess is you don't need it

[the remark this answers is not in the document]
taliaedX: yeah I just went ahead and removed it from my PR.

kdmccormick:
hey @taliaedX -- when we're back on Tuesday, do you mind taking a look at https://github.com/edx/edx-platform/pull/26084 ? It'll pull both of our recent changes into edx-platform.

PR #1117, opened 2021-01-13
Add enterprise_customer field to the SystemWideEnterpriseUserRoleAssignment model

iloveagent57:
…and updates Django Admin definitions around same.

* On the `SystemWideEnterpriseUserRoleAssignment` model, adds an `enterprise_customer` FK (nullable).
* Updates the admin to show the "effective" customer in the detail view, and the explicit value in the list view.  The effective value is the deprecated way we currently determine role assignment - by implicitly assigning the role on every customer to which the user is linked.
* In the detail view/form, the "Enterprise customer" dropdown contains only customers to which the user is currently linked.
* If adding a new assignment via the admin form, the customer dropdown will contain no customers to choose from at all - the administrator must first provide a user and role, click "Save and continue editing", and then a list of linked customers will be present. The help text of the field mentions this.  Rationale: we don't want to provide a drop-down list of _every_ customer that exists in production when creating a new assignment.  The alternative would be to make this form more "dynamic", which would require more development work than provided value.

To test this locally:
* Run `./manage.py lms migrate enterprise` to apply the migration.
* Visit http://localhost:18000/admin/enterprise/systemwideenterpriseuserroleassignment/
* Change a user role assignment by having it apply to a specific enterprise to which the user is linked.
* The `effective_enterprise_customer` value should change to include only that customer.

role-assignment-admin-list

role-assignment-admin-detail

**Merge checklist:**

**Post merge:**


enterprise/admin/forms.py
```diff
@@ -637,7 +637,23 @@ class SystemWideEnterpriseUserRoleAssignmentForm(UserRoleAssignmentAdminForm):
...
         model = SystemWideEnterpriseUserRoleAssignment
-        fields = ['user', 'role']
+        fields = [
+            'user',
+            'role',
+            'enterprise_customer',
+            'applies_to_all',
+        ]
+
+    def __init__(self, *args, **kwargs):
```
kellmarie, line 647: We could probably add a test in `test_forms` for this behavior.

enterprise/models.py
```diff
@@ -2554,6 +2554,25 @@ class SystemWideEnterpriseUserRoleAssignment(EnterpriseRoleAssignmentContextMixi
 
     role_class = SystemWideEnterpriseRole
 
+    enterprise_customer = models.ForeignKey(
```
kellmarie, line 2557: Should we be making this be required now or is it too early to do that?

iloveagent57: It's too early, and there will be some rows for which this column must be null (i.e. with the `enterprise_openedx_operator` role).  A PR some time after this one will put in some validation logic around this field and `applies_to_all`, though.

enterprise/models.py
```diff
@@ -2554,6 +2555,28 @@ class SystemWideEnterpriseUserRoleAssignment(EnterpriseRoleAssignmentContextMixi
...
+        related_name="system_wide_role_assignments",
+        on_delete=models.CASCADE,
+        help_text=_(
+            'The EnterpriseCustomer for which the role is assigned for the user.  If creating a new assignment, '
+            'you must first provide a user email and role, then click "Save and continue editing" '
+            'BEFORE selecting from this dropdown.'
+        ),
+    )
+
+    applies_to_all = models.BooleanField(
```
brittneyexline, line 2571: not something you need to change now, but curious about your thoughts on this field being something that is part of the parent model definition in edx-rbac? especially since the concept of the '*' for context already exists at that level, and this seems to be a more explicit way of defining that on the model?

iloveagent57: Hmm, yeah, that's a good idea.  I think it's not a ton of work to pull that out, lemme give it a shot.

adamstankiewicz (approved):

chavesj (approved):
LGTM

PR #1118, opened 2021-01-14
Setting courses to the Canvas public index by default

alex-sheehan-edx:
In order to make the integrated course discoverable on Canvas, we need to set it to be a part of the public index. 

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/canvas/client.py
+++ b/integrated_channels/canvas/client.py
@@ -221,8 +221,9 @@ class CanvasAPIClient(IntegratedChannelApiClient):
                     self.enterprise_configuration.canvas_base_url,
                     course_id,
                 )
+                # Providing the param `event` and setting it to `offer` is equivalent to publishing the course.
                 self._put(url, json.dumps({
-                    'course': {'image_url': content_metadata_item['image_url']}
+                    'course': {'image_url': content_metadata_item['image_url'], 'event': 'offer'}
                 }).encode('utf-8'))
         except Exception:  # pylint: disable=broad-except
             # we do not want course image update to cause failures
--- a/integrated_channels/canvas/exporters/content_metadata.py
+++ b/integrated_channels/canvas/exporters/content_metadata.py
@@ -24,6 +24,7 @@ class CanvasContentMetadataExporter(ContentMetadataExporter):
         'is_public': 'is_public',
         'self_enrollment': 'self_enrollment',
         'course_code': 'key',
+        'indexed': 'indexed'
     }
 
     LONG_STRING_LIMIT = 2000
@@ -60,7 +61,9 @@ class CanvasContentMetadataExporter(ContentMetadataExporter):
 
     def transform_is_public(self, content_metadata_item):  # pylint: disable=unused-argument
         """
-        Whether to make the course visible in the public Canvas index by default.
+        Whether to make the course content visible to students by default. Note that setting this feature to false
+        will not remove the course from discovery, but will rather remove all visible content from the course if
+        any student attempts to access it.
         """
         return True
 
@@ -69,3 +72,9 @@ class CanvasContentMetadataExporter(ContentMetadataExporter):
         Whether to allow students to self enroll. Helps students enroll via link or enroll button.
         """
         return True
+
+    def transform_indexed(self, content_metadata_item):  # pylint: disable=unused-argument
+        """
+        Whether to make the course default to the public index or not.
+        """
+        return 1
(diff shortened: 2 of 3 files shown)


integrated_channels/canvas/exporters/content_metadata.py
```diff
@@ -24,6 +24,7 @@ class CanvasContentMetadataExporter(ContentMetadataExporter):
         'is_public': 'is_public',
```
alex-sheehan-edx, line 24: In the docs of the `transform` method below for this `is_public` param, we claim that it is to set the course to the public index, but after some investigation I'm fairly certain that it is not the `is_public` param but instead the `indexed` param. According to the Canvas API docs `is_public` is "Set to true if course is public to both authenticated and unauthenticated users." Not sure exactly what that does or if it's actually needed. Any clues? Perhaps it can be removed

alex-sheehan-edx: did some discovery- The `is_public` param allows students to view to content of the course. Without the `is_public` param the course is discoverable in the sense that it will show up in the catalog, but no content shows up when navigating to the actual course: ie 
image

when the `is_public` parameter is included then the student can view the course content: ie 
image

So I feel as though we want to keep this parameter but perhaps a doc change so it more closely reflects the actual behavior?

binodpant: yes indeed their doc sucks! this was admitted by canvas folks themselves hence I pasted this note in the ticket for easy access

```
[2] We find we are unable to set the following settings automatically when creating a course:
        "Include this course in the public course index": true (is false by default)
        Context: It may be a great ability to have to auto publish and auto include in public index. (unless the customer does not want us to do so)
I'm not sure if there's a specific product reason why this isn't documented – but you can do this by including the `course[indexed]` parameter when creating a course, and setting the value to '1'
```

integrated_channels/canvas/exporters/content_metadata.py
```diff
@@ -69,3 +72,9 @@ def transform_self_enrollment(self, content_metadata_item):  # pylint: disable=u
         Whether to allow students to self enroll. Helps students enroll via link or enroll button.
         """
         return True
+
+    def transform_indexed(self, content_metadata_item):  # pylint: disable=unused-argument
+        """
+        Whether to make the course default to the public index or not.
+        """
+        return 1
```
binodpant, line 80: interesting, 1 vs True?

integrated_channels/canvas/client.py
```diff
@@ -221,8 +221,9 @@ def _update_course_details(self, course_id, serialized_data):
                     self.enterprise_configuration.canvas_base_url,
                     course_id,
                 )
+                # Providing the param `event` and setting it to `offer` is equivalent to publishing the course.
```
binodpant, line 224: this looks good for now great you found this! we can always re eval if autopublish will be on/off later.

integrated_channels/canvas/exporters/content_metadata.py
```diff
@@ -69,3 +72,9 @@ def transform_self_enrollment(self, content_metadata_item):  # pylint: disable=u
         Whether to allow students to self enroll. Helps students enroll via link or enroll button.
         """
         return True
+
+    def transform_indexed(self, content_metadata_item):  # pylint: disable=unused-argument
```
binodpant, line 76: be nice to add a quick test case or addition to existing test case for exporter?

integrated_channels/canvas/exporters/content_metadata.py
alex-sheehan-edx: The posted answers on all Canvas forums had it explicitly listed as `1`

tests/test_integrated_channels/test_canvas/test_exporters/test_content_metadata.py
```diff
@@ -139,3 +139,12 @@ def test_transform_self_enrollment(self):
         content_metadata_item = GENERIC_CONTENT_METADATA_ITEM
         exporter = CanvasContentMetadataExporter('fake-user', self.config)
         assert exporter.transform_self_enrollment(content_metadata_item) is True
+
+    @responses.activate
+    def test_transform_indexed(self):
+        """
+        'CanavsContentMEtadataExproter``'s ''transform_indexed'' returns 1 as a value
```
binodpant, line 146: nit: typo

binodpant (approved):
I am good, just posted a nit

PR #1121, opened 2021-01-19
Making Degreed integration configuration fields required

alex-sheehan-edx:
Setting all configuration fields to required. Turns out, the Degreed factory already accounts for all fields so no changes required.

I took a look at staging and saw that there were no configurations. I'm assuming that any configuration on prod has all the fields, but perhaps we should check? As to make sure nothing breaks when the required fields go live.

**Merge checklist:**

**Post merge:**


integrated_channels/degreed/migrations/0009_auto_20210119_1546.py
```diff
@@ -0,0 +1,73 @@
...
+        ),
+        migrations.AlterField(
+            model_name='historicaldegreedenterprisecustomerconfiguration',
+            name='key',
+            field=models.CharField(help_text='The API Client ID provided to edX by the enterprise customer to be used to make API calls to Degreed on behalf of the customer.', max_length=255, verbose_name='API Client ID'),
+        ),
+        migrations.AlterField(
+            model_name='historicaldegreedenterprisecustomerconfiguration',
+            name='secret',
+            field=models.CharField(help_text='The API Client Secret provided to edX by the enterprise customer to be used to make API calls to Degreed on behalf of the customer.', max_length=255, verbose_name='API Client Secret'),
```
taliaedX, line 71: part of me is thinking "let's make the field less of a scroll-y boy", but also ehhh.

taliaedX (approved):
One slight nit, but also I don't care too strongly.
One thing to do - if you haven't already done it - is to add a degreed config locally BEFORE this migration is applied (go ahead and fill in all fields) and verify it doesn't throw an error for some reason (it really shouldn't, but JUST IN CASE).

Also to your PR description around verifying in Prod: I did verify all configs have all fields filled in on Prod prior to making the ticket, however, if you aren't sure how to verify and would love to see it with your own eyes, let me know! I can show you my process.

alex-sheehan-edx:
> add a degreed config locally BEFORE this migration is applied (go ahead and fill in all fields) and verify it doesn't throw an error for some reason

Goooooood idea... I've just already applied my migrations. Gonna have to remember how to step back

> I did verify all configs have all fields filled in on Prod prior to making the ticket

Nice! 

> however, if you aren't sure how to verify and would love to see it with your own eyes, let me know

This would be accessing Django Admin on prod? Or I guess checking the DB? Sure! Let's find a time to go through it together.

integrated_channels/degreed/migrations/0009_auto_20210119_1546.py
alex-sheehan-edx: do you mean the length of the line? I can happily move it to multiple lines.

alex-sheehan-edx: done!

PR #1127, opened 2021-01-27
ENt-3992 | added check for linked identity provider.

HammadAhmadWaqas:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/decorators.py
+++ b/enterprise/decorators.py
@@ -128,11 +128,7 @@ def enterprise_login_required(view):
         enterprise_uuid = kwargs['enterprise_uuid']
         enterprise_customer = get_enterprise_customer_or_404(enterprise_uuid)
         query_params = request.GET
-        # Check if tpa_hint was passed as query param
-        tpa_hint = query_params.get('tpa_hint')
-        if not tpa_hint and not enterprise_customer.has_multiple_idps:
-            tpa_hint = enterprise_customer.identity_provider\
-                if enterprise_customer.identity_provider else None
+        tpa_hint_param = query_params.get('tpa_hint')
 
         # Now verify if the user is logged in. If user is not logged in then
         # send the user to the login screen to sign in with an
@@ -141,7 +137,7 @@ def enterprise_login_required(view):
             parsed_current_url = urlparse(request.get_full_path())
             parsed_query_string = parse_qs(parsed_current_url.query)
             parsed_query_string.update({
-                'tpa_hint': tpa_hint,
+                'tpa_hint': enterprise_customer.get_tpa_hint(tpa_hint_param),
                 FRESH_LOGIN_PARAMETER: 'yes'
             })
             next_url = '{current_path}?{query_string}'.format(
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -366,15 +366,14 @@ class EnterpriseCustomer(TimeStampedModel):
     @property
     def identity_provider(self):
         """
-        Return the identity providers associated with this enterprise customer.
+        Return the first identity provider id associated with this enterprise customer.
         """
         # pylint: disable=no-member
-        try:
-            # pylint: disable=no-member
-            identity_provider = self.enterprise_customer_identity_providers.first()
-            return identity_provider and identity_provider.provider_id
-        except ObjectDoesNotExist:
-            return None
+        identity_provider = self.enterprise_customer_identity_providers.first()
+        if identity_provider:
+            return identity_provider.provider_id
+        LOGGER.info("No linked identity provider found for enterprise customer: %s", self.uuid)
+        return None
 
     @property
     def identity_providers(self):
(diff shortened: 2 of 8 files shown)


enterprise/models.py
```diff
@@ -366,7 +366,7 @@ def enterprise_customer_identity_provider(self):  # pylint: disable=invalid-name
     @property
     def identity_provider(self):
         """
-        Return the identity providers associated with this enterprise customer.
+        Return the identity provider id associated with this enterprise customer.
```
zamanafzal, line 369: shouldn't it be `Return the identity **providers** id associated with this enterprise customer.`

enterprise/decorators.py
```diff
@@ -129,10 +129,13 @@ def wrapper(request, *args, **kwargs):
         enterprise_customer = get_enterprise_customer_or_404(enterprise_uuid)
         query_params = request.GET
         # Check if tpa_hint was passed as query param
-        tpa_hint = query_params.get('tpa_hint')
-        if not tpa_hint and not enterprise_customer.has_multiple_idps:
-            tpa_hint = enterprise_customer.identity_provider\
-                if enterprise_customer.identity_provider else None
+        tpa_hint_param = query_params.get('tpa_hint')
```
zamanafzal, line 132: Can we please add a comment here. So that it can be understandable why we are doing it.

zamanafzal (approved):
LGTM.

enterprise/models.py
HammadAhmadWaqas: its returning id of only *one* identity_provider(i.e first one) so think it should 
`Return the identity provider id associated with this enterprise customer.`
or
`Return the identity provider's id associated with this enterprise customer.`

enterprise/decorators.py
HammadAhmadWaqas: is not it self-explanatory?

mattdrayer: I think a comment that provides context would be helpful to future collaborators.

enterprise/models.py
mattdrayer: Is this where the check for a "default" IdP would happen?

mattdrayer: Also this seems like a good place to add some logging information -- for example what if no IdP is found.

enterprise/views.py
```diff
@@ -1071,8 +1071,14 @@ def get(self, request):
 
         # Check if tpa_hint was passed as query param then redirect to that IDP login
         # Otherwise check if enterprise is linked to one IDP, then page will be redirected there
-        tpa_hint = query_params.get('tpa_hint') or enterprise_customer.identity_providers.first().provider_id if \
-            enterprise_customer.identity_providers.count() == 1 else None
+
+        tpa_hint_param = query_params.get('tpa_hint')
```
mattdrayer, line 1075: duplicate code?

tests/test_enterprise/views/test_enterprise_proxy_login_view.py
```diff
@@ -99,28 +99,50 @@ def test_tpa_redirect(self, use_next):
...
+        ('fake_idp', 'idp-1', True, None),
+        ('idp-1', 'idp-1', False, None),
+        (None, None, False, None),
     )
     @ddt.unpack
-    def test_tpa_hint_redirect(self, use_tpa_hint, tpa_hint):
+    def test_tpa_redirects_using_tpa_hint_param(
+            self,
+            tpa_hint_param,
+            identity_provided_id,
```
mattdrayer, line 117: should this be `identity_provider_id`?

enterprise/models.py
```diff
@@ -383,6 +383,13 @@ def identity_providers(self):
         """
         return self.enterprise_customer_identity_providers.all()
 
+    @property
+    def identity_provider_ids(self):
+        """
+        Return the identity provider Ids associated with this enterprise customer.
+        """
+        return str(self.identity_providers.values_list('provider_id', flat=True))
```
irfanuddinahmad, line 391: Shouldn't we return a list rather than a string?

HammadAhmadWaqas: @irfanuddinahmad good catch

enterprise/decorators.py
HammadAhmadWaqas: moved this logic to a class method with comments.

enterprise/views.py
HammadAhmadWaqas: moved into a common method

enterprise/models.py
HammadAhmadWaqas: Yes, it seems a good place for a default check. added logger.

tests/test_enterprise/views/test_enterprise_proxy_login_view.py
HammadAhmadWaqas: good catch, updated.

mattdrayer (approved):

PR #1128, opened 2021-01-27
Conditionally allows the deletion of individual EnterpriseCourseEnrollment records

iloveagent57:
Conditionally allows the deletion of individual EnterpriseCourseEnrollment records and related
LicensedEnterpriseCourseEnrollment records via the Django Admin site, so that site admins can manually
delete enterprise enrollments that were created in error.
This is only allowed if a Django settings feature flag is set to True.

How I manually tested - 
1. Create a licensed enterprise enrollment for the `verified@example.com` user in the pied-piper customer: 
1-licensed-enrollment-learner-portal

2. As a site admin, visit the Django Admin for enterprise enrollments, see that I'm allowed to delete this one (I toggled the feature on in my local settings):
2-admin-now-allows-deletion

3. Click delete, get the confirmation screen that shows me I'll delete both the Enterprise Course Enrollment and the LicensedEnterpriseCourseEnrollment record: 
3-confirm-deletion-of-licensed-enrollment

4. After confirming deletion, noticed that my enrollment is gone from the list: 
4-enrollment-is-gone-from-admin

5. As `verified@example.com`, notice that the course is no longer on my learner portal dashboard: 
5-course-is-gone-from-learner-portal

6. As the same learner, see that the course remains on my "personal" dashboard, indicating that the underlying `student.CourseEnrollment` record still exists: 
6-course-is-still-in-personal-dashboard

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/admin/__init__.py
+++ b/enterprise/admin/__init__.py
@@ -10,6 +10,7 @@ from edx_rbac.admin import UserRoleAssignmentAdmin
 from simple_history.admin import SimpleHistoryAdmin
 from six.moves.urllib.parse import urlencode  # pylint: disable=import-error
 
+from django.conf import settings
 from django.conf.urls import url
 from django.contrib import admin, auth
 from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
@@ -20,6 +21,7 @@ from django.utils.html import format_html
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext as _
 
+from enterprise import constants
 from enterprise.admin.actions import export_as_csv_action, refresh_catalog
 from enterprise.admin.forms import (
     EnterpriseCustomerAdminForm,
@@ -586,6 +588,7 @@ class EnterpriseCourseEnrollmentAdmin(admin.ModelAdmin):
         'enterprise_customer_user',
         'course_id',
         'saved_for_later',
+        'license_uuid',
     )
 
     list_display = (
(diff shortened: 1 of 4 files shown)


enterprise/admin/__init__.py
```diff
@@ -597,6 +600,12 @@ class Meta:
     change_list_template = "enterprise/admin/enterprise_course_enrollments_list.html"
     search_fields = ('enterprise_customer_user__user_id', 'course_id',)
 
+    def license_uuid(self, obj):
+        """
+        Return the subscription license UUID (if any exists)  associated with this enrollment.
+        """
+        return str(obj.license.license_uuid)
```
kellmarie, line 607: You checked this doesn't break on enrollments that don't have a license right?

iloveagent57: Yup, `license` is a property that safely returns  `None` if no associated `LicensedEnterpriseCourseEnrollment` exists.  But good general point that this could use a test.

enterprise/admin/__init__.py
```diff
@@ -607,6 +616,9 @@ def has_delete_permission(self, request, obj=None):
         """
         Disable deletion for EnterpriseCourseEnrollment.
         """
+        features = getattr(settings, 'FEATURES', {})
+        if features.get(constants.ALLOW_ADMIN_ENTERPRISE_COURSE_ENROLLMENT_DELETION, False):
+            return True
```
kellmarie, line 621: Idk if I'm completely missing something obvious but couldn't we just do `return features.get(constants.ALLOW_ADMIN_ENTERPRISE_COURSE_ENROLLMENT_DELETION, False)`

iloveagent57: hah, yes :)

enterprise/admin/__init__.py
iloveagent57: I'm gonna punt on this for now in the interest of time, and since there's no existing tests for most of the base ModelAdmin definitions.

kellmarie (approved):

PR #1129, opened 2021-01-28
version 3.17.15 | Update get_context()

iloveagent57:
In ``SystemWideEnterpriseUserRoleAssignment``, Use either ``applies_to_all_contexts`` or ``enterprise_customer``
  if they are True or non-null, respectively, in determining the result of ``get_context()``,
  but continue to return list of all linked enterprise customer UUIDs if not, (which is the current behavior).
  This is a small step on our journey to explicitly defining user-role assignments

*Manual Testing*
1. Give a user 2 explicit assignments for `enterprise_learner`, to two different enterprise customers, make sure there's a JWT role entry for both:
Screen Shot 2021-01-29 at 10 52 02 AM
Screen Shot 2021-01-29 at 10 52 21 AM

2. Give a user the `enterprise_admin` role with `applies_to_all_contexts` set to true, see that their JWT role has an `enterprise_admin:*` entry:
Screen Shot 2021-01-29 at 10 53 09 AM
Screen Shot 2021-01-29 at 10 59 51 AM

3. Give a user the `enterprise_openedx_operator` role, set `applies_to_all_contexts` to false, see that they still have `enterprise_openedx_operator:*` in their JWT roles:
Screen Shot 2021-01-29 at 10 55 40 AM
Screen Shot 2021-01-29 at 10 57 16 AM

**Merge checklist:**

**Post merge:**

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -1991,6 +1991,16 @@ class TestSystemWideEnterpriseUserRoleAssignment(unittest.TestCase):
     Tests SystemWideEnterpriseUserRoleAssignment.
     """
 
+    def _create_and_link_user(self, user_email, *enterprise_customers):
+        """
+        Helper that creates a User with the given email, then links that user
+        to each of the given EnterpriseCustomers.
+        """
+        user = factories.UserFactory(email=user_email)
+        for enterprise_customer in enterprise_customers:
+            EnterpriseCustomerUser.objects.link_user(enterprise_customer, user_email)
+        return user
+
     @ddt.data(
         {
             'role_name': ENTERPRISE_LEARNER_ROLE,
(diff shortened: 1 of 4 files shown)


tests/test_models.py
```diff
@@ -2035,12 +2043,9 @@ def test_get_context_for_multiple_contexts(self, role_name, expected_context):
         Verify that `SystemWideEnterpriseUserRoleAssignment.get_context` method works as expected for user with
         multiple contexts.
         """
-        user_email = 'edx@example.com'
         enterprise_customer_1 = factories.EnterpriseCustomerFactory(uuid='47130371-0b6d-43f5-01de-71942664de2b')
         enterprise_customer_2 = factories.EnterpriseCustomerFactory(uuid='47130371-0b6d-43f5-01de-71942664de2c')
-        user = factories.UserFactory(email=user_email)
-        EnterpriseCustomerUser.objects.link_user(enterprise_customer_1, user_email)
-        EnterpriseCustomerUser.objects.link_user(enterprise_customer_2, user_email)
+        user = self._create_and_link_user('edx@example.com', enterprise_customer_1, enterprise_customer_2)
```
adamstankiewicz, line 2048: Nice cleanup here.

adamstankiewicz (approved):
LGTM

PR #1130, opened 2021-01-28
feat: removing references to sync_enterprise_catalog_query boolean field…

chavesj:
…of the EnterpriseCustomerCatalog model.

### Ticket:
[ENT-3959](https://openedx.atlassian.net/browse/ENT-3959)

### Testing Instructions:
- Install edx-enterprise package, run migrations from the LMS shell
- Ensure the enterprise-catalog service is running on your devstack setup
- Open the Django Admin UI for both the LMS and enterprise-catalog services. Make changes to Catalogs and Catalog Queries in the LMS and observe that they are always updated in the enterprise-catalog service

**Merge checklist:**

**Post merge:**


enterprise/migrations/0122_remove_sync_enterprise_catalog_query_bool.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('enterprise', '0121_systemwiderole_add_ent_cust_field'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='enterprisecustomercatalog',
```
brittneyexline, line 14: it would be better to introduce this migration in a separate pr: this initial pr should focus on removing references to the field. see "How to Drop a column section" here: https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations

brittneyexline (commented):
logic looks good, just one note about how to safely drop the field

enterprise/migrations/0122_remove_sync_enterprise_catalog_query_bool.py
chavesj: I'm not sure it'll be that simple - check the test results from my initial commit which didn't include the migration https://github.com/edx/edx-enterprise/runs/1787640214 (`NOT NULL constraint failed`). Is there some sort of best-of-both-worlds scenario here?

chavesj: Given those failures it seems there will need to be 3 PRs/deployments:

1. Remove NOT NULL constraint on field
2. Remove all references to field
3. Perform migration

Lmk if there's anything I'm missing

adamstankiewicz: @chavesj On your initial commit, I believe you were seeing `NOT NULL constraint failed` error because the `sync_enterprise_catalog_query` field was removed from the model itself at the same time as removing its references. If this PR keeps that field in tact for, we should be able to just remove the references to it first.

If your initial commit did not remove the `sync_enterprise_catalog_query` field from the model and only removed its references throughout the code first, that should work. Doubling back to remove the field itself alongside creating the migration would then be done as a separate PR.

adamstankiewicz (approved):

PR #1132, opened 2021-01-29
SAP Bug Fix: Handle Different Error Format.

taliaedX:
SAP's documentation (and our code) dictate an error format in the response of:
```
{error: {code: '', message: ''}}
```
However, certain errors are getting sent simply as a string, such as: `error: 'invalidResponse'`
This wraps the logging we were doing to allow us to log those string errors as well.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/sap_success_factors/client.py
+++ b/integrated_channels/sap_success_factors/client.py
@@ -320,14 +320,24 @@ class SAPSuccessFactorsAPIClient(IntegratedChannelApiClient):  # pylint: disable
             return None
 
         if 'error' in sap_inactive_learners:
-            LOGGER.warning(
-                'SAP searchStudent API for customer %s and base url %s returned response with '
-                'error message "%s" and with error code "%s".',
-                self.enterprise_configuration.enterprise_customer.name,
-                self.enterprise_configuration.sapsf_base_url,
-                sap_inactive_learners['error'].get('message'),
-                sap_inactive_learners['error'].get('code'),
-            )
+            try:
+                LOGGER.warning(
+                    'SAP searchStudent API for customer %s and base url %s returned response with '
+                    'error message "%s" and with error code "%s".',
+                    self.enterprise_configuration.enterprise_customer.name,
+                    self.enterprise_configuration.sapsf_base_url,
+                    sap_inactive_learners['error'].get('message'),
+                    sap_inactive_learners['error'].get('code'),
+                )
+            except AttributeError:
+                LOGGER.warning(
+                    'SAP searchStudent API for customer %s and base url %s returned response with '
+                    'error message "%s" and with error code "%s".',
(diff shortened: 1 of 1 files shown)


binodpant:
JIRA ticket link?

integrated_channels/sap_success_factors/client.py
```diff
@@ -320,14 +320,24 @@ def _call_search_students_recursively(self, sap_search_student_url, all_inactive
...
+                LOGGER.warning(
+                    'SAP searchStudent API for customer %s and base url %s returned response with '
+                    'error message "%s" and with error code "%s".',
+                    self.enterprise_configuration.enterprise_customer.name,
+                    self.enterprise_configuration.sapsf_base_url,
+                    sap_inactive_learners['error'].get('message'),
+                    sap_inactive_learners['error'].get('code'),
+                )
+            except AttributeError:
+                LOGGER.warning(
```
binodpant, line 333: looks like this fixes the immediate issue. Is this because sap_inactive_learners sometimes does not contain `error` and `code` attributes?

binodpant (approved):
nothing blocking, just had a couple of clarifying questions/jira thought

integrated_channels/sap_success_factors/client.py
taliaedX: I specified my findings in the description, but afaict certain error conditions are not providing this dictionary format. They updated error handling, so it is possible they missed certain errors or something. I'm not sure. They also occasionally spit back garbled Java, but we're already handling a return of complete nonsense.

taliaedX:
@binodpant the ticket is the branch name. It is also linked back here in the ticket because of that. I double-checked our how we work doc, but we never really decided to standardize on adding a ticket link separately in the PR.

The ticket is here though: https://openedx.atlassian.net/browse/ENT-3567

integrated_channels/sap_success_factors/client.py
taliaedX: So to answer what I think you may be asking about it not returning a dictionary in general, no, it definitely does. It's that sometimes the _subdictionary_ in the error key is missing.

edx-abolger:
> @binodpant the ticket is the branch name. It is also linked back here in the ticket because of that. I double-checked our how we work doc, but we never really decided to standardize on adding a ticket link separately in the PR.
> 
> The ticket is here though: https://openedx.atlassian.net/browse/ENT-3567

I still have an action on this to make it automatic so you don't have to.

edx-abolger (approved):

taliaedX:
@edx-abolger I think that might have been in reference to the branch name linking to the ticket, which may or may not happen in every repo. I don't know if there is capability to have the other direction (ie a link in github from jira in the github by mentioning the github pr or branch or whatever)

PR #1135, opened 2021-02-01
Catching client errors in SAP IC.

alex-sheehan-edx:
Bringing SAP integrated channel's client exception handling up to date with the other channels. Adding status code safeguard check to the `is already transmitted` method.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/exporters/learner_data.py
+++ b/integrated_channels/integrated_channel/exporters/learner_data.py
@@ -272,9 +272,10 @@ class LearnerExporter(Exporter):
         # Fetch course details from the Course API, and cache between calls.
         course_details = None
 
+        enrollment_ids_to_export = [enrollment.id for enrollment in enrollment_queryset]
         generate_formatted_log(
             'Beginning export of enrollments: {enrollments}.'.format(
-                enrollments=list(enrollment_queryset.values()),
+                enrollments=enrollment_ids_to_export,
             ),
             channel_name=channel_name,
             enterprise_customer_identifier=self.enterprise_customer.name
@@ -415,7 +416,22 @@ class LearnerExporter(Exporter):
                 # failure of SSO or similar). In such a case, `get_learner_data_record`
                 # would return None, and we'd simply skip yielding it here.
                 for record in records:
+                    # Because we export a course and course run under the same enrollment, we can only remove the
+                    # enrollment from the list of enrollments to export, once.
+                    try:
+                        enrollment_ids_to_export.pop(enrollment_ids_to_export.index(enterprise_enrollment.id))
+                    except ValueError:
+                        pass
+
                     yield record
+
+        generate_formatted_log(
+            'Finished exporting enrollments. Skipped enrollments: {enrollments}.'.format(
+                enrollments=enrollment_ids_to_export,
+            ),
+            channel_name=channel_name,
+            enterprise_customer_identifier=self.enterprise_customer.name
+        )
 
     def get_learner_assessment_data_records(
             self,
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,12 @@ Change Log
 Unreleased
 ----------
 
+[3.17.20]
+---------
+
+* Adds better exception handling to the SAP integrated channels.
+* Adds better logging to the base transmission process in the integrated channels.
+
 [3.17.19]
 ---------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.17.19"
+__version__ = "3.17.20"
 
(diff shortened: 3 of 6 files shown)


integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -272,15 +272,15 @@ def export(self, **kwargs):  # pylint: disable=R0915
...
-        generate_formatted_log(
-            'Beginning export of enrollments: {enrollments}.'.format(
-                enrollments=list(enrollment_queryset.values()),
-            ),
-            channel_name=channel_name,
-            enterprise_customer_identifier=self.enterprise_customer.name
-        )
-
         for enterprise_enrollment in enrollment_queryset:
+            generate_formatted_log(
```
edx-abolger, line 276: I'm mildly concerned about this generating 10000 individual log messages on a job, but if this is how the other integrated channels functions log bulk jobs then it is fine.  If not, perhaps move the log message back to where it was but only log the ID instead of the entire queryset?  @jlajoie do you have logging best practices for edX anywhere?

alex-sheehan-edx: I needed a way to verify that the particular enrollment was being exported. Agreed on not loving logging each individual message but not sure of what could be the better way. Perhaps like I first said, we could log them all together but only the ID and not the rest of the enrollment objects attributes. My only thought is that we don't want to look through the whole list of enrollments another time just to build the log statement. Happy to do either way though if you or @jlajoie feel strongly.

alex-sheehan-edx:
in a similar train of thought- how do we feel about changing https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/transmitters/learner_data.py#L156 from `debug` to `info`? That way we can confirm that our "trouble transmission" is either making it through the exporter (passed DSC and the like) or not.

alex-sheehan-edx:
> in a similar train of thought- how do we feel about changing https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/transmitters/learner_data.py#L156 from `debug` to `info`? That way we can confirm that our "trouble transmission" is either making it through the exporter (passed DSC and the like) or not.

We can hopefully add formatted logs here at some point but am leaving it out for now as that's a different ticket.

integrated_channels/integrated_channel/exporters/learner_data.py
jlajoie: I don't think we have actual best practices around logging here, but I think as long as we log only things are actionable that's what is important. This seems like we're just wanting to log the keys here to verify that they've run? Are there other identifiers we could use to say "hey this didn't run", like the modified timestamp in MySQL or something similar? Because if everything is green these logs kind of go to waste and we won't use them. 

That being said, I'm pretty sure our log storage is cheap, and if this can be valuable in some instances this is probably fine to leave in.

binodpant: Suggestion: may not work well but let's see:

+1 to not logging all items! 
While there is no silver bullet here, to address Alex's case, would the following work?

* log total enrollment count, and total exported enrollment count outside the loop (but collect the 'failed' item ids while in your loop). so logging could be like `{x} items exported of {y}. Here are the skipped enrollment ids : { ... }`

beyond this there is a larger discussion around how we may be able to better track 'lost' exports...

alex-sheehan-edx: Logging is useful here because the exporter is such a mess of conditionals, it's difficult to track what happens with one particular job. Logging the beginning isn't necessarily super important, however what about logging that a particular job has completed the export and is beginning the transmission? Since there are a number of ways to fail (sometimes silently) in the exporter, having an easily searchable log for completing the export/starting the transmission, I think, would be handy. How would you all feel about removing the `Beginning export` log and changing the debug on https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/transmitters/learner_data.py#L156 to info/a formatted log?

binodpant: `the exporter is such a mess of conditionals` big +1 to this!. Yeah I am also mainly just against the logging inside the loop. but if you think logging all the items at a job level makes life easier in the meantime, that is better than not logging. However, did we not lose info when we did that due to truncation? how do you plan to get around this if you log at the job level?

alex-sheehan-edx: hmm, @binodpant I like the idea of logging skipped enrollments.

alex-sheehan-edx: Although, we need to be careful. According to Adrienne and George, we should be careful in logging skipped transmissions.

binodpant (approved):
ids are better than logging entire objects and exception handling is better!

binodpant (approved):

PR #1142, opened 2021-02-08
ENT-4085- add default field in enterprise customer identity provider table

zamanafzal:
Add default field in enterprise customer identity provider table.
Change can be seen locally at the following URL
http://localhost:18000/admin/enterprise/enterprisecustomer/enterprisecustomer_uuid/change

Screenshot 2021-02-08 at 1 57 09 PM

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -1456,6 +1456,10 @@ class EnterpriseCustomerIdentityProvider(TimeStampedModel):
         unique=True,
         help_text="Slug field containing a unique identifier for the identity provider.",
     )
+    default = models.BooleanField(
```
sameenfatima78, line 1459: Wouldn't we need to make any code changes in the enterprise SSO login flow to make decisions based on this flag?

sameenfatima78 (approved):
LGTM, just one question

HammadAhmadWaqas:
how would we make sure users would not select more than one IDP as default?

zamanafzal:
@HammadAhmadWaqas This is something that users will not do but the enterprise admin and we need to educate them. But yes we can discuss this in standup tomorrow.

enterprise/models.py
zamanafzal: @sameenfatima78 That is part of another ticket.

enterprise/admin/forms.py
```diff
@@ -522,6 +523,20 @@ def clean(self):
                 ),
             )
 
+        # There can be only default provider in case an enterprise has multiple providers.
+        if default_provider is True:
```
HammadAhmadWaqas, line 527: can we just write `if default_provider`

enterprise/admin/forms.py
```diff
@@ -522,6 +523,20 @@ def clean(self):
...
+            default_provider_forms = int(self.data.get('enterprise_customer_identity_providers-TOTAL_FORMS'))
+            default_provider_values = [self.data.get('enterprise_customer_identity_providers-' +
```
HammadAhmadWaqas, line 529: as list comprehension has been spread to multiple lines should not we use simple for loop?

enterprise/admin/forms.py
```diff
@@ -522,6 +523,20 @@ def clean(self):
...
+                                                    str(i) + '-default_provider')
```
HammadAhmadWaqas, line 530: can we use a formatted string here for better readability?

enterprise/admin/forms.py
```diff
@@ -522,6 +523,20 @@ def clean(self):
...
+                                      for i in range(default_provider_forms)]
+            providers_marked_default = default_provider_values.count('on')
+            if providers_marked_default>1:
```
HammadAhmadWaqas, line 533: space is missing from both side of `>` operator

HammadAhmadWaqas: after updating this snippet will look like this.
```
        if default_provider:
            identity_provider_total_forms = int(self.data.get('enterprise_customer_identity_providers-TOTAL_FORMS'))
            default_provider_values = []
            for form_num in range(identity_provider_total_forms):
                default_provider_values.append(
                    self.data.get(f'enterprise_customer_identity_providers-{form_num}-default_provider')
                )
            providers_marked_default = default_provider_values.count('on')
            if providers_marked_default > 1:
                raise ValidationError(
                    {
                        'Please select only one default provider.'
                    }
                )
```

HammadAhmadWaqas (approved):
your code is working, a few small changes if you consider it right we can make.

enterprise/admin/forms.py
zamanafzal: In that case, we need to create an empty list and then append the data in that. I think it looks cleaner than that solution.

PR #1152, opened 2021-02-11
mgmt command to backfill enterprise role assignment data

iloveagent57:
https://openedx.atlassian.net/browse/ENT-4099

Adds a management command for updating enterprise admin and learner role assignments
with a foreign key to an EnterpriseCustomer, or an explicit boolean 
flag indicating that the assignment grants the role to the user
for every EnterpriseCustomer.

**Testing Instructions**
First, make sure to apply migrations for `enterprise` - we've added a config model that can define the arguments passed to this new management command.  This is generally how we parameterize management commands that run via Jenkins and typically need to accept arguments.

Do a dry run for admin assignments of some customer: 
```
./manage.py lms update_role_assignments_with_customers --role=enterprise_admin --dry-run --enterprise-customer-uuid=00000000-1111-2222-3333-444444444444
```
Do a real run for all operators: 
```
./manage.py lms update_role_assignments_with_customers --role=enterprise_openedx_operator
```
Process everything for everyone:
```
./manage.py lms update_role_assignments_with_customers
```
Undo the changes: 
```
./manage.py lms update_role_assignments_with_customers --undo
```

Define some arguments in the database now at http://localhost:18000/admin/enterprise/updateroleassignmentswithcustomersconfig/

Run the command using arguments from from that config model:
```
./manage.py lms update_role_assignments_with_customers --args-from-database
```

**Merge checklist:**

**Post merge:**


enterprise/management/commands/update_role_assignments_with_customers.py
```diff
@@ -0,0 +1,338 @@
+"""
+Management command for updating enterprise user role assignments.
+There are better words below.
```
kellmarie, line 3: I don't think you need the "There are better words below." sentence lol

enterprise/management/commands/update_role_assignments_with_customers.py
```diff
@@ -0,0 +1,338 @@
...
+        """
+        Does all of the actual work on assignments of a given role.
+        """
+        if role_name == ENTERPRISE_OPERATOR_ROLE:
+            self._handle_operators(dry_run)
+            return
+
+        ent_customer_user_ids = self._get_all_enterprise_customer_user_ids(enterprise_customer_uuid)
+
+        log.info(f'There are {len(ent_customer_user_ids)} total enterprise customer user ids to process.')
```
kellmarie, line 142: why not change this string to also be 'string'.format so the pylint error doesn't need to be disabled?

enterprise/management/commands/update_role_assignments_with_customers.py
```diff
@@ -0,0 +1,338 @@
...
+        for ecu_ids_batch in batch(ent_customer_user_ids, batch_size=batch_size):
+            log.info('Processing batch of ECU ids between {} and {}'.format(ecu_ids_batch[0], ecu_ids_batch[-1]))
+
+            customers_by_user_id = self._get_customers_by_user_id(ecu_ids_batch)
+
+            assignments_by_user_id_role = self._get_role_assignments_by_user_id_and_role(
+                customers_by_user_id.keys(),
+                role_name,
+            )
+            log.info(
```
kellmarie, line 153: same here

enterprise/management/commands/update_role_assignments_with_customers.py
```diff
@@ -0,0 +1,338 @@
...
+        If `dry_run` is true, we don't perform a database update; instead, a message will be logged
+        indicating how many assignments would be updated.
+        """
+        updated_assignments = []
+        queryset = SystemWideEnterpriseUserRoleAssignment.objects.filter(role__name=ENTERPRISE_OPERATOR_ROLE)
+        for assignment in queryset:
+            assignment.applies_to_all_contexts = True
+            updated_assignments.append(assignment)
+
+        log.info(f'There are {len(updated_assignments)} operator role assignments to update.')
```
kellmarie, line 178: also here

enterprise/management/commands/update_role_assignments_with_customers.py
```diff
@@ -0,0 +1,338 @@
...
+        if role_name == ENTERPRISE_OPERATOR_ROLE:
+            self._handle_operators(dry_run)
+            return
+
+        ent_customer_user_ids = self._get_all_enterprise_customer_user_ids(enterprise_customer_uuid)
+
+        log.info(f'There are {len(ent_customer_user_ids)} total enterprise customer user ids to process.')
+
+        for ecu_ids_batch in batch(ent_customer_user_ids, batch_size=batch_size):
+            log.info('Processing batch of ECU ids between {} and {}'.format(ecu_ids_batch[0], ecu_ids_batch[-1]))
```
kellmarie, line 145: But not here? this one uses .format

enterprise/management/commands/update_role_assignments_with_customers.py
```diff
@@ -0,0 +1,338 @@
...
+                        role=roles_by_name[role_name],
+                        enterprise_customer=linked_customers_by_uuid[customer_uuid],
+                    )
+                )
+
+        if dry_run:
+            log.info('This is a dry run, no updates or creates will happen for this batch.')
+            return
+
+        log.info(
```
kellmarie, line 250: ^

enterprise/management/commands/update_role_assignments_with_customers.py
iloveagent57: The newer version of edx-lint actually allows f-strings, so I'll just delete the pylint pragma.

iloveagent57: ...but maybe we need to get on that new version first?  I'm putting the disable back in for now.

kellmarie (approved):

chavesj (approved):
LGTM

enterprise/config/models.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    Model that specifies parameters for the
+    ``update_role_assignments_with_customers`` management command.
+
+    .. no_pii:
+    """
+    class Meta:
+        app_label = 'enterprise'
+
+    role = models.TextField(
```
adamstankiewicz, line 23: nit: To prevent typos when writing out the name of a user role assignment, I'm wondering if this model field should provide a list of known choices/roles rather than an open text field?

While there is a check in the management command itself that this provided role is one of the expected valid roles, I feel this may cause some potential confusion if I run the management command / job with a typo'd role name within this `UpdateRoleAssignmentsWithCustomersConfig` configuration model.

adamstankiewicz (approved):
LGTM, though with a question/nit around the `role` field in the `UpdateRoleAssignmentsWithCustomersConfig` configuration model.

enterprise/config/models.py
iloveagent57: Oh, hmm, yeah that's a good idea.

PR #1157, opened 2021-02-17
allow linking the same learner to multiple enterprises

kellmarie:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -136,10 +136,6 @@ class ValidationMessages:
     NO_FIELDS_SPECIFIED = _(
         "Either \"Email or Username\" or \"CSV bulk upload\" must be "
         "specified, but neither were.")
-    PENDING_USER_ALREADY_LINKED = _(
-        "Pending user with email address {user_email} is already linked with another Enterprise {ec_name}, "
-        "you will be able to add the learner once the user creates account or other enterprise "
-        "deletes the pending user")
     USER_ALREADY_REGISTERED = _(
         "User with email address {email} is already registered with Enterprise "
         "Customer {ec_name}")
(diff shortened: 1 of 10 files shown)


enterprise/utils.py
```diff
@@ -1228,19 +1224,20 @@ def delete_data_sharing_consent(course_id, customer_uuid, user_email):
...
-        * Checks that it is not already linked to any Enterprise Customer
+        * Checks that it is not already linked to the provided Enterprise Customer
 
     Arguments:
         email (str): user email to link
+        enterprise_customer (EnterpriseCustomer): the enterprise customer to link the email to
         raw_email (str): raw value as it was passed by user - used in error message.
         message_template (str): Validation error template string.
-        ignore_existing (bool): If True to skip the check for an existing Enterprise Customer
+        raise_exception (bool): whether to raise an exception when an email is invalidated
```
iloveagent57, line 1240: Nit: I kind of like `ignore_existing` better, I think it more accurately represents what this method does.  If `validate_email()` fails, an exception is raised regardless of the value of `raise_exception`.  `raise_exception` only has an impact if there is an existing ECU record.  Maybe call it `raise_on_existing_link` instead, and update the docstring to indicate that this only has an impact if there is an existing ECU for the given email and customer.

iloveagent57 (approved):
This is a positive change, I only have the one tiny complaint mentioned below.  Nice job.

enterprise/utils.py
kellmarie: I feel like ignore_existing wasn't really good name because what it did was skip/raise an exception when an email was already existing. I feel like a name like ignore_existing would make more sense if it actually meant something like "is this email formatted correctly and if it is then return True regardless of whether or not it's currently linked to an enterprise." What this is currently doing is returning True if the email is formatted correctly and linked with an enterprise, and returning False if it's formatted corrected and not linked with an enterprise regardless of what `ignore_existing` is set to.

Overall I think it's named badly and what it returns should be more clear but that's a really large and high impact change to make right now.

adamstankiewicz (approved):
LGTM!

PR #1158, opened 2021-02-17
allowing for channel retrieval to filter for assessment reporting support.

alex-sheehan-edx:
Bulk assessment level data transmissions were being sent out to all channels, including those that don't support it.

Similarly single learner transmissions would be sent to unsupported channels.

Added a fix to support filtering on channel retrieval and unit testing to make sure newly supported channels are filtered.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/tasks.py
+++ b/integrated_channels/integrated_channel/tasks.py
@@ -164,7 +164,7 @@ def transmit_single_subsection_learner_data(username, course_run_id, subsection_
     # Starting Export. N customer is usually 1 but multiple are supported in codebase.
     for enterprise_customer_uuid in enterprise_customer_uuids:
         for channel in channel_utils.get_integrated_channels(
-                {'channel': None, 'enterprise_customer': enterprise_customer_uuid}
+                {'channel': None, 'enterprise_customer': enterprise_customer_uuid, 'assessment_level_support': True}
         ):
             integrated_channel = INTEGRATED_CHANNEL_CHOICES[channel.channel_code()].objects.get(pk=channel.pk)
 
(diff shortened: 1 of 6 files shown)


integrated_channels/integrated_channel/tasks.py
```diff
@@ -164,7 +164,7 @@ def transmit_single_subsection_learner_data(username, course_run_id, subsection_
     # Starting Export. N customer is usually 1 but multiple are supported in codebase.
     for enterprise_customer_uuid in enterprise_customer_uuids:
         for channel in channel_utils.get_integrated_channels(
-                {'channel': None, 'enterprise_customer': enterprise_customer_uuid}
+                {'channel': None, 'enterprise_customer': enterprise_customer_uuid, 'assessment_level': True}
```
taliaedX, line 167: This is really pedantic and in no way a blocker, but can we use either `assessment_level` or `assessment_level_support` consistently? I think it's pretty easy still to just search the actual method to see what the parameter name is and given `assessment_level` is contained within `assessment_level_support` it's still even searchable, but it might make the trail just a little bit easier to follow if the parameter name is the same. It's also an internal parameter afaict (ie we aren't sending "assessment_level" to Canvas) so I don't think there's any reason really for them to not match (unless I'm missing something, in which case ignore me!)

taliaedX (approved):
1 extremely minor comment, but nothing blocking. LGTM otherwise!

integrated_channels/integrated_channel/tasks.py
alex-sheehan-edx: hmm yeah, I'm all for consistency. I think this became desynced when I had a pylint error on variable naming conventions... Will see what I can do to make it consistent.

PR #1160, opened 2021-02-17
omitting integrated assessment level canvas grades from final grade

alex-sheehan-edx:
Jira ticket: https://openedx.atlassian.net/browse/ENT-4148

Since assessment level and completion reporting are both implemented independently of each other and both use the same Canvas assignment object system to post their data, if one or both aren't omitted from the learner's final grade then the points the user earned will be inconsistent.

Example of current behavior:
image

Because both assessment data and completion data are included in the user's final grade, we have the appropriate percentage, but the `Out of` points fraction has points from both.

There are two ways around this, we can either exclude the edx "final grade" or the assessment level grades from the user's final grade. As the assessments can be weighted inconsistently with the edX learner's total grade, and the "final grade" from edx is accurate to the student's score in the class, I have decided to omit the assessment level grades.

Expected behavior:
image 

Here, with the assessment level grades omitted, the learner is still able to view the amount of points they got on a particular subsection, but those points will not factor into the user's Canvas final grade.

It is important to note that with this change, the user won't be able to see in Canvas their overall grade in the course until they complete it on edX.

Unit testing is tricky to add as this is a field supplied to Canvas that has an effect on user experience **on Canvas**, not edX.
**Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,11 @@ Change Log
 Unreleased
 ----------
 
+[3.17.38]
+
+* Omitting assessment level reporting from integrated Canvas learners final grade to not have redundant reported points
+  between final grades and subsection grades.
+
 [3.17.37]
 ---------
 
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,5 +2,5 @@
 Your project description goes here.
 """
 
-__version__ = "3.17.37"
+__version__ = "3.17.38"
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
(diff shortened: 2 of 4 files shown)


taliaedX (approved):
LGTM. And yeah there isn't really much to test here, but one option is just making sure the `_handle_canvas_assignment_retrieval` method does have `is_assessment_grade=True` when using assessment reporting. It's hardcoded when called via create_assessment_reporting so I mean it _should_

alex-sheehan-edx:
> `_handle_canvas_assignment_retrieval` method does have `is_assessment_grade=True` when using assessment reporting. It's hardcoded when called via create_assessment_reporting so I mean it _should_

United tests added  ¯\_(ツ)_/¯

taliaedX (approved):
Still looks good. Thanks for adding some tests!

PR #1165, opened 2021-02-19
[ENT-4036] reorder some enrollment logic and fix some logs

adamstankiewicz:
https://openedx.atlassian.net/browse/ENT-4036

Ensure we always create an `EnterpriseCourseEnrollment` _after_ the regular `CourseEnrollment` is successfully created/updated. This may help with the `EnterpriseCourseEnrollments` existing without an associated `CourseEnrollment` record. Interestingly enough, of these affected enrollments, none of them have any corresponding data in the `CourseEnrollment` history table.

This PR also fixes a couple logger messages that weren't logging much useful for debugging purposes.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -971,20 +971,20 @@ class EnterpriseCustomerUser(TimeStampedModel):
 
         if enrolled_in_course and is_upgrading:
             LOGGER.info(
-                "[Enroll] Trying to upgrade the enterprise user [{learner_id}] in course [{course_run_id}] in "
+                "[Enroll] Trying to upgrade the enterprise customer user [{ecu_id}] in course [{course_run_id}] in "
                 "[{mode}] mode".format(
-                    learner_id=self.id,
+                    ecu_id=self.id,
                     course_run_id=course_run_id,
-                    mode=mode
+                    mode=mode,
                 )
             )
         if not enrolled_in_course:
             LOGGER.info(
-                "[Enroll] Trying to enroll the enterprise user [{learner_id}] in course [{course_run_id}] in "
+                "[Enroll] Trying to enroll the enterprise customer user [{ecu_id}] in course [{course_run_id}] in "
                 "[{mode}] mode".format(
-                    learner_id=self.id,
+                    ecu_id=self.id,
                     course_run_id=course_run_id,
-                    mode=mode
+                    mode=mode,
                 )
             )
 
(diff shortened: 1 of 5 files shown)


enterprise/models.py
```diff
@@ -971,64 +971,41 @@ def enroll(self, course_run_id, mode, cohort=None, source_slug=None, discount_pe
...
                     course_run_id=course_run_id,
-                    mode=mode
+                    mode=mode,
                 )
             )
             try:
                 enrollment_api_client.enroll_user_in_course(self.username, course_run_id, mode, cohort=cohort)
             except HttpClientError as exc:
                 succeeded = False
-                if created:
```
adamstankiewicz, line 1030: The `enterprise_course_enrollment` only gets created if the learner was enrolled in the course, creating/updating a `CourseEnrollment` record.

enterprise/models.py
```diff
@@ -1038,7 +1015,41 @@ def enroll(self, course_run_id, mode, cohort=None, source_slug=None, discount_pe
...
+                LOGGER.info(
+                    "[Enroll] LMS enrollment API succeeded for user {username} in course {course_run_id} in "
+                    " mode {mode}".format(
+                        username=self.username,
+                        course_run_id=course_run_id,
+                        mode=mode,
+                    )
+                )
+                try:
+                    enterprise_course_enrollment, created = EnterpriseCourseEnrollment.objects.get_or_create(
```
adamstankiewicz, line 1029: `EnterpriseCourseEnrollment` is only created is course enrollment was successful.

enterprise/tasks.py
```diff
@@ -49,8 +49,16 @@ def create_enterprise_enrollment(course_id, enterprise_customer_user_id):
         # We have made changes elsewhere to avoid this issue, but in the mean time
         # we believe a Source of ENROLLMENT_TASK is more clear.
 
-        EnterpriseCourseEnrollment.objects.create(
-            course_id=course_id,
-            enterprise_customer_user=enterprise_customer_user,
-            source=EnterpriseEnrollmentSource.get_source(EnterpriseEnrollmentSource.ENROLLMENT_TASK)
-        )
+        try:
+            EnterpriseCourseEnrollment.objects.get_or_create(
```
adamstankiewicz, line 53: Making this bit more defensible given the

https://github.com/edx/edx-enterprise/pull/1165/files#diff-d950991245f878f58adc157895d6431af4385d522ab994aa25726e9abc2b434dR1037-R1045

enterprise/views.py
```diff
@@ -1698,22 +1700,47 @@ def post(self, request, enterprise_uuid, course_id):
...
+            try:
+                client.enroll_user_in_course(
+                    request.user.username,
+                    course_id,
+                    selected_course_mode_name,
+                    cohort=cohort_name
+                )
+            except HttpClientError as exc:
+                succeeded = False
+                default_message = 'No error message provided'
```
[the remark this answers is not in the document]
adamstankiewicz: To be honest, I'm not sure! I decided to just continue using the convention we have in place for handling these HttpClientErrors elsewhere, which gracefully handles when an HttpClientError comes back without a message.

Figured it'd be safer to continue using what's in place than reinvent the wheel here. That said, the message "No error message provided" hasn't appeared in our Splunk logs at least in the past 7 days so I agree these HttpClientErrors are likely always coming back with a message attached.

I feel it makes sense to keep the exception handling defensive juuust in case there's some error while trying to parse the exception content.

kellmarie (approved):

enterprise/views.py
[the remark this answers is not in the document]
kellmarie: Works for me, better safe than sorry. Especially with something as important and enrollments.

enterprise/views.py
```diff
@@ -1698,22 +1700,47 @@ def post(self, request, enterprise_uuid, course_id):
...
+                try:
+                    error_message = json.loads(exc.content.decode()).get('message', default_message)
+                except ValueError:
+                    error_message = default_message
+                LOGGER.exception(
+                    'Error while enrolling user %(user)s: %(message)s',
+                    dict(user=self.user_id, message=error_message)
+                )
```
iloveagent57, line 1722: Does `str(exc)` not contain a representation of the `message`?

iloveagent57 (approved):

enterprise/views.py
adamstankiewicz: You would think it does, right? Here's an example log where this message is extracted out from the exception content in another place within the code:

```
2021-02-19 10:26:56,374 ERROR 27886 [enterprise.models] [user 36854525] [ip 27.122.14.83] models.py:1037 - Error while enrolling user 36854525: An error occurred while creating the new course enrollment for user 'Damiend836' in course 'course-v1:IITBombayX+Linux101.1x+1T2021'
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.8/site-packages/enterprise/models.py", line 1027, in enroll
    enrollment_api_client.enroll_user_in_course(self.username, course_run_id, mode, cohort=cohort)
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.8/site-packages/enterprise/api_client/lms.py", line 104, in inner
    return func(self, *args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.8/site-packages/enterprise/api_client/lms.py", line 229, in enroll_user_in_course
    return self.client.enrollment.post(
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.8/site-packages/slumber/__init__.py", line 167, in post
    resp = self._request("POST", data=data, files=files, params=kwargs)
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.8/site-packages/slumber/__init__.py", line 101, in _request
    raise exception_class("Client Error %s: %s" % (resp.status_code, url), response=resp, content=resp.content)
slumber.exceptions.HttpClientError: Client Error 400: https://courses.edx.org/api/enrollment/v1/enrollment
```

My understanding is that `LOGGER.exception` still prints out the stack trace even though we don't explicitly tell it to print with `str(exc)`; extracting out the underlying error message from `exc.content` to surface the message seems reasonable here as I think the main exception error/message is coming from slumber (i.e., "Client Error 400") which isn't as useful compared to the underlying error message.

iloveagent57: Ah, I missed the `LOGGER.exception` - yeah, that'll get you the stack trace.

PR #1166, opened 2021-02-19
Replaced jsonfield2 with jsonfield

iamsobanjaved:
We are going to replace jsonfield2 with jsonfield as the earlier one has archived and merged back in the latter one.

**Merge checklist:**

**Post merge:**


enterprise/migrations/0124_auto_20210219_0841.py
```diff
@@ -0,0 +1,38 @@
...
+        migrations.AlterField(
+            model_name='historicalenterprisecustomercatalog',
+            name='content_filter',
+            field=jsonfield.fields.JSONField(blank=True, default={}, dump_kwargs={'cls': jsonfield.encoder.JSONEncoder, 'ensure_ascii': False, 'indent': 4, 'separators': (',', ':')}, help_text="Query parameters which will be used to filter the discovery service's search/all endpoint results, specified as a Json object. An empty Json object means that all available content items will be included in the catalog.", load_kwargs={'object_pairs_hook': collections.OrderedDict}, null=True, validators=[enterprise.validators.validate_content_filter_fields]),
+        ),
+        migrations.AlterField(
+            model_name='historicalenterprisecustomercatalog',
+            name='enabled_course_modes',
+            field=jsonfield.fields.JSONField(default=enterprise.constants.json_serialized_course_modes, help_text='Ordered list of enrollment modes which can be displayed to learners for course runs in this catalog.'),
+        ),
```
mraarif, line 37: is there any explanation in release notes for these migrations being generated?

iamsobanjaved: I think this change is the cause of this migration.

> Remove default `help_text` string


https://github.com/rpkilby/jsonfield/blob/master/CHANGES.rst#v310-02222020

UsamaSadiq (approved):

adamstankiewicz (approved):
LGTM

PR #1171, opened 2021-02-25
ENT-4091: handle email_csv case and also multi line when pasting emails on multiple lines in textfield

binodpant:
JIRA: ENT-4091

Fixes:

- [x] Issue with email_csv not being properly processed even after parsing
- [x] Issue with multi line email list in text area not being processed correctly, added `splitlines()` to handle that,  it now works with one or more emails
- [x] priority is still given to 'email' when both 'email' and 'email_csv' are present
- [x] adds a message to the response that says `N learners enrolled` for debugging/reporting/testing etc
- [x] test updates

Testing notes:

Tested by

[1] typing in multiline emails in textarea and hitting enroll learners and checking enrollments were created
[2] uploading csv file that looked like attached and repeating
```
email
learner1@example.com
learner2@example.com
```

as per required format (email first one is used as a header and also dict key)

at http://localhost:1991/test-enterprise/admin/catalog_management?features=BULK_ENROLLMENT

checked frontend response too like '2 learners enrolled'

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -254,8 +254,6 @@ class EnterpriseCustomerViewSet(EnterpriseReadWriteModelViewSet):
                             already_linked_emails.append((email, already_linked.enterprise_customer))
                         elif email in emails:
                             duplicate_emails.append(email)
-                        else:
-                            emails.add(email)
             except ValidationError as exc:
                 errors.append(exc)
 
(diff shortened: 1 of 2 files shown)


binodpant:
will do changelog before merging to avoid multi rebases

enterprise/api/v1/views.py
```diff
@@ -229,13 +232,17 @@ def enterprise_learners(self, request, pk):
...
         )
         if serializer.is_valid(raise_exception=True):
-            singular_email = serializer.validated_data.get('email')
-            emails = set()
             already_linked_emails = []
             duplicate_emails = []
             errors = []
+            enrolled_count = 0
+
+            singular_email = serializer.validated_data.get('email')
```
edx-abolger, line 240: Since this no longer represents a singular email, can you please just remove this temp variable?

enterprise/api/v1/views.py
```diff
@@ -219,6 +219,9 @@ def course_enrollments(self, request, pk):
     def enterprise_learners(self, request, pk):
         """
         Creates a set of enterprise_learners by enrolling them in the specified course.
+        If request.data contains 'email' field it's considered prioritized
```
edx-abolger, line 222: This method is now very long. Is it possible to split it into a few smaller methods?  I am worried about legibility.

binodpant: did a couple of refactors to clean it up

enterprise/api/v1/views.py
```diff
@@ -245,10 +251,6 @@ def enterprise_learners(self, request, pk):
                     else:
                         if already_linked:
                             already_linked_emails.append((email, already_linked.enterprise_customer))
-                        elif email in emails:
```
binodpant, line 248: don't see why we would want to add email back to the emails set where it came from so removed

binodpant: unused duplicate_emails

binodpant: @jlajoie want to ensure I am not breaking anything by removing these here ...

binodpant:
also had the thought to not accept both 'email' and 'email_csv' but not doing that now since it may break MFE. Also Alex and I need to chat before any more changes to this endpoint

enterprise/api/v1/views.py
binodpant: looks good chatted offline

edx-abolger (approved):

PR #1176, opened 2021-03-01
Bulk subscription enrollments into multiple courses endpoint

alex-sheehan-edx:
**Jira**: https://openedx.atlassian.net/browse/ENT-4226

New endpoint similar to https://github.com/edx/edx-enterprise/blob/8045ae8e94012df377331f1f32e4ae228b2eb7f2/enterprise/api/v1/views.py#L219 except that it will take a list of course keys and instead of a string list of user emails, the new endpoint expects a dictionary of <user emails> keys and <license uuids> as the values and will enroll all users in all courses.

**Oddities to note:**
- All users are assumed to have either pending or active subscriptions associated with either their active or "pending" account.
- While [License Manager](https://github.com/edx/license-manager/pull/189) (the MFE that is planned on using this endpoint) will have transactional constraints (all or none). This endpoint, for now, does not have similar support. If a user fails the process, then all users before will have enrollments and  subscription enrollments, or pending enrollments, pending users and pending subscription enrollments. 

**Current Outstanding questions:**
- ?

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -1244,6 +1245,24 @@ def _complete_user_enrollment():
                         discount_percentage=enrollment.discount_percentage,
                         sales_force_id=enrollment.sales_force_id,
                     )
+
+                    # If the pending enrollment was using a license, generate a license enrollment alongside the
+                    # enterprise enrollment.
+                    if enrollment.license_uuid:
+                        source = EnterpriseEnrollmentSource.get_source(EnterpriseEnrollmentSource.ENROLLMENT_URL)
```
alex-sheehan-edx, line 1252: I'm not 100% sure what this source is, if anyone could enlighten me as to why we use it in the first bulk enrollment, I'd appreciate it 😄

binodpant: I think it is just to track who created this enrollment, sometimes it's via django admin sometimes via an api call etc.. not 100% clear how it's used

enterprise/api/v1/serializers.py
```diff
@@ -952,3 +952,43 @@ def validate(self, data):  # pylint: disable=arguments-differ
         if not data.get('email') and not data.get('email_csv'):
             raise serializers.ValidationError('Must include either email or email_csv in request.')
         return data
+
+# pylint: disable=abstract-method
+class EnterpriseCustomerBulkSubscriptionEnrollmentsSerializer(serializers.Serializer):
+    """
+    Serializes an email field and for bulk enrollment requests.
+    """
+    emails = serializers.DictField(
```
alex-sheehan-edx, line 961: I don't have much experience creating dictionary fields, I'm fairly certain I got this right but would love if someone could confirm that I've got this the way we want.

enterprise/api/v1/serializers.py
```diff
@@ -952,3 +952,43 @@ def validate(self, data):  # pylint: disable=arguments-differ
...
+                ("professional", _("Professional Education")),
+                ("no-id-professional", _("Professional Education (no ID)")),
+                ("credit", _("Credit")),
+                ("honor", _("Honor")),
+            ],
+            required=False,
+        ),
+        required=True,
+    )
+    reason = serializers.CharField(required=False)
```
alex-sheehan-edx, line 981: besides `courses` and `emails`, most of this is copied from the existing bulk enrollment endpoint.

enterprise/utils.py
```diff
@@ -1431,6 +1435,11 @@ def enroll_users_in_course(
                     course_id,
                     role=MANUAL_ENROLLMENT_ROLE,
                 )
+
+            if license_uuids:
+                # create licensed enrollment if a subscription license is supplied
+                enterprise_customer_user = get_enterprise_customer_user(user.id, enterprise_customer.uuid)
+                get_create_ent_enrollment(course_id, enterprise_customer_user, license_uuids.get(user).get(course_id))
```
alex-sheehan-edx, line 1442: if we have license uuids, this is a bulk subscription enrollment request and all users are assumed to have licenses for all requested courses.

enterprise/api/v1/views.py
```diff
@@ -213,6 +213,79 @@ def course_enrollments(self, request, pk):
 
         return Response(serializer.errors, status=HTTP_400_BAD_REQUEST)
 
+    @action(detail=True, methods=['post'], permission_classes=[permissions.IsAuthenticated])
+    @permission_required('enterprise.can_enroll_learners', fn=lambda request, pk: pk)
+    # pylint: disable=invalid-name,unused-argument
+    def enroll_learners_in_courses(self, request, pk):
+        """
+        Creates a set of licensed enterprise_learners by bulk enrolling them in the specified course.
```
binodpant, line 221: I would doc here some flavor of the non transctional-ness of this endpoint so we know what to expect when sending a request here. e.g, if some of the enrollments fail, we will still have enrolled a partial set of users, correct? stuff like that.

binodpant: also if it does fail to enroll some of the learners, will the caller be expected to know which learner+course enrollments failed, or just which learners? in the least I think a requirement is to send back the set of learners that failed to enroll, to the MFE. will the license manager endpoint handle that vs this endpoint?

enterprise/api/v1/views.py
```diff
@@ -213,6 +213,79 @@ def course_enrollments(self, request, pk):
...
+            data=request.data,
+            context={
+                'enterprise_customer': enterprise_customer,
+                'request_user': request.user,
+            }
+        )
+        if serializer.is_valid(raise_exception=True):
+            errors = []
+            enrolled_count = 0
+            license_uuids = serializer.validated_data.get('emails')
```
binodpant, line 234: perhaps it's just the naming but we get a field called `emails` which is assigned to `license_uuids`. not sure but is it best to rename the emails field to sth like `license_info` or something? what you have further down is clear . e.g, `course_run_keys = serializer.validated_data.get('courses')`

also I think even if you keep this as is, perhaps having a quick example json in the docstring will help make it easier to see this?

enterprise/utils.py
```diff
@@ -1431,10 +1477,16 @@ def enroll_users_in_course(
...
```
binodpant, line 1484: will this work? isn't the key to license_uuids supposed to be the email address?

enterprise/models.py
```diff
@@ -1244,6 +1250,24 @@ def _complete_user_enrollment():
                         discount_percentage=enrollment.discount_percentage,
                         sales_force_id=enrollment.sales_force_id,
                     )
+
+                    # If the pending enrollment was using a subscription license, generate a license enrollment
+                    # alongside the enterprise enrollment.
+                    if enrollment.license_uuid:
+                        source = EnterpriseEnrollmentSource.get_source(EnterpriseEnrollmentSource.ENROLLMENT_URL)
```
binodpant, line 1257: I see enrollment_source is alrady passed as a kwargs to this function when called elsewhere in the pr, do you think we need to re-fetch this here?

enterprise/api/v1/views.py
alex-sheehan-edx: A failure at this point is bad because learners are expected to be pre-vetted by License manager in terms of their eligibility to enroll in the desired courses. What are we supposed to do if an enrollment is successfully created, but a licensed enrollment isn't? Beyond experiencing an exception that bubbles up and completely crashes the request, we return the number of users enrolled in the number of courses, which would give the frontend an idea of what happened.

binodpant: yeah, I think eventually we have to be transactionally safe in this endpoint, but for this PR I am not too worried. If we do fail to create licesnsed enrollment objects, I wonder if re enrolling them at a later time again will work, will it? if so we are not too worried yet I would say

however if an attempt to enrol the same learner into the same course again fails, then we would get into a bad state

enterprise/utils.py
alex-sheehan-edx: `license_uuids.get(<user>).get(<course_id>)`
should be getting the license uuid from the structure:
```
                license_info: {
                    <user email>: {
                        <course_id>: <license UUID that I'm trying to fetch>                    },
                }
```
no?

enterprise/models.py
alex-sheehan-edx: Sorry, do you know where in this PR? I'm looking now.

alex-sheehan-edx: are you talking about how the enrollment from the pending enrollment has a source?

enterprise/api/v1/views.py
edx-abolger: Ok, so is there a test for this case? The test case being a failure that succeeds later?

enterprise/api/v1/views.py
```diff
@@ -213,6 +213,110 @@ def course_enrollments(self, request, pk):
...
+        if serializer.is_valid(raise_exception=True):
+            errors = []
+            enrolled_count = 0
+            license_uuids = serializer.validated_data.get('license_info')
+            course_run_keys = serializer.validated_data.get('courses')
+            users_missing_licenses = []
+
+            # Get a list of user emails from the license info.
+            emails = set(license_uuids.keys())
+            try:
```
edx-abolger, line 262: Why do you have 2 nested try/except loops here for the same error type? Please remove the outside one, I am pretty sure it doesnt do anything

enterprise/api/v1/views.py
```diff
@@ -213,6 +213,110 @@ def course_enrollments(self, request, pk):
...
+                for email in emails:
+                    try:
+                        validate_email_to_link(email, enterprise_customer, raise_exception=False)
```
edx-abolger, line 265: If this is raise_exception is false, where do you get an error? Will this be silently failing on us? Does it return the errors in a list we aren't reading back?

edx-abolger: Alternately, since you wrap it in a try/except, change the flag.

enterprise/api/v1/views.py
```diff
@@ -213,6 +213,110 @@ def course_enrollments(self, request, pk):
...
+                        if len(license_uuids[email]) != len(course_run_keys):
```
edx-abolger, line 266: This needs a comment. Why would the number of licenses assigned to a user need be the same as the number of courses you are trying to enroll them in?

enterprise/api/v1/views.py
edx-abolger: Please either always use try/except or always use raise_exception=False for consistency.

enterprise/api/v1/serializers.py
edx-abolger: Why can you not _extend_ the existing class? Are there any items from the existing endpoint that do NOT exist on this new one?

enterprise/api/v1/views.py
```diff
@@ -213,6 +213,110 @@ def course_enrollments(self, request, pk):
...
+    @permission_required('enterprise.can_enroll_learners', fn=lambda request, pk: pk)
+    # pylint: disable=invalid-name,unused-argument
+    def enroll_learners_in_courses(self, request, pk):
+        """
+        Creates a set of licensed enterprise_learners by bulk enrolling them in the specified course.
+
+        Note: This endpoint isn't transactional in that if an enrollment fails, there can potentially be a scenario
+        with a subset of users enrolled. Requested learners' eligibility to enroll are expected to be pre-vetted.
+
+        Expected params:
```
edx-abolger, line 226: Why have you chosen this endpoint structure? Is there a really good reason not have this properly flattened into a list of desired enrollments of the form: [{email: A, course_key: B, license_uuid: C}, {email: D, course_key: E, license_uuid: F}] ?  This design choice is hard to read and requires dictionary traversal to do something that should be done in a loop.

enterprise/api/v1/serializers.py
alex-sheehan-edx: we don't use the fields 
```
    email = serializers.CharField(required=False)
    email_csv = Base64EmailCSVField(required=False)
    course_run_key = serializers.CharField(required=False)
    course_mode = serializers.ChoiceField(
```
None of those are required by default, but I feel it would be confusing to have seeming duplicates in a serializer class, no?

enterprise/api/v1/views.py
```diff
@@ -213,6 +213,110 @@ def course_enrollments(self, request, pk):
...
+                    try:
+                        validate_email_to_link(email, enterprise_customer, raise_exception=False)
+                        if len(license_uuids[email]) != len(course_run_keys):
+                            users_missing_licenses.append(email)
+                    except ValidationError as error:
+                        errors.append(error)
+            except ValidationError as exc:
+                errors.append(exc)
+
+            if users_missing_licenses:
```
edx-abolger, line 273: Do our errors not require a standard return format? Can you use a string for one and a list for the other?

edx-abolger (changes requested):

enterprise/api/v1/views.py
alex-sheehan-edx: You highlight exactly what the small voice in my head was quietly nagging about. The parameter structure for the endpoint I think became unwieldy when I realized we needed more information. Allow me to do a refactor and we can assess a flattened structure.

enterprise/api/v1/views.py
alex-sheehan-edx: FWIW the email validation logic here is copied from the original bulk enrollment endpoint.

enterprise/api/v1/views.py
alex-sheehan-edx: FWIW the email validation logic here is copied from the original bulk enrollment endpoint.

enterprise/api/v1/views.py
alex-sheehan-edx: we would also have to include the course mode as a key in those sets of dictionaries.

enterprise/api/v1/views.py
```diff
@@ -213,6 +213,110 @@ def course_enrollments(self, request, pk):
...
+    def enroll_learners_in_courses(self, request, pk):
+        """
+        Creates a set of licensed enterprise_learners by bulk enrolling them in the specified course.
+
+        Note: This endpoint isn't transactional in that if an enrollment fails, there can potentially be a scenario
+        with a subset of users enrolled. Requested learners' eligibility to enroll are expected to be pre-vetted.
+
+        Expected params:
+            - license_info (Dict): user emails as keys and the license UUIDs associated with each of the courses the
+            user's being enrolled in as values. Note, each user is expected to have exactly one license per course
```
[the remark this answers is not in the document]
alex-sheehan-edx: The idea is that being a bulk enrollment, the admin enduser selects/enters a list of users and a list of courses from the frontend. They then expect to enroll all of those users in all of those courses. License manager then confirms that every user in that list has **at least** one valid license for each of the courses they being enrolled in. Therefore we need a mapping of 1:1 between licenses and courses. These licenses don't have to be unique, but they can be.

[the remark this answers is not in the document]
alex-sheehan-edx: although, that rule of `all users in all courses` is specific to this admin portal bulk enrollment behavior and I don't think there's any reason to restrict this endpoint to enrolling all users in all courses, simply many users in many courses. Therefore I can remove the restriction and comment, given that's the direction we want to head.

enterprise/api/v1/views.py
alex-sheehan-edx: A standardized format would be nice, but I haven't come across a required one. I'm down to discuss stricter behavioral patterns if you wish. Any suggestions?

enterprise/api/v1/views.py
alex-sheehan-edx: Will talk with Jeff about the original implementation and reasonings behind it.

enterprise/api/v1/views.py
edx-abolger: That's fine- over time we may find we actually have to add other properties as well, which gets un-readable unless the structure is very flat.

enterprise/api/v1/views.py
```diff
@@ -213,6 +213,110 @@ def course_enrollments(self, request, pk):
...
+        """
+        enterprise_customer = self.get_object()
+        serializer = serializers.EnterpriseCustomerBulkSubscriptionEnrollmentsSerializer(
+            data=request.data,
+            context={
+                'enterprise_customer': enterprise_customer,
+                'request_user': request.user,
+            }
+        )
+        if serializer.is_valid(raise_exception=True):
```
edx-abolger, line 253: Change needed: There is a pattern here I would like you to change to and start using in general (called the Bouncer Pattern https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement)  which is to invert your logic and _exit_ right here if this fails instead of nesting the next 80 lines of code. It makes it clearer what's going on.

 When you make this change, also remember to capture the serializer errors and send them back, which is missing in the call on line 318

enterprise/api/v1/views.py
edx-abolger: Ugh, another item for the architecture meeting.  I think part of my problems are happening because we normally try to have validation errors in the serializer but this is validating semantic content, not formatting problems. However, in an effort to be consistent with what the rest of the Django setup is doing, it looks returning a list of errors matches the Django REST pattern in use elsewhere.

enterprise/api/v1/views.py
```diff
@@ -213,6 +213,110 @@ def course_enrollments(self, request, pk):
...
+                      'enrolled in. Users missing licenses: {}'.format(users_missing_licenses)
+                return Response(msg, status=HTTP_400_BAD_REQUEST)
+
+            if errors:
+                return Response(errors, status=HTTP_400_BAD_REQUEST)
+
+            for email in emails:
+                models.EnterpriseCustomerUser.objects.link_user(enterprise_customer, email)
+
+            if list(emails):
```
edx-abolger, line 284: If emails isn't a list, you still return a 202. Also you know emails is a list because you loop through it right above, which would crash if it weren't a list. Why is this check here?

enterprise/api/v1/views.py
```diff
@@ -213,6 +213,110 @@ def course_enrollments(self, request, pk):
...
+                for course_run_key, mode in course_run_keys.items():
+                    discount = serializer.validated_data.get('discount', 0.0)
+                    enrollment_reason = serializer.validated_data.get('reason')
+                    succeeded, pending, _ = enroll_users_in_course(
```
edx-abolger, line 288: You are skipping the failed case here which you can't do: if it fails, we need to know about it and return the error. I know that this creates design difficulties, but there are Django transaction options we can talk about wrapping this loop in if we need to.

enterprise/api/v1/views.py
```diff
@@ -213,6 +213,110 @@ def course_enrollments(self, request, pk):
...
+                            users=succeeded + pending,
+                        )
+                    self._create_ecom_orders_for_enrollments(
+                        course_run_key,
+                        mode,
+                        discount,
+                        serializer.validated_data.get('salesforce_id'),
+                        succeeded,
+                    )
+            return Response(
```
edx-abolger, line 313: You need to be tracking failed as you go around the loop here same as you are successes.

enterprise/api/v1/views.py
```diff
@@ -213,6 +213,110 @@ def course_enrollments(self, request, pk):
...
+
+            if errors:
+                return Response(errors, status=HTTP_400_BAD_REQUEST)
+
+            for email in emails:
+                models.EnterpriseCustomerUser.objects.link_user(enterprise_customer, email)
+
+            if list(emails):
+                for course_run_key, mode in course_run_keys.items():
+                    discount = serializer.validated_data.get('discount', 0.0)
```
edx-abolger, line 286: Is there a strong reason not to use _enroll_users here instead of re-implementing the loop yourself?

enterprise/models.py
```diff
@@ -1244,6 +1250,24 @@ def _complete_user_enrollment():
...
+                        enterprise_course_enrollment = EnterpriseCourseEnrollment.objects.get(
```
edx-abolger, line 1258: I don't understand something here: does the enterprise_customer_user.enroll function called above not return the successful enrollment? I don't think you should have to query for it again, I think you should already have a reference to it. That would save you the call to source as well.

enterprise/utils.py
```diff
@@ -1381,6 +1388,42 @@ def enroll_user(enterprise_customer, user, course_mode, *course_ids, **kwargs):
     return succeeded
 
 
+def get_create_ent_enrollment(
```
edx-abolger, line 1391: Did anything here get modified or is this just a refactor move?

enterprise/utils.py
```diff
@@ -1390,6 +1433,7 @@ def enroll_users_in_course(
         enrollment_reason=None,
         discount=0.0,
         sales_force_id=None,
+        license_uuids=None,
```
edx-abolger, line 1436: Where do you actually use this modified function ? Did you modify it and not call it as future proofing?

enterprise/utils.py
edx-abolger: This comment doesn't belong in the PR: it's part of the design spec you've agreed on. Add it to the doc string parameters. Also, if you are saying "if any license_uuids are included, all of them in the bulk enrollment MUST have one, then you need a better error message to return: you are going to have to validate them all beforehand and if one user is missing a license, then stop the entire transaction.

edx-abolger: Also it looks like this isn't actually required based on the lines of code below it?

tests/test_enterprise/api/test_views.py
```diff
@@ -2696,6 +2700,93 @@ def test_bulk_enrollment(self, body, expected_code, expected_body):
             response_content_string = json.dumps(self.load_json(response.content))
             self.assertIn(expected_body, response_content_string)
 
+    @ddt.data(
+        # Validation failure cases
```
edx-abolger, line 2704: Given how complex the license_uuid logic you wrote is, please include validation cases with invalid license_uuid data

enterprise/utils.py
binodpant: yup what I had meant was is `user` above the email address or the entire user object? just maybe good to confirm.

enterprise/models.py
binodpant: yeah I was confused why we fetched this source here since it looks like source is being passed as a param too? https://github.com/edx/edx-enterprise/pull/1176/files#diff-e3af55e5dcd947b5e06b20381920b6c81bb4f189e2cceed0c6f2309c4051ff84R1487 I could be misreading this but worth a quick pass on correctness

enterprise/models.py
edx-abolger: Resolved: enroll uses the API instead of a DB call and does not return the enrollment)

enterprise/api/v1/views.py
binodpant: I like the simplification made by making the params a simple array of things.In fact, that 'thing' can probably use a good name! Should we call it something like `licensed_enrollment_request`? Having a name can really help us talk about what specifically we are speaking of. 

Naming rationale: It's a request for enrollment, it may or may not be fulfilled. It has a license. It may have other fields but that is fine.

alex-sheehan-edx: as of the time writing this comment, we're using `licenses_info`

enterprise/api/v1/views.py
```diff
@@ -213,6 +215,110 @@ def course_enrollments(self, request, pk):
 
         return Response(serializer.errors, status=HTTP_400_BAD_REQUEST)
 
+    @action(detail=True, methods=['post'], permission_classes=[permissions.IsAuthenticated])
+    @permission_required('enterprise.can_enroll_learners', fn=lambda request, pk: pk)
+    # pylint: disable=invalid-name,unused-argument
+    def enroll_learners_in_courses(self, request, pk):
+        """
+        Creates a set of licensed enterprise_learners by bulk enrolling them in the specified course. This endpoint is
```
binodpant, line 223: `in the specified course` -> `in all the specified courses` ?

enterprise/api/v1/views.py
```diff
@@ -213,6 +215,110 @@ def course_enrollments(self, request, pk):
...
+        try:
+            serializer.is_valid(raise_exception=True)
+        except ValidationError:
+            return Response(serializer.errors, status=HTTP_400_BAD_REQUEST)
+
+        errors = []
+        licenses_info = serializer.validated_data.get('licenses_info')
+
+        # Default subscription discount is 100%
+        discount = serializer.validated_data.get('discount', 100.00)
```
binodpant, line 271: I am not super clear on how we will use discount here, looks like it will always be 100, for now anyway

enterprise/api/v1/views.py
```diff
@@ -213,6 +215,110 @@ def course_enrollments(self, request, pk):
...
+        serializer = serializers.EnterpriseCustomerBulkSubscriptionEnrollmentsSerializer(
+            data=request.data,
+            context={
+                'enterprise_customer': enterprise_customer,
+                'request_user': request.user,
+            }
+        )
+        try:
+            serializer.is_valid(raise_exception=True)
+        except ValidationError:
```
binodpant, line 264: hmm, is catching ValidationError needed here, vs just checking the output is false? https://www.django-rest-framework.org/api-guide/serializers/#validation

binodpant: actually , hold on. Based on their doc 

```
# Return a 400 response if the data was invalid.
serializer.is_valid(raise_exception=True)
```

reads to me as if simply calling is_valid with that signature automatically take care of throwing back 400, or no? https://www.django-rest-framework.org/api-guide/serializers/#raising-an-exception-on-invalid-data

enterprise/api/v1/views.py
```diff
@@ -213,6 +215,110 @@ def course_enrollments(self, request, pk):
...
+                course_run['mode'],
+                discount,
+                serializer.validated_data.get('salesforce_id'),
+                existing_users,
+            )
+        if results['failures']:
+            return Response(results, status=HTTP_409_CONFLICT)
+        if results['pending']:
+            return Response(results, status=HTTP_202_ACCEPTED)
+        return Response(results, status=HTTP_200_OK)
```
binodpant, line 320: do we need two clauses and response codes for pending vs not? if no faliures, can we always to a 202_accepted?

enterprise/utils.py
```diff
@@ -1381,6 +1389,127 @@ def enroll_user(enterprise_customer, user, course_mode, *course_ids, **kwargs):
...
+            failures: [{ 'email': <email>, 'course_run_key': <key> } ... ]
+        }
+    """
+    results = {
+        'successes': [],
+        'pending': [],
+        'failures': [],
+    }
+    for licensed_user_info in licensed_users_info:
+        user = User.objects.filter(email=licensed_user_info['email']).first()
```
binodpant, line 1461: isn't there a more specific, find a user by email query instead? not clear if expect multiple User object per email....

enterprise/utils.py
```diff
@@ -1381,6 +1389,127 @@ def enroll_user(enterprise_customer, user, course_mode, *course_ids, **kwargs):
...
+
+                    enterprise_customer_user = get_enterprise_customer_user(user.id, enterprise_customer.uuid)
+                    get_create_ent_enrollment(
+                        licensed_user_info.get('course_run_key'),
+                        enterprise_customer_user,
+                        license_uuid=licensed_user_info.get('license_uuid')
+                    )
+                else:
+                    results['failures'].append({
+                        'email': licensed_users_info['email'], 'course_run_key': licensed_user_info['course_run_key']
```
binodpant, line 1485: maybe store the `licensed_user_info.get('course_run_key')` and use it in these two places?

enterprise/utils.py
```diff
@@ -1381,6 +1389,127 @@ def enroll_user(enterprise_customer, user, course_mode, *course_ids, **kwargs):
...
+                    })
+            else:
+                pending_user = enterprise_customer.enroll_user_pending_registration(
+                    licensed_user_info['email'],
+                    licensed_user_info['course_mode'],
+                    licensed_user_info['course_run_key'],
+                    enrollment_source=enterprise_enrollment_source_model().get_source(
```
binodpant, line 1492: guess we don't have a better source to use here for now than 'manual'?

tests/test_enterprise/api/test_views.py
```diff
@@ -2696,6 +2700,226 @@ def test_bulk_enrollment(self, body, expected_code, expected_body):
...
+            'body': {
+                'licenses_info': [{
+                    'email': 'BADLYFORMATTEDEMAIL',
+                    'course_run_key': 'course-v1:edX+DemoX+Demo_Course',
+                    'course_mode': 'verified',
+                    'license_uuid': '5a88bdcade7c4ecb838f8111b68e18ac'
+                }]
+            },
+            'expected_code': 400,
+            'expected_response': [['BADLYFORMATTEDEMAIL does not appear to be a valid email address.']],
```
binodpant, line 2743: nice!

enterprise/api/v1/serializers.py
```diff
@@ -952,3 +952,39 @@ def validate(self, data):  # pylint: disable=arguments-differ
...
+    salesforce_id = serializers.CharField(required=False)
+    discount = serializers.DecimalField(None, 5, required=False)
+    notify = serializers.BooleanField(default=False)
+
+    def create(self, validated_data):
+        return validated_data
+
+    def validate(self, data):  # pylint: disable=arguments-differ
+        if not data.get('licenses_info'):
```
edx-abolger, line 977: why do you have to do this here? Why can you not make required=False at line 966 instead?

tests/test_enterprise/api/test_views.py
```diff
@@ -2696,6 +2700,226 @@ def test_bulk_enrollment(self, body, expected_code, expected_body):
...
+        response = self.client.post(
+            settings.TEST_SERVER + ENTERPRISE_CUSTOMER_BULK_ENROLL_LEARNERS_IN_COURSES_ENDPOINT,
+            data=json.dumps(body),
+            content_type='application/json',
+        )
+        self.assertEqual(response.status_code, expected_code)
+        if expected_response:
+            response_json = response.json()
+            self.assertEqual(expected_response, response_json)
+        self.assertEqual(len(PendingEnrollment.objects.all()), expected_num_pending_licenses)
```
binodpant, line 2868: just to be even stricter, always better to check this length before performing the operation too.. 0 is what we expect in this case I guess?

enterprise/api/v1/views.py
edx-abolger: Binod is right. It is likely just a copy paste error: there is a fair amount of code in the API that doesn't seem to leverage the Django Serializer base functionality.

tests/test_enterprise/api/test_views.py
```diff
@@ -2696,6 +2700,226 @@ def test_bulk_enrollment(self, body, expected_code, expected_body):
...
+                },
+            ]
+        }
+
+        response = self.client.post(
+            settings.TEST_SERVER + ENTERPRISE_CUSTOMER_BULK_ENROLL_LEARNERS_IN_COURSES_ENDPOINT,
+            data=json.dumps(body),
+            content_type='application/json',
+        )
+        self.assertEqual(response.status_code, 409)
```
binodpant, line 2920: SUGGEST: Using the status enums from rest_framework e.g., https://github.com/edx/edx-enterprise/blob/b55a50473775be18505565b99d7966389fb81ca9/tests/test_consent/api/test_permissions.py#L47

enterprise/api/v1/views.py
edx-abolger: This is fine, while it is almost always 100 for now it technically can be configured as something different in Django admin.

enterprise/api/v1/views.py
```diff
@@ -213,6 +215,110 @@ def course_enrollments(self, request, pk):
...
+
+        # Get a list of user emails from the license info.
+        emails = {license_info['email'] for license_info in licenses_info}
+        for email in emails:
+            try:
+                validate_email_to_link(email, enterprise_customer, raise_exception=False)
+            except exceptions.ValidationError as error:
+                errors.append(error)
+
+        if errors:
```
edx-abolger, line 281: I think we don't want this line at 281 anymore. If there are 2 bad emails in a batch of 100 it will return the 2 bad ones but won't complete the enrollments on the other 98.

enterprise/api/v1/views.py
```diff
@@ -213,6 +215,110 @@ def course_enrollments(self, request, pk):
...
+            return Response(errors, status=HTTP_400_BAD_REQUEST)
+
+        for email in emails:
```
edx-abolger, line 284: If you end up changing line 281 you'll need to change this line as well to skip the error emails

binodpant:
don't see linked in JIRA, maybe title needs to have ticket number?

enterprise/api/v1/serializers.py
alex-sheehan-edx: I was following the patterns in the other bulk enrollment endpoint. Happy to make the change if you think it's worth it.

enterprise/api/v1/views.py
alex-sheehan-edx: Yeah, I was trying to keep the endpoints consistent between `enterprise_learners` and `enroll_learners_in_courses` but I agree it's counter intuitive. Happy to change to proper handling.

enterprise/api/v1/views.py
alex-sheehan-edx: yup, the idea is that the value defaults to 100.00 but gives us the freedom to customize going forward if needed.

enterprise/api/v1/views.py
alex-sheehan-edx: doesn't match the work I did to allow for bad enrollments later on. 👍

alex-sheehan-edx: How would you like these returned along with the results of `enroll_licensed_users_in_courses`?

enterprise/api/v1/views.py
alex-sheehan-edx: I liked the idea of being able to know the state of the endpoints work just from the status code, but can go either way. @edx-abolger do you have an opinion here?

enterprise/api/v1/serializers.py
edx-abolger: Please go ahead and change it and make sure there's a test to cover your validation if it isn't there yet.

enterprise/utils.py
alex-sheehan-edx: I could probably use a `.get()` here, right? But also I don't think you can have multiple accounts to one email?

enterprise/api/v1/views.py
edx-abolger: Oh wait, I wonder if serializer.errors doesn't get returned as the whole message if you use the built in one. I would leave this alone as well.

enterprise/api/v1/views.py
edx-abolger: We're returning a list of error strings, correct? Or are we returning a list of emails? If it's the former, something simple like 'Invalid email address: {}' as the error string, appended to the existing error list should be fine.

enterprise/api/v1/views.py
```diff
@@ -213,6 +215,110 @@ def course_enrollments(self, request, pk):
...
+
+        # Retrieve all unique course key/course mode pairings from licenses_info
+        course_runs = [
+            {'key': course[0], 'mode': course[1]} for course in {
+                (info['course_run_key'], info['course_mode']) for info in licenses_info
+            }
+        ]
+        for course_run in course_runs:
+            pending_users = {
+                result.pop('user') for result in results['pending'] if result['course_run_key'] == course_run['key']
```
edx-abolger, line 297: This is fine for now, but check out https://www.geeksforgeeks.org/filter-in-python/ for the future

enterprise/api/v1/views.py
edx-abolger: I think the different state methods are fine but if you are going to make the status messages mean something then the 200 should be a 201 (created).

enterprise/api/v1/serializers.py
alex-sheehan-edx: Could you elaborate please? Are you saying I should keep the way in which we were validating the serializer?

edx-abolger: Yes.

enterprise/api/v1/views.py
alex-sheehan-edx: 👍  will check it out!

enterprise/api/v1/serializers.py
edx-abolger: I did some research and while the required=True thing might work, the 'list of objects' validation requires more finagling then it might actually be worth to make it happen automagically.

enterprise/api/v1/views.py
alex-sheehan-edx: Currently we are returning a dictionary of success, pendings, and failures. All arrays. I can add a `invalid_email` key to that dictionary if that works for you?

alex-sheehan-edx: extra field added. Let me know if you're good with naming conventions 👍

enterprise/utils.py
binodpant: yeah figured there was no way, since email should be a unique key for User I hope! but this works too, just want to check if there is a more optimal way here, sounds like get could work (if user is invalid it might just throw right away)

binodpant (approved):
minus a few small asks/checks, this is looking good to me, if we discover any issues in our LM or MFE testing we will ticket those out

binodpant:
and reminder to do the version changelog etc

enterprise/api/v1/views.py
```diff
@@ -213,6 +216,115 @@ def course_enrollments(self, request, pk):
...
+                    },
+                    ...
+                ]
+
+        Optional params:
+            - discount (int): the percent discount to be applied to all enrollments. Defaults to 100.
+
+        Expected Return Values:
+            Success cases:
+                - All users exist and are enrolled - [], 200
```
edx-abolger, line 248: Nit: this should now be 201

enterprise/utils.py
edx-abolger: @binodpant if you want to take 15 minutes to look at those source options and then file a ticket if we need more, thats fine

PR #1182, opened 2021-03-10
removing course mode from bulk enrollment required params 

alex-sheehan-edx:
fetch the course modes instead and select the best fitting.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -62,6 +62,7 @@ from enterprise.utils import (
     NotConnectedToOpenEdX,
     enroll_licensed_users_in_courses,
     enroll_users_in_course,
+    get_best_mode_from_course_key,
     get_ecommerce_worker_user,
     get_request_value,
     validate_email_to_link,
@@ -234,7 +235,6 @@ class EnterpriseCustomerViewSet(EnterpriseReadWriteModelViewSet):
                     {
                         'email': 'newuser@test.com',
                         'course_run_key': 'course-v1:edX+DemoX+Demo_Course',
-                        'course_mode': 'verified',
                         'license_uuid': '5b77bdbade7b4fcb838f8111b68e18ae'
                     },
                     ...
@@ -274,8 +274,17 @@ class EnterpriseCustomerViewSet(EnterpriseReadWriteModelViewSet):
         # Default subscription discount is 100%
         discount = serializer.validated_data.get('discount', 100.00)
 
-        # Get a list of user emails from the license info.
-        emails = {license_info['email'] for license_info in licenses_info}
+        emails = set()
+
+        # Retrieve and store course modes for each unique course provided
+        course_runs_modes = {license_info['course_run_key']: None for license_info in licenses_info}
+        for course_run in course_runs_modes:
+            course_runs_modes[course_run] = get_best_mode_from_course_key(course_run)
+
+        for index, info in enumerate(licenses_info):
+            emails.add(info['email'])
+            licenses_info[index]['course_mode'] = course_runs_modes[info['course_run_key']]
+
         for email in emails:
             try:
                 validate_email_to_link(email, enterprise_customer, raise_exception=False)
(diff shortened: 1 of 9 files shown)


tests/test_enterprise/api/test_views.py
```diff
@@ -2725,7 +2725,7 @@ def test_bulk_enrollment(self, body, expected_code, expected_body):
             'expected_response': {
                 'non_field_errors': [
                     "All license_info dicts must contain an email, course_run_key, course_mode and license_uuid. "
```
edx-abolger, line 2727: Please update the error message as well.

enterprise/api/v1/views.py
```diff
@@ -274,8 +273,20 @@ def enroll_learners_in_courses(self, request, pk):
...
 
-        # Get a list of user emails from the license info.
-        emails = {license_info['email'] for license_info in licenses_info}
+        enrollment_api_client = EnrollmentApiClient()
+
+        emails = set()
+        course_runs = {}
+        for index, info in enumerate(licenses_info):
+            emails.add(info['email'])
+            course_modes = [mode['slug'] for mode in enrollment_api_client.get_course_modes(info['course_run_key'])]
```
edx-abolger, line 282: Unfortunately as much as I want code reuse, this is going to heavily slow down your enrollment because you are now using the API client, and calling it individually for every single enrollment to do this. This is 100 API calls if you have 100 learners, even if you only have 1 course, to something you should be able to access from a database. I foresee this causing major problems.  Instead please do 2 things: (1) Add another loop before line 280. In this first loop, loop through your licenses and get all the courses. Use this to build your course_runs dictionary ahead of time and do NOT re-run the api call if you already have the course mode for a given course, since it doesn't vary.  (2) modify enrollment_api_client _in place_ to NOT make an HTTP call: have it directly make a database call. This will follow our general plan to incrementally rip out http calls where we should be using DB calls instead.

edx-abolger (changes requested):

enterprise/api/v1/views.py
```diff
@@ -274,8 +273,20 @@ def enroll_learners_in_courses(self, request, pk):
...
+            course_mode = 'verified' if 'verified' in course_modes \
+                else 'professional' if 'professional' in course_modes \
+                else 'no-id-professional' if 'no-id-professional' in course_modes \
+                else 'audit'
```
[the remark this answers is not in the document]
alex-sheehan-edx: no, we can probably throw this into a util function.

enterprise/api/v1/views.py
```diff
@@ -313,8 +318,8 @@ def enroll_learners_in_courses(self, request, pk):
                 )
 
             self._create_ecom_orders_for_enrollments(
-                course_run['key'],
-                course_run['mode'],
+                course_run,
+                course_runs[course_run],
```
binodpant, line 322: if this actually represents the course_mode, I would just call it that here.. not obvious from this.. but have to read the function def which says 

```
                                            course_run_key,
                                            mode,
                                            discount,
                                            salesforce_id,
                                            succeeded_enrollments
```

not sure if this works but is renaming `course_runs` to `course_modes` going to help? looks like `course_runs` is now different from what it was and only holds course modes?

binodpant:
@alex-sheehan-edx I did update the diagram at https://miro.com/app/board/o9J_lRc201I=/ for this but if I missed anything lmk

enterprise/api/v1/views.py
```diff
@@ -274,8 +273,20 @@ def enroll_learners_in_courses(self, request, pk):
         # Default subscription discount is 100%
         discount = serializer.validated_data.get('discount', 100.00)
 
-        # Get a list of user emails from the license info.
-        emails = {license_info['email'] for license_info in licenses_info}
+        enrollment_api_client = EnrollmentApiClient()
+
+        emails = set()
+        course_runs = {}
+        for index, info in enumerate(licenses_info):
```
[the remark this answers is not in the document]
alex-sheehan-edx: I was finding that `license_info` was too similar to `licenses_info` and distinguishing between them became tedious

enterprise/api/v1/views.py
binodpant: yeah I had the same flag earlier, I just was not sure if we wanted to do it as part of this PR or we had a ticket, but we might as well take care of it now, edx-enterprise calling api to get enrollments is not great

alex-sheehan-edx: update 🤗

enterprise/api/v1/views.py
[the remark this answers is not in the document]
binodpant: haha no worries, usually the s at the end helps with plural vs singular. license_infos vs license_info works too, and we can sometimes break english rules as programmers. But not too worried of this one right now

tests/test_enterprise/api/test_views.py
```diff
@@ -2724,8 +2724,8 @@ def test_bulk_enrollment(self, body, expected_code, expected_body):
             'expected_code': 400,
             'expected_response': {
                 'non_field_errors': [
-                    "All license_info dicts must contain an email, course_run_key, course_mode and license_uuid. "
-                    "Missing fields: ['course_mode', 'license_uuid']"
+                    "All license_info dicts must contain an email, course_run_key and license_uuid. "
+                    "Missing fields: ['license_uuid']"
```
binodpant, line 2728: 🎉

tests/test_enterprise/api/test_views.py
```diff
@@ -2857,6 +2851,7 @@ def test_bulk_enrollment_in_bulk_courses(
 
         permission = Permission.objects.get(name='Can add Enterprise Customer')
         self.user.user_permissions.add(permission)
+        mock_get_course_mode.return_value = 'verified'
```
binodpant, line 2854: small ask to make the 'verified' a test file level constant

binodpant: or import from somewhere if it exists

binodpant (approved):
I can't spot anything else to change here! made a small ask but looking great to me

tests/test_enterprise/api/test_views.py
alex-sheehan-edx: I added it to the file 😄

edx-abolger (approved):

PR #1185, opened 2021-03-15
Refactoring django serializers and bug clean up of bulk enrollment

alex-sheehan-edx:
[Jira](https://openedx.atlassian.net/browse/ENT-4313) bugbash 
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -954,17 +954,35 @@ class EnterpriseCustomerBulkEnrollmentsSerializer(serializers.Serializer):
         return data
 
 
+class LicensesInfoSerializer(serializers.Serializer):
+    """
+    Nested serializer class to allow for many license info dictionaries.
+    """
+    email = serializers.CharField(required=False)
+    course_run_key = serializers.CharField(required=False)
+    license_uuid = serializers.CharField(required=False)
+
+    def create(self, validated_data):
+        return validated_data
+
+    def validate(self, data):  # pylint: disable=arguments-differ
+        missing_fields = []
+        for key in self.fields.keys():
+            if not data.get(key):
+                missing_fields.append(key)
+
+        if missing_fields:
+            raise serializers.ValidationError('Found missing licenses_info field(s): {}.'.format(missing_fields))
+
+        return data
+
+
 # pylint: disable=abstract-method
 class EnterpriseCustomerBulkSubscriptionEnrollmentsSerializer(serializers.Serializer):
     """
     Serializes a licenses info field for bulk enrollment requests.
     """
-    licenses_info = serializers.ListField(
-        child=serializers.DictField(
-            child=serializers.CharField(required=False),
-        ),
-        required=False,
-    )
+    licenses_info = LicensesInfoSerializer(many=True, required=False)
     reason = serializers.CharField(required=False)
     salesforce_id = serializers.CharField(required=False)
     discount = serializers.DecimalField(None, 5, required=False)
@@ -974,17 +992,8 @@ class EnterpriseCustomerBulkSubscriptionEnrollmentsSerializer(serializers.Serial
         return validated_data
 
     def validate(self, data):  # pylint: disable=arguments-differ
-        if not data.get('licenses_info'):
-            raise serializers.ValidationError('Must include the "license_info" parameter in request.')
-
-        # validate that each license info has the required keys
-        for license_info in data.get('licenses_info'):
-            required_info = {'email', 'course_run_key', 'license_uuid'}
-            if not set(license_info.keys()) == required_info:
-                missing_fields = list(required_info - set(license_info.keys()))
-                missing_fields.sort()
-                raise serializers.ValidationError(
-                    'All license_info dicts must contain an email, course_run_key and license_uuid. '
-                    'Missing fields: {}'.format(missing_fields)
-                )
+        if data.get('licenses_info') is None:
+            raise serializers.ValidationError(
+                'Must include the "licenses_info" parameter in request.'
+            )
         return data
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -245,8 +245,10 @@ class EnterpriseCustomerViewSet(EnterpriseReadWriteModelViewSet):
 
         Expected Return Values:
             Success cases:
-                - All users exist and are enrolled - [], 201
-                - Some or none of the users exist but are enrolled - [], 202
+                - All users exist and are enrolled -
+                    {'successes': [], 'pending': [], 'failures': []}, 201
+                - Some or none of the users exist but are enrolled -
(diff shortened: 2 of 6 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -954,17 +954,38 @@ def validate(self, data):  # pylint: disable=arguments-differ
...
+    email = serializers.CharField(required=False)
+    course_run_key = serializers.CharField(required=False)
+    license_uuid = serializers.CharField(required=False)
+
+    def create(self, validated_data):
+        return validated_data
+
+    def validate(self, data):  # pylint: disable=arguments-differ
+        missing_fields = []
+        LOGGER.info("MEMES MEMES MEMES MEMES")
```
binodpant, line 970: need to remove these logs?

alex-sheehan-edx: you can see me getting frustrated at debugging haha

enterprise/api/v1/serializers.py
```diff
@@ -974,17 +992,8 @@ def create(self, validated_data):
...
-            if not set(license_info.keys()) == required_info:
-                missing_fields = list(required_info - set(license_info.keys()))
-                missing_fields.sort()
-                raise serializers.ValidationError(
-                    'All license_info dicts must contain an email, course_run_key and license_uuid. '
-                    'Missing fields: {}'.format(missing_fields)
-                )
+        if data.get('licenses_info') is None:
+            raise serializers.ValidationError(
+                'Must include the "license_info" parameter in request.'
```
binodpant, line 997: `license_info` -> `licenses_info`

enterprise/api/v1/serializers.py
```diff
@@ -954,17 +954,35 @@ def validate(self, data):  # pylint: disable=arguments-differ
...
     """
     Serializes a licenses info field for bulk enrollment requests.
     """
-    licenses_info = serializers.ListField(
-        child=serializers.DictField(
-            child=serializers.CharField(required=False),
-        ),
-        required=False,
-    )
+    licenses_info = LicensesInfoSerializer(many=True, required=False)
```
binodpant, line 985: is using required=True here better than checking validation down below like you added? or did I miss sth? not sure if we get the same result .... but yours certainly have a custom message ... so worth checking which is better

enterprise/api/v1/views.py
```diff
@@ -308,11 +311,12 @@ def enroll_learners_in_courses(self, request, pk):
             existing_users = {
                 result.pop('user') for result in results['successes'] if result['course_run_key'] == course_run
             }
+            LOGGER.info("Successfully bulk enrolled learners: {}".format(pending_users | existing_users))
```
binodpant, line 314: what does the pipe do here? did you intend to + instead?

binodpant: nvm I think I missed they were dicts

enterprise/api/v1/serializers.py
```diff
@@ -954,17 +954,35 @@ def validate(self, data):  # pylint: disable=arguments-differ
...
+        return validated_data
+
+    def validate(self, data):  # pylint: disable=arguments-differ
+        missing_fields = []
+        for key in self.fields.keys():
+            if not data.get(key):
+                missing_fields.append(key)
+
+        if missing_fields:
+            raise serializers.ValidationError('Found missing licenses info field(s): {}.'.format(missing_fields))
```
binodpant, line 975: `licenses info` -> `licenses_info` to match exact param name would be great

binodpant: also, 'found missing' seems confusing. Perhaps just "missing parameter `licenses_info` in request" or similar?

enterprise/utils.py
```diff
@@ -1471,7 +1471,7 @@ def enroll_licensed_users_in_courses(enterprise_customer, licensed_users_info, d
         user = User.objects.filter(email=licensed_user_info['email']).first()
         try:
             if user:
-                succeeded = enroll_user(enterprise_customer, user_email, course_mode, course_run_key)
+                succeeded = enroll_user(enterprise_customer, user, course_mode, course_run_key)
```
binodpant, line 1474: aargh, I hate that the docstring for this method does not indicate types!! if you are willing to, pls update that with types too. https://github.com/edx/edx-enterprise/blob/8625b3a3753ea0c2315b1cd3c2a8521b45c306e9/enterprise/utils.py#L1346

binodpant:
not for a change in this PR but is this correct?

in docstrings I see for example, https://github.com/edx/edx-enterprise/blob/73bef386a3ffce2b6117321d71c66d4d13a45701/enterprise/api/v1/views.py#L249
```
- All users exist and are enrolled - [], 201
- ```

for success cases looks like it returns a body of `{}` and not `[]` or did I misread?
if so please update docstrings

File: enterprise/api/v1/serializers.py (hunk 3 above)
Comment (alex-sheehan-edx) [reply enterprise/api/v1/serializers.py:985]: I was basing the pattern off of the existing single course bulk enrollment endpoint's serializer. I believe you're right in that it was done this way to handle customer messages as well as possibly the issue that Adrienne was investigating on the original PR I had for bulk enrollment. Do you have a preference here one way or the other?

File: enterprise/api/v1/views.py (hunk 4 above)
Comment (alex-sheehan-edx) [reply enterprise/api/v1/views.py:314]: you can't add sets together I've learned 🙃 the correct operation is to join them with the `|` character.

Comment (alex-sheehan-edx) [reply enterprise/api/v1/views.py:314]: ie 
```
Python 3.8.5 (default, Jul 21 2020, 10:48:26) 
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> setA = {'a'}
>>> setB = {'b'}
>>> setA | setB
{'b', 'a'}
>>> 
```
and 

```
>>> setA + setB
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'set' and 'set'
>>> 
```

enterprise/api/v1/serializers.py
binodpant: Nah I would not change it now. If the behavior you want is being met let's leave it. But semantically required=True would be nice. I just don't know enough right now to recommend setting it to true since you have a custom message. It seems that there is another way to get custom messages (https://www.django-rest-framework.org/api-guide/fields/#error_messages) which we can investigate if we want

enterprise/api/v1/views.py
binodpant: yup good one thanks so sets not dicts yes, thanks for verifying this

binodpant (approved):
looks good! please test locally as you can and include testing notes in the notes, but otherwise lgtm wrt fixing the 500 error we got. I will test it more thoroughly when we work on error handling later down the line too

binodpant:
version and changelog reminder

PR #1186, opened 2021-03-15
fix: ENT-3954 filter out unpublished courses from enrollment

binodpant:
With current logic, even unpublished courses are considered when deciding which course run a course should resolve to, when a user hits the enrollment landing page. This change takes out the unpublished (or more accurately course_runs with status != 'published' ) from consideration

There are no known use cases when unpublished courses should be considered for this page. See: https://edx-internal.slack.com/archives/CDXAQ8G02/p1615818761060100

This is expected to fix course `course-v1:NYIF+CR.3x+2T2020`

There are two other courses mentioned in the ticket, which may or may not be handled by this fix. But I would love to verify from stage/prod before making any more changes.

Also this causes older course runs like the one reported in the ticket to be picked up which may not even have content metadata and can lead to bad user experience.

With this change, if a user with no enrollments visits the page they should now be led to the more recent published course.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/tests/test_utilities.py
+++ b/tests/test_utilities.py
@@ -1541,6 +1541,36 @@ class TestEnterpriseUtils(unittest.TestCase):
 
     @ddt.data(
         (
+            # Test with two Current runs, but one not published. Should choose published one.
+            {
+                "course_runs": [
+                    fake_catalog_api.create_course_run_dict(
+                        availability="Current",
+                        start="2015-10-15T13:11:03Z",
+                        status="unpublished",
+                        end="2014-10-15T13:11:03Z",
+                        enrollment_start=None,
+                        enrollment_end=None,
+                    ),
+                    fake_catalog_api.create_course_run_dict(
+                        availability="Current",
+                        start="2015-10-15T13:11:03Z",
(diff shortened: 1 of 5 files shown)


tests/test_utilities.py
```diff
@@ -1540,6 +1540,36 @@ def test_is_course_run_about_to_end(self, course_run, expected_boolean):
...
+                "course_runs": [
+                    fake_catalog_api.create_course_run_dict(
+                        availability="Current",
+                        start="2015-10-15T13:11:03Z",
+                        status="unpublished",
+                        end="2014-10-15T13:11:03Z",
+                        enrollment_start=None,
+                        enrollment_end=None,
+                    ),
+                    fake_catalog_api.create_course_run_dict(
```
binodpant, line 1555: this is not an exact but close approx. of the case at https://discovery-internal.edx.org/api/v1/courses/NYIF+CR.3x/

edx-abolger (approved):

iloveagent57 (approved):
Looks good, just needs a changelog entry and version bump.

PR #1187, opened 2021-03-16
fix: The update_role_assignments_with_customers command no longer updates records

iloveagent57:
It only creates new records, which helps de-risk the operation.

See updates AC in https://openedx.atlassian.net/browse/ENT-4099 for explanation of future steps.


chavesj (approved):
LGTM

PR #1189, opened 2021-03-18
Add support for creating multiple pending users

raq929:
This PR also re-organizes tests; it may be best to review it commit by commit.
Much of the "new" code in the test file may just be code that is has been moved; please focus on the PendingEnterpriseCustomerUser stuff.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -400,7 +400,7 @@ class EnterpriseCustomerUserWriteSerializer(serializers.ModelSerializer):
             error_message = ('[Enterprise API] Saving to EnterpriseCustomerUser failed'
                              ' due to non-existing user. User: {}').format(value)
             LOGGER.error(error_message)
-            raise serializers.ValidationError(self.USER_DOES_NOT_EXIST)
+            raise serializers.ValidationError(self.USER_DOES_NOT_EXIST) from None
 
         return value
 
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -433,46 +433,34 @@ class PendingEnterpriseCustomerUserSerializer(serializers.ModelSerializer):
             'enterprise_customer', 'user_email'
         )
 
-    def validate(self, attrs):
-        """
-        Validate if the EnterpriseCustomerUser record already exists.
-        """
-        enterprise_customer = attrs.get('enterprise_customer')
-        user_email = attrs.get('user_email')
-        try:
-            user = User.objects.get(email=user_email)
-            models.EnterpriseCustomerUser.objects.get(
-                user_id=user.pk,
-                enterprise_customer=enterprise_customer
-            )
-        except (User.DoesNotExist, models.EnterpriseCustomerUser.DoesNotExist):
-            pass
-        else:
-            raise serializers.ValidationError('EnterpriseCustomerUser record already exists')
+    def to_representation(self, instance):
+        '''
+        Because we are returning whether or not the instance was created from the create method, we must use the
+        instance for to_representation and ignore the "created" half of the tuple
+        '''
+        return super().to_representation(instance[0])
 
-        return attrs
-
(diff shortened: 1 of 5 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -432,46 +433,34 @@ class Meta:
...
-                user_id=user.pk,
-                enterprise_customer=enterprise_customer
-            )
-        except (User.DoesNotExist, models.EnterpriseCustomerUser.DoesNotExist):
-            pass
-        else:
-            raise serializers.ValidationError('EnterpriseCustomerUser record already exists')
-
-        return attrs
+    def to_representation(self, instance):
```
alex-sheehan-edx, line 436: I'm assuming that Django handles this method? I haven't actually used `to_representation` before, although I feel like I could reasonably assume what it does?

alex-sheehan-edx (approved):
A couple of questions mostly for my own understanding and only v minor formatting. Otherwise LGTM

enterprise/api/v1/serializers.py
raq929: Yeah, it's a DRF method. It expects an instance of the model to be returned, and creates a JSON serializable representation of it. However, this serializer is very weird and sometimes creates one model, sometimes another, and sometimes nothing at all. And we want to know if something was created or not. So I'm having `create` return both the instance *and* whether or not a user was created, and then telling `to_representation` how to deal with that.

raq929: Because I have it open, here's what it does. https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L493

enterprise/api/v1/serializers.py
```diff
@@ -378,6 +378,7 @@ class EnterpriseCustomerUserWriteSerializer(serializers.ModelSerializer):
     """
     Serializer for writing to the EnterpriseCustomerUser model.
     """
+    USER_DOES_NOT_EXIST = "User does not exist"
```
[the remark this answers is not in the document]
raq929: Yeah, I'm trying to remove "magic strings" whereever possible, but don't have time atm for a broader assesment.

enterprise/api/v1/views.py
```diff
@@ -811,42 +811,40 @@ class PendingEnterpriseCustomerUserViewSet(EnterpriseReadWriteModelViewSet):
...
-
     queryset = models.PendingEnterpriseCustomerUser.objects.all()
     filter_backends = (filters.OrderingFilter, DjangoFilterBackend)
     serializer_class = serializers.PendingEnterpriseCustomerUserSerializer
     permission_classes = (permissions.IsAuthenticated, permissions.IsAdminUser)
 
     FIELDS = (
-        'enterprise_customer', 'user_email',
-    )
+          'enterprise_customer', 'user_email',
```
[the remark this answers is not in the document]
raq929: idk, I'll take a look at it.

enterprise/api/v1/views.py
```diff
@@ -811,42 +811,40 @@ class PendingEnterpriseCustomerUserViewSet(EnterpriseReadWriteModelViewSet):
...
-                        break
-
-            if not return_status:
-                raise validation_error
-        else:
-            created = serializer.save()
+    UNIQUE = 'unique'
+    USER_EXISTS_ERROR = 'EnterpriseCustomerUser record already exists'
+
+    def _get_return_status(self, serializer, many):
```
edx-abolger, line 828: opinion: This is a nice pattern. I'm not thrilled with the name since it actually does the saving as well, but I would be pleased to see the handle one-or-many DRF apis in more places.

edx-abolger (approved):

enterprise/api/v1/views.py
raq929: yeaaaah, it started out as one thing and ended up as another, I'm not thrilled with the name either. But yeah, I stole the serializer `many` pattern and it seems to be OK?

PR #1191, opened 2021-03-22
Add new endpoint for ent admins to link learners

raq929:
Allows admins to link learners (ie, create EnterpriseCustomerUsers or PendingEnterpriseCustomerUsers) for their enterprise

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -468,14 +468,17 @@ class LinkLearnersSerializer(PendingEnterpriseCustomerUserSerializer):
     Extends the PendingEnterpriseCustomerSerializer to validate that the enterprise customer uuid
     matches the uuid the user has permissions to update
     """
+    NOT_AUTHORIZED_ERROR = 'Not authorized for this enterprise'
 
-    def validate_enterprise_customer(self, data):
+    def validate_enterprise_customer(self, value):
         """
         Check that the enterprise customer is the same as the one the user has permissions for
+        The value recieved is an EnterpriseCustomer object
         """
-        if data['enterprise_customer'] != self.context.get('enterprise__uuid'):
-            raise serializers.ValidationError('Not authorized for this enterprise')
-        return data
+
+        if str(value.uuid) != self.context.get('enterprise_customer__uuid'):
+            raise serializers.ValidationError(self.NOT_AUTHORIZED_ERROR)
+        return value
 
 
 class CourseDetailSerializer(ImmutableStateSerializer):
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -463,6 +463,21 @@ class PendingEnterpriseCustomerUserSerializer(serializers.ModelSerializer):
         return new_user, created
 
 
+class LinkLearnersSerializer(PendingEnterpriseCustomerUserSerializer):
+    """
+    Extends the PendingEnterpriseCustomerSerializer to validate that the enterprise customer uuid
+    matches the uuid the user has permissions to update
+    """
+
+    def validate_enterprise_customer(self, data):
+        """
+        Check that the enterprise customer is the same as the one the user has permissions for
+        """
+        if data['enterprise_customer'] != self.context.get('enterprise__uuid'):
+            raise serializers.ValidationError('Not authorized for this enterprise')
+        return data
+
+
 class CourseDetailSerializer(ImmutableStateSerializer):
     """
     Serializer for course data retrieved from the discovery service course detail API endpoint.
--- a/enterprise/api/v1/urls.py
+++ b/enterprise/api/v1/urls.py
@@ -35,7 +35,7 @@ urlpatterns = [
     url(
         r'link_pending_enterprise_users/(?P<enterprise_uuid>[A-Za-z0-9-]+)/?$',
         views.PendingEnterpriseCustomerUserEnterpriseAdminViewSet.as_view({'post': 'link_learners'}),
-        name='link-pending-enteprise-learner'
+        name='link-pending-enterprise-learner'
     ),
     url(
         r'^enterprise_catalog_query/(?P<catalog_query_id>[\d]+)/$',
--- a/enterprise/api/v1/urls.py
+++ b/enterprise/api/v1/urls.py
@@ -33,6 +33,11 @@ router.register(
 
 urlpatterns = [
     url(
+        r'link_pending_enterprise_users/(?P<enterprise_uuid>[A-Za-z0-9-]+)/?$',
+        views.PendingEnterpriseCustomerUserEnterpriseAdminViewSet.as_view({'post': 'link_learners'}),
+        name='link-pending-enteprise-learner'
+    ),
+    url(
         r'^enterprise_catalog_query/(?P<catalog_query_id>[\d]+)/$',
         views.CatalogQueryView.as_view(),
         name='enterprise-catalog-query'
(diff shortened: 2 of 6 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -463,6 +463,21 @@ def create(self, attrs):  # pylint: disable=arguments-differ
...
+    """
+    Extends the PendingEnterpriseCustomerSerializer to validate that the enterprise customer uuid
+    matches the uuid the user has permissions to update
+    """
+
+    def validate_enterprise_customer(self, data):
+        """
+        Check that the enterprise customer is the same as the one the user has permissions for
+        """
+        if data['enterprise_customer'] != self.context.get('enterprise__uuid'):
```
edx-abolger, line 476: Do you need to handle the super case where the lack of a uuid means the user has permission to edit all? @iloveagent57 would have the expertise here.

enterprise/api/v1/urls.py
```diff
@@ -44,7 +49,7 @@
     ),
     url(
         r'^tableau_token$',
-        views.TableauAuthViewSet.as_view(),
+        views.TableauAuthView.as_view(),
```
edx-abolger, line 52: Why did this change?

enterprise/api/v1/views.py
```diff
@@ -1203,7 +1241,7 @@ def post(self, request):
             )
 
 
-class TableauAuthViewSet(generics.GenericAPIView):
+class TableauAuthView(generics.GenericAPIView):
```
edx-abolger, line 1244: Again, why name change? Rebase issue?

enterprise/api/v1/views.py
```diff
@@ -843,12 +844,49 @@ def _get_return_status(self, serializer, many):
         return status.HTTP_204_NO_CONTENT
 
     def create(self, request, *args, **kwargs):
+        """
+        If a Django user exists, but the correspinding EnterpriseUser does not, creates a PendingEnterpriseUser for
+        the email in question.
+        Can accept one user or a list of users.
+
+        Returns 201 if any users were created, 204 if no users were created.
```
edx-abolger, line 852: +1 for documentation

enterprise/api/v1/serializers.py
raq929: This method isn't defined on the parent case, so there's no need to call super here. DRF will run the regular validation and then call this method, is my understanding of it.

enterprise/api/v1/views.py
```diff
@@ -843,12 +844,49 @@ def _get_return_status(self, serializer, many):
...
+    Admin must be an administrator for the enterprise in question
+    """
+    permission_classes = (permissions.IsAuthenticated,)
+    serializer = serializers.LinkLearnersSerializer
+
+    @action(methods=['post'], detail=False)
+    @permission_required('enterprise.can_access_admin_dashboard', fn=lambda request, enterprise_uuid: enterprise_uuid)
+    def link_learners(self, request, enterprise_uuid):
+        """
+        If a Django user exists, but the correspinding EnterpriseUser does not, creates a PendingEnterpriseUser for
```
edx-abolger, line 873: Nit: 'corresponding'

enterprise/api/v1/urls.py
raq929: Ah, because `TableauAuthVew` is a `View` and not a `Viewset` and it Tricked me into believing that I could call `as_view()` on a `ViewSet`

enterprise/api/v1/views.py
edx-abolger, line 873: I have to ask, why does a Django user have to exist? I did not think a Django user had to exist for a pending enterprise customer to be created if we did it from Django admin?

enterprise/api/v1/views.py
raq929: Yeah, as mentioned above, making the name accurate.

tests/test_enterprise/api/test_views.py
```diff
@@ -767,6 +774,308 @@ def test_post_pending_enterprise_customer_user_logged_out(self):
...
+        )
+
+        response = self.client.post(
+            settings.TEST_SERVER + reverse('link-pending-enteprise-learner', kwargs={'enterprise_uuid': ent_uuid}),
+            data=data
+        )
+        assert response.status_code == status_code
+        response = self.load_json(response.content)
+        self.assertDictEqual(data, response)
+        if not user_exists:
```
edx-abolger, line 857: Did we want to start a trend where assertions within "if" statements are discouraged?  Is this easier to test as a few separate tests?

enterprise/api/v1/views.py
raq929: Right. But if the EnterpriseUser already exists, we don't create a new PendingEnterpriseUser for that Django user.

How about:
Creates a PendingEnterpriseUser if no User exists or if the User exists but no EnterpriseUser exists
Does not create a PendingEnterpriseUser if the EnterpriseUser already exists.

Hmmm... I don't love that either; suggestions welcome.

raq929: lol

tests/test_enterprise/api/test_views.py
raq929: Heh, this is *after* I took some ifs out of here. It's slightly annoying to do as separate tests. OH. WAIT. I HAVE IDEAS.

edx-abolger (commented):
Need a few questions answered, but possibly no changes if Alex D. can weigh in on case of a super user

enterprise/api/v1/views.py
iloveagent57: "Creates a `PendingEnterpriseCustomerUser` if no `EnterpriseCustomerUser` for the given (customer, email) combination(s) exists."

tests/test_enterprise/api/test_views.py
raq929: Nevermind, my ideas did not pan out. This one is trickier to pull apart; when I went through them the first time, I decided it wasn't worth it for this one.

enterprise/api/v1/views.py
raq929: Yeah, that seems accurate, thanks!

enterprise/api/v1/serializers.py
iloveagent57, line 476: I think you need `self.context.get('enterprise_customer__uuid')` in here, right?

iloveagent57 (commented):
Looking good, one suggestion for the serializer.

enterprise/api/v1/serializers.py
raq929: yeah, I think that might be a better name. Although, I'm confused about why my tests passed when I'm giving it.
```
context = {'enterprise_uuid': enterprise_uuid}
```
Looking into it now.

raq929: If the user is a superuser or staff user, they can use the original endpoint that LicenseManager is using now; it doesn't make this check. That way they can use it for multiple enterprises.

raq929: UPDATE: The tests are clearly not testing what I think they're testing. So that's fun.

edx-abolger: Got it, makes sense.

edx-abolger (approved):

iloveagent57 (approved):
Nice job with all those unit tests - very comprehensive and easy to follow.

PR #1196, opened 2021-03-26
ENT-2010: Improve efficiency of the data sharing consent flow

sameenfatima78:
JIRA: https://openedx.atlassian.net/browse/ENT-2010

**Merge checklist:**

**Post merge:**

### DIFF
--- a/consent/models.py
+++ b/consent/models.py
@@ -6,8 +6,6 @@ Models for edX Enterprise's Consent application.
 import logging
 
 import six
-from opaque_keys import InvalidKeyError
-from opaque_keys.edx.keys import CourseKey
 from simple_history.models import HistoricalRecords
 
 from django.core.exceptions import ImproperlyConfigured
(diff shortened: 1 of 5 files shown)


consent/models.py
```diff
@@ -44,39 +42,32 @@ def proxied_get(self, *args, **kwargs):
...
         if 'course_id' in kwargs:
             try:
-                # Check if we have a course ID or a course run ID
-                course_run_key = str(CourseKey.from_string(kwargs['course_id']))
-            except InvalidKeyError:
-                # The ID we have is for a course instead of a course run; fall through
-                # to the second check.
-                pass
-            else:
+                # Try to get the record for the course OR course run, depending on what we got in kwargs,
```
edx-abolger, line 48: You seem to have deleted the check about whether the item passed in is a course id or a course run id. This endpoint must handle both of those, that is why the extra logic about the course run key is here. Your logic does not handle a customer passing in a course run key. Please revert this and find a different way to clean up.

edx-abolger (changes requested):

consent/models.py
sameenfatima78: This endpoint would still handle both use cases. First it's checking whether a record exists against the course_id or course_run_id, whichever is passed, if it doesn't then it moves to the next step which is returning a record for course instead, if that too doesn't exist then it returns the proxy record. Please let me know if you think that's not the case here and why.

edx-abolger: I see: I assumed that `course_identifier=kwargs['course_id']` would fail in the case of a course run because you removed this formatter: `str(CourseKey.from_string(kwargs['course_id']))`.  I did not realize that the .get_course_id function would handle a course id as well, I see it already handles the logic you removed.

edx-abolger (approved):

muhammad-ammar (approved):
LGTM

PR #1199, opened 2021-03-29
ENT-4222 : replace REST api calls to LMS for certificate and grade data fetch

binodpant:
Replace api calls with utils/database calls to fetch grade and cert info

### Not including (see comments below) in this PR:

These two remaining api calls should clear out all LMS api calls from IC code base (leaving just api calls to LMS-external services):
- [ ] Call lms util instead of REST api to get assessment grades (current call is `grades_api.get_course_assessment_grades`)
- [ ] Call lms util instead of REST Api for `get_course_details` , this one uses the `CourseApiClient` whose usage can also be removed from this codebase

Reason:  I want to test this outgoing PR in staging with at least one LMS first then will add a new PR to handle this piece to avoid making this PR too large to test.

### Included work:

- [x] Gather up calls from integrated_channels exporter to lms into integrated_channels.lms_utils file
- [x] Gather up calls from integrated_channels exporter  to catalog service into integrated_channels.catalog_service_utils file
- [x] Call lms util instead of REST api to get certificate for user 
- [x] Calls lms util instead of RET api to implement `_collect_grades_data`
- [x] Refactor code for better readability and reduced complexity
- [x] Refactor out some logging calls to improve readability and better prep for future logging enhancements
- [x] Removes the already unused (`self.enterprise_api`) property that refers to `EnterpriseApiClient` in Exporter 

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/catalog_service_utils.py
+++ b/integrated_channels/catalog_service_utils.py
@@ -1,0 +1,14 @@
+"""
+A utility collection for calls from integrated_channels to Catalog service
+"""
+from enterprise.api_client.discovery import get_course_catalog_api_service_client
+
+
+def get_course_id_for_enrollment(enterprise_enrollment):
+    """
+    Fetch course_id for a given enterprise enrollment
+    """
+    course_catalog_client = get_course_catalog_api_service_client(
+        site=enterprise_enrollment.enterprise_customer_user.enterprise_customer.site
+    )
+    return course_catalog_client.get_course_id(enterprise_enrollment.course_id)
--- a/integrated_channels/catalog_service_utils.py
+++ b/integrated_channels/catalog_service_utils.py
@@ -7,6 +7,7 @@ from enterprise.api_client.discovery import get_course_catalog_api_service_clien
 def get_course_id_for_enrollment(enterprise_enrollment):
     """
     Fetch course_id for a given enterprise enrollment
+    Returns None if no course id found for course_id associated with input enterprise_enrollment
     """
     course_catalog_client = get_course_catalog_api_service_client(
         site=enterprise_enrollment.enterprise_customer_user.enterprise_customer.site
@@ -16,7 +17,8 @@ def get_course_id_for_enrollment(enterprise_enrollment):
 
 def get_course_run_for_enrollment(enterprise_enrollment):
     """
-    course_run associated with the enrollment
+    Fetch course_run associated with the enrollment
+    Returns empty dict {} if no course_run found for enterprise_enrollment.course_id
     """
     course_catalog_client = get_course_catalog_api_service_client(
         site=enterprise_enrollment.enterprise_customer_user.enterprise_customer.site
--- a/integrated_channels/catalog_service_utils.py
+++ b/integrated_channels/catalog_service_utils.py
@@ -12,3 +12,14 @@ def get_course_id_for_enrollment(enterprise_enrollment):
         site=enterprise_enrollment.enterprise_customer_user.enterprise_customer.site
     )
     return course_catalog_client.get_course_id(enterprise_enrollment.course_id)
+
+
+def get_course_run_for_enrollment(enterprise_enrollment):
+    """
+    course_run associated with the enrollment
+    """
+    course_catalog_client = get_course_catalog_api_service_client(
+        site=enterprise_enrollment.enterprise_customer_user.enterprise_customer.site
+    )
+    course_run = course_catalog_client.get_course_run(enterprise_enrollment.course_id)
+    return course_run
--- a/integrated_channels/integrated_channel/exporters/learner_data.py
+++ b/integrated_channels/integrated_channel/exporters/learner_data.py
@@ -8,6 +8,7 @@ enterprise customer.
 """
 
 from logging import getLogger
+
 from opaque_keys import InvalidKeyError
 from slumber.exceptions import HttpNotFoundError
 
@@ -19,12 +20,8 @@ from consent.models import DataSharingConsent
 from enterprise.api_client.lms import CourseApiClient, GradesApiClient
 from enterprise.models import EnterpriseCourseEnrollment
 from integrated_channels.integrated_channel.exporters import Exporter
-from integrated_channels.utils import (
-    generate_formatted_log,
-    is_already_transmitted,
-    parse_datetime_to_epoch_millis,
-)
 from integrated_channels.lms_utils import get_course_certificate
+from integrated_channels.utils import generate_formatted_log, is_already_transmitted, parse_datetime_to_epoch_millis
 
 LOGGER = getLogger(__name__)
 
--- a/integrated_channels/integrated_channel/exporters/learner_data.py
+++ b/integrated_channels/integrated_channel/exporters/learner_data.py
@@ -12,6 +12,7 @@ from logging import getLogger
 from opaque_keys import InvalidKeyError
 from slumber.exceptions import HttpNotFoundError
 
+from django.contrib import auth
 from django.apps import apps
 from django.utils import timezone
 from django.utils.dateparse import parse_datetime
@@ -24,6 +25,7 @@ from integrated_channels.lms_utils import get_course_certificate, get_single_use
 from integrated_channels.utils import generate_formatted_log, is_already_transmitted, parse_datetime_to_epoch_millis
 
 LOGGER = getLogger(__name__)
+User = auth.get_user_model()  # pylint: disable=invalid-name
 
 
 class LearnerExporter(Exporter):
@@ -565,8 +567,8 @@ class LearnerExporter(Exporter):
         """
 
         course_id = enterprise_enrollment.course_id
-        username = enterprise_enrollment.enterprise_customer_user.user.username
-        user_id = enterprise_enrollment.enterprise_customer_user.user_id
+        lms_user_id = enterprise_enrollment.enterprise_customer_user.user_id
+        user = User.objects.get(pk=lms_user_id)
 
         completed_date = None
         grade = self.grade_incomplete
@@ -574,13 +576,13 @@ class LearnerExporter(Exporter):
         percent_grade = None
 
         try:
-            certificate = get_course_certificate(course_id, username)
+            certificate = get_course_certificate(course_id, user)
         except InvalidKeyError:
-            self._log_courseid_not_found(course_id, enterprise_enrollment, user_id)
+            self._log_courseid_not_found(course_id, enterprise_enrollment, lms_user_id)
             return completed_date, grade, is_passing, percent_grade
 
         if not certificate:
-            self._log_cert_not_found(course_id, enterprise_enrollment, user_id)
+            self._log_cert_not_found(course_id, enterprise_enrollment, lms_user_id)
             return completed_date, grade, is_passing, percent_grade
 
         completed_date = certificate.get('created_date')
@@ -688,10 +690,10 @@ class LearnerExporter(Exporter):
         """
 
         course_id = enterprise_enrollment.course_id
-        username = enterprise_enrollment.enterprise_customer_user.user.username
         lms_user_id = enterprise_enrollment.enterprise_customer_user.user_id
+        user = User.objects.get(pk=lms_user_id)
 
-        grades_data = get_single_user_grade(course_id, username)
+        grades_data = get_single_user_grade(course_id, user)
 
         if grades_data is None:
             LOGGER.error('[Integrated Channel] Grades data not found.'
--- a/integrated_channels/integrated_channel/exporters/learner_data.py
+++ b/integrated_channels/integrated_channel/exporters/learner_data.py
@@ -20,7 +20,7 @@ from consent.models import DataSharingConsent
 from enterprise.api_client.lms import CourseApiClient, GradesApiClient
 from enterprise.models import EnterpriseCourseEnrollment
 from integrated_channels.integrated_channel.exporters import Exporter
-from integrated_channels.lms_utils import get_course_certificate
+from integrated_channels.lms_utils import get_course_certificate, get_single_user_grade
 from integrated_channels.utils import generate_formatted_log, is_already_transmitted, parse_datetime_to_epoch_millis
 
 LOGGER = getLogger(__name__)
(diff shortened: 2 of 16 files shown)


alex-sheehan-edx (commented):
A couple comments, I also think there are some tests that need to be fixed, but overall love the cleanup!

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -253,27 +254,12 @@ def export(self, **kwargs):  # pylint: disable=R0915
             channel_name=channel_name,
             enterprise_customer_identifier=self.enterprise_customer.name
         )
-        enrollment_queryset = EnterpriseCourseEnrollment.objects.select_related(
```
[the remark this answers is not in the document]
binodpant: it was not removed, just refactored into a separate method for better readability in this method, the `create_enrollment_queryset` was extracted from the removed blocks of code here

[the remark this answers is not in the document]
binodpant: another related reason was I wanted to simplify what happens in the for loop that follows, so now that loop runs on a pre filtered set, rather than doing filtering within the loop

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_learner_data.py
```diff
@@ -675,7 +673,7 @@ def test_learner_exporter_with_skip_transmitted(
         # Check that LMS enrollment populated as part of model used in audit check:
         expected_result = mock_course_enrollment_class.objects.get.return_value
         self.assertEqual(expected_result, enterprise_course_enrollment.course_enrollment)
-        assert mock_course_enrollment_class.objects.get.call_count == 2
```
binodpant, line 678: I am not 100% sure why this mock call count went from 2 to 1. but I did debug through the code while running the test and verified that the queryset is correct and the rest of the stuff works as expected. Ideas?

binodpant:
@alex-sheehan-edx  @edx-abolger  I am done replacing api calls for two of the fetches in `learner_data.py`, but I think the last one `_collect_assessment_grades_data`  is going to be a fair amount of work due to the amount of logic contained in the endpoint it uses which is here https://github.com/edx/edx-platform/blob/96f0915b0fbde221dadb0b37fba6f186894260b4/lms/djangoapps/grades/rest_api/v1/gradebook_views.py#L511. 


I suggest we do that separately in a new ticket, I have already been working on this ticket a few days, think there’s good improvement here already

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -439,6 +426,60 @@ def export(self, **kwargs):  # pylint: disable=R0915
...
+    ):
+        """
+        Given an enrollment_queryset, returns only enrollments that are not already transmitted
+        """
+        included_enrollments = set()
+        for enterprise_enrollment in enrollment_queryset:
+            if transmission_audit and \
+                    is_already_transmitted(transmission_audit, enterprise_enrollment.id, grade):
+                # We've already sent a completion status for this enrollment
+                generate_formatted_log(
```
binodpant, line 444: for now we can keep this log per item, I want to revisit this as part of the logging cleanup

integrated_channels/catalog_service_utils.py
```diff
@@ -0,0 +1,25 @@
+"""
+A utility collection for calls from integrated_channels to Catalog service
+"""
+from enterprise.api_client.discovery import get_course_catalog_api_service_client
+
+
+def get_course_id_for_enrollment(enterprise_enrollment):
+    """
+    Fetch course_id for a given enterprise enrollment
```
edx-abolger, line 9: For both functions please add a docstring about what happens if one is not found (since this is a remote client): specific exception, returns none?

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -287,21 +273,19 @@ def export(self, **kwargs):  # pylint: disable=R0915
...
-                generate_formatted_log(
-                    'Skipping export of previously sent enterprise enrollment. '
-                    'EnterpriseEnrollment: {enterprise_enrollment_id}'.format(
-                        enterprise_enrollment_id=enterprise_enrollment.id
-                    ),
-                    channel_name=channel_name,
-                    enterprise_customer_identifier=self.enterprise_customer.name
-                )
-                continue
+        if TransmissionAudit and skip_transmitted:
```
edx-abolger, line 276: This refactor seems logical, does it pass any existing tests (are there existing tests for the export() function?)

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -287,21 +273,19 @@ def export(self, **kwargs):  # pylint: disable=R0915
...
+            enrollments_to_transmit = self._filter_out_pre_transmitted_enrollments(
+                enrollment_queryset,
+                channel_name,
+                grade,
+                TransmissionAudit
+            )
+        else:
+            enrollments_to_transmit = enrollment_queryset
```
edx-abolger, line 284: The one thing that I don't like about this refactor is that your new _filter_out_pre_transmitted_enrollments function actually _resolves_ the query set (ie will make a database call) because it iterates over the results and this means that we could get slightly different behaviors at line 286, since if the code path goes through 284 it will have a different query.  That's one reason this was likely all kept together in the first place: your refactor will result in a hidden side effect that might make debugging later harder than needed.  

Basically, can you please change your _filter_out_pre_transmitted_enrollments to accept a _resolved_ queryset, so instead of passing it enrollment_queryset you pass it the result, like 'all_enrollments = enrollment_queryset.list()'  and then pass in 'all_enrollments'. And then you can change line 284 to also use 'all_enrollments' ?

edx-abolger: Further reading in this area: https://docs.djangoproject.com/en/3.1/ref/models/querysets/

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -375,12 +359,15 @@ def export(self, **kwargs):  # pylint: disable=R0915
...
                         course_id=course_id,
-                        user_id=enterprise_enrollment.enterprise_customer_user.user_id,
+                        user_id=enterprise_user_id,
                         enterprise=enterprise_enrollment.enterprise_customer_user.enterprise_customer.slug
                     ),
                     channel_name=channel_name,
                     enterprise_customer_identifier=self.enterprise_customer.name
                 )
+
+            # todo: we need to refactor so single learner is not so mixed in with bulk case
```
edx-abolger, line 369: Please do not leave TODOs in code: file a ticket and leave the ticket here if necessary.

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -439,6 +426,60 @@ def export(self, **kwargs):  # pylint: disable=R0915
...
+            enterprise_customer_user__enterprise_customer=self.enterprise_customer,
+            enterprise_customer_user__active=True,
+        )
+        if learner_to_transmit and course_run_id:
+            enrollment_queryset = enrollment_queryset.filter(
+                course_id=course_run_id,
+                enterprise_customer_user__user_id=learner_to_transmit.id,
+            )
+            generate_formatted_log(
+                'Exporting single learner. Course: {course_run}, User: {user_id}'.format(
```
edx-abolger, line 473: If this is the lms_user_id can you label it as the LMS User Id so that looking at logs later no one has to go to the code to look up what the number means?

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -516,39 +557,64 @@ def _collect_certificate_data(self, enterprise_enrollment):
...
+        grade = self.grade_incomplete
+        is_passing = False
+        percent_grade = None
 
-            # For consistency with _collect_grades_data, we only care about Pass/Fail grades. This could change.
-            is_passing = certificate.get('is_passing')
-            percent_grade = certificate.get('grade')
-            grade = self.grade_passing if is_passing else self.grade_failing
+        try:
+            certificate = get_course_certificate(course_id, username)
```
edx-abolger, line 570: Why do we get course certificate by Username here? We should always be getting these items by lms_user_id.

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -516,39 +557,64 @@ def _collect_certificate_data(self, enterprise_enrollment):
...
+            return completed_date, grade, is_passing, percent_grade
+
+        completed_date = certificate.get('created_date')
+        if completed_date:
+            completed_date = parse_datetime(completed_date)
+        else:
+            completed_date = timezone.now()
 
-        except HttpNotFoundError:
-            LOGGER.error('[Integrated Channel] Certificate data not found.'
```
edx-abolger, line 539: Thank you for removing the PII from this log!

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -516,39 +557,64 @@ def _collect_certificate_data(self, enterprise_enrollment):
...
+        # For consistency with _collect_grades_data, we only care about Pass/Fail grades. This could change.
+        is_passing = certificate.get('is_passing')
+        percent_grade = certificate.get('grade')
+        grade = self.grade_passing if is_passing else self.grade_failing
 
         return completed_date, grade, is_passing, percent_grade
 
+    def _log_cert_not_found(self, course_id, enterprise_enrollment, user_id):
+        """
+        Standardized logging for no certificate found (refactor candidate)
```
edx-abolger, line 594: Again, please file tickets rather than leaving 'todo' like comment in the code.

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -516,39 +557,64 @@ def _collect_certificate_data(self, enterprise_enrollment):
...
-            is_passing = False
-            percent_grade = None
+        # For consistency with _collect_grades_data, we only care about Pass/Fail grades. This could change.
+        is_passing = certificate.get('is_passing')
+        percent_grade = certificate.get('grade')
+        grade = self.grade_passing if is_passing else self.grade_failing
 
         return completed_date, grade, is_passing, percent_grade
 
+    def _log_cert_not_found(self, course_id, enterprise_enrollment, user_id):
```
edx-abolger, line 592: Nit: can we change 'user_id' to 'lms_user_id' everywhere?

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -516,39 +557,64 @@ def _collect_certificate_data(self, enterprise_enrollment):
...
+        """
+        Standardized logging for no certificate found (refactor candidate)
+        """
+        LOGGER.error('[Integrated Channel] Certificate not found for user'
+                     ' Course: {course_id}, EnterpriseEnrollment: {enterprise_enrollment}, '
+                     ' Learner id {user_id}'
```
edx-abolger, line 598: Nit: can we label as Learner LMS User Id in the log so it's easy to tell where to go look this up?

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -608,39 +674,24 @@ def _collect_grades_data(self, enterprise_enrollment, course_details, is_audit_e
...
-                    # This means the user has an enterprise enrollment record but is not enrolled in the course yet
-                    LOGGER.info(
-                        '[Integrated Channel] User is not enrolled in the course.'
-                        ' Course: {course_id}, EnterpriseEnrollment: {enterprise_enrollment},'
-                        ' Username: {username}'.format(
-                            course_id=course_id,
-                            username=username,
-                            enterprise_enrollment=enterprise_enrollment.pk))
-                    return None, None, None, None
+        grades_data = get_single_user_grade(course_id, username)
```
edx-abolger, line 687: Again, can we  use user_id to get this information instead of username?

integrated_channels/lms_utils.py
```diff
@@ -0,0 +1,55 @@
+"""
```
edx-abolger, line 1: This new module needs tests.

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -608,39 +674,24 @@ def _collect_grades_data(self, enterprise_enrollment, course_details, is_audit_e
...
 
+        if grades_data is None:
             LOGGER.error('[Integrated Channel] Grades data not found.'
                          ' Course: {course_id}, EnterpriseEnrollment: {enterprise_enrollment},'
-                         ' Username: {username}'.format(
+                         ' Userid: {user_id}'.format(
```
edx-abolger, line 692: +1 for PII removal, thank you!

edx-abolger (changes requested):
First, great job removing PII from logs!  

Changes requested: 
1.  main issue is lack of small unit tests for new modules.  You should be able to make the tests look similar to existing tests that make sure the right database model value comes back. 
2.  a few log message update requests
3. one small change to a refactor you did to not pass in a queryset as a parameter

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -9,16 +9,18 @@
...
+from opaque_keys import InvalidKeyError
 from slumber.exceptions import HttpNotFoundError
 
 from django.apps import apps
 from django.utils import timezone
 from django.utils.dateparse import parse_datetime
 
 from consent.models import DataSharingConsent
-from enterprise.api_client.lms import CertificatesApiClient, CourseApiClient, GradesApiClient
+from enterprise.api_client.lms import CourseApiClient, GradesApiClient
```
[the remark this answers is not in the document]
binodpant: exactly yes!

integrated_channels/integrated_channel/exporters/learner_data.py
binodpant: yup it passes existing tests (in fact I had to edit existing tests to ensure they passes as they no longer can use the mocks for the api clients

integrated_channels/integrated_channel/exporters/learner_data.py
binodpant: got it had to read it twice but yes I had a lingering thought in this area too, I think it's best to avoid the discrepancy.. will take a look thanks

integrated_channels/integrated_channel/exporters/learner_data.py
binodpant: ah not sure thatt will work given the api call I am using is expecting a username it appears

```
CourseGradeFactory().read(grade_user, course_key=course_key)
```

edx-abolger: That's a "user" as in the object "Auth.user", not a string. I think the user id should be used to get you a user and username shouldn't work anyway, let me look into that for a second.

edx-abolger: Alright so the original call you replaced used the username for who knows what reason. But you don't have to do that, and anyway the grade_user you are looking at there is definitely a user object, not a username string. So you can use django to get the correct user by id and pass it in, and just remove the username usage here completely.

binodpant: got it, now passing the user object instead!

edx-abolger (approved):
Would still like tickets filed for todos added to comments, but changes are approved.

PR #1203, opened 2021-04-01
ENT-4310: Added the ability to link/unlink enterprise customer catalogs with enterprise reporting configuration via its API endpoint.

saleem-latif:
**Jira Ticket:** [ENT-4310](https://openedx.atlassian.net/browse/ENT-4310)

__Description:__
This PR adds the ability to link/unlink enterprise customer catalogs with enterprise reporting configuration via its API endpoint.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -579,7 +579,7 @@ class EnterpriseCustomerReportingConfigurationSerializer(serializers.ModelSerial
             'enterprise_customer', 'enterprise_customer_id', 'active', 'delivery_method', 'email', 'frequency',
             'day_of_month', 'day_of_week', 'hour_of_day', 'include_date', 'encrypted_password', 'sftp_hostname',
             'sftp_port', 'sftp_username', 'encrypted_sftp_password', 'sftp_file_path', 'data_type', 'report_type',
-            'pgp_encryption_key', 'enterprise_customer_catalogs', 'uuid'
+            'pgp_encryption_key', 'enterprise_customer_catalogs', 'uuid', 'enterprise_customer_catalog_uuids',
         )
 
     encrypted_password = serializers.CharField(required=False, allow_blank=False, read_only=False)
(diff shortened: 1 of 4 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -591,11 +591,53 @@ class Meta:
         write_only=True
     )
     enterprise_customer_catalogs = EnterpriseCustomerCatalogSerializer(many=True, read_only=True)
+    enterprise_catalogs = serializers.ListField(write_only=True, child=serializers.UUIDField(), default=[])
```
sameenfatima78, line 594: To avoid confusing it with the already existing variable in the serializer `enterprise_customer_catalogs`, this variable name `enterprise_catalogs` could be changed to something a bit more specific. What exactly does this  return?

enterprise/api/v1/serializers.py
```diff
@@ -606,6 +648,21 @@ def validate(self, data):  # pylint: disable=arguments-differ
...
+        enterprise_customer = data['enterprise_customer']
+        if enterprise_catalogs:
+            catalog_count = enterprise_customer.enterprise_customer_catalogs.filter(
+                uuid__in=enterprise_catalogs
+            ).count()
+
+            if catalog_count != len(enterprise_catalogs):
+                raise serializers.ValidationError({
+                    'enterprise_catalogs': [
+                        'You can only link catalogs belonging to the provided enterprise customer.',
```
sameenfatima78, line 662: nit: "Only those catalogs can be linked that belong to the enterprise customer. "

enterprise/api/v1/serializers.py
```diff
@@ -606,6 +648,21 @@ def validate(self, data):  # pylint: disable=arguments-differ
             if 'email' in data and not bool(data['email']):
                 raise serializers.ValidationError({'email': ['This field is required']})
 
+        # validate that enterprise customer catalogs exist in the system.
+        enterprise_catalogs = data.get('enterprise_catalogs')
+        enterprise_customer = data['enterprise_customer']
```
sameenfatima78, line 653: Will this ever return `None`? Just want to make sure the query below to fetch enterprise catalogs doesn't throw any exception.

sameenfatima78 (approved):
LGTM, just some minor suggestions

enterprise/api/v1/serializers.py
saleem-latif: @sameenfatima78 I have renamed the field name to `enterprise_customer_catalog_uuids` as it will contain UUIDs if of the enterprise customer catalog. Good catch 👍 previous name was very ambiguous.

enterprise/api/v1/serializers.py
saleem-latif: `enterprise_customer` field is required by default (ref: https://github.com/edx/edx-enterprise/blob/master/enterprise/models.py#L2342) so it can not be `None`.

enterprise/api/v1/serializers.py
```diff
@@ -591,11 +591,57 @@ class Meta:
...
+        Returns:
+            (EnterpriseCustomerReportingConfiguration): Instance of the newly created enterprise customer
+                reporting configuration.
+        """
+        ec_catalog_uuids = validated_data.pop('enterprise_customer_catalog_uuids', [])
+        instance = super().create(validated_data)
+
+        # update enterprise customer catalogs on the reporting configuration instance
+        instance.enterprise_customer_catalogs.set(ec_catalog_uuids)
+        instance.save()
```
HammadAhmadWaqas, line 620: No need to call .save() here

enterprise/api/v1/serializers.py
```diff
@@ -591,11 +591,57 @@ class Meta:
...
+            (EnterpriseCustomerReportingConfiguration): Instance of the newly created enterprise customer
+                reporting configuration.
+
+        """
+        ec_catalog_uuids = validated_data.pop('enterprise_customer_catalog_uuids', [])
+        instance = super().update(instance, validated_data)
+
+        # update enterprise customer catalogs on the reporting configuration instance
+        instance.enterprise_customer_catalogs.set(ec_catalog_uuids)
+        instance.save()
```
HammadAhmadWaqas, line 642: No need to call .save() here

HammadAhmadWaqas (approved):
Just small changes, Otherwise good.

PR #1204, opened 2021-04-01
refactor: Refactor part 2 to remove additional remaining api calls to LMS via REST

binodpant:
## Includes

Update:
Going to do a separate PR for this one:
   - [ ] Call lms util instead of REST api to get assessment grades (current call is `grades_api.get_course_assessment_grades`)
 The above needs a edx-platform PR to be merged first

This PR only includes the following:
- [x] Remaining refactors from https://github.com/edx/edx-enterprise/pull/1199
   - [x] Call lms util instead of REST Api for `get_course_details` , this one uses the `CourseApiClient` whose usage can also be removed from this codebase
- [x] Bug fix for bug found in staging testing and test

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/exporters/learner_data.py
+++ b/integrated_channels/integrated_channel/exporters/learner_data.py
@@ -17,6 +17,10 @@ from django.contrib import auth
 from django.utils import timezone
 from django.utils.dateparse import parse_datetime
 
+try:
+    from openedx.core.djangoapps.content.course_overviews.models import CourseOverview
+except ImportError:
+    CourseOverview = None
 from consent.models import DataSharingConsent
 from enterprise.api_client.lms import GradesApiClient
 from enterprise.models import EnterpriseCourseEnrollment
@@ -171,7 +175,14 @@ class LearnerExporter(Exporter):
         if not (TransmissionAudit and already_transmitted) and LearnerExporter.has_data_sharing_consent(
                 enterprise_enrollment):
 
-            course_details = get_course_details(course_run_id)
+            try:
+                course_details = get_course_details(course_run_id)
+            except InvalidKeyError:
+                LearnerExporter._log_courseid_not_found(course_run_id, enterprise_enrollment)
+            except CourseOverview.DoesNotExist:
+                LearnerExporter._log_course_details_not_found(
+                    course_run_id,
+                )
 
             if course_details:
                 assessment_grade_data = self._collect_assessment_grades_data(enterprise_enrollment)
@@ -274,19 +285,23 @@ class LearnerExporter(Exporter):
             course_id = enterprise_enrollment.course_id
 
             # Fetch course details from Courses API
-            # pylint: disable=unsubscriptable-object
-            if course_details:
+            try:
+                course_details = get_course_details(course_id)
                 generate_formatted_log(
                     'Currently exporting for course: {curr_course}, '
-                    'but course details already found: {course_details}'.format(
+                    'Course_details: {course_details} '.format(
                         curr_course=course_id,
                         course_details=course_details
                     ),
                     channel_name=channel_name,
                     enterprise_customer_identifier=self.enterprise_customer.name
                 )
-
-            course_details = get_course_details(course_run_id)
+            except InvalidKeyError:
+                LearnerExporter._log_courseid_not_found(course_id, enterprise_enrollment)
+            except CourseOverview.DoesNotExist:
+                LearnerExporter._log_course_details_not_found(
+                    course_run_id,
+                )
 
             if course_details is None:
                 # Course not found, so we have nothing to report.
@@ -551,11 +566,11 @@ class LearnerExporter(Exporter):
         try:
             certificate = get_course_certificate(course_id, user)
         except InvalidKeyError:
-            self._log_courseid_not_found(course_id, enterprise_enrollment, lms_user_id)
+            LearnerExporter._log_courseid_not_found(course_id, enterprise_enrollment, lms_user_id)
             return completed_date, grade, is_passing, percent_grade
 
         if not certificate:
-            self._log_cert_not_found(course_id, enterprise_enrollment, lms_user_id)
+            LearnerExporter._log_cert_not_found(course_id, enterprise_enrollment, lms_user_id)
             return completed_date, grade, is_passing, percent_grade
 
         completed_date = certificate.get('created_date')
@@ -571,7 +586,8 @@ class LearnerExporter(Exporter):
 
         return completed_date, grade, is_passing, percent_grade
 
-    def _log_cert_not_found(self, course_id, enterprise_enrollment, lms_user_id):
+    @staticmethod
+    def _log_cert_not_found(course_id, enterprise_enrollment, lms_user_id):
         """
         Standardized logging for no certificate found (refactor candidate)
         """
@@ -584,7 +600,8 @@ class LearnerExporter(Exporter):
                          user_id=lms_user_id,
                      ))
 
-    def _log_courseid_not_found(self, course_id, enterprise_enrollment, lms_user_id):
+    @staticmethod
+    def _log_courseid_not_found(course_id, enterprise_enrollment, lms_user_id=None):
         """
         Standardized logging for no certificate found  (refactor candidate)
         """
@@ -596,6 +613,14 @@ class LearnerExporter(Exporter):
                          enterprise_enrollment=enterprise_enrollment,
                          user_id=lms_user_id,
                      ))
+
+    @staticmethod
+    def _log_course_details_not_found(course_id):
+        """
+        course_id valid but not course details are found
+        """
+        LOGGER.error('[Integrated Channel] course details not found for '
+                     'Course: {course_id}'.format(course_id=course_id))
 
     def _collect_assessment_grades_data(self, enterprise_enrollment):
         """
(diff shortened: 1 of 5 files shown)


integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -87,22 +87,6 @@ def grade_audit(self):
         """
         return self.GRADE_AUDIT
 
-    def fetch_course_details(self, course_details, enrollment_course_id):
```
binodpant, line 90: this caching was not being used so simplified it by just calling the new method now

integrated_channels/lms_utils.py
```diff
@@ -51,5 +62,25 @@ def get_single_user_grade(course_id, user):
     if not CourseGradeFactory:
         return None
     course_key = CourseKey.from_string(course_id)
-    course_grade = CourseGradeFactory().read(user.username, course_key=course_key)
+    course_grade = CourseGradeFactory().read(user, course_key=course_key)
     return course_grade
+
+
+def get_course_details(course_id):
```
binodpant, line 69: I need to add a test for this method, will get that in soon

tests/test_integrated_channels/test_lms_utils.py
```diff
@@ -51,6 +52,10 @@ def test_get_single_user_grade_success(self, mock_course_grade_factory):
         mock_course_grade_factory.return_value.read.return_value = expected_grade
         single_user_grade = get_single_user_grade(A_GOOD_COURSE_ID, self.user)
         assert single_user_grade == expected_grade
+        mock_course_grade_factory.return_value.read.assert_called_with(
```
binodpant, line 55: this test covers the bug fix for a bug I found today in prior PR

edx-abolger (commented):
Will wait on additional tests before approving.

integrated_channels/lms_utils.py
```diff
@@ -14,6 +15,16 @@
...
+    CourseOverview = None
+
+
+# Constants
+COURSE_OVERVIEW_NOT_FOUND = 'CourseOverview could not be found for course_key: {course_key}'
+COURSE_KEY_INVALID = 'Invalid course_key: {course_key}'
+
 
 def get_course_certificate(course_id, user):
     """
```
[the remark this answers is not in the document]
binodpant: Ah I double checked, so yes it is used like follows, either it throws that key error or returns an invalid cert

```
try:
            certificate = get_course_certificate(course_id, user)
        except InvalidKeyError:
            self._log_courseid_not_found(course_id, enterprise_enrollment, lms_user_id)
            return completed_date, grade, is_passing, percent_grade

        if not certificate:
            self._log_cert_not_found(course_id, enterprise_enrollment, lms_user_id)
            return completed_date, grade, is_passing, percent_grade
```


the impl in edx-platform of `get_certificate_for_user` looks like this so yeah looks like we are covered? 

```
try:
        cert = GeneratedCertificate.eligible_certificates.get(
            user__username=username,
            course_id=course_key
        )
    except GeneratedCertificate.DoesNotExist:
        return None

    if format_results:
        return format_certificate_for_user(username, cert)
    else:
        return cert
```

integrated_channels/lms_utils.py
```diff
@@ -51,5 +62,25 @@ def get_single_user_grade(course_id, user):
     if not CourseGradeFactory:
```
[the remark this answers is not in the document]
binodpant: did you mean line 38 with this? 

```
user_cert = get_certificate_for_user(username=user.username, course_key=course_key)
```

checked that method again in edx-platform and indeed it expects username instead of user (odd)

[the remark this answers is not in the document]
binodpant:
```
def get_certificate_for_user(username, course_key, format_results=True):
    """
    Retrieve certificate information for a particular user for a specific course.

    Arguments:
        username (unicode): The identifier of the user.
        course_key (CourseKey): A Course Key.
```

integrated_channels/lms_utils.py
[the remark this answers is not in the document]
edx-abolger: I meant the docstring says one thing and the code does something else. Also the code behaves differently under 2 different error conditions, which means using this function will require the caller to handle None and also handle an exception. Can you make the function more consistent?

integrated_channels/lms_utils.py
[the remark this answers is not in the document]
edx-abolger: Ew. not great but obviously outside the scope of this ticket and our team. Will close comment.

integrated_channels/lms_utils.py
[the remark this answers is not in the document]
binodpant: yeah that case is interesting since it's due an importError. Perhaps it's best indicated if I actually threw an ImportError

```
if not get_certificate_for_user:
        raise ImportError("Could not load get_certificate_for_user function")
```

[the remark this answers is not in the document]
binodpant: Ah I see we have this pattern in other plcaes 

```
if get_certificate_for_user is None:
            raise NotConnectedToOpenEdX(
                _('To use this EnterpriseCourseEnrollmentSerializer, this package must be '
                  'installed in an Open edX environment.')
            )
```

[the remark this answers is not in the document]
binodpant: yeah so going with this pattern here too basically the user of this api only has to handle InvalidKeyError now, and updated docstring

binodpant:
This PR (the grades api client part) is blocked until I release edx-platform to offer a python api to fetch single learner grade

integrated_channels/lms_utils.py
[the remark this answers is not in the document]
edx-abolger: Thank you. This is much more consistent as Python has nice syntax for things like this: 
```

try:
    # do something that may fail
except IDontLikeYouException:
    # say please
except YouAreTooShortException:
    # stand on a ladder
```

integrated_channels/lms_utils.py
edx-abolger: Looks like you have added one, resolving this comment.

integrated_channels/lms_utils.py
```diff
@@ -42,14 +56,44 @@ def get_course_certificate(course_id, user):
...
+        )
     course_key = CourseKey.from_string(course_id)
-    course_grade = CourseGradeFactory().read(user.username, course_key=course_key)
+    course_grade = CourseGradeFactory().read(user, course_key=course_key)
     return course_grade
+
+
+def get_course_details(course_id):
+    """
+    Returns:
```
edx-abolger, line 77: Again, this is not a clean interface. You just moved the problems: you are mixing "Return an error code OR throw an exception, the same 20 line function."  That is 2 different problems the caller now has to handle. Either ALWAYS return none+error code or ALWAYS let the exceptions pass through.

edx-abolger: To explain: the problem with this is that now the calling code will have to write _extra_ code and do _more_ weird error handling, not to mention the code author will have to go look inside this function to figure out how to write the error handler for it.

binodpant: Ah I see what you are saying, so the current usage looks like this with a mixed bag of None vs exceptions

```
course_details = get_course_details(course_run_id)
if course_details is None:
    # Course not found, so we have nothing to report.
```

Instead we can go:

```
try:
    course_details = get_course_details(course_run_id)
except InvalidKeyError:                             # Course not found, log error
except CourseOverview.DoesNotExist:   # Course key is invalid, log error
```

This will also make this function consistent with the other ones in the pr

binodpant: note, the caller has to still handle these two exceptions, since we call this function in a loop of enrollments, we want to log and continue the loop in these two cases. The NotConnectedException is different, it should be treated as runtime exception since something is really bad, no point in continuing. But rather than swallow these exceptions to is better regardless, to propagate them from the get_course_details function.

binodpant:
also I just discovered a mistake I had made. the methods need to process course_key not course_id. The names are confusing me, as used in the code. The enterprise_enrollment.course_id are actually course_keys !

binodpant:
and the `course_id` expected by the lms apis are actually instances of CourseKey objects

```
course_id = CourseKey.from_string(course_key)
```

integrated_channels/lms_utils.py
```diff
@@ -67,31 +67,29 @@ def get_single_user_grade(course_id, user):
             'To use this function, this package must be '
             'installed in an Open edX environment.'
         )
-    course_key = CourseKey.from_string(course_id)
-    course_grade = CourseGradeFactory().read(user, course_key=course_key)
+    course_id = CourseKey.from_string(course_key)
+    course_grade = CourseGradeFactory().read(user, course_key=course_id)
```
binodpant, line 71: note: every thing that says `course_id` in the grading apis refers to instances of CourseKey which have to be obtained via CourseKey.from_string() from the course_key strings provided from learner_data.py (those enterprise_enrollment.course_id are actually course key strings)

binodpant:
updated with a test fix, and an actual bug I found when running the task since I had misunderstood the `course_id` params that go into the LMs apis

added a proposal to our proposal listing: we don't have integration level tests right now, I think we will want at least one for each type of transmit task for better lockdown of bugs

edx-abolger (approved):

integrated_channels/lms_utils.py
binodpant: @edx-abolger sorry for the multiple rounds here, but this new pattern look better and more consistent hopefully

alex-sheehan-edx (approved):
G2G 👍

PR #1209, opened 2021-04-07
feat: Add UUID field to EnterpriseCatalogQuery model

taliaedX:
**Description**
Adds UUID field for catalog queries and generates them for all existing queries. There are only 137 in stage and 8 in prod, so there is no performance concern with running this generation function.
Why are we generating existing ones here and not copying from catalog? Too much effort. The ones in catalog aren't being used currently either so we'll replace what catalog has as LMS should be the source of truth.

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -1864,6 +1864,11 @@ class EnterpriseCatalogQuery(TimeStampedModel):
         ),
         validators=[validate_content_filter_fields]
     )
+    uuid = models.UUIDField(
+        unique=True,
+        default=uuid4,
+        editable=False,
```
iloveagent57, line 1870: Consider making `editable=True` and having it be a read-only field in the admin site, so that it's easy to find a UUID for pasting into config files without having to go to the read-replica.

iloveagent57 (commented):
Looks good, just one suggestion.

iloveagent57:
Oh, also: version and changelog entry if you please.

enterprise/models.py
taliaedX: ooo fair point.

alex-sheehan-edx (approved):
LGTM 👍

iloveagent57 (approved):

PR #1210, opened 2021-04-08
fix: :bug: use CourseOverview model correctly ENT-4222

binodpant:
Return value of new get_course_details is no longer a dict but should model a django model
Return value of get_single_learner_grade also is no longer a dict but a `CourseGrade` object hence make those fixes of tests and code

This PR includes:

- [x] fix for the bug mentioned 
- [x] tests updated to reflect mock as close to the (unavailable) edx-platform courseOverview model as possible
- [x] refactor of learner_data such that all filtering of enrollmentss happens befor ethe main loop begings which simplifies code a bit
- [x] no longer check get_course_detail in the single learner assessment data portion

### Tested by

running transmit_learner_data with one enrollment (it did not transmit but I ran througn entire export() workflow) for a course with pacing='self'

```
./manage.py lms transmit_learner_data --api_user edx --channel CANVAS
```

```
2021-04-08 01:50:14,342 INFO 86209 [integrated_channels.utils] [user None] [ip None] utils.py:252 - [Integrated Channel] [ENT CUSTOMER: edx] [CHANNEL: CANVAS]: Batch processing learners for integrated channel. Configuration: <CanvasEnterpriseCustomerConfiguration for Enterprise Test Enterprise>
2021-04-08 01:50:14,342 INFO 86209 [integrated_channels.utils] [user None] [ip None] utils.py:252 - [Integrated Channel] [ENT CUSTOMER: Test Enterprise] [CHANNEL: canvas]: Starting Export. CompletedDate: None, Course: None, Grade: None, IsPassing: False, User: None
2021-04-08 01:50:15,242 INFO 86209 [integrated_channels.utils] [user None] [ip None] utils.py:252 - [Integrated Channel] [ENT CUSTOMER: Test Enterprise] [CHANNEL: canvas]: Beginning export of enrollments: [1].
2021-04-08 01:50:15,252 ERROR 86209 [enterprise.models] [user None] [ip None] models.py:1689 - <EnterpriseCourseEnrollment for user test-enterprise_learner_2 in course with ID course-v1:edX+E2E-101+course> does not have a matching student.CourseEnrollment
2021-04-08 01:50:15,267 INFO 86209 [integrated_channels.utils] [user None] [ip None] utils.py:252 - [Integrated Channel] [ENT CUSTOMER: Test Enterprise] [CHANNEL: canvas]: Currently exporting for course: course-v1:edX+E2E-101+course, Course_details: course-v1:edX+E2E-101+course
2021-04-08 01:50:15,288 ERROR 86209 [integrated_channels.integrated_channel.exporters.learner_data] [user None] [ip None] learner_data.py:629 - [Integrated Channel] Certificate not found for user Course: course-v1:edX+E2E-101+course, EnterpriseEnrollment: <EnterpriseCourseEnrollment for user test-enterprise_learner_2 in course with ID course-v1:edX+E2E-101+course>,  Learner LMS User Id: 22
2021-04-08 01:50:15,289 INFO 86209 [integrated_channels.utils] [user None] [ip None] utils.py:252 - [Integrated Channel] [ENT CUSTOMER: Test Enterprise] [CHANNEL: canvas]: Received data from certificate api. CompletedDate: None, Course Id: course-v1:edX+E2E-101+course, Enterprise: test-enterprise,  Grade: In Progress, IsPassing: False, Learner LMS User Id: 22
2021-04-08 01:50:16,427 INFO 86209 [integrated_channels.integrated_channel.transmitters.learner_data] [user None] [ip None] learner_data.py:161 - Skipping in-progress enterprise enrollment 1
2021-04-08 01:50:16,430 INFO 86209 [integrated_channels.utils] [user None] [ip None] utils.py:252 - [Integrated Channel] [ENT CUSTOMER: Test Enterprise] [CHANNEL: canvas]: Finished exporting enrollments. Skipped enrollments: [].
2021-04-08 01:50:16,435 INFO 86209 [integrated_channels.utils] [user None] [ip None] utils.py:252 - [Integrated Channel] [ENT CUSTOMER: edx] [CHANNEL: CANVAS]: Batch learner data transmission task finished. Configuration: <CanvasEnterpriseCustomerConfiguration for Enterprise Test Enterprise>, Duration: 2.1168627738952637
2021-04-08 01:50:16,437 INFO 86209 [celery.app.trace] [user None] [ip None] trace.py:125 - Task integrated_channels.integrated_channel.tasks.transmit_learner_data[d3230306-34fb-4f5d-ab6b-af65a31bbe54] succeeded in 2.120410199975595s: None
```

None

**Merge checklist:**

**Post merge:**


binodpant:
forgot about those other tests, will be back with a fix up for those

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -175,6 +173,7 @@ def single_assessment_level_export(self, **kwargs):
         if not (TransmissionAudit and already_transmitted) and LearnerExporter.has_data_sharing_consent(
                 enterprise_enrollment):
 
+            course_details = None
```
taliaedX, line 176: I think the loss of caching might affect something, but Alex will have more insight into that bit. 
For this particular method though, I have to ask: why are we even retrieving course_details? We seem to be using it for an if condition, but I feel like there must be a way to do that without another API call as we aren't actually using any of the contents in this method and we're not setting it into the class variable so it can be reused without another API call now, so I'm just unclear why the heavy lift of another API call here.

I also know this PR doesn't address that, so it's something maybe Alex can explain or we can address at another time.

taliaedX (commented):
Your log printout for testing also has errors in it so I want to double-check what we're expecting to fix. I would love a log or just inline code message in the PR description that shows what the actual error was (I assume it's something like `AttributeError: 'tuple' object has no attribute 'get'`, but just want to verify).

Other than that it looks a-okay to me, will probably defer to Alex for approval though.

integrated_channels/integrated_channel/exporters/learner_data.py
binodpant: good point!

binodpant: yeah I too not clear now that you bring it up, why we fetch it here, thoughts @alex-sheehan-edx ?

binodpant:
> Your log printout for testing also has errors in it so I want to double-check what we're expecting to fix. I would love a log or just inline code message in the PR description that shows what the actual error was (I assume it's something like `AttributeError: 'tuple' object has no attribute 'get'`, but just want to verify).
> 
> Other than that it looks a-okay to me, will probably defer to Alex for approval though.

yeah I think it's how I created the enrollments maybe? I might have had trouble enrolling via learner portal/admin portal so I must have done sth hacky hence those errors.

cert error is interesting, not sure if I know how to setup locally to test the certificate based path but intend to do so in staging for sure, I think my course is in verified track locally but cert is not issued... or something like that

integrated_channels/integrated_channel/exporters/learner_data.py
binodpant: I see it's used as a check right after, that is the only use of it in this function. also get_course_Details is no longer a REST api call so that should help us avoid perf issues

binodpant: also course_details would be per course_run_id, so if we cached, we would have to do per course id or course_run_id anyway

alex-sheehan-edx: In terms of caching, this is a single enrollment transmission, so I don't think it would ever actually have an effect (?). Meaning that we would cache bulk transmissions by retaining the fetched course details to be reused if applicable, but since we're only transmitting for one learner in one course, I don't think it would have an effect.

(I also took a look at the commit when I had originally added this method and it included this:
https://github.com/edx/edx-enterprise/commit/ca1750a93b187f3a9e998965269fbcfb4212f77b#diff-f488bfc03a35af19cc710b17a06ab3014e240abef4ad52114822c8b0e081b5caR177-R181 )
 
Having said that... As to why we're even fetching the course details here... I'm really not entirely sure 🙃. The assessment level exports (bulk and single) were largely based on the (not so perfect) completion data export method, which uses the course details to:
- confirm that the course exists before attempting to export
- uses the `pacing` attribute of `course_details` to determine whether it should pull certificate data or grades data
- uses the `end` attribute in retrieving grades data (for self paced courses)
- some light logging, giving details as to the specific transmission happening

This `single_assessment_level_export` does not need:
- to differentiate between pacings, we simple grab the assessment grade data 
- to use `end` as these transmissions happen throughout the learners progress through the course
- any kind of caching (that I can tell?) because we are only transmitting data related to a single course (we will only have one course_details throughout the entire transmission)

So the only thing, as you all have rightly pointed out, that it's using the course details for is to verify that the course exists. However having said that, if we take a look at `bulk_assessment_level_export` above this method, we are accomplishing essentially the same goal but for many learners and we never fetch course details a single time. 

So TL:DR I think fetching the course details was a consequence of building off of existing, not perfect architecture and can probably be removed entirely from this method, if we're ok with not validating that the course exists.

taliaedX: Is there a scenario where we'd get data to export for a course that does not in fact exist?
I would imagine we'd not get to this point if we just didn't have a course? Or we could also just verify the course details where we actually use them.

alex-sheehan-edx: We start the whole export when a user has gone onto our site and physically completed the course, so I really don't see how that could ever happen without the course existing.

alex-sheehan-edx: In fact, we use the `course run id` provided by platform's signal to get those course details so if we're worried about not having a valid course, I would ask how the learner was able to complete a subsection of the course linked to the course run id provided by the signal (which directly pulls it from the course)

binodpant: I agree there too, why fetch again if we are here as a result of a course's completion.. right?

binodpant: going to bite the bullet and try removing it ...

binodpant: but this also. makes me ask why this?

```
       course_run_id = kwargs.get('course_run_id', None)
```

seems like course_run_id should then be a required arg , do we want to at least check course_run_id is not None here before proceeding?

binodpant: or we can just let the enrollment_queryset be empty if it's None, don't know if that is going to be the case

binodpant: I would say let's just not get course_details for now and see...

binodpant: okie check out the update!!! I still have not been able to test this locally (damn publisher/ecom etc!) so trying to run that task directly to see what happens ... with manually fed data

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_learner_data.py
```diff
@@ -555,21 +554,19 @@ def get_course_grade(course_id, username):
             assert report1.grade == LearnerExporter.GRADE_INCOMPLETE
 
         assert learner_data[2].course_id == self.course_key
-        assert learner_data[3].course_id == self.course_id
+        assert learner_data[3].course_id == course_id2
```
binodpant, line 557: these changes are because I am now using set() in two places and the ordering of enrollments has apparently changed.

binodpant:
ok I ananged to test the single assessment task by doing this


```
transmit_single_subsection_learner_data('enterprise_openedx_operator','course-v1:edX+E2E-101+course',1,0.8)
```


I was doing one enrollment, and passed random values

```
2021-04-08 20:53:56,917 INFO 89550 [integrated_channels.utils] [user None] [ip None] utils.py:252 - [Integrated Channel] [ENT CUSTOMER: None] [CHANNEL: None]: Beginning single exportation of learner data for enrollment: 3
2021-04-08 20:53:57,633 INFO 89550 [integrated_channels.integrated_channel.tasks] [user None] [ip None] tasks.py:189 - [Integrated Channel] Single learner data transmission task finished.Customer: 4c44681c-06db-4e4f-b4bf-418be997d92f, Course: course-v1:edX+E2E-101+course, Duration: 25.33825945854187, Username: enterprise_openedx_operator
```

It went through without a failure upto data sharing consent check but trying to go past that I started to getthese kinds of failures

```
 File "/edx/app/edxapp/edx-platform/openedx/core/djangoapps/oauth_dispatch/jwt.py", line 120, in _create_jwt
    'preferred_username': user.username,
AttributeError: 'NoneType' object has no attribute 'username'
```

So I think maybe this is best tested from staging, not worth this setup right now I think

integrated_channels/integrated_channel/exporters/learner_data.py
alex-sheehan-edx: `course_run_id = kwargs.get('course_run_id', None)` is another relic of the combined single and bulk completion exports. Most likely we can just change it to a required param.

binodpant:
alright reviewers! Thanks @alex-sheehan-edx for helping testing this on his devstack.

We created a course, we finished a subsection, and ensured the signal hit the single subsectoin code path
We also (inadvertently) tested the main export() method by calling the transmit_learner_data() workflow which generally runs during a full course completion

when we completed a subsection, we got some output like this :

```
edx.devstack.lms                  | 2021-04-09 15:46:41,149 INFO 2814 [integrated_channels.integrated_channel.transmitters.learner_data] [user 35] [ip 172.18.0.1] learner_data.py:57 - edX+816
edx.devstack.lms                  | 2021-04-09 15:46:41,150 INFO 2814 [integrated_channels.integrated_channel.transmitters.learner_data] [user 35] [ip 172.18.0.1] learner_data.py:58 - block-v1:edX+816+3T2020+type@sequential+block@05584d82f5354b938bede0edaaef24b1
```

which verified for us that the transmission audit is correct after sending data over to blackboard (which was the test LMS here)

We found and fixed two bugs after that which will you see a fix for in the latest commits

Now I feel much better about this! But we really do need to get local testing setup fixed up so any one can run these steps

more discussion on that later

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_learner_data.py
alex-sheehan-edx: it's weird to me that we'd be hard coding the ordering the learner data here but  ¯\_(ツ)_/¯

alex-sheehan-edx (approved):
with the small bugs we found in manual testing fixed this LGTM 👍

binodpant:
also checked that the error reports from ENT-4452 are also related to this bug and should be fixed by the PR

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_learner_data.py
binodpant: yeah I did not like the way this test was to begin with.. I think we need to find a better way than to rely on order here.. unless we determine order is important for some reason, I can't see why

PR #1214, opened 2021-04-13
fix: ENT-4383: proxy login view now correctly passes along tpa_hint param to /login

binodpant:
Passes tpa_hint param to the next url in the redirect sequence. Before the fix, it was being urlencoded as part of the 'next' param and therefore not picked up by the /login endpoint. This change retains the prior behavior but also adds `tpa_hint` as a url param to the redirect_uri which needs it

All in all, this really just amounts to adding the param `&tpa_hint=saml-default-saml-config-bpant` to the redirect_uri as it stands today without changing the value of the `next` parameter passed to it

JIRA: https://openedx.atlassian.net/browse/ENT-4383

###  Testing notes 

Locally, verified that, on visiting learner portal: http://localhost:8734/saml-customer 

* before fix: I ended up at a url like this:  `http://localhost:18000/login?next=http%3A%2F%2Flocalhost%3A8734%2Fsaml-customer%3Ftpa_hint%3Dsaml-default-saml-config-bpant`
* after fix: I now get redirected to: `/login?next=http%3A%2F%2Flocalhost%3A8734%2Fsaml-customer%3Ftpa_hint%3Dsaml-default-saml-config-bpant&tpa_hint=saml-default-saml-config-bpant`

**Merge checklist:**
- [] Translations updated (see docs/internationalization.rst but also this isn't blocking for merge atm)

**Post merge:**

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -1067,15 +1067,14 @@ class EnterpriseProxyLoginView(View):
             # Add the tpa_hint to the redirect's 'next' query parameter
             # Redirect will be to the Enterprise Customer's TPA provider
             query_dict['tpa_hint'] = tpa_hint
-            new_query = urlencode(query_dict, doseq=True)
         else:
             # If there's no linked IDP or multiple IDPs are linked and no tpa_hint provided in query_param
             # Add Enterprise Customer UUID and proxy_login to the redirect's query parameters
             # Redirect will be to the edX Logistration with Enterprise Proxy Login sidebar
             query_dict['enterprise_customer'] = [str(enterprise_customer.uuid)]
             query_dict['proxy_login'] = [True]
-            new_query = urlencode(query_dict, doseq=True)
 
+        new_query = urlencode(query_dict, doseq=True)
         new_redirect_to = urlunsplit((scheme, netloc, path, new_query, fragment))
         return redirect(new_redirect_to)
 
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -75,6 +75,7 @@ from enterprise.utils import (
     is_course_run_enrollable,
     track_enrollment,
     ungettext_min_max,
+    update_query_parameters,
 )
 from integrated_channels.cornerstone.utils import create_cornerstone_learner_data
 
@@ -1067,6 +1068,12 @@ class EnterpriseProxyLoginView(View):
             # Add the tpa_hint to the redirect's 'next' query parameter
             # Redirect will be to the Enterprise Customer's TPA provider
             query_dict['tpa_hint'] = tpa_hint
+
+            # also add tpa_hint to the next url being sent
+            tpa_next_param = {
+                'tpa_hint': tpa_hint,
+            }
+            query_dict['next'] = update_query_parameters(str(query_dict['next']), tpa_next_param)
         else:
             # If there's no linked IDP or multiple IDPs are linked and no tpa_hint provided in query_param
             # Add Enterprise Customer UUID and proxy_login to the redirect's query parameters
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -75,7 +75,6 @@ from enterprise.utils import (
     is_course_run_enrollable,
     track_enrollment,
     ungettext_min_max,
-    update_query_parameters,
 )
 from integrated_channels.cornerstone.utils import create_cornerstone_learner_data
 
@@ -1067,18 +1066,16 @@ class EnterpriseProxyLoginView(View):
         if tpa_hint:
             # Add the tpa_hint to the redirect's 'next' query parameter
             # Redirect will be to the Enterprise Customer's TPA provider
-            tpa_next_param = {
-                'tpa_hint': tpa_hint,
-            }
-            query_dict['next'] = update_query_parameters(str(query_dict['next']), tpa_next_param)
+            query_dict['tpa_hint'] = tpa_hint
+            new_query = urlencode(query_dict, doseq=True)
         else:
             # If there's no linked IDP or multiple IDPs are linked and no tpa_hint provided in query_param
             # Add Enterprise Customer UUID and proxy_login to the redirect's query parameters
             # Redirect will be to the edX Logistration with Enterprise Proxy Login sidebar
             query_dict['enterprise_customer'] = [str(enterprise_customer.uuid)]
             query_dict['proxy_login'] = [True]
+            new_query = urlencode(query_dict, doseq=True)
 
-        new_query = urlencode(query_dict, doseq=True)
         new_redirect_to = urlunsplit((scheme, netloc, path, new_query, fragment))
         return redirect(new_redirect_to)
 
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -1065,11 +1065,11 @@ class EnterpriseProxyLoginView(View):
         tpa_hint = enterprise_customer.get_tpa_hint(tpa_hint_param)
 
         if tpa_hint:
-            # Add the tpa_hint to the redirect's 'next' query parameter
+            # Add the tpa_hint to the redirect's query parameters listing
             # Redirect will be to the Enterprise Customer's TPA provider
             query_dict['tpa_hint'] = tpa_hint
 
-            # also add tpa_hint to the next url being sent
+            # Also add tpa_hint to the next url being sent
             tpa_next_param = {
                 'tpa_hint': tpa_hint,
             }
--- a/tests/test_enterprise/views/test_enterprise_proxy_login_view.py
+++ b/tests/test_enterprise/views/test_enterprise_proxy_login_view.py
@@ -82,24 +82,33 @@ class TestEnterpriseProxyLoginView(TestCase):
         expected_url = LMS_LOGIN_URL + '?' + query_params.urlencode()
         self.assertRedirects(response, expected_url, fetch_redirect_response=False)
 
-    @ddt.data(
-        {'use_next': True},  # 'next' query param is passed in URL
-        {'use_next': False},  # 'next' query param not passed, default to LP URL with slug
-    )
-    @ddt.unpack
-    def test_tpa_redirect(self, use_next):
+    def test_tpa_redirect_no_next_param(self):
         """
-        Verify the view adds the next param and the tpa_hint to the redirect if the enterprise has an identity provider.
+        When 'next' url param is absent:
+          The proxy login redirects to default learner portal url,
+          url should contain: ?next=<learner portal url>&<tpa_hint=customer idp>
         """
-        url = self._get_url_with_params(use_next=use_next)
+        url = self._get_url_with_params(use_next=False)
         response = self.client.get(url)
         query_params = QueryDict(mutable=True)
-        if use_next:
-            next_url = self.next_url
-        else:
-            learner_portal_url = settings.ENTERPRISE_LEARNER_PORTAL_BASE_URL
-            next_url = learner_portal_url + '/' + self.enterprise_customer.slug
-        query_params['next'] = f'{next_url}?tpa_hint={self.enterprise_customer.identity_provider}'
+        learner_portal_url = settings.ENTERPRISE_LEARNER_PORTAL_BASE_URL
+        next_url = learner_portal_url + '/' + self.enterprise_customer.slug
+        query_params['next'] = next_url
+        query_params['tpa_hint'] = self.enterprise_customer.identity_provider
+        expected_url = LMS_LOGIN_URL + '?' + query_params.urlencode()
+        self.assertRedirects(response, expected_url, fetch_redirect_response=False)
+
+    def test_tpa_redirect_with_next_param(self):
+        """
+        When 'next' url param is present:
+          The proxy login redirects to the provided next url,
+          url should contain: ?next=<provided next param>&<tpa_hint=customer idp>
+        """
+        url = self._get_url_with_params(use_next=True, next_override=self.next_url)
+        response = self.client.get(url)
+        query_params = QueryDict(mutable=True)
+        query_params['next'] = self.next_url
+        query_params['tpa_hint'] = self.enterprise_customer.identity_provider
         expected_url = LMS_LOGIN_URL + '?' + query_params.urlencode()
         self.assertRedirects(response, expected_url, fetch_redirect_response=False)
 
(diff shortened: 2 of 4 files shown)


tests/test_enterprise/views/test_enterprise_proxy_login_view.py
```diff
@@ -82,24 +82,33 @@ def test_redirect_no_tpa(self, use_next):
...
-            next_url = learner_portal_url + '/' + self.enterprise_customer.slug
-        query_params['next'] = f'{next_url}?tpa_hint={self.enterprise_customer.identity_provider}'
+        learner_portal_url = settings.ENTERPRISE_LEARNER_PORTAL_BASE_URL
+        next_url = learner_portal_url + '/' + self.enterprise_customer.slug
+        query_params['next'] = next_url
+        query_params['tpa_hint'] = self.enterprise_customer.identity_provider
+        expected_url = LMS_LOGIN_URL + '?' + query_params.urlencode()
+        self.assertRedirects(response, expected_url, fetch_redirect_response=False)
+
+    def test_tpa_redirect_with_next_param(self):
```
binodpant, line 101: I felt like having conditionals in the test for logic branching is not a good idea, hence I split these two cases out

enterprise/views.py
```diff
@@ -1067,18 +1066,16 @@ def get(self, request):
         if tpa_hint:
             # Add the tpa_hint to the redirect's 'next' query parameter
             # Redirect will be to the Enterprise Customer's TPA provider
-            tpa_next_param = {
-                'tpa_hint': tpa_hint,
-            }
-            query_dict['next'] = update_query_parameters(str(query_dict['next']), tpa_next_param)
+            query_dict['tpa_hint'] = tpa_hint
```
edx-abolger, line 1069: This looks like you are removing a level of indirection that was in use before. Is it possible to do a quick git blame and figure out why it was done this way initially? Maybe we need the value both encoded for the next URL level AND at this current level? (Unless the original coder was just mistaken on where the value went)

tests/test_enterprise/views/test_enterprise_proxy_login_view.py
```diff
@@ -133,13 +142,17 @@ def test_tpa_redirects_using_tpa_hint_param(
...
-        url = self._get_url_with_params(enterprise_slug_override=enterprise_customer.slug, tpa_hint=tpa_hint_param)
+        url = self._get_url_with_params(
+            enterprise_slug_override=enterprise_customer.slug,
+            tpa_hint=tpa_hint_param,
+        )
         response = self.client.get(url)
         query_params = QueryDict(mutable=True)
         next_url = self.next_url
 
         if redirected_tpa_hint:
```
edx-abolger, line 153: Again, see my previous comment above- it's not clear to me if we need to have tpa_hint at 2 possible levels of redirect (where before we only had 1) or if your replacement for a single level of redirect is fixing something that was built incorrectly the first time. I'd just like a double check?

enterprise/views.py
binodpant: this is a good question. I now wonder if the tpa_hint is _also_ needed in the 'next' url we are forwarding to... will check

binodpant: since @chavesj made a prior change, and looks like this will undo that, I am going to pull him and consult before making this change

binodpant: esp since this PR had a fix in which required the tpa_hint in the `next` url it looks like https://github.com/edx/edx-enterprise/pull/961/files

enterprise/views.py
```diff
@@ -1067,18 +1066,16 @@ def get(self, request):
...
+            new_query = urlencode(query_dict, doseq=True)
         else:
             # If there's no linked IDP or multiple IDPs are linked and no tpa_hint provided in query_param
             # Add Enterprise Customer UUID and proxy_login to the redirect's query parameters
             # Redirect will be to the edX Logistration with Enterprise Proxy Login sidebar
             query_dict['enterprise_customer'] = [str(enterprise_customer.uuid)]
             query_dict['proxy_login'] = [True]
+            new_query = urlencode(query_dict, doseq=True)
```
taliaedX, line 1077: what is the point of moving and duplicating this code? I may be completely missing something, but neither part of the if/else returns something and it's not like you're defining a different "query_dict" variable.

binodpant: good catch it's a holdover from a prior round I forgot to address

enterprise/views.py
chavesj: After taking a look at [this commit](https://github.com/edx/edx-enterprise/commit/2f34adac7b2ad87b9e3a2d749be05e2aadce5270) it seems like the `tpa_hint` param is needed as part of the `next` query param. Based on the commit message it seems needed for the correct welcome template to be rendered after authenticating with TPA.

taliaedX (commented):
I'm in the same boat as Adrienne. It feels like the solution would have to do with modifying update_query_parameters or changing the urlencoding of the query params prior to sending to fix the urlencoding on the tpa_hint (which was my understanding at least of the reported issue), but instead the solution presented here is just dropping particular redirect functionality altogether, which seems like probably not the best path.

tests/test_enterprise/views/test_enterprise_proxy_login_view.py
chavesj: The conditional in conjunction with DDT was to avoid having this duplicate logic in tests. I don't agree with this change but this also isn't a hill I'm willing to die on so it's up to you

enterprise/views.py
binodpant: Thanks @chavesj going to add that back as well so I don't break that use case. To be clear, I will also add it to the `next` url, but also maintain it when redirecting to the /login url as I am doing here.. .we on the same page?

binodpant: so plan is to add it to next as a urlencoded value like before, but also have tpa_hint be passed to the redirect_uri...

binodpant: so yeah it will look like this now 

```
http://lms.example.com/login?next=http%3A%2F%2Flocalhost%3A18000%3Ftpa_hint%3Didp-1&tpa_hint=idp-1
```

binodpant: please check latest commit

tests/test_enterprise/views/test_enterprise_proxy_login_view.py
binodpant: I hear you on the duplication. It is a fine balance sometimes between choosing duplication vs choosing conditional logic - esp in tests. Typically I choose to avoid the latter and also we can address duplication such as this (if more than two tests for example), by util methods and such... I decided not to do that here yet but yeah if we do more of these, worth addressing the dup too. Also, sorry just me, I just have not yet gained a liking for ddt (parameterized testing in general) unless a large no. of data driven scenarios are being handled.

binodpant:
note: also updated the url I was taken to, afte r the latest change,s in the testing notes

/login?next=http%3A%2F%2Flocalhost%3A8734%2Fsaml-customer%3Ftpa_hint%3Dsaml-default-saml-config-bpant&tpa_hint=saml-default-saml-config-bpant

binodpant:
and my local testing setup gave this login session log in case reviewers spot anything odd about it: It looks okay to me now and the login worked via saml

but if there are other cases I should test I can do that (I am going to do that in staging as well)


```
edx.devstack.lms                  | [14/Apr/2021 16:32:33] "GET /login?next=http%3A%2F%2Flocalhost%3A8734%2Fsaml-customer%3Ftpa_hint%3Dsaml-default-saml-config-bpant&tpa_hint=saml-default-saml-config-bpant HTTP/1.1" 302 0
edx.devstack.lms                  | 2021-04-14 16:32:33,472 INFO 3006 [tracking] [user None] [ip 172.20.0.1] logger.py:41 - {"name": "/auth/login/tpa-saml/", "context": {"user_id": null, "path": "/auth/login/tpa-saml/", "course_id": "", "org_id": ""}, "username": "", "session": "3d78d1a9d9ddd44ac8aa8451630031f7", "ip": "172.20.0.1", "agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.183 Safari/537.36", "host": "localhost:18000", "referer": "http://localhost:8734/", "accept_language": "en-US,en;q=0.9", "event": "{\"GET\": {\"auth_entry\": [\"login\"], \"next\": [\"http://localhost:8734/saml-customer?tpa_hint=saml-default-saml-config-bpant\"], \"idp\": [\"default-saml-config-bpant\"]}, \"POST\": {}}", "time": "2021-04-14T16:32:33.472215+00:00", "event_type": "/auth/login/tpa-saml/", "event_source": "server", "page": null}
edx.devstack.lms                  | [14/Apr/2021 16:32:33] "GET /auth/login/tpa-saml/?auth_entry=login&next=http%3A%2F%2Flocalhost%3A8734%2Fsaml-customer%3Ftpa_hint%3Dsaml-default-saml-config-bpant&idp=default-saml-config-bpant HTTP/1.1" 302 0
edx.devstack.lms                  | 2021-04-14 16:33:48,804 INFO 3006 [tracking] [user None] [ip 172.20.0.1] logger.py:41 - {"name": "/auth/complete/tpa-saml/", "context": {"user_id": null, "path": "/auth/complete/tpa-saml/", "course_id": "", "org_id": ""}, "username": "", "session": "", "ip": "172.20.0.1", "agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.183 Safari/537.36", "host": "localhost:18000", "referer": "", "accept_language": "en-US,en;q=0.9", "event": "{\"GET\": {}, \"POST\": {\"RelayState\": [\"default-saml-config-bpant\"], \"SAMLResponse\": [\"PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz4KPHNhbWwycDpSZXNwb25zZSBEZXN0aW5hdGlvbj0iaHR0cDovL2xvY2FsaG9zdDoxODAwMC9hdXRoL2NvbXBsZXRlL3RwYS1zYW1sLyIgSUQ9Il9jOTNhMmI3NmI3MTkwZmJjZDY5NDM4ODk3ZjAzMmE1OSIgSW5SZXNwb25zZVRvPSJPTkVMT0dJTl84MzYzZTYxNzZiOTQzZjcyN2M4M2FkMDViOTRhMmM5MDM3NzBjNGY3IiBJc3N1ZUluc3RhbnQ9IjIwMjEtMDQtMTRUMTY6MzM6NDguNDAyWiIgVmVyc2lvbj0iMi4wIiB4bWxuczpzYW1sMnA9InVybjpvYXNpczpuYW1lczp0YzpTQU1MOjIuMDpwcm90", "time": "2021-04-14T16:33:48.804498+00:00", "event_type": "/auth/complete/tpa-saml/", "event_source": "server", "page": null}
edx.devstack.lms                  | 2021-04-14 16:33:49,461 INFO 3006 [celery.app.trace] [user None] [ip 172.20.0.1] trace.py:125 - Task lms.djangoapps.email_marketing.tasks.update_user[83e1108e-6391-488a-a63a-20adf24a295f] succeeded in 0.09649610001360998s: None
edx.devstack.lms                  | [14/Apr/2021 16:33:54] "POST /auth/complete/tpa-saml/ HTTP/1.1" 302 0
edx.devstack.lms                  | 2021-04-14 16:33:54,739 INFO 3006 [tracking] [user None] [ip 172.20.0.1] logger.py:41 - {"name": "/auth/complete/tpa-saml/", "context": {"user_id": null, "path": "/auth/complete/tpa-saml/", "course_id": "", "org_id": ""}, "username": "", "session": "82da900ab2c6700b8802a4df17128f1e", "ip": "172.20.0.1", "agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.183 Safari/537.36", "host": "localhost:18000", "referer": "", "accept_language": "en-US,en;q=0.9", "event": "{\"GET\": {}, \"POST\": {}}", "time": "2021-04-14T16:33:54.738953+00:00", "event_type": "/auth/complete/tpa-saml/", "event_source": "server", "page": null}
edx.devstack.lms                  | 2021-04-14 16:33:55,089 INFO 3006 [audit] [user 23] [ip 172.20.0.1] models.py:2566 - Login success - rick (rsanchez@samltest.id)
edx.devstack.lms                  | [14/Apr/2021 16:33:56] "GET /auth/complete/tpa-saml/? HTTP/1.1" 302 0
```

edx-abolger (approved):

binodpant:
Need to investigate why this url in prod does not work ( I simulated what would happen if this PR ships, in prod) and got a failure by way of js error and page loads but not the login form:

https://courses.edx.org/login?next=https%3A%2F%2Fenterprise.edx.org%2Fpied-piper-non-integrated-test%3Ftpa_hint%3Dsaml-pied-piper-non-integrated-test&tpa_hint=saml-pied-piper-non-integrated-test

I see 

```
Uncaught TypeError: Cannot read property 'name' of undefined
    at eval (eval at _.template (lms-base-vendor.188b06b0a06d.js:6), <anonymous>:8:130)
    at template (lms-base-vendor.188b06b0a06d.js:6)
    at lms-base-vendor.188b06b0a06d.js:8
    at child.render (logistration_factory.3f1b66764e6b.js?raw:1)
    at child.hinted_login (logistration_factory.3f1b66764e6b.js?raw:2)
    at child.loadForm (logistration_factory.3f1b66764e6b.js?raw:2)
    at child.postRender (logistration_factory.3f1b66764e6b.js?raw:2)
    at child.render (logistration_factory.3f1b66764e6b.js?raw:2)
    at child.initialize (logistration_factory.3f1b66764e6b.js?raw:2)
    at child.Backbone.View (lms-base-vendor.188b06b0a06d.js:10)
```

binodpant:
After a chat with @chavesj  and @georgebabey this afternoon, here are a couple of open questions to explore before we can merge this PR:

(Setup a tentative 11 am tomorrow )

* Do we need to pass in `tpa_hint` as a parameter to the `next` url being formed in this proxy login view, if so why? looks like passing it into the`next` parameter was introduced in this PR https://github.com/edx/edx-enterprise/commit/2f34adac7b2ad87b9e3a2d749be05e2aadce5270, so let's follow up where / how it's used and if we need to pass this along as well 
* If we ship this PR as of current state it adds `tpa_hint` to both the `next` url and also as a param to the /login redirect url. It fixes the 404 but the logistration branded pages open up and gives a javascript error (listed above in PR comment).
* If we remove the extra `tpa_hint` added to the end of the redirect url (meaning don't even submit this PR), this bug reported in this PR occurs (404 so user does not even get a chance to login/register)

So we have one broken flow in each case. Hence need to regroup and find a correct fix.

Also see: https://edx-internal.slack.com/archives/C014MDBL8JW/p1618426083483500 for some context/discussion

binodpant:
how the /login endpoint processes this request: https://github.com/edx/edx-platform/blob/3152c14192dcccac79b7b7a76eba91b380eba78a/openedx/core/djangoapps/user_authn/views/login_form.py#L166

binodpant:
After another full day of investigation, 

I think we can distill all this into one question now:

Do we want to take a user to the hinted login page, when an enterprise_customer has been located based on a tpa_hint present in the url?

If yes, we will need to fix the page to submit my fix. If no, we should change the logic in the flow to take them to a different page. Only then will my proposed fix be unblocked (otherwise users will always be led to the broken hinted login page with my PR)

binodpant:
all updates in JIRA (noting so this does not look like a dead PR)

binodpant:
ok, so edx-platform is brewing a fix for the hinted login page error

once we fix / merge that, we can release this PR hopefully without issues

https://github.com/edx/edx-platform/pull/27412

binodpant:
closing in favor of this PR https://github.com/edx/edx-platform/pull/27431

binodpant:
closing in favor of https://github.com/edx/edx-platform/pull/27431

PR #1215, opened 2021-04-13
perf: :zap: add an index to the username field in data sharing consen…

edx-abolger:
…t so queries using the username don't take so long

To be used when running update username scripts and other username based lookups.

**Merge checklist:**
- [ N/A ] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files)
- [ N/A  ] `make static` has been run to update webpack bundling if any static content was updated
- [ x ] `./manage.py makemigrations` has been run

**Post merge:**


binodpant:
This seems fine to me! If we query a lot based off of username, this makes total sense. I have not added an index in edx models before but looks like this is fully hands-off once done, based on https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations#EverythingAboutDatabaseMigrations-Howtoaddindextoexistingtable(AWSAurora)

binodpant (approved):
lgtm

consent/migrations/0005_auto_20210413_1644.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('consent', '0004_datasharingconsenttextoverrides'),
+    ]
+
+    operations = [
+        migrations.AddIndex(
+            model_name='datasharingconsent',
+            index=models.Index(fields=['username'], name='consent_dat_usernam_fae23a_idx'),
```
raq929, line 15: What's behind this name for it?

edx-abolger: That's randomly generated by the makemigration tool.

iloveagent57 (approved):
Looks good, needs boilerplate.

edx-abolger:
Not going to merge this, the plan was to merge this, run the job that this sped up, and then possibly remove the index, but the job has now been completed without this enhancement, and future related items should be designed as a group.

PR #1218, opened 2021-04-15
fix: improves performance of enterprise role assignment admin page

iloveagent57:
### Overview
* Deletes custom get_search_results() method, since `enterprise_customer__name` is now a viable search field
* Improves pagination by asking for an estimated row count from Mysql INFORMATION_SCHEMA.TABLES
* Turns 1 + N query into 1 query via proper use of list_select_related

https://openedx.atlassian.net/browse/ENT-4384

### Queries now performed in the list view
1. Select all the roles (very quick, small number of roles) for dropdown/filter.
2. Get an estimate of total rows from INFORMATION_SCHEMA (very quick)
2a. If the list is actually a search result, we'll do a real `count()` against the filtered queryset.  This will typically be a small queryset - most usage is to find assignments for specific learners, or all assignments for an enterprise customer.
3. One query to get assignments and all related objects, ordered by the PK of the main table, limited to 25 records at a time - this takes 10 seconds or less on the read-replica.  Prior to this change, we were effectively making this query twice: once to get the total count of records, and once to load the page of 25 records.

### To test locally
* Install this branch of enterprise into your local LMS container.  Restart the devserver.
* Go to http://localhost:18000/admin/enterprise/systemwideenterpriseuserroleassignment/?q=piper
* Open the Django Toolbar and click on the SQL tab - you'll get a list of all queries run when loading a page.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/admin/__init__.py
+++ b/enterprise/admin/__init__.py
@@ -4,6 +4,7 @@ Django admin integration for enterprise app.
 """
 
 import json
+import logging
 
 from config_models.admin import ConfigurationModelAdmin
 from django_object_actions import DjangoObjectActions
@@ -15,6 +16,8 @@ from django.conf import settings
 from django.conf.urls import url
 from django.contrib import admin, auth
 from django.core.exceptions import MultipleObjectsReturned, ObjectDoesNotExist
+from django.core.paginator import Paginator
+from django.db import connection
 from django.db.models import Q
 from django.http import HttpResponseRedirect
 from django.urls import reverse
@@ -70,6 +73,9 @@ try:
 except ImportError:
     EnrollmentAttributeOverrideView = None
 User = auth.get_user_model()
+
+
+logger = logging.getLogger(__name__)
 
 
 class EnterpriseCustomerBrandingConfigurationInline(admin.StackedInline):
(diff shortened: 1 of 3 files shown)


enterprise/admin/__init__.py
```diff
@@ -831,22 +833,59 @@ def get_fields(self, request, obj=None):
...
+        """
+        query = self.object_list.query
+
+        if query.where:
+            return super().count
+        if not getattr(self, '_whole_table_count', None):
+            try:
+                with connection.cursor() as cursor:
+                    cursor.execute(
+                        'SELECT TABLE_ROWS FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = %s',
```
adamstankiewicz, line 857: Does the MySQL user for enterprise-catalog have permissions for `INFORMATION_SCHEMA` or only `enterprise_catalog` in stage/prod? I imagine it does but figured it's worth asking as I'm not too certain whether our databases have individual MySQL users per service or if they share a root MySQL user, for example.

enterprise/admin/__init__.py
```diff
@@ -831,22 +833,59 @@ def get_fields(self, request, obj=None):
...
+                        [query.model._meta.db_table]
+                    )
+                    self._whole_table_count = int(cursor.fetchone()[0])
+            except Exception:  # pylint: disable=broad-except
+                self._whole_table_count = self.ARBITRARILY_LARGE_NUMBER
```
adamstankiewicz, line 862: nit: Is it worth adding a `logger.exception` here for more visibility into when the raw MySQL query raises an exception for whatever reason? Without it, would the only way we know something went wrong be based on the fact that the table says there are exactly `self.ARBITRARILY_LARGE_NUMBER` number of rows?

adamstankiewicz (approved):
Looks good to me, with a clarification/curiosity question and a tiny nit. 

Great refactor to include `enterprise_customer__name` as a `search_field` to remove the less performant `get_search_results` 🚀

enterprise/admin/__init__.py
```diff
@@ -831,22 +833,59 @@ def get_fields(self, request, obj=None):
...
+
+    # pylint: disable=attribute-defined-outside-init
+    @property
+    def count(self):  # pylint: disable=invalid-overridden-method
+        """
+        Returns the number of items in the object list (possibly an estimate).
+        """
+        query = self.object_list.query
+
+        if query.where:
```
adamstankiewicz, line 851: [clarification] Does `query.where` also include non-search filters (e.g., by role)?

image

enterprise/admin/__init__.py
iloveagent57: There are multiple MySQL/Aurora instances we use - there's a shared one that several IDAs, including enterprise-catalog are on; LMS/Studio have their own instance.  It tends to be a single user per service for all of the connections.

enterprise/admin/__init__.py
iloveagent57: Yeah, it does, and the example you provided will be a pretty large queryset in prod (assuming there's no search term to go along with that filter).  I'm inclined to say "let's see what happens" for now.

PR #1221, opened 2021-04-22
fix: add list of enterprise catalog uuids to EnterpriseCustomerSerializer

adamstankiewicz:
### Ticket

[ENT-4154](https://openedx.atlassian.net/browse/ENT-4154)

### Example API response (truncated)

```json
{
    "next": null,
    "previous": null,
    "count": 2,
    "num_pages": 1,
    "current_page": 1,
    "start": 0,
    "results": [
        {
            "uuid": "bdd488ad-3f6e-49b7-9fb4-99443afad65f",
            "name": "Test Enterprise",
            "slug": "test-enterprise",
            "enterprise_customer_catalogs": [
                "d0e5d354-205c-4cae-85c2-eb9abdf5b382"
            ]
        }
    ]
}
```

### Testing instructions

1. Ensure your devstack has at least one `EnterpriseCustomer` configured with an enterprise catalog.
2. Visit http://localhost:18000/enterprise/api/v1/enterprise-customer/.
3. Observe additional `enterprise_customer_catalogs` field in the response containing a list of catalog UUIDs.

### **Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * Nothing.
 
+[3.22.9]
+--------
+* Expose enterprise catalog uuids associated with an Enterprise Customer in the ``enterprise-customer`` API endpoint.
+
 [3.22.8]
 --------
 * Add dashboard admin rbac role permission on tableau auth view so that only
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.22.8"
+__version__ = "3.22.9"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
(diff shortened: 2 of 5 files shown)


chavesj (approved):
🚀

iloveagent57 (approved):
Just a note from a conversation I had in my head:
Why not include `catalog_query_uuids` in this serializer, too, since we now sync those between edx-enterprise and enterprise-catalog?  Wouldn't this save an HTTP request to enterprsie-catalog?
Answer: in edx-enterprise, not every catalog has an associated catalog query (i.e. many catalogs have a direct `content_filter` field).  In enterprise-catatalog, every catalog is associated with a catalog query object, so we should still rely on the sequence of
1. Get all the catalog UUIDs for a given customer from edx-enterprise.
2. Ask enterprise-catalog for all of the query UUIDs associated with those catlaog UUIDs.

adamstankiewicz:
@iloveagent57 Yep, that's also my understanding of the next steps for this work. 

In learner portal, we'll get this additional `enterprise_customer_catalogs` field in the API response so it's accessible throughout the app, and will need to make that additional API call to enterprise-catalog to fetch the catalog query uuids for the search filtering, given these catalog uuids associated with the customer. All of this will should be covered through [ENT-4155](https://openedx.atlassian.net/browse/ENT-4155).

PR #1236, opened 2021-05-10
feat: :construction: Add enterprise to seeded users and roles

edx-abolger:
Allows for repeated running of the following command for easy creation of fresh enterprise setup on devstack:
By default no system level enterprise_admin user will exist in seed data.

Usage: 
1. `make lms-shell` to get into the lms. 
2. `./manage.py lms seed_enterprise_devstack_data --name '"My Test Enterprise for Ticket 1234"`

Result will be an admin, learners, enterprise etc all keyed to the new enterprise only. 

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/management/commands/seed_enterprise_devstack_data.py
+++ b/enterprise/management/commands/seed_enterprise_devstack_data.py
@@ -112,7 +112,7 @@ class Command(BaseCommand):
         """ Ensures the `ENTERPRISE_ENROLLMENT_API_ACCESS_GROUP` is created """
         Group.objects.get_or_create(name=ENTERPRISE_ENROLLMENT_API_ACCESS_GROUP)
 
-    def _create_enterprise_user(self, username, role):
+    def _create_enterprise_user(self, username, role, enterprise_customer=None):
         """
         Creates a new user with the specified `username` and `role` (e.g.,
         'enterprise_learner'). The newly created user is added to the
@@ -140,7 +140,10 @@ class Command(BaseCommand):
             if user:
                 self._add_name_to_user_profile(user)
                 self._add_user_to_groups(user=user, role=role)
-                self._create_system_wide_role_assignment(user=user, role=role)
+                self._create_system_wide_role_assignment(
+                    user=user,
+                    role=role, 
+                    enterprise_customer=enterprise_customer)
                 self._create_feature_role_assignments(user=user, role=role)
                 return {
                     "user": user,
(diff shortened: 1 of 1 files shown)


enterprise/management/commands/seed_enterprise_devstack_data.py
```diff
@@ -254,22 +257,31 @@ def handle(self, *args, **options):
...
                 role=ENTERPRISE_ADMIN_ROLE,
+                enterprise_customer=enterprise_customer
+
             ),
             self._create_enterprise_user(
-                username=ENTERPRISE_OPERATOR_ROLE,
-                role=ENTERPRISE_OPERATOR_ROLE
+                username="{}_{}".format(ENTERPRISE_OPERATOR_ROLE,slug),
+                role=ENTERPRISE_OPERATOR_ROLE,
+                enterprise_customer=enterprise_customer
```
taliaedX, line 281: Given:
```
def has_access_to_all_contexts(self):
        """
        Returns true if the role for this assignment is ``ENTERPRISE_OPERATOR_ROLE``,
        or if ``applies_to_all_contexts`` is true; returns false otherwise.
        """
        return (self.role.name == ENTERPRISE_OPERATOR_ROLE) or self.applies_to_all_contexts
```
I don't know if enterprise here matters?

enterprise/management/commands/seed_enterprise_devstack_data.py
```diff
@@ -212,14 +217,12 @@ def _create_enterprise_customer_user(self, username, enterprise_customer):
         )
         return enterprise_customer_user
 
-    def _create_enterprise(self, enterprise_name, enterprise_users):
+    def _create_enterprise_linked_data(self, enterprise_users, enterprise_customer):
         """
-        Creates an enterprise and its associated data, including the
-        EnterpriseCustomer, an enterprise catalog, and initial users of
+        Creates enterprise data for a given EnterpriseCustomer and its 
+        associated data, including an enterprise catalog, and initial users of
```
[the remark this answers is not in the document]
edx-abolger: I moved the creation of the EnterpriseCustomer outside the function. But will clean up

enterprise/management/commands/seed_enterprise_devstack_data.py
edx-abolger: Good catch, will remove.

iloveagent57 (approved):
LGTM (agreed with Talia's comments).

PR #1237, opened 2021-05-11
fix: :bug: ENT-4525 do not process enterprise enrollments without matching courseEnrollment

binodpant:
Why?
Learners cannot even get into a course page or the learner portal in general in any useful way. until this is fixed. This case occurs when the `enterprise_course_enrollments` endpoint is called and it encounters even 1 EnterpriseCourseEnrollment records, whose self.course_enrollment is None. This can happen for direct integrations, even though the exact root cause is not yet known. Still since the learner is not even enrolled officially yet, we should really not block the learner

This is a high value add change that makes learners able to use the learner poratl even while we investigate the underluing causes of the bad records.

What is included?

* - [x] The fix: Just a filtering of the enrollments once they are fetched, to exclude any with absent course_enrollment properties (it's a cached property not a foreign key)
* - [x] Needed test changes

#### Accessing learner portal

I locally changed the value of the course id in an existing `student.CourseEnrollment` record that I already had, in the database for LMS locally, to an incorrect value, which basically amounts to the same result

I assigned one code to the learner using ecommerce page so I had something to work with, there is no license subsidy here
Then I visited http://localhost:8734/test-enterprise while logged in as that learner test_enterprise_2@example.com

reproduces the issue since the query to fetch course_enrollment records returns empty in that case

Once I apply this fix I am able to get into the learner portal and see this page

Screen Shot 2021-05-11 at 12 46 13 PM

#### Enrolling in a course thus accessed once learner is in this odd state:

I actually reproduced the situatoin in local by first ensureing there is a valid enterprisecourseenrollment for the learner but messing up the record in student.courseEnrollment table (changing the course_id) so that effectively there is no longer a courseEnrollment to be found

when I do that I get this screen instead of the screen I got from the case with courseEnrollment intact. It now says 'Enroll' and that makes sens ebecause the response of the `enterprise_course_enrollments` is now an empty array, instead of the prior response of 500

Screen Shot 2021-05-11 at 4 36 28 PM

Now attempted to enroll the learner in this course:

I get the screen to enroll, but unable as of yet to proceed due to ecommerce running into problems. I might have reprovision or reset it some other way to verify enrollment fully

Screen Shot 2021-05-11 at 4 47 08 PM

JIRA: ENT-4525

**Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * Nothing.
 
+[3.23.1]
+---------
+* Fix: filter out EnterpriseCourseEnrollments without corresponding CourseEnrollment records in learner portal view.
+
 [3.23.0]
 ---------
 * Added support for ``--enrollment-before`` and ``--no-commit`` params in ``email_drip_for_missing_dsc_records`` command.
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.23.0"
+__version__ = "3.23.1"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/enterprise_learner_portal/api/v1/views.py
+++ b/enterprise_learner_portal/api/v1/views.py
@@ -83,10 +83,12 @@ class EnterpriseCourseEnrollmentView(APIView):
             enterprise_customer_user=enterprise_customer_user
         )
 
-        course_overviews = get_course_overviews(enterprise_enrollments.values_list('course_id', flat=True))
+        filtered_enterprise_enrollments = [record for record in enterprise_enrollments if record.course_enrollment]
+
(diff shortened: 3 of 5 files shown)


iloveagent57 (approved):
Seems reasonable.  It would be good to add a unit test that covers the case where there's an ECE without a `course_enrollment`, but don't let that block you from shipping this change now.

iloveagent57 (approved):

binodpant:
ok locally for me, I was able to finally load up a course page in L P delete the courseEnrollment record, hit enroll again in L P and verified it worked

Also hit up on Alex D's machine removed the course enrollment record then revisited learner portal and even though we could not enroll a learner (since there were other issues with catalog server and what not) I am going to go with this improvement and we will see how far this takes us for this learner! rest of testing will be in staging and the real testing will be in the customer's account in prod

binodpant:
Tested enrollment from staging for my learner, normal enrollment flow is working. Was able to enroll in this course from the learner portal

https://courses.stage.edx.org/courses/course-v1:edx+1244+1T2021/course/

No errors in the endpoint, enrollments come back fine in the normal use case (without this kind of issue)

Once this is in prod will test the bug case with Josefina

PR #1240, opened 2021-05-13
docs: ADR for bulk enrollment endpoint

binodpant:
ADR for bulk enrollment backend in edx-enterprise (with reference also to the one in license-manager)

**Merge checklist:**

**Post merge:**

### DIFF
--- a/docs/decisions/0005-bulk-enrollment.rst
+++ b/docs/decisions/0005-bulk-enrollment.rst
@@ -15,13 +15,18 @@ needs a suitable backend.
 Various enrollment apis exist, but they do not cater to or optimize for multiple enrollments.
 
 There is an `EnrollmentApiClient` which has an `enroll_user_in_course()` method used by:
-  * models::EntrpriseCustomerUser::enroll()
-  * utils::enroll_user()
+  * models::EnterpriseCustomerUser::enroll()
+  * utils::enroll_user() ( handles multiple courses )
   * views::CourseEnrollmentView::post()
   * views::HandleConsentEnrollment::get()
   * views::GrantDataSharingPermission::_enroll_learner_in_course()
 
-An api endpoint is thus needed that caters to the bulk use case without causing request overload.
+There is also a method to create pending user enrollments for non-edX users at:
+  `models::EnterpriseCustomer::enroll_user_pending_registration()`
+This method also handles multiple courses.
+
+Due to the scale and volume considerations, an api endpoint is needed that caters to the bulk
+use case without causing request overload.
 
 Decisions
 =========
@@ -32,7 +37,7 @@ Decisions
   `EnrollmentApiClient` for now, but there is a cleanup effort anticipated to replace it with
   python api calls which is not in scope for this effort.
 * For emails not matching existing users in edX, we will use the
-  `models::enroll_user_pending_registration()` method which also handles multiple courses.
+  `models::EnterpriseCustomer::enroll_user_pending_registration()` method
 * We will create an ecom order for each successful enrollment of an existing edX user.
 * Any failures in individual enrollments will not cause failures of the entire batch. In other
   words, the endpoint will be non transactional. This avoid wasteful and confusing workflow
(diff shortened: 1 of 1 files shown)


docs/decisions/0005-bulk-enrollment.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Accepted
+
+Context
+=======
+
+Business need: The admin portal feature to enroll multiple learners in multiple subscription-courses
+needs a suitable backend.
+
+Various enrollment apis exist, but they do not cater to or optimize for multiple enrollments.
```
edx-abolger, line 15: suggestion: You may wish to cross check this part with the (not publicly available so can't just link to it) Miro board that covers our enrollment paths, just to see if everything is covered https://miro.com/welcomeonboard/ce4L6lSmyZ1mMN8FX9XghAihnyk6nuBDGezLZ3AXnyJrAehZvUtpqbrq410uB0mO

docs/decisions/0005-bulk-enrollment.rst
```diff
@@ -0,0 +1,59 @@
...
+  * views::CourseEnrollmentView::post()
+  * views::HandleConsentEnrollment::get()
+  * views::GrantDataSharingPermission::_enroll_learner_in_course()
+
+There is also a method to create pending user enrollments for non-edX users at:
+  `models::EnterpriseCustomer::enroll_user_pending_registration()`
+This method also handles multiple courses.
+
+Due to the scale and volume considerations, an api endpoint is needed that caters to the bulk
+use case without causing request overload.
```
edx-abolger, line 29: I would add "rate limiting" here as well since that is the technical vocabulary definition of what we ran into.

docs/decisions/0005-bulk-enrollment.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Decisions
+=========
+
+* We will add a new endpoint to edx-enterprise to handle bulk enrollment in a single request.
+* For existing edX users, we will reuse the `utils::enroll_user()` method for this use case.
+  This method handles multiple courses already. That method still ends up calling the `EnrollmentApiClient` for now.
+  But there is a cleanup effort anticipated to replace it with python api calls which is not in scope for this effort.
```
edx-abolger, line 37: Replace the pronouns with descriptive words so I don't have to guess what 'it' is (I only know what you mean because I am deeply involved in the project right now). Example:  "to replace the existing HttpClient calls in enroll_user() with direct calls to the enroll_user() python functions defined in the parent edx-platform Django application."

docs/decisions/0005-bulk-enrollment.rst
```diff
@@ -0,0 +1,59 @@
...
+* For emails not matching existing users in edX, we will use the `models::EnterpriseCustomer::enroll_user_pending_registration()` method
```
edx-abolger, line 38: I would change 'edX' to 'edx-platform' or 'the edX LMS Django application'.  'edX' is the company, not the technical component, and this library is open source.

docs/decisions/0005-bulk-enrollment.rst
```diff
@@ -0,0 +1,59 @@
...
+* We will create an ecom order for each successful enrollment of an existing edX user.
```
edx-abolger, line 39: ecom => 'ecommerce'.

docs/decisions/0005-bulk-enrollment.rst
binodpant: thanks! yes I went over that, looks like the paths mentioned there lead to enrolling via the GrantDatasharingPermissions view which calls one of the methods I listed here, this is good to see though!

docs/decisions/0005-bulk-enrollment.rst
binodpant: indicated the method we call and fixed copy thanks

docs/decisions/0005-bulk-enrollment.rst
```diff
@@ -0,0 +1,61 @@
...
+* We will not have MFEs call this endpoint directly. Rather they will call a license-manager endpoint which will call this endpoint.
+  There are several reasons for this:
+  #. Only licensed learners should be enrolled in the subscription's courses. This design will guard against any frontend requests containing non-desired learners.
+  #. Calling license-manager from edx-enterprise requires edx-enterprise to know about license-manager which is counter to the archiectural design for edx-enterprise
+  #. Finally, semantically it makes sense to request a license management service to authorize license based enrollments.
+* The edx-enterprise endpoint will return a response with these pieces of information:
+  `{'successes': [], 'pending': [], 'failures': []}`. Note, pending enrollments are supported.
+  This will allow the license-manager to inform the user if any enrollments failed, while successfully (meaning: success or pending) enrolling the rest, in a single request cycle.
+* The endpoint will be safe to be invoked with pre-existing enrollment pairs (meaning learner email + course_id)
```
edx-abolger, line 52: Clarification: this statement means that if you try to enroll a learner in a course they are already enrolled in, you will get a successes:[] result for that pair, correct? If so can you add that as a note? Othere than that, LGTM

edx-abolger (approved):

docs/decisions/0005-bulk-enrollment.rst
binodpant: indeed, I added that clarification to this point thanks

PR #1246, opened 2021-05-20
feat: Use new email template for bulk enrollment ENT-4322

binodpant:
Included:

- Added  a model that holds a template. This template will be used only for bulk enrollment. We only use the first item from the queryset results so holding multiple records in the table at the moment is useless. We have it as a model only because we want to avoid updating code update the template. This code may get removed/replaced if we decide to use Braze in the future for these emails.
- Refactored the existing model `EnrollmentEmailNotificationTemplate` code out into an abstract base model which both this new template model, and the existing enrollment email template model now inherit from.
- Added a default False `bulk_enroll_mode` parameter to allow max. reuse of existing code
- Changed bulk enrollment `enroll_learners_in_courses` method to use this template instead using above switch

JIRA: ENT-4322

### Testing notes:

- bulk enrolled two learners into a course in localhost:1991/test-enterprise
- checked the email folder for a new email (devstack by default will direct django mail to file) 
```
vim /edx/src/ace_messages/20210601-032635-140670805145344.log
```
- saved the contents of the file into an html file and opened in browser
looks like this (I will cleanup CSS of the My Dashboard in the Django template content which is not being checked into git)

Screen Shot 2021-05-31 at 11 32 07 PM

**Merge checklist:**

**Post merge:**


enterprise/api/v1/views.py
```diff
@@ -318,6 +318,7 @@ def enroll_learners_in_courses(self, request, pk):
                     catalog_api_user=request.user,
                     course_id=course_run,
                     users=pending_users | existing_users,
+                    bulk_enroll_mode=True,
```
binodpant, line 321: this allows us to use a different template without having to refactor most of the remaining code, in a backward compatible way

binodpant:
Need to cleanup two things:
* center the layout of the social links
* fix link CSS for 'My Dashboard'

enterprise/models.py
```diff
@@ -2271,19 +2278,58 @@ def render_template(self, template_text, kwargs):
...
+        """
+        return self.__str__()
+
+
+@python_2_unicode_compatible
+class BulkEnrollmentNotificationEmailTemplate(BaseEnrollmentNotificationEmailTemplate):
+    """
+        Why a new template instead of using 'EnrollmentNotificationEmailTemplate'?
+        1: we do not need customer linking, since template will not be per customer
+        2: we still want to do a model to avoid having template in text/html files on git
```
taliaedX, line 2293: 1. Make it optional then
2. Why would you need to do this? If you just add a type column to the existing model you can then map that to your `bulk_enroll_mode` flag and just update the EnrollmentNotificationEmailTemplate uses to also pick that type.

enterprise/utils.py
```diff
@@ -438,25 +438,14 @@ def get_notification_subject_line(course_name, template_configuration=None):
         return stock_subject_template.format(course_name=course_name)
 
 
-def send_email_notification_message(user, enrolled_in, enterprise_customer, email_connection=None):
```
taliaedX, line 441: again not convinced we couldn't re-use this.

enterprise/utils.py
```diff
@@ -533,13 +606,34 @@ def enterprise_course_enrollment_model():
...
     """
     returns the ``LicensedEnterpriseCourseEnrollment`` class.
     """
     return apps.get_model('enterprise', 'LicensedEnterpriseCourseEnrollment')  # pylint: disable=invalid-name
 
 
+def get_enterprise_bulk_enroll_email_template():
+    """
+    Get the first instance found of `BulkEnrollmentNotificationEmailTemplate`
+    Currently we only expect to use one of these, since it's not per customer.
```
taliaedX, line 626: What is the future of this? Is there a future of this or do we really only ever expect 1? Because if we only ever expect 1, I'm extra not a fan of a new model.
We should honestly design this for the future unless that future is 100% "we're just gonna stick it in Braze"

taliaedX (commented):
This seems functional, but the comments explaining the design decisions don't really convince me about the decisions made.

I'd like to see us determine what the future use-case of this email template could be, because in most cases I'd expect us to refactor the existing model rather than make a one off model that contains a single row and we only ever expect to contain a single row.

enterprise/utils.py
binodpant: I think the future is going to be we use Braze.. since it looks like we are standardizing on it. So this I see as a stop gap only. but who knows how long this will stay here? I also confirmed with Joe/Jen that we are not looking for per customer customization of this template at this time.

enterprise/models.py
binodpant: Ah I did not want to use the existing model since it's customer specific and needs to fill in a customer uuid. this one is going to be a fixed template (for now anyway) for all bulk enrollment.

Also, for this one I decided it's best to not send emails if there is no template, as opposed to current design of defaulting to the text files (since I would then also have to add text based template as a fallback)... :(

enterprise/utils.py
binodpant: yeah I guess I can still reuse this method with an optional bulk enroll flag passed to it? that seems doable, but it will still need me to add branching logic since for bulk enroll I was going to not use the fallback text/html files...but other than that it's largely reusable

enterprise/models.py
binodpant: Having said that.. there is an option: just exploring...

I confirmed that the `EnrollmentNotificationEmailTemplate` model does require a enterprise_customer to be specified when creating...so the only way to re-use all this setup for bulk enroll, is to also go with text/html based templates as a default fallback for bulk enrollment, and add a column (like type='bulk_enrollment') to the existing model. In bulk enroll mode we find the template that matches that type. Presumably only one exists per type (we can also unique key those fields type + enterprise_customer maybe). When a row with type='bulk_enrollment' is found we use that... then we won't need a new model. When Lael and I looked at this first, it seemed to use it was cleaner to use a new model, but this option can also be made to work. Honestly I am equally sold on both.

What's your pref based on what we know of the use cases? This option seems totally doable as an alternative to the current PR approach of only using one row of the table which I agree is odd.

binodpant:
To avoid confusion I am going to create a new PR instead of updating this one, with the new approach, coming up!

PR #1248, opened 2021-05-26
edx-abolger/ent 4103 log standardization

edx-abolger:
refactor(integrated_channels): 🔊 configure top level channel logs to always print standard config information

Types of changes you will see in this PR:
1. Modify top level/batch integrated channels logs to always log a standard set of information of integrated channels.
2. Modify as many log statements as possible to remove LMS username and LMS user email in favor of LMS user id. 
3. Modify tests to avoid test acceptance criteria based on logs.
4. Adding standardized stack trace information to Logger.error statements instead of either ignoring it or suppressing it in favor of home-grown error formats.
5. Updates to docstrings and log messages that give Python object type information about what has gone wrong for better debugging.
6. A standardized format for non-error messages of the form `<method_name> <started/finished/failed>. <details>` for easier tracing from logs of what is going on.
7. Some additional parameters added to methods if it was necessary for complete logging.
8. A specific format for logging batch tasks in some places to distinguish the log from logging something related to a particular learner or course of the form: `Batch <method_name> <started/finished/failed>. <details>`
9. Does change code logic in the case of duration time logged so that start and end logs have times that actually match (in some log cases, records for starting+finishing a task were not 1-1, making the existing duration field useless as a tool for log exploration

This PR does _not_:
(1) fully replace Integrated channels logs for all 7 integrated channels with the new format. Those can be done 1 by 1 as needed now that a format exists.
(2) Remove 100% of formatted email from logs _in cases where there is no matching LMS User, as if it did (in the case of a misspelling, for example) we would not be able to debug the problem- there is no external reference to be had in some of these cases._ 
(3) Re-enable the existing but skipped log-based tests around learner data integration, because those tests are some kind of catch all with no criteria in the source code, and should be removed and written with actual criteria. 

**Merge checklist:**
- [ n/a ] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files)
- [ n/a ] `make static` has been run to update webpack bundling if any static content was updated
- [ n/a ] `./manage.py makemigrations` has been run
- [ n/a ] [Version](https://github.com/edx/edx-enterprise/blob/master/enterprise/__init__.py) bumped
- [n/a  ] Translations updated (see docs/internationalization.rst but also this isn't blocking for merge atm)

**Post merge:**

### DIFF
--- a/integrated_channels/blackboard/exporters/learner_data.py
+++ b/integrated_channels/blackboard/exporters/learner_data.py
@@ -34,8 +34,9 @@ class BlackboardLearnerExporter(LearnerExporter):
         enterprise_customer_user = enterprise_enrollment.enterprise_customer_user
         if enterprise_customer_user.user_email is None:
             LOGGER.debug(
-                'No learner data was sent for user [%s] because a Blackboard user ID could not be found.',
-                enterprise_customer_user.username
+                'No learner data was sent for LMS User [%s] because a Blackboard user ID could not be found for customer [%s]',
+                enterprise_customer_user.user_id,
+                enterprise_customer_user.enterprise_customer.name
             )
             return None
         percent_grade = kwargs.get('grade_percent', None)
@@ -79,8 +80,9 @@ class BlackboardLearnerExporter(LearnerExporter):
         if enterprise_enrollment.enterprise_customer_user.user_email is None:
             # We need an email to find the user on blackboard.
             LOGGER.debug(
-                'No learner data was sent for user [%s] because a blackboard user ID could not be found.',
-                enterprise_enrollment.enterprise_customer_user.username
+                'No learner data was sent for LMS User [%s] because a Blackboard user ID could not be found for customer [%s]',
+                enterprise_enrollment.enterprise_customer_user.user_id,
+                enterprise_enrollment.enterprise_customer_user.enterprise_customer.name
             )
             return None
 
--- a/integrated_channels/blackboard/exporters/learner_data.py
+++ b/integrated_channels/blackboard/exporters/learner_data.py
@@ -34,7 +34,7 @@ class BlackboardLearnerExporter(LearnerExporter):
         enterprise_customer_user = enterprise_enrollment.enterprise_customer_user
         if enterprise_customer_user.user_email is None:
             LOGGER.debug(
-                'No learner data was sent for LMS User [%s] because a Blackboard user ID could not be found for customer [%s]',
+                'No learner data was sent for LMS User [%s] because Blackboard User ID not found for customer [%s]',
                 enterprise_customer_user.user_id,
                 enterprise_customer_user.enterprise_customer.name
             )
@@ -80,7 +80,7 @@ class BlackboardLearnerExporter(LearnerExporter):
         if enterprise_enrollment.enterprise_customer_user.user_email is None:
             # We need an email to find the user on blackboard.
             LOGGER.debug(
-                'No learner data was sent for LMS User [%s] because a Blackboard user ID could not be found for customer [%s]',
+                'No learner data was sent for LMS User [%s] because Blackboard User ID not found for customer [%s]',
                 enterprise_enrollment.enterprise_customer_user.user_id,
                 enterprise_enrollment.enterprise_customer_user.enterprise_customer.name
             )
--- a/integrated_channels/blackboard/exporters/learner_data.py
+++ b/integrated_channels/blackboard/exporters/learner_data.py
@@ -39,9 +39,9 @@ class BlackboardLearnerExporter(LearnerExporter):
                 enterprise_customer_user.user_id,
                 None,
                 ('get_learner_data_records finished. No learner data was sent because '
-                'Blackboard User ID not found for [{name}]'.format(
-                    name=enterprise_customer_user.enterprise_customer.name
-                ))))
+                 'Blackboard User ID not found for [{name}]'.format(
+                     name=enterprise_customer_user.enterprise_customer.name
+                 ))))
             return None
         percent_grade = kwargs.get('grade_percent', None)
         completed_timestamp = None
@@ -89,9 +89,9 @@ class BlackboardLearnerExporter(LearnerExporter):
                 enterprise_enrollment.enterprise_customer_user.user_id,
                 None,
                 ('get_learner_assessment_data_records finished. No learner data was sent because '
-                'Blackboard User ID not found for [{name}]'.format(
-                    name=enterprise_enrollment.enterprise_customer_user.enterprise_customer.name
-                ))))
+                 'Blackboard User ID not found for [{name}]'.format(
+                     name=enterprise_enrollment.enterprise_customer_user.enterprise_customer.name
+                 ))))
             return None
 
         blackboardLearnerAssessmentDataTransmissionAudit = apps.get_model(  # pylint: disable=invalid-name
--- a/integrated_channels/blackboard/exporters/learner_data.py
+++ b/integrated_channels/blackboard/exporters/learner_data.py
@@ -9,7 +9,7 @@ from django.apps import apps
 
 from integrated_channels.catalog_service_utils import get_course_id_for_enrollment
 from integrated_channels.integrated_channel.exporters.learner_data import LearnerExporter
-from integrated_channels.utils import parse_datetime_to_epoch_millis
+from integrated_channels.utils import generate_formatted_log, parse_datetime_to_epoch_millis
 
 LOGGER = getLogger(__name__)
 
@@ -33,11 +33,15 @@ class BlackboardLearnerExporter(LearnerExporter):
         """
         enterprise_customer_user = enterprise_enrollment.enterprise_customer_user
         if enterprise_customer_user.user_email is None:
-            LOGGER.debug(
-                'No learner data was sent for LMS User [%s] because Blackboard User ID not found for customer [%s]',
+            LOGGER.debug(generate_formatted_log(
+                'blackboard',
+                enterprise_customer_user.enterprise_customer.uuid,
                 enterprise_customer_user.user_id,
-                enterprise_customer_user.enterprise_customer.name
-            )
+                None,
+                ('get_learner_data_records finished. No learner data was sent because '
+                'Blackboard User ID not found for [{name}]'.format(
+                    name=enterprise_customer_user.enterprise_customer.name
+                ))))
             return None
         percent_grade = kwargs.get('grade_percent', None)
         completed_timestamp = None
@@ -79,11 +83,15 @@ class BlackboardLearnerExporter(LearnerExporter):
         """
         if enterprise_enrollment.enterprise_customer_user.user_email is None:
             # We need an email to find the user on blackboard.
-            LOGGER.debug(
-                'No learner data was sent for LMS User [%s] because Blackboard User ID not found for customer [%s]',
+            LOGGER.debug(generate_formatted_log(
+                'blackboard',
+                enterprise_enrollment.enterprise_customer_user.enterprise_customer.uuid,
                 enterprise_enrollment.enterprise_customer_user.user_id,
-                enterprise_enrollment.enterprise_customer_user.enterprise_customer.name
-            )
+                None,
+                ('get_learner_assessment_data_records finished. No learner data was sent because '
+                'Blackboard User ID not found for [{name}]'.format(
+                    name=enterprise_enrollment.enterprise_customer_user.enterprise_customer.name
+                ))))
             return None
 
         blackboardLearnerAssessmentDataTransmissionAudit = apps.get_model(  # pylint: disable=invalid-name
--- a/integrated_channels/blackboard/exporters/learner_data.py
+++ b/integrated_channels/blackboard/exporters/learner_data.py
@@ -38,7 +38,7 @@ class BlackboardLearnerExporter(LearnerExporter):
                 enterprise_customer_user.enterprise_customer.uuid,
                 enterprise_customer_user.user_id,
                 None,
-                ('get_learner_data_records finished. No learner data was sent because '
+                ('get_learner_data_records finished. No learner data was sent for this LMS User Id because '
                  'Blackboard User ID not found for [{name}]'.format(
                      name=enterprise_customer_user.enterprise_customer.name
                  ))))
@@ -88,8 +88,8 @@ class BlackboardLearnerExporter(LearnerExporter):
                 enterprise_enrollment.enterprise_customer_user.enterprise_customer.uuid,
                 enterprise_enrollment.enterprise_customer_user.user_id,
                 None,
-                ('get_learner_assessment_data_records finished. No learner data was sent because '
-                 'Blackboard User ID not found for [{name}]'.format(
+                ('get_learner_assessment_data_records finished. No learner data was sent for this LMS User Id because'
+                 ' Blackboard User ID not found for [{name}]'.format(
                      name=enterprise_enrollment.enterprise_customer_user.enterprise_customer.name
                  ))))
             return None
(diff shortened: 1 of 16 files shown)


edx-abolger:
Final squashed commit message for merge: 

refactor(integrated_channels): :loud_sound: configure top level channel logs to always print standard config information

Modified top level/batch integrated channels logs to always log a standard set of information of integrated channels. Modified tests to avoid test acceptance criteria based on logs.

integrated_channels/blackboard/exporters/learner_data.py
```diff
@@ -34,8 +34,9 @@ def get_learner_data_records(
         enterprise_customer_user = enterprise_enrollment.enterprise_customer_user
         if enterprise_customer_user.user_email is None:
             LOGGER.debug(
-                'No learner data was sent for user [%s] because a Blackboard user ID could not be found.',
-                enterprise_customer_user.username
+                'No learner data was sent for LMS User [%s] because Blackboard User ID not found for customer [%s]',
+                enterprise_customer_user.user_id,
```
taliaedX, line 38: So I appreciate and understand removing the user email since that is PII, however, since we now have an additional step to be able to tell what user is affected or find the user affected, I'd prefer a value that we can easily hunt down. Ie I'd prefer if we could use the UUID here.

If I search by UUID, I will find all relevant logs showing this user's journey. However, if I use their ID I will get *a lot* of irrelevant logs that don't even have anything to do with a user.

integrated_channels/integrated_channel/tasks.py
```diff
@@ -73,110 +104,102 @@ def transmit_learner_data(username, channel_code, channel_pk):
...
-    )
+    _log_batch_task_finish('transmit_learner_data', channel_code, api_user.id, integrated_channel, duration)
 
 
 @shared_task
 @set_code_owner_attribute
-def transmit_single_learner_data(username, course_run_id):
+def transmit_single_learner_data(learner_username, course_run_id):
     """
-    Task to send single learner data to each linked integrated channel.
```
taliaedX, line 104: given this is the `single_learner` method and I do think we have a bulk one somewhere, is there a reason to remove "single" here?

integrated_channels/integrated_channel/tasks.py
```diff
@@ -73,110 +104,102 @@ def transmit_learner_data(username, channel_code, channel_pk):
...
-                    user_id=user.id,
-                    enterprise_uuid=enterprise_customer_uuid))
+
+            LOGGER.info(generate_formatted_log(
+                integrated_channel.channel_code(),
+                enterprise_customer_uuid,
+                user.id,
+                course_run_id,
+                'transmit_single_learner_data for channels {} started.'
+                .format([c.channel_code() for c in enterprise_integrated_channels])
```
taliaedX, line 146: Isn't this (and the below log line) going to be generated for each channel? So why would you use a loop here?

Ie we'll see like 4-6 log lines that consist of:
`transmit_single_learner_data for channels Canvas, Blackboard, Moodle, etc. started.` 
That seems extremely redundant and I imagine them being separated helps us verify that a particular channel did actually attempt to transmit something.

taliaedX (commented):
I have two issues that I only commented in the first spot I saw, however they occur many times throughout this PR. Those issues: 
- I'd love if we could use user UUID instead of ID
- we're going from channel specific tracking to listing all channels.

These 2 changes actually make navigating logs more difficult.

integrated_channels/integrated_channel/tasks.py
taliaedX: I see the below you added for the `finish` **does** use the specific channel code, so not sure why we're not doing the same here.

edx-abolger: Probably just a mistake on my part, should just be a single channel.

edx-abolger:
> I have two issues that I only commented in the first spot I saw, however they occur many times throughout this PR. Those issues:
> 
> * I'd love if we could use user UUID instead of ID
> * we're going from channel specific tracking to listing all channels.
> 
> These 2 changes actually make navigating logs more difficult.


- So, lms users don't have UUIDs AFAICT. Do they? (https://open-edx-proposals.readthedocs.io/en/latest/oep-0032-arch-unique-identifier-for-users.html?highlight=lms%20user%20id#user-identifier-candidate-comparison) 
- We should not be listing all the channels: I will re-review and change anything that lists multiple channels when there should only be one.

integrated_channels/blackboard/exporters/learner_data.py
edx-abolger: I can add the Enteprise Customer UUID here.

taliaedX:
>So, lms users don't have UUIDs AFAICT. Do they? (https://open-edx-proposals.readthedocs.io/en/latest/oep-0032-arch-unique-identifier-for-users.html?highlight=lms%20user%20id#user-identifier-candidate-comparison)

@edx-abolger 
We're using enterprise_customer_user here not the `user` table. Confusing I know, but enterprise_customer_user does have a uuid field.

edx-abolger:
Oh yes, I can definitely add the UUID for this set of logs.

binodpant (commented):
sorry for late review, I forgot to click submit review!

integrated_channels/moodle/exporters/learner_data.py
```diff
@@ -38,8 +38,9 @@ def get_learner_data_records(
 
         if enterprise_customer.user_email is None:
             LOGGER.debug(
-                'No learner data was sent for user [%s] because a Moodle user ID could not be found.',
-                enterprise_customer.username
+                'No learner data sent for LMS User [%s] because a Canvas user ID was not found for customer [%s]',
```
binodpant, line 41: should be Moodle instead of Canvas?

integrated_channels/sap_success_factors/exporters/learner_data.py
```diff
@@ -131,12 +149,8 @@ def unlink_learners(self):
         if not sap_inactive_learners:
             return None
 
-        providers = enterprise_customer.identity_providers
-        if not enterprise_customer.has_identity_providers:
-            LOGGER.info(
-                'Enterprise customer [%s] has no associated identity provider',
-                enterprise_customer.name
-            )
+        providers = self._get_identity_providers()
```
binodpant, line 152: 🎉

binodpant (commented):
adding some comments..

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -496,6 +430,7 @@ def get_learner_assessment_data_records(
     ):
         """
         Generate a learner assessment data transmission audit with fields properly filled in.
+        Returns a list of LearnerDataTransmissionAudit objects.
```
alex-sheehan-edx, line 433: Do you think it's worth adding params here as you did above a few times?

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -557,7 +492,8 @@ def _collect_certificate_data(self, enterprise_enrollment):
 
         Args:
             enterprise_enrollment (EnterpriseCourseEnrollment): the enterprise enrollment record for which we need to
-            collect completion/grade data
+            collect completion/grade data,
+            channel_name: labeled for relevant integrated channel this is being called for to enhance logging.
```
alex-sheehan-edx, line 496: When do we want `channel_name` to be a part of the doc string? I noticed above that it wasn't included.

integrated_channels/integrated_channel/tasks.py
```diff
@@ -185,48 +208,40 @@ def transmit_single_subsection_learner_data(username, course_run_id, subsection_
...
 @shared_task
 @set_code_owner_attribute
-def transmit_subsection_learner_data(username, channel_code, channel_pk):
+def transmit_subsection_learner_data(job_username, channel_code, channel_pk):
     """
     Task to send assessment level learner data to a linked integrated channel.
 
     Arguments:
-        username (str): The username of the User to be used for making API requests for learner data.
+        job_username (str): The username of the User making API requests for learner data.
```
alex-sheehan-edx, line 232: That's a great clarification 👍 has caused me headaches even being familiar with the processes

test_utils/__init__.py
```diff
@@ -407,3 +407,26 @@ class EmptyCacheMixin:
     def setUp(self):
         super().setUp()
         caches['default'].clear()
+
+
+class ReturnValueSpy:
```
alex-sheehan-edx, line 412: I ❤️ this

alex-sheehan-edx (commented):
Couple of nits and questions but overall this looks great and I'm excited to have to work less hard to dive into all the incredibly frustrating splunk logs 😄

integrated_channels/integrated_channel/exporters/learner_data.py
edx-abolger: Yes, will do.

integrated_channels/integrated_channel/exporters/learner_data.py
edx-abolger: I updated params when I added them- existing params were left alone.

integrated_channels/integrated_channel/tasks.py
edx-abolger: reverting

integrated_channels/integrated_channel/tasks.py
edx-abolger: My rule is to NEVER use 'user' alone: roles are better: 'learner_user', 'admin_user', 'api_user' etc

integrated_channels/blackboard/exporters/learner_data.py
```diff
@@ -33,11 +33,15 @@ def get_learner_data_records(
         """
         enterprise_customer_user = enterprise_enrollment.enterprise_customer_user
         if enterprise_customer_user.user_email is None:
-            LOGGER.debug(
-                'No learner data was sent for LMS User [%s] because Blackboard User ID not found for customer [%s]',
+            LOGGER.debug(generate_formatted_log(
+                'blackboard',
+                enterprise_customer_user.enterprise_customer.uuid,
```
taliaedX, line 38: so when I asked for uuid I meant for the user, not the enterprise.
The key thing I am trying to solve for is getting a ticket from ECS that says "some user (probably their email address) is having some issue" and being able to track them down with *some* identifier. The reason I said ID was pointless was because if I try a search in Splunk that is like `error 6` I will get a lot, A LOT of logs that have nothing to do with a *user* id of 6. Hell I'll get logs that don't even have to do with any ID and just happen to have a 6 in them.
Thus this is why I was trying to insinuate using uuid *instead* of id, which is functionally useless, but was probably not clear enough about that.

So this line would be `enterprise_customer_user.uuid`

taliaedX (changes requested):
We're still not getting the right value to hunt down a user when we inevitably get a ticket from ECS to do exactly that.

integrated_channels/blackboard/exporters/learner_data.py
taliaedX: I think given our high user ids for LMS at this stage, this is probably not actually an issue. And I forgot to mention I do see value in having the enterprise's actual uuid. I'm not 100% sure if we use LMS user id everywhere (we seem to use it almost everywhere), so I'm fine with standardizing on that. In general I'd prefer us using uuids but the direct link from enterprise customer user to LMS is the LMS user id, and I don't know that we'll ever get a ticket about a user that actually has an id less than 100,000

integrated_channels/blackboard/exporters/learner_data.py
```diff
@@ -79,11 +83,15 @@ def get_learner_assessment_data_records(
...
+            LOGGER.debug(generate_formatted_log(
+                'blackboard',
+                enterprise_enrollment.enterprise_customer_user.enterprise_customer.uuid,
                 enterprise_enrollment.enterprise_customer_user.user_id,
-                enterprise_enrollment.enterprise_customer_user.enterprise_customer.name
-            )
+                None,
+                ('get_learner_assessment_data_records finished. No learner data was sent because '
+                'Blackboard User ID not found for [{name}]'.format(
+                    name=enterprise_enrollment.enterprise_customer_user.enterprise_customer.name
```
taliaedX, line 93: Isn't this if block referring to the individual user?
`enterprise_enrollment.enterprise_customer_user.user_email`
Thus shouldn't this be like `Blackboard User ID not found for [lms user id, so we aren't using a user's real name?]` It definitely cancels out of the learner assessment audit altogether, but I _think_ it's just the block for a specific user still?

taliaedX: I know the generated formatted log standardizes including the user id, but just reading this log is confusing, since it implies the entire enterprise's learner data fails because _the enterprise_ doesn't have a user id in blackboard.

integrated_channels/blackboard/exporters/learner_data.py
edx-abolger: EnterpriseCustomerUser does not currently have a UUID field.  The table is only a link between an EnterpriseCustomer (which has a UUID) and the LMS User id, for reasons in the OEP I linked before. I think that it is reasonable to bring up adding a UUID specifically to EnterpriseCustomerUser, maybe as a discovery or Enterprise Arch ticket, but it's not part of this logging ticket. 

Since we don't have a UUID for a user when searching logs, and you might have to filter out a lot of unrelated errors, the other solution I propose that is in scope is I could change the text in the formatter to be less generic to make false positives unlikely:  

Right now the formatter logs `[User: {lms_user_id}]`, so a typical debug process would have to be (1) lookup the LMS user id in the Read Replica by email, then (2) in Splunk search for the text `"Integrated Channel" [User: <value>]`. 

I could change the `[User: <value>]` to be a single, fuller identifier like `[integrated_channel_lms_user: <value>]` for fewer other potential log hits?

taliaedX: Ack. You're right - what the heck table was I looking at earlier. I swore I saw a uuid on ECU. Sorry about that!

I think my initial concern was overlap in ids of varying kinds and being able to track a single metric/identifier across different services. This usually involves some unique token attributed to an entire process.
I mostly just want edX to be consistent in that as much as we can be (I know for instance, with pending enterprise users they may or may not have a user id and we *have* to use email addresses).

To your suggestion, that definitely allows for the ability to structure a search to isolate integrated channel logs by user, however, it still requires knowing that entire log prefix ahead of time.
My ideal is being able to take a user id/uuid, drop it into Splunk and get ONLY logs related to that user's interaction and processes in all of edX. If I have a user id, but not entirely sure what I'm looking for is integrated channels related, I will have to sift through results to get an example of that log output (or find this code).

Most likely, the people doing the digging are very familiar with integrated channels and probably already use a search term of `integrated_channels` (I know I sure do!) and our LMS user ids are unique enough at this stage that I can (and have) just plopped one into a Splunk search with no other search terms and began isolating down what I'm looking for, so this is all not actually a big deal at all and I feel non-blocking at this stage. It [standardized logging across systems] is also an initiative that needs to be driven at a top down level (likely by Arch)

taliaedX: After writing a different comment, I think your suggestion is a good one just in the fact we can then facet that specific "field" and be able to do something like `integrated_channel_lms_user=<id> "specific error we're hunting down"` in Splunk. We could also then filter just by the user id and see any other integrated channel logs that may or may not be occurring with that user.

integrated_channels/blackboard/exporters/learner_data.py
taliaedX: So I think bulk assessment level jobs would also hit this same log line and I'm not entirely sure it would cancel out of the entire job or just return nothing for this single user. So I'm still leaning into rewording this line so it doesn't insinuate the entire job fails, ie maybe: `No learner data was sent *for this user* because a Blackboard User ID could not be found for them.` Since we prefix user id and Customer Name, we probably don't need the log line itself to contain those contents as well. I'm also realizing if we want an easy way to determine if certain errors are happening at high rates, we can actually track those numbers better with a standardized log message. (ie we can use the full error message as a search term, instead of being able to only pull part of it or have to use regex to parse out the dynamic user ids, names, etc.)

taliaedX (approved):
I don't actually have anything blocking at this point.
May defer to Binod though.

PR #1252, opened 2021-06-01
Fix: ENT-4545 | updated time objects to be timezone aware in email_drip_for_missing_dsc_records

HammadAhmadWaqas:
We were using time zone aware time in the test but not in command that could have caused flakiness.

### DIFF
--- a/enterprise/management/commands/email_drip_for_missing_dsc_records.py
+++ b/enterprise/management/commands/email_drip_for_missing_dsc_records.py
@@ -4,12 +4,13 @@ Django management command for sending an email to learners with missing DataShar
 """
 import datetime
 import logging
-from datetime import date, timedelta
+from datetime import timedelta
 from urllib.parse import urljoin
 
 from django.conf import settings
 from django.core.management import BaseCommand
 from django.urls import reverse
+from django.utils import timezone
 
 from consent.models import DataSharingConsent, ProxyDataSharingConsent
 from enterprise import utils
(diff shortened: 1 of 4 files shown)


enterprise/management/commands/email_drip_for_missing_dsc_records.py
```diff
@@ -88,7 +89,7 @@ def get_enterprise_course_enrollments(self, options):
                 enterprise_customer_user__enterprise_customer__enable_data_sharing_consent=True
             )
         else:
-            past_date = date.today() - timedelta(days=past_num_days)
+            past_date = timezone.now().date() - timedelta(days=past_num_days)
```
muhammad-ammar, line 92: I would recommend to not change the management command. Instead update the test only.

tests/test_enterprise/management/test_email_drip_for_missing_dsc_records.py
```diff
@@ -71,14 +71,16 @@ def create_enrollments(self, num_learners, enrollment_date):
                 enterprise_customer_user=enterprise_customer_user,
                 course_id=course_id,
             )
-            enterprise_course_enrollment.created = enrollment_date
+            enterprise_course_enrollment.created = enrollment_time
             enterprise_course_enrollment.save()
 
     def setUp(self):
         super().setUp()
-        today = timezone.now().date()
```
HammadAhmadWaqas, line 79: This should be a whole DateTime object instead of just Date as we save this value in `created` field.

tests/test_enterprise/management/test_email_drip_for_missing_dsc_records.py
```diff
@@ -71,14 +71,16 @@ def create_enrollments(self, num_learners, enrollment_date):
...
-        self.create_enrollments(num_learners=3, enrollment_date=today - timedelta(days=1))
-        self.create_enrollments(num_learners=5, enrollment_date=today - timedelta(days=10))
+        now = timezone.now()
+        # creating enrollments for yesterday.
```
HammadAhmadWaqas, line 80: added comments for better readability

irfanuddinahmad (approved):

muhammad-ammar (approved):
LGTM

saleem-latif (approved):

PR #1257, opened 2021-06-03
feat: Use updated template model for bulk enrollment ENT-4322

binodpant:
This approach is basically to add a`template_type` column to the model holding email templates.
Also removes use of file based templates to single source templates from db.

It allows us to single source templates in one place - the database

Also a data migration added to populate default template for enterprise_customer=None, for both template_type values (SELF_ENROLL and ADMIN_ENROLL). The latter is used by bulk enrollment emails. The former template is used by by any other calls to the `notify_learners()` function that does NOT specify `admin_enrollment=True`.

__Basic approach:__

We now fetch the model with type='ADMIN_ENROLL' if bulk enroll case is specified, otherwise we default to the 'SELF_ENROLL' (or default) template (templates fetched from the database table for `EnrollEmailNotificationTemplate` model).

In case template override per customer is not found in the db for the given template_type, we use the fallback template from the db where enterprise_customer=None and type=Desired_Type 

The `admin_enrollment=True` switch passed to `notify_learners` (with default value of False) instructs the code to pick up the admin enrollment template if it exists in db, or from the admin enrollment template db records, as a fallback.

Also we log warning when we cannot find a template:

e.g.,
```
WARNING  enterprise.utils:utils.py:497 Cannot find email templates for EdX, template_type: DEFAULT_ENROLL. Not sending notification email.
```

Tests added to cover template finding logic

### Manual testing notes

_with no templates at all_: I only get a logging message as above, and no email attempts made according to email log file

_with only learner initiated default enroll template but no admin default template, still no per enterprise templates_: again, for bulk enroll case, no attempt to email was made, only logging seen. But if I remove the `admin_enrollment=true` to switch modes, I now see it picking up the self-enroll template instead

_with default templates for both learner and admin initiated templates_: for bulk enroll, the correct template was chosen

in all cases, checked the /edx/src/ace_messages in the lms container.

__Email looks like this__

Screen Shot 2021-06-08 at 10 19 18 PM

My local django admin vie with three templates (two from migration default):

Screen Shot 2021-06-09 at 8 27 39 AM

ENT-4322

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -2235,14 +2237,27 @@ class Meta:
...
+    template_type_choices = [
+        (DEFAULT_TEMPLATE_TYPE, 'Default Enrollment Template'),
+        (BULK_ENROLL_TEMPLATE_TYPE, 'Bulk Enrollment Template'),
+    ]
+
     plaintext_template = models.TextField(blank=True, help_text=BODY_HELP_TEXT)
     html_template = models.TextField(blank=True, help_text=BODY_HELP_TEXT)
     subject_line = models.CharField(max_length=100, blank=True, help_text=SUBJECT_HELP_TEXT)
-    enterprise_customer = models.OneToOneField(
+    enterprise_customer = models.ForeignKey(
```
binodpant, line 2250: in contrast to before, now we can have multiple templates per customer with type specified.

enterprise/templates/enterprise/emails/bulk_enroll_notification.html
```diff
@@ -0,0 +1,371 @@
+<html>
```
binodpant, line 1: this template may need some final tuneup but it's quite close to the look we want

enterprise/models.py
```diff
@@ -652,14 +652,15 @@ def clear_pending_registration(self, email, *course_ids):
         else:
             PendingEnrollment.objects.filter(user=pending_ecu, course_id__in=course_ids).delete()
 
-    def notify_enrolled_learners(self, catalog_api_user, course_id, users):
+    def notify_enrolled_learners(self, catalog_api_user, course_id, users, bulk_enrollment=False):
```
edx-abolger, line 655: Suggestion: I would change the names in several places in this PR. I know we call the Quokkas feature 'bulk enrollment', but the reason we have this new template isn't "single vs bulk",  it's "Self Enrolled" or "Admin Enrolled". Reading the code for all of this stuff, it would be a lot clearer that the important part to this feature is that the templates change based on who did the enrolling. Thoughts?

enterprise/templates/enterprise/emails/bulk_enroll_notification.txt
```diff
@@ -0,0 +1,8 @@
+Great News! You've been enrolled in {{enrolled_in.name}} by {{organization_name}}
+
+Visit {{enrolled_in.url}} to access your course.
+
+This course is a free benefit offered especially for you, and we re excited for you to meet your learning community on edX.
```
edx-abolger, line 5: Type: missing "we re" should be "we are" or "we're"

edx-abolger (changes requested):
This seems clean to me. See 2 comments- 1 typo and one about naming everything, up to you what you do with 2nd comment.

enterprise/migrations/0132_auto_20210603_1544.py
```diff
@@ -0,0 +1,29 @@
...
+        ),
+        migrations.AddField(
+            model_name='historicalenrollmentnotificationemailtemplate',
+            name='template_type',
+            field=models.CharField(choices=[('DE', 'Default Enrollment Template'), ('BE', 'Bulk Enrollment Template')], default='DE', help_text='Use either DE (default) or BE (bulk enrollment)', max_length=2),
+        ),
+        migrations.AlterField(
+            model_name='enrollmentnotificationemailtemplate',
+            name='enterprise_customer',
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='enterprise_enrollment_templates', to='enterprise.EnterpriseCustomer'),
```
taliaedX, line 27: OH. I was originally wondering this in your first PR, but you don't need to convert from a OnetoOne to foreign key. You can in fact still make a one to one optional, which helps in my other suggestions.

enterprise/utils.py
```diff
@@ -377,6 +382,8 @@ def build_notification_message(template_context, template_configuration=None):
         template_context (dict): A set of data to render
         template_configuration: A database-backed object with templates
             stored that can be used to render a notification.
+        default_text_template: template to use for text emails, if template_configuration is not found.
+        default_html_template: template to use for html emails, if template_configuration is not found.
```
taliaedX, line 386: The more I go through this PR the more I lose my mind.

Okay, so:
1. We're using 2 storage methods for templates - file and database
2. We previously (before this PR) had the "default" as a file based template because we associated the templates ONLY with an enterprise customer.
3. You're adding _another_ file based template, which is not what I want and I think we can get rid of ALL file-based templates and just consistently use the database.
4. We can leave the OneToOne logic be and just make it nullable.

So for build_notification_message, I'd actually get rid of code, not add any, it should look like this:
```
def build_notification_message(
        template_context,
        template_configuration=None):
    return template_configuration.render_all_templates(template_context)
```
And where you figure out your template configuration, the logic should just be:
```
if bulk_enrollment:
    enterprise_template_config = enterprise_customer.enterprise_enrollment_templates.filter(
            template_type=BULK_ENROLLMENT_TEMPLATE_TYPE
    ).first()
else:
    try:
        enterprise_template_config = enterprise_customer.enterprise_enrollment_template

    except (ObjectDoesNotExist, AttributeError):
        enterprise_template_config = enterprise_customer.enterprise_enrollment_templates.filter(
            template_type=DEFAULT_TEMPLATE_TYPE
        ).first()
```
Do you see what I'm getting at with this?
The database already stores html and plaintext template and subject and all that jazz, we should just add bulk enrollment and the default to that and then figure out which template to call from the database rather than make a database call and fallback to a filesystem.

enterprise/models.py
taliaedX: This is actually not a required change. We just need to remove the requirement that it be present. (ie make it nullable)

enterprise/models.py
```diff
@@ -2235,14 +2237,27 @@ class Meta:
...
+    ]
+
     plaintext_template = models.TextField(blank=True, help_text=BODY_HELP_TEXT)
     html_template = models.TextField(blank=True, help_text=BODY_HELP_TEXT)
     subject_line = models.CharField(max_length=100, blank=True, help_text=SUBJECT_HELP_TEXT)
-    enterprise_customer = models.OneToOneField(
+    enterprise_customer = models.ForeignKey(
         EnterpriseCustomer,
-        related_name="enterprise_enrollment_template",
+        related_name="enterprise_enrollment_templates",
```
taliaedX, line 2252: This can be left on what it was originally

enterprise/utils.py
```diff
@@ -438,7 +445,22 @@ def get_notification_subject_line(course_name, template_configuration=None):
...
 
 
-def send_email_notification_message(user, enrolled_in, enterprise_customer, email_connection=None):
+def resolve_fallback_email_templates(bulk_enrollment=False):
+    """
+    Returns text and html template locations to use when rendering email templates from file.
+    """
+    if bulk_enrollment:
+        return 'enterprise/emails/bulk_enroll_notification.txt', 'enterprise/emails/bulk_enroll_notification.html'
+    return 'enterprise/emails/user_notification.txt', 'enterprise/emails/user_notification.html'
```
taliaedX, line 454: This whole method can be tossed.

taliaedX (changes requested):
I didn't quite grasp that we already had a database solution and thus were mixing file and database templates and I'd really, really rather we didn't. We can simplify things a lot by just using the database consistently.

enterprise/models.py
binodpant: yeah this is a great point, self vs admin is the key distinction indeed. going to do this...

enterprise/models.py
```diff
@@ -2235,14 +2237,27 @@ class Meta:
         'placeholder {course_name} will be replaced with the name of the course or program that was enrolled in.'
     )
 
+    DEFAULT_TEMPLATE_TYPE = 'DE'
+    BULK_ENROLL_TEMPLATE_TYPE = 'BE'
+    template_type_choices = [
+        (DEFAULT_TEMPLATE_TYPE, 'Default Enrollment Template'),
+        (BULK_ENROLL_TEMPLATE_TYPE, 'Bulk Enrollment Template'),
+    ]
```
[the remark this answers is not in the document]
binodpant: sure, yeah this is a good suggestion, will change the names from DE and BE to DefaultEnrollment and AdminEnrollment (going to rename bulk enrollment to admin enrollment soon per Adrienne's suggestion)

enterprise/models.py
```diff
@@ -2235,14 +2237,27 @@ class Meta:
...
     subject_line = models.CharField(max_length=100, blank=True, help_text=SUBJECT_HELP_TEXT)
-    enterprise_customer = models.OneToOneField(
+    enterprise_customer = models.ForeignKey(
         EnterpriseCustomer,
-        related_name="enterprise_enrollment_template",
+        related_name="enterprise_enrollment_templates",
         on_delete=models.deletion.CASCADE
     )
+    template_type = models.CharField(
+        max_length=2,
```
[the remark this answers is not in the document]
binodpant: yeah ok with the upcoming change we will want a larger value for sure, will do 255 in case we add other values in the future.

enterprise/utils.py
```diff
@@ -481,11 +503,24 @@ def send_email_notification_message(user, enrolled_in, enterprise_customer, emai
         'organization_name': enterprise_customer.name,
     }
     try:
-        enterprise_template_config = enterprise_customer.enterprise_enrollment_template
+        if bulk_enrollment:
+            template_type = 'BE'
+        else:
+            template_type = 'DE'
```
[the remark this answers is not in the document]
binodpant: yeah good catch! need to use those

enterprise/utils.py
binodpant: hmm I would love that too actually, except there is n issue we will hit. The issue will be that there will have to be a template model instance for every enterprise customer, and then have a way to fall back if that does not exist. If we remove the file based template, we will then need to furnish two things: an alternative fallback method (which would be another database record somewhere?) .

not sure if I clearly explained but mainly the issue is that there is only one record in prod in this table, so almost all enterprises are using the 'fallback' from files as of now.  So it's almost as if we never meant to use the db models haha... 

Now, if there was a fallback template also in the database, then I think we can make it work all database based and eliminate the file based system. Would that be a route we are interested in? Even if we do, I think we may want to do that separately not part of this PR.. thoughts?

binodpant: in other words, if we were to redo this from scratch, I think we will need: (1) a default fallback template for both types in db, and (2) a per enterprise template like today. No?

enterprise/models.py
binodpant: hmm, so if we store two different templates for a given customer, with different types, can we still use onetoone with a nullable value, or do we need the foreign key/ manyToOne method?

taliaedX: Great point! If we see the possibility of like customer customized bulk enrollment templates, then yeah it should actually be a many to one (the many being templates, the one being customer)

taliaedX: The current structure seems to be default, bulk, *or* customer, so in this structure customer would be a type and there wouldn't be a many to one. If we do see having a type of bulk with a customer specification then we need to figure out what type our existing customer templates are - default? Makes the use of "default" a bit odd no?

enterprise/models.py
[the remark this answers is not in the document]
taliaedX: yeah I don't think it necessarily has to be 255. It can be whatever. I appreciate the consideration of the actual data length and I don't see us writing a paragraph for a label (I will veto that anyways), so I think you can stick with your gut to limit it, I just wouldn't go with 2.

enterprise/utils.py
taliaedX: Right so (and this is mentioned in another comment) our current logic even before bulk enrollment, is default OR customer. If bulk enrollment is just another template (which is sort of how it was being treated, your code actually breaks existing customer templates, so we need to fix that either way), then that's where my logic comes into play. It uses bulk enrollment templates first and foremost when that is specified, if not, it tries to fetch a customer template and then will fall back to the default.

In the case of us having a bulk enrollment template customized for a customer then yes this logic needs to change a little, but it still roughly follows the above.

binodpant: Ah so I see where we have a disconnect, there is a dimension I did not clarify before, which I think may clarify this better?

The thing is that we will still need to support a template per ' enrollment type'. For example:

- if bulk enrollment: use template default OR customer
- else regular enrollment: use template default OR customer

So each enrollment type will have a default template, and also a possible customer override (which I am honestly thinking is not going to be used much).

Given that, does the proposed setup then make sense?

enterprise/models.py
binodpant: yup exactly, we are setting up the bulk enroll templates the same way as the other templates, default in files, then per customer override (if desired) in the table

enterprise/models.py
binodpant: done!

enterprise/models.py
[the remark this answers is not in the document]
binodpant: >  I don't see us writing a paragraph for a label

meaning help_text or?

binodpant:
made a couple feedback changes, can follow up on Talia's thoughts about the setup tomorrow, also need test fixes but yeah looking closer now!

enterprise/utils.py
binodpant: also my logic see below, basically does what you mentioned, but later it still uses file based fallback if this results in None.

```
try:
        if admin_enrollment:
            template_type = ADMIN_ENROLL_EMAIL_TEMPLATE_TYPE
        else:
            template_type = DEFAULT_ENROLL_EMAIL_TEMPLATE_TYPE
        # we only support one template per type
        enterprise_template_config = enterprise_customer.enterprise_enrollment_templates.filter(
            template_type=template_type
        ).first()
    except (ObjectDoesNotExist, AttributeError):
        enterprise_template_config = None
```


I think really the only moving part here is whether we use file based templates at all. Replacing that with a db based template is possible but it requires another refactor plus a new model, we cannot fit it into the current model which is designed for per customer overrides only.

edx-abolger: @binodpant am I correct that the full "proposed setup" after this conversation is: 
 1.  putting both defaults in the DB
      1a. Write a migration that put the defaults in the new DB immediately on deploy of this change. 
 2. Remove file-based templates completely
 3. Keeping the existing customer overrides for a learner-initiated enrollment (which we already have but we don't think is used much), and apply those under condition "learner-iniated enrollment AND customer has _default type_ template" 
 4. (this is the one I am muddiest on from the above conversation: do we need this or do all bulk-enrolled learners just get 1 email? ) Add capability to store new customer templates for the bulk-enrollment initiated enrollments. Apply this template under condition "bulk/admin-iniated enrollment AND customer has _bulk/admin type_ template"

binodpant: so resolving this, since we are now trying completely database based templates. no more files.

tests upcoming but ...

enterprise/api/v1/views.py
```diff
@@ -318,6 +318,7 @@ def enroll_learners_in_courses(self, request, pk):
                     catalog_api_user=request.user,
                     course_id=course_run,
                     users=pending_users | existing_users,
+                    admin_enrollment=True,
```
binodpant, line 321: all other call sites still just default to `admin_enrollment=False`

enterprise/utils.py
```diff
@@ -365,39 +369,6 @@ def get_catalog_admin_url_template(mode='change'):
     return None
 
 
-def build_notification_message(template_context, template_configuration=None):
```
binodpant, line 368: this function boils down to one line, so I removed it.. and we just now call the template_config method in the calling site

enterprise/utils.py
```diff
@@ -519,6 +534,13 @@ def enterprise_enrollment_source_model():
     return apps.get_model('enterprise', 'EnterpriseEnrollmentSource')  # pylint: disable=invalid-name
 
 
+def enroll_notification_email_template():
+    """
+    Returns the ``EnrollmentNotificationEmailTemplate`` class.
+    """
+    return apps.get_model('enterprise', 'EnrollmentNotificationEmailTemplate')  # pylint: disable=invalid-name
```
taliaedX, line 541: what's the reason behind a separate method that just fetches a model?

enterprise/utils.py
```diff
@@ -438,7 +409,40 @@ def get_notification_subject_line(course_name, template_configuration=None):
...
+    except (ObjectDoesNotExist, AttributeError):
+        enterprise_template_config = None
+
+    if not enterprise_template_config:
+        # use the fallback template instead
+        template_queryset = enrollment_template.objects.filter(
+            enterprise_customer=None,
+            template_type=template_type,
+        )
+        enterprise_template_config = template_queryset.first()
```
taliaedX, line 434: Can't you just attach `[0]` or `first()` on the previous equation instead of having to make 2 separate variables?

taliaedX (commented):
this looks pretty good. I'm still not certain on the naming choice of one of the template types.

enterprise/utils.py
binodpant: guess I just blindly followed the examples I saw for the other models. I changed it to call `apps.get_model('enterprise', 'EnrollmentNotificationEmailTemplate')` directly for now as esp. there is only one call site for it anyway

enterprise/models.py
```diff
@@ -2235,13 +2240,28 @@ class Meta:
         'placeholder {course_name} will be replaced with the name of the course or program that was enrolled in.'
     )
 
+    template_type_choices = [
+        (DEFAULT_ENROLL_EMAIL_TEMPLATE_TYPE, 'Default Enrollment Template'),
```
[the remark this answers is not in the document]
binodpant: ah yeah good call. I forgot to rename these. The default name does not really make sense here.. we just happen to use it as default in the code flows here. renamed to 'SELF_' instead of 'default_', and strings accordingly too

tests/test_utilities.py
```diff
@@ -572,180 +561,6 @@ def test_send_email_notification_message_with_site_defined_values(
                 assert getattr(mail.outbox[0], field) == val
             assert mail.outbox[0].connection is conn
 
-    @ddt.data(
```
binodpant, line 575: I am removing this test since I feel like it's no longer necessary to test against a template that is no longer in file form in git, we are doing templates in the db, so whoever changes those templates will need to test them out against the version of code being tested. It would be hard to keep the test and the database templates in sync, and we are going to go Braze so I don't think losing this test case for this time, is going to lose us anything.

If you feel strongly we need this test back, I am going to have to extra work to bring the template into the factory somehow, I used simpler templates in my factory to only cover some essential context variables

binodpant:
Have to follow up with someone but for whatever weird reason, many tests depend on the presence of the email template file, https://github.com/edx/edx-enterprise/blob/1f898a9d1b3232daf6a309262cd4dfaccb4ac1ef/test_utils/__init__.py don't want to follow that rabbithole quite for this PR but will ask the author(s) on slack

enterprise/migrations/0132_auto_20210604_1556.py
```diff
@@ -0,0 +1,29 @@
...
+            model_name='historicalenrollmentnotificationemailtemplate',
+            name='template_type',
+            field=models.CharField(choices=[('DEFAULT_ENROLL', 'Default Enrollment Template'), ('ADMIN_ENROLL', 'Admin Enrollment Template')], default='DEFAULT_ENROLL', help_text='Use either DEFAULT_ENROLL or ADMIN_ENROLL', max_length=255),
+        ),
+        migrations.AlterField(
+            model_name='enrollmentnotificationemailtemplate',
+            name='enterprise_customer',
+            field=models.OneToOneField(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='enterprise_enrollment_template', to='enterprise.EnterpriseCustomer'),
+        ),
+    ]
```
edx-abolger, line 29: @binodpant were you not going to add a migration that populated default templates?

enterprise/models.py
```diff
@@ -2275,9 +2295,12 @@ def __str__(self):
         """
         Return human-readable string representation.
         """
-        return '<EnrollmentNotificationEmailTemplate for EnterpriseCustomer with UUID {}>'.format(
-            self.enterprise_customer.uuid
-        )
+        if self.enterprise_customer:
```
edx-abolger, line 2298: If this is the string notification of the model object, please add the object's .id field so when you see this in logs you can actually go look at a database object.

enterprise/templates/enterprise/emails/user_notification.html
```diff
@@ -1,15 +1,25 @@
 {% load i18n %}<html>
+
```
edx-abolger, line 2: Did you mean to change this file? These file changes appear to be why your tests are all failing.

enterprise/utils.py
```diff
@@ -438,7 +408,43 @@ def get_notification_subject_line(course_name, template_configuration=None):
         return stock_subject_template.format(course_name=course_name)
 
 
-def send_email_notification_message(user, enrolled_in, enterprise_customer, email_connection=None):
+def find_enroll_email_template(enterprise_customer, template_type):
+    """
+    Find email template from the template database represented by EnrollmentNotificationEmailTemplate model.
```
edx-abolger, line 413: Can you add a docstring note for what 'template_type' is, and should it have a default kwarg?

enterprise/migrations/0132_auto_20210604_1556.py
taliaedX: Oh yeah I thought "default" in my brain for some reason was filling this in already, but it absolutely does not.
Yeah we need a migration for the bulk enrollment and default template that was file based

enterprise/templates/enterprise/emails/user_notification.html
binodpant: oh no, I had meant to restore this file but somehow have ended up changing it, this was just attempt to check if the test failures get fixed by restoring the file.. probably should try restoring it correctly first

enterprise/migrations/0132_auto_20210604_1556.py
binodpant: ah yes that is needed of course! otherwise we will have empty templates..

enterprise/utils.py
binodpant: added docstring! I decided it's best to make the args required so in fact added raise's to check inputs. Since for this function passing in both is the only correct way to use it.

enterprise/models.py
binodpant: added! look like these in adjango admin 

```
<EnrollmentNotificationEmailTemplate (id: 5) Default template for type DEFAULT_ENROLL>
<EnrollmentNotificationEmailTemplate (id: 6) for EnterpriseCustomer with UUID 873fec72-5a21-4f95-ac19-e3d67f9c4777>
```

tests/test_utilities.py
```diff
@@ -576,8 +563,9 @@ def test_send_email_notification_message_with_site_defined_values(
         """
         enrolled_in['start'] = datetime.datetime.strptime(enrolled_in['start'], '%Y-%m-%d')
         enterprise_customer = EnterpriseCustomerFactory(name=enterprise_customer_name)
+        # since migration alread sets up fallback templates, use per customer template
         EnrollmentNotificationEmailTemplateFactory(
-            enterprise_customer=None,
+            enterprise_customer=enterprise_customer,
```
taliaedX, line 568: can we set one of these tests up (or add a test) that doesn't just rely on the default?
Like I see for the fallbacks test we have logic using the Admin (well at least creating an Admin template too):
```
EnrollmentNotificationEmailTemplateFactory(            enterprise_customer=None,            template_type=SELF_ENROLL_EMAIL_TEMPLATE_TYPE        )        

EnrollmentNotificationEmailTemplateFactory(            enterprise_customer=None,            template_type=ADMIN_ENROLL_EMAIL_TEMPLATE_TYPE        )
```
But I don't really see us checking for Admin templates and since that one requires actually passing in a value I'm more curious about it than the default. Maybe that's in a different commit than the range I'm looking at this exact moment, so if we do have that already, feel free to ignore me!

enterprise/migrations/0133_auto_20210608_1931.py
```diff
@@ -0,0 +1,326 @@
...
+                    </td>
+                    </tr>
+                </tbody>
+            </table>
+        </body>
+        </html>
+        """,
+        subject_line='',
+        enterprise_customer=None,
+        template_type='ADMIN_ENROLL',
```
taliaedX, line 314: Have you by chance had the ability to actually see the rendered email from this? I'm just curious if this hideous monstrosity actually does save and render properly (I know it's not your doing for why it is hideous, that's just what it has to look like. I'm just wondering if it does render properly to and from the database). If you haven't been able to test, let's just double-check that on stage before we deploy this to prod when we get there. 👍

tests/test_utilities.py
```diff
@@ -641,7 +629,6 @@ def test_send_email_notification_message_with_site_from_email_override(
             )
 
         enterprise_customer = EnterpriseCustomerFactory(site=site)
-        EnrollmentNotificationEmailTemplateFactory(enterprise_customer=None)
```
taliaedX: I'm curious how we can just not make an email template here now and the test still work?

enterprise/migrations/0133_auto_20210608_1931.py
binodpant: ah yes sorry sharing image in description section

tests/test_utilities.py
binodpant: Ah due to migration I think default self template already exists and is used

looks like this in test

```
(Pdb++) enterprise_template_config
<EnrollmentNotificationEmailTemplate (id: 1) Default template for type SELF_ENROLL>
```

enterprise/migrations/0133_auto_20210608_1931.py
binodpant: I do expect to have to make changes to the template btw so this may not be final, I am certain Jen or Joe are going to have feedback. I also need to fix that My Dashboard style but I wanted to see if I can do that from django admin once this is live...

tests/test_utilities.py
binodpant: Test added for per customer admin template!

tests/test_utilities.py
taliaedX: ahh right. I forgot tests here run the migrations!

taliaedX (approved):
Assuming you get tests to pass this looks good to me!

binodpant:
Ah the latest failure is a known sporadic test issue 

```
FAILED tests/test_enterprise/management/test_email_drip_for_missing_dsc_records.py::EmailDripForMissingDscRecordsCommandTests::test_email_drip_for_missing_dsc_records
```

rerunning job

edx-abolger (approved):

PR #1258, opened 2021-06-03
feat: log more specific information about HTTP client errors during enrollment

iloveagent57:
Log more specific information about HTTP client errors that are caught when using the LMS
enrollment API.  Also send an exception event to the monitoring service when this happens, even
though we handle the exception "gracefully".

https://openedx.atlassian.net/browse/ENT-4455

The current logged exception message is
```
 Unable to create an LMS enrollment from the DSC view: 
Client Error 400: https://courses.edx.org/api/enrollment/v1/enrollment
```
...which is not helpful.  In the DSC `get()` handler, we do log some info about licensed enrollments failing (when catching the exception raised from `_enroll_learner_in_course()`, but that's only in the case that the enterprise customer has DSC disabled, and it doesn't tell us about _why_ the enrollment API objects to the client request.  This change makes it so that both the `get()` and `post()` handler of the DSC view will know _why_ enrollments fail.

With this change, we'll now get errors like this in our logs:
```
2021-06-04 14:10:01,343 ERROR 1365 [enterprise.views] [user 2] [ip 172.21.0.1] views.py:620 - Client Error while enrolling user edx in course-v1:edX+DemoX+Demo_Course via enrollment API: The [verified] course mode is expired or otherwise unavailable for course run [course-v1:edX+DemoX+Demo_Course].
  File "/edx/src/edx-enterprise/enterprise/views.py", line 597, in _enroll_learner_in_course
    enrollment_api_client.enroll_user_in_course(
  File "/edx/src/edx-enterprise/enterprise/api_client/lms.py", line 104, in inner
    return func(self, *args, **kwargs)
  File "/edx/src/edx-enterprise/enterprise/api_client/lms.py", line 229, in enroll_user_in_course
    return self.client.enrollment.post(
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.8/site-packages/slumber/__init__.py", line 167, in post
    resp = self._request("POST", data=data, files=files, params=kwargs)
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.8/site-packages/slumber/__init__.py", line 101, in _request
    raise exception_class("Client Error %s: %s" % (resp.status_code, url), response=resp, content=resp.content)
slumber.exceptions.HttpClientError: Client Error 400: http://edx.devstack.lms:18000/api/enrollment/v1/enrollment
```
and we'll see errors accumulate in New Relic.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -12,6 +12,7 @@ from uuid import UUID
 import pytz
 import waffle  # pylint: disable=invalid-django-waffle-import
 from dateutil.parser import parse
+from edx_django_utils import monitoring
 from edx_rest_api_client.exceptions import HttpClientError
 from ipware.ip import get_client_ip
 from opaque_keys import InvalidKeyError
@@ -606,11 +607,21 @@ class GrantDataSharingPermissions(View):
                                 course_mode=course_mode
                             )
                         )
-                    except Exception as exc:  # pylint: disable=broad-except
-                        LOGGER.error(
-                            'Unable to create an LMS enrollment from the DSC view: {exc}'.format(
-                                exc=exc
-                            )
+                    except HttpClientError as exc:
+                        monitoring.record_exception()
+
+                        default_message = 'No error message provided'
+                        try:
+                            error_message = json.loads(exc.content.decode()).get('message', default_message)
+                        except ValueError:
+                            error_message = default_message
+                        LOGGER.exception(
+                            'Client Error while enrolling user %(user)s in %(course)s via enrollment API: %(message)s',
+                            {
+                                'user': request.user.username,
+                                'course': course_id,
+                                'message': error_message,
+                            }
                         )
                         raise
             try:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,12 @@ Change Log
 Unreleased
 ----------
 
+[3.23.11]
+---------
+* Log more specific information about HTTP client errors that are caught when using the LMS
+  enrollment API.  Also send an exception event to the monitoring service when this happens, even
+  though we handle the exception "gracefully".
+
 [3.23.10]
 ---------
 * Send long dsc url in missing DSC email as individual params.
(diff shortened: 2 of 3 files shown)


enterprise/views.py
```diff
@@ -606,11 +607,21 @@ def _enroll_learner_in_course(
                                 course_mode=course_mode
                             )
                         )
-                    except Exception as exc:  # pylint: disable=broad-except
-                        LOGGER.error(
-                            'Unable to create an LMS enrollment from the DSC view: {exc}'.format(
-                                exc=exc
-                            )
+                    except HttpClientError as exc:
+                        monitoring.record_exception()
```
adamstankiewicz, line 611: I think this is my first time seeing `monitoring.record_exception` being used. Where does this exception get recorded?

enterprise/views.py
```diff
@@ -606,11 +607,21 @@ def _enroll_learner_in_course(
...
+
+                        default_message = 'No error message provided'
+                        try:
+                            error_message = json.loads(exc.content.decode()).get('message', default_message)
+                        except ValueError:
+                            error_message = default_message
+                        LOGGER.exception(
+                            'Client Error while enrolling user %(user)s in %(course)s via enrollment API: %(message)s',
```
adamstankiewicz, line 619: Much more information of an exception message than before 💯

adamstankiewicz (approved):

enterprise/views.py
iloveagent57: New Relic!

iloveagent57: https://github.com/edx/edx-django-utils/blob/10df5a3cfcb94842edab6de48901a57cafbfd11d/edx_django_utils/monitoring/internal/utils.py#L85-L98

>    Records a caught exception to the monitoring system.    Note: By default, only unhandled exceptions are monitored. This function    can be called to record exceptions as monitored errors, even if you handle    the exception gracefully from a user perspective.

PR #1260, opened 2021-06-09
ENT-2966 | added flag `enable_compression`

HammadAhmadWaqas:
https://openedx.atlassian.net/browse/ENT-2966

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -2368,6 +2371,13 @@ class EnterpriseCustomerReportingConfiguration(TimeStampedModel):
         verbose_name=_("Delivery Method"),
         help_text=_("The method in which the data should be sent.")
     )
+    enable_compression = models.BooleanField(
+        default=True,
+        help_text=_(
+            "Specifies whether report should be compressed. Without compression files will not be password protected "
+            "or encrypted."
```
muhammad-ammar, line 2378: I know compression is needed to set a password but how is encryption dependent on compression?

muhammad-ammar (approved):
👍 once build is green

enterprise/models.py
```diff
@@ -2325,9 +2325,12 @@ class EnterpriseCustomerReportingConfiguration(TimeStampedModel):
         (DATA_TYPE_GRADE, DATA_TYPE_GRADE),
         (DATA_TYPE_COMPLETION, DATA_TYPE_COMPLETION),
         (DATA_TYPE_COURSE_STRUCTURE, DATA_TYPE_COURSE_STRUCTURE),
-
     )
 
+    # This list will specifies the Data Types that are allowed to send without compression.
+    # All other Data Types must be compressed.
```
sameenfatima78, line 2331: Nit:
```python
# This list will specifies the Data Types that are allowed to send without compression.    
# All other Data Types must be compressed.
-> 
# Data types that are allowed to be sent without compression, all other data types must be compressed.

Comment (sameenfatima78) [reply enterprise/models.py:2331]: Also, shouldn't we handle the `invalid-name` warning?

File: enterprise/models.py (hunk 1 above)
Comment (sameenfatima78) [reply enterprise/models.py:2378]: shouldn't this change in model have a migration? I can't seem to find any migration file in the PR

File: enterprise/models.py [hunk 3]
```diff
@@ -2579,6 +2590,13 @@ def clean(self):
                     'Decrypted SFTP password must be set if the delivery method is SFTP.'
                 )
 
+        # Check compression is enabled for compression only Data Types.
+        if not self.enable_compression and self.data_type not in self.ALLOWED_NON_COMPRESSION_DATA_TYPES:
```
sameenfatima78, line 2594: Maybe we can add parenthesis b/w conditions to add more readability

sameenfatima78 (changes requested):
Left some comments, all good but migration file is missing

enterprise/models.py
HammadAhmadWaqas: @sameenfatima78 actually its name is too long. the valid name should be less than 30 chars.

I am unable to think of any name that is less than 30 chars any, meaningful.
you can suggest any if you have any.

HammadAhmadWaqas:
@sameenfatima78 could you please re-reveiw.

sameenfatima78 (approved):
LGTM!

PR #1262, opened 2021-06-10
bug: properly handling Canvas API pagination responses

alex-sheehan-edx:
[jira](https://openedx.atlassian.net/browse/ENT-4622)

Canvas API's default to 10 objects per response with [pagination](https://canvas.instructure.com/doc/api/file.pagination.html) handled through the [Link header](https://www.w3.org/Protocols/9707-link-header.html)

Because of this, when any student completed an assignment or triggered a course completion after 10 other assignments/course completion had transmitted, the response for assignments within the course from the Canvas API would cap out at 10 and never return the 11'th assignment. Thus our client would believe the assignment did not exist and would create another duplicate. 

This PR adds pagination logic into the assignment retrieval/creation process that allows for request of additional pages of data if the assignment isn't found. 

Note: this PR will not fix the situation of already effected customers and additional work will be needed to fully rectify their situation. This will be done through manual API calls to delete the objects on the customer's instance (curl TBI)

Testing notes:
 In order to test you will need the following:
- a working local LMS and edx-enterprise
- a working Canvas customer configuration entry within LMS; this includes having an enterprise customer, learners and fully integrated Canvas instance
- a graded edX course with more than 10 assignments, integrated into the Canvas instance

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/canvas/client.py
+++ b/integrated_channels/canvas/client.py
@@ -378,37 +378,66 @@ class CanvasAPIClient(IntegratedChannelApiClient):
             course_id (str) : the Canvas course ID relating to the course which the client is currently
             transmitting learner data to.
         """
-        # First, check if the course assignment already exists
+        # Check if the course assignment already exists
         canvas_assignments_url = '{canvas_base_url}/api/v1/courses/{course_id}/assignments'.format(
             canvas_base_url=self.enterprise_configuration.canvas_base_url,
             course_id=course_id
         )
         resp = self.session.get(canvas_assignments_url)
 
-        if resp.status_code >= 400:
-            raise ClientError(
-                "Something went wrong retrieving assignments from Canvas. Got response: {}".format(
-                    resp.text,
-                ),
-                resp.status_code
-            )
+        more_pages_present = True
+        current_page_count = 0
+        assignment_id = ''
 
-        assignments_resp = resp.json()
-        assignment_id = None
-        for assignment in assignments_resp:
-            try:
-                if assignment['integration_id'] == integration_id:
-                    assignment_id = assignment['id']
-                    break
-            # The validation check above should ensure that we have a 200 response from Canvas, but sanity catch if we
-            # have a unexpected response format
-            except (KeyError, ValueError, TypeError) as error:
+        # current_page_count serves as a timeout, limiting to a max of 150 pages of requests
+        while more_pages_present and current_page_count < 150:
+            if resp.status_code >= 400:
                 raise ClientError(
                     "Something went wrong retrieving assignments from Canvas. Got response: {}".format(
                         resp.text,
                     ),
                     resp.status_code
-                ) from error
+                )
+
+            assignments_resp = resp.json()
+            for assignment in assignments_resp:
+                try:
+                    if assignment['integration_id'] == integration_id:
+                        assignment_id = assignment['id']
+                        break
+
+                # The integration ID check above should ensure that we have a 200 response from Canvas,
+                # but sanity catch if we have a unexpected response format
+                except (KeyError, ValueError, TypeError) as error:
+                    raise ClientError(
+                        "Something went wrong retrieving assignments from Canvas. Got response: {}".format(
+                            resp.text,
+                        ),
+                        resp.status_code
+                    ) from error
+
+            if not assignment_id:
+                # Canvas pagination headers come back as a string ie:
+                # 'headers': {
+                #     'Link': '<{assignment_url}?page=2&per_page=10>; rel="current",' \
+                #             '<{assignment_url}?page=1&per_page=10>; rel="prev",' \
+                #             '<{assignment_url}?page=1&per_page=10>; rel="first",' \
+                #             '<{assignment_url}?page=2&per_page=10>; rel="last"' \
+                # }
+                # so we have to parse out the linked list of assignment pages
+                assignment_page_results = resp.headers['Link'].split(',')
+                pages = {}
+                for page in assignment_page_results:
+                    page_type = page.split('; rel=')[1].strip('"')
+                    pages[page_type] = page.split(';')[0].strip('<>')
+
+                if pages.get('current') == pages.get('last', None):
+                    more_pages_present = False
+                else:
+                    resp = self.session.get(pages.get('next'))
+                    current_page_count += 1
+            else:
+                more_pages_present = False
 
         # Canvas requires a course assignment for a learner to be assigned a grade.
         # If no assignment has been made yet, create it.
--- a/tests/test_integrated_channels/test_canvas/test_client.py
+++ b/tests/test_integrated_channels/test_canvas/test_client.py
@@ -87,6 +87,10 @@ class TestCanvasApiClient(unittest.TestCase):
             refresh_token=self.refresh_token,
         )
         self.integration_id = 'course-v1:{course_id}+2T2020'.format(course_id=self.course_id)
+        self.integration_id_2 = 'course-v2:{course_id}+2T2020'.format(course_id=self.course_id)
+        self.integration_id_3 = 'course-v3:{course_id}+2T2020'.format(course_id=self.course_id)
+        self.integration_id_4 = 'course-v4:{course_id}+2T2020'.format(course_id=self.course_id)
+
         self.course_completion_date = datetime.date(
             2020,
             random.randint(1, 10),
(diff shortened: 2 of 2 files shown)


integrated_channels/canvas/client.py
```diff
@@ -378,37 +378,56 @@ def _handle_canvas_assignment_retrieval(
...
-        assignment_id = None
-        for assignment in assignments_resp:
-            try:
-                if assignment['integration_id'] == integration_id:
-                    assignment_id = assignment['id']
-                    break
-            # The validation check above should ensure that we have a 200 response from Canvas, but sanity catch if we
-            # have a unexpected response format
-            except (KeyError, ValueError, TypeError) as error:
+        while more_pages_present:
```
edx-abolger, line 391: Never leave infinite while loops in code: please add a timeout condition or a max number of tries (150 or something, for a max of 1500 assignments per user per course).

integrated_channels/canvas/client.py
```diff
@@ -378,37 +378,56 @@ def _handle_canvas_assignment_retrieval(
...
+                    raise ClientError(
+                        "Something went wrong retrieving assignments from Canvas. Got response: {}".format(
+                            resp.text,
+                        ),
+                        resp.status_code
+                    ) from error
+
+            if not assignment_id:
+                # Canvas pagination headers come back as a string,
+                # so we have to parse out the linked list of assignment pages
```
edx-abolger, line 419: This parsing is hard to read as a reader of code, please add a response example or a URL link to canvas docs for why you know this parser is correct.

integrated_channels/canvas/client.py
```diff
@@ -378,37 +378,56 @@ def _handle_canvas_assignment_retrieval(
...
+                assignment_page_results = resp.headers['Link'].split(',')
+                current_page = assignment_page_results[0].split(';')[0].strip('<>')
+                last_page = assignment_page_results[3].split(';')[0].strip('<>')
+
+                if current_page == last_page:
+                    more_pages_present = False
+                else:
+                    next_page_url = assignment_page_results[1].split(';')[0].strip('<>')
```
edx-abolger, line 427: Nit: I don't love the loop style here where you have 1 get request outside the loop and 1 inside it. It's harder to reason about the code when I come back to read it. Next time consider a loop like this:
more_pages_present = True
while more_pages_present: 
   url = some_url_string_logic()
   resp = self.session(url)
   done = do_stuff_with response()
   if done: break

tests/test_integrated_channels/test_canvas/test_client.py
```diff
@@ -444,6 +448,71 @@ def test_create_course_success(self):
             assert status_code == 201
             assert response_text == expected_resp
 
+    def test_assignment_retrieval_pagination(self):
+        """
+        Placeholder
```
edx-abolger, line 453: docstring?

integrated_channels/canvas/client.py
```diff
@@ -378,37 +378,58 @@ def _handle_canvas_assignment_retrieval(
...
-        assignment_id = None
-        for assignment in assignments_resp:
-            try:
-                if assignment['integration_id'] == integration_id:
-                    assignment_id = assignment['id']
-                    break
-            # The validation check above should ensure that we have a 200 response from Canvas, but sanity catch if we
-            # have a unexpected response format
-            except (KeyError, ValueError, TypeError) as error:
+        while more_pages_present and current_page_count < 150:
```
edx-abolger, line 392: Comment on why this is here as a timeout, otherwise it looks random

edx-abolger (changes requested):

integrated_channels/canvas/client.py
alex-sheehan-edx: Agreed, I shot myself in the foot by trying to preserve the original beginning of the function. If you feel as though it's too unreadable I could be convinced to rearrange things but do at least believe that this should function as is.

integrated_channels/canvas/client.py
```diff
@@ -378,37 +378,65 @@ def _handle_canvas_assignment_retrieval(
...
+            assignments_resp = resp.json()
+            for assignment in assignments_resp:
+                try:
+                    if assignment['integration_id'] == integration_id:
+                        assignment_id = assignment['id']
+                        break
+
+                # The integration ID check above should ensure that we have a 200 response from Canvas,
+                # but sanity catch if we have a unexpected response format
+                except (KeyError, ValueError, TypeError) as error:
```
binodpant, line 411: what happens if some other error not in this list happens?

binodpant: guess not, I just realized the code in try is simple so mark this resolved!

integrated_channels/canvas/client.py
```diff
@@ -378,37 +378,65 @@ def _handle_canvas_assignment_retrieval(
             course_id (str) : the Canvas course ID relating to the course which the client is currently
             transmitting learner data to.
         """
-        # First, check if the course assignment already exists
+        # Check if the course assignment already exists
```
binodpant, line 381: a quick thought: shouldn't we increase the per_page size to larger value so we won't even need to loop in most cases? of course we still want the logic you added. Just a thought on whether simply increasing per_page will in many cases get us a hit the first time?

binodpant: I don't know what an average assignment count for a user will be, but looks like getting at least 50 or so in a single call may be worthwhile to reduce end-end calls? presuming it won't be too much to ask for a single request

integrated_channels/canvas/client.py
```diff
@@ -378,37 +378,65 @@ def _handle_canvas_assignment_retrieval(
...
+                # 'headers': {
+                #     'Link': '<{assignment_url}?page=2&per_page=10>; rel="current",' \
+                #             '<{assignment_url}?page=1&per_page=10>; rel="prev",' \
+                #             '<{assignment_url}?page=1&per_page=10>; rel="first",' \
+                #             '<{assignment_url}?page=2&per_page=10>; rel="last"' \
+                # }
+                # so we have to parse out the linked list of assignment pages
+                assignment_page_results = resp.headers['Link'].split(',')
+                current_page = assignment_page_results[0].split(';')[0].strip('<>')
+                last_page = assignment_page_results[3].split(';')[0].strip('<>')
```
binodpant, line 430: I see this in their doc `rel="last" may also be excluded if the total count is too expensive to compute on each request.`. Guess we won't hit this often. But is there a chance the 'last' page will be absent in some cases?

integrated_channels/canvas/client.py
```diff
@@ -378,37 +378,65 @@ def _handle_canvas_assignment_retrieval(
...
-        for assignment in assignments_resp:
-            try:
-                if assignment['integration_id'] == integration_id:
-                    assignment_id = assignment['id']
-                    break
-            # The validation check above should ensure that we have a 200 response from Canvas, but sanity catch if we
-            # have a unexpected response format
-            except (KeyError, ValueError, TypeError) as error:
+        # current_page_count serves as a timeout, limiting to a max of 150 pages of requests
+        while more_pages_present and current_page_count < 150:
```
binodpant, line 393: I now realize this is assignments for a single course right? so it say 2000 students are enrolled in the course over time, we expect to get say 2000 assignments in this response total? if per_page=10, we are going to possibly make 200 calls.... not sure if these numbers are realistic but now I am like if we keep per_page=10 we could end up making a lot of requests, assuming large enrollment counts

integrated_channels/canvas/client.py
alex-sheehan-edx: According to the Canvas documentation:

> There is an unspecified limit to how big you can set per_page to, so be sure to always check for the Link header.

As it reads, it seems to me that the limit is either purposely hidden or varies instance to instance. I'm worried about over setting the page limit and causing issues. After that, it was a matter debating what is a number that's not _**too**_ big yet big enough and from discussions with @edx-abolger, we figured that extra API calls isn't too detrimental and we know that the default size will always work with whatever instance we're integrated with.

integrated_channels/canvas/client.py
alex-sheehan-edx: this is an interesting point. Let me investigate behaviors but yeah I should probably account for this.

integrated_channels/canvas/client.py
alex-sheehan-edx: The assignments are per subsection on edX, we shouldn't be creating dups per student.

integrated_channels/canvas/client.py
binodpant: haha yeah that line read like per_page=1million even works lol. ok great, sounds like you have already gone over this consideration! Just wanted to ensure we thought about this given the no. of assignments returned will likely grow over time for a given course, and can be large for large enrollment size courses (bulk enrollment in Canvas lol!) 

But yeah,  we should look at real data once we launch this anyway before more changes, this setup of per_page 10 may turn out to be ideal even!

integrated_channels/canvas/client.py
edx-abolger: @binodpant this ends up being fine because the API doesn't work the way you expect (I also had to look this up because I too was expecting the API call to need a student_id filter or similar on it). Instead, an 'assignment' object here is an object that belongs to a course, there is a different API (not the one we call) called 'List assignments for user' that does what we would expect would happen.

binodpant: thanks if we’re not doing crazy requests/loops per course we should be fine, we can also collect data. I also wish canvas had better search apis so we coukd look for specific assignments for presence but ok...let’s sendcthis out so we resolve the duplicate issue in the least

edx-abolger (approved):

PR #1263, opened 2021-06-11
ENT-4655 |unlink enterprise learners command

HammadAhmadWaqas:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/management/commands/unlink_enterprise_customer_learners.py
+++ b/enterprise/management/commands/unlink_enterprise_customer_learners.py
@@ -1,0 +1,104 @@
+# -*- coding: utf-8 -*-
+"""
+Django management command to unlink the learners, Delete the enrollments and remove the DSC.
+"""
+import csv
+import logging
+
+from django.contrib import auth
+from django.core.management import BaseCommand
+
+from enterprise.models import (
+    EnterpriseCourseEnrollment,
+    EnterpriseCustomer,
+    EnterpriseCustomerUser,
+    PendingEnterpriseCustomerUser,
+)
+from enterprise.utils import delete_data_sharing_consent
+
+LOGGER = logging.getLogger(__name__)
+
+User = auth.get_user_model()
+
+
+class Command(BaseCommand):
+    """
+    Django management command to unlink the learners, Delete the enrollments and remove the DSC
+
+    Example usage:
+    ./manage.py lms unlink_enterprise_customer_learners -e 22b783f5-e1d9-4228-87e4-5cf1cf110320 --data-csv path/file.csv
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--data-csv',
+            action='store',
+            dest='data_csv',
+            help='Path of csv to read emails to be unlinked.',
+            type=str,
+        )
+
+        parser.add_argument(
+            '-e',
+            '--enterprise_customer_uuid',
+            action='store',
+            dest='enterprise_customer_uuid',
+            help='Run this command for only the given EnterpriseCustomer UUID.'
+        )
+
+    def handle(self, *args, **options):
+        csv_path = options['data_csv']
+        enterprise_customer_uuid = options['enterprise_customer_uuid']
+        enterprise_customer = EnterpriseCustomer.objects.get(uuid=enterprise_customer_uuid)
+
+        LOGGER.info('[Unlink and Remove DSC]  Process started.')
+
+        results = {}
+
+        with open(csv_path) as csv_file:
+            reader = csv.DictReader(csv_file)
+            for row in reader:
+                email = row['email']
+                results[email] = {'Unlinked': None, 'Removed_DSC': None, 'Removed_EnterpriseCourseEnrolments': None}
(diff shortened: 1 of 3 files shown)


enterprise/management/commands/unlink_enterprise_customer_learners.py
```diff
@@ -0,0 +1,86 @@
...
+    def handle(self, *args, **options):
+        csv_path = options['data_csv']
+        enterprise_customer_uuid = options['enterprise_customer_uuid']
+        enterprise_customer = EnterpriseCustomer.objects.get(uuid=enterprise_customer_uuid)
+
+        LOGGER.info('Started Unlinking and removing DSC')
+
+        results = {}
+
+        with open(csv_path) as csv_file:
```
georgebabey, line 53: For a one-time thing this is probably fine, but I would recommend reading the file first and then doing the operations outside of the `with open` so you don't keep the file handler open longer than you need to.

enterprise/management/commands/unlink_enterprise_customer_learners.py
```diff
@@ -0,0 +1,86 @@
...
+                        email=email, ec_name=enterprise_customer.name
+                    )
+                    results[email]['Unlinked'] = message
+                    continue
+
+                user = User.objects.get(email=email)
+                enterprise_customer_user = get_enterprise_customer_user(user.id, enterprise_customer_uuid)
+                data_sharing_consent_records = enterprise_customer_user.data_sharing_consent_records
+
+                # Deleting the DCS cache
```
georgebabey, line 74: But: DSC not DCS

enterprise/management/commands/unlink_enterprise_customer_learners.py
```diff
@@ -0,0 +1,86 @@
...
+from enterprise.utils import get_enterprise_customer_user
+
+User = auth.get_user_model()  # pylint: disable=invalid-name
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Django management command for creating DataSharingConsent records
```
sameenfatima78, line 24: Nit: Incorrect docstring

enterprise/management/commands/unlink_enterprise_customer_learners.py
```diff
@@ -0,0 +1,86 @@
...
+                    results[email]['Unlinked'] = message
+                    continue
+
+                user = User.objects.get(email=email)
+                enterprise_customer_user = get_enterprise_customer_user(user.id, enterprise_customer_uuid)
+                data_sharing_consent_records = enterprise_customer_user.data_sharing_consent_records
+
+                # Deleting the DCS cache
+                deleted_course_ids = []
+                for data_sharing_consent in data_sharing_consent_records:
```
georgebabey, line 76: We also need to delete the actual DSC record, not just the cache (although the cache was a great catch I had forgotten about)

enterprise/management/commands/unlink_enterprise_customer_learners.py
```diff
@@ -0,0 +1,86 @@
+# -*- coding: utf-8 -*-
+"""
+Django management command for unlink the learners and removing DSC.
```
sameenfatima78, line 3: nit: for unlinking

georgebabey (changes requested):

enterprise/management/commands/unlink_enterprise_customer_learners.py
```diff
@@ -0,0 +1,86 @@
...
+            dest='enterprise_customer_uuid',
+            help='Run this command for only the given EnterpriseCustomer UUID.'
+        )
+
+    def handle(self, *args, **options):
+        csv_path = options['data_csv']
+        enterprise_customer_uuid = options['enterprise_customer_uuid']
+        enterprise_customer = EnterpriseCustomer.objects.get(uuid=enterprise_customer_uuid)
+
+        LOGGER.info('Started Unlinking and removing DSC')
```
zamanafzal, line 49: Can we please add logs with a constant primary string like 
LOGGER.info('[Unlink and Remove DSC] Started Unlinking and removing DSC')

enterprise/management/commands/unlink_enterprise_customer_learners.py
zamanafzal, line 24: nit: Please update the docstring.

enterprise/management/commands/unlink_enterprise_customer_learners.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    Django management command for creating DataSharingConsent records
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--data-csv',
+            action='store',
+            dest='data_csv',
+            help='Path of csv to read enterprise learner data',
```
zamanafzal, line 32: nit: Path of csv to read enterprise learner email.

georgebabey (approved):

sameenfatima78 (approved):

PR #1272, opened 2021-06-16
feat: Added upgrade python requirements workflow

Jawayria:
JIRA: https://openedx.atlassian.net/browse/BOM-2596


.github/workflows/upgrade-python-requirements.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: make upgrade
+        run: |
+          cd $GITHUB_WORKSPACE
+          make upgrade
+
+      - name: setup testeng-ci
+        run: |
+          git clone https://github.com/edx/testeng-ci.git
+          cd $GITHUB_WORKSPACE/testeng-ci
+          pip install -r requirements/base.txt
```
UsamaSadiq, line 44: edx-enterprise doesn't have `base.txt` in it so you should use `dev.txt` instead.

mraarif: this is meant to be `base.txt` in `testeng-ci` as you can see we're moving into `testeng-ci` in step above

.github/workflows/upgrade-python-requirements.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Send failure notification
+        if: ${{ failure() }}
+        uses: dawidd6/action-send-mail@v3
+        with:
+          server_address: email-smtp.us-east-1.amazonaws.com
+          server_port: 465
+          username: ${{secrets.EDX_SMTP_USERNAME}}
+          password: ${{secrets.EDX_SMTP_PASSWORD}}
+          subject: Upgrade python requirements workflow failed in ${{github.repository}}
+          to: arbi-bom@edx.org
```
mraarif, line 66: we should update this email to send notifications to the owning team

Jawayria: Jenkins job was using this email.

mraarif: we should use enterprise team's email

Jawayria: @edx/enterprise-titans can you please mention team's complete email address so that I could mention here?

iloveagent57: 👍  I left a comment in the ticket.

Jawayria: Thanks

.github/workflows/upgrade-python-requirements.yml
```diff
@@ -0,0 +1,68 @@
...
+  upgrade_requirements:
+    runs-on: ubuntu-20.04
+
+    strategy:
+      matrix:
+        python-version: ["3.8"]
+
+    steps:
+      - name: setup target branch
+        run: echo "target_branch=$(if ['${{ github.event.inputs.branch }}' = '']; then echo 'master'; else echo '${{ github.event.inputs.branch }}'; fi)" >> $GITHUB_ENV
```
macdiesel, line 24: Let me preface this by saying I am just now learning about github actions: Question, why would you specify inputs.branch.default = 'master' but then set 'master' again here manually?

macdiesel (approved):
Just one question about the setup. Otherwise looks good. LMK and I can merge.

.github/workflows/upgrade-python-requirements.yml
mraarif: this adds the ability to trigger this workflow manually and specify a different branch than `master` to run the workflow
as you can see here: https://github.com/edx/edx-platform/actions/workflows/upgrade-python-requirements.yml
if the workflow is run by corn schedule it'd run the workflow against `master`

mraarif:
Hi @macdiesel please let us know if you've any further questuions

PR #1276, opened 2021-06-21
ENT4379: Route to default_idp if no tpa_hint is added in query_param

zamanafzal:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -407,7 +407,7 @@ class EnterpriseCustomer(TimeStampedModel):
     @property
     def has_multiple_idps(self):
         """
-        Return True if there are any identity providers associated with this enterprise customer.
+        Return True if there are multiple identity providers associated with this enterprise customer.
         """
         # pylint: disable=no-member
         return self.enterprise_customer_identity_providers.count() > 1
@@ -453,6 +453,8 @@ class EnterpriseCustomer(TimeStampedModel):
         # if there is only one identity provider linked with the customer and tpa_hint_param was not provider.
         if self.has_single_idp and not tpa_hint_param:
             return self.identity_providers.first().provider_id
+        if self.has_multiple_idps and not tpa_hint_param:
+            return self.default_provider_idp.provider_id if self.default_provider_idp else None
         # Now if there is not any linked identity provider OR there are multiple identity providers.
         return None
 
@@ -677,8 +679,11 @@ class EnterpriseCustomer(TimeStampedModel):
         course_path = '/courses/{course_id}/course'.format(course_id=course_id)
         params = {}
         # add tap_hint if there is only one IdP attached with enterprise_customer
-        if self.identity_providers.count() == 1:
+        if self.has_single_idp:
             params = {'tpa_hint': self.identity_providers.first().provider_id}
+
+        elif self.has_multiple_idps and self.default_provider_idp:
+            params = {'tpa_hint': self.default_provider_idp.provider_id}
         course_path = urlquote("{}?{}".format(course_path, urlencode(params)))
 
         lms_root_url = utils.get_configuration_value_for_site(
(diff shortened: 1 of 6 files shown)


enterprise/models.py
```diff
@@ -453,6 +453,8 @@ def get_tpa_hint(self, tpa_hint_param=None):
         # if there is only one identity provider linked with the customer and tpa_hint_param was not provider.
         if self.has_single_idp and not tpa_hint_param:
             return self.identity_providers.first().provider_id
+        if self.has_multiple_idps and not tpa_hint_param:
+            return self.default_provider_idp.provider_id
```
HammadAhmadWaqas, line 457: For the safe side we should check if the `self.default_provider_idp` exists before excessing its `provider_id`.

enterprise/models.py
```diff
@@ -677,8 +679,10 @@ def notify_enrolled_learners(self, catalog_api_user, course_id, users, admin_enr
         course_path = '/courses/{course_id}/course'.format(course_id=course_id)
         params = {}
         # add tap_hint if there is only one IdP attached with enterprise_customer
-        if self.identity_providers.count() == 1:
+        if self.has_single_idp:
             params = {'tpa_hint': self.identity_providers.first().provider_id}
+        elif self.has_multiple_idps:
+            params = {'tpa_hint': self.default_provider_idp.provider_id}
```
HammadAhmadWaqas, line 685: same concern as above.

tests/test_models.py
```diff
@@ -234,6 +234,40 @@ def test_identity_providers(self):
...
+            default_provider=True
+        )
+        ent_idp_two = factories.EnterpriseCustomerIdentityProviderFactory(enterprise_customer=customer)
+        assert customer.get_tpa_hint() == ent_idp_one.provider_id
+        ent_idp_one.default_provider = False
+        ent_idp_one.save()
+        ent_idp_two.default_provider = True
+        ent_idp_two.save()
+        assert customer.get_tpa_hint() == ent_idp_two.provider_id
+
```
HammadAhmadWaqas, line 270: Can we add a test if there are multiple attached but not default IDP exists.

HammadAhmadWaqas (changes requested):

enterprise/models.py
zamanafzal: If an enterprise wants to link more than one IDP then it has to made default one of them. Otherwise, it can not link it.

zamanafzal: Screenshot 2021-06-22 at 2 31 45 PM

HammadAhmadWaqas: @zamanafzal You are 100% right, we can't link multiple IDPs without defining default from the admin, But this restriction is only in admin view. But this constraint is not on the Model level. In future, there might be scenarios when we link multiple IDPs with default so I would recommend better checking if default IDP is there before using it.

zamanafzal: I think this behaviour is not expected if we look at the default IDP ticket. But still for the safe side, I have added this check and test case accordingly.

HammadAhmadWaqas (approved):

enterprise/models.py
```diff
@@ -677,8 +679,11 @@ def notify_enrolled_learners(self, catalog_api_user, course_id, users, admin_enr
         course_path = '/courses/{course_id}/course'.format(course_id=course_id)
         params = {}
         # add tap_hint if there is only one IdP attached with enterprise_customer
-        if self.identity_providers.count() == 1:
+        if self.has_single_idp:
             params = {'tpa_hint': self.identity_providers.first().provider_id}
+
+        elif self.has_multiple_idps and self.default_provider_idp:
```
muhammad-ammar, line 685: What if there are `multiple idps` but there is no `default provider idp`?

muhammad-ammar (approved):

saleem-latif (approved):

enterprise/models.py
zamanafzal: This is discussed at the first comment in the PR.
https://github.com/edx/edx-enterprise/pull/1276#discussion_r656038394

PR #1278, opened 2021-06-22
fix(integrated_channels): :bug: fix forward for parameter rename bug

edx-abolger:
Caused in 3.26.7

**Merge checklist:**
- [n/a ] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files)
- [n/a ] `make static` has been run to update webpack bundling if any static content was updated
- [n/a] `./manage.py makemigrations` has been run
- [x ] [Version](https://github.com/edx/edx-enterprise/blob/master/enterprise/__init__.py) bumped
- [x ] [Changelog](https://github.com/edx/edx-enterprise/blob/master/CHANGELOG.rst) record added
- [n/a ] Translations updated (see docs/internationalization.rst but also this isn't blocking for merge atm)

**Post merge:**

### DIFF
--- a/tests/test_integrated_channels/test_integrated_channel/test_transmitters/test_learner_data.py
+++ b/tests/test_integrated_channels/test_integrated_channel/test_transmitters/test_learner_data.py
@@ -10,6 +10,7 @@ import mock
 from pytest import mark
 
 from integrated_channels.integrated_channel.exporters.learner_data import LearnerExporter
+from integrated_channels.integrated_channel.tasks import transmit_single_learner_data
 from integrated_channels.integrated_channel.transmitters.learner_data import LearnerTransmitter
 from test_utils import factories
 
@@ -37,6 +38,23 @@ class TestLearnerDataTransmitter(unittest.TestCase):
         )
 
         self.learner_transmitter = LearnerTransmitter(self.enterprise_config)
+
+    def test_transmit_single_learner_data_signal_kwargs(self):
+        """ transmit_single_learner_data is called with kwargs as a shared task from OpenEdx,
+        so test that interface hasn't changed. """
+
+        edx_platform_api_signal_kwargs = {
+            'username': "fake_username",
+            'course_run_id': "TEST_COURSE_RUN_KEY"
+        }
+        # If you change the arg names to transmit_single_learner_data
+        # this test will now fail with a TypeError:
+        try:
+            transmit_single_learner_data(**edx_platform_api_signal_kwargs)
+        except TypeError:
+            assert False  # if we see a Type error, the test should fail because we failed with bad keyword parameters
+        except Exception:   # pylint: disable=broad-except
+            pass  # otherwise it's because we set the test up with no data.
 
     @mock.patch("integrated_channels.integrated_channel.models.LearnerDataTransmissionAudit")
     @mock.patch("integrated_channels.utils.is_already_transmitted")
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * Nothing
 
+[3.26.10]
+--------
+* Fix forward for parameter rename changing the signals API in 3.26.7
+
 [3.26.9]
 --------
 * Added support to use default idp in Enterprise slug login if there are multiple.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,7 +18,7 @@ Unreleased
 * Nothing
 
 [3.26.10]
---------
+---------
 * Fix forward for parameter rename changing the signals API in 3.26.7
 
 [3.26.9]
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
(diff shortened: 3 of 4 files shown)


tests/test_integrated_channels/test_integrated_channel/test_transmitters/test_learner_data.py
```diff
@@ -38,6 +39,23 @@ def setUp(self):
...
+            'username': "fake_username",
+            'course_run_id': "TEST_COURSE_RUN_KEY"
+        }
+        # If you change the arg names to transmit_single_learner_data
+        # this test will now fail with a TypeError:
+        try:
+            transmit_single_learner_data(**edx_platform_api_signal_kwargs)
+        except TypeError:
+            assert False  # if we see a Type error, the test should fail because we failed with bad keyword parameters
+        except Exception:   # pylint: disable=broad-except
```
taliaedX, line 56: This feels like a weird pass condition. I'm also generally not a fan of highly specific tests that only really avoid 1 scenario we've already hit. Ideally I'd rather us actually test the logic of this method but yeah that's probably nightmarish.

taliaedX (approved):
I'm really not a fan of the test, but I don't have better suggestions without doing some substantial digging and as we want this to stop failing asap, I'll approve it.

tests/test_integrated_channels/test_integrated_channel/test_transmitters/test_learner_data.py
edx-abolger: Oh this does not test the function at all: the problem is that unlike for REST API violations, no test exists to catch if we violate the plugin interface with the wrong params.

iloveagent57: You could do something like the converse: call it with the wrong keyword(s) and do a `with self.assertRaises(...):`

taliaedX: That wouldn't have caught this issue though would it? The issue was that we were in fact passing the wrong keywords. Thus a test that asserts that error gets raised is saying "yes this fails because these keywords are wrong and we don't want them. That is correct and thus this test passes!" would also have passed before this code change?

macdiesel:
FWIW I don't think a test is necessary here and I would advocate for not adding one at all to cover this edge case. Changing APIs (in this case the method signature) is a risky proposition. Adding a test to cover this doesn't really improve the testing and makes the runtime longer. Both things we should be avoiding. 

For me the take away on this is that changing method signatures is risky and not high value. I think that adding a more detailed doc string would have solved the same problem. We should  avoid doing so in the future unless we have a really compelling case to do so and if we make that decision we should grep the other code bases that depend on this function/method/API. 

Last thought, when API changes like this are made, it should be a major version update (the number to the left of the dot) and not a minor version upgrade as suggested by [semantic versioning.](https://semver.org/)

macdiesel:
For edX Semantic Versioning is covered under [OEP-47](https://open-edx-proposals.readthedocs.io/en/latest/oep-0047-bp-semantic-versioning.html)

tests/test_integrated_channels/test_integrated_channel/test_transmitters/test_learner_data.py
edx-abolger: @taliaedX you are correct: the goal of the test was defensive coding that would raise an error if someone did what I did and changed a method signature. However, I do like @macdiesel 's solution going forward of updating the doc strings to note that the method signatures are actually an API contract instead, as that provides defensive coding right where the programmer can see it. I will make another small doc-only PR with a docstring note in anything in integrated channels we call as a kwarg.

edx-abolger:
> FWIW I don't think a test is necessary here and I would advocate for not adding one at all to cover this edge case. Changing APIs (in this case the method signature) is a risky proposition. Adding a test to cover this doesn't really improve the testing and makes the runtime longer. Both things we should be avoiding.
> 
> For me the take away on this is that changing method signatures is risky and not high value. I think that adding a more detailed doc string would have solved the same problem. We should avoid doing so in the future unless we have a really compelling case to do so and if we make that decision we should grep the other code bases that depend on this function/method/API.
> 
> Last thought, when API changes like this are made, it should be a major version update (the number to the left of the dot) and not a minor version upgrade as suggested by [semantic versioning.](https://semver.org/)

Would the major bump you're advocating for here have been when I changed the signature in the first place? In retrospect I agree, (had I realized it was an interface change) but the bug fix to put it back I still consider a patch? I suppose semantic versioning can't really have opinions beyond 'always bump the version' about  "versions of code you intend to not have any breaking changes but actually do ...".

macdiesel:
> Would the major bump you're advocating for here have been when I changed the signature in the first place? 

Correct.

>In retrospect I agree, (had I realized it was an interface change) but the bug fix to put it back I still consider a patch? I suppose semantic versioning can't really have opinions beyond 'always bump the version' about "versions of code you intend to not have any breaking changes but actually do ...".

Yeah it kind of ends up in strange territory. I think that the minor version bump for the fix is fine since the API change was pulled back.

PR #1279, opened 2021-06-23
feat: When a course is found in concluded/completed state, update instead of create ENT-4594

binodpant:
This change eliminates the failures encountered during create_content_metadata when a course is previously concluded (via a previous delete attempt). Instead, now we do a detection of course state before attempting to create it, and instead we update it if it's one of the following states in Canvas:
`'unpublished', 'completed' or 'available' `

- [x] This also brings a course back from completed to 'offer' state after which the course is now once again available for enrolling in Canvas
- [x] Also add detection of 'deleted' courses previously missed, since that allows us to log that a course has previojsly been 'deleted' as opposed to say 'concluded' (or 'completed'). We take no action on deleted courses other than logging for now
- [x] This can now even locate courses in other subaccount under the root account, as a fallback, in case not found under current subaccount.
- [x] Also, update_content_metadata and delete_content_metadata also use the same method I refactored `_get_course_id_from_edx_course_id` hence they will also benefit from being able to find a course in other subaccounts
- [x] Automated tests added. 

# Testing notes

After removing a course from catalog, verified it was in 'completed' workflow_state now. Then added back to catalog and ran transmit_content_metadata task and saw this in logs

```
2021-06-23 03:34:26,542 WARNING 49226 [integrated_channels.canvas.client] [user None] [ip None] client.py:129 - Course with canvas_id = 6072, integration_id = course-v1:edX+99887766+2T2021 found in workflow_state=completed,attempting to update instead of creating it
2021-06-23 03:34:27,026 WARNING 49226 [integrated_channels.canvas.client] [user None] [ip None] client.py:129 - Course with canvas_id = 6074, integration_id = edX+99887766 found in workflow_state=completed,attempting to update instead of creating it
2021-06-23 03:34:27,259 INFO 49226 [integrated_channels.integrated_channel.tasks] [user None] [ip None] tasks.py:47 - [Integrated Channel: CANVAS] Batch transmit_content_metadata finished in 1.3078796863555908 (api user: 3). Configuration: <CanvasEnterpriseCustomerConfiguration for Enterprise Test Enterprise>.
```

Then verified course in canvas went from ('concluded' or) 'completed' state to 'available'
Also verified via Canvas UI that course button now says 'Conclude this course', instead of 'Unconclude this course'

This payload obtained from `https://edx.instructure.com//api/v1/courses/6072?include=course_image&include=self_enrollment` shows the course state after the second round of transmit_content_metadata (there were no failures!)

```
{
    "id": 6072,
    "name": "bulk enroll test course",
    "account_id": 109,
    "uuid": "HkGD3EsTg8ihhoWArHlQRDMY8Aqy1hLqUfXUm62A",
    "start_at": "2021-05-26T16:00:00Z",
    "grading_standard_id": null,
    "is_public": true,
    "created_at": "2021-05-27T18:48:25Z",
    "course_code": "course-v1:edX+99887766+2T2021",
    "default_view": "syllabus",
    "root_account_id": 1,
    "enrollment_term_id": 1,
    "self_enrollment": true,
    "license": null,
    "grade_passback_setting": null,
    "end_at": "2022-08-27T16:00:00Z",
    "public_syllabus": false,
    "public_syllabus_to_auth": false,
    "storage_quota_mb": 500,
    "is_public_to_auth_users": false,
    "homeroom_course": false,
    "course_color": null,
    "hide_final_grades": false,
    "apply_assignment_group_weights": false,
    "calendar": {
        "ics": "https://edx.instructure.com/feeds/calendars/course_HkGD3EsTg8ihhoWArHlQRDMY8Aqy1hLqUfXUm62A.ics"
    },
    "time_zone": "America/Denver",
    "blueprint": false,
    "template": false,
    "sis_course_id": null,
    "sis_import_id": null,
    "integration_id": "course-v1:edX+99887766+2T2021",
    "enrollments": [],
    "workflow_state": "available",
    "restrict_enrollments_to_course_dates": false
}
```

__Also checked deleted course state handling__

Deleted a course in Canvas using the 'delete' (not conclude) option. Verified course returned now as 'workflow_state' = 'deleted' then ran a job (after clearing local tx records. This time instead of faiing to create it simply logged

```
2021-06-24 14:53:33,220 WARNING 51939 [integrated_channels.canvas.client] [user None] [ip None] client.py:122 - Course with integration_id = course-v1:edX+99887766+2T2021 found in deleted state,not attempting to create/update
```

__Finding course under root account instead of current subaccount__

I have to now add automated tests, but tested this out manually with various cases such as:

- deleted a course in canvas, then verified it can find it and log/ignore in that case (we still add a success tx record in this case)
- made a course unpublished in canvas, and it correctly detects state and updates course, instead of creating it
- forced it to try and find course under root account (by using a different canvas account id where course does not exist, and it did go find the course under root account id in that case)

```
2021-06-25 01:57:38,882 INFO 61047 [integrated_channels.canvas.client] [user None] [ip None] client.py:437 - [integrated_channel:canvas][integrated_channel_enterprise_customer_uuid:873fec72-5a21-4f95-ac19-e3d67f9c4777][integrated_channel_lms_user:None][integrated_channel_course_key:course-v1:edX+99887766+2T2021]Found course under root Canvas account
2021-06-25 01:57:38,883 ERROR 61047 [integrated_channels.canvas.client] [user None] [ip None] client.py:123 - [integrated_channel:canvas][integrated_channel_enterprise_customer_uuid:873fec72-5a21-4f95-ac19-e3d67f9c4777][integrated_channel_lms_user:None][integrated_channel_course_key:course-v1:edX+99887766+2T2021]Course with integration_id = course-v1:edX+99887766+2T2021 found in deleted state, not attempting to create/update
```

__Also I manually moved a course to a different subaccount in Canvas side__

Then remove the tx record to force it to create a course and voila! it found the course and instead updated it back from the concluded to offer state

ENT-4594

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/canvas/client.py
+++ b/integrated_channels/canvas/client.py
@@ -19,6 +19,9 @@ from integrated_channels.utils import generate_formatted_log, refresh_session_if
 LOGGER = logging.getLogger(__name__)
 
 
+MESSAGE_WHEN_COURSE_WAS_DELETED = 'Course was deleted previously, skipping create/update'
+
+
 class CanvasAPIClient(IntegratedChannelApiClient):
     """
     Client for connecting to Canvas.
@@ -136,7 +139,7 @@ class CanvasAPIClient(IntegratedChannelApiClient):
                     )
                 )
                 status_code = 200
-                response_text = 'Course was deleted previously, skipping create/update'
+                response_text = MESSAGE_WHEN_COURSE_WAS_DELETED
             else:
                 # 'unpublished', 'completed' or 'available' cases
                 LOGGER.warning(
--- a/integrated_channels/canvas/client.py
+++ b/integrated_channels/canvas/client.py
@@ -73,7 +73,7 @@ class CanvasAPIClient(IntegratedChannelApiClient):
     @staticmethod
     def course_fetch_endpoint(canvas_base_url, canvas_account_id):
         """
-        Returns endpoint to GET to for course fetch
+        Returns endpoint to fetch all courses for the specified Canvas account
         """
         return '{}/api/v1/accounts/{}/courses'.format(
             canvas_base_url,
@@ -338,6 +338,10 @@ class CanvasAPIClient(IntegratedChannelApiClient):
           under the current subaccount
           Will even return courses that are in the 'deleted' state in Canvas, so we can correctly
           skip these courses in logic for create_or_update when needed.
+
+          Note: we do not need to follow pagination here since it would be extremely unlikely
+          that searching by a specific edx_course_id results in many records, we generally only
+          expect 1 record to come back anyway.
         """
         url = "{}/api/v1/accounts/{}/courses/?search_term={}".format(
             self.enterprise_configuration.canvas_base_url,
--- a/integrated_channels/canvas/client.py
+++ b/integrated_channels/canvas/client.py
@@ -2,6 +2,7 @@
 """
 Client for connecting to Canvas.
 """
+import logging
 import json
 from http import HTTPStatus
 
@@ -14,6 +15,8 @@ from django.apps import apps
 from integrated_channels.exceptions import ClientError
 from integrated_channels.integrated_channel.client import IntegratedChannelApiClient
 from integrated_channels.utils import refresh_session_if_expired
+
+LOGGER = logging.getLogger(__name__)
 
 
 class CanvasAPIClient(IntegratedChannelApiClient):
@@ -68,6 +71,16 @@ class CanvasAPIClient(IntegratedChannelApiClient):
         )
 
     @staticmethod
+    def course_fetch_endpoint(canvas_base_url, canvas_account_id):
+        """
+        Returns endpoint to GET to for course fetch
+        """
+        return '{}/api/v1/accounts/{}/courses'.format(
+            canvas_base_url,
+            canvas_account_id,
+        )
+
+    @staticmethod
     def course_update_endpoint(canvas_base_url, course_id):
         """
         Returns endpoint to PUT to for course update
@@ -78,14 +91,52 @@ class CanvasAPIClient(IntegratedChannelApiClient):
         )
 
     def create_content_metadata(self, serialized_data):
+        """
+        Creates a course in Canvas.
+        If course is not found ,easy!  create it as usual
+        If course found, it will be one of these workflow_states per doc /the current state
+            of the course one of 'unpublished', 'available', 'completed', or 'deleted'
+                If available: issue an update with latest field values
+                If completed: this happens if a course has been concluded.
+                    update it to change status to offer as per course[event]=offer
+                If ‘unpublished’ - still just update
+                If ‘deleted’ - take no action for now.
+        https://canvas.instructure.com/doc/api/courses.html#method.courses.update
+        """
         self._create_session()
 
-        # step 1: create the course
-        status_code, response_text = self._post(self.course_create_url, serialized_data)
-        created_course_id = json.loads(response_text)['id']
+        course_details = json.loads(serialized_data.decode('utf-8'))['course']
+        edx_course_id = course_details['integration_id']
+        located_course = self._find_course_by_course_id(edx_course_id)
 
-        # step 2: upload image_url and any other details
-        self._update_course_details(created_course_id, serialized_data)
+        if not located_course:
+            # Course does not exist: Create the course
+            status_code, response_text = self._post(self.course_create_url, serialized_data)
+            created_course_id = json.loads(response_text)['id']
+
+            # step 2: upload image_url and any other details
+            self._update_course_details(created_course_id, course_details)
+        else:
+            workflow_state = located_course['workflow_state']
+            if workflow_state.lower() == 'deleted':
+                LOGGER.warning(
+                    'Course with integration_id = %s found in deleted state,'
+                    'not attempting to create/update',
+                    edx_course_id,
+                )
+            else:
+                # 'unpublished', 'completed' or 'available' cases
+                LOGGER.warning(
+                    'Course with canvas_id = %s, integration_id = %s found in workflow_state=%s,'
+                    'attempting to update instead of creating it',
+                    located_course['id'],
+                    edx_course_id,
+                    workflow_state,
+                )
+                status_code, response_text = self._update_course_details(
+                    located_course['id'],
+                    course_details,
+                )
 
         return status_code, response_text
 
@@ -93,7 +144,7 @@ class CanvasAPIClient(IntegratedChannelApiClient):
         self._create_session()
 
         integration_id = self._extract_integration_id(serialized_data)
-        course_id = self._get_course_id_from_integration_id(integration_id)
+        course_id = self._get_course_id_from_edx_course_id(integration_id)
 
         url = CanvasAPIClient.course_update_endpoint(
             self.enterprise_configuration.canvas_base_url,
@@ -106,7 +157,7 @@ class CanvasAPIClient(IntegratedChannelApiClient):
         self._create_session()
 
         integration_id = self._extract_integration_id(serialized_data)
-        course_id = self._get_course_id_from_integration_id(integration_id)
+        course_id = self._get_course_id_from_edx_course_id(integration_id)
 
         url = '{}/api/v1/courses/{}'.format(
             self.enterprise_configuration.canvas_base_url,
@@ -183,28 +234,33 @@ class CanvasAPIClient(IntegratedChannelApiClient):
         pass
 
     # Private Methods
-
-    def _update_course_details(self, course_id, serialized_data):
+    def _update_course_details(self, course_id, course_details):
         """
         Update a course for image_url (and possibly other settings in future)
-        This is used only for settings that are not settable in the initial course creation
+        Also sets course to 'offer' state by sending 'course[event]=offer'
+
+        Arguments:
+          - course_details (dict): { 'image_url' } : optional used if present to course[image_url]
         """
+        response_code = None
+        response_text = None
+        # Providing the param `event` and setting it to `offer` is equivalent to publishing the course.
+        update_payload = {'course': {'event': 'offer'}}
         try:
             # there is no way to do this in a single request during create
             # https://canvas.instructure.com/doc/api/all_resources.html#method.courses.update
-            content_metadata_item = json.loads(serialized_data.decode('utf-8'))['course']
-            if "image_url" in content_metadata_item:
-                url = CanvasAPIClient.course_update_endpoint(
-                    self.enterprise_configuration.canvas_base_url,
-                    course_id,
-                )
-                # Providing the param `event` and setting it to `offer` is equivalent to publishing the course.
-                self._put(url, json.dumps({
-                    'course': {'image_url': content_metadata_item['image_url'], 'event': 'offer'}
-                }).encode('utf-8'))
+            url = CanvasAPIClient.course_update_endpoint(
+                self.enterprise_configuration.canvas_base_url,
+                course_id,
+            )
+            if "image_url" in course_details:
+                update_payload['course']['image_url'] = course_details['image_url']
+
+            response_code, response_text = self._put(url, json.dumps(update_payload).encode('utf-8'))
         except Exception:  # pylint: disable=broad-except
             # we do not want course image update to cause failures
             pass
+        return response_code, response_text
 
     def _post(self, url, data):
         """
@@ -273,18 +329,20 @@ class CanvasAPIClient(IntegratedChannelApiClient):
 
         return integration_id
 
-    def _get_course_id_from_integration_id(self, integration_id):
+    def _find_course_by_course_id(self, edx_course_id):
         """
-        To obtain course ID we have to request all courses associated with the integrated
-        account and match the one with our integration ID.
-
-        Args:
-            integration_id (string): The ID retrieved from the transmission payload.
+        Search course by edx_course_id (used as integration_id in canvas) under given account.
+          Issue with this is that if this course exists in canvas with this edx_course_id
+          under a different subaccount, we will not detect it.
+          Hence we will try to find a course under other subaccounts in case we don't find it
+          under the current subaccount
+          Will even return courses that are in the 'deleted' state in Canvas, so we can correctly
+          skip these courses in logic for create_or_update when needed.
         """
         url = "{}/api/v1/accounts/{}/courses/?search_term={}".format(
             self.enterprise_configuration.canvas_base_url,
             self.enterprise_configuration.canvas_account_id,
-            quote(integration_id),
+            quote(edx_course_id),
         )
         resp = self.session.get(url)
         all_courses_response = resp.json()
@@ -295,20 +353,38 @@ class CanvasAPIClient(IntegratedChannelApiClient):
                 all_courses_response.status_code
             )
 
-        course_id = None
+        course_found = None
         for course in all_courses_response:
-            if course['integration_id'] == integration_id:
-                course_id = course['id']
+            if course['integration_id'] == edx_course_id:
+                course_found = course
                 break
+        return course_found
 
-        if not course_id:
+        # now let's try under all subaccounts
+        """url = "{}/api/v1/accounts/{}/courses/?search_term={}".format(
+            self.enterprise_configuration.canvas_base_url,
+            self.enterprise_configuration.canvas_account_id,
+            quote(edx_course_id),
+        )"""
+
+    def _get_course_id_from_edx_course_id(self, edx_course_id):
+        """
+        To obtain course ID we have to request all courses associated with the integrated
+        account and match the one with our edx_course_id (integration_id).
+
+        Args:
+            edx_course_id (string): Course ID to search by
+        """
+        course = self._find_course_by_course_id(edx_course_id)
+
+        if not course:
             raise ClientError(
-                "No Canvas courses found with associated integration ID: {}.".format(
-                    integration_id
+                "No Canvas courses found with associated edx course ID: {}.".format(
+                    edx_course_id
                 ),
                 HTTPStatus.NOT_FOUND.value
             )
-        return course_id
+        return course['id']
 
     def _search_for_canvas_user_by_email(self, user_email):
         """
@@ -470,12 +546,12 @@ class CanvasAPIClient(IntegratedChannelApiClient):
         Helper method to take necessary learner data and post to Canvas as a submission to the correlated assignment.
         """
         submission_url = '{base_url}/api/v1/courses/{course_id}/assignments/' \
-                         '{assignment_id}/submissions/{user_id}'.format(
-                             base_url=self.enterprise_configuration.canvas_base_url,
-                             course_id=course_id,
-                             assignment_id=assignment_id,
-                             user_id=canvas_user_id
-                         )
+            '{assignment_id}/submissions/{user_id}'.format(
+                base_url=self.enterprise_configuration.canvas_base_url,
+                course_id=course_id,
+                assignment_id=assignment_id,
+                user_id=canvas_user_id
+            )
 
         # The percent grade from the grades api is represented as a decimal
         submission_data = {
--- a/integrated_channels/canvas/client.py
+++ b/integrated_channels/canvas/client.py
@@ -59,6 +59,7 @@ class CanvasAPIClient(IntegratedChannelApiClient):
             self.enterprise_configuration.canvas_base_url,
             self.enterprise_configuration.canvas_account_id
         )
+        self.root_canvas_account = self._find_root_canvas_account()
 
     @staticmethod
     def course_create_endpoint(canvas_base_url, canvas_account_id):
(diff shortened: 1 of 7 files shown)


integrated_channels/canvas/client.py
```diff
@@ -295,20 +353,38 @@ def _get_course_id_from_integration_id(self, integration_id):
...
         for course in all_courses_response:
-            if course['integration_id'] == integration_id:
-                course_id = course['id']
+            if course['integration_id'] == edx_course_id:
+                course_found = course
                 break
+        return course_found
 
-        if not course_id:
+        # now let's try under all subaccounts
```
binodpant, line 363: Need to do this in case course was posted under a different subaccount / account previously

integrated_channels/canvas/client.py
```diff
@@ -67,6 +70,16 @@ def course_create_endpoint(canvas_base_url, canvas_account_id):
             canvas_account_id,
         )
 
+    @staticmethod
+    def course_fetch_endpoint(canvas_base_url, canvas_account_id):
+        """
+        Returns endpoint to GET to for course fetch
```
edx-abolger, line 76: nit: Can you make this docstring a bit more helpful?

integrated_channels/canvas/client.py
```diff
@@ -67,6 +70,16 @@ def course_create_endpoint(canvas_base_url, canvas_account_id):
...
+        """
+        return '{}/api/v1/accounts/{}/courses'.format(
```
edx-abolger, line 78: It looks like from here: https://canvas.instructure.com/doc/api/courses.html. that this API is an alias for the more straightforward GET /api/v1/courses/:id.   I would recommend you use this alias instead to keep the canvas account id out of the main logic and as part of the authentication setup only unless there is a good reason.

binodpant: ah yes good callout. The only issue is that we don't have access to the canvas course id :( until we make a course fetch. This is something we should maybe think of doing, perhaps storing the canvas ids on our side would be nice as an association record. But we have not really put a place/table to do that as of yet. e.g. we could store the canvas course id + edx course id association in a table if we wanted sth like

```
customer_lms_configuration_id | edx_course_id | external_lms_course_id
```

or sth like it maybe as a future idea

integrated_channels/canvas/client.py
binodpant: I did not implement this part yet because I want to hear from reviewers what they think of this. Basically doing this means that we won't possibly be able to support multiple canvas accounts per customer., because we would be reaching out to find courses under all other subaccounts too. Meaning effecitvely merging in all subaccounts as if. they are all one set of courses (or ignoring the subaccounts in some way).

Should we only support one subaccount per customer? Should we try to find courses under other subaccounts?

With customers we don't yet have an issue since we only use one subaccount per customer so far. But our own demo instance uses multiple subaccounts. If we assume a course can only be posted under one subaccount, then we don't need to do this and we can simply acknowledge a failure to create a course in that case (after checking response of course)

binodpant: Also, FYI it's easy to move course between subaccounts in Canvas (manually or via API), if needed

integrated_channels/canvas/client.py
```diff
@@ -183,28 +234,33 @@ def delete_course_completion(self, user_id, payload):  # pylint: disable=unused-
...
-            if "image_url" in content_metadata_item:
-                url = CanvasAPIClient.course_update_endpoint(
-                    self.enterprise_configuration.canvas_base_url,
-                    course_id,
-                )
-                # Providing the param `event` and setting it to `offer` is equivalent to publishing the course.
-                self._put(url, json.dumps({
-                    'course': {'image_url': content_metadata_item['image_url'], 'event': 'offer'}
-                }).encode('utf-8'))
+            url = CanvasAPIClient.course_update_endpoint(
```
binodpant, line 252: this refactor is here because we always want to set course to offer state, but the image_url detection of optional value is a separate concern, we don't want to mix the two

integrated_channels/canvas/client.py
```diff
@@ -273,18 +331,26 @@ def _extract_integration_id(self, data):
...
+          skip these courses in logic for create_or_update when needed.
+
+          Note: we do not need to follow pagination here since it would be extremely unlikely
+          that searching by a specific edx_course_id results in many records, we generally only
+          expect 1 record to come back anyway.
+
+          The `&state[]=all` is added so we can also fetch priorly 'delete'd courses
         """
-        url = "{}/api/v1/accounts/{}/courses/?search_term={}".format(
+        url = "{}/api/v1/accounts/{}/courses/?search_term={}&state[]=all".format(
```
binodpant, line 350: just later added this since without this param, 'deleted' courses will by default be excluded, and I have a clause elsewhere to specifically detect and log deleted courses

integrated_channels/canvas/client.py
alex-sheehan-edx: Do we need a course ID for that endpoint? I have a feeling that we can get all courses through `GET /api/v1/courses` but I'm not sure as to the scope of the courses returned. As in, if the `/api/v1/accounts/<account ID>/courses` endpoint actually limits to the sub-account or just returns all courses.

alex-sheehan-edx: side note, are you using this method anywhere?

alex-sheehan-edx: actually, assuming it's for https://github.com/edx/edx-enterprise/pull/1279/files#diff-bd39e800f794c671737477556e9d2d2e5855a2a4d5a83b7c764aba54ec49ece7R350 , I don't think we even need to worry about too many results since we're filtering by course title.

integrated_channels/canvas/client.py
```diff
@@ -78,22 +91,62 @@ def course_update_endpoint(canvas_base_url, course_id):
         )
 
     def create_content_metadata(self, serialized_data):
+        """
+        Creates a course in Canvas.
+        If course is not found ,easy!  create it as usual
```
alex-sheehan-edx, line 96: nit, I don't mind the `easy!` being there, but comma placement

integrated_channels/canvas/client.py
```diff
@@ -78,22 +91,62 @@ def course_update_endpoint(canvas_base_url, course_id):
...
+        If course found, it will be one of these workflow_states per doc /the current state
```
alex-sheehan-edx, line 97: `If course found, it will be one of these workflow_states per doc /the current state of the course one of 'unpublished', 'available', 'completed', or 'deleted'`

I'd maybe reword to something like
 
> If course is found, it will have one of the follow workflow_states: 
>                 `available`: issue an update with latest field values
>                 `completed`: this happens if a course has been concluded.
>                 ...
> for information of Canvas workflow_states see:
> https://canvas.instructure.com/doc/api/courses.html#method.courses.update

integrated_channels/canvas/client.py
alex-sheehan-edx: In with the convo going on with Adrienne's [comment](https://github.com/edx/edx-enterprise/pull/1279/files#r657913079), I'm wondering if we are able to hit the generic `GET /api/v1/courses` and get all courses in all sub-accounts. I think we should test out the functionality because it could simplify things.

integrated_channels/canvas/client.py
```diff
@@ -183,28 +236,33 @@ def delete_course_completion(self, user_id, payload):  # pylint: disable=unused-
         pass
 
     # Private Methods
-
-    def _update_course_details(self, course_id, serialized_data):
+    def _update_course_details(self, course_id, course_details):
         """
         Update a course for image_url (and possibly other settings in future)
-        This is used only for settings that are not settable in the initial course creation
+        Also sets course to 'offer' state by sending 'course[event]=offer'
```
alex-sheehan-edx, line 242: what does the offer state mean in this context? Does it have an effect on how we interact with the course?

integrated_channels/canvas/client.py
```diff
@@ -183,28 +236,33 @@ def delete_course_completion(self, user_id, payload):  # pylint: disable=unused-
...
+
+        Arguments:
+          - course_details (dict): { 'image_url' } : optional used if present to course[image_url]
```
alex-sheehan-edx, line 245: if you're going to add one of the params to the docstring I'd at least add the other even if it's straightforward as to not confuse anyone?

integrated_channels/canvas/client.py
```diff
@@ -183,28 +236,33 @@ def delete_course_completion(self, user_id, payload):  # pylint: disable=unused-
...
         """
+        response_code = None
+        response_text = None
+        # Providing the param `event` and setting it to `offer` is equivalent to publishing the course.
```
alex-sheehan-edx, line 249: I'd maybe include this in the docstring!

integrated_channels/canvas/client.py
```diff
@@ -183,28 +236,33 @@ def delete_course_completion(self, user_id, payload):  # pylint: disable=unused-
...
-            if "image_url" in content_metadata_item:
-                url = CanvasAPIClient.course_update_endpoint(
-                    self.enterprise_configuration.canvas_base_url,
-                    course_id,
-                )
-                # Providing the param `event` and setting it to `offer` is equivalent to publishing the course.
-                self._put(url, json.dumps({
-                    'course': {'image_url': content_metadata_item['image_url'], 'event': 'offer'}
-                }).encode('utf-8'))
+            url = CanvasAPIClient.course_update_endpoint(
```
alex-sheehan-edx, line 254: random nit, but the url generation is sitting in between two pieces of `update_payload` logic, maybe move this up above line 250 and keep things together?

alex-sheehan-edx (commented):
Couple of comments and questions! But looking pretty good 😄

integrated_channels/canvas/client.py
edx-abolger: It boooooothers :P

integrated_channels/canvas/client.py
```diff
@@ -78,22 +91,62 @@ def course_update_endpoint(canvas_base_url, course_id):
...
+            # Course does not exist: Create the course
+            status_code, response_text = self._post(self.course_create_url, serialized_data)
+            created_course_id = json.loads(response_text)['id']
+
+            # step 2: upload image_url and any other details
+            self._update_course_details(created_course_id, course_details)
+        else:
+            workflow_state = located_course['workflow_state']
+            if workflow_state.lower() == 'deleted':
+                LOGGER.warning(
```
edx-abolger, line 122: Can you please use the new generate_formatted_log function for new logs in this file? user id will be none, but other items can be filled in.

edx-abolger: Also can you raise this to "ERROR" ? If we have this it's going to cause sync problems for users, we want to know about it in the logs and possibly alert on it.

integrated_channels/canvas/client.py
edx-abolger: This could be resolved by adding a link to the relevant Canvas API.

integrated_channels/canvas/client.py
```diff
@@ -183,28 +236,33 @@ def delete_course_completion(self, user_id, payload):  # pylint: disable=unused-
...
+            url = CanvasAPIClient.course_update_endpoint(
+                self.enterprise_configuration.canvas_base_url,
+                course_id,
+            )
+            if "image_url" in course_details:
+                update_payload['course']['image_url'] = course_details['image_url']
+
+            response_code, response_text = self._put(url, json.dumps(update_payload).encode('utf-8'))
         except Exception:  # pylint: disable=broad-except
             # we do not want course image update to cause failures
```
edx-abolger, line 263: Please add an Error log here for alerting and tracking: I already have 2 different backlog tickets for bad course image URLs for other LMS systems, I don't want to add canvas to that list of problems I can't track.

integrated_channels/canvas/client.py
```diff
@@ -295,20 +361,38 @@ def _get_course_id_from_integration_id(self, integration_id):
...
+        # now let's try under all subaccounts
+        """url = "{}/api/v1/accounts/{}/courses/?search_term={}".format(
+            self.enterprise_configuration.canvas_base_url,
+            self.enterprise_configuration.canvas_account_id,
+            quote(edx_course_id),
+        )"""
+
+    def _get_course_id_from_edx_course_id(self, edx_course_id):
+        """
+        To obtain course ID we have to request all courses associated with the integrated
```
edx-abolger, line 380: This docstring scared me until I realized that it is not accurate. We do not get all courses and match ourselves, we call the remote API that does this. I would change this to say "To obtain the Canvas course ID, we have to use the Canvas API to search by edx_course_id."

edx-abolger (changes requested):

integrated_channels/canvas/client.py
binodpant: I hear you @alex-sheehan-edx on the GET /api/v1/courses, but I checked it again and it only returns more specicifc results, which is weird. The doc also says only for this user. Meaning only courses where the current user (as in, which user? tracked by dev token I think), is enrolled or somehow associated So I could not get it to return the deleted or other course as returned by the search endpoint for example

I agree it would be nice to just call a generic, get me all courses, endpoint, I have not found one! still going to confirm

binodpant: also checking this one https://canvas.instructure.com/doc/api/search.html#method.search.all_courses even though it says 'only courses in public index'

binodpant: and yes, it does return courses across subaccounts! so this resolves one of the issues. The only issue is it does_not_return that deleted course :( if we are to accept that we won't see deleted courses (without using that other search method) and just not want to fetch 'deleted' courses, we can use this one .. I tried {{canvas_url}}/api/v1/search/all_courses?per_page=1000 and I get courses from multiple accounts...


Also if we use this endpoint which may return a large no. of courses, we may need to call it once per `client` lifecycle rather than once per call to `create_content_metadata`

integrated_channels/canvas/client.py
binodpant: both make sense! will do

integrated_channels/canvas/client.py
binodpant: sure, it will now look like this on error

```
021-06-24 18:18:20,270 ERROR 55966 [integrated_channels.canvas.client] [user None] [ip None] client.py:278 - [integrated_channel:canvas][integrated_channel_enterprise_customer_uuid:873fec72-5a21-4f95-ac19-e3d67f9c4777][integrated_channel_lms_user:None][integrated_channel_course_key:edX+99887766]Failed to update details for course, edx_course_id=edX+99887766, canvas_course_id=6074. Details: a test only
```

integrated_channels/canvas/client.py
binodpant: there may be one possible way to resolve this


we fetch the root account id (or even just have it stored in canvas config) : I can also try to fetch by calling /api/v1/accounts and using the one with parent_account_id=null.

if we have the root account id, then searching against that id, instead of the subaccount id like I do now, seems to find the course under various subaccount ids!

alex-sheehan-edx: instead of fetching the root account ID, could we just ask the user to supply the root account ID within the configuration?

binodpant: So after going over various options, here is a final idea I have to resolve all of this more cleanly:

```
def find_course_in_canvas(edx_course_id):
  course = find_course_in_account(current_subaccount)
  if not course:
     log(..)
      root_account = find_root_account()
      course = find_course_in_account(root_account) # this may or may not work even
 if not course:
    log(...)
  return course
```

with this, we will still mostly find courses under current subaccount (the most common case), but we will hit the root_account case but if we fail to find a course under current subaccount we will scour at a larger scope under the root account

edx-abolger: @alex-sheehan-edx this config is already clearly too difficult for users as it is, I would much prefer we use the API call to get it if possible.

binodpant: alrightie! added code to also check under root account id

integrated_channels/canvas/client.py
```diff
@@ -358,33 +365,34 @@ def _extract_integration_id(self, data):
...
-            self.enterprise_configuration.canvas_account_id,
+            canvas_account_id,
             quote(edx_course_id),
         )
         resp = self.session.get(url)
         all_courses_response = resp.json()
 
         if resp.status_code >= 400:
+            message = ''
+            if 'reason' in all_courses_response:
```
binodpant, line 390: it seems that when we hit a 404 we don't get a reason back, we just get 'errors' as an array

binodpant:
there are a few test failures due to the changes to use a new utils file, will fix those up. A few mock changes are needed.

integrated_channels/canvas/client.py
```diff
@@ -78,22 +94,85 @@ def course_update_endpoint(canvas_base_url, course_id):
...
         integration_id = self._extract_integration_id(serialized_data)
-        course_id = self._get_course_id_from_integration_id(integration_id)
+        course_id = CanvasUtil.get_course_id_from_edx_course_id(
+            self.enterprise_configuration,
+            self.session,
+            integration_id,
+        )
 
         url = CanvasAPIClient.course_update_endpoint(
             self.enterprise_configuration.canvas_base_url,
```
mattdrayer, line 178: I recommend creating a getter for the `canvas_base_url` and ensuring that it appends a trailing slash to the end  of the value.  We definitely get burned on this from time to time, especially with SAPSF.

integrated_channels/canvas/utils.py
```diff
@@ -0,0 +1,173 @@
...
+
+from integrated_channels.exceptions import ClientError
+from integrated_channels.utils import generate_formatted_log
+
+LOGGER = logging.getLogger(__name__)
+
+
+class CanvasUtil:
+    """
+    A util to make various util functions related to Canvas easier and co-located.
```
mattdrayer, line 15: Sorry, what's the point of having a static object, versus this simply being a library of functions?  It's not like we have to have objects for everything in Python-land.

binodpant: yeah good point, I had started off as a regular class with non static methods (since I had thought to do the fetch of root acount in the constructor and hold on to it), but for simplicity for now, I decided to go stateless, and I ended up with all static methods, and never thought to switch away from the class.

alex-sheehan-edx (approved):
I have the same questions/concern over base url as @mattdrayer  but otherwise this is looking good to me!

integrated_channels/canvas/client.py
binodpant: sure! I will take a look after pr to see if we manage that or not. but good callout, I think stripping makes sense to be safe to manual errors, e.g.

binodpant: I can even do a micro pr for that

binodpant:
@edx-abolger lmk if any concerns/flags, hoping ready to merge and test in staging today!

integrated_channels/canvas/utils.py
binodpant: ever better, (in another PR), I think we want to use urllib.parse. 

```
>>> import urllib.parse
>>> urllib.parse.urljoin('http://test1','/abc')
'http://test1/abc'
>>> urllib.parse.urljoin('http://test1/','/abc')
'http://test1/abc'
>>> urllib.parse.urljoin('http://test1/','abc')
'http://test1/abc'
>>> urllib.parse.urljoin('http://test1','abc')
'http://test1/abc'
```

integrated_channels/canvas/client.py
binodpant: ever better, (in another PR), I think we want to use urllib.parse. 

```
>>> import urllib.parse
>>> urllib.parse.urljoin('http://test1','/abc')
'http://test1/abc'
>>> urllib.parse.urljoin('http://test1/','/abc')
'http://test1/abc'
>>> urllib.parse.urljoin('http://test1/','abc')
'http://test1/abc'
>>> urllib.parse.urljoin('http://test1','abc')
'http://test1/abc'
```

integrated_channels/canvas/client.py
```diff
@@ -78,22 +94,85 @@ def course_update_endpoint(canvas_base_url, course_id):
...
 
-        # step 2: upload image_url and any other details
-        self._update_course_details(created_course_id, serialized_data)
+        if not located_course:
+            # Course does not exist: Create the course
+            status_code, response_text = self._post(self.course_create_url, serialized_data)
+            created_course_id = json.loads(response_text)['id']
+
+            # step 2: upload image_url and any other details
+            self._update_course_details(created_course_id, course_details)
```
edx-abolger, line 125: Looks like we aren't even logging potential failures on this call. Can we at least add a log statement for failures here?

integrated_channels/utils.py
```diff
@@ -247,7 +247,6 @@ def generate_formatted_log(
     Arguments:
     - channel_name (str): The name of the integrated channel
     - enterprise_customer_uuid (str): UUID of the relevant EnterpriseCustomer
-    - enterprise_customer_identifer (str): Identifying name of the EnterpriseCustomer
```
edx-abolger, line 250: Thanks for the catch here!

edx-abolger (approved):
LGTM at this time.

integrated_channels/utils.py
binodpant: nw, I figured I would take care of this while I was here and the editor was pointing it out too, so great.

PR #1284, opened 2021-06-30
ENT-4666: Fixed issue with CourseEnrollment receiver when learner has multiple enterprises

irfanuddinahmad:
…enterprises

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/signals.py
+++ b/enterprise/signals.py
@@ -5,7 +5,6 @@ Django signal handlers.
 
 from logging import getLogger
 
-from django.core.exceptions import ObjectDoesNotExist
 from django.db.models.signals import post_delete, post_save, pre_delete, pre_save
 from django.dispatch import receiver
 
@@ -329,9 +328,8 @@ def create_enterprise_enrollment_receiver(sender, instance, **kwargs):     # pyl
     """
     if kwargs.get('created') and instance.user:
         user_id = instance.user.id
-        try:
-            ecu = EnterpriseCustomerUser.objects.get(user_id=user_id)
-        except ObjectDoesNotExist:
+        ecu = EnterpriseCustomerUser.objects.filter(user_id=user_id, active=True).first()
+        if not ecu:
             return
         logger.info((
             "User %s is an EnterpriseCustomerUser. "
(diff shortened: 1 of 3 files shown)


enterprise/signals.py
```diff
@@ -329,9 +328,8 @@ def create_enterprise_enrollment_receiver(sender, instance, **kwargs):     # pyl
     """
     if kwargs.get('created') and instance.user:
         user_id = instance.user.id
-        try:
-            ecu = EnterpriseCustomerUser.objects.get(user_id=user_id)
-        except ObjectDoesNotExist:
+        ecu = EnterpriseCustomerUser.objects.filter(user_id=user_id, active=True)
```
HammadAhmadWaqas, line 331: `ecu` is here a `queryset`, and later on we are using it  as `ecu .id` at [line](https://github.com/edx/edx-enterprise/pull/1284/files#diff-b4b6533b246e5794f1a7051c1bc5b1848746e918dad92f5ea5c6ad09ada2fdfdR342). I think code will break here.

HammadAhmadWaqas (changes requested):

HammadAhmadWaqas:
it would be great if can add test for it.

enterprise/signals.py
irfanuddinahmad: Good catch! ... fixed

irfanuddinahmad:
> it would be great if can add test for it.

Test is already present at https://github.com/edx/edx-enterprise/blob/f14a13ed9ef0dfe36b4ef8e5b2f0d47be996f1ea/tests/test_enterprise/test_signals.py#L698

zamanafzal (approved):

HammadAhmadWaqas (approved):
LGTM

edx-abolger:
@irfanuddinahmad, I see that there was a bug, but the solution here seems like it could create an Enterprise Course Enrollment for the WRONG enterprise if the wrong EnterpriseCustomer is chosen?  I think this may be a bigger bug than how you fixed it: what if the EnterpriseCourseEnrollment to create should actually be for the other course? (@iloveagent57 , is there an architectural workaround here? It seems like we might actually be missing the information in the signal that would get the right Enterprise selected, which would have to be passed in somehow from a browser session?)

PR #1287, opened 2021-06-30
feat: Add history tables for EnterpriseCustomerUser and SystemWideEnterpriseUserRoleAssignment

taliaedX:
***Background***
While attempting to troubleshoot issues around missing user roles, we struggled to get an accurate picture of the process and suspected (from limited log lines) that roles may have been created, but then later deleted (at least in some of the cases). Adding history tables to both ECU and role tables allows us to more accurately piece together what is happening from some of these signal initiated processes and be able to articulate actual solutions.

Due to errors with SimpleHistory, this PR also changes the RoleAssignment table's `role_class` field to explicitly `role` and changes it to a foreign key. It being assigned a class, but never explicitly made a field was causing it to be missed by the SimpleHistory class instantiation and the generated migration. For the base class, it was inheriting from an edx-rbac class which converts the `role_class` variable into a foreign key - thus we're just skipping this step (and it handles `role` already existing in edx-rbac as well. See here for the exact code I'm now bypassing: https://github.com/edx/edx-rbac/blob/master/edx_rbac/models.py#L24)

***Changes***
- Adds history tables for EnterpriseCustomerUser and SystemWideEnterpriseUserRoleAssignment
- Updates SystemWideEnterpriseUserRoleAssignment to just define the role as a foreign key instead of relying on edx-rbac to do that exact same thing.

**Merge checklist:**

**Post merge:**


enterprise/migrations/0136_auto_20210629_2129.py
```diff
@@ -0,0 +1,70 @@
...
+                ('modified', model_utils.fields.AutoLastModifiedField(default=django.utils.timezone.now, editable=False, verbose_name='modified')),
+                ('applies_to_all_contexts', models.BooleanField(default=False, help_text='If true, indicates that the user is effectively assigned their role for any and all contexts. Defaults to False.')),
+                ('history_id', models.AutoField(primary_key=True, serialize=False)),
+                ('history_date', models.DateTimeField()),
+                ('history_change_reason', models.CharField(max_length=100, null=True)),
+                ('history_type', models.CharField(choices=[('+', 'Created'), ('~', 'Changed'), ('-', 'Deleted')], max_length=1)),
+                ('enterprise_customer', models.ForeignKey(blank=True, db_constraint=False, help_text='The EnterpriseCustomer for which the role is assigned for the user.  If creating a new assignment, you must first provide a user email and role, then click "Save and continue editing" BEFORE selecting from this dropdown.', null=True, on_delete=django.db.models.deletion.DO_NOTHING, related_name='+', to='enterprise.EnterpriseCustomer')),
```
kiram15, line 35: If you wanted to make this line more readable, you could save the help_text as a variable earlier

enterprise/migrations/0136_auto_20210629_2129.py
```diff
@@ -0,0 +1,70 @@
...
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='systemwideenterpriseuserroleassignment',
+            name='role',
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='system_wide_role_assignments', to='enterprise.SystemWideEnterpriseRole'),
+        ),
+        migrations.CreateModel(
+            name='HistoricalSystemWideEnterpriseUserRoleAssignment',
```
kiram15, line 25: Do we add in any additional tests for history models (doubtful)? and is there a way you could mimic the old behavior in a test to prove in this PR that the changes will fix it?

enterprise/migrations/0136_auto_20210629_2129.py
taliaedX: This is just what django generates. It's a good suggestion in general, but I think for a migration it's not super necessary. No one is looking at this for the help text lol.

enterprise/migrations/0136_auto_20210629_2129.py
taliaedX: To answer your first question: We don't have any existing tests. It'd probably be possible to verify the model gets instantiated when the original is saved, but mostly we'd be testing that simple history does what it's supposed to and that's what the tests they presumably have in their own repo should be doing. 😄 

For the second: Not really. For a few reasons:
- Now that we have a history table, we can't really mimic what it was like without it.
- Tests that look for the absence of something are often pretty problematic as I tend to question if it's correctly not finding the object or if it's just that test conditions wouldn't have generated it anyways (this is especially true as you get into mocks and frontend code. I have many, many times found my test looking for an absence is only functioning because the positive case is also not working 😞 ). You typically want to format your tests in a way that they can look for something's existence (usually we look for exceptions and I guess I could custom make one to detect this but that might get pretty complicated)
- [extremely personal opinion]: I don't like adding tests to detect an issue that I've already fixed. Unless it's a major issue (ie if we pushed code that would have otherwise failed this new test and it breaks production), or it's a particular utility/method/library that is used in a variety of places, a hyper specific test just adds time to our build process for no particular benefit.

kiram15 (approved):

enterprise/models.py
```diff
@@ -2688,7 +2689,11 @@ class SystemWideEnterpriseUserRoleAssignment(EnterpriseRoleAssignmentContextMixi
     .. no_pii:
     """
 
-    role_class = SystemWideEnterpriseRole
+    role = models.ForeignKey(
```
alex-sheehan-edx, line 2692: the FK method is one I'm a little more familiar with than (what seems to be?) assigning the object to the field, could I get a tl:dr on what the `role_class` was doing before this?

alex-sheehan-edx (approved):
one none blocking q for my own understanding but otherwise LGTM

enterprise/models.py
alex-sheehan-edx: oh just saw your description- nvm 😄

taliaedX: ah this was answered via dm, but role_class is a field value picked up by the base class (class inheritance ultimately leads back to here: https://github.com/edx/edx-rbac/blob/master/edx_rbac/models.py#L24).
When role_class is present, it adds a field named `role` to the model that is a foreign key to the model associated with the value of role_class here. In much simpler verbiage: it does exactly what I'm doing in this pr. When role is present, it does nothing, because it doesn't need to add a foreign key field where we already have that exact foreign key field.

PR #1289, opened 2021-07-01
feat: Add segment tracking for bulk enrollment

taliaedX:
***Background***
This adds bulk enrollment specific segment tracking to the enrollment action that occurs only via bulk enrollment.

***Screenshot***
Screen Shot 2021-07-01 at 3 33 31 PM

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -64,6 +64,7 @@ from enterprise.utils import (
     get_best_mode_from_course_key,
     get_ecommerce_worker_user,
     get_request_value,
+    track_enrollment,
     validate_email_to_link,
 )
 from enterprise_learner_portal.utils import CourseRunProgressStatuses, get_course_run_status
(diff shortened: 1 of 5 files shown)


enterprise/api/v1/views.py
```diff
@@ -313,6 +314,7 @@ def enroll_learners_in_courses(self, request, pk):
                 result.pop('user') for result in results['successes'] if result['course_run_key'] == course_run
             }
             LOGGER.info("Successfully bulk enrolled learners: {}".format(pending_users | existing_users))
+            track_enrollment('customer-admin-enrollment', request.user.id, course_run)
```
binodpant, line 317: if possible we should add a test for this tracking call too

binodpant (approved):
well yeah this one looks nice and easy! yeah let's find out more about the source field requirements for pending and regular enrollments. but this meets the criteria for AC1 in the ticket so I think we can go in. Just made a suggestion about unit test.

enterprise/api/v1/views.py
```diff
@@ -312,14 +313,16 @@ def enroll_learners_in_courses(self, request, pk):
...
             }
-            LOGGER.info("Successfully bulk enrolled learners: {}".format(pending_users | existing_users))
-            if serializer.validated_data.get('notify'):
-                enterprise_customer.notify_enrolled_learners(
-                    catalog_api_user=request.user,
-                    course_id=course_run,
-                    users=pending_users | existing_users,
-                    admin_enrollment=True,
-                )
+            if len(pending_users | existing_users) > 0:
```
alex-sheehan-edx, line 316: Assuming this is to prevent the empty log lines, nice!

alex-sheehan-edx (approved):
nice tests, lgtm

enterprise/api/v1/views.py
taliaedX: yes that is this bit!

PR #1299, opened 2021-07-14
Updating bulk enrollment email template

kiram15:
Screenshots for email differences are posted in original ticket [here](https://openedx.atlassian.net/browse/ENT-4653?atlOrigin=eyJpIjoiOWVkYmRkM2FkNTFmNDhiNWIwMTUyYmQ5ZjgyZTEwOTEiLCJwIjoiamlyYS1zbGFjay1pbnQifQ). 

**Merge checklist:**

**Post merge:**


enterprise/migrations/0133_auto_20210608_1931.py
```diff
@@ -50,11 +47,8 @@ def create_templates(apps, schema_editor):
     EmailTemplate.objects.get_or_create(
         plaintext_template="""
         Great News! You've been enrolled in {{enrolled_in.name}} by {{organization_name}}
-
         This course is a free benefit offered especially for you, and we are excited for you to meet your learning community on edX.
-
         Visit this link to see and enroll in your course, {{enrolled_in.url}}
-
```
taliaedX: do we no longer want the spaces? 🤔

enterprise/migrations/0137_enrollment_email_update.py
```diff
@@ -0,0 +1,230 @@
...
+    )
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('enterprise', '0136_auto_20210629_2129'),
+    ]
+
+    operations = [
+        migrations.RunPython(code=update_email, reverse_code=update_email)
```
kiram15, line 228: The reason I have the reverse_code calling the same method is that I didn't see the need to paste the old HTML in this one, because no functionality will change based up if this one is migrated back. However, I didn't want a migration that could be rolled back.

enterprise/migrations/0137_enrollment_email_update.py
```diff
@@ -0,0 +1,230 @@
+from logging import getLogger
+from django.db import migrations
+
+LOGGER = getLogger(__name__)
+
+def update_email(apps, schema_editor):
```
taliaedX, line 6: why not just do this in Django Admin?
IE we shouldn't need to be updating templates anyways...
I get for initial population, but we're adjusting an already existing one. No need to deal with migration headaches for that.

admin/enterprise/enrollmentnotificationemailtemplate/ is the path by the way. ie (`http://localhost:18000/admin/enterprise/enrollmentnotificationemailtemplate/` or courses-internal.stage.edx.org/admin/enterprise/enrollmentnotificationemailtemplate/)

enterprise/models.py
```diff
@@ -685,15 +685,16 @@ def notify_enrolled_learners(self, catalog_api_user, course_id, users, admin_enr
...
             'LMS_ROOT_URL',
             settings.LMS_ROOT_URL
         )
-        destination_url = '{site}/{login_or_register}?next={course_path}'.format(
-            site=lms_root_url,
-            login_or_register='{login_or_register}',  # We don't know the value at this time
+        enterprise_url = '{site}{course_path}'.format(
+            site=get_configuration_value(
+                'ENTERPRISE_LEARNER_PORTAL_BASE_URL',
+                settings.ENTERPRISE_LEARNER_PORTAL_BASE_URL
```
taliaedX, line 696: what if they don't have the learner portal enabled?

taliaedX (commented):
I have some questions around the URL and the necessity of a migration here.

enterprise/migrations/0137_enrollment_email_update.py
taliaedX: (This shouldn't really be a task for developers) given we actually expose these templates

kiram15: This makes sense (and how I was doing it initially when testing), but I figured for it to be updated on everyone's local it would be better to make a migration and tests? But I can also scratch this entirely and just post the html into stage and prod

binodpant: I would say any further updates to templates can be done in Django admin. Not as nice as having a git history but there is Django history too I think

enterprise/models.py
binodpant: good point, le'ts get clarify on that case of learner poratl off. I think it's going to be the B2C link instead which is similar to https://learning.edx.org/course/course-v1:MichiganX+py4e101x+3T2020/home?utm_source=braze&utm_medium=email&utm_campaign=enrollmentconfirmation (this example is from a prior Sailthru email I got on enrollment, for reference only)

binodpant: idk about all those utm params don't think we need those for this use case. Braze will handle separately later.. maybe best to confirm with enterprise-engg or titans folks what url makes sense here... the simplest with no adornment is of the form `https://learning.edx.org/course/course-v1:MichiganX+py4e101x+3T2020/`

enterprise/models.py
```diff
@@ -711,15 +712,16 @@ def notify_enrolled_learners(self, catalog_api_user, course_id, users, admin_enr
         with mail.get_connection() as email_conn:
             for user in users:
                 login_or_register = 'register' if isinstance(user, PendingEnterpriseCustomerUser) else 'login'
-                destination_url = destination_url.format(login_or_register=login_or_register)
+                enterprise_url = enterprise_url.format(login_or_register=login_or_register)
```
binodpant, line 715: Re: the Start My Course link: For me two questions are pertinent:
learner portal on/off:
pending vs pre-existing edx user
The learners in the bulk enroll set can be in either of these boats?
For ref: the enterprise-catalog has a piece of logic to determine enrollment_url which is used in Integrated channels for example. https://github.com/edx/enterprise-catalog/blob/3ad56d1bbad03c86123f74f7e67fcc6275a40238/enterprise_catalog/apps/catalog/models.py See get_content_enrollment_url . This one has logic to check learner portal on/off we can model from? But these urls are for enrollment. Hence I am unsure about that second question from above. Maybe Titans folks have more context too? (edited)

tests/test_utilities.py
```diff
@@ -607,6 +613,7 @@ def test_send_email_notification_message_with_site_defined_values(
         Test that we can successfully render and send an email message.
         """
         enrolled_in['start'] = datetime.datetime.strptime(enrolled_in['start'], '%Y-%m-%d')
+        dashboard = 'http://lms.example.com'
```
binodpant, line 616: nit: my pref for this var name would be dashboard_url

enterprise/models.py
```diff
@@ -711,15 +712,16 @@ def notify_enrolled_learners(self, catalog_api_user, course_id, users, admin_enr
...
                 utils.send_email_notification_message(
                     user=user,
                     enrolled_in={
                         'name': course_name,
-                        'url': destination_url,
+                        'url': enterprise_url,
```
binodpant, line 720: So will your change keep the functionality intact for cases that aren't bulk enrollment (if you search for usages of this method without the admin_enrollment=True)? The only other case is the self_enroll case ... I wonder if the `url` that was used before is still applicable to that case.

If we don't want to have to deal with too many changes or not affect those other cases, we can also not use the `url` param and instead use a new `enterprise_course_url` which is only used in the bulk enrollment template, which leaves the other use cases untouched.

enterprise/models.py
```diff
@@ -676,27 +676,42 @@ def notify_enrolled_learners(self, catalog_api_user, course_id, users, admin_enr
...
+                'ENTERPRISE_LEARNER_PORTAL_BASE_URL',
+                settings.ENTERPRISE_LEARNER_PORTAL_BASE_URL
+            )
+            destination_url = '{site}/{login_or_register}?next=/{slug}/{course_path}'.format(
+                site=dashboard_url,
+                login_or_register='{login_or_register}',  # We don't know the value at this time
+                slug=self.slug,
+                course_path=course_path
+            )
+        else:
```
binodpant, line 694: FYI I had suggested to specialize this url for enterprise to bulk enrollment case only sice we don't know how the other use cases are expecting urls to be. and we don't want to affect those right now

tests/test_utilities.py
```diff
@@ -679,6 +689,7 @@ def test_send_email_notification_message_with_site_from_email_override(
         utils.send_email_notification_message(
             user,
             enrolled_in,
+            dashboard,
```
binodpant, line 692: for consistently I would also change these vars to dashboard_url but otherwise lgtm!

enterprise/models.py
binodpant: SUGGEST: Add a comment on the idea that we currently only support bulk enroll for subs which means we can assume Learner portal is enabled for the enterprise, somewhere here

binodpant (approved):
Made a couple of suggestions, but lgtm

PR #1301, opened 2021-07-16
fix: SAP: use separate endpoint and user for certain customers

taliaedX:
### Background:
- At least one customer's SAP instance does not allow learner grades to be submitted "by" the learner.
- Our current functionality submits the grade with a user id of the learner (aka submits the grades by the learner) for all requests.

Thus, this needs to be changed, potentially for all customers, but for now we want it to be conditional on a flag.

The specific logic (please review @mattdrayer) is:
- When the `prevent_self_submit_grades` flag is False (which it will be for all existing customers) we use the existing functionality and endpoint which is to submit grades *as* the learner and using the `learning/public-api/rest/v1/current-user/item/learning-event` endpoint.
- When this flag is *True*, we use a different endpoint (`learning/odatav4/public/admin/learningevent-service/v1/OCNLearningEvents`) and send with the base `sapsf_user_id` (which is defined in the customer's enterprise configuration)

### Code Changes in this PR
- Adds a new Boolean field to the SAPSuccessFactorsEnterpriseConfiguration model called `prevent_self_submit_grades`
- Adds this as an editable field to Django Admin.
- Defaults the field to *FALSE*. (see the next section about this because I know this differs from the ticket)
- Adds a new test verifying the logic swapping to the new url
- Modifies an existing test to have it verify the old url is still the one we're using.

### What this PR does NOT do
- Does not do the False/True shuffle as requested in the ticket.
Setting the default value of a new field in Django also updates existing records. Thus to be able to accomplish what the ticket is asking we'd end up updating all records twice. (Django Boolean Fields are not allowed to be Nullable, which is the only way I thought I could make this logic work without multiple migrations)

It's not that much on paper, but it's an extra migration for really little benefit that I can see:
- from the ticket, we are not clear if this approach will totally work.
- It's also not clear we'll want this for every customer down the line.
- even if we do, at that point this field should just be removed and the logic updated

**Merge checklist:**

**Post merge:**


integrated_channels/sap_success_factors/client.py
```diff
@@ -139,8 +141,19 @@ def create_course_completion(self, user_id, payload):
...
             HTTPError: if we received a failure response code from SAP SuccessFactors
         """
-        url = self.enterprise_configuration.sapsf_base_url + self.global_sap_config.completion_status_api_path
-        return self._call_post_with_user_override(user_id, url, payload)
+        base_url = self.enterprise_configuration.sapsf_base_url
+
+        if self.enterprise_configuration.prevent_self_submit_grades:
+            return self._call_post_with_session(
+                base_url + self.GENERIC_COURSE_COMPLETION_PATH,
+                payload
```
mattdrayer, line 149: Are we certain that the payload schema is the exact same for these two requests?

integrated_channels/sap_success_factors/client.py
```diff
@@ -28,6 +28,8 @@ class SAPSuccessFactorsAPIClient(IntegratedChannelApiClient):  # pylint: disable
 
     SESSION_TIMEOUT = 5
 
+    GENERIC_COURSE_COMPLETION_PATH = 'learning/odatav4/public/admin/learningevent-service/v1/OCNLearningEvents'
```
mattdrayer, line 31: any particular reason we are specifying this here and not in `global_sap_configuration`?

integrated_channels/sap_success_factors/client.py
taliaedX: Well the old endpoint's documentation doesn't appear to exist, but our included fields are named the same as the new API's documented naming scheme. There are some additional fields in the API's endpoint documentation such as instructorName that I cannot tell are optional or not (it doesn't say optional or required).

I will not really have 100% confidence in the fields without being able to hit the API unfortunately.

integrated_channels/sap_success_factors/client.py
taliaedX: 1. The global config as a whole just feels like an unnecessary object. To store in a database allows us to track changes, but if API endpoints change, it's also extremely likely we'd need to change our code to work with those anyways. Thus I don't really want API endpoints editable in Django Admin. We could expose it in the admin screen but without being able to be edited, but we should just add that to any documentation ECS uses.
2. We decided against the global configs in all of our new integrated channels (see above). I'd love to get rid of the existing endpoints, but don't really go through the refactor at this moment. If this whole thing works out well and we end up wanting to remove the existing course completion url, I'll probably refactor at that point.

integrated_channels/sap_success_factors/client.py
taliaedX: so yeah the existing attribute names are the same and I am not sure if instructorName, cpeHours, comments, price, etc. were existing or not. Their documentation does say that userID is required, so maybe if the rest were actually required, they would also mention that in the description of the fields. I hope?

integrated_channels/sap_success_factors/client.py
taliaedX: Oh I forgot reason 3 and sort of 4: by making it a constant we remove an unnecessary (IMO) database call. And (4) a database table that is only ever expected to have a single row just should not exist. I mean our table doesn't because it also acts as a history/audit table, but that has also caused problems in other instances and I'm also not a fan of it.

mattdrayer: Haha, all good and I agree with all points made -- thanks!

integrated_channels/sap_success_factors/client.py
mattdrayer: Through my own testing with our SAP tenant which is sort of still accessible via Postman I *think* I figured out that we actually have to specify more fields in this new request vs. the old one.  This is what I ended up with as a payload:
`{
"userID": "ChanA",
"courseID":"HarvardX+CS50x1",
"providerID": "EDX",
"courseCompleted":true,
"completedTimestamp":1538766064479,
"instructorName":"Drayer",
"grade":"Pass",
"totalHours":40,
"creditHours":10,
"contactHours":8,
"cpeHours":2,
"comments":"Testing new OCN API completion events endpoint",
"price":0.00,
"currency":"USD"
}
`

But this is tedious to test/confirm because you only get one shot at creating a learning history record and then you have to pick a new course.  I'm going to spend a couple minutes seeing if I can determine which fields are required more concretlely.

mattdrayer: OKAY, I managed to figure out a reasonably straightforward way to test this using our BestRun tenant, which is STILL available for our use via Postman (crazy).  I've confirmed that the minimum fields we need to send in order to record a completion event are:

```
{
"completedTimestamp":1538766064479,
"courseCompleted":true,
"userID": "ChanA",
"courseID":"MITx+3.086x",
"providerID": "EDX"
}
```

It seems that everything else is nice to have but not necessarily critical, at least for our tenant.

mattdrayer: The response I receive when I submit this very basic payload looks like this:

```
{
    "@odata.context": "$metadata#OCNLearningEvents/$entity",
    "@odata.metadataEtag": "W/\"e51c4bea-1e33-44f2-97ef-6ca1c39eb9f0\"",
    "userID": "ChanA",
    "courseID": "MITx+3.086x",
    "providerID": "EDX",
    "courseCompleted": true,
    "completedTimestamp": 1538766064479,
    "instructorName": null,
    "grade": null,
    "totalHours": 0.0,
    "creditHours": 0.0,
    "contactHours": 0.0,
    "cpeHours": 0.0,
    "comments": null,
    "price": 0.0,
    "currency": null
}
```

mattdrayer (approved):
Looks good to me!

integrated_channels/sap_success_factors/client.py
taliaedX: Interesting. I was under the impression our instance was completely inaccessible. Had I known otherwise, could have tested instead of making you do so. Thank you for that confirmation though!

So this confirms the data and format we're sending is still fine. I will double-check handling of the response though.

taliaedX: yeah we're not really doing anything with the response unless it's a non-200 status code.

alex-sheehan-edx (approved):
LGTM 👍

integrated_channels/sap_success_factors/client.py
mattdrayer: So, a small discrepancy that we overlooked during this conversation, and that we are now seeing in a new error message being returned from the SAPSF tenants:  We are sending the value for `courseCompleted` as `"true"` with quotes, however it seems for the new endpoint the value must be sent as `true` without quotes.  

This was a subtle breaking change in the endpoint schema from the previous version that I must have quickly revised during my testing and didn't realize was actually a breaking change -- ie, I probably thought that I just had specified it wrong the first time I tried to submit a payload to the new endpoint.

@binodpant @alex-sheehan-edx we can talk about it more at our upcoming integrations meeting.

mattdrayer: Error message being captured in our audit logs:  `{"error":{"code":null,"message":"Invalid value for property 'courseCompleted'."}}`

taliaedX: I can try to get it to format that correctly. Something about the 2 methods we use for post are formatting it differently.

taliaedX: nvm. It seems the new endpoint just expects actual booleans whereas the old one didn't.
And we're doing that override in the base integrated channels class.
https://github.com/edx/edx-enterprise/blob/master/integrated_channels/integrated_channel/models.py#L262

This might be a little trickier to override. Will take a look

taliaedX: and yet also here:
https://github.com/edx/edx-enterprise/blob/master/integrated_channels/sap_success_factors/models.py#L301

So we're overriding the payload data that - judging by the comment - was written for SAP.
Ugh

PR #1302, opened 2021-07-19
bug: fixing client to handle paginated responses from Blackboard

alex-sheehan-edx:
[Jira](https://openedx.atlassian.net/browse/ENT-4635)

**Background:**
Through some investigations of the API documentation, we discovered that BB does indeed paginate their responses. In testing with our local instance it seems like the default size for an instance is `200` but it is unclear if this is customizable by the customer, so to be safe all sizes should be taken into account. Before this PR, if the number of assignments within a course got too high, we would not be able to verify the existence of further assignments. This would trigger our gradebook column (Blackboard's representation of our assignments) creation process, which would fail with a unique key integrity error. As such all further grade reporting would fail to be reported. 

**Code Changes in this PR**
- The Blackboard client now checks for the existence of a `next page` within the API response
- The Blackboard client will request next pages from the customer's instance until it:
    -  a) finds the assignment gradebook column
    -  b) exhausts all pages from the API
    - c) hits the page limit (250)
- Unit tests that cover the pagination traversal

**What this PR does NOT do**
The [Jira](https://openedx.atlassian.net/browse/ENT-4635) ticket specifies that we have transmitted duplicate assignments as a result of the lack of pagination handling and thus all customer instances need to be sanitize of duplicate assessment reporting (as Canvas needed recently). Through testing I have determined that this is improbable as Blackboard, unlike Canvas, has unique key restraints on the gradebook columns making it impossible to retransmit assignments with the same `externalID`. Thus any deduplication work has been left out of this PR.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,18 +18,21 @@ Unreleased
 * Nothing
 
 
+[3.27.4]
+--------
+* Add pagination handling to integrated channels Blackboard client
+
 [3.27.3]
----------
+--------
 * Adds flag to SAP Success Factors customer configuration to switch SAP endpoints for learner completion calls.
 
 [3.27.2]
----------
-
+--------
 * Ensure deletion and unlinking of a ``EnterpriseCustomerUser`` record only deletes the ``enterprise_learner`` system-wide role for that
   particular ``EnterpriseCustomerUser``, as opposed to all ``enterprise_learner`` roles associated with the user.
 
 [3.27.1]
----------
+--------
 * Updates bulk enrollment email template.
 
 [3.27.0]
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.27.3"
+__version__ = "3.27.4"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/integrated_channels/blackboard/client.py
+++ b/integrated_channels/blackboard/client.py
@@ -31,6 +31,8 @@ COURSES_V3_PATH = '/learn/api/public/v3/courses'
 COURSE_CONTENT_PATH = '/learn/api/public/v1/courses/{course_id}/contents'
 COURSE_CONTENT_CHILDREN_PATH = '/learn/api/public/v1/courses/{course_id}/contents/{content_id}/children'
 COURSE_CONTENT_DELETE_PATH = '/learn/api/public/v1/courses/{course_id}/contents/{content_id}'
+GRADEBOOK_COLUMN_DESC = "edX learner's grade"
+PAGE_TRAVERSAL_LIMIT = 250
 
 
 class BlackboardAPIClient(IntegratedChannelApiClient):
@@ -354,6 +356,33 @@ class BlackboardAPIClient(IntegratedChannelApiClient):
         """
         return str(abs(hash(external_id)))
 
+    def generate_blackboard_gradebook_column_data(self, external_id, grade_column_name, points_possible):
+        """
+        Properly formatted json data to create a new gradebook column in a blackboard course
+
+        Note: Potential customization here per-customer, if the need arises.
+        """
+        return {
+            "externalId": external_id,
+            "name": grade_column_name,
+            "displayName": grade_column_name,
+            "description": GRADEBOOK_COLUMN_DESC,
+            "externalGrade": False,
+            "score": {
+                "possible": points_possible
+            },
+            "availability": {
+                "available": "Yes"
+            },
+            "grading": {
+                "type": "Manual",
+                "scoringModel": "Last",
+                "anonymousGrading": {
+                    "type": "None",
+                }
+            },
+        }
+
     def generate_gradebook_url(self, course_id):
         """
         Blackboard API url helper method.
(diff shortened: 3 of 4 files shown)


enterprise/__init__.py
```diff
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.27.1"
+__version__ = "3.27.2"
```
taliaedX, line 5: womp

CHANGELOG.rst
```diff
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * Nothing
 
+[3.27.2]
```
taliaedX, line 20: 3.27.3 now. 😅

integrated_channels/blackboard/client.py
```diff
@@ -525,37 +552,49 @@ def _get_or_create_integrated_grade_column(self, bb_course_id, grade_column_name
...
+            #         ...
+            #     }, {
+            #         'id', 2,
+            #         ...
+            #     }],
+            #     'paging': {
+            #         'nextPage': '/learn/api/public/v1/courses/<courseID>/gradebook/columns?offset=200'
+            #     }
+            # }
+            if parsed_response.get('paging') and not grade_column_id:
```
taliaedX, line 584: can you ELI5 why grade_column_id not existing affects whether we check for more pages or not. I did read the description, but I may be too hazy to comprehend it if it was mentioned there.

integrated_channels/blackboard/client.py
```diff
@@ -525,37 +552,49 @@ def _get_or_create_integrated_grade_column(self, bb_course_id, grade_column_name
...
+                gradebook_column_url = '{}{}'.format(
+                    self.enterprise_configuration.blackboard_base_url,
+                    parsed_response.get('paging').get('nextPage'),
+                )
+            else:
+                more_pages_present = False
+
+            current_page_count += 1
```
taliaedX, line 592: is there a point in incrementing this if we set `more_pages_present = False`
To state another way, why not stick this in the `if` statement above (and I mean the one right above this Else. The other one I understand is within a For loop)

integrated_channels/blackboard/client.py
taliaedX: Is it that we go through pages of results until we get a grade_column_id and then we stop going through the pages?

taliaedX (commented):
I mostly have questions for my understanding! Thus just commenting for now, but can revisit after that and when maybe my brain functions better if you still need an additional approver.

integrated_channels/blackboard/client.py
```diff
@@ -354,6 +354,33 @@ def generate_blackboard_course_id(self, external_id):
         """
         return str(abs(hash(external_id)))
 
+    def generate_blackboard_gradebook_column_data(self, external_id, grade_column_name, points_possible):
```
edx-abolger, line 357: nit: If you don't use 'self', make this a static or class method

CHANGELOG.rst
alex-sheehan-edx: 😡

enterprise/__init__.py
alex-sheehan-edx: 👍

integrated_channels/blackboard/client.py
alex-sheehan-edx: Yeah the idea is that we either go through every page or we find the ID on a particular page. Blackboard will only include the `paging` response if there exists another page, so that line is saying "if we haven't found ID yet and if there are more pages available".

integrated_channels/blackboard/client.py
alex-sheehan-edx: TBH you're right, there's no need to increment for both cases. If we have no more pages present then there's no point in incrementing the page count. It doesn't necessarily break anything to have it there, but for the sake of readability I can change.

taliaedX (approved):
LGTM! Understand if you want to wait until Adrienne reviews though

tests/test_integrated_channels/test_blackboard/test_client.py
```diff
@@ -685,3 +685,125 @@ def test_content_child_customization_failure(self):
...
+                json=blackboard_page_two_response
+            )
+            rsps.add(
+                responses.POST,
+                client.generate_create_grade_column_url(self.blackboard_course_id),
+                json={'id': self.blackboard_grade_column_id}
+            )
+
+            # traverse over all pages, post a new column if column isn't found
+            column_id = client._get_or_create_integrated_grade_column(  # pylint: disable=protected-access
```
edx-abolger, line 803: Can you add some kind of test here that we call the pages the correct number of times? I don't see a test here that actually confirms that we make that call to page 2?

integrated_channels/blackboard/client.py
```diff
@@ -354,6 +354,33 @@ def generate_blackboard_course_id(self, external_id):
...
+        """
+        Properly formatted json data to create a new gradebook column in a blackboard course
+
+        Note: Potential customization here per-customer, if the need arises.
+        """
+        return {
+            "externalId": external_id,
+            "name": grade_column_name,
+            "displayName": grade_column_name,
+            "description": "edX learner's grade.",
```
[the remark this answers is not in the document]
alex-sheehan-edx: The field isn't displayed on the immediate UI and is more of a body div type display so I thought it lent itself to a full sentence structure. That being said I also see it being completely unnecessary and differently patterned from most of the other content we display so I'll remove it.

tests/test_integrated_channels/test_blackboard/test_client.py
alex-sheehan-edx: `responses.RequestsMock` by default asserts that all mock created within the context must be executed before leaving. If we didn't make the call to page 2 then the test would fail with `AssertionError: Not all requests have been executed [('GET', 'https://base.url/learn/api/public/v1/courses/723/gradebook/columns')]`

edx-abolger: Sweet, TIL!

edx-abolger (approved):

PR #1307, opened 2021-07-23
feat: check for and only show enabled IDPs in Django Admin

taliaedX:
## Background
We want to ensure only certain IDPs are selectable for enterprise customers. The particular location we want to change is the Django Admin Enterprise Customer form where you can select an Identity Provider.
With the changes in this PR, if an IDP is disabled for use in enterprise, it will not be visible in the dropdown.

## Code Changes
- Adds conditional logic to the `get_idp_choices` method to utilize the new flag on the ProviderConfig model.
- Adds a test to actually test this method (it _kind_ of exists in the `test_forms` file, but not really testing the logic of the method).

## Testing Instructions
- In order to test this you need the branch for PR https://github.com/edx/edx-platform/pull/28261 locally
- Apply the migration from that PR
- Pick or make a provider (search for `Provider Configuration` in Django Admin and you'll find the 3 types under Third Party Auth. Pick any type, add or modify a provider) and check the box for "Disabled for Enterprise TPA" on the provider form.
- Verify this provider is not in the list of selectable providers in the Identity Provider section of an Enterprise Customer's details screen.
- Turn the setting back off and verify the provider _is_ selectable.

### Provider Config Admin Form
Screen Shot 2021-07-23 at 11 32 04 AM
### Enterprise IDP Selection
Screen Shot 2021-07-23 at 11 32 26 AM

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -257,14 +257,6 @@ def get_identity_provider(provider_id):
         Instance of ProviderConfig or None.
     """
     try:
-        # pylint: disable=redefined-outer-name,import-outside-toplevel
-        from common.djangoapps.third_party_auth.provider import Registry
-    except ImportError as exception:
-        LOGGER.warning("Could not import Registry from common.djangoapps.third_party_auth.provider")
-        LOGGER.warning(exception)
-        Registry = None  # pylint: disable=redefined-outer-name
-
-    try:
         return Registry and Registry.get(provider_id)
     except ValueError:
         return None
@@ -277,17 +269,10 @@ def get_idp_choices():
     Return:
         A list of choices of all identity providers, None if it can not get any available identity provider.
     """
-    try:
-        # pylint: disable=redefined-outer-name,import-outside-toplevel
-        from common.djangoapps.third_party_auth.provider import Registry
-    except ImportError as exception:
-        LOGGER.warning("Could not import Registry from common.djangoapps.third_party_auth.provider")
-        LOGGER.warning(exception)
-        Registry = None  # pylint: disable=redefined-outer-name
 
     first = [("", "-" * 7)]
     if Registry:
-        return first + [(idp.provider_id, idp.name) for idp in Registry.enabled()]
+        return first + [(idp.provider_id, idp.name) for idp in Registry.enabled() if not idp.disable_for_enterprise_sso]
     return None
 
 
--- a/tests/test_utilities.py
+++ b/tests/test_utilities.py
@@ -5,6 +5,7 @@ Tests for the `edx-enterprise` utility functions.
 
 import datetime
 import unittest
+from collections import namedtuple
 
 import ddt
 import mock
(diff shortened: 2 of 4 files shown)


tests/test_utilities.py
```diff
@@ -1710,6 +1711,15 @@ def test_get_active_course_runs(self, course, users_all_enrolled_courses, expect
         """
         assert utils.get_active_course_runs(course, users_all_enrolled_courses) == expected_course_run
 
+    @mock.patch('enterprise.utils.Registry')
+    def test_get_idp_choices(self, mock_registry):
+        Provider = namedtuple('SAMLProvider', 'provider_id, name, disable_for_enterprise_sso')
+        enabled_provider = Provider('12345', 'google', False)
+        disabled_provider = Provider('333333', 'facebook', True)
+
+        mock_registry.enabled.return_value = [enabled_provider, disabled_provider]
```
binodpant, line 1720: is `enabled` a property on the Registry? I did not quite make the connectoin between the enabled and how get_idp_choices` works, but if this works in testing I guess it's just me trying to understand this better

binodpant (approved):
love this! the namedtuples are looking very clean and readable vs regular tuples

just added a clarifying q for my own learning since I have not looked at this code before

enterprise/utils.py
```diff
@@ -277,17 +269,10 @@ def get_idp_choices():
...
-        from common.djangoapps.third_party_auth.provider import Registry
-    except ImportError as exception:
-        LOGGER.warning("Could not import Registry from common.djangoapps.third_party_auth.provider")
-        LOGGER.warning(exception)
-        Registry = None  # pylint: disable=redefined-outer-name
 
     first = [("", "-" * 7)]
     if Registry:
-        return first + [(idp.provider_id, idp.name) for idp in Registry.enabled()]
+        return first + [(idp.provider_id, idp.name) for idp in Registry.enabled() if not idp.disable_for_enterprise_sso]
```
binodpant, line 275: ok yeah I see this is the new field you are adding in platform, so 🎉

tests/test_utilities.py
taliaedX: Enabled appears to be a method in the Registry class. I honestly could not find the actual code that creates this class. It is somewhere deep in one of the authentication django libraries I assume (I tried the social auth and related repos, but couldn't find Registry defined).
It is called in the `get_idp_choices` method as in: `for idp in Registry.enabled()`
The output of that call is in this format: ``` [<SAMLProviderConfig: SAMLProviderConfig object (149)>, <SAMLProviderConfig: SAMLProviderConfig object (152)>, <SAMLProviderConfig: SAMLProviderConfig object (123)>]```

binodpant: ah I see, so basically enabled() return value is what we are mocking which seems like the correct abstraction anyway, thanks for explaining

binodpant: oh yea just saw that `for idp in Registry.enabled() if not`

PR #1309, opened 2021-07-23
fix: SAP requires true without quotes in JSON payload.

taliaedX:
## Background
SAP's new endpoint appears to use true/false without quotes now.
As we're still not using the endpoint everywhere, instead of modifying `_payload_data` I've just modified the payload in the client in our particular conditional for using the new endpoint.

## Code Changes
- loads and re-dumps Payload to trigger correct format of boolean field
- Also comments on this with both explanation and TODO
- updates test to match

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/sap_success_factors/client.py
+++ b/integrated_channels/sap_success_factors/client.py
@@ -4,6 +4,7 @@ Client for connecting to SAP SuccessFactors.
 """
 
 import datetime
+import json
 import logging
 import time
 
@@ -144,6 +145,15 @@ class SAPSuccessFactorsAPIClient(IntegratedChannelApiClient):  # pylint: disable
         base_url = self.enterprise_configuration.sapsf_base_url
 
         if self.enterprise_configuration.prevent_self_submit_grades:
+            # TODO:
+            #   if/when we decide we should use the generic course completion path
+            #   for all customers, we should update the _payload_data method on the
+            #   SapSuccessFactorsLearnerDataTransmissionAudit class instead of doing this json load/dump
+            # Explanation:
+            #  json.loads loads "true" as True and when dumped again will be `true` (without quotes)
+            #  The initial payload explicitly sets the value to "true" *before* the dump, and thus it dumps with quotes.
+            payload_to_update = json.loads(payload)
+            payload = json.dumps(payload_to_update)
             return self._call_post_with_session(
                 base_url + self.GENERIC_COURSE_COMPLETION_PATH,
                 payload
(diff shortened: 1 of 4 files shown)


tests/test_integrated_channels/test_sap_success_factors/test_client.py
```diff
@@ -340,11 +340,14 @@ def test_client_uses_prevent_learner_submit_flag(self):
         )
 
         expected_response = 200, json.dumps(expected_response_body)
+        transformed_payload = self.completion_payload
+        transformed_payload['courseCompleted'] = True
```
mattdrayer, line 344: Any concerns about `True` versus `true`?

mattdrayer: Discussed out-of-band -- we are sending `true` in the payload versus `True` -- good to go!

mattdrayer (approved):
LGTM!

integrated_channels/sap_success_factors/client.py
```diff
@@ -144,6 +145,15 @@ def create_course_completion(self, user_id, payload):
         base_url = self.enterprise_configuration.sapsf_base_url
 
         if self.enterprise_configuration.prevent_self_submit_grades:
+            # TODO:
+            #   if/when we decide we should use the generic course completion path
+            #   for all customers, we should update the _payload_data method on the
+            #   SapSuccessFactorsLearnerDataTransmissionAudit class instead of doing this json load/dump
+            # Explanation:
+            #  json.loads loads "true" as True and when dumped again will be `true` (without quotes)
```
edx-abolger, line 153: OMG JSON loads why you do that.

edx-abolger (approved):

integrated_channels/sap_success_factors/client.py
taliaedX: well in this case I want it to do that haha.
I originally had logic doing `courseCompleted = True if courseCompleted == "true" else False` and the test kept failing and that's when I took a closer look and realized just loading it was enough to get the boolean I wanted. 🙃 
I just felt the need to explain because I'm otherwise not doing anything here besides loading and re-dumping it lol.

tests/test_integrated_channels/test_sap_success_factors/test_client.py
taliaedX: Yeah just to clarify this exact test case:
The actual code logic takes a payload formatted as `completedCourse: "true"` and loads it which gives us `completedCourse: True` and then dumps it back out to get `completedCourse: true`

The 2 lines above here and the one below this comment mimic that logic to form the payload we should be sending to SAP (and in this test that we send to the `_call_post_with_session` method)

taliaedX: It(the test) just has slightly different logic from the actual code because completion_payload here is a dictionary whereas in the code we would receive a serialized version (aka a string)

PR #1311, opened 2021-07-25
maint: no longer make calls to enterprise catalog when no updates needed

alex-sheehan-edx:
[Proposal](https://openedx.atlassian.net/wiki/spaces/SOL/pages/3013509195/transmit+content+metadata+hammers+the+enterprise-catalog+DB+-+make+it+stop)

[Jira](https://openedx.atlassian.net/browse/ENT-4643)

**Background**
Through several investigations, we've found that the content metadata exporter is hammering the enterprise catalog API every single time the content metadata task is run in the integrated channels. This was due to the fact that despite the need, or lack thereof, for an update (a customer's catalog or it's metadata changes), we would still retrieve the entire catalog via the enterprise catalog's API. Additionally, often the customers would be requesting `all-content` meaning that with pagination, we would be making 150~ calls to the catalog **per customer, per transmission** in order to fetch the content metadata. This along with the fact that we were running these transmissions every 10 minutes for all customers meant that we were wasting a lot of CPU usage for essentially NOOPs. 

Simultaneously, we also saw several odd behaviors from Cornerstone customers' transmissions. Firstly, CSOD customers were being included in scheduled transmissions, despite the fact that CSOD is a pull, not push based system and thus was completely unnecessary. This has been fixed with the merging of https://github.com/edx/edx-enterprise/pull/1304. Secondly, we noticed that even when CSOD customers were being successfully transmitted to, there were no content metadata item audits being stored. Lastly, it's important to note that CSOD customers have been told to expect the `cornerstone/course-list` endpoint will always return a current state of the course catalog, meaning that any missing content will be treated as deletes.

**Notes and outstanding questions**
- Now that we're creating another endpoint for CSOD customers, we could potentially be a bit smarter with how we format our data. Is it worth returning parsed `created`, `updated` and `deleted` pieces of data? Is there any other way we'd want to format the response that could be helpful to customers? @mattdrayer @binodpant 

**What this PR does:**
- adds a check to the content metadata `export` function that will query enterprise catalog for the most recent time the customer's catalog has been changed
- will remove any of the customer's catalogs from the transmission if they don't require updates (based on the timestamp of the most recent transmission audit
- adds an `export_force_all_catalogs` method to the content metadata exporter that effectively functions exactly the same as the old `export` method. 
- For the Cornerstone integrated channel:
  - Adds a new view endpoint `course-updates` that functions exactly as `course-list` except it will exclude metadata from catalogs that do not require updates
  - Maintains the exact same behavior in the `course-list` endpoint
  - Creates, updates and deletes `ContentMetadataItemTransmission` objects on transmissions the same way all other channels do

**Merge checklist:**

**Post merge:**


docs/decisions/0006-transmit-updates-only.rst
```diff
@@ -0,0 +1,87 @@
+Transmitting Only Updates IDA
+-----------------------------
+
+Status
+======
+
+Draft
```
edx-abolger, line 7: If you submit this PR it won't be a draft, it will be accepted because it's implemented here :)

alex-sheehan-edx: made it a draft while we go through review 😄

docs/decisions/0006-transmit-updates-only.rst
```diff
@@ -0,0 +1,87 @@
...
+
+Context
+=======
+The transmit_content_metadata jenkins job (along with the underlying integrated_channels.transmit_content_metadata()
+task and export() method) makes an uncomfortably large number of calls to the the enterprise-catalog service’s
+/api/v1/enterprise-catalog/<uuid>/get_content_metadata endpoint with a relatively low number of customers. The resulting
+response payloads are usually quite large which results in the service’s shared Aurora DB often being pinned at 100% CPU
```
edx-abolger, line 14: Simplify for OS community: "...which results in unacceptably high database load"

docs/decisions/0006-transmit-updates-only.rst
```diff
@@ -0,0 +1,87 @@
...
+In order to address these performance issues and to more effectively handle on-boarding more customers to the Integrated
+Channels, we have decided to address requests to the enterprise catalog on two fronts:
+
+1. Increase the page size of requests going to `/api/v1/enterprise-catalog/<uuid>/get_content_metadata` from a base of
+10 to 100. This will reduce the number of calls necessary to retrieve the entire catalog from 150~ to about 15
+(as of 07/2021).
+
+2. Add an additional request/check to the Integrated Channel's exporter that will retrieve the last time either a
+catalog or the associated content metadata of a catalog has been changed (whichever is most recent). This timestamp will
+be compared to the most recent time the customer has had a successful transmission of that particular catalog. If there
```
edx-abolger, line 53: Yesterday we decided on "the most recent catalog update timestamp that has been successfully exported already" instead of what you have here, correct?

docs/decisions/0006-transmit-updates-only.rst
```diff
@@ -0,0 +1,87 @@
...
+is no updated needed, then the exporter will refrain from querying enterprise catalog for the catalog's metadata.
+
+Exceptions
+==========
+
+All CSOD customers have been told to expect the cornerstone/course-list endpoint will always return a current state of
```
edx-abolger, line 59: Remove customer-specifics, state as " The cornerstone/course-list endpoint will always...."

docs/decisions/0006-transmit-updates-only.rst
```diff
@@ -0,0 +1,87 @@
...
+1. Increase the page size of requests going to `/api/v1/enterprise-catalog/<uuid>/get_content_metadata` from a base of
+10 to 100. This will reduce the number of calls necessary to retrieve the entire catalog from 150~ to about 15
+(as of 07/2021).
+
+2. Add an additional request/check to the Integrated Channel's exporter that will retrieve the last time either a
+catalog or the associated content metadata of a catalog has been changed (whichever is most recent). This timestamp will
+be compared to the most recent time the customer has had a successful transmission of that particular catalog. If there
+is no updated needed, then the exporter will refrain from querying enterprise catalog for the catalog's metadata.
+
+Exceptions
```
edx-abolger, line 56: "Exception: The Cornerstone (CSOD) LMS Integration will not use the 'transmit only updates' rule."

docs/decisions/0006-transmit-updates-only.rst
```diff
@@ -0,0 +1,87 @@
...
+
+4. making create, update and deletes of ContentMetadataItemTransmission objects on all CSOD transmissions the same way
+all other channels currently do.
+
+Consequences
+============
+
+The transmit-content-metadata jenkins job can continue to run every 10 minutes, so that it catches and transmits updated
+content metadata when the metadata actually changes. However, the cost of the average run in terms of CPU usage will go
+down significantly. That being said, f any metadata associated with a catalog has changed between job run N and N+1,
```
edx-abolger, line 80: nit: "if"

docs/decisions/0006-transmit-updates-only.rst
```diff
@@ -0,0 +1,87 @@
...
+then we’ll transmit the full catalog to the integrated channel during run N+1. Typically, the metadata in
+enterprise-catalog is updated once per day, so we’d expect to only be transmitting catalog data to these integrated
```
edx-abolger, line 82: Lines 82-87 are business processes, they should not go in the ADR. If we change the exception for Cornerstone in the future, we can write a new ADR.

docs/decisions/0006-transmit-updates-only.rst
alex-sheehan-edx: yes sorry, forgot to refactor

docs/decisions/0006-transmit-updates-only.rst
alex-sheehan-edx: I've rewritten the `Consequences` section as I feel it was more Alex D's and my separate thoughts copied into a block. It should now read more cleanly.

docs/decisions/0006-transmit-updates-only.rst
```diff
@@ -0,0 +1,73 @@
...
+be compared to the `catalog_last_changed` field of the ContentMetadataItemTransmission object, which is the `catalog
+last modified` value retrieved from the customer's most recent successful transmission. If there
+is no updated needed, then the exporter will refrain from querying enterprise catalog for the catalog's metadata.
+
+Exceptions
+==========
+
+1. The Cornerstone (CSOD) LMS Integration will not use the 'transmit only updates' rule."
+
+2. CSOD related transmissions will record ContentMetadataItemTransmission objects, same as all other channels.
```
edx-abolger, line 63: 😍

enterprise/api_client/enterprise_catalog.py
```diff
@@ -143,37 +143,53 @@ def delete_enterprise_catalog(self, catalog_uuid):
             return {}
 
     @JwtLmsApiClient.refresh_token
-    def get_content_metadata(self, enterprise_customer, enterprise_catalogs=None):
+    def get_content_metadata(self, enterprise_customer_catalogs, catalogs_last_modified=None):
```
edx-abolger, line 146: So, this is an API call used in a bunch of external repositories. You can't just change the Interface to not have an enterprise_customer UUID.  You need to put that field back to avoid a bunch of unintended consequences.

edx-abolger: What I would do is add your new catalogs_last_modified optional parameter, and if enterprise_catalogs is none, keep the existing behavior where that means all catalogs.

enterprise/api_client/enterprise_catalog.py
```diff
@@ -143,37 +143,53 @@ def delete_enterprise_catalog(self, catalog_uuid):
...
+                        content_catalog_last_modified[content_id] = catalogs_last_modified[catalog_uuid]
             except (SlumberBaseException, ConnectionError, Timeout) as exc:
                 LOGGER.exception(
                     'Failed to get content metadata for Catalog [%s] in enterprise-catalog due to: [%s]',
                     catalog_uuid, str(exc)
                 )
                 raise
 
-        return list(content_metadata.values())
+        return list(content_metadata.values()), content_catalog_last_modified
```
edx-abolger, line 192: The other thing you chould choose to do for better backwards compatibility is to not return the 2nd parameter if catalogs_last_modified is passed in as None.

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_content_metadata.py
```diff
@@ -52,34 +54,46 @@ def test_content_exporter_export(self, mock_get_content_metadata):
             FAKE_UUIDS[3],
         ])
 
-    @mock.patch("enterprise.api_client.enterprise_catalog.EnterpriseCatalogApiClient.get_content_metadata")
-    def test_export_with_catalogs_to_transmit(self, mock_get_content_metadata):
+    @mock.patch('integrated_channels.integrated_channel.exporters.content_metadata.EnterpriseCatalogApiClient')
+    def test_export_with_catalogs_to_transmit(self, mock_ent_catalog_api):
```
edx-abolger, line 58: Again, see comments above about making sure other calls to get_content_metadata dont violate existing API contracts. I would leave this test alone and make sure it passes AS-IS for backwards compatibility- your 'test_export_fetches_content_only_when_update_needed' covers the new updates.

test_utils/factories.py
```diff
@@ -823,3 +823,28 @@ class Meta:
     text = factory.LazyAttribute(lambda x: FAKER.word())
     start_date = factory.Faker('date_object')
     expiration_date = factory.Faker('date_object')
+
+
+class ContentMetadataItemTransmissionFactory(factory.django.DjangoModelFactory):
```
edx-abolger, line 828: 👍  for adding a factory for this important data structure!

tests/test_enterprise/api_client/test_enterprise_catalog.py
```diff
@@ -276,7 +277,7 @@ def test_get_content_metadata():
     }
     responses.add(responses.GET, second_url, json=second_expected_response)
 
-    results = client.get_content_metadata(catalog.enterprise_customer, [catalog])
+    results, _ = client.get_content_metadata([catalog_data])
```
edx-abolger, line 280: Change back, see top level API compatibility comments

edx-abolger (changes requested):
I'm about 1/2 way through all files on review, but I'd like you to revisit your api_client changes and fix those first.

docs/decisions/0006-transmit-updates-only.rst
```diff
@@ -0,0 +1,73 @@
...
+
+In order to address these performance issues and to more effectively handle on-boarding more customers to the Integrated
+Channels, we have decided to address requests to the enterprise catalog on two fronts:
+
+1. Increase the page size of requests going to `/api/v1/enterprise-catalog/<uuid>/get_content_metadata` from a base of
+10 to 100. This will reduce the number of calls necessary to retrieve the entire catalog from 150~ to about 15
+(as of 07/2021).
+
+2. Add an additional request/check to the Integrated Channel's exporter that will retrieve the last time either a
+cataclog or the associated content metadata of a atalog has been changed (whichever is most recent). This timestamp will
```
taliaedX, line 53: cataclog 👞

taliaedX: also `atalog`

docs/decisions/0006-transmit-updates-only.rst
```diff
@@ -0,0 +1,73 @@
...
+Draft
+
+Context
+=======
+The transmit_content_metadata jenkins job (along with the underlying integrated_channels.transmit_content_metadata()
+task and export() method) makes an uncomfortably large number of calls to the the enterprise-catalog service’s
+/api/v1/enterprise-catalog/<uuid>/get_content_metadata endpoint with a relatively low number of customers which results
+in an unacceptably high database load.
+
+usage and causing a myriad of alerts and warnings.
```
taliaedX, line 16: Is this supposed to be a continuation of the previous sentence?

docs/decisions/0006-transmit-updates-only.rst
```diff
@@ -0,0 +1,73 @@
...
+
+The `export()` method is the main method responsible for fetching said content metadata. It will take all listed
+`catalogs to transmit` under the customer's configurations (otherwise will default to all the customer's catalogs if
+`catalogs to transmit` is not provided) and request content metadata from the enterprise catalog. These catalogs can
+range from "all possible content" to a specific, single course.
+
+Decisions
+=========
+
+In order to address these performance issues and to more effectively handle on-boarding more customers to the Integrated
```
taliaedX, line 45: onboarding doesn't need a dash.

enterprise/api_client/enterprise_catalog.py
taliaedX: I don't disagree and we need to be careful with changing APIs, but I want to make sure I'm not missing something: what external repositories? The only one using edx-enterprise's get_content_metadata call is enterprise_reporting AFAICT (and yes this change as written would break it). Everything else is using Catalog's `get_content_metadata` endpoint.

integrated_channels/cornerstone/views.py
```diff
@@ -12,6 +12,112 @@
...
+
+class CornerstoneCoursesUpdates(BaseViewSet):
+    """
+        **Use Cases**
+
+            Get a list of courses that need updated to share on cornerstone filtered by an enterprise customer.
+
+        **Example Requests**
+
+            GET /cornerstone/course-updates?ciid={customer_uuid}
```
taliaedX, line 39: are we going to tell Cornerstone customers to use this new endpoint? Or did I miss it being used elsewhere in this PR? This is as dumb of a question as it looks :D

integrated_channels/cornerstone/views.py
```diff
@@ -71,16 +177,6 @@ class CornerstoneCoursesListView(generics.ListAPIView):
             empty result.
 
     """
-    permission_classes = (permissions.IsAuthenticated,)
-
-    OAuth2Authentication = get_oauth2authentication_class()
```
taliaedX, line 76: potentially stupid question: why did we get rid of all the auth stuff here?

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -71,12 +74,46 @@ def __init__(self, user, enterprise_configuration):
...
+                    catalog_last_modified = enterprise_catalog.get('catalog_modified')
+                else:
+                    catalog_last_modified = None
+            elif not enterprise_catalog.get('catalog_modified'):
+                catalog_last_modified = enterprise_catalog.get('content_last_modified')
+            else:
+                catalog_last_modified = max(
+                    enterprise_catalog.get('content_last_modified'),
+                    enterprise_catalog.get('catalog_modified')
+                )
```
taliaedX, line 97: This if block is breaking my brain.
So after my 5th or so read, if I'm reading this right, why not just do like:
```
if not (enterprise_catalog.get('content_last_modified') AND enterprise_catalog.get('catalog_modified'):
    catalog_last_modified = None
else:
   catalog_last_modified = max(
          enterprise_catalog.get('content_last_modified', '0'),
          enterprise_catalog.get('catalog_modified', '0')
    )
```
(Ignore my hideous space/indentation count, I'm writing this on github)

taliaedX (commented):
I'm gonna need to read this over again when my brain is more functional, but for now I have very nitty nitpicks and some questions

integrated_channels/integrated_channel/exporters/content_metadata.py
taliaedX: (From the test input it seems like we get a string at this stage which is why I use a default stringified 0 instead of say an int or None since that breaks the max operation)

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_content_metadata.py
alex-sheehan-edx: the issue is that the exporter has changed how it retrieves catalogs, and there's a new API call to `get_enterprise_catalog`, so regardless I have to account for the new mock needed. This is also a test on how the `exporter`'s `export()` method calls the API, not necessarily testing the API itself. I think as such the changes to this test are ok, no?

docs/decisions/0006-transmit-updates-only.rst
alex-sheehan-edx: I thiiiink I just meant to delete it haha

enterprise/api_client/enterprise_catalog.py
alex-sheehan-edx: method has been reverted to be backwards compatible, the minor issue is now there's basically two use cases

integrated_channels/cornerstone/views.py
alex-sheehan-edx: to avoid redundancies, the auth stuff was added to a base class above

integrated_channels/cornerstone/views.py
alex-sheehan-edx: No dumb questions 🙃. The idea is that we can't change the behaviors of existing endpoints because customers are expecting and handling the current patterns. So I added this new endpoint that we can roll out to customer's when they change their systems to handle it. There's a monthly meeting between CSOD and Matt D and the team- I'm going to be joining that call to talk about the new endpoint and any potential improvements that could be made (transmitting formatted data like `updated courses`, `created courses` and `deleted courses`. But until that conversation happens, this will basically be unused.

integrated_channels/integrated_channel/exporters/content_metadata.py
alex-sheehan-edx: Lol yeah this was as much a pain to write as it is probably to read. My only concern is comparing `0` to datetimes, but this could potentially be a great improvement to readability.

integrated_channels/cornerstone/views.py
taliaedX: oh somehow I missed the re-adding of it elsewhere. 🤦

integrated_channels/integrated_channel/exporters/content_metadata.py
taliaedX: are they true datetimes though? From the test data they appear to be strings that _can_ be read in as datetime by a method that parses them that way. We don't seem to convert them to actual datetimes anywhere up to this point as far as I could tell. Thus string comparisons :p

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -68,11 +70,14 @@ def _partition_items(self, channel_metadata_item_map):
...
             channel_metadata = item.channel_metadata
             transmitted_item = transmission_map.get(content_id, None)
+
             if transmitted_item is not None:
                 if diff(channel_metadata, transmitted_item.channel_metadata):
                     items_to_update[content_id] = channel_metadata
+                    items_catalog_updated_times[content_id] = item.catalog_last_changed
             else:
                 items_to_create[content_id] = channel_metadata
+                items_catalog_updated_times[content_id] = item.catalog_last_changed
```
taliaedX, line 80: Is there a point in putting this line in the if block? You define it the same in both conditions.

taliaedX: Also yeah what if there is no existing `catalog_last_changed` value? I know at some point we set it to `catalog_last_modified` or something which is also None by default (I assume it will take on a value though)

taliaedX (commented):
A couple of minor questions, but otherwise looks fine to me. I defer to Adrienne as I know there was some back and forth questions around backwards compatibility.

enterprise/api_client/enterprise_catalog.py
```diff
@@ -143,36 +143,68 @@ def delete_enterprise_catalog(self, catalog_uuid):
...
+        handles two cases for content metadata retrieval. Although all parameters are listed as optional, a combination
+        is always required based on use case.
+
+        Use Case (1): If enterprise_customer is specified, then enterprise_catalogs is optional but must be a list of
+        EnterpriseCustomerCatalog objects if used. catalogs_last_modified cannot be used if enterprise_customer is
+        supplied. If enterprise_catalogs is used, this method will fetch content metadata for only the provided
+        catalogs, else all catalogs under the provided enterprise_customer will be fetched.
+
+        Use Case (2): If catalogs_last_modified is specified, then enterprise_catalogs MUST be provided. However,
+        enterprise_catalogs has be a list of dictionaries containing the a key/value pairs of `uuid: <catalog UUID>`.
```
iloveagent57, line 158: Would it be worth making a new function like `get_content_metadata_modified_since()` to make some of this branching logic easier to reason about?  It'd also help you get around the problem of sometimes returning one thing (a list) and sometimes two (a list and a timestamp).

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -71,12 +74,42 @@ def __init__(self, user, enterprise_configuration):
...
+        Export transformed content metadata if there has been an update to the consumer's catalogs
+        """
+        enterprise_customer_catalogs = self.enterprise_configuration.customer_catalogs_to_transmit or \
+            self.enterprise_customer.enterprise_customer_catalogs.all()
+
+        catalogs_to_transmit = []
+        catalogs_last_modified = {}
+        for enterprise_customer_catalog in enterprise_customer_catalogs:
+            enterprise_catalog = self.enterprise_catalog_api.get_enterprise_catalog(enterprise_customer_catalog.uuid)
+            if not enterprise_catalog.get('content_last_modified') and enterprise_catalog.get('catalog_modified'):
```
iloveagent57, line 86: Do you need parentheses for this condition?
```python
>>> not True and False
False
>>> not (True and False)
True
```

integrated_channels/integrated_channel/exporters/content_metadata.py
iloveagent57: @alex-sheehan-edx I'd suggest forcing `catalog_last_modified` and `last_successful_transmission` to always be datetime objects to make this whole thing simpler.  

* Have `_get_most_recent_catalog_update_time()` return `datetime.min` by default instead of `None`.
* Deserialize `content_last_modified` and `catalog_modified` to datetimes inside of `enterprise_catalog_api get_enterprise_catalog()` - you can probably use `dateparse.parse_datetime()`.  Set them to `datetime.min` if they're not present in the response.

After doing that, you can just do
```python
catalog_last_modified = max(
      enterprise_catalog['content_last_modified'],
      enterprise_catalog['catalog_modified'],
)
last_successful_transmission = self._get_most_recent_catalog_update_time()
if last_successful_transmission < catalog_last_modified:
    # do your thing
```

iloveagent57 (commented):
Still reading, but here are a few suggestions to start with.

integrated_channels/integrated_channel/transmitters/content_metadata.py
alex-sheehan-edx: the if block is supposed to determine if there needs to be an update or create, there's the case were it's neither (course already created and no updated needed) and in those instances I don't want to add the catalog last changed timestamps to the dict. Basically wherever we add things to either `items_to_update` or `items_to_create` (not `items_to_delete` as there's no timestamps saved for those, we're just deleting the record). 

If the `catalog_last_changed` is None at this point, then this had to have been from a CSOD transmission- basically it reads the payload for the `catalog_last_changed` value and the exporter creates the catalog_last_changed value from the request from ent-catalog. If there's a None value then it will simply save the transmissions without the the value. This is roughly the same as requiring the next transmission to update.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -147,13 +167,51 @@ def _transform_item(self, content_metadata_item):
...
         return transformed_item
 
+    def _get_enterprise_catalog_last_changed(self,  enterprise_catalog):
+        """
+        Retrieves catalog metadata, specifically selecting for `content_last_modified` and `catalog_modified`.
+        """
+        enterprise_catalog_data = self.enterprise_catalog_api.get_enterprise_catalog(
+            enterprise_catalog.uuid
+        )
+        content_last_modified = enterprise_catalog_data.get('content_last_modified', None)
```
iloveagent57, line 177: Nit: the second argument of `dict.get()` defaults to `None` already.

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -38,10 +38,11 @@ def transmit(self, payload, **kwargs):
         """
         Transmit content metadata items to the integrated channel.
         """
-        items_to_create, items_to_update, items_to_delete, transmission_map = self._partition_items(payload)
+        items_to_create, items_to_update, items_to_delete, transmission_map, items_catalog_updated_times = \
```
iloveagent57, line 41: Nit: `items_catalog_updated_times` is kind of a misnomer, since this is really a mapping of `content_id` to the time that _content_ last changed.  Maybe something like `content_update_times_by_id`?

iloveagent57 (commented):
Couple nits and check that we're sorting on the correct field.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -147,13 +167,51 @@ def _transform_item(self, content_metadata_item):
...
+        # pylint: disable=invalid-name
+        ContentMetadataItemTransmission = apps.get_model(
+            'integrated_channel',
+            'ContentMetadataItemTransmission'
+        )
+        past_transmissions = ContentMetadataItemTransmission.objects.filter(
+            enterprise_customer=self.enterprise_configuration.enterprise_customer,
+            integrated_channel_code=self.enterprise_configuration.channel_code(),
+            content_last_changed__isnull=False,
+        ).values('content_last_changed').order_by('-modified')
```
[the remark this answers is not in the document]
alex-sheehan-edx: it's a timestamped model, so this seems like either would work in this situation? but content_last_changed would be more clear.

integrated_channels/cornerstone/exporters/content_metadata.py
```diff
@@ -47,11 +47,16 @@ class CornerstoneContentMetadataExporter(ContentMetadataExporter):  # pylint: di
...
+        """
+        return self.export(force_retrieve_all_catalogs=True)
+
     def transform_organizations(self, content_metadata_item):
         """
-        Return the transformed version of the course organizations
-        by converting each organization into cornerstone course partner object.
-        or default Partner if no partner found
+        Return the transformed version of the course organizations by converting each organization into cornerstone
+        course partner object. or default Partner if no partner found
```
edx-abolger, line 59: +1 for docstrings

integrated_channels/cornerstone/views.py
```diff
@@ -12,6 +12,112 @@
 from integrated_channels.cornerstone.models import CornerstoneEnterpriseCustomerConfiguration
 
 
+class BaseViewSet(generics.ListAPIView):
```
edx-abolger, line 15: You say this, but this isn't used for class CornerstoneCoursesListView(generics.ListAPIView). Is that an oversight?

integrated_channels/integrated_channel/exporters/content_metadata.py
edx-abolger: Unresolving this because I agree with Talia that the crazy if statements hurt my brain.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -71,28 +75,44 @@ def __init__(self, user, enterprise_configuration):
...
+                # Retrieve whatever modification happened more recently
+                catalog_last_modified = max(
+                    content_last_modified,
+                    catalog_modified
+                )
+                last_successful_transmission = self._get_most_recent_catalog_update_time()
+
+            # If the last successful transmission was since the last time the catalog was updated, there's no need for
+            # an update
+            if force_retrieve_all_catalogs or not last_successful_transmission or \
```
edx-abolger, line 100: I know this will work b/c force_retrieve_all_catalogs means you won't eval the rest of the if statement, but it's really hard to figure out that if statement when half the time the variables in it will be undefined. Can you abstract out it:

```
need_catalog_update = False
if not force_retrieve_all_catalogs:
  do_stuff;
  need_catalog_update = (not last_successful_transmission) or (last_successful_transmission < catalog_last_modified)
  
 if not force_retrieve_all_catalogs or need_catalog_update: 
   do_more_stuff()
```

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -71,28 +75,44 @@ def __init__(self, user, enterprise_configuration):
...
+                    last_successful_transmission < catalog_last_modified:
+                content_metadata_item = self.enterprise_catalog_api.get_content_metadata(
+                    self.enterprise_customer,
+                    [enterprise_customer_catalog]
+                )
+                for item in content_metadata_item:
+                    transformed = self._transform_item(item)
+                    LOGGER.debug(
```
edx-abolger, line 108: Plug: since this is a new log call, is it a lot of work to use the new log format here?

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -71,28 +75,44 @@ def __init__(self, user, enterprise_configuration):
...
+                        'Exporting content metadata item with plugin configuration [%s]: [%s]',
+                        self.enterprise_configuration,
+                        json.dumps(transformed, indent=4),
+                    )
+                    content_last_modified = item.pop('content_last_modified')
+                    content_metadata_item_export = ContentMetadataItemExport(item, transformed, content_last_modified)
+                    content_metadata_export[content_metadata_item_export.content_id] = content_metadata_item_export
         return OrderedDict(sorted(content_metadata_export.items()))
```
edx-abolger, line 116: If I understand this correctly, you have keys representing course keys: `content_metadata_export[content_metadata_item_export.content_id]`
Why are you sorting the data by course key? Do we really need data sorted alphabetically by course key or should this be some other field?

edx-abolger: Edit: I know it was already like that, but why?

integrated_channels/cornerstone/views.py
```diff
@@ -12,6 +12,112 @@
...
+        enterprise_config = CornerstoneEnterpriseCustomerConfiguration.objects.get(
+            enterprise_customer=enterprise_customer
+        )
+        exporter = enterprise_config.get_content_metadata_exporter(worker_user)
+        transmitter = enterprise_config.get_content_metadata_transmitter()
+        data = transmitter.transmit(exporter.export())
+        return Response(data)
+
+
 class CornerstoneCoursesListView(generics.ListAPIView):
```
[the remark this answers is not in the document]
alex-sheehan-edx: yeah.. that's a nasty mistake. I can look into some unit tests.

integrated_channels/integrated_channel/exporters/content_metadata.py
alex-sheehan-edx: Good point- my only thought is that it could have something to do with the customers who's transmission chunk size is larger and they somehow/for some reason expect the data in sorted order. But I can't point to any one example of that...

edx-abolger (approved):

iloveagent57 (approved):

PR #1312, opened 2021-07-27
feat: emails in a celery task, for calls to `EnterpriseCustomer#notify_enrolled_learners()`

binodpant:
## Included in PR
Summary of changes (updated from what I had since last round of review):
- Added a task to send email notifications in `enterprise/tasks.py`
- Extracted out a method to only serialize the email_items list so that we can use it to call a celery task
- Now calling the celery task from the model's `notify_enrolled_learners` function so every caller of the notify_enrolled_learners function gets this ability, not just bulk enrollment
- Moved a function `parse_lms_api_datetime` to utils. Strictly speaking not needed for this PR but I decided to leave it there 

For bulk enrollment, no. of tasks created = no. of courses. Each task will email all learners for enrollment into that course

ENT-4779

## Testing notes:

* ADDED: Unit test for the new task
* UPDATED: Update tests that call the view to ensure the task is invoked correctly

`Staging`:
Once this (via edx-platform) gets to staging, test plan is to put a pause on edx-platform prod pipeline, run a test with enrollment of these kinds:

- bulk enroll with emails on and off
- regular enroll in a course from courseware page
- learner portal enroll in a course

`Manual`:

Executed bulk enrollment, and verified a task was fired for each course and also an email file was generated per course 

```
ls -al /edx/src/ace_messages/
total 1296
drwxr-xr-x 5 root root    160 Jul 28 03:16 .
drwxr-xr-x 4 root root    128 Jan 14  2021 ..
-rw-r--r-- 1 root root 438845 Jul 28 03:16 20210728-031634-139685836518400.log
-rw-r--r-- 1 root root 438843 Jul 28 03:16 20210728-031636-139685832905728.log
-rw-r--r-- 1 root root 440297 Jul 28 03:16 20210728-031638-139685848429376.log
```

in logs I see 

```
2021-07-29 13:38:53,667 INFO 32082 [enterprise.api.v1.views] [user 17] [ip 172.18.0.16] views.py:315 - Successfully bulk enrolled learners: {<User: ecommerce_worker>, <User: edx>, <User: verified>} into course course-v1:edX+909878+2T2021
2021-07-29 13:38:54,398 INFO 32082 [celery.app.trace] [user 17] [ip 172.18.0.16] trace.py:125 - Task enterprise.tasks.send_enterprise_email_notification[0c160b49-e277-4654-9a8c-da6a3bd78bf8] succeeded in 0.042356400001153816s: None
2021-07-29 13:38:54,399 INFO 32082 [enterprise.api.v1.views] [user 17] [ip 172.18.0.16] views.py:315 - Successfully bulk enrolled learners: {<User: ecommerce_worker>, <User: edx>, <User: verified>} into course course-v1:edX+303707808+2T2021
2021-07-29 13:38:55,056 INFO 32082 [celery.app.trace] [user 17] [ip 172.18.0.16] trace.py:125 - Task enterprise.tasks.send_enterprise_email_notification[a0583741-2b13-4e00-8a66-5685bb492fd5] succeeded in 0.043135300002177246s: None
2021-07-29 13:38:55,057 INFO 32082 [enterprise.api.v1.views] [user 17] [ip 172.18.0.16] views.py:315 - Successfully bulk enrolled learners: {<User: ecommerce_worker>, <User: edx>, <User: verified>} into course course-v1:edX+808404707+2T2021
2021-07-29 13:38:55,740 INFO 32082 [celery.app.trace] [user 17] [ip 172.18.0.16] trace.py:125 - Task enterprise.tasks.send_enterprise_email_notification[4a8eb86b-e88d-4289-a6a0-28ce30d3468a] succeeded in 0.04192120000516297s: None
[29/Jul/2021 13:38:55] "POST /enterprise/api/v1/enterprise-customer/9a5afb8e-63bc-456a-a600-81b60777e5aa/enroll_learners_in_courses/ HTTP/1.1" 201 789
```

Each file represents emails send to all 25 learners I used in testing.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -3,6 +3,7 @@
 Views for enterprise api version 1 endpoint.
 """
 
+from enterprise.tasks import notify_enrolled_learners
 from logging import getLogger
 from smtplib import SMTPException
 
@@ -318,12 +319,13 @@ class EnterpriseCustomerViewSet(EnterpriseReadWriteModelViewSet):
                 ))
                 track_enrollment(PATHWAY_CUSTOMER_ADMIN_ENROLLMENT, request.user.id, course_run)
                 if serializer.validated_data.get('notify'):
-                    enterprise_customer.notify_enrolled_learners(
+                    email_items = enterprise_customer.prepare_notification_content(
                         catalog_api_user=request.user,
                         course_id=course_run,
                         users=pending_users | existing_users,
                         admin_enrollment=True,
                     )
+                    notify_enrolled_learners.delay(enterprise_customer.uuid, True, email_items)
 
         if email_errors:
             results['invalid_email_addresses'] = email_errors
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -320,10 +320,10 @@ class EnterpriseCustomerViewSet(EnterpriseReadWriteModelViewSet):
                 track_enrollment(PATHWAY_CUSTOMER_ADMIN_ENROLLMENT, request.user.id, course_run)
                 if serializer.validated_data.get('notify'):
                     email_items = enterprise_customer.prepare_notification_content(
-                        catalog_api_user=request.user,
-                        course_id=course_run,
-                        users=pending_users | existing_users,
-                        admin_enrollment=True,
+                        request.user,
+                        course_run,
+                        pending_users | existing_users,
+                        True,
                     )
                     notify_enrolled_learners.delay(enterprise_customer.uuid, True, email_items)
 
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -3,7 +3,6 @@
 Views for enterprise api version 1 endpoint.
 """
 
-from enterprise.tasks import notify_enrolled_learners
 from logging import getLogger
 from smtplib import SMTPException
 
@@ -58,6 +57,7 @@ from enterprise.api.v1.permissions import IsInEnterpriseGroup
 from enterprise.api_client.lms import EnrollmentApiClient
 from enterprise.constants import COURSE_KEY_URL_PATTERN, PATHWAY_CUSTOMER_ADMIN_ENROLLMENT
 from enterprise.errors import AdminNotificationAPIRequestError, CodesAPIRequestError
+from enterprise.tasks import notify_enrolled_learners
 from enterprise.utils import (
     NotConnectedToOpenEdX,
     enroll_licensed_users_in_courses,
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -57,7 +57,6 @@ from enterprise.api.v1.permissions import IsInEnterpriseGroup
 from enterprise.api_client.lms import EnrollmentApiClient
 from enterprise.constants import COURSE_KEY_URL_PATTERN, PATHWAY_CUSTOMER_ADMIN_ENROLLMENT
 from enterprise.errors import AdminNotificationAPIRequestError, CodesAPIRequestError
-from enterprise.tasks import notify_enrolled_learners
 from enterprise.utils import (
     NotConnectedToOpenEdX,
     enroll_licensed_users_in_courses,
@@ -319,13 +318,12 @@ class EnterpriseCustomerViewSet(EnterpriseReadWriteModelViewSet):
                 ))
                 track_enrollment(PATHWAY_CUSTOMER_ADMIN_ENROLLMENT, request.user.id, course_run)
                 if serializer.validated_data.get('notify'):
-                    email_items = enterprise_customer.prepare_notification_content(
-                        request.user,
-                        course_run,
-                        pending_users | existing_users,
-                        True,
+                    enterprise_customer.notify_enrolled_learners(
+                        catalog_api_user=request.user,
+                        course_id=course_run,
+                        users=pending_users | existing_users,
+                        admin_enrollment=True,
                     )
-                    notify_enrolled_learners.delay(enterprise_customer.uuid, True, email_items)
 
         if email_errors:
             results['invalid_email_addresses'] = email_errors
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -61,6 +61,7 @@ from enterprise.utils import (
     CourseEnrollmentDowngradeError,
     CourseEnrollmentPermissionError,
     NotConnectedToOpenEdX,
+    create_dict_from_user,
     get_configuration_value,
     get_ecommerce_worker_user,
     get_enterprise_worker_user,
@@ -657,16 +658,26 @@ class EnterpriseCustomer(TimeStampedModel):
         else:
             PendingEnrollment.objects.filter(user=pending_ecu, course_id__in=course_ids).delete()
 
-    def notify_enrolled_learners(self, catalog_api_user, course_id, users, admin_enrollment=False):
+    def prepare_notification_content(self, catalog_api_user, course_id, users, admin_enrollment=False):
         """
-        Notify learners about a course in which they've been enrolled.
+        Prepare serializable contents to send emails with (if using tasks to send emails)
 
-        Args:
-            catalog_api_user: The user for calling the Catalog API
-            course_id: The specific course the learners were enrolled in
-            users: An iterable of the users or pending users who were enrolled
-            admin_enrollment: Default False. Set to true if using bulk enrollment, for example.
-                When true, we use the admin enrollment template instead.
+        Returns: A list of dictionary objects that are of the form:
+          {
+              "user": user
+                "enrolled_in": {
+                    'name': course_name,
+                    'url': destination_url,
+                    'type': 'course',
+                    'start': course_start,
+                },
+                "dashboard_url": dashboard_url,
+                "enterprise_customer_uuid": self.uuid,
+                "admin_enrollment": admin_enrollment,
+          }
+          where user is one of
+              - 1: { 'first_name': name, 'username': user_name, 'email': email } (similar to a User object)
+              - 2: { 'user_email' : user_email } (similar to a PendingEnterpriseCustomerUser object)
         """
         course_details = CourseCatalogApiClient(catalog_api_user, self.site).get_course_run(course_id)
         if not course_details:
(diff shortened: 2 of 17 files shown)


enterprise/api/v1/views.py
```diff
@@ -318,12 +319,13 @@ def enroll_learners_in_courses(self, request, pk):
                 ))
                 track_enrollment(PATHWAY_CUSTOMER_ADMIN_ENROLLMENT, request.user.id, course_run)
                 if serializer.validated_data.get('notify'):
-                    enterprise_customer.notify_enrolled_learners(
-                        catalog_api_user=request.user,
-                        course_id=course_run,
-                        users=pending_users | existing_users,
-                        admin_enrollment=True,
+                    email_items = enterprise_customer.prepare_notification_content(
```
edx-abolger, line 322: Nit: can you change the name of 'prepare_notification_content' to 'serialize_notification_content' ?

enterprise/models.py
```diff
@@ -723,20 +733,51 @@ def notify_enrolled_learners(self, catalog_api_user, course_id, users, admin_enr
...
+                    'type': 'course',
+                    'start': course_start,
+                },
+                "dashboard_url": dashboard_url,
+                "enterprise_customer_uuid": self.uuid,
+                "admin_enrollment": admin_enrollment,
+            })
+        return email_items
+
+    def notify_enrolled_learners(self, catalog_api_user, course_id, users, admin_enrollment=False):
```
edx-abolger, line 754: Don't name the notify_enrolled_learners task and this function the same thing, add _task if you are going to keep both items in 2 separate modules.

enterprise/utils.py
```diff
@@ -455,20 +455,40 @@ def find_enroll_email_template(enterprise_customer, template_type):
     return enterprise_template_config
 
 
+def create_dict_from_user(user):
```
edx-abolger, line 458: Are you sure that the Django user object doesn't already have a serializer? Home writing one like this adds complexity. Can you look at
```

from django.forms.models import model_to_dict
model_to_dict(instance)
```
function and see if it meets your needs here?

edx-abolger (changes requested):
I understand caution, but you should go ahead and make the new task used everywhere, not just for bulk enrollment. The conservative approach here actually makes it harder to understand, I would rather place that conservative approach into adding a unit test for your new task (which you should make one of anyway). The module circular dependency can be solved a different way, let me know if you need help on that.

enterprise/utils.py
binodpant: thanks, thinks I should be able to make that work and avoid a custom function, will be back!

binodpant:
> I understand caution, but you should go ahead and make the new task used everywhere, not just for bulk enrollment. The conservative approach here actually makes it harder to understand, I would rather place that conservative approach into adding a unit test for your new task (which you should make one of anyway). The module circular dependency can be solved a different way, let me know if you need help on that.

Sure thing! I was, like you said, being risk averse. There is opportunity to apply the tasking to other emails sent out too and get it done in this round. I do see the value of doing it now rather than later, so all the emails get this benefit, it isn't that big a stretch to apply and test it out by enrolling directly from the courseware, learner portal etc in addition to enrolling from bulk enroll page. Yes the task portion also needs some unit testing, so going to look into that. So I accept your proposal. Be back with this!

binodpant:
also on the circular dep looks like I might already have worked myself out of that, but if I get stuck or sth will let you know thanks

enterprise/utils.py
```diff
@@ -1962,3 +2069,28 @@ def get_best_mode_from_course_key(course_key):
         else CourseModes.AUDIT
 
     return best_course_mode
+
+
+def parse_lms_api_datetime(datetime_string, datetime_format=LMS_API_DATETIME_FORMAT):
```
binodpant, line 2074: I moved it here to utils since calling EnterpriseCatalogApiClient  from tasks was causing a circular dep. I think this is a better location for this function anyway

binodpant:
A final round of testing manually as well as fixing up one test failure remains but I largely made the changes I needed to make I think, so that now I can keep the model function in place and also this change to use celery will apply to all calls of that function

tests/test_enterprise/api/test_views.py
```diff
@@ -3339,7 +3339,8 @@ class TestBulkEnrollment(BaseTestEnterpriseAPIViews):
                     'email': 'abc@test.com',
                     'course_run_key': 'course-v1:edX+DemoX+Demo_Course',
                     'license_uuid': '5a88bdcade7c4ecb838f8111b68e18ac'
-                }]
+                }],
+                'notify': 'true',
```
binodpant, line 3343: inserting this here so we can hit the code path that sends notifications

enterprise/tasks.py
```diff
@@ -8,25 +8,60 @@
...
+                item['dashboard_url'],
+                enterprise_customer_uuid,
+                email_connection=email_conn,
+                admin_enrollment=admin_enrollment,
+            )
+
+
 @shared_task
 @set_code_owner_attribute
 def create_enterprise_enrollment(course_id, enterprise_customer_user_id):
```
edx-abolger, line 55: Why were these changes needed? This is an unrelated shared task to the one you are modifying?

binodpant: oh these were because the direct use of the models in this task was causing circular ref errors since the model now calls the task

binodpant: ah just read your latest comment, yes will do that thanks

enterprise/utils.py
```diff
@@ -455,20 +460,121 @@ def find_enroll_email_template(enterprise_customer, template_type):
...
+            site=lms_root_url,
+            login_or_register='{login_or_register}',  # We don't know the value at this time
+            course_path=course_path
+        )
+
+    try:
+        course_start = parse_lms_api_datetime(course_details.get('start'))
+    except (TypeError, ValueError):
+        course_start = None
+        LOGGER.exception(
```
edx-abolger, line 537: +1 for using Logger.exception so we get the stack trace!!!

tests/test_enterprise/api/test_views.py
```diff
@@ -3469,6 +3473,38 @@ def test_bulk_enrollment_in_bulk_courses_pending_licenses(
         else:
             mock_track_enroll.assert_not_called()
 
+        # verify notification sent correctly for each course to applicable learners
+        if 'notify' in body:
```
edx-abolger, line 3477: nit: This is already a long test and I don't love conditional logic in tests, it makes them breakable. Can you please refactor the existing test by splitting it in half (old_test = 1 new function, if conditional code into 2nd new function) and then make it like this: 

```

def _original_test_logic():
   stuff

def notify_test():
   your new code

@<data>
def test_bulk_enrollment_in_bulk_courses_pending_licenses():
  original_code(params)
  
@<data you want to use for notifications>
def test_bulk_enrollment_in_bulk_courses_pending_licenses_with_notifications():
  original_code(params)
  new_notification_code(params)

```


Another option is to just run the notify verification on every test case.

tests/test_enterprise/test_tasks.py
```diff
@@ -99,3 +105,38 @@ def test_create_enrollment_task_no_create_duplicates(self, catalog_api_client_mo
...
+        admin_enrollment = True
+
+        mail_conn = mock.MagicMock()
+        mock_email_conn.return_value.__enter__.return_value = mail_conn
+
+        email_items = serialize_notification_content(
+            enterprise_customer,
+            course_details,
+            self.FAKE_COURSE_ID,
+            users + [pending_user],
```
edx-abolger, line 126: I like that both cases are covered here!

edx-abolger (changes requested):
2 small remaining requests, almost ready to go, looks very clean.

tests/test_enterprise/api/test_views.py
binodpant: I like this! let me get back with this change, so the notification use case testing is separate

binodpant: done!

edx-abolger (approved):
LGTM!

enterprise/tasks.py
```diff
@@ -8,25 +8,60 @@
...
+        *
+        * email_items: list of dictionary objects with keys:
+        *   user (dict): a dict with either of the following forms:
+              - 1: { 'first_name': name, 'username': user_name, 'email': email } (similar to a User object)
+              - 2: { 'user_email' : user_email } (similar to a PendingEnterpriseCustomerUser object)
+        *   enrolled_in (dict): name and optionally other keys needed by templates
+        *   dashboard_url (str)
+    """
+    with mail.get_connection() as email_conn:
+        for item in email_items:
```
macdiesel, line 42: Since this isn't wrapped with a try/catch, I believe that an error in one email send will prevent all from being sent. Is this behavior OK for your use case?

macdiesel (commented):
One other comment about this. Do you need to be concerned about messages that fail? I'm perhaps you might want to [add a retry](https://docs.celeryproject.org/en/stable/userguide/tasks.html#retrying) in case the error is transient? 

Would you want to retry sending these emails later should they failure? If so I would suggest saving these messages off to a dead letter queue so they could be re-run once the code/infra/whatever is fixed at a later point in time. 

Perhaps beyond the MVP for this but right now with the synchronous nature of this call the user will know when the send fails via the webapp. Since this is being moved to an async task the user will no longer have this feedback. This could cause customer confusion and potentially increase tickets to ECS.

enterprise/tasks.py
edx-abolger: probably not ok. May as well fix that here, with at least a logged exception, it existed in the prior version but that doesn't mean it wasn't a problem.  For the beyond MVP stuff we want to move this to Braze, so I wouldn't try to provide failed send feedback here.

binodpant: Sorry I had merged this by the time I read this. But yeah agree with Brian's observation. I can create a separate PR to do that. And I will likely merge and test that version tomorrow, which should make this sturdier

binodpant: I don't know exactly what Exceptions are going to be thrown by the underlying code yet though.. a quick look at the mail.send() api reveals nothing. will a `except Exception be` the safest option?

macdiesel: That hammer will catch it all, sounds good to me.

binodpant: to close the loop this update is now being released via https://github.com/edx/edx-platform/pull/28324

PR #1313, opened 2021-07-27
Fix: SAP CourseCompleted Flag - This time using actual payload data

taliaedX:
With Alex S.' help, we were able to get actual payload data so I could be certain that the test data matched and also go through the SAP transformations step by step.

The last pr didn't work because my test data ended up in a bad state (I didn't realize Python assigns references rather than performing a copy when you try to assign a variable to an existing dictionary) and I mistakenly matched the base code to the test data.

So this one fixes the test data, but also I used actual payload data to verify the functionality outside of the tests to ensure confidence in the code and thus verify the test data was the problem.

The very original problem still needed to be resolved (the endpoint does not support `"true"` as a parameter and it needed to be formatted for JS (`true` without quotes), but with this new code we now actually do that.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/sap_success_factors/client.py
+++ b/integrated_channels/sap_success_factors/client.py
@@ -148,15 +148,12 @@ class SAPSuccessFactorsAPIClient(IntegratedChannelApiClient):  # pylint: disable
             # TODO:
             #   if/when we decide we should use the generic course completion path
             #   for all customers, we should update the _payload_data method on the
-            #   SapSuccessFactorsLearnerDataTransmissionAudit class instead of doing this json load/dump
-            # Explanation:
-            #  json.loads loads "true" as True and when dumped again will be `true` (without quotes)
-            #  The initial payload explicitly sets the value to "true" *before* the dump, and thus it dumps with quotes.
+            #   SapSuccessFactorsLearnerDataTransmissionAudit class instead of doing this
             payload_to_update = json.loads(payload)
-            payload = json.dumps(payload_to_update)
+            payload_to_update['courseCompleted'] = bool(payload_to_update['courseCompleted'] == 'true')
             return self._call_post_with_session(
                 base_url + self.GENERIC_COURSE_COMPLETION_PATH,
-                payload
+                json.dumps(payload_to_update)
             )
 
         return self._call_post_with_user_override(
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * Nothing
 
+[3.27.9]
+--------
+* Fix SAP Course Completion payload format again.
+
 [3.27.8]
 --------
 * Fix SAP Course Completion payload format.
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.27.8"
+__version__ = "3.27.9"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"  # pylint: disable=invalid-name
--- a/tests/test_integrated_channels/test_sap_success_factors/test_client.py
+++ b/tests/test_integrated_channels/test_sap_success_factors/test_client.py
@@ -340,9 +340,10 @@ class TestSAPSuccessFactorsAPIClient(unittest.TestCase):
         )
 
         expected_response = 200, json.dumps(expected_response_body)
-        transformed_payload = self.completion_payload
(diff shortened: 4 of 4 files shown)


edx-abolger (approved):

integrated_channels/sap_success_factors/client.py
```diff
@@ -148,15 +148,12 @@ def create_course_completion(self, user_id, payload):
...
             #   if/when we decide we should use the generic course completion path
             #   for all customers, we should update the _payload_data method on the
-            #   SapSuccessFactorsLearnerDataTransmissionAudit class instead of doing this json load/dump
-            # Explanation:
-            #  json.loads loads "true" as True and when dumped again will be `true` (without quotes)
-            #  The initial payload explicitly sets the value to "true" *before* the dump, and thus it dumps with quotes.
+            #   SapSuccessFactorsLearnerDataTransmissionAudit class instead of doing this
             payload_to_update = json.loads(payload)
-            payload = json.dumps(payload_to_update)
+            payload_to_update['courseCompleted'] = bool(payload_to_update['courseCompleted'] == 'true')
```
mattdrayer, line 153: So, I've also seen in the transmission audit records that we are sending `courseCompleted = false` in addition to `courseCompleted = true`.  In my testing with both our pseudo-tenant and the customer tenant I was able to send a completion event with a value of `false`.  I was able to then send a second completion event with value of `true`.  At that point the completion record became "locked" and I couldn't send any more completion events without triggering the "duplicate completions" response from the tenant. 

I guess my point is here that our system is not only sending `true`, but it's also sending `false` for the `courseCompleted` value, and it's not clear to me if this latest implementation only handles the true scenario or if it handles both.  Can you clarify?

taliaedX: Quick overview of process (as of the changes in this pr, not the previous bad code I wrote):
1. The audit model forms the payload with the following: `courseCompleted="true" if self.course_completed else "false",` (ie it specifically makes boolean values into strings) (this is in the `_payload_data` method of the audit model)
2. The payload is exported as a dictionary (in the`_payload_data` method of the audit model)
3. Then json dumped which turns the whole dictionary into a string. (in the `serialize` method of the audit model)
4. In the client when using the new endpoint, we then "load" the payload (this translates a string back into a dictionary so we can access specific values easily)
5. change the string boolean values into actual booleans
4. dump the payload (which puts them into javascript-style booleans: ie `true` and `false`)

Example of this logic:
```
>>> dict = {"truevalue": "true", "falsevalue": "false"}  # This is what the payload looks like before the serialize method
>>> dump = json.dumps(dict) 
>>> dump
'{"truevalue": "true", "falsevalue": "false"}'  # This is what it looks like after the serialize method
>>> loaded = json.loads(dump)  # This is the first thing we do in this client logic
>>> loaded
{'truevalue': 'true', 'falsevalue': 'false'}
>>> loaded['truevalue'] = bool(loaded['truevalue'] == 'true')
>>> loaded
{'truevalue': True, 'falsevalue': 'false'}
>>> loaded['falsevalue'] = bool(loaded['falsevalue'] == 'true')
>>> loaded
{'truevalue': True, 'falsevalue': False}
>>> dumped_final = json.dumps(loaded)
>>> dumped_final
'{"truevalue": true, "falsevalue": false}'
```
(This might be confusing given I have `falsevalue == 'true'` in the logic there, but I wanted to show how each boolean appears in the final outcome, even when using the same condition as we are using in the actual code. 

The original change in this PR that the style guide decided was unnecessary which might be a little more obvious was:
`payload_to_update['courseCompleted'] = True if payload_to_update['courseCompleted'] == 'true' else False`

To specifically answer your question, both Python booleans are translated into their lowercase versions without quotes when dumped back into the string version of the payload.

What I did previously to this pr was simply dump the payload's value which was `"true"` without converting to a valid boolean, so it actually ended up in the payload again as `"true"`. Which makes sense. Load > dump > load > dump should end up with the same values throughout if making no other changes. I don't know where in my original testing it somehow magically became a valid boolean. I think this was due to the test conditions being wrong.

But, with an actual valid boolean in a json.dumps call it will be converted to JS style booleans `true/false` with no quotes.

mattdrayer: Awesome, thanks for the detailed explanation!  The dropped `else False` would have 100% alleviated my concern here -- I guess I don't see why it would be better style to not include it.  I mean I get that the computer will handle the false condition just fine, but it's the humans that read the code who appreciate these crutches...

PR #1315, opened 2021-07-29
Catalog query deletes logging and bulk query ID updates

alex-sheehan-edx:
[jira](https://openedx.atlassian.net/browse/ENT-4783)

background:
deleting catalog query's is currently breaking for a few processes, so we want to log an exception if one get's deleted. Additionally, in order to ever rectify this kind of event in the future, we want to add management command to bulk change enterprise catalog's queries. 

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/management/commands/bulk_update_catalog_query_id.py
+++ b/enterprise/management/commands/bulk_update_catalog_query_id.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+"""
+Django management command for bulk updating EnterpriseCustomerCatalog record's EnterpriseCatalogQuery's.
+"""
+
+import logging
+
+from django.core.management import BaseCommand
+
+from enterprise.models import EnterpriseCatalogQuery, EnterpriseCustomerCatalog
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Takes a new and old Enterprise Catalog Query ID, find all most recent records within the historical Enterprise
+    Catalog table where enterprise catalog query ID is the old ID provided and updates the corresponding Enterprise
+    Catalog table entries with the new ID.
+    """
+    help = "Updates specified EnterpriseCatalog's enterprise catalog query ID records with a provided new ID"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'old_id',
+            action='store',
+            help='Old catalog query ID to be replaced.'
+        )
+        parser.add_argument(
+            'new_id',
+            action='store',
+            help='New catalog query ID to replace the old one.'
+        )
+
+    def handle(self, *args, **options):
+        """
+        Entry point for management command execution.
+        """
+        LOGGER.info("Starting bulk update of EnterpriseCatalog's enterprise catalog query IDs")
+
+        old_id = options['old_id']
+        new_id = options['new_id']
+        uuids_to_change = []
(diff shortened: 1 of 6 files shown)


enterprise/management/commands/bulk_update_catalog_query_id.py
```diff
@@ -0,0 +1,62 @@
...
+        old_id = options['old_id']
+        new_id = options['new_id']
+        uuids_to_change = []
+        if EnterpriseCatalogQuery.objects.filter(id=new_id).first():
+            catalogs_to_change = EnterpriseCustomerCatalog.history.filter(enterprise_catalog_query_id=old_id).order_by(
+                '-modified'
+            ).distinct()
+            for catalog in catalogs_to_change:
+                if not catalog.next_record and catalog.get_history_type_display() != 'Deleted':
+                    uuids_to_change.append(str(catalog.uuid))
```
taliaedX, line 50: I don't think this actually has to be a string.
```
>>> uuids.append(catalog1.uuid)
>>> uuids.append(catalog2.uuid)
>>> EnterpriseCustomerCatalog.objects.filter(uuid__in=uuids)
<QuerySet [<EnterpriseCustomerCatalog 'All Content' with UUID '6cbd3a0e-dc96-4d67-821e-53ca39d992c0' for EnterpriseCustomer Moodle Test Enterprise>, <EnterpriseCustomerCatalog 'Test UUID' wi
```

enterprise/models.py
```diff
@@ -1912,6 +1912,16 @@ def __str__(self):
         """
         return "<EnterpriseCatalogQuery '{title}' >".format(title=self.title)
 
+    def delete(self, *args, **kwargs):
+        """
+        Deletes this ``EnterpriseCatalogQuery``.
```
taliaedX, line 1917: Didn't the query end up deleted in this case anyways? *Do* we want to explicitly delete it? A catalog query can be associated with another catalog. If someone deleted the catalog (and it didn't delete the query) we could just hook the existing query up to the new catalog. I'm not sure I see the utility in allowing deletion. Just seems like another way for someone to then accidentally the query now too lol

taliaedX (commented):
Some minor nits

enterprise/management/commands/bulk_update_catalog_query_id.py
alex-sheehan-edx: oh yeeeeah. Cool thx 👍

enterprise/models.py
alex-sheehan-edx: From the ticket AC:

> Add an alert to detect if a catalog query has been deleted

which is why I didn't _**prevent**_ the delete so to speak. But I'm more than happy to prevent it. Agreed, the management command helps us rectify any deletes that would otherwise need to happen.

taliaedX: I am dumb. I forgot delete is already inherent to the class and it *was* a query that was deleted. You're just adding an override that adds actual logging which is exactly what we want. 👍

taliaedX (approved):
I personally have no further qualms here so approving! Good luck on your quality checks bro 🤞

enterprise/models.py
```diff
@@ -1912,6 +1912,16 @@ def __str__(self):
...
+        """
+        super().delete(*args, **kwargs)
+        LOGGER.exception(
```
edx-abolger, line 1920: Yay!

edx-abolger (approved):

PR #1316, opened 2021-07-29
feat: Follow up to ENT-4779 PR to handle in loop exceptions

binodpant:
follow up to https://github.com/edx/edx-enterprise/pull/1312 per Brian's feedback

If one email send fails for any reason, avoid failure and log exception, includes a test

ENT-4779

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/tasks.py
+++ b/enterprise/tasks.py
@@ -40,14 +40,23 @@ def send_enterprise_email_notification(
     """
     with mail.get_connection() as email_conn:
         for item in email_items:
-            send_email_notification_message(
-                item['user'],
-                item['enrolled_in'],
-                item['dashboard_url'],
-                enterprise_customer_uuid,
-                email_connection=email_conn,
-                admin_enrollment=admin_enrollment,
-            )
+            course_name = item['enrolled_in']['name']
+            try:
+                send_email_notification_message(
+                    item['user'],
+                    item['enrolled_in'],
+                    item['dashboard_url'],
+                    enterprise_customer_uuid,
+                    email_connection=email_conn,
+                    admin_enrollment=admin_enrollment,
+                )
+            except Exception as exc:  # pylint: disable=broad-except
+                LOGGER.exception(
+                    f"Failed notifying user: "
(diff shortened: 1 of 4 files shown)


binodpant:
will do version and changelog per merge

enterprise/tasks.py
```diff
@@ -40,14 +40,23 @@ def send_enterprise_email_notification(
...
+            try:
+                send_email_notification_message(
+                    item['user'],
+                    item['enrolled_in'],
+                    item['dashboard_url'],
+                    enterprise_customer_uuid,
+                    email_connection=email_conn,
+                    admin_enrollment=admin_enrollment,
+                )
+            except Exception as exc:  # pylint: disable=broad-except
```
binodpant, line 53: open to thoughts but I don't see much of a way around catching all exceptions, I have no idea what specific exceptions to catch here

tests/test_enterprise/test_tasks.py
```diff
@@ -106,6 +107,47 @@ def test_create_enrollment_task_no_create_duplicates(self, catalog_api_client_mo
...
+
+        # check exception is logged at ERROR level twice, in this case
+        # but causes no failures otherwise
+        with self.assertLogs('enterprise.tasks', level='ERROR') as log_cm:
+            send_enterprise_email_notification(
+                enterprise_customer.uuid,
+                True,
+                email_items,
+            )
+            self.assertEqual(2, len(log_cm.records))
```
binodpant, line 148: just checking in this case that two logger records got created

macdiesel (approved):
Cool, thank you!

enterprise/tasks.py
```diff
@@ -52,9 +56,10 @@ def send_enterprise_email_notification(
                 )
             except Exception as exc:  # pylint: disable=broad-except
                 LOGGER.exception(
-                    f"Failed notifying user: "
-                    f"enterprise_customer_uuid: {enterprise_customer_uuid}"
-                    f"of enterprise enrollment in course {course_name}",
+                    "Failed notifying user: {}"
```
binodpant, line 59: do not want to print email etc pii, so just printing username if available

PR #1317, opened 2021-08-02
fix: Fix Start my course URL for admin enrollment email

binodpant:
Should be proper learner portal course link, also add tests for both admin_enroll and self_enroll cases

### OPEN QUESTION!!
What about when learner portal is off for enterprise?  Should this url be made different in that case?

ANSWERED: we want the prior behavior of sending users direct to register or login, with next url being the courseware page url

JIRA: ENT-4789

### Testing Notes

After fixing ENT-4789

#### Existing learner:

Took me to this page when I click on Start my Course link (from devstack email file)
http://localhost:18000/login?next=/courses/course-v1%3AedX%2B808404707%2B2T2021/course%3F

which redirected me to
http://localhost:18000/courses/course-v1:edX+808404707+2T2021/course/

#### Pending learner:

Invited non edx learner (abc@mouse.com) to subscription
Then directly enrolled them using bulk enrollment into a course
Email link was this:
http://localhost:18000/register?next=/courses/course-v1%3AedX%2B909878%2B2T2021/course%3F

When visited it prompted me to register new account
I registered as abc@mouse.com as expected

Note: they have not activated the license yet.

#### Using manage learners to invite a new learner not yet in edX
I got an email with the activation link which works

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -467,6 +467,17 @@ def is_pending_user(user):
     return hasattr(user, 'user_email') and not hasattr(user, 'first_name')
 
 
+def get_learner_portal_url(enterprise_customer):
+    """
+    Learner portal url for enterprise_customer
+    """
+    return get_configuration_value_for_site(
+        enterprise_customer.site,
+        'ENTERPRISE_LEARNER_PORTAL_BASE_URL',
+        settings.ENTERPRISE_LEARNER_PORTAL_BASE_URL
+    )
+
+
 def serialize_notification_content(
     enterprise_customer,
     course_details,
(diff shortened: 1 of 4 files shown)


tests/test_enterprise/test_utils.py
```diff
@@ -238,3 +250,85 @@ def test_enroll_pending_licensed_users_in_courses_succeeds(self):
...
+        email_items = serialize_notification_content(
+            ent_customer,
+            course_details,
+            course_id,
+            users,
+            admin_enrollment=True,
+        )
+
+        def expected_email_item(user):
+            course_path = 'course/{course_id}'.format(course_id=course_details.get('course'))
```
binodpant, line 314: this is the learner portal url for the course. We need to determine/finalize what we want when learner portal is off for enterprise

binodpant: resolved: we are not going to use the learner portal link in the 'start my course' link. it's only used for the My Dashboard link still for bulk enrollment use case

enterprise/utils.py
```diff
@@ -495,40 +514,24 @@ def serialize_notification_content(
     """
     dashboard_url = None
     course_name = course_details.get('title')
+    course_path = '/courses/{course_id}/course'.format(course_id=course_id)
+    params = {}
+
     if admin_enrollment:
```
binodpant, line 520: For ref: this logic used to be in models.py until this PR
https://github.com/edx/edx-enterprise/pull/1299/files#diff-d950991245f878f58adc157895d6431af4385d522ab994aa25726e9abc2b434dL679

so other than dashboard_url being learner portal, all other behavior should be intact

tests/test_enterprise/test_utils.py
```diff
@@ -238,3 +250,70 @@ def test_enroll_pending_licensed_users_in_courses_succeeds(self):
...
+        return ent_customer, users, course_id, course_details
+
+    @ddt.unpack
+    @ddt.data(
+        (True, 'http://test.learner.portal'),
+        (False, None),
+    )
+    @mock.patch("enterprise.utils.get_configuration_value_for_site")
+    @mock.patch("enterprise.utils.get_learner_portal_url")
+    def test_serialize_notification_content(
```
edx-abolger, line 277: +1 for additional tests

enterprise/utils.py
edx-abolger: I copied some code from that PR to make my review simpler. I'm checking the new changes against this:
```
course_path = '/courses/{course_id}/course'.format(course_id=course_id)
        params = {}
        # add tap_hint if there is only one IdP attached with enterprise_customer
        if self.has_single_idp:
            params = {'tpa_hint': self.identity_providers.first().provider_id}

        elif self.has_multiple_idps and self.default_provider_idp:
            params = {'tpa_hint': self.default_provider_idp.provider_id}
        course_path = urlquote("{}?{}".format(course_path, urlencode(params)))

        lms_root_url = utils.get_configuration_value_for_site(
            self.site,
            'LMS_ROOT_URL',
            settings.LMS_ROOT_URL
        )
        destination_url = '{site}/{login_or_register}?next={course_path}'.format(
            site=lms_root_url,
            login_or_register='{login_or_register}',  # We don't know the value at this time
            course_path=course_path
        )
```

enterprise/utils.py
```diff
@@ -495,40 +514,24 @@ def serialize_notification_content(
...
-            settings.LMS_ROOT_URL
-        )
-        destination_url = '{site}/{login_or_register}?next={course_path}'.format(
-            site=lms_root_url,
-            login_or_register='{login_or_register}',  # We don't know the value at this time
-            course_path=course_path
-        )
+        dashboard_url = get_learner_portal_url(enterprise_customer)
+
+    # add tap_hint if there is only one IdP attached with enterprise_customer
```
edx-abolger, line 523: Nit: this should be tpa

enterprise/utils.py
```diff
@@ -542,9 +545,12 @@ def serialize_notification_content(
 
     email_items = []
     for user in users:
-        is_pending_user = hasattr(user, 'user_email') and not hasattr(user, 'first_name')
-        login_or_register = 'register' if is_pending_user else 'login'
-        destination_url = destination_url.format(login_or_register=login_or_register)
+        login_or_register = 'register' if is_pending_user(user) else 'login'
+        destination_url = '{site}/{login_or_register}?next={course_path}'.format(
+            site=lms_root_url,
+            login_or_register=login_or_register,
```
edx-abolger, line 551: Nice, and the refactor actually cleans up that weird format thing in the old code.

edx-abolger (approved):
LGTM, 1 nit.

enterprise/utils.py
binodpant: ah ok thanks for doing this

PR #1320, opened 2021-08-03
feat: Added support for django3.2

Jawayria:
JIRA: https://openedx.atlassian.net/browse/BOM-2692


tests/test_models.py
```diff
@@ -1739,55 +1738,6 @@ def test_string_conversion(self, method):
         assert expected_to_str == method(dsc)
 
 
-@ddt.ddt
-@mark.django_db
-class TestEnterpriseCustomerPluginConfiguration(unittest.TestCase):
```
awais786, line 1744: These tests are for `EnterpriseCustomerPluginConfiguration` abstract model. Django has removed abstract model instantiating with this error `Instantiating an abstract model now raises TypeError.` in [django3.2](https://docs.djangoproject.com/en/3.2/releases/3.2/#miscellaneous). Django [PR](https://github.com/django/django/pull/13267) in which it is implemented.

So no need to test this functionality here.

tests/test_migrations.py
```diff
@@ -2,11 +2,11 @@
 """
 Tests for migrations, especially potentially risky data migrations.
 """
+from six import StringIO
```
mzulqarnain1, line 5: You can use `from io import StringIO` instead.

mzulqarnain1 (changes requested):
There's a warning in Django 2.2 tests
```
RemovedInDjango30Warning: Remove the context parameter from MultiEmailField.from_db_value(). Support for it will be removed in Django 3.0.
```
Please fix that too.

awais786 (approved):

mzulqarnain1 (approved):

PR #1323, opened 2021-08-05
fix(integrated_channels): :mute: fix bug causing log noise due to cou…

edx-abolger:
…rse metadata not containing a start or end field.

**Merge checklist:**
- [N/A ] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files)
- [ N/A] `make static` has been run to update webpack bundling if any static content was updated
- [N/A ] `./manage.py makemigrations` has been run
- [N/A ] [Version](https://github.com/edx/edx-enterprise/blob/master/enterprise/__init__.py) bumped
- [x ] [Changelog](https://github.com/edx/edx-enterprise/blob/master/CHANGELOG.rst) record added
- [ N/A] Translations updated (see docs/internationalization.rst but also this isn't blocking for merge atm)

**Post merge:**
- [ N/A can wait until next normal bump] Tag pushed and a new [version](https://github.com/edx/edx-enterprise/releases) released
- [N/A ] After versioned build finishes in [GitHub Actions](https://github.com/edx/edx-enterprise/actions), verify version has been pushed to [PyPI](https://pypi.org/project/edx-enterprise/)
- [N/A ] PR created in [edx-platform](https://github.com/edx/edx-platform) to upgrade dependencies (including edx-enterprise)

### DIFF
--- a/integrated_channels/canvas/exporters/content_metadata.py
+++ b/integrated_channels/canvas/exporters/content_metadata.py
@@ -26,6 +26,7 @@ class CanvasContentMetadataExporter(ContentMetadataExporter):
         'course_code': 'key',
         'indexed': 'indexed'
     }
+    SKIP_KEY_IF_NONE = True
 
     LONG_STRING_LIMIT = 2000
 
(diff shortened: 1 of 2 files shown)


integrated_channels/canvas/exporters/content_metadata.py
```diff
@@ -78,3 +78,17 @@ def transform_indexed(self, content_metadata_item):  # pylint: disable=unused-ar
         Whether to make the course default to the public index or not.
         """
         return 1
+
+    def transform_start(self, content_metadata_item):
```
alex-sheehan-edx, line 82: nice fix 🥳

alex-sheehan-edx (approved):
LGTM :shipit:

PR #1324, opened 2021-08-05
feat: Safe encode/decode functions for course key reversible transform for use with LMS (cornerstone mainly)

binodpant:
This PR only addresses part of the requirements to make Cornerstone course pulls safe
- [x] util function to encode/decode reversibly and avoid unsafe chars like `. | \ / < > ” % & and whitespace`

There is an additional set of work, which will be done in a new PR, 
- [ ] uses this util in Cornerstone exporter to convert edx course keys into safe format, when returning data from cornerstone endpoint
- [ ] update tests to use the unsafe course keys to ensure we got coverage
- [ ] address the issue of how we will handle cases where courses were already sent to CSOD so we don't ask CSOD to create dup entries
- [ ] Double check on learner module if we need to take any action (esp if we look up courses on CSOD side, if we don't then it may not be an issue)
 
ENT-4663

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/utils.py
+++ b/integrated_channels/utils.py
@@ -3,13 +3,13 @@
 Utilities common to different integrated channels.
 """
 
+import base64
 import math
 import re
 from datetime import datetime, timedelta
 from itertools import islice
 from logging import getLogger
 from string import Formatter
-import base64
 
 import requests
 from six.moves import range
--- a/integrated_channels/utils.py
+++ b/integrated_channels/utils.py
@@ -31,7 +31,7 @@ def encode_course_key_into_base64(edx_course_key):
     Base64 encodes edx course key (string) into a form safe (string) for use with LMS such as Cornerstone
     e.g., Cornerstone does not allow some chars
     For Base64, the urlsafe version is used, since it only uses a pretty limited charset, minus the /
-    edX course keys allow these chars `ALLOWED_ID_CHARS = r'[\w\-~.:]'` per: opaque_keys/edx/locator.py
+    edX course keys allow these chars `ALLOWED_ID_CHARS` per: opaque_keys/edx/locator.py
     """
     if edx_course_key is None:
         raise ValueError('Cannot process an undefined edx_course_key')
--- a/integrated_channels/utils.py
+++ b/integrated_channels/utils.py
@@ -26,7 +26,7 @@ UNIX_MAX_DATE_STRING = '2038-01-19T03:14:07Z'
 LOGGER = getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


tests/test_integrated_channels/test_utils.py
```diff
@@ -20,6 +20,18 @@ class TestIntegratedChannelsUtils(unittest.TestCase):
     Test utility functions used by integration channels.
     """
 
+    @ddt.data(
```
binodpant, line 23: I picked some of these values from the prod replica

integrated_channels/utils.py
```diff
@@ -25,6 +26,23 @@
 LOGGER = getLogger(__name__)
 
 
+def encode_course_key_for_lms(edx_course_key):
```
mattdrayer, line 29: Suggestion:  You don't need to specify `for_lms` and `from_lms` here -- I would instead say `to_base64` and `from_base64`.  Whether or not these utils are used for encoding keys for passing to/from an LMS is sort of irrelevant.

binodpant: done, also added error detection now.. don't want to inadvertently process spaces only kinds of inputs since they translate to the same base64 value each time. I hope we dn't have such bad content keys but who knows right?

mattdrayer: Do we need to be so specific regarding the invalidity of the key?  You could probably just do a more general ValueError try/except around the actual base64 conversion and log-out the issue instead of packing it into the return statement.

binodpant: the only issue there is the '  ' case. That one always encodes successfully to a fixed string .. so wanted to avoid that. Although I could trim that before sending to base64 then we can also do what you mentioned.

binodpant:
My next steps were going to be:

- 1: apply this function to a new transform function in exporter for CSOD, sth like 

```
    def transform_key(self, content_metadata_item):
        """
        Encodes the course key into a base64 urlsafe format
        """
        edx_course_key = content_metadata_item.get('key')
        return encode_course_key_into_base64(edx_course_key)

```

- 2: update tests of cornerstone to also use these interesting course ids (I actually put some in the unit tests for ref)


But to make life easier , I am going to merge this PR and we will do a new PR for the rest of the work

tests/test_integrated_channels/test_utils.py
binodpant: example query I used 
```
SELECT id from wwc.course_overviews_courseoverview ov where id like '%/%' order by id desc limit 500; 
```

alex-sheehan-edx (approved):
LGTM :shipit:

PR #1326, opened 2021-08-06
bug: invalid CSOD related course keys are now encoded

alex-sheehan-edx:
[jira](https://openedx.atlassian.net/browse/ENT-4663)

CSOD customers cannot accept courses that contain keys with invalid chars- thus for any **invalid course key**  we export, encoded it. When exporting learner data, make sure to encode the course ID if it's invalid as well in order to find the right course.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/cornerstone/client.py
+++ b/integrated_channels/cornerstone/client.py
@@ -12,6 +12,7 @@ import requests
 from django.apps import apps
 
 from integrated_channels.integrated_channel.client import IntegratedChannelApiClient
+from integrated_channels.utils import convert_invalid_course_ids
 
 LOGGER = logging.getLogger(__name__)
 
@@ -77,6 +78,11 @@ class CornerstoneAPIClient(IntegratedChannelApiClient):
         json_payload = json.loads(payload)
         callback_url = json_payload['data'].pop('callbackUrl')
         session_token = json_payload['data'].pop('sessionToken')
+
+        # When exporting content metadata, we encode course keys that contain invalid chars.
+        course_id = json_payload['data'].get('courseId')
+        json_payload['data']['courseId'] = convert_invalid_course_ids(course_id)
+
         url = '{base_url}{callback_url}{completion_path}?sessionToken={session_token}'.format(
             base_url=self.enterprise_configuration.cornerstone_base_url,
             callback_url=callback_url,
--- a/integrated_channels/cornerstone/exporters/content_metadata.py
+++ b/integrated_channels/cornerstone/exporters/content_metadata.py
@@ -14,6 +14,7 @@ from enterprise.utils import get_closest_course_run, get_language_code
 from integrated_channels.integrated_channel.constants import ISO_8601_DATE_FORMAT
 from integrated_channels.integrated_channel.exporters.content_metadata import ContentMetadataExporter
 from integrated_channels.utils import (
+    convert_invalid_course_ids,
     get_duration_from_estimated_hours,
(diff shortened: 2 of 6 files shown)


integrated_channels/cornerstone/client.py
```diff
@@ -94,6 +101,17 @@ def create_course_completion(self, user_id, payload):  # pylint: disable=unused-
         )
         return response.status_code, response.text
 
+    def _convert_invalid_course_ids(self, course_id):
+        """
+        Regex check a course ID to see if it contains any invalid chars. If it does then encode the string, otherwise
+        return the original course ID.
+        """
+        re2 = re.compile(r"[|<>.&%\s\\/\“]+")
+        if re2.search(course_id):
```
mattdrayer, line 110: I guess I feel like, if we are converting course keys to send to CSOD, we should just convert all of the keys since it might look weird/buggy if some keys are in the traditional format and some keys are in the encoded format.  Thoughts?

integrated_channels/cornerstone/exporters/content_metadata.py
```diff
@@ -53,6 +55,32 @@ def export_force_all_catalogs(self):
         """
         return self.export(force_retrieve_all_catalogs=True)
 
+    def encode_invalid_keys(self, content_key):
```
mattdrayer, line 58: Why is this here when there is something similar in `client.py` above?  This seems like more of a utility function -- maybe it should be moved to `cornerstone/utils.py`

tests/test_integrated_channels/test_cornerstone/test_exporters/test_content_metadata.py
```diff
@@ -113,6 +114,50 @@ def test_transform_description(self, item_description, expected_description):
...
+            encode_course_key_into_base64('edx+|'),
+        ),
+        (
+            {'key': 'edx+/'},
+            encode_course_key_into_base64('edx+/'),
+        ),
+    )
+    @responses.activate
+    @ddt.unpack
+    def test_transform_ke (self, item_key, expected_id):
```
edx-abolger, line 153: Please add a test case where invalid characters appear in the beginning and middle of the keys, not just at the end.

integrated_channels/cornerstone/client.py
edx-abolger: The problem is that regenerating course keys has far reaching effects since I believe those keys are used as IDs on the Cornerstone side. It is possible any dropped and recreated courses would lose linked learner data and enrollments on the Cornerstone side. At a minimum, there would suddenly be two of every course and we'd have to handle that with additional manual clean up code. This solution is for backwards compatibility. (A comment on this in code would not be amiss, @alex-sheehan-edx )

edx-abolger (changes requested):
Small requests, should be easily resolvable

integrated_channels/cornerstone/client.py
mattdrayer: Makes perfect sense -- I neglected to consider the fact that there are already edX courses in CSOD tenants and we would have effectively abandoned those records on the CSOD side.

integrated_channels/cornerstone/exporters/content_metadata.py
alex-sheehan-edx: good point! Yeah, this should really be in a utils file. 👍

edx-abolger (approved):

PR #1328, opened 2021-08-09
bug: 404s on integrated channel metadata deletions should not error

alex-sheehan-edx:
[jira](https://openedx.atlassian.net/browse/ENT-4082)

When the integrated channels sends a delete request to external LMS' and receive a 404 meaning the course is not found, it shouldn't be treated as a failure, but instead a success. 

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/blackboard/client.py
+++ b/integrated_channels/blackboard/client.py
@@ -147,7 +147,9 @@ class BlackboardAPIClient(IntegratedChannelApiClient):
         BlackboardAPIClient._validate_channel_metadata(channel_metadata_item)
         external_id = channel_metadata_item.get('externalId')
         course_id = self._resolve_blackboard_course_id(external_id)
-        BlackboardAPIClient._validate_course_id(course_id, external_id)
+
+        if not course_id:
+            return HTTPStatus.OK.value, 'Course:{} already removed.'.format(external_id)
 
         LOGGER.info("Deleting course with courseId: %s", course_id)
         update_url = self.generate_course_update_url(course_id)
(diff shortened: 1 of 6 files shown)


integrated_channels/blackboard/client.py
```diff
@@ -147,7 +147,9 @@ def delete_content_metadata(self, serialized_data):
         BlackboardAPIClient._validate_channel_metadata(channel_metadata_item)
         external_id = channel_metadata_item.get('externalId')
         course_id = self._resolve_blackboard_course_id(external_id)
-        BlackboardAPIClient._validate_course_id(course_id, external_id)
+
```
binodpant, line 150: I suggested adding a unit test to test we don't fail when course id is not found from LMS side, for Bb client

tests/test_integrated_channels/test_moodle/test_client.py
```diff
@@ -101,8 +102,13 @@ def test_delete_content_metadata(self):
 
         client = MoodleAPIClient(self.enterprise_config)
         client._post = unittest.mock.MagicMock(name='_post', return_value=SUCCESSFUL_RESPONSE)  # pylint: disable=protected-access
-        client.get_course_id = unittest.mock.MagicMock(name='_get_course_id')
-        client.get_course_id.return_value = self.moodle_course_id
+        client._get_courses = unittest.mock.MagicMock(name='_get_courses')  # pylint: disable=protected-access
```
binodpant, line 105: I would also add a negative test case to cover missing course id from moodle side

binodpant (approved):
LGTM test coverage wise and also followed up on not needing this for canvas with @alex-sheehan-edx yesterday I will merge it in

PR #1331, opened 2021-08-16
fix/temp: Add logging to ECU and logistration to try to track where user discrepencies are occurring.

taliaedX:
**Description**:
per ENT-4690, we noticed the log prefixed user id would be an expected user, but the actual system role assignment was being done to our automated users (such as enterprise-worker). Through digging I noticed this wasn't just happening with role assignment but also through consent and the admin notification banner. All of these link back generically to the tpa pipeline and/or EnterpriseCustomerUser, so I added logging of the user id (LMS) to try to add some more points of data of either right or wrong user ids.

EDIT: This PR also became an explosion of linter updates due to me wanting to use f-strings and being mad at the linter for yelling about it.

**Merge checklist:**

**Post merge:**


enterprise/tpa_pipeline.py
```diff
@@ -61,6 +61,7 @@ def handle_enterprise_logistration(backend, user, **kwargs):
         **kwargs: Any remaining pipeline variables
 
     """
+    LOGGER.info(f'Beginning enterprise logistration for user {user.id}')
```
binodpant, line 64: in this log, it may be better to also replicate wording on like `LMS User Id` similar to the other one

binodpant (approved):
logging is cool, made a small wording suggestion on one of the logs, let's hope this is going to help us find out more info and cross correlate logs information.

enterprise/tpa_pipeline.py
taliaedX: yeah I can add that. I specifically added the word LMS there though because the log is on ECU which also has an id field and is of the user variety.

tests/test_enterprise/test_tasks.py
```diff
@@ -153,8 +153,7 @@ def test_enterprise_email_notification_failsafe(self, mock_send_notification, mo
     def test_send_enterprise_email_notification(self, mock_send_notification, mock_email_conn):
         enterprise_customer = EnterpriseCustomerFactory()
         pending_user = PendingEnterpriseCustomerUserFactory()
-        range_list = range(1, 10)
-        users = [UserFactory(username=f'user{user_id}') for user_id in range_list]
```
taliaedX, line 157: this was causing a linter error, also seems functionally irrelevant.

tests/test_enterprise/api_client/test_enterprise_catalog.py
```diff
@@ -226,7 +226,7 @@ def test_partial_successful_refresh_catalog():
     client = enterprise_catalog.EnterpriseCatalogApiClient('staff-user-goes-here')
     refreshed_catalogs, failed_to_refresh_catalogs = client.refresh_catalogs(EnterpriseCustomerCatalog.objects.all())
     assert failed_to_refresh_catalogs[0] == catalog2.uuid
-    assert len(refreshed_catalogs.items()) == 1
```
taliaedX, line 229: length of a dictionary and length of its items should be the same.

tests/test_management.py
```diff
@@ -137,6 +137,7 @@ class TestTransmitCourseMetadataManagementCommand(unittest.TestCase, EnterpriseM
     """
     Test the ``transmit_content_metadata`` management command.
     """
+    # pylint: disable=line-too-long
```
taliaedX, line 140: ^ = the strategy for being able to disable the line too long complaints everywhere in this file instead of it being interpreted bizarrely. (This is a known [open] bug with pylint)

enterprise/management/commands/update_role_assignments_with_customers.py
```diff
@@ -141,7 +141,7 @@ def _handle(self, batch_size, role_name, enterprise_customer_uuid=None, dry_run=
             customers_by_user_id = self._get_customers_by_user_id(ecu_ids_batch)
 
             assignments_by_user_id_role = self._get_role_assignments_by_user_id_and_role(
-                customers_by_user_id.keys(),
+                customers_by_user_id.keys(),  # pylint: disable=dict-keys-not-iterating
```
taliaedX, line 144: I think one of these "`.keys()` as a method parameter" had a sub method that still was not iterating through the keys, but was making a set from them, so it seemed fine to leave be.

tests/test_management.py
binodpant: interesting, this means we can remove this someday once we upgrade pylint I presume

binodpant (approved):
As long as this passes tests and pylint checks I am all set!  love the cleanup

tests/test_management.py
taliaedX: Not really no. The original usage was just specifying this on each line where it applied. However, that doesn't work because of the bug for some reason. 

I think enabling line-too-long at all is debatable, but if it's going to be enabled by the rules, currently anything long added to this file will be ignored, which may or may not be desired. 

To put it simply: instead of being able to specify the disable commands for the only 2 lines affected, I had to apply it to the entire class because of the pylint bug.

iloveagent57 (approved):
👏

enterprise/admin/actions.py
```diff
@@ -74,7 +74,7 @@ def refresh_catalog(self, request, queryset):  # pylint: disable=unused-argument
 
     # display catalog and task ids that were successfully started refreshing
     updated_message = ''
-    if refreshed_catalogs.items():
+    if refreshed_catalogs:
```
edx-abolger, line 77: Arent these things slightly different logically? I think your change is more correct, b/c otherwise line 77 could throw an Attribute error, but did the linter catch this?

taliaedX: Sorry I completely missed this.
Yes this was a complaint from the linter. It falls under the rule of using dictionary iterators without actually iterating.

In some of the instances, I just turned that rule off, but here since we're just checking for existence, items cannot exist without the base and the base doesn't validate as true without items, so it is equitable.

PR #1333, opened 2021-08-19
feat: adding customer_admin enrollment source

kiram15:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/migrations/0140_update_enrollment_sources.py
+++ b/enterprise/migrations/0140_update_enrollment_sources.py
@@ -1,18 +1,18 @@
 from django.db import migrations
 
-ENTERPRISE_ENROLLMENT_SOURCES = {
-    'Admin Enterprise Enrollment': 'customer_admin'
+ADDITIONAL_ENTERPRISE_ENROLLMENT_SOURCES = {
+    'Customer Admin Enrollment': 'customer_admin'
 }
 
-def add_enterprise_enrollment_source(apps, schema_editor):
+def add_new_enterprise_enrollment_source(apps, schema_editor):
     enrollment_sources = apps.get_model('enterprise', 'EnterpriseEnrollmentSource')
-    for name, slug in ENTERPRISE_ENROLLMENT_SOURCES.items():
+    for name, slug in ADDITIONAL_ENTERPRISE_ENROLLMENT_SOURCES.items():
         enrollment_sources.objects.update_or_create(name=name, slug=slug)
 
 
-def drop_enterprise_enrollment_source(apps, schema_editor):
+def drop_new_enterprise_enrollment_source(apps, schema_editor):
     enrollment_sources = apps.get_model('enterprise', 'EnterpriseEnrollmentSource')
-    enrollment_sources.objects.filter(name__in=ENTERPRISE_ENROLLMENT_SOURCES).delete()
+    enrollment_sources.objects.filter(name__in=ADDITIONAL_ENTERPRISE_ENROLLMENT_SOURCES).delete()
 
 class Migration(migrations.Migration):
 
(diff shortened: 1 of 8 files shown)


enterprise/migrations/0140_update_enrollment_sources.py
```diff
@@ -0,0 +1,26 @@
+from django.db import migrations
+
+ENTERPRISE_ENROLLMENT_SOURCES = {
```
edx-abolger, line 3: nit: might rename to 'ADDITIONAL_ENTERPRISE_ENROLLMENT_SOURCES'

enterprise/views.py
```diff
@@ -555,10 +555,13 @@ def create_enterprise_course_enrollment(request, enterprise_customer, course_id,
             enterprise_customer=enterprise_customer,
             user_id=request.user.id
         )
+        enterprise_enrollment_source = EnterpriseEnrollmentSource.get_source(
+            EnterpriseEnrollmentSource.ENROLLMENT_URL)
```
alex-sheehan-edx, line 559: I think we usually put closing parentheses on new lines... unless something's changed

tests/test_enterprise/api/test_views.py
```diff
@@ -3458,6 +3458,9 @@ def test_bulk_enrollment_in_bulk_courses_pending_licenses(
             self.assertEqual(expected_response, response_json)
         self.assertEqual(len(PendingEnrollment.objects.all()), expected_num_pending_licenses)
 
+        if expected_num_pending_licenses == 1:
```
edx-abolger, line 3461: Please add similar logic to check the other enrollment source you changed in the test of the view that isn't bulk enrollment.

edx-abolger (changes requested):
1 more test and I think this will be gtg

enterprise/migrations/0140_update_enrollment_sources.py
alex-sheehan-edx: If the dict has the keyword `additional` then perhaps also a change to the methods below as well? `add_new_enterprise_enrollment_source` or something similar. I haven't entirely convinced myself yet so feel free to push back.

enterprise/utils.py
```diff
@@ -1660,14 +1660,15 @@ def customer_admin_enroll_user(enterprise_customer, user, course_mode, course_id
...
+        enterprise_enrollment_source,
         license_uuid=None,
 ):
     """
     Get or Create the Enterprise Course Enrollment.
 
     If ``license_uuid`` present, will also create a LicensedEnterpriseCourseEnrollment record.
     """
-    source = enterprise_enrollment_source_model().get_source(enterprise_enrollment_source_model().ENROLLMENT_URL)
+    source = enterprise_enrollment_source
```
binodpant, line 1671: should  we validate or fail here if source is None?

enterprise/migrations/0140_update_enrollment_sources.py
```diff
@@ -0,0 +1,26 @@
...
+    'Admin Enterprise Enrollment': 'customer_admin'
```
binodpant, line 4: Name wise, Enterprise Admin Enrollment makes more sense to me, but open to thoughts (that's just me!)

binodpant: Or just `Customer Admin Enrollment`, since that is the name used all over anyway

enterprise/utils.py
```diff
@@ -1660,14 +1660,15 @@ def customer_admin_enroll_user(enterprise_customer, user, course_mode, course_id
 def get_create_ent_enrollment(
         course_id,
         enterprise_customer_user,
+        enterprise_enrollment_source,
```
binodpant, line 1663: Wow so no existing test for `get_create_ent_enrollment` fails when we add this? I have not checked but sure would be nice to have a unit test that tested this on its own (if too large scope for your PR we can file it away but if easy may not be a bad opportunity to add at least this one case!)

Yeah on quick check I don't see this being called from _any_ test, which makes me sad

binodpant: I think we have a good opp. to add one if it's missing... please push back if this extends scope in a large way!

edx-abolger (approved):

PR #1339, opened 2021-08-27
feat: Make title unique and pass it to the catalog service.

taliaedX:
### Description

This PR is here to ultimately allow us to use query titles in Algolia/the public enterprise catalog/"Explore Catalog"

In order to facilitate that, we need the existing query titles to be unique and to also actually get passed to the enterprise catalog service.

Actual changes of this PR:
- EnterpriseCatalogQuery model update: Makes the Title field nullable, blank, and unique (yes Mysql supports nullable & unique at the same time, other dbs do not.)
- If there are duplicates in the titles, it will append the id of the query and a random number
  - Note: We verified there are no duplicates on Stage or Prod. This is being added exclusively for devstack users. I also added the random generator just now because I realize you can't apply the same migration multiple times otherwise.
- Adds query_title as a parameter in the post_save action on EnterpriseCatalog and the migrate_enterprise_catalogs command. 

**Note:** This PR is dependent on https://github.com/edx/enterprise-catalog/pull/309

**Merge checklist:**

**Post merge:**


enterprise/migrations/0141_make_enterprisecatalogquery_title_unique.py
```diff
@@ -0,0 +1,34 @@
...
+    # I've verified there are no duplicates in Stage and Prod.
+    # This was brought up as a thing devs might need for their local devstack.
+    EnterpriseCatalogQuery = apps.get_model('enterprise', 'EnterpriseCatalogQuery')
+    duplicate_queries = EnterpriseCatalogQuery.objects.raw(
+        'SELECT cq1.title, cq1.id from enterprise_enterprisecatalogquery cq1 '
+        'JOIN (select count(*), title from enterprise_enterprisecatalogquery GROUP BY title having count(*) > 1) cq2 '
+        'ON cq1.title=cq2.title'
+    )
+    for query in duplicate_queries:
+        query.title = f'{query.title}{query.id}{randint(0, 600)}'
```
alex-sheehan-edx, line 16: what is the randint guaranteeing here?

alex-sheehan-edx (approved):
LGTM- is there an order of ops here with enterprise catalog vs enterprise side of things? I see you have default None type params so I guess this is backwards compatible, I just want to be sure.

enterprise/migrations/0141_make_enterprisecatalogquery_title_unique.py
taliaedX: In thinking about it, idk. I hit a weird case, but that was due to my own manipulation of the db so I don't think this even possible in the wild.

If we applied and then unapplied the migration (which does nothing to the titles, I can't roll those back), it wouldn't matter since the titles would no longer be the same so they wouldn't be duplicates.

enterprise/signals.py
```diff
@@ -255,7 +255,8 @@ def update_enterprise_catalog_data(sender, instance, **kwargs):     # pylint: di
     """
     catalog_uuid = instance.uuid
     catalog_query_uuid = str(instance.enterprise_catalog_query.uuid) if instance.enterprise_catalog_query else None
-
+    query_title = instance.enterprise_catalog_query.title if instance.enterprise_catalog_query and \
+        hasattr(instance.enterprise_catalog_query, 'title') else None
```
iloveagent57, line 259: NIt, consider
```python
query_title = getattr(instance.enterprise_catalog_query, 'title', None) if instance.enterprise_catalog_query else None
```

enterprise/migrations/0141_make_enterprisecatalogquery_title_unique.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 2.2.24 on 2021-08-27 17:56
+
+from django.db import migrations, models
+
+def force_unique_title(apps, schema_editor):
+    # I've verified there are no duplicates in Stage and Prod.
+    # This was brought up as a thing devs might need for their local devstack.
```
iloveagent57, line 7: Thank you for this note, and for thinking of dev environments (or maybe Open edX environments?).

iloveagent57 (approved):
Looks good, one nit.  don't forget changelog/version.

enterprise/migrations/0141_make_enterprisecatalogquery_title_unique.py
taliaedX: Alex S was the one to note this and yeah our thought was development, but definitely also may apply to Open edX!

taliaedX:
Oh yeah I forgot:
```

LGTM- is there an order of ops here with enterprise catalog vs enterprise side of things? I see you have default None type params so I guess this is backwards compatible, I just want to be sure.
```

To answer your question, this is in the PR description as a note: `Note: This PR is dependent on edx/enterprise-catalog#309` so yeah I would like the catalog pr to go first.

PR #1340, opened 2021-08-29
refactor: Refactor data-sharing consent GET and POST handlers...

iloveagent57:
...to not have too many statements, because readability matters.

Relates to https://openedx.atlassian.net/browse/ENT-4564 in that it will make writing unit tests
for new functionality less daunting.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -110,6 +110,45 @@ LMS_LOGIN_URL = urljoin(settings.LMS_ROOT_URL, '/login')
 ENTERPRISE_GENERAL_ERROR_PAGE = 'enterprise/enterprise_error_page_with_messages.html'
 
 
+CATALOG_API_CONFIG_ERROR_CODE = 'ENTGDS004'
+CUSTOMER_DOES_NOT_EXIST_ERROR_CODE = 'ENTGDS008'
+CONTENT_ID_DOES_NOT_EXIST_ERROR_CODE = 'ENTGDS000'
+LICENSED_ENROLLMENT_ERROR_CODE = 'ENTGDS100'
+NO_CONSENT_RECORD_ERROR_CODE = 'ENTGDS0002'
+REDIRECT_URLS_MISSING_ERROR_CODE = 'ENTGDS003'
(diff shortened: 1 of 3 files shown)


enterprise/views.py
```diff
@@ -110,6 +110,44 @@
...
+        'exception: {exception}'
+    ).format(error_message=error_message, error_code=error_code, exception=exception)
+
+    for key, value in kwargs.items():
+        log_message += ', {key}: {value}'.format(key=key, value=value)
+
+    if info_only:
+        LOGGER.info(log_message)
+    else:
+        LOGGER.error(log_message)
```
macdiesel, line 147: Should this be LOGGER.exception if the exception is present so we capture the stack trace?

enterprise/views.py
```diff
@@ -192,6 +230,16 @@ def render_page_with_error_code_message(request, context_data, error_code, log_m
     )
 
 
+def should_upgrade_to_licensed_enrollment(consent_record, license_uuid):
+    """
+    In the event that a learner did enroll into audit from B2C and they consented
+    on the data sharing consent page, we want to upgrade their audit enrollment
+    into verified when they view the course in the learner portal and hit
+    the DSC GET endpoint again.
```
macdiesel, line 238: Does this mean we end up with a side effecting GET? Same or similar to the one discussed here: https://edx-internal.slack.com/archives/CD3L7U2SZ/p1630002556001900?

macdiesel (approved):
2 nits/questions.

enterprise/views.py
iloveagent57: We're spitting out a lot of info in these log messages, I don't think a `logger.exception()` adds much.  Also, you can only get into these code paths by making a request directly to the GET or POST handler - there's nothing interesting higher up the stack, just Django request-handling stuff.

enterprise/views.py
iloveagent57: Yeah, it is.  This whole DSC flow is not great, particularly in how we hijacked it for licensed-enrollment.  It can be changed, but would require coordinated changes in frontend-app-learner-portal and license-manager.

PR #1343, opened 2021-09-02
feat: Integrated channels: ENT-4847 Correctly derive course_completed state for audit courses

binodpant:
ADR included

As of today, for all integrated channels, if course is audit we may never get completion_date set hence we never say course is complete. However a user may have completed all 'non gated' content. This change will allow that state to be reported as 'course_completed=True' for audit track courses.

Per chat with @ccmelo there _may_ be changes in how we report audit completion (perhaps based on `passed_timestamp` but since we may not always have that value for courses where passing grade hasn't been offered yet, learners who have fully seen all nongated content still may not have that field populated in the model `PersistentCourseGrade`

The evaluation of course_completed, now is no longer the responsibility of `get_learner_data_records` instead it's fed this value as an input after being evaluated outside of calling it via a new function which is now used for all Integ channels. This also smooths out how we evaluate this value currently ( some channels currently use `is_passing`, some don't )

If course enrollment for 'audit' mode is detected, fetched the `incomplete_count` which is then used to derive course_complete state. Rest of the information (like grades) is still obtained the same way 

ENT-4847

**Merge checklist:**

**Post merge:**


integrated_channels/moodle/exporters/learner_data.py
```diff
@@ -23,7 +23,7 @@ def get_learner_data_records(
             self,
             enterprise_enrollment,
             completed_date=None,
-            is_passing=False,
+            course_completed=False,
```
alex-sheehan-edx, line 26: update doc string?

integrated_channels/sap_success_factors/exporters/learner_data.py
```diff
@@ -30,8 +30,7 @@ def get_learner_data_records(
             enterprise_enrollment,
             completed_date=None,
             grade=None,
-            is_passing=False,
-            grade_percent=None
+            course_completed=False,
     ):
         """
         Return a SapSuccessFactorsLearnerDataTransmissionAudit with the given enrollment and course completion data.
```
alex-sheehan-edx, line 36: update? 👀

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -317,43 +326,52 @@ def export(self, **kwargs):
...
-                grade=grade,
-                is_passing=is_passing,
-                grade_percent=grade_percent
+                completed_date=completed_date_from_api,
+                grade=grade_from_api,
+                course_completed=is_course_completed(
+                    enterprise_enrollment,
+                    completed_date,
+                    is_passing_from_api,
+                    incomplete_count,
```
alex-sheehan-edx, line 373: `incomplete_count` will only be present if this is an audit enrollment, right?

alex-sheehan-edx (commented):
A couple of comments, some reworking and I think there are a couple test changes needed. Otherwise looking good

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -317,43 +326,52 @@ def export(self, **kwargs):
...
             # Apply the Source of Truth for Grades
-            grade = grade_from_api
-            completed_date = completed_date_from_api
-            is_passing = is_passing_from_api
             records = self.get_learner_data_records(
                 enterprise_enrollment=enterprise_enrollment,
-                completed_date=completed_date,
-                grade=grade,
-                is_passing=is_passing,
-                grade_percent=grade_percent
```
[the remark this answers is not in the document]
binodpant: updated!

integrated_channels/integrated_channel/exporters/learner_data.py
binodpant: yup, the function that uses this value validate for None, so if we pass None for audit case it will fail in case we pass None for audit

integrated_channels/cornerstone/exporters/learner_data.py
```diff
@@ -34,9 +33,6 @@ def get_learner_data_records(
 
         CornerstoneLearnerDataTransmissionAudit object should exit already if not then return None.
         """
-        course_completed = False
-        if completed_date is not None:
```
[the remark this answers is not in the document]
binodpant: take care of

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -317,43 +326,52 @@ def export(self, **kwargs):
                     )))
                 continue
 
+            # For audit courses, check if 100% completed
```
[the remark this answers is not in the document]
binodpant: refactored two pieces into helper functions

integrated_channels/cornerstone/exporters/learner_data.py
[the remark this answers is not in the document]
binodpant: gone!

enterprise/models.py
```diff
@@ -1651,13 +1651,17 @@ def is_active(self):
         """
         Returns True iff this enrollment is currently active.
         """
+        if not self.course_enrollment:
```
alex-sheehan-edx, line 1654: 👍

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -254,6 +264,68 @@ def export_unique_courses(self):
...
+        Fetch grades info using either certificate api, or grades api.
+        Note: This logic is going to be refactored, so that audit enrollments are treated separately
+        '''
+        is_audit_enrollment = enterprise_enrollment.is_audit_enrollment
+        lms_user_id = enterprise_enrollment.enterprise_customer_user.user_id
+        enterprise_customer_uuid = enterprise_enrollment.enterprise_customer_user.enterprise_customer.uuid
+        course_id = enterprise_enrollment.course_id
+
+        # For instructor-paced and non-audit courses, let the certificate determine course completion
+        if course_details.pacing == 'instructor' and not is_audit_enrollment:
```
alex-sheehan-edx, line 283: to be removed in the future! But thank you for moving this to a helper function 🥳

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -254,6 +264,68 @@ def export_unique_courses(self):
...
+                channel_name, enterprise_customer_uuid, lms_user_id, course_id,
+                f'_collect_grades_data finished with: CourseMode: {enterprise_enrollment.mode}, '
+                f' CompletedDate: {completed_date_from_api},'
+                f' Grade: {grade_from_api},'
+                f' IsPassing: {is_passing_from_api},'
+                f' Audit Mode?: {is_audit_enrollment}'
+            ))
+        return completed_date_from_api, grade_from_api, is_passing_from_api, grade_percent
+
+    def get_incomplete_content_count(self, enterprise_enrollment, channel_name):
```
alex-sheehan-edx, line 305: 👍

integrated_channels/utils.py
```diff
@@ -215,8 +217,8 @@ def is_already_transmitted(transmission, enterprise_enrollment_id, grade, subsec
         if subsection_id:
             already_transmitted = already_transmitted.filter(subsection_id=subsection_id)
 
-        if already_transmitted.latest('id') and getattr(already_transmitted.latest('id'), 'grade', None) == grade:
-            return True
+        latest_transmitted_tx = already_transmitted.latest('id')
+        return latest_transmitted_tx and getattr(latest_transmitted_tx, 'grade', None) == grade
     except transmission.DoesNotExist:
         pass
```
alex-sheehan-edx, line 223: can we just return false here?

integrated_channels/utils.py
```diff
@@ -334,3 +336,36 @@ def refresh_session_if_expired(oauth_access_token_function, session=None, expire
...
+    """
+    for seat in course_run.get('seats', []):
+        if seat.get('type') == 'verified':
+            return parse_datetime_handle_invalid(seat.get('upgrade_deadline'))
+    return None
+
+
+def is_course_completed(enterprise_enrollment, completed_date, is_passing, incomplete_count):
+    '''
+    For non audit, this requires passing and completed_date
```
alex-sheehan-edx, line 353: nice doc string

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -317,43 +387,28 @@ def export(self, **kwargs):
...
-                    .format(
-                        completed_date=completed_date_from_api,
-                        grade=grade_from_api,
-                        is_passing=is_passing_from_api
-                    )))
+            # For audit courses, check if 100% completed
+            # which we define as: no non-gated content is remaining
+            incomplete_count = self.get_incomplete_content_count(enterprise_enrollment, channel_name)
+
+            completed_date_from_api, grade_from_api, is_passing_from_api, grade_percent = self.get_grades_summary(
```
alex-sheehan-edx, line 394: maybe add a comment here on how this collects data from only verified enrollments?

alex-sheehan-edx (approved):
A couple of nits, and I think there's one more doc string to update, but is look good to me after those small changes!

enterprise/models.py
binodpant: yeah the tests were not passing without this anyway

integrated_channels/utils.py
binodpant: yeah think so basically the pass can also just be return False. If it fails in any other way rn it will fail anyway

integrated_channels/integrated_channel/exporters/learner_data.py
binodpant: I did add docstring to the function itself, but this function will also return data for audit mode, just that it won't call the cert api, as of now so not sure what I can write here? but I did add to the docstring of `get_grades_summary`

PR #1347, opened 2021-09-09
feat: get_remote_id accepts optional idp_id, used by integrated channels ENT-4784

binodpant:
Commits

- [X] get_remote_id gets an optional idp_id param which is used to shortcut locating of provider
- [X] django form and model changes to accept optional idp_id field
- [x] Integrated channels code uses id_id param when calling get_remote_id

Screen Shot 2021-09-09 at 1 29 34 PM

ENT-4784

**Merge checklist:**

**Post merge:**


tests/test_models.py
```diff
@@ -2112,8 +2141,8 @@ def test_enable_compression_clean(
                 config.clean()
             data_types = ", ".join(EnterpriseCustomerReportingConfiguration.ALLOWED_NON_COMPRESSION_DATA_TYPES)
             assert context.exception.messages[0] == \
-                   f'Compression can only be disabled for the following data types: {data_types} and ' \
-                   f'delivery method: {EnterpriseCustomerReportingConfiguration.DELIVERY_METHOD_SFTP}'
+                f'Compression can only be disabled for the following data types: {data_types} and ' \
```
binodpant, line 2144: autoformatter did this

mattdrayer (approved):

integrated_channels/blackboard/migrations/0007_auto_20210909_1536.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.24 on 2021-09-09 15:36
+
```
mattdrayer, line 2: Why so many migration files for a single PR?

mattdrayer: oh, nvm, they are for different models.py files for different integrated channels

mattdrayer (approved):

integrated_channels/blackboard/migrations/0007_auto_20210909_1536.py
binodpant: yeah idk, when I ran migrate, it generated a new one for each model, is that even typical?

binodpant: I am guessing since they are in different folders?

binodpant: yeah since each subclass is in its own folder and migrations also reside in multiple folders

alex-sheehan-edx (approved):
LGTM 👍

PR #1348, opened 2021-09-10
feat: pass idp_id to get_remote_id for degreed

binodpant:
After a recent change, we can now pass idp_id (the slug of the idp provider) into the get_remote_id() call. If the Integrated channel config does not provide this value it will be passed as None or empty and we get toda's behavior. If we do set it though, we can now resolve the correct remote id based on idp slug.

I am hoping this will fix issues with not being able to send learner data completions for Degreed going forward. It won't fix previously missed completions. we have to find a different way to re-send those.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/degreed/exporters/learner_data.py
+++ b/integrated_channels/degreed/exporters/learner_data.py
@@ -34,7 +34,9 @@ class DegreedLearnerExporter(LearnerExporter):
         """
         # Degreed expects completion dates of the form 'yyyy-mm-dd'.
         completed_timestamp = completed_date.strftime("%F") if isinstance(completed_date, datetime) else None
-        if enterprise_enrollment.enterprise_customer_user.get_remote_id() is not None:
+        if enterprise_enrollment.enterprise_customer_user.get_remote_id(
+            self.enterprise_configuration.idp_id
+        ) is not None:
             DegreedLearnerDataTransmissionAudit = apps.get_model(
                 'degreed',
                 'DegreedLearnerDataTransmissionAudit'
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * Nothing
 
+[3.28.7]
+---------
+* Degreed integrated channel now uses idp_id explicitly when calling get_remote_id()
+
 [3.28.6]
 ---------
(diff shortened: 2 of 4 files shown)


integrated_channels/degreed/exporters/learner_data.py
```diff
@@ -34,7 +34,9 @@ def get_learner_data_records(
         """
         # Degreed expects completion dates of the form 'yyyy-mm-dd'.
         completed_timestamp = completed_date.strftime("%F") if isinstance(completed_date, datetime) else None
-        if enterprise_enrollment.enterprise_customer_user.get_remote_id() is not None:
+        if enterprise_enrollment.enterprise_customer_user.get_remote_id(
```
mattdrayer, line 37: Do we need to include a test?

mattdrayer:
For missed completions, do you mean there are records in the LTA table showing succesfull transmission, but the records have not actually been transmitted?  In the case of existing audit records the only way to re-transmit would be to delete those audit records, then the system would re-transmit.

binodpant:
> For missed completions, do you mean there are records in the LTA table showing succesfull transmission, but the records have not actually been transmitted? In the case of existing audit records the only way to re-transmit would be to delete those audit records, then the system would re-transmit.

exactly! I think that’s what we’d need to do. I have never even tried to change stuff in prod

integrated_channels/degreed/exporters/learner_data.py
binodpant: totally, going to emulate what I did for sapsf shortly

binodpant:
> > For missed completions, do you mean there are records in the LTA table showing succesfull transmission, but the records have not actually been transmitted? In the case of existing audit records the only way to re-transmit would be to delete those audit records, then the system would re-transmit.
> 
> exactly! I think that’s what we’d need to do. I have never even tried to change stuff in prod

assuming my fix works we can try retransmission after this goes in?

alex-sheehan-edx (approved):
LGTM (with a test please)

integrated_channels/degreed/exporters/learner_data.py
binodpant: done

mattdrayer (approved):

PR #1351, opened 2021-09-14
feat: :goal_net: catch/report status code >=400 for create_course_completion calls

binodpant:
Applies to all channels and we get free logging from base class

Also renamed `payload` to `exporter` in transmit() method docstrings

ENT-4929

there is a except block that catches and handles logging for ClientError around the completion call, so I took advantage of that:
```
except ClientError as client_error:
```
does not fix any existing issue, this is logging-only that makes up for missing logs in some cases when we don't necessarily catch and re-raise/log >=400 cases

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/transmitters/learner_data.py
+++ b/integrated_channels/integrated_channel/transmitters/learner_data.py
@@ -4,6 +4,7 @@ Generic learner data transmitter for integrated channels.
 """
 
 import logging
+from http import HTTPStatus
 
 from django.apps import apps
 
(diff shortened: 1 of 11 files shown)


integrated_channels/integrated_channel/transmitters/learner_data.py
```diff
@@ -219,6 +220,9 @@ def transmit(self, payload, **kwargs):
                     getattr(learner_data, kwargs.get('remote_user_id')),
                     serialized_payload
                 )
+                if code >= HTTPStatus.BAD_REQUEST.value:
```
binodpant, line 223: torn on if we need that message or we can just use `body` here

alex-sheehan-edx: Given that we've seen mixed response types coming from the differing LMS', I feel like it would be worth to have some constant identifying chars with the log.

alex-sheehan-edx: I see, so this catches anytime where a `400+` slips through any of the channels without us raising a client error. Okay, I like this approach 👍

alex-sheehan-edx (approved):
lgtm

binodpant:
> lgtm

yup on the code in logs, I think `handle_transmission_error` already prints it out

PR #1354, opened 2021-09-16
feat: fill in now() for completed date for audit: ENT-4918

binodpant:
if audit course completed, and no date from api, then fill in with now() for completed audit courses
this is needed since we get None completed_date for audit courses in some cases when we decide to call it completed based on incomplete_count == 0

JIRA: https://openedx.atlassian.net/browse/ENT-4918

if for some rason a date does come back from api it still honors it
if audit complete is not completed (based on incomplete_count), it still avoid filling in
if track isn't even audit, it avoid filling date in

so only sets date to now() when audit + completed + date_not_coming_from_api

ENT-4918

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/exporters/learner_data.py
+++ b/integrated_channels/integrated_channel/exporters/learner_data.py
@@ -329,7 +329,7 @@ class LearnerExporter(Exporter):
         is_audit_enrollment = enterprise_enrollment.is_audit_enrollment
 
         # The decision to not get incomplete count for non audit enrollments can be questioned
-        # Right now we don't use this number for non audit. But this conditoin can be just removed
+        # Right now we don't use this number for non audit. But this condition can be just removed
         # if we decide we need this for non audit enrollments too
         if not is_audit_enrollment:
             return incomplete_count
--- a/integrated_channels/integrated_channel/exporters/learner_data.py
+++ b/integrated_channels/integrated_channel/exporters/learner_data.py
@@ -269,6 +269,7 @@ class LearnerExporter(Exporter):
         course_details,
         enterprise_enrollment,
         channel_name,
+        incomplete_count=None,
     ):
         '''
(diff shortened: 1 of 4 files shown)


integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -312,6 +327,10 @@ def get_incomplete_content_count(self, enterprise_enrollment, channel_name):
         '''
         incomplete_count = None
         is_audit_enrollment = enterprise_enrollment.is_audit_enrollment
+
+        # The decision to not get incomplete count for non audit enrollments can be questioned
+        # Right now we don't use this number for non audit. But this conditoin can be just removed
```
alex-sheehan-edx, line 332: `condition`?

integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -269,13 +269,17 @@ def get_grades_summary(
         course_details,
         enterprise_enrollment,
         channel_name,
+        incomplete_count=None,
```
alex-sheehan-edx, line 272: Remind me who the consumer of this is and why it's getting added?

alex-sheehan-edx (approved):
LGTM other than one minor comment nit and a context question!

integrated_channels/integrated_channel/exporters/learner_data.py
binodpant: it's used to decide when to fill in the date automatically, among other things, later in the method. for audit we only want to fill if course is complete based on `incomplete_count==0`

binodpant: we don't use it for non audit right now, I think we can further refactor so audit based code is compeletely separate from non audit...

PR #1359, opened 2021-09-21
fix: Moodle: Integrated channel experiences "unknown error" scenario

johnnagro:
- adding additional logging to better understand scope of error type and customer impact
- ensure parsed timestamp represents UTC time even if system's timezone is something else (test stability)

[ENT-4902](https://openedx.atlassian.net/browse/ENT-4902)

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/moodle/client.py
+++ b/integrated_channels/moodle/client.py
@@ -36,14 +36,10 @@ def moodle_request_wrapper(method):
             # This cannot be converted to JSON thus the above fails miserably.
             # The above can fail with different errors depending on the format of the returned page.
             # Moodle of course does not tell us what is wrong in any part of this HTML.
-            log_format_keys = {'method': method.__name__,
-                               'code': response.status_code,
-                               'enterprise_configuration_id': self.enterprise_configuration.id}
-            log_format_string = '''\
-                                Moodle API task "{method}" for enterprise_configuration "{enterprise_configuration_id}"
-                                failed due to unknown error with code "{code}".
-                                '''
-            raise ClientError(log_format_string.format(**log_format_keys), response.status_code) from error
+            log_msg = (f'Moodle API task "{method.__name__}" '
+                       f'for enterprise_customer_uuid "{self.enterprise_configuration.enterprise_customer.uuid}" '
+                       f'failed due to unknown error with code "{response.status_code}".')
+            raise ClientError(log_msg, response.status_code) from error
         if isinstance(body, list):
             # On course creation (and ONLY course creation) success,
             # Moodle returns a list of JSON objects, because of course it does.
--- a/integrated_channels/moodle/client.py
+++ b/integrated_channels/moodle/client.py
@@ -36,8 +36,8 @@ def moodle_request_wrapper(method):
             # This cannot be converted to JSON thus the above fails miserably.
             # The above can fail with different errors depending on the format of the returned page.
             # Moodle of course does not tell us what is wrong in any part of this HTML.
-            raise ClientError('Moodle API task "{method}" failed due to unknown error.'.format(
-                method=method.__name__), response.status_code
+            raise ClientError('Moodle API task "{method}" for enterprise_configuration "{enterprise_configuration_id}" failed due to unknown error with code "{code}".'.format(
+                method=method.__name__, code=response.status_code, enterprise_configuration_id=self.enterprise_configuration.id), response.status_code
             ) from error
         if isinstance(body, list):
             # On course creation (and ONLY course creation) success,
--- a/integrated_channels/moodle/client.py
+++ b/integrated_channels/moodle/client.py
@@ -36,10 +36,10 @@ def moodle_request_wrapper(method):
             # This cannot be converted to JSON thus the above fails miserably.
             # The above can fail with different errors depending on the format of the returned page.
             # Moodle of course does not tell us what is wrong in any part of this HTML.
-            log_format_keys = { 'method': method.__name__, 
-                                'code': response.status_code, 
+            log_format_keys = { 'method': method.__name__,
+                                'code': response.status_code,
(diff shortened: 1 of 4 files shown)


integrated_channels/moodle/client.py
```diff
@@ -36,8 +36,8 @@ def inner(self, *args, **kwargs):
             # This cannot be converted to JSON thus the above fails miserably.
             # The above can fail with different errors depending on the format of the returned page.
             # Moodle of course does not tell us what is wrong in any part of this HTML.
-            raise ClientError('Moodle API task "{method}" failed due to unknown error.'.format(
-                method=method.__name__), response.status_code
+            raise ClientError('Moodle API task "{method}" for enterprise_configuration "{enterprise_configuration_id}" failed due to unknown error with code "{code}".'.format(
+                method=method.__name__, code=response.status_code, enterprise_configuration_id=self.enterprise_configuration.id), response.status_code
```
johnnagro, line 40: i wanted to share some information on the customer impacted via the enterprise config record being used

integrated_channels/moodle/exporters/content_metadata.py
```diff
@@ -103,7 +104,7 @@ def transform_start(self, content_metadata_item):
         """
         start_date = content_metadata_item.get('start', None)
         if start_date:
-            return int(parse(start_date).timestamp())
+            return int(parse(start_date).replace(tzinfo=timezone.utc).timestamp())
```
johnnagro, line 107: my local system's timezone was causing these timestamps to return with a 5 hour offset in testing. i changed this to ensure we're always using a UTC datetime

integrated_channels/moodle/client.py
```diff
@@ -36,9 +36,14 @@ def inner(self, *args, **kwargs):
...
-                method=method.__name__), response.status_code
-            ) from error
+            log_format_keys = {'method': method.__name__,
+                               'code': response.status_code,
+                               'enterprise_configuration_id': self.enterprise_configuration.id}
+            log_format_string = '''\
+                                Moodle API task "{method}" for enterprise_configuration "{enterprise_configuration_id}"
+                                failed due to unknown error with code "{code}".
+                                '''
+            raise ClientError(log_format_string.format(**log_format_keys), response.status_code) from error
```
alex-sheehan-edx, line 46: Love this logging!

alex-sheehan-edx: I think this looks great, and works perfectly well- but in general we're trying to trend towards f strings instead of `format` and I think it might make things a little clearer here. Feel free to format this a bit more but my thoughts are something along the lines of 
```
       log_format_string = (f'Moodle API task "{method.__name__}" for enterprise_configuration "{self.enterprise_configuration.id}}" '
                             f'failed due to unknown error with code "{response.status_code}".')
```

alex-sheehan-edx: I'm also curious if config ID is the best identifier here- obviously that field does exist in the table, but perhaps a more helpful way to easily know which configuration customer would be to log `enterprise_customer_id`. Maybe we could do both? IDK, what are your thoughts? Only because we don't do much with the config IDs and they're not readily available unless you're querying read replica

binodpant: I have recently been using that `generate_formatted_log` as much as possible, sometimes we can't fill in all the info but mostly it works great, and makes logs consistent too. and the `message` part of that I now use f`` strings yes generally

integrated_channels/moodle/client.py
```diff
@@ -36,9 +36,10 @@ def inner(self, *args, **kwargs):
             # This cannot be converted to JSON thus the above fails miserably.
             # The above can fail with different errors depending on the format of the returned page.
             # Moodle of course does not tell us what is wrong in any part of this HTML.
-            raise ClientError('Moodle API task "{method}" failed due to unknown error.'.format(
-                method=method.__name__), response.status_code
-            ) from error
+            log_msg = (f'Moodle API task "{method.__name__}" '
```
binodpant, line 39: love this!

integrated_channels/moodle/client.py
```diff
@@ -36,9 +36,10 @@ def inner(self, *args, **kwargs):
...
+                       f'for enterprise_customer_uuid "{self.enterprise_configuration.enterprise_customer.uuid}" '
```
binodpant, line 40: let's hope this logging helps. It does contain the enterprise customer id but will we know which catalog, course etc was being processed? I guess this wrapper currently cannot access those so.. fwiw this logging on its own is useful. I just wonder if we also need more logging around other places (we can do this later too) to pinpoint course, catalog, user etc. The `generate_formatted_log` is an attempt to standardize that .. but it cannot always be used without sourcing all the required inputs

binodpant (approved):
lgtm!

integrated_channels/moodle/client.py
binodpant: also if this is an example of this 
```
 response = self._get_courses(serialized_data['courses[0][shortname]'])
  File "/edx/app/edxapp/venvs/edxapp/lib/python3.8/site-packages/integrated_channels/moodle/client.py", line 39, in inner
    raise ClientError('Moodle API task "{method}" failed due to unknown error.'.format(
integrated_channels.exceptions.ClientError: Moodle API task "_get_courses" failed due to unknown error.
```

then perhaps printing out what we can of the response may also be helpful?

alex-sheehan-edx: `then perhaps printing out what we can of the response or response code from moodle`

> Moodle of course does not tell us what is wrong in any part of this HTML.

Sounds like Talia found out that the response body really doesn't give us much. The code is getting logged here, and I'm not entirely sure if we're able to get more (useful) information out of it.

alex-sheehan-edx (approved):
looks great to me!

PR #1361, opened 2021-09-22
feat: ENT-4936 | adding management command to backfill role assignments

christopappas:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * Nothing
 
+[3.28.20]
+---------
+* feat: Added enterprise_learner role assignment backfill management command
+
 [3.28.19]
 ---------
 * fix: additional Moodle field changes
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.28.19"
+__version__ = "3.28.20"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"
--- a/enterprise/management/commands/backfill_learner_role_assignments.py
+++ b/enterprise/management/commands/backfill_learner_role_assignments.py
@@ -1,0 +1,94 @@
+"""
+Management command for assigning enterprise_learner roles
+to existing linked enterprise users that are missing them.
+"""
+
+import logging
+from time import sleep
+
+from django.contrib import auth
+from django.core.management.base import BaseCommand
+
+from enterprise.constants import ENTERPRISE_LEARNER_ROLE
+from enterprise.models import EnterpriseCustomerUser, SystemWideEnterpriseRole, SystemWideEnterpriseUserRoleAssignment
+from enterprise.utils import batch
+
+log = logging.getLogger(__name__)
+User = auth.get_user_model()
+
+
+class Command(BaseCommand):
+    """
+    Management command for assigning enterprise_learner roles to existing enterprise users.
+
+    Example usage:
+        $ ./manage.py backfill_learner_role_assignments
+    """
+    help = 'Assigns enterprise_learner role to linked enterprise users missing them.'
+
+    def add_arguments(self, parser):
+        """
+        Entry point for subclassed commands to add custom arguments.
+        """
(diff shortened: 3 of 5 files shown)


enterprise/management/commands/backfill_role_assignments.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def backfill_role_assignments(self, options):
+        """
+        Assigns enterprise_learner role to users.
+        """
+        batch_limit = options['batch_limit']
+        batch_sleep = options['batch_sleep']
+
+        role = SystemWideEnterpriseRole.objects.get(name=ENTERPRISE_LEARNER_ROLE)
+        ecus = EnterpriseCustomerUser.objects.select_related('enterprise_customer').filter(linked=True).iterator()
```
christopappas, line 58: just FYI: `linked=True` here is actually redundant with the fact we're using `.objects.` I believe. Seems like the model manager (EnterpriseCustomerUserManager) by default uses `linked=True`. Explicit > Implicit though #python

enterprise/management/commands/backfill_role_assignments.py
```diff
@@ -0,0 +1,94 @@
...
+                log.info(
+                    'Created SystemWideEnterpriseUserRoleAssignment %s for enterprise user %s',
+                    role_assignment, ecu
+                )
+            else:
+                log.info(
+                    'Did not create role assignment for enterprise user %s', ecu
+                )
+
+            if records_processed % batch_limit:
```
adamstankiewicz, line 83: Does `records_processed` need to be incremented?

enterprise/management/commands/backfill_role_assignments.py
adamstankiewicz, line 58: [clarification] Should the batching apply to the querying of `EnterpriseCustomerUser` records? As is, we'll be querying for all of them at once, no? If so, is this OK?

enterprise/management/commands/backfill_role_assignments.py
christopappas: good catch! will add

enterprise/management/commands/backfill_role_assignments.py
christopappas: I'll reread the docs, but my thought that's where `.iterator()` is valuable, because you don't get the whole table at once

christopappas: it definitely doesn't _cache_ the whole table in the queryset, but now i'm doubting myself on what it does at the DB level

christopappas: `iterator()` will prevent crazy amount of memory being taken for the application, but not DB caching. That will still happen. I'll change this to be a more traditional batch

enterprise/management/commands/backfill_role_assignments.py
```diff
@@ -0,0 +1,94 @@
...
+                else:
+                    log.info(
+                        'Did not create role assignment for enterprise user %s', ecu
+                    )
+
+            sleep(batch_sleep)
+
+    def handle(self, *args, **options):
+        """
+        Entry point for managment command execution.
```
adamstankiewicz, line 88: nit: `managment` ➡️  `management`

enterprise/management/commands/backfill_role_assignments.py
```diff
@@ -0,0 +1,94 @@
...
+log = logging.getLogger(__name__)
+User = auth.get_user_model()
+
+
+class Command(BaseCommand):
+    """
+    Management command for assigning enterprise_learner roles to existing enterprise users.
+
+    Example usage:
+        $ ./manage.py backfill_role_assignments
```
adamstankiewicz, line 25: nit: maybe rename the command to `backfill_learner_role_assignments` just to be more explicit?

adamstankiewicz (approved):
2 small, non-blocking nits but otherwise LGTM!

enterprise/management/commands/backfill_role_assignments.py
christopappas: 🍝

PR #1363, opened 2021-09-23
feat: allow customer configuration to disable learner data transmissions

alex-sheehan-edx:
[Jira](https://openedx.atlassian.net/browse/ENT-4914)

updated Django admin form
image

What:
allowing enterprise configurations to disable all learner data transmission. Log if transmissions are skipped.

**Merge checklist:**

**Post merge:**


tests/test_integrated_channels/test_integrated_channel/test_transmitters/test_learner_data.py
```diff
@@ -108,3 +108,33 @@ def test_raises_client_error_on_status_code(self, is_already_tx, mock_lms_id):
...
+        self.learner_transmitter.client.create_course_completion = Mock()
+
+        LearnerExporterMock = LearnerExporter
+        self.learner_transmitter.single_learner_assessment_grade_transmit(
+            LearnerExporterMock,
+            remote_user_id='user_id'
+        )
+        assert not self.learner_transmitter.client.create_assessment_reporting.called
+
+        self.learner_transmitter.assessment_level_transmit(
```
alex-sheehan-edx, line 129: I could theoretically call both assessment level transmission jobs and then assert that `create_assessment_reporting` is never called but I'm asserting twice for readability

integrated_channels/integrated_channel/transmitters/learner_data.py
```diff
@@ -62,37 +62,46 @@ def single_learner_assessment_grade_transmit(self, exporter, **kwargs):
         )
         kwargs.update(channel_name=app_label)
 
-        # Even though we're transmitting a learner, they can have records per assessment (multiple per course).
-        for learner_data in exporter.single_assessment_level_export(**kwargs):
+        if not self.enterprise_configuration.disable_learner_data_transmissions:
```
johnnagro, line 65: would it be clearer to use an if block around the log + an early return rather than an if around the whole function?

integrated_channels/integrated_channel/transmitters/learner_data.py
```diff
@@ -115,56 +124,65 @@ def assessment_level_transmit(self, exporter, **kwargs):
...
-
-            try:
-                code, body = self.client.create_assessment_reporting(
-                    getattr(learner_data, kwargs.get('remote_user_id')),
-                    serialized_payload
-                )
-                LOGGER.info(generate_formatted_log(
-                    app_label, enterprise_customer_uuid, lms_user_id, learner_data.course_id,
-                    'create_assessment_reporting successfully completed for EnterpriseCourseEnrollment {}'.format(
+        if not self.enterprise_configuration.disable_learner_data_transmissions:
```
johnnagro, line 127: similarly here, perhaps it would be more clear to use an `if` block that concisely logs + returns early before the meat of the method

integrated_channels/integrated_channel/transmitters/learner_data.py
```diff
@@ -185,68 +203,78 @@ def transmit(self, payload, **kwargs):
...
-                    serialized_payload
-                )
-                if code >= HTTPStatus.BAD_REQUEST.value:
-                    raise ClientError(f'Client create_course_completion failed: {body}', code)
-
-                LOGGER.info(generate_formatted_log(
-                    app_label, enterprise_customer_uuid, lms_user_id, learner_data.course_id,
-                    'Successfully sent completion status call for enterprise enrollment {}'.format(
-                        enterprise_enrollment_id,
+        if not self.enterprise_configuration.disable_learner_data_transmissions:
```
johnnagro, line 207: same

johnnagro (commented):
good work - i think this code accomplishes the goal. i left some feedback on how we might make the code more clear/readable by using an early return rather than a branch in a few places. i find its easier to understand since the if condition, logging, and early return logic are all in one place rather than wrapped around the body of the method. it also avoids adding more indentation to the body. happy to discuss further.

integrated_channels/integrated_channel/models.py
```diff
@@ -78,6 +78,12 @@ class EnterpriseCustomerPluginConfiguration(TimeStampedModel):
             "selected from the newest catalog."
         ),
     )
+    disable_learner_data_transmissions = models.BooleanField(
+        default=False,
+        verbose_name="Learner Data Transmission Feature Gate",
```
binodpant, line 83: will this be better named like a verb similar to our other booleans? such as 

```
Disable Learner Data Transmission
```

e.g., we have things like `Enable audit data reporting` , `Hide course original price`

so just thinking it will look a little bit more consistent

integrated_channels/integrated_channel/transmitters/learner_data.py
alex-sheehan-edx: Fair enough!

integrated_channels/integrated_channel/models.py
alex-sheehan-edx: consistent and I think it's clearer to the user who's toggling this in Django Admin

binodpant (approved):
I am good once John's feedback and the label change!

johnnagro (approved):
lgtm ⛵

PR #1367, opened 2021-09-30
BUG: dropping not updated catalogs results in expected behavior 

alex-sheehan-edx:
WIP - doc strings need to be updated. Pointers welcomed with open arms

[jira](https://openedx.atlassian.net/browse/ENT-5013)

TLDR:
enterprise customers' catalogs are being checked for the most recent time they've been updated before we fetch all the content metadata under the catalog. When a catalog's been found to not have any changes, we were explicitly excluding all courses contained within them from the list of exported content.

 Unfortunately, the transmitter uses the existence (or lack thereof) of content items within the exporter's payload  to determine if:
 - a catalog needs creating (is present within the list courses generated by the exporter and not present with the history of transmitted objects)
 - a catalog needs deleting (is present within the history of transmitted objects and not present with the list of courses generated by the exporter)
 - a catalog should have a NOOP (is present within both the history and payload from the exporter and their respective channel metadata _**match exactly**_)
 
 Because catalogs were being excluded, they were counted as `content to remove`
 
 This PR:
 - adds catalog uuid to the content metadata transmission audit
 - if a catalog is found to not have any updates, the exporter will look up all past transmission audits under that catalog's uuid and copy the item for the current exporter payload, thus resulting in a `No changes needed` from the transmitter, instead of a `delete needed`. 

**Merge checklist:**

**Post merge:**


integrated_channels/integrated_channel/models.py
```diff
@@ -296,6 +297,7 @@ class ContentMetadataItemTransmission(TimeStampedModel):
         blank=True,
         null=True
     )
+    enterprise_customer_catalog_uuid = models.UUIDField()
```
mattdrayer, line 300: You'll need to commit a migration file for this and also it probably should be set as nullable/blank.

mattdrayer: Also not sure if we should consider referential integrity here and specify the UUID as an FK to the ECC table.

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -38,10 +38,10 @@ def transmit(self, payload, **kwargs):
         """
         Transmit content metadata items to the integrated channel.
         """
-        items_to_create, items_to_update, items_to_delete, transmission_map, content_update_times_by_id = \
-            self._partition_items(payload)
+        items_to_create, items_to_update, items_to_delete, transmission_map, content_update_times_by_id, \
```
mattdrayer, line 41: Any particular reason you decided not to go with the `items_to_skip` idea?

alex-sheehan-edx: How would that work? Skipping an item is determined by its presence within the exporter's payload and in the audit table. `Partition items` already decides what items need to be skipped, the trick here is that we need catalog uuid's associated with each particular content item so that when we create new audit items (in `_transmit_create`) we can add the catalog uuid to the record to be used by future skipped catalogs.

integrated_channels/integrated_channel/models.py
alex-sheehan-edx: Yup- migration is in the works now. The trick is that I want to backfill all previous metadata transmission audits with their specific catalog's uuid. This way the field can be not nullable or blank and we won't have to worry about recovering from our present state.

alex-sheehan-edx: I'm not sure if we can FK here since the tables are across db's (ie platform vs ECC)

integrated_channels/integrated_channel/transmitters/content_metadata.py
binodpant: hmm so how / where exactly are we detecting that the content has not changed and there can be skipped? sorry I read and may have missed the logic. also the addition of the extra field is auto handled by the diff() we do, no? just need to ensure that.

binodpant: is it more like `Skipping an item is determined by its presence within the exporter's payload and in the audit table.` AND `the content item is the same` ?

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -105,11 +105,11 @@ def export(self, **kwargs):
                 )
 
             if force_retrieve_all_catalogs or need_catalog_update:
-                content_metadata_item = self.enterprise_catalog_api.get_content_metadata(
+                content_metadata_items = self.enterprise_catalog_api.get_content_metadata(
```
binodpant, line 108: great, good change to naming!

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -226,14 +246,36 @@ def _get_most_recent_catalog_update_time(self):
             return past_transmissions[0]['content_last_changed']
         return None
 
+    def update_content_transmissions_catalog_uuids(self):
+        """
+        Placeholder
```
binodpant, line 251: does this method need to be in the exporter

binodpant: I would consider removing this method once this issue is over, like we discussed, that way we can keep it here for now for easier work

binodpant:
My high level thoughts: I think this approach will work. I do not like the new complexity of the exporter, and the fact that we now ask audit logs in two places, and also we fetch the superset of audit logs in transmitter, but a subset (per catalog( in the exporter, but changing all that out is going to require more refactors and we don't want to do that right now

I think with proper unit testing and local testing, this can go out. So I am good to approve once we have some unit tests

I also recommend removing the task related code once we are all done, since it's in the exporter now and that kind of muddies what should be in an exporter. But it does exactly what is needed for now

Sounds like plan is to:

- keep channels off in prod
- check in these changes
- run management command in jenkins to populate catalog uuids
- enable one customer in prod and test it out
- if all works, enable all configs in prod


if this is not the case, happy to be corrected!
good luck, and great job @alex-sheehan-edx

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -130,12 +130,32 @@ def export(self, **kwargs):
...
+                ContentMetadataItemTransmission = apps.get_model(
+                    'integrated_channel',
+                    'ContentMetadataItemTransmission'
+                )
+                past_catalog_transmission_queryset = ContentMetadataItemTransmission.objects.filter(
+                    enterprise_customer=self.enterprise_configuration.enterprise_customer,
+                    integrated_channel_code=self.enterprise_configuration.channel_code(),
+                    enterprise_customer_catalog_uuid=enterprise_customer_catalog.uuid
+                )
+                for past_transmission in past_catalog_transmission_queryset:
```
mattdrayer, line 153: Might be useful to capture a log entry here -- eg, "Found X past transmission items."

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -105,11 +105,11 @@ def export(self, **kwargs):
                 )
 
             if force_retrieve_all_catalogs or need_catalog_update:
-                content_metadata_item = self.enterprise_catalog_api.get_content_metadata(
+                content_metadata_items = self.enterprise_catalog_api.get_content_metadata(
                     self.enterprise_customer,
                     [enterprise_customer_catalog]
                 )
-                for item in content_metadata_item:
+                for item in content_metadata_items:
```
mattdrayer, line 112: Might be good to capture a log entry here, eg "Found X content metadata items"

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -226,14 +246,36 @@ def _get_most_recent_catalog_update_time(self):
...
+                self.enterprise_customer,
+                [enterprise_customer_catalog]
+            )
+            content_ids = [get_content_metadata_item_id(item) for item in content_metadata_items]
+            ContentMetadataItemTransmission = apps.get_model(
+                'integrated_channel',
+                'ContentMetadataItemTransmission'
+            )
+            transmission_items = ContentMetadataItemTransmission.objects.filter(enterprise_customer=self.enterprise_configuration.enterprise_customer, integrated_channel_code=self.enterprise_configuration.channel_code(), content_id__in=content_ids)
+            for item in transmission_items:
```
mattdrayer, line 267: Log "Found X transmission items" ?

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -226,14 +246,36 @@ def _get_most_recent_catalog_update_time(self):
...
     Object representation of a content metadata item export.
     """
 
-    def __init__(self, content_metadata_item, channel_content_metadata_item, content_last_changed=None):
+    def __init__(self, content_metadata_item, channel_content_metadata_item, enterprise_customer_catalog_uuid=None, content_last_changed=None):
         self.content_id = get_content_metadata_item_id(content_metadata_item)
         self.metadata = content_metadata_item
         self.channel_metadata = channel_content_metadata_item
         self.content_last_changed = content_last_changed
+        self.enterprise_customer_catalog_uuid = enterprise_customer_catalog_uuid
```
mattdrayer, line 281: Is it important to undestand when/why `enterprise_customer_catalog_uuid` would be None?  Might want to log instances of that particular scenario.

integrated_channels/integrated_channel/management/commands/update_content_transmission_catalogs.py
```diff
@@ -0,0 +1,50 @@
...
+        # Before we do a whole bunch of database queries, make sure that the user we were passed exists.
+        try:
+            User.objects.get(username=username)
+        except User.DoesNotExist as no_user_error:
+            raise CommandError('A user with the username {} was not found.'.format(username)) from no_user_error
+
+        channels = self.get_integrated_channels(options)
+
+        for channel in channels:
+            channel_code = channel.channel_code()
```
mattdrayer, line 48: Capture a log entry here?

integrated_channels/cornerstone/transmitters/content_metadata.py
```diff
@@ -31,14 +31,14 @@ def transmit(self, payload, **kwargs):
             payload (OrderedDict): course metadata dictionary corresponding to the result of the content metadata
             exporter's export.
         """
-        items_to_create, items_to_update, items_to_delete, transmission_map, items_catalog_updated_times = \
-            self._partition_items(payload)
+        items_to_create, items_to_update, items_to_delete, transmission_map, items_catalog_updated_times, \
```
binodpant, line 34: we had circled the idea of embedding the catalog uuid in the items_to_create itself but it seemed safer to leave those untouched for the time being

integrated_channels/integrated_channel/exporters/content_metadata.py
binodpant: should None raise a failure / exception?

alex-sheehan-edx: yeah, I agree- we should _**always**_ have a catalog uuid here. I'll make it a required param

integrated_channels/integrated_channel/management/commands/update_content_transmission_catalogs.py
alex-sheehan-edx: We have entry logs at the task level which is invoked by this- they are specifically formatted for starts and endings of a task/command. That feels sufficient?

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -75,10 +77,18 @@ def _partition_items(self, channel_metadata_item_map):
...
                     items_to_update[content_id] = channel_metadata
                     items_catalog_updated_times[content_id] = item.content_last_changed
+                else:
+                    skipped_items.append(content_id)
             else:
                 items_to_create[content_id] = channel_metadata
+                content_catalog_map[content_id] = item.enterprise_customer_catalog_uuid
                 items_catalog_updated_times[content_id] = item.content_last_changed
 
+        if skipped_items:
```
binodpant, line 87: nit: sorry for last minute feedback. All the other items are actual objects/dicts. This is actually a array of scalars (content+ids). Shall we just name it like that `skipped_content_ids`?

binodpant (approved):

PR #1370, opened 2021-09-30
bug: fix cornerstone course key error with table

kiram15:
**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -3008,3 +3008,28 @@ def __repr__(self):
         Return uniquely identifying string representation.
         """
         return self.__str__()
+
+class CornerstoneCourseKeys(models.Model):
```
mattdrayer, line 3012: Suggest singular `CornerstoneCourseKey` for this model -- also, not sure if it should live in `integrated_channels/cornerstone`

integrated_channels/utils.py
```diff
@@ -36,7 +37,11 @@ def convert_invalid_course_ids(course_id):
     re2 = re.compile(r"[|<>.&%\s\\/\“]+")
     if re2.search(course_id):
         # If the course key contains any of the invalid chars, encode the key
-        return encode_course_key_into_base64(course_id)
+        course_id = encode_course_key_into_base64(course_id)
+    # If the encoded or unencoded version of the key are over 50 characters, they will error out
+    # in cornerstone, so we convert them to a uuid.
+    if len(course_id) > 50:
```
mattdrayer, line 43: Should this just be a method on the CornerstoneCourseKey class?

enterprise/models.py
kiram15: Was having some problems with circular imports when it was in there with the implementation I have at the minute, which is why I just put it in models. If you think it's important I will try to work it out!

integrated_channels/utils.py
kiram15: For this, we would have to make different instances in the code for the other usages it has, so while the general utils file may not be the best, I don't know if its the best to only have it available as part of the model

enterprise/models.py
mattdrayer: It would be good to understand and resolve the circular imports -- we really should be defining Cornerstone-specific models in the Cornerstone integrated channel versus at the top-level of enterprise.

mattdrayer:
This LGTM now although you probably want an actual engineering team member to approve the PR.

docs/decisions/0008-course-key-table-for-cornerstone.rst
```diff
@@ -0,0 +1,40 @@
...
+=======
+
+At one point we were having transmitted courses to CSOD tenants fail with invalid chars contained within course metadata. 
+To combat this, we decided to encode the course keys. This has resulted in some of the courses failing validation with a 
+"course key too long" failures. There's a 50 char limit to course keys on the CSOD lms, so we need to limit the size of
+our encoded course keys within the content metadata.
+
+Simply truncating it would not allow for backwards compatibility or ensuring uniqueness. We also considered encoding
+and then compress, but while this would most likely work for every course key (the largest length found in prod was 54 chars),
+it can't properly guarantee the key to be under 50 chars, and we would have to worry with hashing the fact that
```
binodpant, line 19: I would also mention the fact that we started an exploration of whether there are existing uuids we can use to map to course keys. We found that there are uuids in Discovery for course :
https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L768
and 
courseruns: https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L1277
 but we are still in the stages of finding out whether they are stable throughout their lifespan to be used. Turns out only courses with 'published' state in discovery have stable uuids. Because there are some unknowns there, this fix will allow us a more immediate remediation.

integrated_channels/cornerstone/utils.py
```diff
@@ -3,9 +3,28 @@
...
+
+
+def cornerstone_learner_data_transmission_audit():
+    """
+    Returns the ``EnterpriseCustomer`` class.
+    """
+    return apps.get_model('cornerstone', 'CornerstoneLearnerDataTransmissionAudit')
+
+
+def cornerstone_course_key():
```
binodpant, line 22: I would suggest using a `_model` suffix to the function name for better usage clarity

integrated_channels/cornerstone/utils.py
```diff
@@ -31,3 +50,26 @@ def create_cornerstone_learner_data(request, course_id):
         pass
     except Exception as ex:  # pylint: disable=broad-except
         LOGGER.error('Unable to Create/Update CornerstoneLearnerDataTransmissionAudit. {ex}'.format(ex=ex))
+
+
+def convert_invalid_course_ids(course_id):
```
binodpant, line 55: name should be `convert_invalid_course_id` ?

integrated_channels/cornerstone/utils.py
```diff
@@ -31,3 +50,26 @@ def create_cornerstone_learner_data(request, course_id):
...
+    """
+    Regex check a course ID to see if it contains any invalid chars. If it does then encode the string, otherwise
+    return the original course ID.
+    """
+    re2 = re.compile(r"[|<>.&%\s\\/\“]+")
+    if re2.search(course_id):
+        # If the course key contains any of the invalid chars, encode the key
+        course_id = encode_course_key_into_base64(course_id)
```
binodpant, line 63: I think it is better to avoid updating the input argument, and instead return a new variable such as `safe_course_id`

integrated_channels/cornerstone/utils.py
```diff
@@ -31,3 +50,26 @@ def create_cornerstone_learner_data(request, course_id):
...
+    # If the encoded or unencoded version of the key are over 50 characters, they will error out
+    # in cornerstone, so we convert them to a uuid.
+    if len(course_id) > 50:
+        course_id = uuid4()
+    return course_id
+
+
+def add_cornerstone_key_pair(course_id):
+    key_mapping, ___ = cornerstone_course_key().objects.get_or_create(
+        internal_course_id=course_id, defaults={
```
binodpant, line 73: formatting?

integrated_channels/cornerstone/utils.py
```diff
@@ -31,3 +50,26 @@ def create_cornerstone_learner_data(request, course_id):
...
+        # If the course key contains any of the invalid chars, encode the key
+        course_id = encode_course_key_into_base64(course_id)
+    # If the encoded or unencoded version of the key are over 50 characters, they will error out
+    # in cornerstone, so we convert them to a uuid.
+    if len(course_id) > 50:
+        course_id = uuid4()
+    return course_id
+
+
+def add_cornerstone_key_pair(course_id):
```
binodpant, line 71: perhaps this method is better called `get_or_create_key_pair` ?

tests/test_integrated_channels/test_cornerstone/test_exporters/test_content_metadata.py
```diff
@@ -158,14 +159,29 @@ def test_transform_description(self, item_description, expected_description):
...
         assert exporter.transform_course_key(item_content_metadata) == expected_id
 
+    @ddt.data(
+        (
+            {'key': 'testing+out+very+very+very+very+long+course+id+T2020'},
+        )
+    )
+    @responses.activate
+    @ddt.unpack
+    def test_transform_key_2(self, item_key):
```
binodpant, line 177: Think we do need to use ddt here since it's only one arg. Unless you plan to add more cases... it also makes test names harder to read in logs.

integrated_channels/cornerstone/utils.py
kiram15: It's weird but it's the way lint wanted 🙄

PR #1375, opened 2021-10-06
fix: use synchronized instance of channel_config when requesting tokens ENT-4988

binodpant:
if we don't do this, multiple sessions calling the oauth function can invalidate each other's Blackboard tokens

Basically, by using select_for_update and using it in an transaction.atomic() block we ensure that we don't let other sessions using this function, read off a state of refresh_token that is no longer valid (or has been used up). The sessions that come next will await / blocked on this oauth token function call to execute and save refresh_token before proceeding, thus resolving the state issue with refresh tokens. 

__What's the rationale behind having the transactional atomic wrapper only around `_get_oauth_access_token`?__

Will previously issues access)_tokens (each django session is going to store its own access token and expires_at settings) continue to work?

The idea is the previously requested/obtained access tokens still work until they expire. This allows us to store just access_token and expires_at in the django sessoin. However, the `refresh_token` is a much more singleton kind of resource, only one is ever valid wrt a given Blackboard instance. Hence we cannot store that in django sessoin and need to store in db and propertly synchornize.
We verified this by requesting using prior access tokens after having obtained new refresh/access tokens, in postman, and that worked.

__How did I test this?__

First I wrote up a test script in the form of an overly aggressive management command that enforces getting a new token every time the `_create_session` is called (this is not true generally, we await expiry of the token)

```
 def handle(self, *args, **options):
            ent_config = BlackboardEnterpriseCustomerConfiguration.objects.get(pk=1)
            bbclient = client.BlackboardAPIClient(ent_config)
            for i in range(1, 10):
                bbclient.expires_at = datetime.datetime(2021, 8, 10)
                bbclient._create_session()  # pylint: disable=protected-access
                LOGGER.info(f'new expires_at = {bbclient.expires_at}')
                LOGGER.info(f"Bb course id: {bbclient._resolve_blackboard_course_id('559480219276064835')}")  # pylint: disable=protected-access
```

This reproduces the issue when the session that does not use the refresh token that was most recently received, ran into the refresh_token errors mentoined in the ticket, as expected

After adding the changes in this PR I no longer see those errors and I am able to receive and save the refresh token each time consistently, but this time the session that comes later awaits the currently running loop in the other session, to be done, so you can also visibly watch the loop cycles await each other per call

ENT-4988

**Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * Nothing
 
+[3.30.4]
+--------
+* fix: Blackboard integrated channel now correctly synchronizes the one-and-only valid refresh_token
+
 [3.30.3]
 --------
 * fix: content_filter in django admin was broken after the jsonfield upgrade, so this contains fix for that
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.30.3"
+__version__ = "3.30.4"
 
 default_app_config = "enterprise.apps.EnterpriseConfig"


johnnagro (approved):
great work 👏

alex-sheehan-edx (approved):
I'm thinking a test of any kind would be superb but at the same time I can't think of how to test this 🤔. Regardless, love the solution and ✅ :shipit:

PR #1378, opened 2021-10-07
feat: Ability to show/hide labor market data in Learner Portal

sameenfatima78:
**Merge checklist:**

**Post merge:**
- [x] Tag pushed and a new [version](https://github.com/edx/edx-enterprise/releases) released
- [x] After versioned build finishes in [GitHub Actions](https://github.com/edx/edx-enterprise/actions), verify version has been pushed to [PyPI](https://pypi.org/project/edx-enterprise/)
- [x] PR created in [edx-platform](https://github.com/edx/edx-platform) to upgrade dependencies (including edx-enterprise)


enterprise/models.py
```diff
@@ -293,6 +293,11 @@ class Meta:
         help_text=_("Specifies whether the enterprise learner portal site should be made known to the learner.")
     )
 
+    hide_labor_market_data = models.BooleanField(
```
HammadAhmadWaqas, line 296: I think It would be better if we keep the flag as positive instead of negative, (instead of `hide` start it with `show`)
Here are two reasons,
- It will be consistent with other existing flags, you can say they start with enable instead of disable.
- Negative flags can be confusing sometimes, 

Few suggested names are:
- enable_learner_portal_labor_market_data
- enable_labor_market_data
- show_learner_portal_labor_market_data
- show_labor_market_data

sameenfatima78: I think the current approach is more straightforward, also it wouldn't go with business requirements as suggested in the ticket. We are already fetching a similar flag from enterprise configuration on learner portal which follows the same standard, as you can see below: 
Screenshot 2021-10-08 at 2 32 47 PM

PR #1387, opened 2021-10-18
fix: properly weight blackboard grades

johnnagro:
Similar to Canvas, Blackboard should not include most grade columns in the final grade. Fix up pre-existing settings along the way.

[ENT-4394](https://openedx.atlassian.net/browse/ENT-4394)

### DIFF
--- a/integrated_channels/blackboard/client.py
+++ b/integrated_channels/blackboard/client.py
@@ -26,6 +26,7 @@ GRADEBOOK_PATH = '/learn/api/public/v1/courses/{course_id}/gradebook/columns'
 ENROLLMENT_PATH = '/learn/api/public/v1/courses/{course_id}/users'
 COURSE_PATH = '/learn/api/public/v1/courses'
 POST_GRADE_COLUMN_PATH = '/learn/api/public/v2/courses/{course_id}/gradebook/columns'
+PATCH_GRADE_COLUMN_PATH = '/learn/api/public/v2/courses/{course_id}/gradebook/columns/{column_id}'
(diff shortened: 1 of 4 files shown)


tests/test_integrated_channels/test_blackboard/test_client.py
```diff
@@ -763,6 +763,50 @@ def test_content_child_customization_failure(self):
...
+                'name': self.blackboard_grade_column_name,
+                'description': "edX learner's grade.",
+                'created': '2021-07-16T19:46:29.698Z',
+                'score': {'possible': 100.0},
+                'availability': {'available': 'Yes'},
+                'includeInCalculations': True,
+            }],
+        }
+
+        with responses.RequestsMock() as rsps:
```
alex-sheehan-edx, line 787: 👏 👏 👏

integrated_channels/blackboard/client.py
```diff
@@ -630,7 +647,8 @@ def _get_bb_user_id_from_enrollments(self, user_email, course_id):
             HTTPStatus.NOT_FOUND.value
         )
 
-    def _get_or_create_integrated_grade_column(self, bb_course_id, grade_column_name, external_id, points_possible=100):
+    def _get_or_create_integrated_grade_column(self, bb_course_id, grade_column_name, external_id, points_possible=100,
```
alex-sheehan-edx, line 650: Anyone feel free to correct me if I'm wrong but I think our formatting convention for multiline params is
```
def method(
   self,
   arg1,
   arg2,
   arg3
)
```

johnnagro: yeah, i'm far from an expert on this. I was working off this stackoverflow https://stackoverflow.com/a/24988191

PR #1384, opened 2021-10-13
feat: Integrated channels content metadata transmissions V2

alex-sheehan-edx:
details written below, changes to come once enterprise catalog PR goes live.

All questions/concerns welcome

Old implementation:
old IC

new proposed implementation:
new IC

**Merge checklist:**
- [ ] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files)
    - `base.in` if needed in production but edx-platform doesn't install it
    - `test-master.in` if edx-platform pins it, with a matching version
    - `make upgrade && make requirements` have been run to regenerate requirements
- [ ] `make static` has been run to update webpack bundling if any static content was updated
- [ ] `./manage.py makemigrations` has been run
    - Checkout the [Database Migration](https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations) Confluence page for helpful tips on creating migrations.
    - *Note*: This **must** be run if you modified any models.
      - It may or may not make a migration depending on exactly what you modified, but it should still be run.
    - This should be run from either a venv with all the lms/edx-enterprise requirements installed or if you checked out edx-enterprise into the src directory used by lms, you can run this command through an lms shell.
        - It would be `./manage.py lms makemigrations` in the shell.
- [ ] [Version](https://github.com/edx/edx-enterprise/blob/master/enterprise/__init__.py) bumped
- [ ] [Changelog](https://github.com/edx/edx-enterprise/blob/master/CHANGELOG.rst) record added
- [ ] Translations updated (see docs/internationalization.rst but also this isn't blocking for merge atm)

**Post merge:**
- [ ] Tag pushed and a new [version](https://github.com/edx/edx-enterprise/releases) released
    - *Note*: Assets will be added automatically. You just need to provide a tag (should match your version number) and title and description.
- [ ] After versioned build finishes in [GitHub Actions](https://github.com/edx/edx-enterprise/actions), verify version has been pushed to [PyPI](https://pypi.org/project/edx-enterprise/)
    - Each step in the release build has a condition flag that checks if the rest of the steps are done and if so will deploy to PyPi.
    (so basically once your build finishes, after maybe a minute you should see the new version in PyPi automatically (on refresh))
- [ ] PR created in [edx-platform](https://github.com/edx/edx-platform) to upgrade dependencies (including edx-enterprise)
    - This **must** be done after the version is visible in PyPi as `make upgrade` in edx-platform will look for the latest version in PyPi.
[...]

PR #1394, opened 2021-10-27
feat: exporter now only requests metadata for courses that need updating

alex-sheehan-edx:
fast follow to https://github.com/edx/edx-enterprise/pull/1384

To be merged after https://github.com/edx/enterprise-catalog/pull/341 goes live

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api_client/enterprise_catalog.py
+++ b/enterprise/api_client/enterprise_catalog.py
@@ -200,14 +200,35 @@ class EnterpriseCatalogApiClient(JwtLmsApiClient):
             )
             return {}
 
+    def traverse_get_content_metadata(self, endpoint, query, catalog_uuid):
+        """
+        Helper method to traverse over a paginated response from the enterprise-catalog service's `get_content_metadata`
+        endpoint.
+        """
+        content_metadata = OrderedDict()
+        try:
+            response = endpoint.get(**query)
+            for item in utils.traverse_pagination(response, endpoint):
+                content_id = utils.get_content_metadata_item_id(item)
+                content_metadata[content_id] = item
+        except (SlumberBaseException, ConnectionError, Timeout) as exc:
+            LOGGER.exception(
+                'Failed to get content metadata for Catalog [%s] in enterprise-catalog due to: [%s]',
+                catalog_uuid, str(exc)
+            )
+            raise
+
+        return content_metadata
+
     @JwtLmsApiClient.refresh_token
-    def get_content_metadata(self, enterprise_customer, enterprise_catalogs=None):
+    def get_content_metadata(self, enterprise_customer, enterprise_catalogs=None, content_keys_filter=None):
         """
         Return all content metadata contained in the catalogs associated with the EnterpriseCustomer.
 
         Arguments:
             enterprise_customer (EnterpriseCustomer): The EnterpriseCustomer to return content metadata for.
             enterprise_catalogs (EnterpriseCustomerCatalog): Optional list of EnterpriseCustomerCatalog objects.
+            content_keys_filter (List): List of content keys to filter by in the content metadata endpoint
 
         Returns:
             list: List of dicts containing content metadata.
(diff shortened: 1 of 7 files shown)


johnnagro (approved):
⛵

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -213,11 +213,13 @@ def export(self, **kwargs):
 
             # We only need to fetch content metadata if there are items to update or create
             if items_to_create or items_to_update:
-                # Todo: create a content item specific fetch content metadata endpoint in enterprise catalog so all data
-                # isn't required to fetch
+                items_create_keys = [key.get('content_key') for key in items_to_create]
+                items_update_key = list(items_to_update.keys())
```
binodpant, line 217: I think a plural name makes sense here?

binodpant (approved):
just a name-nit but otherwise lgtm

enterprise/api_client/enterprise_catalog.py
```diff
@@ -200,14 +200,35 @@ def delete_enterprise_catalog(self, catalog_uuid):
...
+        endpoint.
+        """
+        content_metadata = OrderedDict()
+        try:
+            response = endpoint.get(**query)
+            for item in utils.traverse_pagination(response, endpoint):
+                content_id = utils.get_content_metadata_item_id(item)
+                content_metadata[content_id] = item
+        except (SlumberBaseException, ConnectionError, Timeout) as exc:
+            LOGGER.exception(
```
binodpant, line 215: incrementally improvable: we can use the existing pattern to log, that uses a nicely extracted out lambda like thingie I recently used in one place, hope to clean that out someday

enterprise/api_client/enterprise_catalog.py
```diff
@@ -200,14 +200,35 @@ def delete_enterprise_catalog(self, catalog_uuid):
...
 
+    def traverse_get_content_metadata(self, endpoint, query, catalog_uuid):
+        """
+        Helper method to traverse over a paginated response from the enterprise-catalog service's `get_content_metadata`
+        endpoint.
+        """
+        content_metadata = OrderedDict()
+        try:
+            response = endpoint.get(**query)
+            for item in utils.traverse_pagination(response, endpoint):
```
binodpant, line 211: oh wait where do you even use the catalog_uuid?

alex-sheehan-edx: `endpoint = getattr(self.client, self.GET_CONTENT_METADATA_ENDPOINT.format(catalog_uuid))` catalog uuid is part of the URL, which is included in the endpoint.

enterprise/api_client/enterprise_catalog.py
```diff
@@ -200,14 +200,35 @@ def delete_enterprise_catalog(self, catalog_uuid):
...
+                content_id = utils.get_content_metadata_item_id(item)
+                content_metadata[content_id] = item
+        except (SlumberBaseException, ConnectionError, Timeout) as exc:
+            LOGGER.exception(
+                'Failed to get content metadata for Catalog [%s] in enterprise-catalog due to: [%s]',
+                catalog_uuid, str(exc)
```
adamstankiewicz, line 217: nit: I believe `LOGGER.exception` already logs the exception without needing to explicitly include it in the message?

enterprise/utils.py
```diff
@@ -900,6 +900,20 @@ def get_cache_key(**kwargs):
     return get_django_cache_key(**kwargs)
 
 
+def chunk_content_keys(content_filter, chunk_size):
```
adamstankiewicz, line 903: Could we make use of the [existing `batch` util method](https://github.com/edx/edx-enterprise/blob/a149e9e55cdae56179ee470ba72ebeab3d34aa49/enterprise/utils.py#L2050) rather than redefining a new one?

enterprise/api_client/enterprise_catalog.py
alex-sheehan-edx:
```
There are three keyword arguments in kwargs which are inspected: exc_info which, if it does not evaluate as false, causes exception information to be added to the logging message.
```

binodpant: yup I believe exc_info is the desired method?

enterprise/api_client/enterprise_catalog.py
binodpant: we can ignore this/resolve, since there's no duplicatoin of args issue here

PR #1397, opened 2021-11-02
fix: allowing records to be cross catalogs

alex-sheehan-edx:
[Jira](https://openedx.atlassian.net/browse/ENT-5097)

A create request may be made for a piece of content that already exists for a customer under a different catalog. Therefore we need to check if a content record exists under another catalog before we attempt to create one. 

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/transmitters/content_metadata.py
+++ b/integrated_channels/integrated_channel/transmitters/content_metadata.py
@@ -156,15 +156,19 @@ class ContentMetadataTransmitter(Transmitter):
 
     def _create_transmissions(self, content_metadata_item_map, content_updated_mapping):
         """
-        Create ContentMetadataItemTransmission models for the given content metadata items.
+        Create ContentMetadataItemTransmission models for the given content metadata items. Because records are soft
+        deleted, before creating new records we must verify that there is no previously deleted record under this
+        customer with this content ID. Additionally we must also check if a record exists under this customer under a
+        separate catalog as content over catalogs is overwritten.
         """
         ContentMetadataItemTransmission = apps.get_model(
             'integrated_channel',
             'ContentMetadataItemTransmission'
         )
         create_transmissions = []
-        reactivate_transmissions = []
         for content_id, channel_metadata in content_metadata_item_map.items():
+            catalog_uuid = content_updated_mapping.get(content_id).get('catalog_uuid')
+            content_last_changed = content_updated_mapping.get(content_id).get('modified')
             LOGGER.info(
                 f'Creating content transmission record for course: {content_id} under enterprise customer: '
                 f'{self.enterprise_configuration.enterprise_customer.uuid}.'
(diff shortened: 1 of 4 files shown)


integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -163,46 +163,64 @@ def _create_transmissions(self, content_metadata_item_map, content_updated_mappi
...
-        ContentMetadataItemTransmission.objects.bulk_update(
-            reactivate_transmissions,
-            ['deleted_at', 'channel_metadata', 'content_last_changed']
-        )
+                    past_transmission.channel_metadata = channel_metadata
+                    past_transmission.content_last_changed = content_updated_mapping.get(content_id).get('modified')
+                    past_transmission.enterprise_customer_catalog_uuid = content_updated_mapping.get(
+                        content_id
+                    ).get('catalog_uuid')
+                    past_transmission.save()
```
binodpant, line 211: ah so no more bulk_update, just a per item update here replaces that right?

tests/test_integrated_channels/test_integrated_channel/test_transmitters/test_content_metadata.py
```diff
@@ -287,3 +289,41 @@ def test_transmit_delete_failure(self):
...
+        create_payload = {content_id: channel_metadata}
+        update_payload = {}
+        delete_payload = {}
+        content_updated_mapping = {
+            content_id: {'catalog_uuid': new_catalog_uuid, 'modified': datetime.now()},
+        }
+        self.create_content_metadata_mock.return_value = (200, '{"success":"true"}')
+        transmitter = ContentMetadataTransmitter(self.enterprise_config)
+        transmitter.transmit(create_payload, update_payload, delete_payload, content_updated_mapping)
+        import pdb;
```
binodpant, line 322: remove debug calls

binodpant (changes requested):

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -174,36 +173,39 @@ def _create_transmissions(self, content_metadata_item_map, content_updated_mappi
                 integrated_channel_code=self.enterprise_configuration.channel_code(),
                 content_id=content_id,
                 deleted_at__isnull=False,
-            )
+            ).first()
```
binodpant, line 176: I would also find it useful to have the docstring for this method quickly explain what it does and the 3 cases it handles and what it does in each case, just briefly is fine

binodpant: mainly think it's useful now since the method is no longer as simple as it used to be, but more accurate!

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -174,36 +173,39 @@ def _create_transmissions(self, content_metadata_item_map, content_updated_mappi
...
 
-            if past_deleted_transmission.first():
+            if past_deleted_transmission:
                 LOGGER.info(
                     f'Found previously deleted content record while creating record for course: {content_id}'
```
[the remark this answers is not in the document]
alex-sheehan-edx: transmissions don't have UUID's 😞 + the pairing of customer and content ID is unique so is in a way the identifying pair. I could log the ID but I just feel like it's not necessary.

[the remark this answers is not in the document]
alex-sheehan-edx: but I will indicate the intentions of the process for what it will do to the record (ie reactivate)

binodpant (approved):
wohoo!

PR #1401, opened 2021-11-03
feat: allow filtering enterprise learners by enterprise uuid or enterprise role

long74100:
- Update EnterpriseCustomerUserFilterBackend to allow filtering by enterprise customer uuid and enterprise user role
- Add created_at to serializer to get activation date

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/filters.py
+++ b/enterprise/api/filters.py
@@ -4,8 +4,11 @@ Filters for enterprise API.
 """
 
 from rest_framework import filters
+from rest_framework.exceptions import ValidationError
 
 from django.contrib import auth
+
+from enterprise.models import SystemWideEnterpriseUserRoleAssignment
 
 User = auth.get_user_model()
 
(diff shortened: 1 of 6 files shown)


macdiesel (approved):

enterprise/api/filters.py
```diff
@@ -33,25 +35,64 @@ def filter_queryset(self, request, queryset, view):
...
+            query_params.update(enterprise_customer_id=enterprise_customer_uuid)
+
+        if role:
+            role_assignment_filters = {'role__name': role}
+            if enterprise_customer_uuid:
+                role_assignment_filters.update(enterprise_customer_id=enterprise_customer_uuid)
+
+            users_with_role = SystemWideEnterpriseUserRoleAssignment.objects.filter(
+                **role_assignment_filters
+            ).values_list('user_id', flat=True)
```
iloveagent57, line 80: This could be a really big result set if `enterprise_customer_uuid` is excluded from the filter.  Maybe we should force `enterprise_customer_uuid` to not be null if `role` is provided?

iloveagent57 (commented):
Looks good, just the one concern about filtering on role name alone.

enterprise/api/filters.py
long74100: good call, I guess it makes sense to throw an error in this case?

enterprise/api/filters.py
```diff
@@ -33,25 +36,68 @@ def filter_queryset(self, request, queryset, view):
...
+            query_params.update(enterprise_customer_id=enterprise_customer_uuid)
+
+        if role:
+            role_assignment_filters = {'role__name': role}
+
+            if not enterprise_customer_uuid:
+                raise ValidationError('Cannot filter by role without providing enterprise_customer_uuid.')
+
+            if enterprise_customer_uuid:
+                role_assignment_filters.update(enterprise_customer_id=enterprise_customer_uuid)
```
iloveagent57, line 81: Nit: `enterprise_customer_uuid` is guaranteed to be truthy at this point.

iloveagent57 (approved):

enterprise/api/filters.py
long74100: :facepalm forgot to remove

PR #1403, opened 2021-11-04
fix: CSOD workaround bugfix

johnnagro:
Fix to a name-spacing error introduced by our CSOD workaround. This call uses the [AppConfig.label](https://docs.djangoproject.com/en/3.2/ref/applications/#django.apps.AppConfig.label) which is the last part of the path. This model is defined under `integrated_channels.cornerstone` rather than `integrated_channel`.

This bug is in production and may be affecting enrollments for CSOD learners. I tested this bug and my fix with a local console. I will try to introduce tests to cover these lines of code.

- [ENT-5115](https://openedx.atlassian.net/browse/ENT-5115)

Log lines to look for:

```
LookupError: App 'integrated_channel' doesn't have a 'CornerstoneEnterpriseCustomerConfiguration' model.
```

```
KeyError: 'cornerstoneenterprisecustomerconfiguration'
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * Nothing
 
+[3.33.6]
+---------
+fix: CSOD API session tokens bugfix
+
 [3.33.5]
 ---------
 fix: CSOD API session tokens are now saved to the customer's configuration instead of individual transmission audits
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.33.5"
+__version__ = "3.33.6"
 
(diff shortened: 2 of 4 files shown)


mattdrayer (approved):
At some point we should clean up the redundant `if` statement on line 2343 but that does not have to be part of this PR.  I also would like to see a stronger qualification of whether or not this is a CSOD request since we are assuming that the presence of a `sessionToken` query string parameter means CSOD right now.  But again, that does not need to be part of this PR.

johnnagro:
> At some point we should clean up the redundant `if` statement on line 2343 but that does not have to be part of this PR. I also would like to see a stronger qualification of whether or not this is a CSOD request since we are assuming that the presence of a `sessionToken` query string parameter means CSOD right now. But again, that does not need to be part of this PR.

thanks for the reminder - i just added the change.

PR #1406, opened 2021-11-16
feat: for SAP channel, do not consider grade when diffing audit records ENT-4752

binodpant:
this allows us to avoid retransmitting completion reports for SAP, when grade changes

What this PR does:

- [x] introduces a new `ChannelSettingsMixin` which will house class level properties that represent settings. I added one setting there called `INCLUDE_GRADE_FOR_COMPLETION_AUDIT_CHECK` which is True by default. This keeps today's behavior when sending completions (meaning it also considers grade changes as a reason to retransmit completions)
- [x] Makes the learner_data Exporter and Transmitter classes inherit this mixin as in `class LearnerExporter(ChannelSettingsMixin, Exporter):`
- [x] Disables this flag for SAP Exporter/Transmitter for learner_data which uses this mixin.
- [x] Uses this flag when passing a flag to the util method `is_already_transmitted` which will in turn ignore grade diff when answering the question, if this flag is False

ENT-4752

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/transmitters/learner_data.py
+++ b/integrated_channels/integrated_channel/transmitters/learner_data.py
@@ -9,6 +9,7 @@ from http import HTTPStatus
 from django.apps import apps
 
 from integrated_channels.exceptions import ClientError
+from integrated_channels.integrated_channel.channel_settings import ChannelSettingsMixin
 from integrated_channels.integrated_channel.client import IntegratedChannelApiClient
 from integrated_channels.integrated_channel.exporters.learner_data import LearnerExporterUtility
 from integrated_channels.integrated_channel.transmitters import Transmitter
@@ -17,7 +18,7 @@ from integrated_channels.utils import generate_formatted_log, is_already_transmi
 LOGGER = logging.getLogger(__name__)
 
 
-class LearnerTransmitter(Transmitter):
+class LearnerTransmitter(ChannelSettingsMixin, Transmitter):
     """
     A generic learner data transmitter.
 
@@ -92,7 +93,7 @@ class LearnerTransmitter(Transmitter):
             except ClientError as client_error:
                 code = client_error.status_code
                 body = client_error.message
-                self.handle_transmission_error(
+                self.log_transmission_error(
                     learner_data,
                     client_error,
                     app_label,
@@ -159,11 +160,13 @@ class LearnerTransmitter(Transmitter):
             )
 
             # Check the last transmission for the current enrollment and see if the old grades match the new ones
+            # note: the property self.include_grade_for_completion_audit_check comes from ChannelSettingsMixin
             if is_already_transmitted(
                 TransmissionAudit,
                 enterprise_enrollment_id,
                 learner_data.grade,
-                learner_data.subsection_id
+                learner_data.subsection_id,
+                detect_grade_updated=self.INCLUDE_GRADE_FOR_COMPLETION_AUDIT_CHECK,
             ):
                 # We've already sent a completion status for this enrollment
                 LOGGER.info(generate_formatted_log(
@@ -191,7 +194,7 @@ class LearnerTransmitter(Transmitter):
             except ClientError as client_error:
                 code = client_error.status_code
                 body = client_error.message
-                self.handle_transmission_error(
+                self.log_transmission_error(
                     learner_data,
                     client_error,
                     app_label,
@@ -275,7 +278,12 @@ class LearnerTransmitter(Transmitter):
                 continue
 
             grade = getattr(learner_data, 'grade', None)
-            if is_already_transmitted(TransmissionAudit, enterprise_enrollment_id, grade):
+            if is_already_transmitted(
+                TransmissionAudit,
+                enterprise_enrollment_id,
+                grade,
+                detect_grade_updated=self.INCLUDE_GRADE_FOR_COMPLETION_AUDIT_CHECK,
+            ):
                 # We've already sent a completion status for this enrollment
                 LOGGER.info(generate_formatted_log(
                     app_label,
@@ -304,7 +312,7 @@ class LearnerTransmitter(Transmitter):
             except ClientError as client_error:
                 code = client_error.status_code
                 body = client_error.message
-                self.handle_transmission_error(
+                self.log_transmission_error(
                     learner_data,
                     client_error,
                     app_label,
@@ -373,8 +381,8 @@ class LearnerTransmitter(Transmitter):
                 payload=learner_data
             )), exc_info=True)
 
-    def handle_transmission_error(self, learner_data, client_exception,
-                                  integrated_channel_name, enterprise_customer_uuid, learner_id, course_id):
+    def log_transmission_error(self, learner_data, client_exception,
+                               integrated_channel_name, enterprise_customer_uuid, learner_id, course_id):
         """Handle the case where the transmission fails."""
         LOGGER.exception(generate_formatted_log(
             integrated_channel_name, enterprise_customer_uuid, learner_id, course_id,
--- a/integrated_channels/integrated_channel/transmitters/learner_data.py
+++ b/integrated_channels/integrated_channel/transmitters/learner_data.py
@@ -93,7 +93,7 @@ class LearnerTransmitter(Transmitter, ChannelSettingsMixin):
             except ClientError as client_error:
                 code = client_error.status_code
                 body = client_error.message
-                self.handle_transmission_error(
+                self.process_transmission_error(
                     learner_data,
                     client_error,
                     app_label,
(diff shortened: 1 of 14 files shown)


tests/test_integrated_channels/test_sap_success_factors/test_exporters/test_learner_data.py
```diff
@@ -57,3 +57,18 @@ def test_call_get_remote_id(self, mock_get_course_run_for_enrollment, mock_get_c
...
+        """
+        If you override any settings to the ChannelSettingsMixin, add a test here for those
+        """
+        user = UserFactory()
+        enterprise_customer = EnterpriseCustomerFactory()
+        enterprise_configuration = SAPSuccessFactorsEnterpriseCustomerConfigurationFactory(
+            enterprise_customer=enterprise_customer,
+            idp_id='test-id'
+        )
+        assert SapSuccessFactorsLearnerExporter(
```
binodpant, line 71: this is what tests that the instance level variable is False for SAP Exporter

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_learner_data.py
```diff
@@ -89,6 +89,12 @@ def setUp(self):
         assert isinstance(self.exporter, LearnerExporter)
         super().setUp()
 
+    def test_default_channel_settings(self):
+        """
+        If you add any settings to the ChannelSettingsMixin, add a test here for the common default value
+        """
+        assert LearnerExporter('fake-user', self.config).INCLUDE_GRADE_FOR_COMPLETION_AUDIT_CHECK is True
```
binodpant, line 96: this is what tests that the instance level variable is False for SAP Exporter

integrated_channels/integrated_channel/transmitters/learner_data.py
```diff
@@ -373,8 +381,8 @@ def _log_exception_supplemental_data(self, learner_data, operation_name,
                 payload=learner_data
             )), exc_info=True)
 
-    def handle_transmission_error(self, learner_data, client_exception,
-                                  integrated_channel_name, enterprise_customer_uuid, learner_id, course_id):
+    def log_transmission_error(self, learner_data, client_exception,
```
alex-sheehan-edx, line 384: 👍  on making this clearer

integrated_channels/sap_success_factors/exporters/learner_data.py
```diff
@@ -25,6 +25,8 @@ class SapSuccessFactorsLearnerExporter(LearnerExporter):
     Class to provide a SAPSF learner data transmission audit prepared for serialization.
     """
 
+    INCLUDE_GRADE_FOR_COMPLETION_AUDIT_CHECK = False
```
alex-sheehan-edx, line 28: Could I get a TLDR on how this works? Will it override the mixin passed to the base exporter?

integrated_channels/sap_success_factors/transmitters/learner_data.py
```diff
@@ -20,6 +20,8 @@ class SapSuccessFactorsLearnerTransmitter(LearnerTransmitter):
     sent to SuccessFactors.
     """
 
+    INCLUDE_GRADE_FOR_COMPLETION_AUDIT_CHECK = False
```
alex-sheehan-edx, line 23: Ditto for this override the mixin for the transmitter!

integrated_channels/sap_success_factors/transmitters/learner_data.py
```diff
@@ -61,5 +63,5 @@ def handle_transmission_error(self, learner_data, client_exception, integrated_c
                     ecu.user_id, ecu.id, ecu.enterprise_customer
                 )
                 return
-        super().handle_transmission_error(learner_data, client_exception,
-                                          integrated_channel_name, enterprise_customer_uuid, learner_id, course_id)
+        super().log_transmission_error(learner_data, client_exception,
```
alex-sheehan-edx, line 66: NIT: I think we can format this closer to our other method/params

integrated_channels/utils.py
```diff
@@ -185,7 +185,13 @@ def get_image_url(content_metadata_item):
     return image_url
 
 
-def is_already_transmitted(transmission, enterprise_enrollment_id, grade, subsection_id=None):
+def is_already_transmitted(
+    transmission,
+    enterprise_enrollment_id,
+    grade,
+    subsection_id=None,
+    detect_grade_updated=True,
```
alex-sheehan-edx, line 193: Genuinely curious, do we need this to have a default value?

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_learner_data.py
alex-sheehan-edx: nice test!

tests/test_integrated_channels/test_sap_success_factors/test_transmitters/test_learner_data.py
```diff
@@ -223,3 +223,11 @@ def test_transmit_by_course_id_success(self):
         assert payloads[0].error_message == 'error occurred'
         assert payloads[1].status == '200'
         assert payloads[1].error_message == ''
+
+    def test_override_channel_settings(self):
```
alex-sheehan-edx, line 227: Thought- is it worth it to leave a comment on the base mixin/here/somewhere about how if anyone in the future wants to override the default mixin they should add a test like this to the channel?

tests/test_integrated_channels/test_utils.py
```diff
@@ -146,3 +147,80 @@ def test_is_course_completed_nonaudit_incomplete(self, mock_get_course_run_for_e
...
+        if grade_should_match_audit:
+            type(mock_audit_record_from_audit_logs).grade = grade_match
+        else:
+            type(mock_audit_record_from_audit_logs).grade = grade_non_matched
+
+        mock_audit_record.latest.return_value = mock_audit_record_from_audit_logs
+
+        transmission.objects.filter.return_value = mock_audit_record
+
+        enterprise_enrollment_id = 'a'
```
alex-sheehan-edx, line 180: NIT: Could we maybe make this a little more ID like?

tests/test_integrated_channels/test_sap_success_factors/test_transmitters/test_learner_data.py
```diff
@@ -223,3 +223,11 @@ def test_transmit_by_course_id_success(self):
...
+        """
+        If you add any settings to the ChannelSettingsMixin, add a test here for the common default value
+        """
+        assert learner_data.SapSuccessFactorsLearnerTransmitter(
+            self.enterprise_config
+        ).INCLUDE_GRADE_FOR_COMPLETION_AUDIT_CHECK is False
```
alex-sheehan-edx, line 233: is it worth it to test that a SAP transmission will skip regardless of whether the grade has changed?

alex-sheehan-edx (approved):
Nice to learn about mixins! Definitely something that we could utilize going forward. Let's try and keep an eye out for similar situations. A couple of nits but after those are addressed this LGTM!

integrated_channels/sap_success_factors/transmitters/learner_data.py
binodpant: see the other answer

integrated_channels/utils.py
binodpant: good q, I don't think so, it was a safety measure in case we didn't pass a value

tests/test_integrated_channels/test_sap_success_factors/test_transmitters/test_learner_data.py
binodpant: yeah I think best place for that is the docstring on the mixin itself

tests/test_integrated_channels/test_sap_success_factors/test_transmitters/test_learner_data.py
binodpant: hmm I can look at that , it should be doable to test that SAP specific transmitter let me check

binodpant: added test!

binodpant:
jenkins run quality

integrated_channels/integrated_channel/transmitters/learner_data.py
```diff
@@ -381,9 +381,30 @@ def _log_exception_supplemental_data(self, learner_data, operation_name,
...
-    def handle_transmission_error(self, learner_data, client_exception,
-                                  integrated_channel_name, enterprise_customer_uuid, learner_id, course_id):
-        """Handle the case where the transmission fails."""
+    def handle_transmission_error(self, learner_data, client_exception):
+        """
+        Subclasses who wish to do additional processing of transmission error
+        before logging, can do so in overrides of this method
+        """
+
+    def process_transmission_error(
```
binodpant, line 390: I am now using what is called the 'template' design pattern here meaning subclasses are only allowed to change parts of this method. In this case, logging always happens for free and subclasses don't need to call super().log_... anymore. If they do want additional processing of the error though, before logging, they can override the handle_ method (which only SAPSF does right now)

PR #1407, opened 2021-11-17
Backfilling course end dates for canvas

kiram15:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/canvas/client.py
+++ b/integrated_channels/canvas/client.py
@@ -349,32 +349,31 @@ class CanvasAPIClient(IntegratedChannelApiClient):
 
         return status_code, message
 
-    def update_participation_types(self, canvas_pk):
+    def update_participation_types(self, canvas_pks):
         """
         For each canvas course provided, send an update with the parameter of
         'restrict_enrollments_to_course_dates' to true in order to set the participation types of
         all canvas courses to "Course" instead of "Term", which allows users to view the end date
 
         Args:
-            - canvas_pk: primary key for canvas integrated channel
+            - canvas_pk: list of primary keys of previously transmitted courses from canvas customers
         """
         self._create_session()
 
-        integration_id = canvas_pk
-        course_id = CanvasUtil.get_course_id_from_edx_course_id(
-            self.enterprise_configuration,
-            self.session,
-            integration_id,
-        )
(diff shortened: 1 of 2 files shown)


integrated_channels/canvas/client.py
```diff
@@ -349,6 +349,33 @@ def cleanup_duplicate_assignment_records(self, courses):
...
         return status_code, message
 
+    def update_participation_types(self, canvas_pk):
+        """
+        For each canvas course provided, send an update with the parameter of
+        'restrict_enrollments_to_course_dates' to true in order to set the participation types of
+        all canvas courses to "Course" instead of "Term", which allows users to view the end date
+
+        Args:
+            - canvas_pk: primary key for canvas integrated channel
```
alex-sheehan-edx, line 359: > primary key for canvas integrated channel

I'm a little confused, is `canvas_pk` the edx course key?

integrated_channels/integrated_channel/management/commands/backfill_course_end_dates.py
```diff
@@ -0,0 +1,63 @@
...
+class Command(IntegratedChannelCommandMixin, BaseCommand):
+    """
+    Update content transmission items to have their respective catalog's uuid.
+    """
+
+    def add_arguments(self, parser):
+        """
+        Add required arguments to the parser.
+        """
+        parser.add_argument(
```
alex-sheehan-edx, line 24: I'm curious if we need a catalog user argument at all, are we making any API calls to the catalog service? Is there a security reason why we'd need this or another user for the job? 🤔

integrated_channels/canvas/client.py
```diff
@@ -349,6 +349,33 @@ def cleanup_duplicate_assignment_records(self, courses):
 
         return status_code, message
 
+    def update_participation_types(self, canvas_pk):
```
alex-sheehan-edx, line 352: If you wanted to be _**extra**_  thorough you could write a test for this method, but I realizing we want to get this out and I don't think it's overly complicated so you can skip for now.

alex-sheehan-edx (approved):
Some doc q's and I'm not sure if a user is necessary but otherwise this looks good to me!

PR #1415, opened 2021-11-30
Mm/ent 5047

manny-m:
**Merge checklist:**

**Post merge:**


enterprise/api/v1/views.py
```diff
@@ -1483,3 +1483,42 @@ def update(self, request, *args, **kwargs):
...
+        try:
+            enterprise_customer_key_match = models.EnterpriseCustomerInviteKey.objects.get(pk=enterprise_customer_key)
+
+            if not enterprise_customer_key_match.is_valid:
+                return Response(
+                    {"detail": "Enterprise customer invite key is not valid"},
+                    status=status.HTTP_422_UNPROCESSABLE_ENTITY
+                )
+
+            enterprise_customer = models.EnterpriseCustomer.objects.get(
```
christopappas, line 1514: Because EnterpriseCustomer is a foreign key on EnterpriseCustomerInviteKey, you can actually remove this and instead drop `enterprise_customer_key_match.enterprise_customer` directly into the get_or_create for the EnterpriseCustomerUser

```
models.EnterpriseCustomerUser.objects.get_or_create(
                user_id=request.user.pk,
                enterprise_customer=enterprise_customer_key_match.enterprise_customer,
            )
```

enterprise/api/v1/views.py
```diff
@@ -1483,3 +1483,42 @@ def update(self, request, *args, **kwargs):
...
+                pk=enterprise_customer_key_match.enterprise_customer.uuid
+            )
+            models.EnterpriseCustomerUser.objects.get_or_create(
+                user_id=request.user.pk,
```
christopappas, line 1518: I don't think it's wrong, but we do seem to use `user.id` more than `user.pk` throughout the codebase, might be worth changing for consistency

enterprise/api/v1/views.py
```diff
@@ -1483,3 +1483,42 @@ def update(self, request, *args, **kwargs):
...
+    permission_classes = (permissions.IsAuthenticated,)
+
+    @action(methods=['post'], detail=False)
+    def link_user(self, request, enterprise_customer_key):
+        """
+        Post
+        Links user using enterprise_customer_key
+        """
+        try:
+            enterprise_customer_key_match = models.EnterpriseCustomerInviteKey.objects.get(pk=enterprise_customer_key)
```
christopappas, line 1506: is `enterprise_customer_key` a uuid? even if the EnterpriseCustomerInviteKey's primary key is a uuid, i think it would be clearer to use `uuid=enterprise_customer_key` instead of `pk=enterprise_customer_key`

christopappas:
Do we still need to add the `enterprise_customer_key` field on the EnterpriseCustomerUser model? (Assuming it's foreign key?) Not 100% sure, just peaking at the AC

enterprise/api/v1/views.py
```diff
@@ -1483,3 +1483,42 @@ def update(self, request, *args, **kwargs):
...
+                    status=status.HTTP_422_UNPROCESSABLE_ENTITY
+                )
+
+            enterprise_customer = models.EnterpriseCustomer.objects.get(
+                pk=enterprise_customer_key_match.enterprise_customer.uuid
+            )
+            models.EnterpriseCustomerUser.objects.get_or_create(
+                user_id=request.user.pk,
+                enterprise_customer=enterprise_customer,
+            )
```
christopappas, line 1520: I think we need to create the enterprise learner role for the EnterpriseCustomerUser before we return the response. Might need to be a get_or_create? In case this endpoint gets called twice for the same enterprise we only have the 1 role 🤔

manny-m: So it looks like we have a [signal that does this](https://github.com/edx/edx-enterprise/blob/140c8b61c0400baa8b1f008382f389cc48740cac/enterprise/signals.py#L142) for us! I tested this in my local env and when enterprise learners are created their roles are automatically assigned.

christopappas: iiiinteresting. good find!

tests/test_enterprise/api/test_views.py
```diff
@@ -4977,3 +4977,90 @@ def test_put_method_not_allowed(self):
...
+            expiration_date=datetime.utcnow() + timedelta(days=365)
+        )
+        invalid_enterprise_customer_invite_key = factories.EnterpriseCustomerInviteKeyFactory(
+            enterprise_customer=enterprise_customer,
+            expiration_date=datetime.utcnow() - timedelta(days=365)
+        )
+
+        self.enterprise_customer = enterprise_customer
+        self.enterprise_customer_invite_key = enterprise_customer_invite_key
+        self.invalid_enterprise_customer_invite_key = invalid_enterprise_customer_invite_key
```
christopappas, line 5011: You can actually drop these and assign the values returned by the factories directly to `self. enterprise_customer` etc above

CHANGELOG.rst
```diff
@@ -17,6 +17,9 @@ Unreleased
 ----------
 * Nothing
 
+[3.35.2]
```
adamstankiewicz, line 20: nit: since this a new feature, should the version be bumped to 3.36.0?

enterprise/api/v1/views.py
```diff
@@ -1483,3 +1484,54 @@ def update(self, request, *args, **kwargs):
...
+
+    Given a enterprise_customer_key, link user to the appropriate enterprise.
+
+    If the key is not found, returns 400
+    If the ket is not valid, return 422
+    """
+    permission_classes = (permissions.IsAuthenticated,)
+
+    @action(methods=['post'], detail=False)
+    def link_user(self, request, enterprise_customer_key):
```
adamstankiewicz, line 1502: Curious to hear your thoughts on whether this should be its own view class? An alternative might be for it to be an action on the existing `EnterpriseCustomerInviteKeyViewSet` (since it already has to do with the EnterpriseCustomerInviteKey), where the endpoint URL might be:

```/enterprise/api/v1/enterprise-customer-invite-key/<enterprise_customer_invite_key_uuid>/link-user/```

File: enterprise/api/v1/views.py [hunk 8]
```diff
@@ -1483,3 +1484,54 @@ def update(self, request, *args, **kwargs):
...
+
+            if not enterprise_customer_key_match.is_valid:
+                return Response(
+                    {"detail": "Enterprise customer invite key is not valid"},
+                    status=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                )
+
+            enterprise_customer = enterprise_customer_key_match.enterprise_customer
+
+            _, created = models.EnterpriseCustomerUser.objects.get_or_create(
```
long74100, line 1518: following up on the discussion we were having about already linked users, we probably can't use get_or_create here with the invite key

enterprise/api/v1/views.py
```diff
@@ -1483,3 +1484,54 @@ def update(self, request, *args, **kwargs):
...
+    permission_classes = (permissions.IsAuthenticated,)
+
+    @action(methods=['post'], detail=False)
+    def link_user(self, request, enterprise_customer_key):
+        """
+        Post
+        Links user using enterprise_customer_key
+        """
+        try:
+            enterprise_customer_key_match = models.EnterpriseCustomerInviteKey.objects.get(uuid=enterprise_customer_key)
```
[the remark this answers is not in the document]
manny-m: My javascript brain wouldn't think of this! 🥜  Thank you.

enterprise/api/v1/views.py
```diff
@@ -1483,3 +1484,54 @@ def update(self, request, *args, **kwargs):
...
+
+            enterprise_customer = enterprise_customer_key_match.enterprise_customer
+
+            _, created = models.EnterpriseCustomerUser.objects.get_or_create(
+                user_id=request.user.id,
+                enterprise_customer=enterprise_customer,
+                invite_key=enterprise_customer_key_match
+            )
+
+            if created:
```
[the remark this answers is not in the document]
manny-m: Good call out! Fixed!

enterprise/api/v1/views.py
manny-m: You convinced me! Fixed!

enterprise/api/v1/views.py
```diff
@@ -1483,3 +1484,51 @@ def update(self, request, *args, **kwargs):
     )
     def destroy(self, request, *args, **kwargs):
         return super().destroy(request, *args, **kwargs)
+
+    @action(methods=['post'], detail=True, url_path='link-user')
+    def link_user(self, request, pk=None):
+        """
+        Post
+        Links user using enterprise_customer_key
+        /enterprise/api/enterprise_user_link/{enterprise_customer_key}/link-user
```
adamstankiewicz, line 1493: nit: The URL is now different with this as an action under `EnterpriseCustomerInviteKeyViewSet`. Should it be?

```
/enterprise/api/v1/enterprise-customer-invite-key/<enterprise_customer_invite_key_uuid>/link-user/
```

tests/test_enterprise/api/test_views.py
```diff
@@ -4977,3 +4989,54 @@ def test_put_method_not_allowed(self):
         self.assertEqual(response.status_code, 405)
         response = response.json()
         self.assertEqual(response['detail'], 'Method "PUT" not allowed.')
+
+    def test_successful_link(self):
```
adamstankiewicz, line 4993: Should we also have test case(s) for when an existing EnterpriseCustomerUser with `linked=False` and `active=False` calls `/link-user/` action, asserting those fields on their existing record should be flipped to `True`?

enterprise/api/v1/views.py
```diff
@@ -1483,3 +1484,51 @@ def update(self, request, *args, **kwargs):
...
+                request.user.id,
+                pk,
+                enterprise_customer.uuid,
+            )
+        elif not user.active:
+            user.active = True
+            user.save()
+
+        return Response(
+            {"enterprise_customer_slug": enterprise_customer.slug},
```
adamstankiewicz, line 1532: [suggestion] Might also be worth including the enterprise_customer.uuid in the response here?

tests/test_enterprise/api/test_views.py
```diff
@@ -4977,3 +4989,54 @@ def test_put_method_not_allowed(self):
...
+        client = APIClient()
+        client.login(username=self.USERNAME, password=TEST_PASSWORD)
+
+        response = client.post(
+            settings.TEST_SERVER + reverse(
+                self.ENTERPRISE_CUSTOMER_INVITE_KEY_ENDPOINT_LINK_USER,
+                kwargs={'pk': self.enterprise_customer_3_invite_key.uuid}
+            )
+        )
+        self.assertEqual(response.status_code, 201)
```
adamstankiewicz, line 5014: Should also assert on the API response (e.g., does it return the customers slug in the response)?

tests/test_enterprise/api/test_views.py
manny-m: Good catch, we check for both now!

enterprise/api/v1/views.py
```diff
@@ -1483,3 +1484,56 @@ def update(self, request, *args, **kwargs):
...
+
+        Given a enterprise_customer_key, link user to the appropriate enterprise.
+
+        If the key is not found, returns 400
+        If the ket is not valid, return 422
+        """
+        try:
+            enterprise_customer_key_match = models.EnterpriseCustomerInviteKey.objects.get(uuid=pk)
+        except models.EnterpriseCustomerInviteKey.DoesNotExist:
+            return Response({"error": "Could not find Enterprise Customer Invite Key"}, status=HTTP_400_BAD_REQUEST)
```
iloveagent57, line 1503: Nit: I think a 404 is more appropriate if the object can't be found.

enterprise/api/v1/views.py
```diff
@@ -1483,3 +1484,56 @@ def update(self, request, *args, **kwargs):
...
+                pk,
+                enterprise_customer.uuid,
+            )
+        elif not enterprise_user.active or not enterprise_user.linked:
+            if not enterprise_user.active:
+                enterprise_user.active = True
+            if not enterprise_user.linked:
+                enterprise_user.linked = True
+            enterprise_user.save()
+
```
iloveagent57, line 1532: You might want to return something other than a 201 in this `elif` block, since the enterprise user wasn't created - maybe a 200?

Also, I think you need an `else` path here to cover the case where the enterprise user already exists, is active, and linked to the enterprise.  And in that case you'd probably want to return a 204 response code.

iloveagent57 (commented):
Nice job on this, and very thorough tests! Just a couple small suggestions.

enterprise/api/v1/views.py
manny-m: This makes sense to me!

adamstankiewicz: There is a `get_object_or_404` method you might use here.

enterprise/api/v1/views.py
manny-m: The blocks now return 200 codes. 

204 (`HTTP_204_NO_CONTENT`) would mean that our response wouldn't have the content `enterprise_customer_slug` and `enterprise_customer_uuid` which we would still need.

enterprise/api/v1/views.py
```diff
@@ -33,7 +34,7 @@
 from django.apps import apps
 from django.conf import settings
 from django.core import exceptions, mail
-from django.http import Http404
+from django.http import Http404, response
```
long74100: nit: wrong/unused response import

enterprise/api/v1/views.py
```diff
@@ -1483,3 +1485,63 @@ def update(self, request, *args, **kwargs):
...
+    def link_user(self, request, pk=None):
+        """
+        Post
+        Links user using enterprise_customer_key
+        /enterprise/api/enterprise-customer-invite-key/{enterprise_customer_key}/link-user
+
+        Given a enterprise_customer_key, link user to the appropriate enterprise.
+
+        If the key is not found, returns 404
+        If the ket is not valid, return 422
```
long74100, line 1498: nit: key typo & inconsistent phrasing for returns/return

iloveagent57 (approved):
lgtm

enterprise/api/v1/views.py
```diff
@@ -1483,3 +1484,63 @@ def update(self, request, *args, **kwargs):
...
+            enterprise_user.invite_key = enterprise_customer_key
+            enterprise_user.save()
+            track_enterprise_user_linked(
+                request.user.id,
+                pk,
+                enterprise_customer.uuid,
+            )
+            return Response(response_body, status=HTTP_201_CREATED, headers=headers)
+
+        elif not enterprise_user.active or not enterprise_user.linked:
```
long74100, line 1536: I think we need to set the invite_key here when reactivating/linking here

long74100 (approved):

enterprise/api/v1/views.py
```diff
@@ -1483,3 +1484,63 @@ def update(self, request, *args, **kwargs):
...
+        headers = self.get_success_headers(response_body)
+
+        if created:
+            enterprise_user.invite_key = enterprise_customer_key
+            enterprise_user.save()
+            track_enterprise_user_linked(
+                request.user.id,
+                pk,
+                enterprise_customer.uuid,
+            )
```
adamstankiewicz, line 1533: Should `track_enterprise_user_linked` also be called below when an EnterpriseCustomerUser record already existed but we flipped `linked/active` to True?

PR #1421, opened 2021-12-03
feat: add is_active on enterprise customer invite key

long74100:
**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -3103,10 +3112,30 @@ def is_valid(self):
         is_not_expired = now < self.expiration_date
         usage_count = self.linked_enterprise_customer_users.count()
         is_usage_under_limit = usage_count < self.usage_limit
-        return is_not_expired and is_usage_under_limit
+        return is_not_expired and is_usage_under_limit and self.is_active
```
iloveagent57, line 3115: nit: I think this reads a little more naturally if `self.is_active` is the first clause of this conjunction.  And it might save a query, too.

enterprise/api/v1/serializers.py
```diff
@@ -1208,19 +1212,26 @@ def save(self):  # pylint: disable=arguments-differ
...
+class EnterpriseCustomerInviteKeyUpdateSerializer(BaseEnterpriseCustomerInviteKeySerializer):
+    """
+    Serializer for updating the EnterpriseCustomerInviteKey model.
+    """
+
+    uuid = serializers.UUIDField(read_only=True)
+    enterprise_customer_uuid = serializers.UUIDField(read_only=True)
+    usage_limit = serializers.IntegerField(read_only=True)
+    expiration_date = serializers.DateTimeField()
+    is_active = serializers.BooleanField()
```
iloveagent57, line 1224: Should any of these be marked as `required=False`?

enterprise/api/v1/serializers.py
```diff
@@ -1208,19 +1212,26 @@ def save(self):  # pylint: disable=arguments-differ
...
-        model = models.EnterpriseCustomerInviteKey
-        fields = (
-            'uuid',
-            'enterprise_customer_uuid',
-            'enterprise_customer_name',
-            'is_valid'
-        )
+    class Meta(BaseEnterpriseCustomerInviteKeySerializer.Meta):
+        fields = BaseEnterpriseCustomerInviteKeySerializer.Meta.fields + \
+            ('enterprise_customer_uuid', 'enterprise_customer_name')
```
iloveagent57, line 1234: nit: `enterprise_customer_uuid` is already in the base class meta's fields.  Or does the `SerializerMethodField` not work without this here?

iloveagent57 (commented):
Looks good, should we add a test for the `update()` action in the viewset?

enterprise/admin/__init__.py
```diff
@@ -1004,13 +1004,26 @@ class EnterpriseCustomerInviteKeyAdmin(admin.ModelAdmin):
         'enterprise_customer',
         'usage_limit',
         'expiration_date',
+        'is_active',
     )
```
adamstankiewicz, line 1008: Unrelated: while we're in here, should we add the ability to search for invite keys via the EnterpriseCustomer.{slug, name, and/or UUID}? Imagine that will be helpful in the not too distant future for ourselves and ECS.

enterprise/api/v1/serializers.py
long74100: it works without, good call

iloveagent57 (approved):

PR #1424, opened 2021-12-07
feat: OAuth Auth link for Blackboard Admin

johnnagro:
# Description

A small win to reduce setup complexity. Generate oauth auth URL in Django Admin once required fields are available.

- [ENT-4389](https://openedx.atlassian.net/browse/ENT-4389)

Screen Shot 2021-12-07 at 3 00 58 PM
)

### DIFF
--- a/integrated_channels/blackboard/admin/__init__.py
+++ b/integrated_channels/blackboard/admin/__init__.py
@@ -13,6 +13,7 @@ from integrated_channels.blackboard.models import BlackboardEnterpriseCustomerCo
 
 LMS_OAUTH_REDIRECT_URL = urljoin(settings.LMS_ROOT_URL, '/blackboard/oauth-complete')
 
+
 @admin.register(BlackboardEnterpriseCustomerConfiguration)
 class BlackboardEnterpriseCustomerConfigurationAdmin(admin.ModelAdmin):
     """
--- a/integrated_channels/blackboard/admin/__init__.py
+++ b/integrated_channels/blackboard/admin/__init__.py
@@ -49,9 +49,9 @@ class BlackboardEnterpriseCustomerConfigurationAdmin(admin.ModelAdmin):
                 being rendered with this admin form.
         """
         if obj.blackboard_base_url and obj.client_id:
-            return f"{obj.blackboard_base_url}/learn/api/public/v1/oauth2/authorizationcode"\
-            f"?redirect_uri=https://courses.edx.org/blackboard/oauth-complete&"\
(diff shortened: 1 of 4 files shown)


integrated_channels/blackboard/admin/__init__.py
```diff
@@ -38,3 +39,19 @@ def enterprise_customer_name(self, obj):
...
+    def oauth_authorization_url(self, obj):
+        """
+        Returns: the oauth authorization url when the blackboard_base_url and client_id are available.
+
+        Args:
+            obj: The instance of BlackboardEnterpriseCustomerConfiguration
+                being rendered with this admin form.
+        """
+        if obj.blackboard_base_url and obj.client_id:
+            return (f"{obj.blackboard_base_url}/learn/api/public/v1/oauth2/authorizationcode"
```
binodpant, line 52: I think this base url also needs to be parameterized? `https://courses.edx.org`. Meaning we may also have staging configurations, or local. Figure this is best obtained from a LMS_BASE_URL settings somewhere?

integrated_channels/blackboard/admin/__init__.py
```diff
@@ -38,3 +39,19 @@ def enterprise_customer_name(self, obj):
                 being rendered with this admin form.
         """
         return obj.enterprise_customer.name
+
+    def oauth_authorization_url(self, obj):
```
binodpant, line 43: this is not strictly speaking needed, but how hard would it be to make this a hyperlink in the django admin world? I do not know either, figure it's not worth it unless easy to do

integrated_channels/blackboard/admin/__init__.py
johnnagro: i believe the `blackboard_base_url` is user supplied as part of the configuration - it points to their instance.

binodpant: oh sorry I meant in the redirect_uri

binodpant: would it be `settings.LMS_BASE_URL/blackboard/oauth-complete' ?

integrated_channels/blackboard/admin/__init__.py
johnnagro: sure!

integrated_channels/blackboard/admin/__init__.py
johnnagro: oh i see what you mean - of course. let me take a look.

binodpant (approved):
lgtm!

PR #1425, opened 2021-12-10
feat: OAuth Auth link for Blackboard Admin

johnnagro:
A better approach to https://github.com/edx/edx-enterprise/pull/1424 which adds the property to the model and serializer so it can be used beyond just django admin (enables display in the self-service portal)

[ENT-4389](https://openedx.atlassian.net/browse/ENT-4389)

Screen Shot 2021-12-07 at 3 00 58 PM
)

Screen Shot 2021-12-10 at 8 30 34 AM

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * Nothing
 
+[3.36.6]
+--------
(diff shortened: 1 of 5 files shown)


binodpant (approved):
love this changed approach! will be very helpful in reusing in the admin form later

PR #1426, opened 2021-12-15
feat: update invite key filter and serializer

long74100:
- Add usage count to serializer
- Update filter backend to support filtering by enterprise customer uuid

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/filters.py
+++ b/enterprise/api/filters.py
@@ -8,8 +8,7 @@ from rest_framework.exceptions import ValidationError
 
 from django.contrib import auth
 
-from enterprise.api.utils import get_enterprise_customer_from_user_id
-from enterprise.models import SystemWideEnterpriseUserRoleAssignment
+from enterprise.models import EnterpriseCustomerUser, SystemWideEnterpriseUserRoleAssignment
 
 User = auth.get_user_model()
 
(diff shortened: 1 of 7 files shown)


adamstankiewicz (approved):
LGTM!

enterprise/api/filters.py
```diff
@@ -128,17 +127,23 @@ def filter_queryset(self, request, queryset, view):
...
-            queryset = queryset.filter(enterprise_customer_id=enterprise_customer_uuid)
+            user_enterprise_customer_uuids = [
+                ec.enterprise_customer_id for ec in EnterpriseCustomerUser.objects.filter(user_id=request.user.id)
+            ]
+            query_params.update(enterprise_customer_id__in=user_enterprise_customer_uuids)
 
-        return queryset
+        enterprise_customer_uuid = request.query_params.get('enterprise_customer_uuid', None)
+        if enterprise_customer_uuid:
+            query_params.update(enterprise_customer_id=enterprise_customer_uuid)
```
macdiesel, line 147: `enterprise_customer_id=enterprise_customer_uuid`
How hard would it be for us to fix the naming mismatch here? How deep does this run?

enterprise/api/filters.py
```diff
@@ -128,17 +127,23 @@ def filter_queryset(self, request, queryset, view):
...
         if not request.user.is_staff:
-            enterprise_customer_uuid = get_enterprise_customer_from_user_id(request.user.id)
-            queryset = queryset.filter(enterprise_customer_id=enterprise_customer_uuid)
+            user_enterprise_customer_uuids = [
+                ec.enterprise_customer_id for ec in EnterpriseCustomerUser.objects.filter(user_id=request.user.id)
+            ]
+            query_params.update(enterprise_customer_id__in=user_enterprise_customer_uuids)
 
-        return queryset
+        enterprise_customer_uuid = request.query_params.get('enterprise_customer_uuid', None)
```
macdiesel, line 145: Since this is on the request object I'm assuming it's user input. Is it safe for us to use this to query this info? Could an actor provide an enterprise_customer_uuid that wasn't theirs and gain access to info they shouldn't have?

Would it be better to get the info enterprise_customer info from the request.user object?

macdiesel (commented):
A few questions.

enterprise/api/filters.py
long74100: If the user is not staff, they can only view keys in enterprises they are part of.

enterprise/api/filters.py
long74100: Django automatically populates the foreign key column name by appending _id. We could use db_column='enterprise_customer_uuid' to specify a name but we don't do it for any of the foreign keys we have. Alternatively query_params.update(enterprise_customer=enterprise_customer_uuid) should also work here.

adamstankiewicz: IIRC @long74100 and I were chatting about these mismatches the other day. fwiw just did a quick search across the edX github org... 

uses of `enterprise_id` in code: 104
uses of `enterprise_uuid` in code:  84
uses of `enterprise_customer_id` in code: 113
uses of `enterprise_customer_uuid` in code:  476

Might be part of a broader tech debt initiative to add to our backlog? 🤔

macdiesel: I can add this as a tech debt item. i would assume that our default should always be the UUID moniker?

enterprise/api/filters.py
macdiesel: Fair enough, I didn't put that bit together.

macdiesel (approved):

enterprise/api/filters.py
long74100: my vote is for uuid 🤔

PR #1428, opened 2021-12-21
feat: allow more than 1 lms configuration per lms-kind

johnnagro:
# Description

- added a uuid to canvas and blackboard configs for use in public urls
- use config-specific oauth state (uuid)
- change views to use new lookup, and fallback
- add oauth authorization url to canvas models

# References

- [ENT-5238](https://openedx.atlassian.net/browse/ENT-5238)

**Merge checklist:**

**Post merge:**


integrated_channels/blackboard/migrations/0010_auto_20211221_1532.py
```diff
@@ -0,0 +1,37 @@
+# Generated by Django 3.2.10 on 2021-12-21 15:32
+
+from django.db import migrations, models
+import uuid
+
+
+def create_uuid(apps, schema_editor):
```
binodpant, line 7: do we need a reverse data migration to undo this for rollbacks?

integrated_channels/canvas/migrations/0013_auto_20211221_1535.py
```diff
@@ -0,0 +1,37 @@
+# Generated by Django 3.2.10 on 2021-12-21 15:35
+
+from django.db import migrations, models
+import uuid
+
+
+def create_uuid(apps, schema_editor):
+    CanvasEnterpriseCustomerConfiguration = apps.get_model('canvas', 'CanvasEnterpriseCustomerConfiguration')
```
binodpant, line 8: ditto: any need for reverse functions?

binodpant (approved):
other than the minor nits I am good with this change... it does not impact the other channels. but we will have to keep this impl intact if we add more oauth cases so maybe in future iterations we can think of some common pattern. but totally not worth right now to DRY beyond this

integrated_channels/blackboard/migrations/0010_auto_20211221_1532.py
johnnagro: we can throw them away.

integrated_channels/canvas/migrations/0013_auto_20211221_1535.py
johnnagro: we can throw them away.

PR #1434, opened 2022-01-11
Adds toggle_universal_link

manny-m:
### [1 of 2 PRs to close 5177](https://openedx.atlassian.net/browse/ENT-5177)

This PR adds the ability to easily toggle the `enable_universal_link` feature in the `EnterpriseCustomer` model. 

Toggling this has side affects as outlined in the ticket.

Test are WIP

**Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,7 +17,7 @@ Unreleased
 ----------
 * Nothing
 
-[3.37.00]
+[3.37.0]
 ---------
 feat: Adds toggle_universal_link endpoint
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * Nothing
 
+[3.37.00]
+---------
+feat: Adds toggle_universal_link endpoint
+
 [3.36.11]
 ---------
 fix: Integrated channels Degreed2 exporter now handles invalid start/end date in content metadata item
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -713,12 +713,13 @@ class EnterpriseCustomer(TimeStampedModel):
         )
         send_enterprise_email_notification.delay(self.uuid, admin_enrollment, email_items)
 
-    def toggle_universal_link(self, enable_universal_link, link_expiration_date):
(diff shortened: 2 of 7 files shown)


CHANGELOG.rst
```diff
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * Nothing
 
+[3.37.00]
```
iloveagent57, line 20: Nit: we typically use single '0's.

enterprise/models.py
```diff
@@ -713,6 +713,37 @@ def notify_enrolled_learners(self, catalog_api_user, course_id, users, admin_enr
...
+
+        # If universal link is being disabled
+        if not enable_universal_link:
+            # Deactivate all EnterpriseCustomerInviteKey
+            EnterpriseCustomerInviteKey.objects.filter(
+                enterprise_customer=self,
+                is_active=True,
+            ).update(is_active=False)
+        # If universal link is being enabled and a date is passed
+        elif bool(link_expiration_date):
```
iloveagent57, line 740: I think you need to check if any `EnterpriseCustomerInviteKeys` already exist when it's being toggled to `True` - if they do, we wouldn't want to create a new one, right?

iloveagent57 (commented):
This is looking good so far, minor suggestion/addition to the business logic.

enterprise/models.py
long74100, line 740: I think we can do ```elif link_expiration_date``` here.

enterprise/models.py
```diff
@@ -713,6 +713,37 @@ def notify_enrolled_learners(self, catalog_api_user, course_id, users, admin_enr
...
         send_enterprise_email_notification.delay(self.uuid, admin_enrollment, email_items)
 
+    def toggle_universal_link(self, enable_universal_link, link_expiration_date):
+        """
+        Sets enable_universal_link
+
+        If there is no change to be made, return
+
+        When enable_universal_link changes to;
+            True: a new link is created
```
long74100, line 723: should we reference customer invite key instead of links here?

enterprise/models.py
iloveagent57: Ah, that's right - the links active state is immutable!  So yeah, leave what you have. 👍

PR #1440, opened 2022-01-20
fix: inactivate other customers when linking pending ECU record

adamstankiewicz:
Ticket: https://openedx.atlassian.net/browse/ENT-5312

**Description**

Through this work, I found a few places where we were not appropriately setting `active` field on `EnterpriseCustomerUser` objects. I also added a management command that will allow us to clean up all enterprise users who have more than 1 active `EnterpriseCustomerUser` objects. When a user has multiple active `EnterpriseCustomerUser` objects, we will reset them all to `active=False` except for the one that was most recently `active`.

The management command will be run once to clean up the data of enterprise users in this incorrect state. I've added some logging as well so we have better observability into when we create enterprise course enrollments for users with multiple active `EnterpriseCustomerUser` objects.

**Merge checklist:**

**Post merge:**


tests/test_enterprise/management/test_ensure_singular_active_enterprise_customer_user.py
```diff
@@ -0,0 +1,57 @@
...
+
+from enterprise.models import EnterpriseCustomerUser
+from test_utils.factories import EnterpriseCustomerFactory, EnterpriseCustomerUserFactory
+
+EXCEPTION = "DUMMY_TRACE_BACK"
+
+
+@mark.django_db
+@ddt.ddt
+class CreateEnterpriseCourseEnrollmentCommandTests(TestCase):
```
adamstankiewicz, line 19: D'oh. Fixing this name now, copy/pasta oversight.

enterprise/management/commands/ensure_singular_active_enterprise_customer_user.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def _historical_enterprise_customer_user_model(self):
+        """ Returns `HistoricalEnterpriseCustomerUser model """
+        HistoricalEnterpriseCustomerUser = apps.get_model(
+            'enterprise',
+            'HistoricalEnterpriseCustomerUser',
+        )
+        return HistoricalEnterpriseCustomerUser
+
+    def _process_user(self, user):
```
manny-m, line 64: Cool! 🚀

manny-m (approved):
You caught lots of edge cases, nice!

enterprise/management/commands/ensure_singular_active_enterprise_customer_user.py
```diff
@@ -0,0 +1,107 @@
...
+        historical records of which EnterpriseCustomerUser object was flipped to `active=True`
+        most recently.
+        """
+        EnterpriseCustomerUser = self._enterprise_customer_user_model()
+        HistoricalEnterpriseCustomerUser = self._historical_enterprise_customer_user_model()
+        lms_user_id = user['user_id']
+        current_active_ecu_count = user['ecu_count']
+        log.info('Processing LMS User ID %s', lms_user_id)
+        active_historical_ecus = HistoricalEnterpriseCustomerUser.objects.filter(user_id=lms_user_id, active=True)
+        most_recent_active_ecu = active_historical_ecus.order_by('-history_date').first()
```
christopappas, line 77: fyi I think you can use the `.history` manager on the enterprise customer user to do something similar
```
my_enterprise_customer_user_object.history.filter(active=True)
```
really just a "btw", in case you wanted to approach it differently 🤷

enterprise/management/commands/ensure_singular_active_enterprise_customer_user.py
```diff
@@ -0,0 +1,107 @@
...
+
+        log.info(
+            'Found %s enterprise users with mulitple active EnterpriseCustomerUser objects',
+            len(user_ids_with_multiple_active_ecus),
+        )
+
+        for user_batch in batch(user_ids_with_multiple_active_ecus, batch_size=batch_limit):
+            for user in user_batch:
+                self._process_user(user)
+            sleep(batch_sleep)
```
iloveagent57, line 107: why sleep?

enterprise/management/commands/ensure_singular_active_enterprise_customer_user.py
christopappas: actually nevermind, there are multiple enterprise customer user objects, scratch this

adamstankiewicz: @christopappas So, the last time I worked with historical models (I think also in a management command), I had first resorted to the approach you're suggesting above since that's what's in the official documentation for django-simple-history. However, I had run into some odd error then (blanking on exactly what it was) so I just stuck with the current approach that worked before here.

christopappas: okay cool, np

christopappas:
So, I'm wondering if we should be adding/modifying a test (outside of the manage command test you added (ty btw)) to prove the new state of the code. Specifically, the tests we had previously didn't catch this regression, so I wonder if it's worth making a test to verify the previous bad state won't happen again

enterprise/models.py
```diff
@@ -1267,13 +1268,23 @@ def link_pending_enterprise_user(self, user, is_user_created):
                     user=user,
                     enterprise_customer=enterprise_customer_user.enterprise_customer,
                 ))
+
+                enterprise_customer_user.active = True
+                enterprise_customer_user.save()
+                enterprise_customer_user.inactivate_other_customers(
```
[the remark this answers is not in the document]
iloveagent57: Could we push this down into a more common place, like `EnterpriseCustomerUser.save()` or the model manager or something?  That way future writers of code that set `active = True` don't have to worry about this?

adamstankiewicz:
> So, I'm wondering if we should be adding/modifying a test (outside of the manage command test you added (ty btw)) to prove the new state of the code. Specifically, the tests we had previously didn't catch this regression, so I wonder if it's worth making a test to verify the previous bad state won't happen again

Valid point! No tests covered it because this is new logic, which makes sense. I will update and/or some tests.

enterprise/signals.py
```diff
@@ -333,9 +333,18 @@ def create_enterprise_enrollment_receiver(sender, instance, **kwargs):     # pyl
...
+        ecu = active_ecus_for_user.first()
         if not ecu:
+            # nothing to do here
             return
+        if len(active_ecus_for_user) > 1:
+            logger.warning(
+                'User %s has more than 1 active EnterpriseCustomerUser object. Continuing with course enrollment'
+                ' but the enrollment may end up associated with an incorrect EnterpriseCustomerUser',
+                user_id,
+            )
```
iloveagent57, line 347: Nit: would be helpful to log the course id here, too.

iloveagent57 (commented):
Awesome work sussing all of this out!  Really just the main suggestion for making the "only-one-active-ECU-ever" invariant happen in the most common place feasible.

enterprise/models.py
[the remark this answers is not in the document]
adamstankiewicz: @iloveagent57 I actually debated this and kept going back and forth about it haha. I ultimately decided not to put it in `EnterpriseCustomerUser.save()` as I felt it might have been inappropriate to tightly couple inactivating other customers on every `EnterpriseCustomerUser` save. The primary concern I had was if ECS or engineering manually updates an `EnterpriseCustomerUser` record or it somehow otherwise gets updated with the expectation of no side effects.

enterprise/management/commands/ensure_singular_active_enterprise_customer_user.py
adamstankiewicz: @iloveagent57 Honestly, just following convention with other batched management commands we have :) Probably safe to remove, but wasn't 100% sure so left it in.

enterprise/models.py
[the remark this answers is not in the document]
iloveagent57: What if we encapsulated the whole thing in a function like `EnterpriseCustomerUser.set_active()`, which would set that record with `active = True` and then call `inactivate_other_customers()`.?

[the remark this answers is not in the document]
adamstankiewicz: @iloveagent57 Then again, I suppose we would only call it when `active` changes from false to true for a given `EnterpriseCustomerUser`. Which wouldn't be on every `EnterpriseCustomerUser.save()`. I'm going to explore this a little further and re-evaluate I think.

iloveagent57 (approved):
Consider my last response a nit and at your discretion. :)

enterprise/models.py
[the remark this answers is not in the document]
adamstankiewicz: @iloveagent57 That's also a possibility. I'll sleep on it and see how ambitious I am in the morning haha!

macdiesel:
I just want to echo Alex's sentiment. Nice job here Adam.

enterprise/management/commands/ensure_singular_active_enterprise_customer_user.py
```diff
@@ -0,0 +1,107 @@
...
+        active_historical_ecus = HistoricalEnterpriseCustomerUser.objects.filter(user_id=lms_user_id, active=True)
+        most_recent_active_ecu = active_historical_ecus.order_by('-history_date').first()
+        ecus_for_user = EnterpriseCustomerUser.objects.filter(user_id=lms_user_id).exclude(id=most_recent_active_ecu.id)
+        for ecu in ecus_for_user:
+            ecu.active = False
+        EnterpriseCustomerUser.objects.bulk_update(ecus_for_user, ['active'])
+        log.info(
+            'Successfully updated %s of %s EnterpriseCustomerUser objects for LMS User ID %s',
+            len(ecus_for_user),
+            current_active_ecu_count,
```
long74100, line 85: is the order flipped here for len(ecus_for_user) and current_active_ecu_count?

brianhw:
Just a request going forward -- that when adding a migration that creates a new column which has a default value, it would be helpful to notify data engineering that the table should be reloaded via Stitch into Snowflake, because Stitch doesn't know about migrations.   Our automation picked this up this morning, but it would be good to head this off sooner.

PR #1441, opened 2022-01-20
fix: Cornerstone missing completion records

johnnagro:
## Description

To work around some encoding and string-length limitations with Cornerstone, we've gone through a few iterations of mapping internal course_ids to external cornerstone course ids. Sometimes they were the same, sometimes base64 encoded, sometimes some form of a uuid4. Meanwhile, because of the way they're created via a click-through, CornerstoneLearnerDataTransmissionAudit would always contain the internal course_id - I checked the data and the code. The way audit records were being looked-up for course-completion events was flawed in such a way that only courses who had the same internal and external course id representation would end up getting transmitted. In other words, newer mappings who were base64 encoded, or using a uuid4, would never find the appropraite audit record to transmit and fail to send completion events. This PR cleans up that lookup logic, always uses a string uuid for external mappings (to simplify logic), and cleans up and adds some comments.

- correct how CornerstoneLearnerDataTransmissionAudit are looked up
when sending completion events
- clean up how course_id to external_course_id records are mapped

## References

- [ENT-5255](https://openedx.atlassian.net/browse/ENT-5255)
- [ENT-4954](https://openedx.atlassian.net/browse/ENT-4954)

### DIFF
--- a/integrated_channels/cornerstone/utils.py
+++ b/integrated_channels/cornerstone/utils.py
@@ -10,14 +10,14 @@ from django.apps import apps
 
 def cornerstone_learner_data_transmission_audit():
     """
-    Returns the ``CornerstoneLearnerDataTransmissionAudit`` class.
+        Returns the ``CornerstoneLearnerDataTransmissionAudit`` class.
     """
     return apps.get_model('cornerstone', 'CornerstoneLearnerDataTransmissionAudit')
 
 
 def cornerstone_course_key_model():
     """
-    Returns the ``CornerstoneCourseKey`` class.
+        Returns the ``CornerstoneCourseKey`` class.
     """
     return apps.get_model('cornerstone', 'CornerstoneCourseKey')
 
--- a/integrated_channels/cornerstone/utils.py
+++ b/integrated_channels/cornerstone/utils.py
@@ -13,14 +13,14 @@ from integrated_channels.utils import encode_course_key_into_base64
 
 def cornerstone_learner_data_transmission_audit():
     """
-    Returns the ``EnterpriseCustomer`` class.
+    Returns the ``CornerstoneLearnerDataTransmissionAudit`` class.
     """
     return apps.get_model('cornerstone', 'CornerstoneLearnerDataTransmissionAudit')
 
 
 def cornerstone_course_key_model():
     """
-    Returns the ``EnterpriseCustomer`` class.
+    Returns the ``CornerstoneCourseKey`` class.
     """
     return apps.get_model('cornerstone', 'CornerstoneCourseKey')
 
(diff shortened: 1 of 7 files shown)


integrated_channels/cornerstone/utils.py
```diff
@@ -53,23 +50,18 @@ def create_cornerstone_learner_data(request, course_id):
 
 def convert_invalid_course_id(course_id):
```
binodpant, line 51: after this PR, do we even use this function anymore?

johnnagro: we do but i might be able to factor it out, here is one of the places we use it https://github.com/edx/edx-enterprise/blob/master/integrated_channels/cornerstone/exporters/content_metadata.py#L63

binodpant (approved):
as long as tests are updated and pass, I am go! discussed the logic offline

binodpant:
also discussed the future change idea of replacing our own uuid with the discovery uuids that already exist for each course, that will further simplify life for us I think

integrated_channels/cornerstone/utils.py
```diff
@@ -2,25 +2,22 @@
...
 from uuid import uuid4
 
 from django.apps import apps
 
-from integrated_channels.utils import encode_course_key_into_base64
-
 
 def cornerstone_learner_data_transmission_audit():
     """
-    Returns the ``EnterpriseCustomer`` class.
```
alex-sheehan-edx, line 16: whoa how'd that docstring slip through the cracks? 😬

integrated_channels/cornerstone/utils.py
```diff
@@ -53,23 +50,18 @@ def create_cornerstone_learner_data(request, course_id):
 
 def convert_invalid_course_id(course_id):
     """
-    Regex check a course ID to see if it contains any invalid chars. If it does then encode the string, otherwise
-    return the original course ID.
+        upsert a CornerstoneCourseKey object, return the external_course_id
```
alex-sheehan-edx, line 53: nit: docstring spacing just to be consistent?

integrated_channels/cornerstone/utils.py
```diff
@@ -53,23 +50,18 @@ def create_cornerstone_learner_data(request, course_id):
...
-    if len(safe_course_id) > 50:
-        safe_course_id = str(uuid4())
-    return safe_course_id
+    key_mapping = get_or_create_key_pair(course_id)
+    return key_mapping.external_course_id
 
 
 def get_or_create_key_pair(course_id):
+    """
+        upsert and return a CornerstoneCourseKey object, always use uuid4 for new external_course_id records
```
alex-sheehan-edx, line 61: nit: docstring spacing

alex-sheehan-edx (approved):
small formatting nits but otherwise LGTM

PR #1446, opened 2022-01-24
feat: mysql8 is generating internal checks for `PositiveInteger field

awais786:
feat: mysql8 is generating internal checks if model has `PositiveIntegerField`.

`error django.db.utils.OperationalError: (1059, "Identifier name 'blackboard_blackboardlearnerassessmentdatatransmissionaudit_chk_1' is too long")`

These check has random names with 64 characters limit. But due to long model names few checks breaks the limit and mysql8 throws error. Mysql [Documentation](https://dev.mysql.com/doc/refman/8.0/en/identifier-length.html)

`For constraint definitions that include no constraint name, the server internally generates a name derived from the associated table name. For example, internally generated foreign key and CHECK constraint names consist of the table name plus _ibfk_ or _chk_ and a number. If the table name is close to the length limit for constraint names, the additional characters required for the constraint name may cause that name to exceed the limit, resulting in an error.`

We need to inform devs to use small table names to avoid these issues.

**Solution 1**
Changed the field types to simple IntegerField.

**Solution 2**
https://docs.djangoproject.com/en/4.0/ref/models/constraints/

Create django meta.checkconstraint
```
constraints = [
            CheckConstraint(
                check=Q(enterprise_course_enrollment_id__gte=0),
                name='bb_enrollment_id_check',
            ),
        ]
```

[BOM-3197](https://openedx.atlassian.net/browse/BOM-3197)

Solution 1 is good enough.

in platform check works fine.
https://github.com/openedx/edx-platform/pull/29784


integrated_channels/blackboard/migrations/0005_blackboardlearnerassessmentdatatransmissionaudit.py
```diff
@@ -15,7 +15,7 @@ class Migration(migrations.Migration):
             fields=[
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                 ('blackboard_user_email', models.CharField(max_length=255)),
-                ('enterprise_course_enrollment_id', models.PositiveIntegerField(db_index=True)),
+                ('enterprise_course_enrollment_id', models.IntegerField(db_index=True)),
```
awais786, line 18: It requires for github migration check otherwise it will break with long check name error.

awais786:
Good article about checks https://dev.mysql.com/blog-archive/mysql-8-0-16-introducing-check-constraint/

integrated_channels/blackboard/migrations/0005_blackboardlearnerassessmentdatatransmissionaudit.py
robrap: Changing existing migrations is dangerous, and I'm not sure it is wanted in this case. If you were to change this, some deployments would have an IntegerField and some would have a PositiveIntegerField when you get to the later AlterField. Does the AlterField really do the right thing in both cases (e.g. remove the constraint for a PositiveIntegerField, and act as a no-op with no error for the IntegerField? I would think this wouldn't create the SQL to remove the constraint if it is going from an IntegerField to IntegerField.

Instead, if you add the AlterField, does the problem go away as soon as you squash migrations? Or, not until you get to phase 2 of squashing, when you can finally delete the old migrations?

If squashing (phase 1) is enough (and that's an open question for me), could we instead do the following:
1. Add the AlterField.
2. Squash migrations.
3. Celebrate no error?!
4. Delete old migrations after the next named release?

Note: If the problem isn't solved until the migration is actually deleted, this plan won't work unless the mysql8 upgrade can wait until after the next named release, because we would need to wait for the next named release before deleting the old migrations.

awais786:
**Data in prod for these tables.** 

`select count(*) from blackboard_blackboardlearnerassessmentdatatransmissionaudit;
6080528
`

`select count(*) from sap_success_factors_sapsuccessfactorslearnerdatatransmission3ce5;
2124578
`

integrated_channels/blackboard/migrations/0005_blackboardlearnerassessmentdatatransmissionaudit.py
awais786: @robrap I did these steps now 

1. Add the AlterField 
2. Generate new migration for alter field
2. Squash migration.
3. Check the hash against mysql8-check migrations in [platform](https://github.com/openedx/edx-platform/runs/4951340194?check_suite_focus=true#step:11:168) ( failed )

Squash migration includes this https://github.com/openedx/edx-enterprise/pull/1446/files#diff-82ed9615cb2cf73f24651c57f6b548ae58b7a99ca64900d4054b38e505f5381aR57 and when it run on mysql8 it crashes.

awais786: In case of squash this issue will appear because it will pick the `PositiveIntegerField` from initial migration how to handle this ?

1. Does the AlterField really do the right thing in both cases (e.g. remove the constraint for a PositiveIntegerField) **Yes it will solve the issue **
2. act as a no-op with no error for the IntegerField to IntegerField ( **no idea need to do discovery** )

robrap: @awais786: The squash migration is auto-generated, but you can manually update it before merging. You would just remove the `AlterField` and copy the change into the `CreateModel`. This is basically what you were wishing to do, but I think doing it inside the squash migration is the place to do it. Thus, any deployment that has some, but not all, of the squashed migrations would just run the rest of the non-squashed migrations, including the AlterField. However, any new deployment (or deployment that hasn't gotten to the replaced migrations) would instead run the squashed migration that would never refer to `PositiveIntegerField`. Hopefully, it is the latter case that would run for `mysql8-check` and would pass.

integrated_channels/blackboard/migrations/0001_initial_squashed_0011_alter_blackboardlearnerassessmentdatatransmissionaudit_enterprise_course_enrollment_id.py
```diff
@@ -0,0 +1,129 @@
...
+                ('error_message', models.TextField(blank=True)),
+                ('status', models.CharField(max_length=100)),
+            ],
+        ),
+        migrations.CreateModel(
+            name='BlackboardLearnerAssessmentDataTransmissionAudit',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('blackboard_user_email', models.CharField(max_length=255)),
+                ('enterprise_course_enrollment_id', models.IntegerField(db_index=True)),
```
awais786, line 57: Changed this from `Positive to Integer`.
Remove alter migrations code from bottom.

integrated_channels/sap_success_factors/migrations/0001_squashed_0022_auto_20200206_1046.py
```diff
@@ -22,7 +22,7 @@ class Migration(migrations.Migration):
             fields=[
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                 ('sapsf_user_id', models.CharField(max_length=255)),
-                ('enterprise_course_enrollment_id', models.PositiveIntegerField(db_index=True)),
+                ('enterprise_course_enrollment_id', models.IntegerField(db_index=True)),
```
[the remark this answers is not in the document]
robrap: Looks good. 
1. Can you confirm if this passes your mysql8 test?
2. Maybe add comments to the squash migration files (or wherever you think) stating that the replaced migrations should not be deleted until after nutmeg. I guess it could go in the replaced migration files as well.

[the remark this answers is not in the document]
awais786: 1: Yes now migration checks work fine with mysql8. Please check these migrations here with mysql8 [blackboard](https://github.com/openedx/edx-platform/runs/4965788270?check_suite_focus=true#step:11:150) and [sap](https://github.com/openedx/edx-platform/runs/4965788270?check_suite_focus=true#step:11:614).

2. Adding comments

integrated_channels/blackboard/migrations/0001_initial_squashed_0011_alter_blackboardlearnerassessmentdatatransmissionaudit_enterprise_course_enrollment_id.py
```diff
@@ -0,0 +1,131 @@
...
+import uuid
+
+
+# NOTE: the replaced migrations should not be deleted until after nutmeg.
+
+# Functions from the following migrations need manual copying.
+# Move them and any dependencies into this file, then update the
+# RunPython operations to refer to the local versions:
+# integrated_channels.blackboard.migrations.0010_auto_20211221_1532
+
```
robrap, line 19: Was this just about `create_uuid`? If so, you can remove this comment since this manual step is complete.

robrap (approved):
Thanks. I think this is ready. Just one minor comment that I think just needs to be deleted.

integrated_channels/blackboard/migrations/0001_initial_squashed_0011_alter_blackboardlearnerassessmentdatatransmissionaudit_enterprise_course_enrollment_id.py
awais786: > Thanks. I think this is mostly ready.
> 
> 1. Just one minor comment that I think just needs to be deleted.
> 2. I imagine you need to bump the version and update the CHANGELOG as well.

I just asked owners for the review. After their review I will bump the version. Would you please share your thoughts on running these migration on prod ? One table has 6 million and other has 2 million records.

robrap: 1. Read the section of https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations#EverythingAboutDatabaseMigrations-CheckingSQLforamigration on Checking SQL. Shouldn't this just be dropping a constraint?
2. Are you concerned if they want to add back the constraint, or about some other migration? I forgot you may want to discuss with them whether or not they want the constraint added back or not.

Am I answering your concern, or have I misunderstood something?

awais786: Thanks for document.
1. It's just dropping a constraint but want to double check with SRE because table has lots of records. Owning team review is important may be they have some different plan for this problem.
2. Comment removed.

integrated_channels/blackboard/migrations/0011_alter_blackboardlearnerassessmentdatatransmissionaudit_enterprise_course_enrollment_id.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('blackboard', '0010_auto_20211221_1532'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='blackboardlearnerassessmentdatatransmissionaudit',
+            name='enterprise_course_enrollment_id',
```
binodpant, line 15: so @awais786 in essence, this PR is about changing this field to IntegerField instead of PositiveIntegerField?

binodpant: just to understand better, it's only an issue for these two tables because the table names + field constraint names combined are too long?

binodpant: I am seeing no reason to reject the PR since it looks like it avoids generating those long constraint by using a regular IntegerField instead of PositiveIntegerField, but looks also like a better longer term fix would be to use shorter model/table names

integrated_channels/blackboard/migrations/0001_initial_squashed_0011_alter_blackboardlearnerassessmentdatatransmissionaudit_enterprise_course_enrollment_id.py
```diff
@@ -0,0 +1,124 @@
...
+def create_uuid(apps, schema_editor):
+    BlackboardEnterpriseCustomerConfiguration = apps.get_model('blackboard', 'BlackboardEnterpriseCustomerConfiguration')
+    for config in BlackboardEnterpriseCustomerConfiguration.objects.all():
+        config.uuid = uuid.uuid4()
+        config.save()
+
+
+class Migration(migrations.Migration):
+
+    replaces = [('blackboard', '0001_initial'), ('blackboard', '0002_auto_20200930_1723'), ('blackboard', '0003_blackboardlearnerdatatransmissionaudit'), ('blackboard', '0004_blackboard_tx_chunk_size_default_1'), ('blackboard', '0005_blackboardlearnerassessmentdatatransmissionaudit'), ('blackboard', '0006_auto_20210708_1446'), ('blackboard', '0007_auto_20210909_1536'), ('blackboard', '0008_auto_20210923_1727'), ('blackboard', '0009_alter_blackboardenterprisecustomerconfiguration_enterprise_customer'), ('blackboard', '0010_auto_20211221_1532'), ('blackboard', '0011_alter_blackboardlearnerassessmentdatatransmissionaudit_enterprise_course_enrollment_id')]
```
binodpant, line 22: not sure what this migration is about?

integrated_channels/blackboard/migrations/0011_alter_blackboardlearnerassessmentdatatransmissionaudit_enterprise_course_enrollment_id.py
binodpant: What I don't understand is why Django generates migrations for the other models that did not change... ?

adzuci (approved):
I think the risk of a slow migration is minimal, but would like to get @nadeemshahzad or @NIXKnight's opinion as well.

integrated_channels/blackboard/migrations/0001_initial_squashed_0011_alter_blackboardlearnerassessmentdatatransmissionaudit_enterprise_course_enrollment_id.py
robrap: This is a squash migration. You can read more about it here: https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations#EverythingAboutDatabaseMigrations-SquashingMigrations. This is how we are able to get the old migration not to run for new deployments, which was causing errors. For older deployments, just the new AlterField migration will be run.

awais786:
@robrap I have a question in case of revert (roll back migration). Due to squash migration roll back is not working locally. Is there any thing require to fix it ? e.g split this PR into two PRs ( 2nd PR will contain squash migrations )

integrated_channels/blackboard/migrations/0011_alter_blackboardlearnerassessmentdatatransmissionaudit_enterprise_course_enrollment_id.py
awais786: - @binodpant On mysql8 when migration runs they generate internal check and due to big table+check name they cross the limit and crashes. 
- We faced this issue only in 2 tables and now [migrations](https://github.com/openedx/edx-platform/runs/4965788270?check_suite_focus=true) are running fine on mysql8.
- changing table is name is risky and time consuming since it has million of records. For new table we will add a check to make sure model name has normal length.
- I have squashed the migrations and generate new one.

nadeemshahzad (approved):

robrap:
@awais786: Good questions.
* If it feels safer for you, you could split the PRs, but I don’t think it is necessary.
* Any deployment will either use the non-squashed migrations (if any of the squashed/replaced migrations have already run), or else it will use the squash migration instead. In our case, we already have some of the squashed migration. So, I think for our deployment, the squash migration for us is just for housekeeping, where Django should mark which migrations were replaced by the squash. This would be used in the future when you delete the old migrations (after nutmeg).
* In your testing of rollback, roll-forward again, does anything seem to go break, or does it just AlterField, restore it, and then alter it again?

awais786:
@robrap Just want to be on safe side to avoid any surprises. I was just testing rollback and I found this issue. In case of `sap_success_factors` app  rollback to `0009` or to `0001_squashed_0022` is not working. It is due to `squashmigrations` because django history table has only 1 entry for this app. Same thing will happen for blackboard. 
When I ran this PR locally migrate command runs only squash migrations
```
Running migrations:
  Applying blackboard.0011_alter_blackboardlearnerassessmentdatatransmissionaudit_enterprise_course_enrollment_id... OK
  Applying sap_success_factors.0009_alter_sapsuccessfactorslearnerdatatransmissionaudit_enterprise_course_enrollment_id... OK
```

Splitting this into two PRs seems safe approach.

integrated_channels/blackboard/migrations/0011_alter_blackboardlearnerassessmentdatatransmissionaudit_enterprise_course_enrollment_id.py
awais786: Removed the squashed migrations those will come in next PR after deployment.

enterprise/__init__.py
```diff
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.39.0"
+__version__ = "3.40.0"
```
iamsobanjaved, line 5: There is a database-related change but that doesn't cause compatibility issues, so minor version change is fine for me.

iamsobanjaved (approved):

integrated_channels/blackboard/migrations/0011_alter_blackboardlearnerassessmentdatatransmissionaudit_enterprise_course_enrollment_id.py
binodpant: Got it, I think we are fine. @awais786  We can also just disable the configs temporarily from django admin, to avoid processing this table, if you feel the need for that. we only have a handful blackboard configs. Depending on timeframe, we can inform @mattdrayer and act accordingly.

binodpant: I meant Blackboard _and_ SAP configs

awais786: My PR is ready we can do this next week @nadeemshahzad.

alex-sheehan-edx (approved):
lgtm

integrated_channels/blackboard/migrations/0011_alter_blackboardlearnerassessmentdatatransmissionaudit_enterprise_course_enrollment_id.py
awais786: @binodpant can you help me in deployment with disabling the configs temporarily from django admin ?

PR #1455, opened 2022-01-31
feat: Add field for Display name for LMS configs

johnnagro:
## Description

- display_name field
- post_save signal to update when blank
- overrideable generator for default display_name
- migrations

## References

- [ENT-5315](https://openedx.atlassian.net/browse/ENT-5315)


kiram15 (approved):
LGTM - Thank you for using blank for Charset instead of null!

PR #1467, opened 2022-02-10
feat: docker-based local work without lms

johnnagro:
## Description

There are sometimes benefits and conveniences to having a containerized local environment versus setting up and maintaining python, virtualenv, and system dependencies (mysql-dev, etc) natively. Sometimes installing your code into the LMS container extends your testing cycle past your patience. Here is some docker/makefile config to enable `make test-shell` to launch a containerized environment containing your local development code so you can `make test` / `make upgrade` / 'profit' / etc inside a container. 

Does not run the web-service nor workers (still need to install your code into an LMS container for that), just wraps local development & dependency management inside docker to make it easier to develop - think of it as dockerized `virtualenv` for your native development system.

## Want to try?

Check this branch out, then run `make test-shell` which should build the `Dockerfile` then drop you into a bash prompt with `dev.txt` dependencies already installed. Then you can run commands such as `make test` or `make upgrade` inside the container.

## References

- [ENT-3934](https://openedx.atlassian.net/browse/ENT-3934)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,91 @@
+FROM ubuntu:focal as app
+MAINTAINER sre@edx.org
+
+
+# Packages installed:
+# git; Used to pull in particular requirements from github rather than pypi,
+# and to check the sha of the code checkout.
+
+# build-essentials; so we can use make with the docker container
+
+# language-pack-en locales; ubuntu locale support so that system utilities have a consistent
+# language and time zone.
+
+# python; ubuntu doesnt ship with python, so this is the python we will use to run the application
+
+# python3-pip; install pip to install application requirements.txt files
+
+# libmysqlclient-dev; to install header files needed to use native C implementation for
+# MySQL-python for performance gains.
+
+# libssl-dev; # mysqlclient wont install without this.
+
+# python3-dev; to install header files for python extensions; much wheel-building depends on this
+
+# gcc; for compiling python extensions distributed with python packages like mysql-client
+
+# If you add a package here please include a comment above describing what it is used for
+RUN apt-get update && apt-get -qy install --no-install-recommends \
+ language-pack-en \
+ locales \
+ python3.8 \
+ python3-pip \
+ python3.8-venv \
+ libmysqlclient-dev \
+ libssl-dev \
+ python3-dev \
+ gcc \
+ build-essential \
+ git
+
+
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -0,0 +1,92 @@
+FROM ubuntu:focal as app
```
christopappas, line 1: hey can you add a comment up here that says that this docker image is only being used for testing as of today's date? that way, there's no confusion later around whether or not this is used in production

docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+version: "2.1"
+services:
+  app:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    container_name: enterprise.test.app
+    hostname: app.test.enterprise
+    volumes:
+      - .:/edx/app/enterprise_catalog/enterprise_catalog
```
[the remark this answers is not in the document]
johnnagro: typo! good catch. i cribbed off how docker/docker-compose work in enterprise-catalog, edx-platform, and enterprise-access.

Makefile
```diff
@@ -171,6 +171,9 @@ pii_clean:
 isort: ## call isort on packages/files that are checked in quality tests
 	isort --skip migrations --recursive tests test_utils enterprise enterprise_learner_portal consent integrated_channels manage.py setup.py
 
+%-shell: ## Run a shell, as root, on the specified service container
```
christopappas, line 174: because there's no actual application running per se, "app" feels a bit misleading to me. maybe something more like `make test-env-shell` or `make test-shell` or something. what do you think?

johnnagro: I had been thinking about that a bit too. I wasn't sure if it was more useful to maintain `make app-shell` since that was familiar to users, or ultimately more confusing since users may expect it to be a real running instance of the project (web service, workers, database, etc).

I like `make test-shell` for now.

adamstankiewicz: +1, while the familiarity of `app-shell` is a plus, it is misleading. Only added feedback here is to maybe add `make test-shell` known somewhere on the README.

johnnagro: I'll add notes to the `local development` parts of `doc/` 👍

adamstankiewicz (approved):
LGTM! Thanks for putting this together 👨‍🍳 💋

PR #1470, opened 2022-02-14
feat: override chunk size default to match channel capability

johnnagro:
## Description

- set new model default where needed
- set fields read-only in /admin
- will follow-up with a settings override too

## References

- [ENT-5347](https://openedx.atlassian.net/browse/ENT-5347)


binodpant (approved):
as long as degreed2 is the only remaining channels where we need this change, lgtm!

binodpant:
also just generally thinking out loud, do we expect to hit any rate limits/throttles with chunk size 1 for these channels? That is not going to block this PR per se, since we already are using chunk size 1 mostly, but we probably still want to think through that state. I believe we ran into throttling with degreed2, and canvas to before, and probably want to drawing board how we want to cleanly handle that or introduce some smartness on how many requests per minute we send out etc (if needed).

johnnagro:
> also just generally thinking out loud, do we expect to hit any rate limits/throttles with chunk size 1 for these channels? That is not going to block this PR per se, since we already are using chunk size 1 mostly, but we probably still want to think through that state. I believe we ran into throttling with degreed2, and canvas to before, and probably want to drawing board how we want to cleanly handle that or introduce some smartness on how many requests per minute we send out etc (if needed).

fwiw these channels are already hard-coded to assume chunk_size of 1 so in theory no change to the current standard operating procedure. we may need to implement some kind of rate limit backoff/retry in the future.

PR #1477, opened 2022-02-17
feat: adding CornerstoneLearnerDataTransmissionAudit admin view

johnnagro:
## Description

- help & fix investigate customer issues

## References

- [ENT-5463](https://openedx.atlassian.net/browse/ENT-5463)


binodpant (approved):
lgtm if tested ok locally!

binodpant:
Think you will also need to add the app-permissions for these models in edx-internal once this is in.. to see in staging/prod etc?

johnnagro:
> Think you will also need to add the app-permissions for these models in edx-internal once this is in.. to see in staging/prod etc?

Yes. I'll send an additional PR.

PR #1480, opened 2022-02-17
feat: Add duration to the schedule of payload for SAP content metadata exporter

binodpant:
Based on reading this doc, it seems the 'schedule' field is what we want
https://help.sap.com/viewer/9d4c9e0d04304afdbe8f1b4480d71403/2111/en-US/2be70bfe31e64399be709de794f639ee.html

### JIRA
https://openedx.atlassian.net/browse/ENT-5367

This PR includes AC no. 2 of that ticket

### What this change does
This change populates the schedule array from one of these:

 - the closes course run if we are processing a course itme
 - the course run items itself

it extracts the values to produce schedule as follows:

```
[{
            'startDate': parse_datetime_to_epoch_millis(start) if start else '',
            'endDate': parse_datetime_to_epoch_millis(end) if end else '',
            'active': current_time_is_in_interval(start, end) if start else False,
            'duration': f"{duration} days",
}
]
```

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -1141,6 +1141,7 @@ def get_course_run_duration_info(course_run):
         )
     return duration_info
 
+
 def get_duration_of_course_or_courserun(content_metadata_item):
     """
     Returns duration start, end dates given a piece of content_metadata item
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -1141,6 +1141,35 @@ def get_course_run_duration_info(course_run):
         )
     return duration_info
 
+def get_duration_of_course_or_courserun(content_metadata_item):
+    """
+    Returns duration start, end dates given a piece of content_metadata item
+    If course item, extracts start, end dates of closest course run based on current timestamp
+    Returns:
+        duration: in days or 0
+        start: start field of closest course run item, or None
+        end: end field of closest course run item, or None
+    """
+    start = None
+    end = None
+    if content_metadata_item.get('content_type') == 'courserun':
+        start = content_metadata_item.get('start')
+        end = content_metadata_item.get('end')
+    elif content_metadata_item.get('content_type') == 'course':
+        course_runs = content_metadata_item.get('course_runs')
+        if course_runs:
+            course_run = get_closest_course_run(course_runs)
+            if course_run:
+                start = course_run.get('start')
+                end = course_run.get('end')
+    if not start:
+        return 0, None, None
+    start_date = parse_datetime_handle_invalid(start)
+    end_date = parse_datetime_handle_invalid(end)
+    if not start_date or not end_date:
+        return 0, None, None
+    return (end_date - start_date).days, start, end
+
 
 def is_course_run_enrollable(course_run):
     """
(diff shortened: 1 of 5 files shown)


enterprise/utils.py
```diff
@@ -1142,6 +1142,36 @@ def get_course_run_duration_info(course_run):
...
+        end: end field of closest course run item, or None
+    """
+    start = None
+    end = None
+    if content_metadata_item.get('content_type') == 'courserun':
+        start = content_metadata_item.get('start')
+        end = content_metadata_item.get('end')
+    elif content_metadata_item.get('content_type') == 'course':
+        course_runs = content_metadata_item.get('course_runs')
+        if course_runs:
```
alex-sheehan-edx, line 1161: when would we see no course runs in the metadata?

integrated_channels/sap_success_factors/exporters/content_metadata.py
```diff
@@ -133,11 +138,19 @@ def transform_revision_number(self, content_metadata_item):  # pylint: disable=u
         """
         return 1
 
-    def transform_schedule(self, content_metadata_item):  # pylint: disable=unused-argument
+    def transform_schedule(self, content_metadata_item):
         """
         Return the schedule of the content item.
         """
-        return []
+
```
alex-sheehan-edx, line 145: tiny nit but do you mind removing the empty space after the doc string?

integrated_channels/sap_success_factors/exporters/content_metadata.py
```diff
@@ -133,11 +138,19 @@ def transform_revision_number(self, content_metadata_item):  # pylint: disable=u
         """
         return 1
 
-    def transform_schedule(self, content_metadata_item):  # pylint: disable=unused-argument
+    def transform_schedule(self, content_metadata_item):
         """
         Return the schedule of the content item.
         """
-        return []
```
alex-sheehan-edx, line 140: did we just assume we couldn't export schedule before? Wish we had a comment explaining the empty array return value

alex-sheehan-edx (approved):
A couple of questions and one small nit request but otherwise looks good

integrated_channels/sap_success_factors/exporters/content_metadata.py
binodpant: yeah I can't find anything like that. The SAP docs as of today do say they prefer schedule if given, otherwise, that 'duration' is used. Since we are only sending schedule anyway I did not worry about duration fwiw.

enterprise/utils.py
binodpant: I actually don't know if this happens in real life, but it can happen and is just defensive programming. Also noted this pattern existed in some other code in integ channels and/or utils.

binodpant: also fwiw: this pattern of 'find a course run if it exists, and extract some field(s) out of it' also seems fairly repeated so perhaps there is an opportunity for us to try DRY too... just did not want to go overboard in this PR

integrated_channels/sap_success_factors/exporters/content_metadata.py
binodpant: in this case I figured it would be better to always send the same shape in this field regardless of if we found a course run or not. I am presuming that is better than not sending anything at all (like rn)

binodpant: it is honestly a bit hard to know what the best answer if for SAP since we don't hav a test environment like the others, so fo this one we will have to transmit, then maybe get @mattdrayer 's help in seeing what it looks like on customers' side

PR #1493, opened 2022-03-10
fix: CornerstoneLearnerDataTransmissionAudit admin view timeout

johnnagro:
## Description

I had attempted to pull in associated objects (non-foreign keys) for some added info on this admin screen but it seems to be breaking (timing out) in prod (works locally). I want to use this view, i don't need this extra info. I wanna dump it for now.

## References

- [ENT-5565](https://openedx.atlassian.net/browse/ENT-5565)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * nothing
 
+[3.40.14]
+---------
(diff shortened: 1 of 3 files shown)


binodpant (approved):
interesting but ok let’s get through this one and we can revisit

PR #1495, opened 2022-03-14
fix: CSOD Learner Audit Django Admin Timeouts

johnnagro:
## Description

- timeout is likely due to django admin trying to create/populate a select drop-down for the user field
- mark user as a 'raw id field'
- add back the enterprise name read-only since it wasnt the issue

## References

- [ENT-5565](https://openedx.atlassian.net/browse/ENT-5565)

### DIFF
--- a/integrated_channels/cornerstone/admin/__init__.py
+++ b/integrated_channels/cornerstone/admin/__init__.py
@@ -76,6 +76,9 @@ class CornerstoneLearnerDataTransmissionAuditAdmin(admin.ModelAdmin):
         "course_id",
         "status",
     )
+    raw_id_fields = (
+        'user',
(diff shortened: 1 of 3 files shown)


integrated_channels/cornerstone/admin/__init__.py
```diff
@@ -77,6 +99,14 @@ class CornerstoneLearnerDataTransmissionAuditAdmin(admin.ModelAdmin):
         "status",
     )
 
+    raw_id_fields = (
```
johnnagro, line 102: this is the part that should fix the weird timeout

alex-sheehan-edx (approved):
lgtm

PR #1501, opened 2022-03-17
fix: chunk cornerstone content updates

johnnagro:
## Description

A customer cannot setup their CSOD integration because the initial pull of their content by CSOD times out. This specific customer has multiple "all-content" catalogs but in theory even smaller catalogs have the potential to timeout since we do a full send every time. 

This set of changes is to add a configurable "max changes to send" size (defaulting to 1000). Whenever CSOD polls, we'll doll out the next N changes based on what our catalog diff tells us. Prioritizing creates over updates. If we have more than N changes to send, the audit records for only those first N will get updated and thus the remaining changes will be sent on the next poll by CSOD.

I tried to implement this will the least number of architectural changes possible. Also tried to keep in mind how/where the proper support for the `If-Modified-Since` header changes would be made.

- set max chunk size to send in one request
- 'correct' modified times when if-modified-since
- add testing, comments

## Considerations

- the request marks the records as transmitted, we'll need to avoid using the endpoint for debugging
- we're about to implement proper support of `If-Modified-Since` which may improve on how we track what needs to be transmitted/re-transmitted

## References
- [ENT-5521](https://openedx.atlassian.net/browse/ENT-5521)

### DIFF
--- a/enterprise/api_client/enterprise_catalog.py
+++ b/enterprise/api_client/enterprise_catalog.py
@@ -111,25 +111,10 @@ class EnterpriseCatalogApiClient(JwtLmsApiClient):
             content_keys (list): List of string content keys
             should_raise_exception (Bool): Optional param for whether or not api response exceptions should be raised.
 
-        Response:
-            items_to_create: [{
-               "course_key": <content key>>
-            }, {
-            ..
-            }]
-
-            items_to_delete: [{
-               "course_key": <content key>>
-            }, {
-            ..
-            }]
-
-            items_found: [{
-                "course_key": <content key>>,
-                "date_updated": <content last modified datetime>
-            }, {
-                ...
-            }]
+        Returns:
+            items_to_create (list): dictionaries of content_keys to create
+            items_to_delete (list): dictionaries of content_keys to delete
+            items_found (list): dictionaries of content_keys and date_updated datetimes of content to update
         """
         catalog_uuid = enterprise_customer_catalog.uuid
         endpoint = getattr(self.client, self.CATALOG_DIFF_ENDPOINT.format(catalog_uuid))
--- a/integrated_channels/cornerstone/urls.py
+++ b/integrated_channels/cornerstone/urls.py
@@ -4,13 +4,10 @@ URL definitions for Cornerstone API.
 
 from django.urls import path
 
-from integrated_channels.cornerstone.views import CornerstoneCoursesListView, CornerstoneCoursesUpdates
+from integrated_channels.cornerstone.views import CornerstoneCoursesListView
 
 urlpatterns = [
     path('course-list', CornerstoneCoursesListView.as_view(),
          name='cornerstone-course-list'
-         ),
-    path('course-updates', CornerstoneCoursesUpdates.as_view(),
-         name='cornerstone-course-updates'
          )
 ]
--- a/integrated_channels/cornerstone/views.py
+++ b/integrated_channels/cornerstone/views.py
@@ -2,16 +2,21 @@
 Views containing APIs for cornerstone integrated channel
 """
 
+import datetime
 from logging import getLogger
 
+from dateutil import parser
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
 from rest_framework import generics, permissions, renderers, status
 from rest_framework.authentication import SessionAuthentication
 from rest_framework.response import Response
 
+from django.utils.http import parse_http_date_safe
+
 from enterprise.api.throttles import ServiceUserThrottle
 from enterprise.utils import get_enterprise_customer, get_enterprise_worker_user, get_oauth2authentication_class
 from integrated_channels.cornerstone.models import CornerstoneEnterpriseCustomerConfiguration
+from integrated_channels.integrated_channel.constants import ISO_8601_DATE_FORMAT
 
 logger = getLogger(__name__)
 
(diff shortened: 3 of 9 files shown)


integrated_channels/cornerstone/views.py
```diff
@@ -209,20 +124,32 @@ def get(self, request, *args, **kwargs):
         )
         exporter = enterprise_config.get_content_metadata_exporter(worker_user)
         transmitter = enterprise_config.get_content_metadata_transmitter()
-        data = transmitter.transmit(*exporter.export_force_all_catalogs())
+
+        get_count = int(request.GET.get('count', exporter.MAX_PAYLOAD_COUNT))
```
johnnagro, line 128: you can pass a count in via a get param, cannot exceed the settings though

integrated_channels/cornerstone/views.py
```diff
@@ -209,20 +124,32 @@ def get(self, request, *args, **kwargs):
...
+                this_last_modified = parser.parse(item['LastModifiedUTC'])
+                if if_modified_since > int(this_last_modified.timestamp()):
+                    # log when we find an incorrect modified date, then "fix" it
+                    logger.info(f'integrated_channel=CSOD, '
+                                f'integrated_channel_enterprise_customer_uuid={enterprise_customer_uuid}, '
+                                f'If-Modified-Since={request.headers.get("If-Modified-Since")}, '
+                                f'integrated_channel_course_key={item["ID"]}, '
+                                f'LastModifiedUTC={item["LastModifiedUTC"]} '
+                                f'fixing modified/header mismatch')
+                    if_modified_since_dt = datetime.datetime.fromtimestamp(if_modified_since)
```
johnnagro, line 152: we track when we need to update content with the audit records, the if-modified-since can be faked a bit until we have full support with the soft-deletes.

integrated_channels/cornerstone/views.py
```diff
@@ -32,96 +37,6 @@ class BaseViewSet(generics.ListAPIView):
     renderer_classes = [renderers.JSONRenderer]
 
 
-class CornerstoneCoursesUpdates(BaseViewSet):
```
johnnagro, line 35: originally, we were going to use a new endpoint for support of the `If-Modified-Since` header. it was never used and I plan to merge that functionality into the existing endpoint instead. removing this code for now.

integrated_channels/cornerstone/urls.py
```diff
@@ -4,13 +4,10 @@
...
 from django.urls import path
 
-from integrated_channels.cornerstone.views import CornerstoneCoursesListView, CornerstoneCoursesUpdates
+from integrated_channels.cornerstone.views import CornerstoneCoursesListView
 
 urlpatterns = [
     path('course-list', CornerstoneCoursesListView.as_view(),
          name='cornerstone-course-list'
-         ),
-    path('course-updates', CornerstoneCoursesUpdates.as_view(),
```
johnnagro, line 13: unused class/route/path

test_utils/fake_catalog_api.py
```diff
@@ -1342,6 +1343,18 @@ def get_fake_catalog_diff_create_w_program():
     ], [], []
 
 
+def get_fake_catalog_diff_w_one_each():
+    """
+    Returns a fake response from the EnterpriseCatalogApiClient.get_catalog_diff where two courses and a program are
+    returned, one in each change-type bucket
+    """
+    # items_to_create, items_to_delete, matched_items
```
alex-sheehan-edx, line 1351: cruft?

tests/test_integrated_channels/test_cornerstone/test_views.py
```diff
@@ -47,14 +51,6 @@ def test_course_list_unauthorized_non_customer(self):
         response = self.client.get(self.course_list_url)
         self.assertEqual(response.status_code, 401)
 
-    def test_course_update_unauthorized_non_customer(self):
```
alex-sheehan-edx, line 50: why was this removed?

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -166,6 +167,20 @@ def _get_catalog_diff(self, enterprise_catalog, content_keys, force_retrieve_all
...
         )
+
+        # if we have more to work with than the allowed space, slice it up
+        if len(items_to_create) + len(items_to_delete) + len(matched_items) > max_item_count:
+            # prioritize creates, then updates, then deletes
+            items_to_create = items_to_create[0:max_item_count]
+            count_left = max_item_count - len(items_to_create)
+            matched_items = matched_items[0:count_left]
+            count_left = count_left - len(matched_items)
+            # in testing, this is sometimes a list, sometimes a dict
```
alex-sheehan-edx, line 178: why is this inconsistent?

alex-sheehan-edx (commented):
couple of questions/comments but otherwise looking good

tests/test_integrated_channels/test_cornerstone/test_views.py
johnnagro: this is for the update version of the endpoint which we're removing in this PR as well. there is a version of this test for the list endpoint - the one in use - immediately above this.

integrated_channels/integrated_channel/exporters/content_metadata.py
johnnagro: its inconsistent in the catalog api - when items to delete is empty, it returns as a dictionary, when its got data, it returns as a list. i don't fully understand why/the history but fixing the catalog api and looking for any downstream consumer that also needed updating i felt was outside the scope of this PR - especially since we're going to probably work on that endpoint to properly support if-modified-since

integrated_channels/cornerstone/views.py
johnnagro: yeah totes. i thought 2 lines was easier to read/comprehend.

test_utils/fake_catalog_api.py
johnnagro: it was in other similar methods, but i can clean it up

enterprise/api_client/enterprise_catalog.py
```diff
@@ -111,25 +111,10 @@ def get_catalog_diff(self, enterprise_customer_catalog, content_keys, should_rai
...
-            ..
-            }]
-
-            items_found: [{
-                "course_key": <content key>>,
-                "date_updated": <content last modified datetime>
-            }, {
-                ...
-            }]
+        Returns:
```
johnnagro, line 114: this was picked up in a master rebase - ignore.

PR #1502, opened 2022-03-18
fix: edx-enterprise readthedocs build broken

johnnagro:
- fixed docstring warnings

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,7 +8,7 @@ version: 2
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: docs/conf.py
-  fail_on_warning: false
+  fail_on_warning: true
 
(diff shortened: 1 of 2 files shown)


binodpant:
ok thanks this will unblock us then. Yeah it is weird that edx-enterprise builds fail on warnings from edx-platform docs ? Not clear why/how that is... Maybe Ned can help us navigate that... it appears the docs are published correctly now thought! 

https://open-edx-enterprise-service-documentation--1502.org.readthedocs.build/en/1502/

so looks good to me

binodpant (approved):

binodpant:
just noted these warnings: 
https://github.com/openedx/edx-enterprise/runs/5601918760?check_suite_focus=true#step:5:95

perhaps these can be fixed easily?

PR #1503, opened 2022-03-21
feat: setting customer identity provider config is_valid on first SSO login

alex-sheehan-edx:
Jira ticket: https://openedx.atlassian.net/browse/ENT-5484

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/tpa_pipeline.py
+++ b/enterprise/tpa_pipeline.py
@@ -3,6 +3,7 @@ Module provides elements to be used in third-party auth pipeline.
 """
 
 import re
+from datetime import datetime
 from logging import getLogger
 
 from django.urls import reverse
@@ -28,14 +29,21 @@ except ImportError:
 LOGGER = getLogger(__name__)
 
 
-def get_enterprise_customer_for_running_pipeline(request, pipeline):
+def get_sso_provider(request, pipeline):
     """
-    Get the EnterpriseCustomer associated with a running pipeline.
+    Helper method to retrieve the sso provider ID from either a user's SSO login request
     """
     sso_provider_id = request.GET.get('tpa_hint')
     if pipeline:
         sso_provider_id = Registry.get_from_pipeline(pipeline).provider_id
-    return get_enterprise_customer_for_sso(sso_provider_id)
+    return sso_provider_id
+
+
+def get_enterprise_customer_for_running_pipeline(request, pipeline):
+    """
+    Get the EnterpriseCustomer associated with a running pipeline.
+    """
+    return get_enterprise_customer_for_sso(get_sso_provider(request, pipeline))
 
 
 def get_enterprise_customer_for_sso(sso_provider_id):
@@ -48,6 +56,17 @@ def get_enterprise_customer_for_sso(sso_provider_id):
         )
     except EnterpriseCustomer.DoesNotExist:
         return None
+
+
+def validate_provider_config(enterprise_customer, sso_provider_id):
+    """
+    Helper method to ensure that a customer's provider config is validated
+    """
+    # With a successful SSO login, validate the enterprise customer's IDP config if it hasn't already been validated
+    enterprise_provider_config = enterprise_customer.identity_providers.filter(provider_id=sso_provider_id).first()
+    if not enterprise_provider_config.identity_provider.was_valid_at:
+        enterprise_provider_config.identity_provider.was_valid_at = datetime.now()
+        enterprise_provider_config.identity_provider.save()
 
 
 @partial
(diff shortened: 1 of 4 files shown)


enterprise/tpa_pipeline.py
```diff
@@ -79,6 +79,10 @@ def handle_enterprise_logistration(backend, user, **kwargs):
             handle_redirect_after_social_auth_login(backend, user)
         return
 
+    if not enterprise_customer.identity_providers.first().identity_provider.is_valid:
```
johnnagro, line 82: any chance `enterprise_customer.identity_providers` could be empty and thus `.first()` would blow up? Its always the first provider?

Also, we renamed this field `was_valid_at`

enterprise/tpa_pipeline.py
```diff
@@ -79,6 +79,10 @@ def handle_enterprise_logistration(backend, user, **kwargs):
...
+        enterprise_customer.identity_providers.first().identity_provider.is_valid = datetime.now()
```
johnnagro, line 83: we renamed this field `was_valid_at`

enterprise/tpa_pipeline.py
binodpant: We may be adding axsecond sso config for example so I think we won’t want to assume first instead if possiblecwe want to use the exact  config we are working with

alex-sheehan-edx: I believe that this `handle_enterprise_logistration` method is only hit when we have an SSO interaction, so the existence of an identity provider should be guaranteed

binodpant: agreed with that, but what I meant is what if there are multiple items in identity_providers? should we not update the correct one? I have not checked, but is it possible to know which provider config we are working with, in this part of the flow ? I can try to look that up

binodpant: I think you just want to use that provider config, based on this id, rather than first()

binodpant:
coming up great! it's a high traffic codebase so I recommend updating/adding a test, perhaps you can just add to one or more of the existing test cases

enterprise/tpa_pipeline.py
alex-sheehan-edx: I've changed the code to filter for the provider id 👍

johnnagro (approved):

enterprise/tpa_pipeline.py
```diff
@@ -63,13 +64,13 @@ def handle_enterprise_logistration(backend, user, **kwargs):
...
     LOGGER.info(f'Beginning enterprise logistration for LMS user {user.id}')
     request = backend.strategy.request
-    enterprise_customer = get_enterprise_customer_for_running_pipeline(
-        request,
-        {
-            'backend': backend.name,
-            'kwargs': kwargs
-        }
-    )
+    pipeline = {'backend': backend.name, 'kwargs': kwargs}
```
binodpant, line 67: this should work too I think it would be a slight bit cleaner if we can reuse the provider_id logic between this method and the get_enterprise_customer_for_running_pipeline method in case we change logic in the future, then we have to update in both places


ideally can we do sth like this?

```
def get_enterprise_customer_for_running_pipeline(request, pipeline):
    """
    Get the EnterpriseCustomer associated with a running pipeline.
    """
    sso_provider_id = resolve_provider_id(request, pipeline)
    return get_enterprise_customer_for_sso(sso_provider_id)
```

then in your code you can also just call

```
 sso_provider_id = resolve_provider_id(request, pipeline)
```

enterprise/tpa_pipeline.py
```diff
@@ -79,6 +80,12 @@ def handle_enterprise_logistration(backend, user, **kwargs):
             handle_redirect_after_social_auth_login(backend, user)
         return
 
+    # With a successful SSO login, validate the enterprise customer's IDP config if it hasn't already been validated
+    enterprise_provider_config = enterprise_customer.identity_providers.filter(provider_id=sso_provider_id).first()
+    if not enterprise_provider_config.identity_provider.was_valid_at:
```
binodpant, line 85: hmm maybe I am just not reading the django right, but should these references be `enterprise_provider_config` or `enterprise_provider_config.identity_provider` ?

alex-sheehan-edx: the enterprise provider config is really just a reference to the provider config from platform. The validation lives in that provider config which we reference through the enterprise provider config

binodpant: ah ok maybe I just did not get the correct syntax etc, if this works locally we are all set!

PR #1508, opened 2022-03-24
fix: add foreign keys to integrated channels audit models

johnnagro:
## Description

Customers who have more than one integration of the same kind will run into disambiguation issues with the audit records.

- add more foreign keys to tie the audits to specific integration configs to help the sync know which audits belong to which config of the same kind
- enterprise customer uuid is a natural partition key which should be on more of our records
- bonus content: clean up LearnerAudit inheritance
- create generic implementations of meta classes for testing + base implementation of logic

## ToDo (probably in a second follow-on PR after this deploys)

- [ ] data migration
- [ ] audit record queries should include config id

## References

- https://openedx.atlassian.net/browse/ENT-5595
- https://openedx.atlassian.net/browse/ENT-5603
- https://openedx.atlassian.net/browse/ENT-5605


integrated_channels/cornerstone/migrations/0018_auto_20220325_1757.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.2.12 on 2022-03-25 17:57
```
alex-sheehan-edx, line 1: same q about separate migrations

integrated_channels/cornerstone/models.py
```diff
@@ -187,11 +188,13 @@ def get_learner_data_exporter(self, user):
...
 
 
-class CornerstoneLearnerDataTransmissionAudit(TimeStampedModel):
+class CornerstoneLearnerDataTransmissionAudit(LearnerDataTransmissionAudit):
     """
     The payload we sent to Cornerstone at a given point in time for an enterprise course enrollment.
 
     """
+
+    # TODO how is this set/used? should it be on the abstract class?
```
alex-sheehan-edx, line 197: my only guess would be this is for individual transmissions where we recieve a signal from a specific user's completion. How it get's set is a mystery to me  ¯\_(ツ)_/¯

alex-sheehan-edx (approved):
a couple of q/comments but otherwise looking good

integrated_channels/integrated_channel/models.py
```diff
@@ -244,26 +244,49 @@ def update_content_transmission_catalog(self, user):
         exporter.update_content_transmissions_catalog_uuids()
 
 
-class LearnerDataTransmissionAudit(models.Model):
+class GenericEnterpriseCustomerPluginConfiguration(EnterpriseCustomerPluginConfiguration):
```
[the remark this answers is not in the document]
binodpant: will this result a db table too? i think we should maybe even remove the table for the base class? seems like it’s an empty table rn

integrated_channels/integrated_channel/models.py
```diff
@@ -328,6 +379,7 @@ class ContentMetadataItemTransmission(TimeStampedModel):
 
     enterprise_customer = models.ForeignKey(EnterpriseCustomer, on_delete=models.CASCADE)
     integrated_channel_code = models.CharField(max_length=30)
+    plugin_configuration_id = models.PositiveIntegerField(blank=True, null=True)
```
binodpant, line 382: I think we ran some issues before with biginteger firld and index name length?

binodpant: ok this one it looks like https://edx-internal.slack.com/archives/C014MDBL8JW/p1643827874773499?thread_ts=1643826987.458379&cid=C014MDBL8JW

johnnagro: I found this field type used elsewhere for a non-foreign-key id join column. i can take a peek to see if we've got any bugs/docs on this...

PR #1513, opened 2022-03-30
feat: mysql8 is generating internal checks if model has `PositiveInteger` field 

awais786:
mysql8 is generating internal checks if model has `PositiveIntegerField`.

These check has random names with 64 characters limit. But due to long model names few checks breaks the limit and mysql8 throws error.


integrated_channels/sap_success_factors/migrations/0004_auto_20220324_1550.py
```diff
@@ -25,7 +25,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='sapsuccessfactorslearnerdatatransmissionaudit',
             name='plugin_configuration_id',
-            field=models.PositiveIntegerField(blank=True, null=True),
+            field=models.IntegerField(blank=True, null=True),
```
awais786, line 28: Don't need this in this commit. We need this in other squash PR.

johnnagro (approved):

PR #1515, opened 2022-03-30
fix: squash SAP migrations

johnnagro:
**Merge checklist:**

**Post merge:**


integrated_channels/sap_success_factors/migrations/0004_auto_20220324_1550_squashed_0006_auto_20220330_1157.py
```diff
@@ -0,0 +1,67 @@
...
+        migrations.AddField(
+            model_name='sapsuccessfactorslearnerdatatransmissionaudit',
+            name='subsection_name',
+            field=models.CharField(max_length=255, null=True),
+        ),
+        migrations.AlterField(
+            model_name='sapsuccessfactorslearnerdatatransmissionaudit',
+            name='enterprise_course_enrollment_id',
+            field=models.IntegerField(blank=True, db_index=True, null=True),
+        ),
```
robrap, line 61: I think this is a repeat of the earlier AlterField, which may be from your manual change? You only need one if they are the same, and I'm wondering if this was a new field in these migrations or not.

iamsobanjaved: This wasn't a new field, it was previously there which we fixed earlier but was removed here so `AlterField` appeared in `0004` migration

https://github.com/openedx/edx-enterprise/commit/b0e418101bfb1b0bbc5188b517adfa58547dbf11#diff-6f4694ec4c254f702264a1cd1e0c2bc82e8914e39d4ec9778a743f47bd00a4c2L277

PR #1516, opened 2022-03-30
feat: management command to backfill missing fks

johnnagro:
## Description

- backfill missing fks on audit records
- data migration as mgmt command rather than traditional migration
- lots of data to change, no schema changes needed
- requires config of a manual jenkins task 

## Testing

- tested locally by hand
- manually test on stage

## References

- [ENT-5605](https://openedx.atlassian.net/browse/ENT-5605)


alex-sheehan-edx (approved):
Not entirely sure why we fetch api user and require it as a param to the command but I know it's done elsewhere so LGTM  ¯\_(ツ)_/¯

PR #1519, opened 2022-04-04
ENT-5632 completions fix for non audit to base on passed_timestamp

binodpant:
Change of logic is basically:

- if audit track: no change of behavior. Still send completions as long as incomplete_count = 0
- if non audit track: only send completion if passed_timestamp is not Null and is_passed=True. This matches with how the analytics tools report completions. This should also fix the bug where we incorrectly sent non audit tracks as completed, solely based on this other completed_date that we calculate, that does _not_ match with analytics logic. We may be able to remove that field in the future but for now I left it there, I just don't use it to calculate is_completed anymore :)
- specifically call CourseGradeFactory with create_if_needed=False, since I don't see why we need to use the get_or_create pattern here, we are simply wanting to read existing grade records

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/exporters/learner_data.py
+++ b/integrated_channels/integrated_channel/exporters/learner_data.py
@@ -299,6 +299,7 @@ class LearnerExporter(ChannelSettingsMixin, Exporter):
         enterprise_customer_uuid = enterprise_enrollment.enterprise_customer_user.enterprise_customer.uuid
         course_id = enterprise_enrollment.course_id
 
+        # breakpoint()
         if is_audit_enrollment:
             completed_date_from_api, grade_from_api, is_passing_from_api, grade_percent, passed_timestamp = \
                 self.collect_grades_data(enterprise_enrollment, course_details, channel_name)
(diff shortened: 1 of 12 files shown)


integrated_channels/integrated_channel/exporters/learner_data.py
```diff
@@ -291,38 +292,47 @@ def get_grades_summary(
...
         '''
         is_audit_enrollment = enterprise_enrollment.is_audit_enrollment
         lms_user_id = enterprise_enrollment.enterprise_customer_user.user_id
         enterprise_customer_uuid = enterprise_enrollment.enterprise_customer_user.enterprise_customer.uuid
         course_id = enterprise_enrollment.course_id
 
-        # For instructor-paced and non-audit courses, let the certificate determine course completion
-        if course_details.pacing == 'instructor' and not is_audit_enrollment:
-            completed_date_from_api, grade_from_api, is_passing_from_api, grade_percent = \
+        # breakpoint()
```
alex-sheehan-edx, line 302: cruft?

alex-sheehan-edx (approved):
one nit but lgtm

johnnagro (approved):

PR #1528, opened 2022-04-11
fix: override get_assignments() so that the order matters

iloveagent57:
Overrides the [SystemWideEnterpriseUserRoleAssignment.get_assignments()](https://github.com/openedx/edx-enterprise/blob/9753181f1efee616affac216efe56eec9dc2a9b6/enterprise/models.py#L2802) method to return list of (role, context) assignments where the first item in the list corresponds to the currently active enterprise for the user.

This approach should allow us to remedy the problem without changing behavior of the consumers of JWT-stored roles (e.g. in ecommerce).

https://openedx.atlassian.net/browse/ENT-5700

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -3,6 +3,7 @@ Database models for enterprise.
 """
 
 import collections
+import itertools
 import json
 import os
 from decimal import Decimal
@@ -1268,11 +1269,27 @@ class EnterpriseCustomerUser(TimeStampedModel):
     @classmethod
     def inactivate_other_customers(cls, user_id, enterprise_customer):
         """
-        Inactive all the enterprise customers of given user except the given enterprise_customer.
+        Mark as inactive all the enterprise customers of given user except the given enterprise_customer.
         """
         EnterpriseCustomerUser.objects.filter(
             user_id=user_id
         ).exclude(enterprise_customer=enterprise_customer).update(active=False)
+
+    @classmethod
+    def get_active_enterprise_users(cls, user_id, enterprise_customer_uuids=None):
+        """
+        Return a queryset of all active enterprise users to which the given user is related.
+        Or, if ``enterprise_customer_uuids`` is non-null, only the enterprise users
+        related to the list of given ``enterprise_customer_uuids``.
+        """
+        kwargs = {
+            'user_id': user_id,
+            'active': True,
+        }
+        if enterprise_customer_uuids:
+            kwargs['enterprise_customer__in'] = enterprise_customer_uuids
+
+        return EnterpriseCustomerUser.objects.filter(**kwargs)
 
 
 class PendingEnterpriseCustomerUser(TimeStampedModel):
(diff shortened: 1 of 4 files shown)


enterprise/models.py
```diff
@@ -2870,6 +2889,69 @@ def get_context(self):
...
+        Return an iterator of (rolename, enterprise customer uuid) for the given
+        user (and maybe role_names).
+
+        Differs from super().get_assignments(...) in that it yields the (role name, customer uuid)
+        pairs that correspond to the currently *active* EnterpriseCustomerUser for the given role first.
+        The resulting generated pairs are sorted by role name, and within role_name, by (active?, customer uuid).
+
+        ('enterprise_admin', ['active-enterpirse-uuid', 'other-active-enterpirse-uuid', 'inactive-enterpirse-uuid'])
+        ('enterprise_learner', ['active-enterpirse-uuid', 'inactive-enterpirse-uuid']),
+        ('enterprise_openedx_operator', ['*'])
```
iloveagent57, line 2917: Sorry, this is inconsistent in terms of how it's talking about the yielded items - I'll clean it up.

enterprise/models.py
```diff
@@ -821,13 +822,15 @@ def get_link_by_email(self, user_email, enterprise_customer):
...
-    def link_user(self, enterprise_customer, user_email):
+    def link_user(self, enterprise_customer, user_email, **kwargs):
         """
         Link user email to Enterprise Customer.
 
         If :class:`django.contrib.auth.models.User` instance with specified email does not exist,
         :class:`.PendingEnterpriseCustomerUser` instance is created instead.
         """
+        defaults = {'active': True}
+        defaults.update(kwargs)
```
iloveagent57, line 833: This was a change to support testing that I'm now realizing is no longer needed, will delete.

enterprise/models.py
```diff
@@ -2870,6 +2887,72 @@ def get_context(self):
...
+        user (and maybe role_names).
+
+        Differs from super().get_assignments(...) in that it yields (role name, customer uuid list) pairs
+        such that the first item in the customer uuid list for each role
+        corresponds to the currently *active* EnterpriseCustomerUser for the user.
+
+        The resulting generated pairs are sorted by role name, and within role_name, by (active, customer uuid).
+        For example:
+
+          ('enterprise_admin', ['active-enterpirse-uuid', 'inactive-enterpirse-uuid', 'other-inactive-enterpirse-uuid'])
```
adamstankiewicz, line 2916: nit: typos in "enterpirse"

tests/test_models.py
```diff
@@ -2287,6 +2299,55 @@ def test_get_context_explicit_enterprise_customer(self, role_name, expected_cont
...
+            role=roles_api.openedx_operator_role(),
+            applies_to_all_contexts=True,
+        )
+
+        expected_assignments = [
+            (
+                ENTERPRISE_ADMIN_ROLE, [
+                    str(delta_customer.uuid), str(alpha_customer.uuid),
+                ],
+            ),
```
adamstankiewicz, line 2339: [clarification] Is the `ENTERPRISE_ADMIN_ROLE` first here because it happens to come before `ENTERPRISE_LEARNER_ROLE` alphabetically? If we by chance have another role in the future that comes before `ENTERPRISE_ADMIN_ROLE` alphabetically, would it then come first? I'm wondering if the sort order of the roles should be more deterministic (e.g., consumers might expect active, `enterprise_learner` roles to always come first)?

enterprise/models.py
```diff
@@ -1268,12 +1269,28 @@ def create_order_for_enrollment(self, course_run_id, discount_percentage, mode,
     @classmethod
     def inactivate_other_customers(cls, user_id, enterprise_customer):
         """
-        Inactive all the enterprise customers of given user except the given enterprise_customer.
+        Mark as active all the enterprise customers of given user except the given enterprise_customer.
```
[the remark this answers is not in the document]
iloveagent57: Hah, yes!

PR #1536, opened 2022-04-14
fix: Revert "Revert "fix: override get_assignments() so that active enterprise uuids come first.""

iloveagent57:
Ensure that EnterpriseSystemWideUserRoleAssignment.get_assignments() can handle null values that might be returned by get_context().  See https://openedx.atlassian.net/browse/ENT-4346 for open technical debt and partial explanation of how `None` is a meaningful context.

This reverts commit 6927277aad53f2497a336730e8af52372ee316cf.

https://openedx.atlassian.net/browse/ENT-5700

Screenshots from manually testing the specific null-pointer defect:
Create a new user with no enterprise link, but _with_ an assignment record where the enterprise id is null and they don't have wildcard access:
image

Login and verify that there are no errors, and the user's JWT should have `roles` with a role name but no context:
image

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -3,6 +3,7 @@ Database models for enterprise.
 """
 
 import collections
+import itertools
 import json
 import os
 from decimal import Decimal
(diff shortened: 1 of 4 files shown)


enterprise/models.py
```diff
@@ -2789,13 +2806,14 @@ def enterprise_customer_uuids(self):
                     self.role.name,
                     self.id
                 ))
-            return None
+            return []
```
iloveagent57, line 2809: This is the change that stops us from hitting a TypeError/None is not iterable.

enterprise/models.py
```diff
@@ -2870,6 +2888,72 @@ def get_context(self):
...
+    def get_distinct_assignments_by_role_name(cls, user, role_names=None):
+        """
+        Returns a mapping of role names to sets of enterprise customer uuids
+        for which the user is assigned that role.
+        """
+        # super().get_assignments() returns pairs of (role name, contexts), where
+        # contexts is a list of 0 or more enterprise uuids (or the ALL_ACCESS_CONTEXT token)
+        assigned_customers_by_role = collections.defaultdict(set)
+        for role_name, customer_uuids in super().get_assignments(user, role_names):
+            assigned_customers_by_role[role_name].update(customer_uuids or [])
```
iloveagent57, line 2901: ...also just being super defensive about the same TypeError down here.

macdiesel (approved):

adamstankiewicz (approved):

adamstankiewicz (approved):

PR #1542, opened 2022-04-20
fix: drop tpa_hint param in redirects when no SSO

johnnagro:
## Description

- when no SSO is configured for an enterprise customer, do not include a `tpa_hint=None` in the enroll -> auth redirect

## References

- [ENT-5728](https://openedx.atlassian.net/browse/ENT-5728)


alex-sheehan-edx (approved):
LGTM

PR #1544, opened 2022-04-27
fix: modify signature of EmbargoApiClient.redirect_if_blocked

iloveagent57:
Make this signature match and use the same signature that
``openedx.core.djangoapps.embargo.redirect_if_blocked()`` now uses.

https://openedx.atlassian.net/browse/ENT-5777

A change was introduced in https://github.com/openedx/edx-platform/pull/30295 that modified the signature of `embargo.redirect_if_blocked()`, this change makes enterprise code respect that new signature.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,12 @@ Unreleased
 ----------
 None
 
+[3.46.0]
+--------
+fix: modify signature of EmbargoApiClient.redirect_if_blocked
+Make this signature match and use the same signature that
+``openedx.core.djangoapps.embargo.api.redirect_if_blocked()`` now uses.
+
 [3.44.4]
 --------
 fix: implement back-off and retry for degreed2
--- a/enterprise/views.py
+++ b/enterprise/views.py
@@ -15,7 +15,6 @@ import waffle  # pylint: disable=invalid-django-waffle-import
 from dateutil.parser import parse
 from edx_django_utils import monitoring
 from edx_rest_api_client.exceptions import HttpClientError
-from ipware.ip import get_client_ip
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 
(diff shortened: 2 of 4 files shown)


johnnagro (approved):
looks like an unused import can be cleaned up

CHANGELOG.rst
```diff
@@ -17,6 +17,12 @@ Unreleased
 ----------
 None
 
+[3.46.0]
+--------
+fix: modify signature of EmbargoApiClient.redirect_if_blocked
+Make this signature match and use the same signature that
+``openedx.core.djangoapps.embargo.redirect_if_blocked()`` now uses.
```
adamstankiewicz, line 24: nit: `openedx.core.djangoapps.embargo.api.redirect_if_blocked()`?

enterprise/views.py
```diff
@@ -1786,7 +1785,10 @@ def get(self, request, enterprise_uuid, course_id):
         """
         # Check to see if access to the course run is restricted for this user.
         embargo_url = EmbargoApiClient.redirect_if_blocked(
-            [course_id], request.user, get_client_ip(request)[0], request.path)
+            request=request,
+            course_run_ids=[course_id],
```
adamstankiewicz, line 1789: [clarification] Does `redirect_if_blocked` accept `course_run_ids`? It looks like it only accepts a single `course_key` kwarg?

```python
def redirect_if_blocked(
    request: Request,
    course_key: CourseKey,
    access_point: str = 'enrollment',
    user: Optional[types.User] = None,
) -> Optional[str]:
```

iloveagent57: phew, good catch.
I should try to manually test this.

iloveagent57: ah, wait, so - `EmbargoApiClient.redirect_if_blocked()` is really just a wrapper around the `embargo_api.redirect_if_blocked()` function.  If you look at the change in `lms.py`, we're using the `course_key` argument - our wrapper calls the real function in a loop, one iteration per course run id.

adamstankiewicz: Oh, yeah, my bad! The similar names of these methods got me conflated :)

adamstankiewicz (approved):

PR #1546, opened 2022-05-02
feat: improvements to admin views

johnnagro:
## Description

- larger list sizes for reviewing data
- hide secrets from list view
- show modified dates in more places

## References

- ENT-5598


integrated_channels/integrated_channel/admin/__init__.py
```diff
@@ -16,7 +16,7 @@ class ContentMetadataItemTransmissionAdmin(admin.ModelAdmin):
         'enterprise_customer',
         'integrated_channel_code',
         'content_id',
-        'channel_metadata'
```
johnnagro, line 19: modified is useful, the giant metadata object is not useful on the list view, slows the performance of this page down

bcitro (approved):

PR #1551, opened 2022-05-09
fix: degreed2 improperly tracking completion status

johnnagro:
## Description

Degreed2 wasn't making use of passed-in completion status from `args` and was defaulting to true which made some learner audit records improperly included for sync. Since the record had no completion timestamp, validation was failing. Solution is to use the `course_completed` passed to the exporter.

## References

- [ENT-5803](https://2u-internal.atlassian.net/browse/ENT-5803)
- similar to https://github.com/edx/edx-enterprise/blob/master/integrated_channels/cornerstone/exporters/learner_data.py#L26

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 None
 
+[3.46.5]
+--------
+fix: degreed2 improperly tracking completion status
+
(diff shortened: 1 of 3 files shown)


alex-sheehan-edx (approved):
Gonna need a version bump but otherwise LGTM

PR #1553, opened 2022-05-09
Return None for context if a role assignment has no enterprise_custom…

long74100:
…er and does not apply to all contexts

https://2u-internal.atlassian.net/browse/ENT-4346

**Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,11 @@ Unreleased
 ----------
 None
 
+[3.49.0]
+--------
+fix: Return None for context if a ``SystemWideEnterpriseUserRoleAssignment`` has no enterprise_customer and does not apply to all contexts.
+     We'll no longer fall back on granting context based on enterprise membership when there is no explicit context.
+
 [3.48.0]
 --------
 chore: add migration to remove is_active from role assignment model schema
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -2792,31 +2792,10 @@ class EnterpriseCustomerReportingConfiguration(TimeStampedModel):
 class EnterpriseRoleAssignmentContextMixin:
     """
     Mixin for RoleAssignment models related to enterprises.
+
+    DEPRECATED: Not removing since it's referenced in a migration (0001_squashed_0092_auto_20200312_1650).
+
     """
-
-    @property
-    def enterprise_customer_uuids(self):
-        """Get the enterprise customer uuids linked to the user."""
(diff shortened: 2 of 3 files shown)


macdiesel (approved):

enterprise/models.py
```diff
@@ -2888,7 +2888,7 @@ def get_context(self):
         if self.enterprise_customer:
             return [str(self.enterprise_customer.uuid)]
 
-        return super().get_context()
```
christopappas, line 2891: https://github.com/openedx/edx-rbac/blob/f4d22abe35bdc3cff82f324167806d924fd95ef1/edx_rbac/models.py#L83

^ the default returns None, and it was overridden in the Mixin on line 2813. Should we instead remove the mixin override?

I guess, do we also want to return none for EnterpriseFeatureUserRoleAssignment as well?

CHANGELOG.rst
```diff
@@ -17,6 +17,14 @@ Unreleased
 ----------
 None
 
+[3.46.6]
+--------
+fix: Return None for context if a role assignment has no enterprise_customer and does not apply to all contexts
```
iloveagent57, line 22: Nit: This is actually a big enough change to warrant a new minor version.  I would also add something along the lines of "we'll no longer fall-back on granting role assignment contexts based on Enterprise membership when role assignment records have no explicit context (enterprise UUID or `*`)"

iloveagent57 (approved):
Looks good - I would do some manual checking around service users (like `enterprise_worker` or `ecommerce_worker`) to make sure their expected role assignments remain unchanged.

CHANGELOG.rst
```diff
@@ -17,6 +17,12 @@ Unreleased
 ----------
 None
 
+[3.49.0]
+--------
+fix: Return None for context if a role assignment has no enterprise_customer and does not apply to all contexts.
```
christopappas, line 22: so actually, hmm, ENT-4348 adds validation to prevent us from being able to save role assignments that have no enterprise_customer and that do not apply to all contexts 🤔 . also as part of ENT-4349, we've removed all existing system wide role assignments that meet this criteria too

long74100: so no need for this anymore?

christopappas: sure. and yeah, leave this in the changelog cause as of this commit it'll still be possible

christopappas (approved):

iloveagent57 (approved):

PR #1554, opened 2022-05-09
fix: add validation to prevent Enterprise systemwideroleassignment fr...

christopappas:
…om being saved with null enterprise_customer if has_access_to_all_contexts is not True

also, removing an old management command and all its tests (confirmed internally it is no longer needed)

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -2849,6 +2849,25 @@ class SystemWideEnterpriseUserRoleAssignment(UserRoleAssignment):
...
+
+        if self.enterprise_customer is None and self.has_access_to_all_contexts() is not True:
+            message = (
+                "Enterprise customer must be set on SystemWideEnterpriseUserRoleAssignment "
+                f"instance {self.id} because has_access_to_all_contexts is not True."
+            )
+            raise ValidationError(message)
+
+    def save(self, *args, **kwargs):
+        self._validate_whether_enterprise_customer_needed()
```
macdiesel, line 2868: I'm worried that this exception message is going to end up getting swallowed and logged in a strange way... Should I be concerned? Would it be better to just put the message to the logger in the `_validate_whether_enterprise_customer_needed` method? 

also this save method now also raises `ValidationError`.

macdiesel (approved):
One nit.

enterprise/models.py
```diff
@@ -2849,6 +2849,25 @@ class SystemWideEnterpriseUserRoleAssignment(UserRoleAssignment):
...
 
+    def _validate_whether_enterprise_customer_needed(self):
+        """
+        Verify that if enterprise_customer is None, then has_access_to_all_contexts
+        must be true.
+
+        Raises ValidationError
+        """
+
+        if self.enterprise_customer is None and self.has_access_to_all_contexts() is not True:
```
long74100, line 2860: nit: can be `and not self.has_access_to_all_contexts()`

long74100 (approved):

enterprise/models.py
christopappas: I can add a log line for sure 👍 

For what it's worth, models' save() method already raises ValidationError if a field is not correct (like if you try to set a bool field to a string)

tests/test_models.py
```diff
@@ -2348,25 +2348,6 @@ def test_get_assignments_many_assignments(self):
...
-        """
-        user_with_no_link = factories.UserFactory(email='zelda@example.com')
-        SystemWideEnterpriseUserRoleAssignment.objects.get_or_create(
-            user=user_with_no_link,
-            role=roles_api.learner_role(),
-            enterprise_customer=None,
-            applies_to_all_contexts=False,
-        )
-
-        self.assertEqual(
```
christopappas, line 2365: making a note here that i did examine https://github.com/openedx/edx-enterprise/blob/master/enterprise/management/commands/update_role_assignments_with_customers.py and it appears that we no longer create role assignments without either an enterprise_customer or applies_to_all_contexts; therefore this test should be safe to remove (without losing functional coverage)

PR #1589, opened 2022-06-21
feat: basic integrated channels task concurrency control

johnnagro:
## Description

Implement a simple tool to allow celery task concurrency control to help prevent integrated channel jobs from overrunning each other. Based on prior art linked in ticket. More abstract, generic library method perhaps in the future.

**Grooming Notes:** 

- scope of this ticket will be to provide a fix for the current issue in the integrated channels.  We will follow up with a future discovery ticket to see if we can abstract out to a more generic solution.
- “Fast” implementation: create a lock key string representing the unique instance of the job, try to set it in memcache with an expiration of the lock timeout length, subsequent runs can check for the existence of the key in memcache and bail out if it exists

## References

- [ENT-5957](https://2u-internal.atlassian.net/browse/ENT-5957)
- [ENT-5901](https://2u-internal.atlassian.net/browse/ENT-5901)

### DIFF
--- a/integrated_channels/integrated_channel/constants.py
+++ b/integrated_channels/integrated_channel/constants.py
@@ -3,3 +3,4 @@ Constants used by the integrated channels
 """
 
 ISO_8601_DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
+TASK_LOCK_EXPIRY_SECONDS = 43200
--- a/integrated_channels/integrated_channel/constants.py
+++ b/integrated_channels/integrated_channel/constants.py
@@ -3,4 +3,4 @@ Constants used by the integrated channels
 """
 
 ISO_8601_DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
-TASK_LOCK_EXPIRY_SECONDS = 43200
+TASK_LOCK_EXPIRY_SECONDS = 60 * 60 * 12
--- a/integrated_channels/integrated_channel/tasks.py
+++ b/integrated_channels/integrated_channel/tasks.py
@@ -31,7 +31,7 @@ def locked(expiry_seconds, lock_name_kwargs):  # lint-amnesty, pylint: disable=m
         def wrapper(*args, **kwargs):  # lint-amnesty, pylint: disable=inconsistent-return-statements
             cache_key = f'{func.__name__}'
             for key in lock_name_kwargs:
-                cache_key = f'{cache_key}-{key}:{kwargs.get(key)}'
+                cache_key += f'-{key}:{kwargs.get(key)}'
             if cache.add(cache_key, "true", expiry_seconds):
                 exception = None
                 try:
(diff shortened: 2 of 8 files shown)


integrated_channels/integrated_channel/tasks.py
```diff
@@ -22,6 +25,34 @@
...
+                exception = None
+                try:
+                    LOGGER.info('Locking task in cache with key: %s for %s seconds', cache_key, expiry_seconds)
+                    return func(*args, **kwargs)
+                except Exception as error:  # lint-amnesty, pylint: disable=broad-except
+                    LOGGER.exception(error)
+                    exception = error
+                finally:
+                    LOGGER.info('Unlocking task in cache with key: %s', cache_key)
+                    cache.delete(cache_key)
```
iloveagent57, line 45: I think this will execute even when the `try` block doesn't produce an exception, thus releasing the lock.  Should the `delete()` be moved to the `except` block?
https://github.com/openedx/edx-enterprise/pull/1589

integrated_channels/integrated_channel/tasks.py
```diff
@@ -22,6 +25,34 @@
 User = auth.get_user_model()
 
 
+def locked(expiry_seconds, lock_name_kwargs):  # lint-amnesty, pylint: disable=missing-function-docstring
+    def task_decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):  # lint-amnesty, pylint: disable=inconsistent-return-statements
+            cache_key = f'{func.__name__}'
+            for key in lock_name_kwargs:
+                cache_key = f'{cache_key}-{key}:{kwargs.get(key)}'
```
iloveagent57, line 34: This is over-writing `cache_key` on each pass of the loop.  Did you want something like this instead?
```python
cache_key += f'-{key}:{kwargs.get(key)}'

File: integrated_channels/integrated_channel/tasks.py (hunk 1 above)
Comment (iloveagent57) [reply integrated_channels/integrated_channel/tasks.py:45]: It depends on the problem, and the last time I thought about locking celery tests for an enterprise-catalog use case, I really wanted to lock the tasks _and_ not let the same `task(args)` run again within the expiry period.
It the goal is just to prevent the same `task(args)` from executing concurrently, then yeah, you can release the lock as soon as `task` is done.

File: integrated_channels/integrated_channel/tasks.py [hunk 3]
```diff
@@ -22,6 +25,34 @@
 User = auth.get_user_model()
 
 
+def locked(expiry_seconds, lock_name_kwargs):  # lint-amnesty, pylint: disable=missing-function-docstring
```
iloveagent57, line 28: Nit: add a docstring ;)

integrated_channels/integrated_channel/constants.py
```diff
@@ -3,3 +3,4 @@
 """
 
 ISO_8601_DATE_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
+TASK_LOCK_EXPIRY_SECONDS = 43200
```
iloveagent57, line 6: last nit: I find `60 * 60 * 12` a little easier/quicker to evaluate as 12 hours (rather than remembering that 86400 seconds is one day and dividing that by 2).

PR #1593, opened 2022-07-11
feat: refactor content metadata jobs to save remote ids, api call status

johnnagro:
## Description

Part 1 of at least 2...

We'd like to persist `ContentMetadataItemTransmission` objects even if they don't successfully get created on the remote system. We'd like to save remote content ids as well as API call response statuses and error messages on `ContentMetadataItemTransmission` objects. Before we can do that, we want to refactor some inconsistencies and complexity out of the `ContentMetadataTransmitter` to create DRY places to introduce this new functionality and persist audit information for all Content (not just successfully created content).  Notes throughout the code. Includes a management command to back-fill the new `remote_created_at` and `remote_updated_at` fields on existing records.

## TLDR

- Create a `ContentMetadataItemTransmission` for every content item we're about to send rather than after the first successful send
- Persist a record (with status/errors) even when the remote system fails to accept the content
- Track `created_at`, `updated_at` in addition to `deleted_at`
- Track remote system API response status codes and bodies 
- Clean up logging

## Deploy...

We need to back-fill the new columns so... 

1. we'd pause content metadata transmissions
2. deploy this PR 
3. add/run the included back-fill job
4. re-enable the content metadata transmissions

## Next Steps

The next steps would be to iimplemented the specific strategy for saving the remote content id yet. Not every remote system has remote ids which need to be tracked. We don't want to force every client to implement a new `create_content_metadata` signature. I think the approach we'll take for saving remote ids is to support an alternative to `client.create_content_metadata` which takes the transmission items directly instead of just the serialized chunk. This will let the individual clients track/update per-item attributes like status and remote content id. We can decide which to call based on which method is implemented by the client (https://stackoverflow.com/a/69192646). This allows legacy client code to run, while newer or updated clients can take advantage of saving remote-ids.

## References

- https://2u-internal.atlassian.net/browse/ENT-5958


integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -71,24 +71,24 @@ def __init__(self, user, enterprise_configuration):
         super().__init__(user, enterprise_configuration)
         self.enterprise_catalog_api = EnterpriseCatalogApiClient(self.user)
 
-    def _log_info(self, msg):
+    def _log_info(self, msg, course_or_course_run_key=None):
```
johnnagro, line 74: optionally pass a course_key into the logging method, which will properly format it in the logs as a splunk variable.

integrated_channels/integrated_channel/models.py
```diff
@@ -401,7 +401,28 @@ class ContentMetadataItemTransmission(TimeStampedModel):
         null=True,
     )
     deleted_at = models.DateTimeField(
-        help_text='Date when the content transmission was deleted',
+        help_text='Date when the content transmission was deleted in the remote API',
+        blank=True,
+        null=True
+    )
+    created_at = models.DateTimeField(
```
johnnagro, line 408: before this refactor, the absence of a record meant it wasn't transmitted - either because an attempt hadn't yet been made or the attempt(s) failed. after this refactor, records are created before a transmission is sent so, we need to differentiate "new and unsent" from created records.

integrated_channels/integrated_channel/models.py
```diff
@@ -401,7 +401,28 @@ class ContentMetadataItemTransmission(TimeStampedModel):
...
+        blank=True,
+        null=True
+    )
+    api_response_status_code = models.CharField(
+        help_text='The most recent remote API call response HTTP status code'
+        max_length=100,
+        blank=True,
+        null=True
+    )
+    api_response_body = models.TextField(
```
johnnagro, line 424: i wanted more descriptive names than `status`. i also want to store all response bodies, not just error response bodies.

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -32,49 +32,48 @@ def __init__(self, enterprise_configuration, client=IntegratedChannelApiClient):
...
                 self.enterprise_configuration.enterprise_customer.uuid,
                 None,
-                None,
+                course_or_course_run_key,
                 msg
             )
         )
 
-    def transmit(self, create_payload, update_payload, delete_payload, content_updated_mapping, **kwargs):
+    def _log_info_for_each_item_map(self, item_map, msg):
```
johnnagro, line 57: helper to DRY up logging a message-per-item in a transmission item map, which is done in a few places.

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -32,49 +32,48 @@ def __init__(self, enterprise_configuration, client=IntegratedChannelApiClient):
...
+    def transmit(self, create_payload, update_payload, delete_payload, **kwargs):
         """
         Transmit content metadata items to the integrated channel. Save or update content metadata records according to
         the type of transmission.
         """
-        self._log_info(
-            f'Transmitting delete payload: {delete_payload} for customer: '
-            f'{self.enterprise_configuration.enterprise_customer.uuid}'
-        )
+        self._log_info_for_each_item_map(delete_payload, 'transmitting delete')
```
johnnagro, line 70: these payloads can be large enough that they're cut-off/truncated in splunk which really diminishes their value - instead we should log a line per item using the helper above. 

later on we can work on tools to tune log verbosity, perhaps allowing changes on a per-customer or per-integration level while we're debugging a specific case while reducing the log load otherwise.

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -95,7 +94,7 @@ def _serialize_items(self, channel_metadata_items):
             sort_keys=True
         ).encode('utf-8')
 
-    def _transmit_create(self, channel_metadata_item_map, content_updated_mapping):
+    def _transmit_create(self, channel_metadata_item_map):
```
johnnagro, line 97: here is an example of an inconsistent method signature - update and delete just take transmission item maps. since we are creating the content transmission records we're about to send before we send them, we can pass an item map in here too.

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -106,7 +105,7 @@ def _transmit_create(self, channel_metadata_item_map, content_updated_mapping):
         for chunk in islice(chunk_items, transmission_limit):
             serialized_chunk = self._serialize_items(list(chunk.values()))
             try:
-                self.client.create_content_metadata(serialized_chunk)
+                response_status_code, response_body = self.client.create_content_metadata(serialized_chunk)
```
johnnagro, line 108: most clients and remote systems treat the success/failure of a whole chunk. in theory, we may later have systems which support a more granular status within that chunk. we may also need that more granular status to track remote content. in that case, we may offer an optional second `create_content_metadata` type method which accepts the chunk of items rather than the `serialized_chunk` and individually track the status, remote content id created, etc.

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -115,8 +114,10 @@ def _transmit_create(self, channel_metadata_item_map, content_updated_mapping):
                     f"Task failed with message [{exc.message}] and status code [{exc.status_code}]"
                 )
                 LOGGER.exception(exc)
-            else:
-                self._create_transmissions(chunk, content_updated_mapping)
+                response_status_code = exc.status_code
```
johnnagro, line 117: existing clients parse the response codes >= 400 and raise a `ClientError` exception, bubbling up the status and an message inside it. this is in contrast to the learner data side where the unadulterated status and body are passed up to this method for parsing. leaving this except statement in place and extracting the status allows us to avoid having to update every existing client to stop emitting a `ClientError`.

integrated_channels/integrated_channel/exporters/content_metadata.py
johnnagro, line 74: optional content key to be passed into the formatted logger which will include it as a splunk variable

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -107,6 +107,7 @@ def _get_catalog_content_keys(self, enterprise_customer_catalog):
             enterprise_customer_catalog_uuid=enterprise_customer_catalog.uuid,
             plugin_configuration_id=self.enterprise_configuration.id,
             deleted_at__isnull=True,
+            created_at__isnull=False,
```
johnnagro, line 110: see model notes on this field.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -160,6 +162,88 @@ def _check_matched_content_updated_at(
                 items_to_update[matched_item.get('content_key')] = item
         return items_to_update
 
+    def _check_matched_content_to_create(
```
johnnagro, line 165: before this refactoring, `ContentMetadataItemTransmission` were only created or undeleted _after_ they were successfully transmitted to the remote system. this refactor pre-creates them instead so we have an object to capture error messages and status on. this method as well as the semantic of pre-creating these records adds more consistency with how data flows.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -160,6 +162,88 @@ def _check_matched_content_updated_at(
...
+        ContentMetadataItemTransmission = apps.get_model(
+            'integrated_channel',
+            'ContentMetadataItemTransmission'
+        )
+        items_to_create = {}
+        create_transmissions = []
+        for matched_item in matched_items:
+            content_id = matched_item.get('content_key')
+            content_last_changed = matched_item.get('date_updated')
+            past_deleted_transmission = ContentMetadataItemTransmission.objects.filter(
```
johnnagro, line 192: look for previously deleted content and undelete it.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -262,7 +351,7 @@ def export(self, **kwargs):
 
             # We only need to fetch content metadata if there are items to update or create
             if items_to_create or items_to_update:
-                items_create_keys = [key.get('content_key') for key in items_to_create]
+                items_create_keys = list(items_to_create.keys())
```
johnnagro, line 354: we're now passing a dict of content_key : `ContentMetadataItemTransmission` item

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -273,16 +362,17 @@ def export(self, **kwargs):
                 for item in content_metadata_items:
                     key = get_content_metadata_item_id(item)
 
-                    # Used to save modified times and catalog uuid's to sent transmission records
```
johnnagro, line 276: this mapping just tracked the catalog modified at datetime as well as the catalog uuid. since we are precreating `ContentMetadataItemTransmission` items, we can store this information on those records and drop this extra dict.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -192,7 +282,7 @@ def _get_catalog_diff(self, enterprise_catalog, content_keys, force_retrieve_all
             enterprise_catalog,
             items_to_delete
         )
-        return items_to_create, content_to_update, content_to_delete
+        return content_to_create, content_to_update, content_to_delete
```
johnnagro, line 285: pass a dict of content_key : transmission item, consistent with the other payloads

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -32,49 +32,48 @@ def __init__(self, enterprise_configuration, client=IntegratedChannelApiClient):
             client=client
         )
 
-    def _log_info(self, msg, content_id=None):
+    def _log_info(self, msg, course_or_course_run_key=None):
```
johnnagro, line 35: optionally pass the course key to be properly formatted as a splunk variable.

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -163,112 +166,60 @@ def _transmit_delete(self, channel_metadata_item_map):
...
-    def _create_transmissions(self, content_metadata_item_map, content_updated_mapping):
+    def _create_transmissions(self, content_metadata_item_map, response_status_code, response_body):
         """
+        TODO FIXME old docs
         Create ContentMetadataItemTransmission models for the given content metadata items. Because records are soft
         deleted, before creating new records we must verify that there is no previously deleted record under this
         customer with this content ID. Additionally we must also check if a record exists under this customer under a
         separate catalog as content over catalogs is overwritten.
         """
-        ContentMetadataItemTransmission = apps.get_model(
```
johnnagro, line 176: we've essentially moved this logic to before the transmission.

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -163,112 +166,60 @@ def _transmit_delete(self, channel_metadata_item_map):
...
-                            content_id=content_id,
-                            channel_metadata=channel_metadata,
-                            content_last_changed=content_last_changed,
-                            enterprise_customer_catalog_uuid=catalog_uuid,
-                            plugin_configuration_id=self.enterprise_configuration.id
-                        )
-                    )
-        ContentMetadataItemTransmission.objects.bulk_create(create_transmissions, batch_size=50, ignore_conflicts=True)
+            transmission.api_response_status_code = response_status_code
+            transmission.api_response_body = response_body
```
johnnagro, line 190: i've decided to save all response bodies, not just failed bodies. we can revisit this later if it becomes a problem - perhaps make it a tunable behavior to allow verbose debugging on a per-customer basis for investigating.

integrated_channels/integrated_channel/models.py
```diff
@@ -401,7 +401,28 @@ class ContentMetadataItemTransmission(TimeStampedModel):
         null=True,
     )
     deleted_at = models.DateTimeField(
```
iloveagent57, line 403: Nit: consider using a prefix like `remote_` for these three columns, to make it extra super-duper clear that it's talking about the CUD remote actions and when they occurred. (Might be too much of a PITA because you'd have to rename the existing `deleted_at` column).

integrated_channels/integrated_channel/transmitters/content_metadata.py
iloveagent57, line 97: nit: these `_transmit_*()` methods are now extremely similar to each other and use the same method signature.  Maybe make a base method and use a `functools.partial` or something to factor this down even more?  e.g.
```python
def _transmit_action(self, item_map, client_method, final_action, action_name):
    # ...do all the work, eventually calling client_method(serialized_chunk)
    # and final_action(chunk, response_status_code, response_body)
    # all the exception handling, etc.
    # use e.g. action_name='delete' for your logging messages.

self._transmit_create = functools.partial(
    self._transmit_action, 
    self=self, 
    client_method=self.client.create_content_metadata, 
    final_action=self._create_transmissions,
    action_name='create',
)
self._transmit_update = functools.partial(
    self._transmit_action, 
    self=self, 
    client_method=self.client.update_content_metadata, 
    final_action=self._update_transmissions,
    action_name='update'
)
self._transmit_delete = functools.partial(
    self._transmit_action, 
    self=self, 
    client_method=self.client.delete_content_metadata, 
    final_action=self._delete_transmissions,
    action_name='delete'
)
```

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -163,112 +166,60 @@ def _transmit_delete(self, channel_metadata_item_map):
...
         """
-        ContentMetadataItemTransmission = apps.get_model(
-            'integrated_channel',
-            'ContentMetadataItemTransmission'
-        )
-        create_transmissions = []
-        for content_id, channel_metadata in content_metadata_item_map.items():
-            catalog_uuid = content_updated_mapping.get(content_id).get('catalog_uuid')
-            content_last_changed = content_updated_mapping.get(content_id).get('modified')
+        created_at = datetime.now()
```
iloveagent57, line 182: nit: prolly utcnow

integrated_channels/integrated_channel/exporters/content_metadata.py
iloveagent57, line 192: nit: could create a convenience classmethod on `ContentMetadataItemTransmission` that takes only `self.enterprise_configuration` and `content_id` as arguments.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -160,6 +162,88 @@ def _check_matched_content_updated_at(
...
+            content_id = matched_item.get('content_key')
+            content_last_changed = matched_item.get('date_updated')
+            past_deleted_transmission = ContentMetadataItemTransmission.objects.filter(
+                enterprise_customer=self.enterprise_configuration.enterprise_customer,
+                plugin_configuration_id=self.enterprise_configuration.id,
+                integrated_channel_code=self.enterprise_configuration.channel_code(),
+                content_id=content_id,
+                deleted_at__isnull=False,
+            ).first()
+            incomplete_transmission = ContentMetadataItemTransmission.objects.filter(
```
iloveagent57, line 199: nit: same idea here.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -160,6 +162,88 @@ def _check_matched_content_updated_at(
...
+                    'Marking record as active.',
+                    course_or_course_run_key=content_id
+                )
+                past_deleted_transmission.api_response_status_code = None
+                past_deleted_transmission.api_response_body = None
+                past_deleted_transmission.deleted_at = None
+                past_deleted_transmission.channel_metadata = None
+                past_deleted_transmission.content_last_changed = content_last_changed
+                past_deleted_transmission.enterprise_customer_catalog_uuid = enterprise_customer_catalog.uuid,
+                past_deleted_transmission.save()
```
iloveagent57, line 218: nit: would be nice as a `past_deleted_transmission.resurrect()` convenience function call.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -160,6 +162,88 @@ def _check_matched_content_updated_at(
...
+                    'Found an unsent content record while creating record. '
+                    'Including record.',
+                    course_or_course_run_key=content_id
+                )
+                incomplete_transmission.api_response_status_code = None
+                incomplete_transmission.api_response_body = None
+                incomplete_transmission.channel_metadata = None
+                incomplete_transmission.content_last_changed = content_last_changed
+                incomplete_transmission.enterprise_customer_catalog_uuid = enterprise_customer_catalog.uuid,
+                incomplete_transmission.save()
```
iloveagent57, line 231: nit: same idea here

iloveagent57 (commented):
I am generally really in favor of this change.  My comments are mostly on the function-organization and reusability level.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -160,6 +162,73 @@ def _check_matched_content_updated_at(
...
+        ContentMetadataItemTransmission = apps.get_model(
+            'integrated_channel',
+            'ContentMetadataItemTransmission'
+        )
+        items_to_create = {}
+        create_transmissions = []
+        for matched_item in matched_items:
+            content_id = matched_item.get('content_key')
+            content_last_changed = matched_item.get('date_updated')
+            past_deleted_transmission = ContentMetadataItemTransmission.deleted_transmissions(
```
johnnagro, line 192: i want to chase down the table constraints to ensure we can only have 1 record per lms config + content id (+ enterprise customer but that is implied via config id)

integrated_channels/integrated_channel/models.py
johnnagro: good idea. i am going to generate the migration last and I will tackle a column rename in that step. shouldnt be a big deal - i'll tweak the code

tests/test_integrated_channels/test_integrated_channel/test_transmitters/test_content_metadata.py
```diff
@@ -219,152 +280,87 @@ def test_transmit_update_failure(self):
...
             enterprise_customer=self.enterprise_config.enterprise_customer,
             plugin_configuration_id=self.enterprise_config.id,
             integrated_channel_code=self.enterprise_config.channel_code(),
-            content_id=content_id,
-            channel_metadata=channel_metadata
-        ).first().deleted_at
-
-    @mark.django_db
-    @ddt.ddt
-    def test_transmitting_create_content_with_previously_deleted_record(self):
```
johnnagro, line 305: the logic of resetting dates has been moved into the Exporter

tests/test_integrated_channels/test_integrated_channel/test_transmitters/test_content_metadata.py
```diff
@@ -219,152 +280,87 @@ def test_transmit_update_failure(self):
...
-        self.create_content_metadata_mock.return_value = (200, '{"success":"true"}')
-        transmitter = ContentMetadataTransmitter(self.enterprise_config)
-        assert past_transmission.deleted_at
-        transmitter.transmit(create_payload, update_payload, delete_payload, content_updated_mapping)
-        past_transmission.refresh_from_db()
-        assert not past_transmission.deleted_at
 
-    @mark.django_db
-    @ddt.ddt
-    def test_transmitting_same_content_over_different_catalogs(self):
```
johnnagro, line 337: the catalog behavior needs to be reworked - it has little value since courses can exist in multiple, overlapping catalogs

integrated_channels/cornerstone/client.py
```diff
@@ -43,24 +43,28 @@ def create_content_metadata(self, serialized_data):
         Create content metadata using the Cornerstone course content API.
         Since Cornerstone is following pull content model we don't need to implement this method
         """
+        return 200, ''
```
johnnagro, line 46: These methods were never called in the past. I am calling them now and its easier to have them return a fake response code and body. Cornerstone pulls updates via an API call versus us sending updates over. I've refactored some of the way we accomplish that to simplify things and remove inconsistencies. This change helps that along.

integrated_channels/cornerstone/transmitters/content_metadata.py
```diff
@@ -15,68 +15,3 @@ def __init__(self, enterprise_configuration, client=CornerstoneAPIClient):
             enterprise_configuration=enterprise_configuration,
             client=client
         )
-
-    def transmit(self, create_payload, update_payload, delete_payload, content_updated_mapping, **kwargs):
```
johnnagro, line 19: Cornerstone pulls content via an API call. In order to accomplish this in the past, it overrode a lot of the `Transmitter`code and added custom methods. I've tweaked the base implementation signature of `transmit` to return the lists of items it worked on (now that we save all content items). This allows us to call the normal `transmit` and work on the results rather than implement an alternative like we did in the past.

integrated_channels/cornerstone/views.py
```diff
@@ -127,7 +127,11 @@ def get(self, request, *args, **kwargs):
 
         get_count = int(request.GET.get('count', exporter.MAX_PAYLOAD_COUNT))
         content_item_count = min(get_count, exporter.MAX_PAYLOAD_COUNT)
-        data = transmitter.transmit(*exporter.export_for_web_polling(max_payload_count=content_item_count))
+        created, updated, _ = transmitter.transmit(
```
johnnagro, line 130: Cornerstone pulls content via an API call. In order to accomplish this in the past, it overrode a lot of the Transmittercode and added custom methods. I've tweaked the base implementation signature of transmit to return the lists of items it worked on (now that we save all content items). This allows us to call the normal transmit and work on the results rather than implement an alternative like we did in the past.

integrated_channels/integrated_channel/admin/__init__.py
```diff
@@ -16,7 +16,7 @@ class ContentMetadataItemTransmissionAdmin(admin.ModelAdmin):
         'enterprise_customer',
         'integrated_channel_code',
         'content_id',
-        'deleted_at',
+        'remote_deleted_at',
```
johnnagro, line 19: column was renamed.

integrated_channels/integrated_channel/management/commands/backfill_remote_action_timestamps.py
```diff
@@ -0,0 +1,76 @@
+"""
+Backfill the new remote_created_at and remote_updated_at content audit record values.
```
johnnagro, line 2: This job allows us to fill in the new `remote_created_at` and `remote_updated_at` fields to adequately correct values.

This is necessary because the refactored logic looks at `remote_updated_at` and `remote_created_at` and `null` values there have meaning - for example a blank `remote_created_at` is assumed to be a record that wasn't completely transmitted on the remote system and a re-create should be attempted.

integrated_channels/integrated_channel/migrations/0015_auto_20220718_2113.py
```diff
@@ -0,0 +1,43 @@
...
+    operations = [
+        migrations.RenameField(
+            model_name='contentmetadataitemtransmission',
+            old_name='deleted_at',
+            new_name='remote_deleted_at'
+        ),
+        migrations.AlterField(
+            model_name='contentmetadataitemtransmission',
+            name='remote_deleted_at',
+            field=models.DateTimeField(blank=True, help_text='Date when the content transmission was deleted in the remote API', null=True),
```
johnnagro, line 21: the alter just updates the `help_text`

integrated_channels/sap_success_factors/exporters/content_metadata.py
```diff
@@ -52,11 +52,14 @@ def transform_provider_id(self, content_metadata_item):  # pylint: disable=unuse
         """
         return self.enterprise_configuration.provider_id
 
-    def transform_status(self, content_metadata_item):  # pylint: disable=unused-argument
+    def transform_status(self, content_metadata_item):
```
johnnagro, line 55: Delete is sent to SAP by setting `status` to `INACTIVE`. This code needs to be aware of that and not just blindly set an `ACTIVE` status every time.

integrated_channels/sap_success_factors/transmitters/content_metadata.py
```diff
@@ -29,103 +24,18 @@ def __init__(self, enterprise_configuration, client=SAPSuccessFactorsAPIClient):
             client=client
         )
 
-    def transmit(self, create_payload, update_payload, delete_payload, content_updated_mapping, **kwargs):
-        """
-        Transmit content metadata items to the integrated channel.
+    def _transmit_action(self, content_metadata_item_map, client_method, action_name):
```
johnnagro, line 27: we don't have action-specific transformation hooks - ie an object is serialized for the remote system the same way whether its being created, updated, or deleted. That is fine for now but, SAP needs to set the `status` to "delete", this override lets us add an `INACTIVE` status to the items we're deleting before they're passed to the real `_transmit_delete`

tests/test_integrated_channels/test_blackboard/test_transmitters/test_content_metadata.py
```diff
@@ -66,15 +66,28 @@ def test_transmit_content_metadata_updates_records(
             integrated_channel_code=self.enterprise_config.channel_code(),
             content_last_changed='2021-07-16T15:11:10.521611Z',
             enterprise_customer_catalog_uuid=self.enterprise_customer_catalog.uuid,
-            channel_metadata={}
+            channel_metadata={},
```
johnnagro, line 69: this same test pattern is copied to almost every channel implementation. we have an opportunity here to DRY these tests into a single base test that takes all the implementations as input.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -273,16 +347,17 @@ def export(self, **kwargs):
...
                     transformed_item = self._transform_item(item)
                     if key in items_create_keys:
                         create_payload[key] = transformed_item
+                        existing_record = items_to_create.get(key)
+                        existing_record.channel_metadata = transformed_item
+                        existing_record.content_last_changed = item.get('content_last_modified')
+                        # Sanity check
+                        existing_record.enterprise_customer_catalog_uuid = enterprise_customer_catalog.uuid
+
+                        create_payload[key] = existing_record
```
alex-sheehan-edx, line 360: if we have this, is [line 353](https://github.com/openedx/edx-enterprise/pull/1593/files#diff-bd1856bd7ffab338eb216ef940de368327f49152f8673cc007eac15557220c78R353) still necessary?

alex-sheehan-edx (commented):
A couple of questions but this looks super cool!

integrated_channels/cornerstone/views.py
johnnagro: In our discussions this week, it sounds like we'll want to include these deleted items but ensure they have an `IsActive` key set to `false` - an attempt at this is implemented with a field transformation, but its unclear if its working properly today.

integrated_channels/integrated_channel/exporters/content_metadata.py
johnnagro: great catch

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -106,7 +106,8 @@ def _get_catalog_content_keys(self, enterprise_customer_catalog):
             integrated_channel_code=self.enterprise_configuration.channel_code(),
             enterprise_customer_catalog_uuid=enterprise_customer_catalog.uuid,
             plugin_configuration_id=self.enterprise_configuration.id,
-            deleted_at__isnull=True,
+            remote_deleted_at__isnull=True,
+            remote_created_at__isnull=False,
```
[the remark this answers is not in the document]
alex-sheehan-edx: I think I figured this out, with your change since we create the transmission audits before they're sent, if for some reason the transmission isn't sent (errors or chunk size or whatever) we would still hold the record but not have a created. Is this right?

integrated_channels/sap_success_factors/transmitters/content_metadata.py
```diff
@@ -29,117 +24,18 @@ def __init__(self, enterprise_configuration, client=SAPSuccessFactorsAPIClient):
...
-            items_to_delete (dict): dict containing the items deleted successfully.
+        Set status to INACTIVE for items that should be deleted.
         """
-        for item in failed_items:
-            content_metadata_id = item['courseID']
-            items_to_create.pop(content_metadata_id, None)
-            items_to_update.pop(content_metadata_id, None)
-            items_to_delete.pop(content_metadata_id, None)
-            update_payload.pop(content_metadata_id, None)
+        if action_name == 'delete':
```
alex-sheehan-edx, line 31: I think we'll need something like this in the cornerstone transmitted, no?

alex-sheehan-edx (approved):
this looks good!

PR #1599, opened 2022-07-20
feat: Added B2C support for nudge emails to enrolled learners

irfanuddinahmad:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/management/commands/nudge_dormant_enrolled_learners.py
+++ b/enterprise/management/commands/nudge_dormant_enrolled_learners.py
@@ -1,5 +1,5 @@
 """
-Django management command to send nudge email to dormant enrolled enterprise learners.
+Django management command to send nudge email to dormant enrolled learners.
 """
 import logging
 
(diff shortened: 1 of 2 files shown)


enterprise/management/commands/nudge_dormant_enrolled_learners.py
```diff
@@ -144,9 +172,13 @@ def get_query_results_from_snowflake(self):
...
         try:
             cs.execute(QUERY)
-            rows = cs.fetchall()
+            rows = cs.fetchmany(BATCH_SIZE)
             for row in rows:
                 yield row
+            while len(rows) > 0:
+                rows = cs.fetchmany(BATCH_SIZE)
+                for row in rows:
+                    yield row
```
muhammad-ammar, line 181: How about

```python
cs.execute(QUERY)
rows = cs.fetchmany(BATCH_SIZE)
while len(rows) > 0:
    for row in rows:
        yield row
    rows = cs.fetchmany(BATCH_SIZE)
```

enterprise/management/commands/nudge_dormant_enrolled_learners.py
```diff
@@ -144,9 +172,13 @@ def get_query_results_from_snowflake(self):
         cs = ctx.cursor()
```
muhammad-ammar, line 172: We can use [DictCursor](https://docs.snowflake.com/en/user-guide/python-connector-example.html#using-dictcursor-to-fetch-values-by-column-name). For each row this will return a `dict` and you can values by specifying the column names instead of using index numbers.

sameenfatima78 (approved):
+1 to the second feedback already added by Ammar, otherwise LGTM.

enterprise/management/commands/nudge_dormant_enrolled_learners.py
HammadAhmadWaqas: I think we don't need batch as we will not be using `fetchall`. 
```
cur = con.cursor(DictCursor)
try:
    cur.execute(QUERY)
    for row in cur:
        yield row
finally:
    cur.close()
```

irfanuddinahmad: @HammadAhmadWaqas DictCursor ... "This is a Cursor class that returns rows as dictionaries and stores the result set in the client."
http://mysql-python.sourceforge.net/MySQLdb-1.2.2/public/MySQLdb.cursors.DictCursor-class.html

HammadAhmadWaqas: @irfanuddinahmad there is [fetchManyDict](http://mysql-python.sourceforge.net/MySQLdb-1.2.2/public/MySQLdb.cursors.CursorDictRowsMixIn-class.html#fetchmanyDict) property. it says we can use fetchMany as well as fetchManyDict is deprecating.

johnnagro:
I am closing old, inactive PRs. If you feel this has been closed in error please comment.

PR #1600, opened 2022-07-20
feat: mark Cornerstone session token fields read-only in admin

johnnagro:
## Description

- mark `session_token` read-only in Django Admin to avoid confusion and errors
- mark `session_token_modified` read-only in Django Admin to avoid confusion and errors

## References

- https://2u-internal.atlassian.net/browse/ENT-6059

### DIFF
--- a/integrated_channels/cornerstone/admin/__init__.py
+++ b/integrated_channels/cornerstone/admin/__init__.py
@@ -66,6 +66,8 @@ class CornerstoneEnterpriseCustomerConfigurationAdmin(admin.ModelAdmin):
 
     readonly_fields = (
         "enterprise_customer_name",
+        "session_token",
+        "session_token_modified",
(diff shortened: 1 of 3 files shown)


integrated_channels/cornerstone/admin/__init__.py
```diff
@@ -66,6 +66,8 @@ class CornerstoneEnterpriseCustomerConfigurationAdmin(admin.ModelAdmin):
 
     readonly_fields = (
         "enterprise_customer_name",
+        "session_token",
```
johnnagro, line 69: the relevant model fields https://github.com/edx/edx-enterprise/blob/master/integrated_channels/cornerstone/models.py#L104

alex-sheehan-edx (approved):
lgtm

PR #1603, opened 2022-07-21
fix: create cornerstone learner audit records with new foreign keys

johnnagro:
## Description

Recent bugs have pointed out that we were creating `CornerstoneLearnerDataTransmissionAudit` records without the newly required `enerprise_customer_uuid` nor `plugin_configuration_id` foreign keys. We use these keys to differentiate records when a customer has more than 1 configuration of the same kind. Cornerstone creates these records in a web view as learners click through - we had missed updating that area. This corrects/amends that. Includes a management back-fill command to update stuff we've already orphaned.

I expanded the testing of the Cornerstone click-through logic. I expanded the logging in the click-through logic. I formatted all the new and existing logging to use the standard splunk key/value pairs we rely on for debugging.

## References

- https://2u-internal.atlassian.net/browse/ENT-6033
- https://2u-internal.atlassian.net/browse/ENT-6051


integrated_channels/integrated_channel/management/commands/backfill_missing_csod_foreign_keys.py
```diff
@@ -0,0 +1,84 @@
+"""
```
alex-sheehan-edx, line 1: you're going to run this through jenkins correct?

tests/test_enterprise/views/test_router_view.py
```diff
@@ -215,8 +225,19 @@ def test_get_redirects_with_course_key_and_sessionToken(
             'course_key': 'fake_course_key',
         }
         self.request.GET['sessionToken'] = 'fake_token'
+        self.request.GET['subdomain'] = 'dummy_subdomain'
```
alex-sheehan-edx, line 228: since this is required to be contained by the base url in the test config defined in the setup, would it be smart to keep it as a constant, perhaps as a class var? then you can construct the base url with said constant

tests/test_management.py
```diff
@@ -1708,3 +1712,81 @@ def cleanup_test_objects(self):
...
+        for _ in range(10):
+            factories.CornerstoneLearnerDataTransmissionAuditFactory(subdomain='csod-one')
+        for _ in range(10):
+            factories.CornerstoneLearnerDataTransmissionAuditFactory(subdomain='csod-two')
+        call_command(
+            'backfill_missing_csod_foreign_keys',
+        )
+        assert 0 == CornerstoneLearnerDataTransmissionAudit.objects.filter(plugin_configuration_id__isnull=True).count()
+
+    def cleanup_test_objects(self):
```
alex-sheehan-edx, line 1787: I thought this was taken care of for you with the fixtures 🤔

alex-sheehan-edx (commented):
looks pretty good, Waiting for small change before approval

integrated_channels/integrated_channel/management/commands/backfill_missing_csod_foreign_keys.py
johnnagro: correct

tests/test_management.py
johnnagro: i saw the same pattern above, so i followed suit

integrated_channels/integrated_channel/management/commands/backfill_missing_csod_foreign_keys.py
alex-sheehan-edx: Cool! Will look for the internal PR

alex-sheehan-edx (approved):
lgtm!

PR #1612, opened 2022-08-04
fix: overlapping content across catalogs constantly causing recreates

alex-sheehan-edx:
[related ticket](https://2u-internal.atlassian.net/browse/ENT-5849)

TLDR- we noticed that certain pieces of content were constantly being "created" (or attempted to be) for customers through the integrated channels. The common factor between all these customers and all this content was that the customers had multiple catalogs that shared the content.

**The problem:** 
The integrated channels (roughly) work as such-
- we iterate over each customer catalog
- we collect all transmission audits under each catalogs, one by one
   - specifically we filter the audit table by customer + catalog uuid
- we take the collected audits and send them to the ent catalog service and ask what work needs to be done
- we take the work that needs to be done and send it to the customer
- after successfully sending a piece of content to the customer, we save the record as a transmission audit
   - IF THERE EXISTS A RECORD OF THAT PIECE OF CONTENT FOR THE CUSTOMER UNDER A DIFFERENT CATALOG, we edit the record instead of creating a new one, swapping the catalog UUID to the new catalog. 

with this logic, on the second transmit content metadata run (and subsequently all future runs), the content will be missing under the list of sent content of one of the overlapping catalogs when we go to fetch all previous transmission audits. Thus when we ask the ent catalog service what works needs to be done, it will report that the content needs to be created. This is what creates the endless stream of "creating content X" for customers.

**The fix**
Unfortunately when collecting all audits, we can't look up content across catalogs. Due to how the catalog diff endpoint functions, we have to report content/ask for work to do catalog by catalog. Therefore the fix is to move down the line- when we get the list of work from the catalog diff endpoint, we have to check any requested creates to see if we have a record of that to-be-created content already existing on the customer's instance.   If it does already exist then we simply remove it from the list of things to do.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/exporters/content_metadata.py
+++ b/integrated_channels/integrated_channel/exporters/content_metadata.py
@@ -160,7 +160,13 @@ class ContentMetadataExporter(Exporter):
                 items_to_update[matched_item.get('content_key')] = item
         return items_to_update
 
-    def _get_catalog_diff(self, enterprise_catalog, content_keys, force_retrieve_all_catalogs, max_item_count):
+    def _get_catalog_diff(
+        self,
+        enterprise_catalog,
+        content_keys,
+        force_retrieve_all_catalogs,
+        max_item_count
+    ):
         """
         From the enterprise catalog API, request a catalog diff based off of a list of content keys. Using the diff,
         retrieve past content metadata transmission records for update and delete payloads.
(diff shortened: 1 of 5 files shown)


integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -169,12 +176,32 @@ def _get_catalog_diff(self, enterprise_catalog, content_keys, force_retrieve_all
...
+        ContentMetadataItemTransmission = apps.get_model(
+            'integrated_channel',
+            'ContentMetadataItemTransmission'
+        )
+        unique_new_items_to_create = []
+
+        # We need to remove any potential create transmissions if the content already exists on the customer's instance
+        # under a different catalog
+        for index, item in enumerate(items_to_create):
+            item_created_in_other_catalog = ContentMetadataItemTransmission.objects.filter(
```
mattdrayer, line 189: This seems like a lot of database I/O work -- does it make more sense to simply pull the full set of audit record content_ids for the customer UUID, that are not "deleted" and then reconcile with that list?

alex-sheehan-edx: good point!

mattdrayer (approved):
LGTM!  We've discussed offline about ways to improve on the `for` loop implementation, which could be handled as a follow-on enhancement.  However, given that this will typically execute as an asynchronous task on a pool worker it's less of a concern.

bcitro (approved):
🚢

johnnagro (approved):
⛵

PR #1616, opened 2022-08-11
fix: refactor the way we send cornerstone content metadata deletes

johnnagro:
## Description

Previously we were not sending deleted content messages to Cornerstone when they pulled a course update. 

This PR:
- adds delete messages into to CSOD update view
- follows the SAP pattern to set an inactive field on delete transmission actions
- moved a useful snippet into a utility class, clean up old copy/paste usage
- cleans up some debug logging and adds comments

## References

- https://2u-internal.atlassian.net/browse/ENT-6121

### DIFF
--- a/integrated_channels/cornerstone/exporters/content_metadata.py
+++ b/integrated_channels/cornerstone/exporters/content_metadata.py
@@ -93,6 +93,9 @@ class CornerstoneContentMetadataExporter(ContentMetadataExporter):
         Return the transformed version of the course is_active status by traversing course runs and setting IsActive to
         True if any of the course runs have availability value set to `Current`, `Starting Soon` or `Upcoming`.
         """
+        # lets not overwrite something we've already tried to set INACTIVE
+        if content_metadata_item.get('IsActive') is not None and content_metadata_item.get('IsActive') == False:
+            return False
         is_active = False
         for course_run in content_metadata_item.get('course_runs', []):
             if course_run.get('availability') in ['Current', 'Starting Soon', 'Upcoming']:
--- a/integrated_channels/cornerstone/exporters/content_metadata.py
+++ b/integrated_channels/cornerstone/exporters/content_metadata.py
@@ -93,9 +93,6 @@ class CornerstoneContentMetadataExporter(ContentMetadataExporter):
         Return the transformed version of the course is_active status by traversing course runs and setting IsActive to
         True if any of the course runs have availability value set to `Current`, `Starting Soon` or `Upcoming`.
         """
-        # lets not overwrite something we've already tried to set INACTIVE
-        if content_metadata_item.get('IsActive') is not None and content_metadata_item.get('IsActive') is False:
-            return False
         is_active = False
         for course_run in content_metadata_item.get('course_runs', []):
             if course_run.get('availability') in ['Current', 'Starting Soon', 'Upcoming']:
--- a/integrated_channels/cornerstone/exporters/content_metadata.py
+++ b/integrated_channels/cornerstone/exporters/content_metadata.py
@@ -94,7 +94,7 @@ class CornerstoneContentMetadataExporter(ContentMetadataExporter):
         True if any of the course runs have availability value set to `Current`, `Starting Soon` or `Upcoming`.
         """
         # lets not overwrite something we've already tried to set INACTIVE
-        if content_metadata_item.get('IsActive') is not None and content_metadata_item.get('IsActive') == False:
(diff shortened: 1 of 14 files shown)


integrated_channels/cornerstone/exporters/content_metadata.py
```diff
@@ -93,6 +93,9 @@ def transform_is_active(self, content_metadata_item):
         Return the transformed version of the course is_active status by traversing course runs and setting IsActive to
         True if any of the course runs have availability value set to `Current`, `Starting Soon` or `Upcoming`.
         """
+        # lets not overwrite something we've already tried to set INACTIVE
+        if content_metadata_item.get('IsActive') is not None and content_metadata_item.get('IsActive') is False:
```
johnnagro, line 97: this is just for safety, might not be 100% needed. eventually we'll want per-action (create, update, delete) transformations which would allow us to drop this kind of logic.

integrated_channels/sap_success_factors/transmitters/content_metadata.py
```diff
@@ -30,7 +25,6 @@ def _transmit_action(self, content_metadata_item_map, client_method, action_name
         """
         if action_name == 'delete':
             for _, item in content_metadata_item_map.items():
-                LOGGER.info(f'_transmit_delete <{item.id}>')
```
johnnagro, line 33: this is just cleaning up old debug logging i left in during another project

tests/test_integrated_channels/test_cornerstone/test_views.py
```diff
@@ -211,8 +211,8 @@ def test_course_list_restricted_count(self, mock_get_catalog_diff, mock_get_cont
         mock_get_content_metadata.return_value = get_fake_content_metadata_for_partial_create_w_program()
         response = self.client.get(url, HTTP_IF_MODIFIED_SINCE=if_modified_since)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        # 1 create, 1 update, 1 delete, no deletes sent, no count restriction, so expect 2
-        self.assertEqual(len(response.data), 2)
+        # 1 create, 1 update, 1 delete, so expect 3
```
johnnagro, line 214: update the tests to expect the delete messages

integrated_channels/integrated_channel/management/commands/backfill_missing_foreign_keys.py
```diff
@@ -67,22 +68,6 @@ class Command(IntegratedChannelCommandMixin, BaseCommand):
     Backfill missing audit record foreign keys.
     ''')
 
-    def batch_by_pk(self, ModelClass, extra_filter=Q(), batch_size=10000):
```
johnnagro, line 70: i moved this to a utility class, cleaning up old management command

integrated_channels/integrated_channel/management/commands/backfill_remote_action_timestamps.py
```diff
@@ -23,22 +23,6 @@ class Command(IntegratedChannelCommandMixin, BaseCommand):
     ./manage.py lms backfill_remote_action_timestamps
     """
 
-    def batch_by_pk(self, ModelClass, extra_filter=Q(), batch_size=10000):
```
johnnagro, line 26: i moved this to a utility class, cleaning up old management command

integrated_channels/utils.py
```diff
@@ -381,3 +382,26 @@ def is_valid_url(url):
         return all([result.scheme, result.netloc])
     except ValueError:
         return False
+
+
+def batch_by_pk(ModelClass, extra_filter=Q(), batch_size=10000):
```
johnnagro, line 387: i wrote this helper and had copy/pasted it twice before. moving it into a utility class for re-use.

bcitro (commented):
One question as I try to fully wrap my head around this - what happens if a course gets set to inactive, but becomes "active" again later (e.g. a new course run is added, or the customer's catalog query is updated)? Is there code somewhere to flip the `IsActive` field back?

johnnagro:
> 

Interesting. Originally I was just considering the delete -> recreate case, which is covered by this reset that gets run before the transformation:

https://github.com/edx/edx-enterprise/blob/jnagro/ENT-6081/0/integrated_channels/integrated_channel/models.py#L521

Since I am not 100% sure we need that safety, and in fact its not hit in testing (and the testing documents a successful delete) I can remote that check for now. Worse case, we're not consistently seeing deletes and we can re-address then.

bcitro (approved):
LGTM!

PR #1618, opened 2022-08-15
fix: adding uuid to logo string

kiram15:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -1519,7 +1519,7 @@ class PendingEnrollment(TimeStampedModel):
         return self.__str__()
 
 
-def logo_path(instance, filename):
+def logo_path(instance, filename):  # pylint: disable=unused-argument
     """
     Delete the file if it already exist and returns the enterprise customer logo image path.
 
@@ -1528,12 +1528,12 @@ def logo_path(instance, filename):
         filename (str): file to upload
 
     Returns:
-        path: path of image file e.g. enterprise/branding/<model.id>/<model_id>_logo.<ext>.lower()
+        path: path of image file e.g. enterprise/branding/logo_<uuid>.<ext>.lower()
 
     """
     extension = os.path.splitext(filename)[1].lower()
-    instance_id = str(instance.id)
-    fullname = os.path.join("enterprise/branding/", instance_id, instance_id + "_logo" + extension)
+    generated_uuid = str(uuid4())
+    fullname = os.path.join("enterprise/branding/logo_" + generated_uuid + extension)
     if default_storage.exists(fullname):
         default_storage.delete(fullname)
     return fullname
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -3,6 +3,7 @@ Tests for the `edx-enterprise` models module.
 """
 
 import json
+import re
 import logging
 import os
 import shutil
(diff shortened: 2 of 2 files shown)


enterprise/models.py
```diff
@@ -1528,12 +1528,12 @@ def logo_path(instance, filename):
...
     Returns:
-        path: path of image file e.g. enterprise/branding/<model.id>/<model_id>_logo.<ext>.lower()
+        path: path of image file e.g. enterprise/branding/logo_<uuid>.<ext>.lower()
 
     """
     extension = os.path.splitext(filename)[1].lower()
-    instance_id = str(instance.id)
-    fullname = os.path.join("enterprise/branding/", instance_id, instance_id + "_logo" + extension)
+    generated_uuid = str(uuid4())
+    fullname = os.path.join("enterprise/branding/logo_" + generated_uuid + extension)
```
johnnagro, line 1536: imho may be nice to add the `enterprise_customer_uuid` in this path too. it better organizes the files and its a natural partition key for the future, etc.

ex:
`os.path.join("enterprise/branding/" + instance.enterprise_customer.uuid + "/logo_" + generated_uuid + extension)`

enterprise/models.py
```diff
@@ -1519,7 +1519,7 @@ def __repr__(self):
         return self.__str__()
 
 
-def logo_path(instance, filename):
+def logo_path(instance, filename):  # pylint: disable=unused-argument
```
johnnagro, line 1522: i had to track down how/where the `logo_path` method gets called. i was confused that it was a free-floating method between classes (not inside a class in this file). could we move it inside the branding model class? or move it to a utility class?

enterprise/models.py
```diff
@@ -1528,12 +1528,12 @@ def logo_path(instance, filename):
...
 
     Returns:
-        path: path of image file e.g. enterprise/branding/<model.id>/<model_id>_logo.<ext>.lower()
+        path: path of image file e.g. enterprise/branding/logo_<uuid>.<ext>.lower()
 
     """
     extension = os.path.splitext(filename)[1].lower()
-    instance_id = str(instance.id)
-    fullname = os.path.join("enterprise/branding/", instance_id, instance_id + "_logo" + extension)
+    generated_uuid = str(uuid4())
```
johnnagro, line 1535: i really like that each upload gets a new uuid because it will help us deal with cache busting because each upload gets a unique asset url.

enterprise/models.py
```diff
@@ -1528,12 +1528,12 @@ def logo_path(instance, filename):
...
+    fullname = os.path.join("enterprise/branding/logo_" + generated_uuid + extension)
     if default_storage.exists(fullname):
```
johnnagro, line 1537: if each new upload has a new unique uuid, i am not sure we need this check and delete - we'll never hit it in real life?

johnnagro (changes requested):
i think your changes to the code are good. I think we also have an opportunity to leave this a little better than we found it in a few places. i left some notes throughout.

enterprise/utils.py
```diff
@@ -2331,3 +2332,22 @@ def parse_lms_api_datetime(datetime_string, datetime_format=LMS_API_DATETIME_FOR
 def localized_utcnow():
     """Helper function to return localized utcnow()."""
     return pytz.UTC.localize(datetime.datetime.utcnow())  # pylint: disable=no-value-for-parameter
+
+
+def logo_path(instance, filename):
+    """
+    Delete the file if it already exist and returns the enterprise customer logo image path.
```
johnnagro, line 2339: we should update this docstring

enterprise/utils.py
```diff
@@ -2331,3 +2332,22 @@ def parse_lms_api_datetime(datetime_string, datetime_format=LMS_API_DATETIME_FOR
...
+
+    Arguments:
+        instance (:class:`.EnterpriseCustomerBrandingConfiguration`): EnterpriseCustomerBrandingConfiguration object
+        filename (str): file to upload
+
+    Returns:
+        path: path of image file e.g. enterprise/branding/logo_<uuid>.<ext>.lower()
```
johnnagro, line 2346: we should update this docstring

johnnagro (approved):
⛵

PR #1625, opened 2022-08-24
feat: expand utility of CSOD deleted_at reset job

johnnagro:
## Description

Originally this job was meant to only reset the `remote_deleted_at` of content which hadn't yet been sent with the newly refactored content metadata code (hence the null remote api status code). Now, we'd like to be able to reset the `remote_deleted_at` for _all_ deleted CSOD content to force a re-transmission of the delete with the new, lower payload size.

## References

### DIFF
--- a/integrated_channels/integrated_channel/management/commands/reset_csod_remote_deleted_at.py
+++ b/integrated_channels/integrated_channel/management/commands/reset_csod_remote_deleted_at.py
@@ -25,7 +25,7 @@ class Command(IntegratedChannelCommandMixin, BaseCommand):
 
     def handle(self, *args, **options):
         """
-        Mark for re-send all CSOD content transmission with a remote_deleted_at but no api_response_status_code
+        Mark for re-send all CSOD content transmission with a remote_deleted_at
         """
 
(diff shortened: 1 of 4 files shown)


integrated_channels/integrated_channel/management/commands/reset_csod_remote_deleted_at.py
```diff
@@ -25,21 +25,20 @@ class Command(IntegratedChannelCommandMixin, BaseCommand):
...
         ContentMetadataItemTransmission = apps.get_model(
             'integrated_channel',
             'ContentMetadataItemTransmission'
         )
 
-        csod_deleted_at_but_null_status = Q(
+        csod_deleted_at = Q(
             integrated_channel_code='CSOD',
-            remote_deleted_at__isnull=False,
-            api_response_status_code__isnull=True
```
johnnagro, line 39: this is set to `200` with the newly refactored code, originally i filtered out stuff that had been "successfully sent" but now, given the timeouts, a bunch of deletes were marked success, given a 200 status, but need to be resent.

bcitro (approved):

PR #1627, opened 2022-08-29
feat: add debug logging to investigate missing CSOD deletes

johnnagro:
## Description

I am investigating a couple of customer reported issues with CSOD content not being properly deleted. We made previous changes which should have fixed this but we're not seeing the results we'd expect. These changes should add some more context to our investigation.

## References

### DIFF
--- a/enterprise/api_client/enterprise_catalog.py
+++ b/enterprise/api_client/enterprise_catalog.py
@@ -125,7 +125,7 @@ class EnterpriseCatalogApiClient(UserAPIClient):
         api_url = self.get_api_url(self.CATALOG_DIFF_ENDPOINT.format(catalog_uuid))
         body = {'content_keys': content_keys}
 
-        items_to_delete = {}
+        items_to_delete = []
         items_to_create = []
         items_found = []
         try:
--- a/integrated_channels/integrated_channel/exporters/content_metadata.py
+++ b/integrated_channels/integrated_channel/exporters/content_metadata.py
@@ -15,7 +15,7 @@ from django.db.models import Q
 from enterprise.api_client.enterprise_catalog import EnterpriseCatalogApiClient
 from enterprise.utils import get_content_metadata_item_id
 from integrated_channels.integrated_channel.exporters import Exporter
-from integrated_channels.utils import generate_formatted_log, resize_item_collections
+from integrated_channels.utils import generate_formatted_log, truncate_item_collections
 
 LOGGER = getLogger(__name__)
 
@@ -285,7 +285,7 @@ class ContentMetadataExporter(Exporter):
                     course_or_course_run_key=item.get('content_key')
                 )
 
-        resized_create, resized_update, resized_delete = resize_item_collections(
+        truncated_create, truncated_update, truincated_delete = truncate_item_collections(
             unique_new_items_to_create,
             matched_items,
             items_to_delete,
@@ -294,16 +294,16 @@ class ContentMetadataExporter(Exporter):
 
         content_to_create = self._check_matched_content_to_create(
             enterprise_catalog,
-            resized_create
+            truncated_create
         )
         content_to_update = self._check_matched_content_updated_at(
             enterprise_catalog,
-            resized_update,
+            truncated_update,
             force_retrieve_all_catalogs
         )
         content_to_delete = self._check_matched_content_to_delete(
             enterprise_catalog,
-            resized_delete
+            truincated_delete
         )
         return content_to_create, content_to_update, content_to_delete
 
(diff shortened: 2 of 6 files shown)


enterprise/api_client/enterprise_catalog.py
```diff
@@ -125,7 +125,7 @@ def get_catalog_diff(self, enterprise_customer_catalog, content_keys, should_rai
         api_url = self.get_api_url(self.CATALOG_DIFF_ENDPOINT.format(catalog_uuid))
         body = {'content_keys': content_keys}
 
-        items_to_delete = {}
```
johnnagro, line 128: Cleaning up a small code smell - I've tested the API and it is always an array, defaulting to a `dict` here was adding complexity elsewhere. Maybe this behaved differently in the past.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -285,32 +285,25 @@ def _get_catalog_diff(
                     course_or_course_run_key=item.get('content_key')
                 )
 
-        # if we have more to work with than the allowed space, slice it up
-        if len(unique_new_items_to_create) + len(items_to_delete) + len(matched_items) > max_item_count:
```
johnnagro, line 289: i suspected that the logic here was somehow dropping deletes as it was trying to re-size so, i moved this into a utility method with some test cases to prove it was working as expected.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -390,10 +383,17 @@ def export(self, **kwargs):
             content_keys = self._get_catalog_content_keys(enterprise_customer_catalog)
 
             self._log_info(
-                f'Retrieved content keys: {content_keys} for past transmissions to customer: '
+                f'Retrieved {len(content_keys)} content keys for past transmissions to customer: '
```
johnnagro, line 386: logging count here, then individual keys on their own line after. logging all the keys in-line like this just led to truncated log lines in splunk, so I cant tell if the course i am looking for is making it to this step.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -390,10 +383,17 @@ def export(self, **kwargs):
...
                 f'{self.enterprise_customer.uuid} under catalog: {enterprise_customer_catalog.uuid}.'
             )
 
+            for content_key in content_keys:
+                self._log_info(
+                    f'Retrieved content key: {content_key} for past transmissions to customer: '
```
johnnagro, line 392: i want to see / properly log each content key so i can query splunk with something like `integrated_channel_course_key=HarvardX+BOK` and see all the data.

integrated_channels/utils.py
```diff
@@ -405,3 +405,15 @@ def batch_by_pk(ModelClass, extra_filter=Q(), batch_size=10000):
         for item in qs:
             start_pk = item.pk
         qs = ModelClass.objects.filter(pk__gt=start_pk).filter(extra_filter).order_by('pk')[:batch_size]
+
+
+def resize_item_collections(items_to_create, items_to_update, items_to_delete, combined_maximum_size):
```
johnnagro, line 410: new utility method broken out so we could properly test the logic.

tests/test_integrated_channels/test_utils.py
```diff
@@ -276,3 +276,36 @@ def test_is_url_valid(self):
     @ddt.unpack
     def test_get_courserun_duration_in_hours(self, course_run, expected_duration_days):
         assert utils.get_courserun_duration_in_hours(course_run) == expected_duration_days
+
+    def test_resize_item_collections(self):
```
johnnagro, line 280: tests of the resize logic, we indirectly tested it in the past as part of other tests but never directly tested the resize/truncate logic independently.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -285,32 +285,25 @@ def _get_catalog_diff(
...
-            unique_new_items_to_create = unique_new_items_to_create[0:max_item_count]
-            count_left = max_item_count - len(unique_new_items_to_create)
-            matched_items = matched_items[0:count_left]
-            count_left = count_left - len(matched_items)
-            # in testing, this is sometimes a list, sometimes a dict
-            # we need it to be a list for our purposes
-            if isinstance(items_to_delete, dict):
-                items_to_delete = list(items_to_delete.values())
-            items_to_delete = items_to_delete[0:count_left]
+        truncated_create, truncated_update, truincated_delete = truncate_item_collections(
```
bcitro, line 288: nit: typo on `truincated_delete`

bcitro (approved):
one typo fix, otherwise LGTM

PR #1629, opened 2022-08-30
feat: add debug logging to investigate missing CSOD deletes

johnnagro:
## Description

I've narrowed down where we seem to be losing the content delete records - this adds additional debug logging.

## References

- #1627


tests/test_integrated_channels/test_utils.py
```diff
@@ -309,3 +309,25 @@ def test_truncate_item_collections(self):
         assert len(out_a) == 3
         assert len(out_b) == 5
         assert len(out_c) == 2
+
+        in_a = []
```
johnnagro, line 313: i wanted to add some test cases where the first couple buckets are small or empty to make sure we'd still get the delete content (bucket c)

kiram15 (approved):
Good luck!

PR #1630, opened 2022-08-31
fix: properly truncate payload to resolve missing CSOD deletes

johnnagro:
## Description

We were truncating our content payloads before we were filtering out items we were not going to update - this was resulting in missed updates and deletes to content in the remote LMS when using `max_payload_count` less than catalog size (typical with CSOD specifically). This PR re-orders the truncate to happen later in the diff process. It also adds a test which replicates the bug behavior.

In the steady synced state, the entire customer catalog is returned as a matched item from the diff endpoint and then we filter that down (to nothing) after we compare last modified dates to determine if an update needs to be send. Because we truncate to `max_payload_count` _before_ we compare dates to filter, we were only ever filtering on an update collection of the first `max_payload_count` and a truncated (empty) delete collection instead of truncating _after_ we collect all the updates and deletes we'd like to send.

## References

- https://github.com/openedx/edx-enterprise/pull/1629
- https://github.com/openedx/edx-enterprise/pull/1627
- https://github.com/openedx/edx-enterprise/pull/1625

### DIFF
--- a/integrated_channels/utils.py
+++ b/integrated_channels/utils.py
@@ -416,9 +416,9 @@ def truncate_item_dicts(items_to_create, items_to_update, items_to_delete, combi
     # if we have more to work with than the allowed space, slice it up
     if len(items_to_create) + len(items_to_delete) + len(items_to_update) > combined_maximum_size:
         # prioritize creates, then updates, then deletes
-        items_to_create = dict(itertools.islice(items_to_create.items(),combined_maximum_size))
+        items_to_create = dict(itertools.islice(items_to_create.items(), combined_maximum_size))
         count_left = combined_maximum_size - len(items_to_create)
-        items_to_update = dict(itertools.islice(items_to_update.items(),count_left))
+        items_to_update = dict(itertools.islice(items_to_update.items(), count_left))
         count_left = count_left - len(items_to_update)
-        items_to_delete = dict(itertools.islice(items_to_delete.items(),count_left))
+        items_to_delete = dict(itertools.islice(items_to_delete.items(), count_left))
     return items_to_create, items_to_update, items_to_delete
--- a/integrated_channels/utils.py
+++ b/integrated_channels/utils.py
@@ -3,6 +3,7 @@ Utilities common to different integrated channels.
 """
 
 import base64
+import itertools
 import json
 import math
 import re
@@ -407,7 +408,7 @@ def batch_by_pk(ModelClass, extra_filter=Q(), batch_size=10000):
         qs = ModelClass.objects.filter(pk__gt=start_pk).filter(extra_filter).order_by('pk')[:batch_size]
 
 
-def truncate_item_collections(items_to_create, items_to_update, items_to_delete, combined_maximum_size):
+def truncate_item_dicts(items_to_create, items_to_update, items_to_delete, combined_maximum_size):
     """
     given the item collections to create, update, and delete on the remote LMS side, truncate the
     collections to keep under a maximum batch size, prioritizing creates, updates, then deletes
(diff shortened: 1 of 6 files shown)


tests/test_integrated_channels/test_integrated_channel/test_exporters/test_content_metadata.py
```diff
@@ -167,6 +168,80 @@ def test_content_exporter_failed_create_export(self, mock_get_catalog_diff, mock
...
+            enterprise_customer_catalog_uuid=self.config.enterprise_customer.enterprise_customer_catalogs.first().uuid,
+            remote_created_at=datetime.datetime.utcnow(),
+        )
+        past_transmission_to_delete.save()
+
+        mock_delete_items = [{'content_key': FAKE_COURSE_RUN2['key']}]
+        mock_api_client.return_value.get_catalog_diff.return_value = mock_create_items, mock_delete_items, \
+            mock_matched_items
+
+        create_payload, update_payload, delete_payload = exporter.export(max_payload_count=1)
```
johnnagro, line 237: at this point in the test, the mock diff endpoint is returning a matched item and a not found (delete) item, with the max payload count of 1, the old code was ultimately not sending the delete because the early truncate was seeing the single matched item as exhausting the may_payload_count available space, even though later in the code we'd determine that we didnt need to send the udpate.

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_content_metadata.py
```diff
@@ -167,6 +168,80 @@ def test_content_exporter_failed_create_export(self, mock_get_catalog_diff, mock
         assert not update_payload
         assert not delete_payload
 
+    @mock.patch('integrated_channels.integrated_channel.exporters.content_metadata.EnterpriseCatalogApiClient')
+    def test_content_exporter_truncation_bug_export(self, mock_api_client):
```
johnnagro, line 172: this test replicates the behavior we were seeing in production. it fails with the old order of truncate operations.

bcitro (approved):

alex-sheehan-edx (approved):
lgtm

tests/test_integrated_channels/test_utils.py
```diff
@@ -277,56 +277,56 @@ def test_is_url_valid(self):
     def test_get_courserun_duration_in_hours(self, course_run, expected_duration_days):
         assert utils.get_courserun_duration_in_hours(course_run) == expected_duration_days
 
-    def test_truncate_item_collections(self):
-        in_a = [1, 2, 3]
-        in_b = [1, 2, 3, 4, 5]
-        in_c = [1, 2, 3, 4, 5, 6, 7]
+    def test_truncate_item_dicts(self):
```
johnnagro, line 280: because we're truncating later in the workflow, we are now working with `dict` data structures instead of `list` - this truncate needed to be reworked and retested

integrated_channels/utils.py
```diff
@@ -407,17 +408,17 @@ def batch_by_pk(ModelClass, extra_filter=Q(), batch_size=10000):
         qs = ModelClass.objects.filter(pk__gt=start_pk).filter(extra_filter).order_by('pk')[:batch_size]
 
 
-def truncate_item_collections(items_to_create, items_to_update, items_to_delete, combined_maximum_size):
+def truncate_item_dicts(items_to_create, items_to_update, items_to_delete, combined_maximum_size):
```
johnnagro, line 411: because we're truncating later in the workflow, we are now working with `dict` data structures instead of `list` - this truncate needed to be reworked and retested

bcitro (approved):

alex-sheehan-edx (approved):
lgtm

PR #1635, opened 2022-09-13
feat: Added POST support for catalog query preview

irfanuddinahmad:
**Description** This PR updates the catalog query preview link to use POST instead of GET to enable large size of arguments

**JIRA** [ENT-6179](https://2u-internal.atlassian.net/browse/ENT-6179)

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/admin/views.py
+++ b/enterprise/admin/views.py
@@ -5,6 +5,9 @@ Custom Django Admin views used in enterprise app.
 import datetime
 import logging
 
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
 from django.conf import settings
 from django.contrib import admin, auth, messages
 from django.contrib.auth import get_permission_codename
@@ -31,11 +34,13 @@ from enterprise.admin.utils import (
     parse_csv,
     split_usernames_and_emails,
 )
+from enterprise.api_client.discovery import get_course_catalog_api_service_client
 from enterprise.api_client.ecommerce import EcommerceApiClient
 from enterprise.constants import PAGE_SIZE
 from enterprise.errors import LinkUserToEnterpriseError
 from enterprise.models import (
     EnrollmentNotificationEmailTemplate,
+    EnterpriseCatalogQuery,
     EnterpriseCustomer,
     EnterpriseCustomerUser,
     PendingEnterpriseCustomerUser,
@@ -100,6 +105,30 @@ class TemplatePreviewView(View):
         Get a human-readable name for the user.
         """
         return request.user.first_name or request.user.username
+
+
+class CatalogQueryPreviewView(APIView):
+    """
+    Renders a search/all response from Discovery for a particular catalog query.
+    """
+
+    def get(self, request, catalog_query_id):
+        """
+        Render the response for a particular catalog query.
+        """
+        enterprise_catalog_query = get_object_or_404(EnterpriseCatalogQuery, pk=catalog_query_id)
+        return Response(self.get_response_from_discovery(request, enterprise_catalog_query))
+
+    @staticmethod
+    def get_response_from_discovery(request, catalog_query):
+        """
+        POST a query to Discovery's search/all endpoint and return the results.
+        """
+        response = get_course_catalog_api_service_client().get_catalog_results(
+            catalog_query.content_filter,
+            traverse_pagination=True
+        )
+        return response
 
 
 class BaseEnterpriseCustomerView(View):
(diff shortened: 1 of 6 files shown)


enterprise/admin/views.py
```diff
@@ -102,6 +104,32 @@ def get_user_name(request):
...
+        if not enterprise_catalog_query:
+            return HttpResponse(status=404)
+        return JsonResponse(self.get_response_from_discovery(request, enterprise_catalog_query))
+
+    @staticmethod
+    def get_response_from_discovery(request, catalog_query):
+        """
+        POST a query to Discovery's search/all endpoint and return the results.
+        """
+        response = get_course_catalog_api_service_client().get_catalog_results(
```
bilaltahir21, line 126: Please pass the `exclude_expired` and `detail_fields` in the `query_params` as this might create an issue with the results of search/all endpoint because we get these parameters from the GET request in the course serializer of search/all API endpoint. For reference of getting parameters from GET request https://github.com/bilaltahir21/course-discovery/blob/e2ea41d45eccd023c13cd78cf36f313046119fc5/course_discovery/apps/course_metadata/search_indexes/serializers/course.py#L67

enterprise/admin/views.py
```diff
@@ -102,6 +107,32 @@ def get_user_name(request):
         return request.user.first_name or request.user.username
 
 
+class CatalogQueryPreviewView(APIView):
```
HammadAhmadWaqas, line 110: Should we add admin-only access validation on this view.

enterprise/admin/views.py
HammadAhmadWaqas: Yes, I think we should see if any of these keys are present in `catalog_query.content_filter` then we should extract them from `catalog_query.content_filter` and pass them in `query_params` of the `get_catalog_results` method.

enterprise/admin/views.py
irfanuddinahmad: These are only for the admin site

enterprise/admin/views.py
irfanuddinahmad: Do we ever use these in the content_filter?

irfanuddinahmad: the GET preview did not apply this processing on the content_filter ...

irfanuddinahmad: I would avoid modifying the clients to fix an issue that is best handled on the server side
 … from an architecture perspective the client side should not have to worry about which filters to send and which not …

enterprise/admin/views.py
```diff
@@ -102,6 +104,32 @@ def get_user_name(request):
...
+    """
+    Renders a search/all response from Discovery for a particular catalog query.
+    """
+
+    def get(self, request, catalog_query_id):
+        """
+        Render the response for a particular catalog query.
+        """
+        enterprise_catalog_query = get_object_or_404(EnterpriseCatalogQuery, pk=catalog_query_id)
+        if not enterprise_catalog_query:
```
HammadAhmadWaqas, line 117: You don't need to handle 404 manually, As we are using `get_object_or_404`

tests/test_admin/test_view.py
```diff
@@ -153,6 +154,58 @@ def test_missing_object(self):
...
+        )
+        result = self.client.get(url)
+        content_count = json.loads(result.content)['count']
+        self.assertEqual(content_count, 3)
+
+    def test_missing_object(self):
+        """
+        Test that a missing template object causes a 404.
+        """
+        assert self.client.login(username=self.user.username, password="QWERTY")
```
HammadAhmadWaqas, line 200: Can we add a test for a non-admin and non-staff users, to make sure they don't have access to this API

HammadAhmadWaqas (dismissed):

enterprise/admin/views.py
```diff
@@ -102,6 +104,32 @@ def get_user_name(request):
...
+            return HttpResponse(status=404)
```
HammadAhmadWaqas, line 118: This line never doing to be executed.

tests/test_admin/test_view.py
```diff
@@ -153,6 +154,79 @@ def test_missing_object(self):
...
+                return_value=fake_catalog_api.FAKE_SEARCH_ALL_RESULTS
+            ),
+        )
+        assert self.client.login(username=self.non_staff_user.username, password="QWERTY")
+        url = reverse(
+            'admin:' + enterprise_admin.utils.UrlNames.PREVIEW_QUERY_RESULT,
+            args=(self.catalog_query.pk,)
+        )
+        result = self.client.get(url)
+        self.assertEqual(result.content, b'')
```
HammadAhmadWaqas, line 227: can we assert the status code as well?

bilaltahir21 (approved):

tests/test_admin/test_view.py
irfanuddinahmad: Yes. However, it is a 302 redirect to login page

saleem-latif (approved):

PR #1644, opened 2022-09-26
feat: tagging exported content records for transmission

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-6339

exporters - always tag content records for transmission
transmitters - always removes tag on content records after transmitting

cornerstone 
                      - now included with all other channels for regularly scheduled transmissions
                      - now has a dummy transmitter transmit method
                      - now has a `transmit for web` transmitter method that functions like the normal transmitter

during scheduled transmissions, cornerstone will default to the dummy transmitter method that will result in a noop. During pulls, it will use the regular transmitter transmit method

**Merge checklist:**

**Post merge:**


integrated_channels/cornerstone/exporters/content_metadata.py
```diff
@@ -53,13 +53,7 @@ def export_for_web_polling(self, max_payload_count=MAX_PAYLOAD_COUNT):
         """
         Return the exported and transformed content metadata as a dictionary for CSDO web pull.
         """
-        return self.export(max_payload_count=max_payload_count)
-
-    def export_force_all_catalogs(self):
```
alex-sheehan-edx, line 58: this isn't used by anything anymore

johnnagro, line 58: note - we dont use this anymore which is why we're removing it

integrated_channels/cornerstone/exporters/content_metadata.py
```diff
@@ -53,13 +53,7 @@ def export_for_web_polling(self, max_payload_count=MAX_PAYLOAD_COUNT):
...
-        """
-        Return the exported and transformed content metadata as a dictionary regardless if there is an update needed.
-        """
-        return self.export(force_retrieve_all_catalogs=True)
+        return self.export(max_payload_count=max_payload_count, enable_cache=True)
```
johnnagro, line 56: i'd pipe `enable_cache` in as a kwarg for `export_for_web_polling`

integrated_channels/cornerstone/transmitters/content_metadata.py
```diff
@@ -16,12 +16,27 @@ def __init__(self, enterprise_configuration, client=CornerstoneAPIClient):
             client=client
         )
 
-    def _transmit_action(self, content_metadata_item_map, client_method, action_name):
+    def transmit(self, create_payload, update_payload, delete_payload, **kwargs):
```
johnnagro, line 19: i dont personally think we should handle the cache via a kwarg here. i'll huddle and suggest an alternative

integrated_channels/integrated_channel/models.py
```diff
@@ -518,6 +525,37 @@ def incomplete_delete_transmissions(
         )
         return ContentMetadataItemTransmission.objects.filter(in_db_but_failed_to_send_query)
 
+    def mark_transmission(self, mark_for):
```
johnnagro, line 528: `mark_transmission` -> `_mark_transmission` if its an internal method, given the existence of things like `mark_for_create`

johnnagro (changes requested):
lets huddle

integrated_channels/cornerstone/transmitters/content_metadata.py
johnnagro: if we use a kwarg here to skip clearing of these marked records, lets make sure it also causes us to skip updating their timestamps, api reponses, and error messages

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -87,14 +87,15 @@ def transmit(self, create_payload, update_payload, delete_payload, **kwargs):
         Transmit content metadata items to the integrated channel. Save or update content metadata records according to
         the type of transmission.
         """
+        remove_mark_for = kwargs.get('remove_mark_for', True)
         self._log_info_for_each_item_map(delete_payload, 'transmitting delete')
-        delete_payload_results = self._transmit_delete(delete_payload)
+        delete_payload_results = self._transmit_delete(delete_payload, remove_mark_for=remove_mark_for)
```
johnnagro, line 92: perhaps pass `**kwargs` instead of a specific kwarg?

johnnagro:
@alex-sheehan-edx and I have had a few huddles to discuss this PR as well as thought-experiment possible alternative approaches. I think we've landed on something good and he's going to send that draft out for review shortly.

integrated_channels/cornerstone/transmitters/content_metadata.py
```diff
@@ -16,6 +16,18 @@ def __init__(self, enterprise_configuration, client=CornerstoneAPIClient):
             client=client
         )
 
+    def transmit(self, create_payload, update_payload, delete_payload):
```
alex-sheehan-edx, line 19: cornerstone transmitter.transmit -> does nothing
cornerstone transmitter.trasnmit_for_web -> does what the normal transmit does

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -209,39 +211,43 @@ def _check_matched_content_to_create(
                 integrated_channel_code=self.enterprise_configuration.channel_code(),
                 content_id=content_id,
             ).first()
-            incomplete_transmission = ContentMetadataItemTransmission.incomplete_create_transmissions(
```
alex-sheehan-edx, line 212: This lookup doesn't have to happen unless we've verified that the previous lookup (`past_deleted_transmission`) does not exist, so I moved it to after the `else`

integrated_channels/integrated_channel/management/commands/__init__.py
```diff
@@ -45,6 +45,7 @@
     for integrated_channel_class in (
         BlackboardEnterpriseCustomerConfiguration,
         CanvasEnterpriseCustomerConfiguration,
+        CornerstoneEnterpriseCustomerConfiguration,
```
alex-sheehan-edx, line 48: Cornerstone would now be included in the regularly scheduled transmit content task

integrated_channels/cornerstone/exporters/content_metadata.py
```diff
@@ -49,17 +49,18 @@ class CornerstoneContentMetadataExporter(ContentMetadataExporter):
...
-    def export_force_all_catalogs(self):
-        """
-        Return the exported and transformed content metadata as a dictionary regardless if there is an update needed.
-        """
-        return self.export(force_retrieve_all_catalogs=True)
+        ContentMetadataItemTransmission = apps.get_model(
+            'integrated_channel',
+            'ContentMetadataItemTransmission'
+        )
+        create_payload = ContentMetadataItemTransmission.objects.filter(marked_for='create')
```
johnnagro, line 60: we also need to filter by `enterprise_customer` and `plugin_config_id`

integrated_channels/integrated_channel/management/commands/__init__.py
johnnagro: you need to update the comment on line 41/42

integrated_channels/cornerstone/transmitters/content_metadata.py
johnnagro: you can probably drop the `**kwargs`

integrated_channels/cornerstone/transmitters/content_metadata.py
```diff
@@ -16,6 +16,18 @@ def __init__(self, enterprise_configuration, client=CornerstoneAPIClient):
             client=client
         )
 
+    def transmit(self, create_payload, update_payload, delete_payload, **kwargs):
+        """
+        Placholder
+        """
+        pass
+
+    def transmit_for_web(self, create_payload, update_payload, delete_payload, **kwargs):
```
johnnagro, line 25: i think you can drop the `**kwargs`

integrated_channels/cornerstone/transmitters/content_metadata.py
```diff
@@ -16,6 +16,18 @@ def __init__(self, enterprise_configuration, client=CornerstoneAPIClient):
...
+        """
+        Placeholder
+        """
+        return super().transmit(create_payload, update_payload, delete_payload, **kwargs)
```
johnnagro, line 29: drop dem `**kwargs`

integrated_channels/cornerstone/transmitters/content_metadata.py
```diff
@@ -16,6 +16,18 @@ def __init__(self, enterprise_configuration, client=CornerstoneAPIClient):
             client=client
         )
 
+    def transmit(self, create_payload, update_payload, delete_payload, **kwargs):
+        """
+        Placholder
+        """
+        pass
```
johnnagro, line 23: lets also log that this has run and we're deliberately doing nothing, i think you can use the `_log_info` here to properly pass all the variables for splunk

johnnagro (changes requested):
looking really good! one critical thing needs to change  - your filtering on the cached export needs to be customer, plugin-specific. otherwise some nits.

integrated_channels/cornerstone/exporters/content_metadata.py
alex-sheehan-edx: agreed

integrated_channels/cornerstone/views.py
```diff
@@ -127,8 +127,14 @@ def get(self, request, *args, **kwargs):
 
         get_count = int(request.GET.get('count', exporter.MAX_PAYLOAD_COUNT))
         content_item_count = min(get_count, exporter.MAX_PAYLOAD_COUNT)
-        created, updated, deleted = transmitter.transmit(
-            *exporter.export_for_web_polling(max_payload_count=content_item_count)
+        # Remove content records `marked for transmission` flags
+        created, updated, deleted = exporter.export_for_web_polling(max_payload_count=content_item_count)
+
+        create_payload = {record.content_id: record for record in created}
```
johnnagro, line 133: ah. this should be inside the `export_for_web_polling()` method. the export signature should be consistent - including the structure of its return values.

alex-sheehan-edx: done

johnnagro (approved):
⛵

PR #1651, opened 2022-09-29
fix: impoving transmission records by moving response body to new table

alex-sheehan-edx:
CAT-1 performance related issues

**IN THIS PR**
We are attempting to increase efficiency of the transmit content metadata task. In order of impact on the system performance we are-

**High Impact**
- Reducing the size each row in the `ContentMetadataItemTransmission` table by migrating it's largest field `api_response_body` to another table (`ApiResponseRecord`). Now, instead of saving the json blob to the transmission record, there is a 1:1 relationship between transmission records and the `ApiResponseRecord` table that houses the response information.
- Adding an index on the `ContentMetadataItemTransmission` table between the `enterprise_customer`,  `integrated_channel_code`, `plugin_configuration_id`, and `content_id` columns. Each task run does an enormous amount of table scans, querying by those fields. As such the index should help with the job run.

**Low Impact**
- slightly reduce the number of saves on the `ContentMetadataItemTransmission` during a content metadata transmission

**Merge checklist:**

**Post merge:**


integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -243,10 +243,10 @@ def _check_matched_content_to_create(
                         channel_metadata=None,
                         content_last_changed=content_last_changed,
                         enterprise_customer_catalog_uuid=enterprise_customer_catalog.uuid,
-                        plugin_configuration_id=self.enterprise_configuration.id
+                        plugin_configuration_id=self.enterprise_configuration.id,
```
alex-sheehan-edx, line 246: saving an additional table save

integrated_channels/integrated_channel/migrations/0021_remove_contentmetadataitemtransmission_api_response_body.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 3.2.15 on 2022-09-29 17:20
+
+from django.db import migrations
```
alex-sheehan-edx, line 3: Removing a relatively large field from the table for performance reasons

integrated_channels/integrated_channel/migrations/0022_alter_contentmetadataitemtransmission_index_together.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.15 on 2022-09-29 19:01
+
+from django.db import migrations
```
alex-sheehan-edx, line 3: Adding an index on 
```
'enterprise_customer', 'integrated_channel_code', 'plugin_configuration_id', 'content_id'
```
to speed up table scans

integrated_channels/integrated_channel/models.py
```diff
@@ -414,6 +439,8 @@ class ContentMetadataItemTransmission(TimeStampedModel):
 
     .. no_pii:
     """
+    class Meta:
```
alex-sheehan-edx, line 442: https://github.com/openedx/edx-enterprise/pull/1651/files#r983927867

johnnagro (approved):

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -184,7 +185,18 @@ def _transmit_action(self, content_metadata_item_map, client_method, action_name
                         course_or_course_run_key=content_id
                     )
                     transmission.api_response_status_code = response_status_code
```
bcitro, line 187: should this line be deleted?

bcitro (approved):
One question otherwise LGTM

PR #1656, opened 2022-10-20
chore: smarter canvas client update course behavior (course deleted)

alfredoguillem:
**Merge checklist:**

**Post merge:**


integrated_channels/canvas/client.py
```diff
@@ -169,13 +169,31 @@ def update_content_metadata(self, serialized_data):
...
             return self.create_content_metadata(serialized_data)
-
-        canvas_course_id = canvas_course.get('id')
-        url = CanvasUtil.course_update_endpoint(
-            self.enterprise_configuration,
-            canvas_course_id,
-        )
-        return self._put(url, serialized_data)
+        else:
+            workflow_state = canvas_course['workflow_state']
```
alex-sheehan-edx, line 173: This should account for the test failures
```
if workflow_state := canvas_course.get('workflow_state']:
    # If the course was deleted, don't update
    if workflow_state.lower() == 'deleted':
       ....
```

alex-sheehan-edx: https://docs.python.org/3/whatsnew/3.8.html have a look at the walrus operator- essentially it assigns the value if the IF statement equates to true.

johnnagro: imho

we are getting a `KeyError` right? given the test conditions, `'workflow_state'` is not a key in the `canvas_course` dictionary. assuming the test conditions don't need to be changed- and you should double check that i may suggest you use a `get()` rather than `[]`? meaning:

`workflow_state = canvas_course.get('workflow_state', '')`

`get()` wont blow up if the key doesn't exist, it returns `None` or whatever you set as the default, in this case an empty string (`''`).

Using this variation, when that key doesn't exist, `worflow_state` would be an empty string so the `.lower()` would still work in the next if statement.

In my opinion, using `get` with a default `''` is more clear than the walrus operator and i _think_ the walrus operator as written would probably just leave the `workflow_state` value as `None`, which would break when `.lower()` was called anyway?





[10:00](https://edx-internal.slack.com/archives/D03R7RHTXD0/p1666792827415729)
which would blow up when you try to call .lower() on it later anyway

tests/test_integrated_channels/test_canvas/test_client.py
```diff
@@ -994,6 +994,35 @@ def test_course_update_creates_when_course_id_not_found(self, mock_find_course_b
         assert status == 200
         assert response == mocked_create_messaged
 
+    @mock.patch.object(CanvasUtil, 'find_course_by_course_id')
```
alex-sheehan-edx, line 997: I __love__ your test!! 🥳

alex-sheehan-edx (approved):
Fantastic work! LGTM

integrated_channels/canvas/client.py
alex-sheehan-edx: heh you raise a fair point. I just got too excited to use a walrus operator 😄

PR #1657, opened 2022-10-20
feat: add signed token generation for plotly

muhammad-ammar:
**JIRA:** https://2u-internal.atlassian.net/browse/ENT-6434

**edx-internal for plotly secret:** https://github.com/edx/edx-internal/pull/7299

**Plotly docs link for auth:** https://enterprise-plotly.edx.org/Docs/embedded-middleware/authentication _This url is behind edx vpn_

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -4,8 +4,10 @@ Views for enterprise api version 1 endpoint.
 
 from logging import getLogger
 from smtplib import SMTPException
+from time import time
 from urllib.parse import quote_plus, unquote
 
+import jwt
 import requests
 from django_filters.rest_framework import DjangoFilterBackend
 from edx_rbac.decorators import permission_required
@@ -35,7 +37,7 @@ from django.apps import apps
 from django.conf import settings
 from django.core import exceptions, mail
 from django.db import transaction
-from django.http import Http404
+from django.http import Http404, JsonResponse
 from django.shortcuts import get_object_or_404
 from django.utils.dateparse import parse_datetime
 from django.utils.decorators import method_decorator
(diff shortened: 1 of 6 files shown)


enterprise/api/v1/views.py
```diff
@@ -1691,3 +1694,44 @@ def link_user(self, request, pk=None):
...
+        # TODO: Update endpoint. The logic in this endpoint is temporarily needed for a POC.
+
+        # This is a new secret key and will be only shared between LMS and our Plotly server.
+        secret_key = settings.ENTERPRISE_PLOTLY_SECRET
+
+        now = int(time())
+        expires_in = 3600 # time in seconds after which token will be expired
+        exp = now + expires_in
+
+        AUDIENCE = 'dash.tests:8050' # TBD: Should be a appropriate value?
```
irfanuddinahmad, line 1722: we can set this to enterprise-plotly.edx.org ... or not at all since AUDIENCE claim in optional

enterprise/api/v1/views.py
```diff
@@ -1691,3 +1694,44 @@ def link_user(self, request, pk=None):
...
+        CLAIMS = {
+            "iss": "DASH EMBEDDED", # TBD: Should be a appropriate value?
```
irfanuddinahmad, line 1724: I think DASH EMBEDDED is a keyword ... we should not change this claim

enterprise/api/v1/views.py
```diff
@@ -1691,3 +1694,44 @@ def link_user(self, request, pk=None):
...
+
+    @permission_required(
+        'enterprise.can_access_admin_dashboard',
+        fn=lambda request, enterprise_uuid: enterprise_uuid
+    )
+    def get(self, request, enterprise_uuid):
+        """
+        Generate auth token for plotly.
+        """
+        # TODO: Update endpoint. The logic in this endpoint is temporarily needed for a POC.
```
irfanuddinahmad, line 1713: I think we are past the POC stage

moconnell1453 (approved):

irfanuddinahmad (approved):
LGTM!

tests/test_enterprise/api/test_views.py
```diff
@@ -5503,3 +5503,60 @@ def test_same_enable_universal_link(self):
...
+        """
+        response = self.client.get(self.url)
+        assert response.status_code == status.HTTP_403_FORBIDDEN
+        assert response.json() == {'detail': 'Missing: enterprise.can_access_admin_dashboard'}
+
+    def test_view_with_admin_user(self):
+        """
+        Verify that an enterprise admin user having `enterprise.can_access_admin_dashboard` role can access the view.
+        """
+        self.set_jwt_cookie(ENTERPRISE_ADMIN_ROLE, self.enterprise_uuid)
```
HammadAhmadWaqas, line 5540: nit: we can move `set_jwt_cookie` to `setUp` and rename `self.enterprise_uuid` as `admin_enterprise_uuid`
`self.enterprise_uuid2` to `self.non_admin_enterprise_uuid`

HammadAhmadWaqas (approved):

tests/test_enterprise/api/test_views.py
muhammad-ammar: but we don't want to set the jwt cookie for the normal user in `test_view_with_normal_user`

PR #1658, opened 2022-10-21
feat: adding variables to the learner data model

kiram15:
[Jira Ticket 
](https://2u-internal.atlassian.net/browse/ENT-6427)
**Merge checklist:**

**Post merge:**


tests/test_integrated_channels/test_integrated_channel/test_exporters/test_learner_data.py
```diff
@@ -51,8 +51,8 @@ class TestLearnerExporter(unittest.TestCase):
     YESTERDAY_TIMESTAMP = NOW_TIMESTAMP - 24 * 60 * 60 * 1000
 
     def setUp(self):
-        self.user = factories.UserFactory(username='C3PO', id=1)
-        self.user_2 = factories.UserFactory(username='R2D2', id=2)
+        self.user = factories.UserFactory(username='C3PO', id=1, email='burneremail@aol.com')
```
johnnagro, line 54: nit: `example.com` email addresses when testing

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_learner_data.py
```diff
@@ -252,6 +252,7 @@ def test_learner_data_instructor_paced_no_certificate(
         assert learner_data[1].course_id == self.course_id
 
         for report in learner_data:
+            assert report.user_email == 'burneremail@aol.com'
```
johnnagro, line 255: nit: rather than `burneremail@aol.com` should we use something like `self.user_2.email` ?

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_learner_data.py
johnnagro, line 54: nit: `example.com` email addresses when testing

johnnagro (approved):
⛵

PR #1662, opened 2022-10-26
feat: adding admin views for all integrated channel learner data audit tables

alex-sheehan-edx:
Jira ticket: https://2u-internal.atlassian.net/browse/ENT-6459

work included:
- django admin views for all channels' learner data transmission audits
- read only field updates (including api_record fields on content metadata transmission audits)
- populating user_email for each individual channel's transmission audit during the learner data export
- made fields returned by learner data transmission audit API a need to need basis

image

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/api/v1/logs/serializers.py
+++ b/integrated_channels/api/v1/logs/serializers.py
@@ -64,7 +64,36 @@ class LearnerSyncStatusSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = LearnerDataTransmissionAudit
-        fields = '__all__'
+        fields = (
+            'user_email',
+            'content_title',
+            'progress_status',
+            'sync_status',
+            'sync_last_attempted_at',
+            'friendly_status_message',
+        )
+
+    sync_status = serializers.SerializerMethodField()
+    sync_last_attempted_at = serializers.SerializerMethodField()
+
+    def get_sync_status(self, obj):
+        """
+        Return a string representation of the sync status.
+        """
+        if obj.status is None:
+            sync_status = 'pending'
+        elif not obj.status.isdigit() or int(obj.status) >= 400:
+            sync_status = 'error'
+        elif int(obj.status) < 400:
+            sync_status = 'okay'
+
+        return sync_status
+
+    def get_sync_last_attempted_at(self, obj):
+        """
+        Return the most recent/youngest sync attempt date.
+        """
+        return obj.modified or obj.created
 
     @classmethod
     def get_class_by_channel_code(this_cls, channel_code):
@@ -85,7 +114,7 @@ class GenericLearnerSyncStatusSerializer(LearnerSyncStatusSerializer):
 
     class Meta:
         model = GenericLearnerDataTransmissionAudit
-        fields = '__all__'
+        fields = "__all__"
 
 
 class BlackboardLearnerSyncStatusSerializer(LearnerSyncStatusSerializer):
@@ -95,7 +124,7 @@ class BlackboardLearnerSyncStatusSerializer(LearnerSyncStatusSerializer):
 
     class Meta:
         model = BlackboardLearnerDataTransmissionAudit
-        fields = '__all__'
+        fields = "__all__"
 
 
 class CanvasLearnerSyncStatusSerializer(LearnerSyncStatusSerializer):
@@ -105,7 +134,7 @@ class CanvasLearnerSyncStatusSerializer(LearnerSyncStatusSerializer):
 
     class Meta:
         model = CanvasLearnerDataTransmissionAudit
-        fields = '__all__'
+        fields = "__all__"
 
 
 class CornerstoneLearnerSyncStatusSerializer(LearnerSyncStatusSerializer):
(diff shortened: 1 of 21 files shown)


integrated_channels/blackboard/admin/__init__.py
```diff
@@ -75,3 +78,43 @@ def customer_oauth_authorization_url(self, obj):
...
+        Returns: the name for the attached EnterpriseCustomer.
+        Args:
+            obj: The instance of BlackboardLearnerDataTransmissionAudit
+                being rendered with this admin form.
+        """
+        EnterpriseCustomer = get_enterprise_customer_model()
+        try:
+            ec = EnterpriseCustomer.objects.filter(
+                enterprise_customer_users__enterprise_enrollments=obj.enterprise_course_enrollment_id
+            ).first()
```
marlonkeating, line 118: Can we make this a helper method since it's used in multiple places, and its purpose is somewhat hard to grok at a glance?

integrated_channels/api/v1/logs/views.py
```diff
@@ -56,6 +56,7 @@ class LearnerSyncStatusViewSet(PermissionRequiredForIntegratedChannelMixin, view
     """
     Sync-status APIs for `LearnerDataTransmissionAudit` implementation items.
     """
+    # serializer_class = LearnerSyncStatusSerializer
```
kiram15: Is this an informational comment or a piece of code left in?

integrated_channels/api/v1/logs/serializers.py
```diff
@@ -66,6 +66,33 @@ class Meta:
         model = LearnerDataTransmissionAudit
         fields = '__all__'
 
+    sync_status = serializers.SerializerMethodField()
+    sync_last_attempted_at = serializers.SerializerMethodField()
+
+    def get_sync_status(self, obj):
```
marlonkeating, line 72: nit: Unit tests for all the branches here?

integrated_channels/api/v1/logs/serializers.py
```diff
@@ -66,6 +66,33 @@ class Meta:
...
+    def get_sync_last_attempted_at(self, obj):
+        """
+        Return the most recent/youngest sync attempt date.
+        """
+        date_list = [obj.created, obj.modified]
+        res = [i for i in date_list if i is not None]
+        if not res:
+            return None
+        else:
+            return max(res)
```
marlonkeating, line 94: nit: Is there a possibility that an object's created date would be later than its modified date?  If not, you could replace this with 
```suggestion
    return obj.modified or obj.created
```

integrated_channels/api/v1/logs/serializers.py
```diff
@@ -85,7 +112,14 @@ class GenericLearnerSyncStatusSerializer(LearnerSyncStatusSerializer):
...
         model = GenericLearnerDataTransmissionAudit
-        fields = '__all__'
+        fields = (
+            'user_email',
+            'content_title',
+            'progress_status',
+            'sync_status',
+            'sync_last_attempted_at',
+            'friendly_status_message',
+        )
```
marlonkeating, line 122: nit: Unless I'm missing something, it seems we're using the same fields for all the LMS providers.  Can we break this list out into a shared constant?

integrated_channels/moodle/exporters/learner_data.py
```diff
@@ -57,6 +57,7 @@ def get_learner_data_records(
             MoodleLearnerDataTransmissionAudit(
```
marlonkeating, line 57: nit: While we're here, maybe add in that comment about using 2 duplicate records, like the other providers?

kiram15 (approved):
Looked at admin locally and everything looks good! Just make sure to link the Jira ticket in the PR description and link the PR in the ticket :)

integrated_channels/api/v1/logs/views.py
alex-sheehan-edx: it's been removed 😄

integrated_channels/api/v1/logs/serializers.py
alex-sheehan-edx: 🤔 good idea!

integrated_channels/api/v1/logs/serializers.py
alex-sheehan-edx: hmmmmm yes I believe you're right!

integrated_channels/blackboard/admin/__init__.py
alex-sheehan-edx: it adds clarity and if you're unfamiliar with the django ORM this can be unapproachable so I'm fine with this 👍

integrated_channels/api/v1/logs/serializers.py
alex-sheehan-edx: we can add these fields to the base `LearnerDataTransmissionAudit` serializer!

integrated_channels/blackboard/admin/__init__.py
```diff
@@ -75,3 +77,38 @@ def customer_oauth_authorization_url(self, obj):
...
+        "content_title",
+        "enterprise_customer_name",
+    )
+
+    list_per_page = 1000
+
+    class Meta:
+        model = BlackboardLearnerDataTransmissionAudit
+
+    def enterprise_customer_name(self, obj):
```
marlonkeating, line 106: This method is replicated across the AuditAdmin classes.  Could we put this method into a base class?

marlonkeating (approved):

PR #1665, opened 2022-11-01
feat: integrated channels customer configs soft delete support

alex-sheehan-edx:
Jira: https://2u-internal.atlassian.net/browse/ENT-6442

**Merge checklist:**

**Post merge:**


integrated_channels/blackboard/models.py
```diff
@@ -96,8 +96,6 @@ def __repr__(self):
 class BlackboardEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguration):
     """
     The Enterprise-specific configuration we need for integrating with Blackboard.
-
```
alex-sheehan-edx, line 99: while poking around I noticed that this is here when the table in fact __does__ contain pii. Not that it really matters as the system that would handle interpreting the docstring is not in use

integrated_channels/integrated_channel/models.py
```diff
@@ -123,6 +176,17 @@ def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
         models.signals.post_save.connect(set_default_display_name, sender=cls)
 
+    def delete(self):
```
alex-sheehan-edx, line 179: very weird behavior here related to how Django handles [inheritance and model managers](https://docs.djangoproject.com/en/4.1/topics/db/managers/). But I couldn't get subclassed channel configs to use the custom manager's delete override. This means that when querying `BlackboardEnterpriseCustomerConfig`, you can use `objects` and `all_objects` but when calling `delete()` on a record, it would grab Django's base `Model` class' usage of `delete` instead of the one defined in the soft delete manager. To remedy this I just overrode/added the three required soft delete methods to the base config.

tests/test_integrated_channels/test_models.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        Test that the all integration customer configs support soft delete
+        """
+        # Assert we have something to work with
+        assert len(channel_config.objects.all()) == 1
+
+        # Soft delete
+        existing_config = channel_config.objects.first()
+        existing_config.delete()
+        assert not channel_config.objects.all()
```
kiram15, line 52: Maybe ensure here that deleted_at gets set as well? And then cleared when you revive it?

tests/test_integrated_channels/test_models.py
```diff
@@ -0,0 +1,59 @@
...
+    @ddt.data(
+        BlackboardEnterpriseCustomerConfiguration,
+        CanvasEnterpriseCustomerConfiguration,
+        CornerstoneEnterpriseCustomerConfiguration,
+        DegreedEnterpriseCustomerConfiguration,
+        Degreed2EnterpriseCustomerConfiguration,
+        MoodleEnterpriseCustomerConfiguration,
+        SAPSuccessFactorsEnterpriseCustomerConfiguration,
+    )
+    def test_integration_customer_config_soft_delete(self, channel_config):
```
kiram15, line 42: I would include a test of the hard delete as well

integrated_channels/sap_success_factors/migrations/0009_sapsuccessfactorsenterprisecustomerconfiguration_deleted_at.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('sap_success_factors', '0008_auto_20221021_0159'),
+    ]
+
+    operations = [
+        migrations.AddField(
```
kiram15, line 13: Does this one not have a historical model?

kiram15 (commented):
Couple questions: Should soft deleted configs show up in the api? Also, how are we going to surface the "revive" function? Will it be available to customers or just to staff?

alex-sheehan-edx:
> Should soft deleted configs show up in the api?

I don't think so, there isn't a real need to surface deleted configs and would probably just confuse consumers of the view.

alex-sheehan-edx:
> Also, how are we going to surface the "revive" function?

My understanding is that it's only available to us, the staff. From the user's perspective the config is deleted and long gone. We could but that increases the scope IMO.

tests/test_integrated_channels/test_models.py
alex-sheehan-edx: I would make the argument that the queryset not finding and then finding the object (under the `all_objects` manager) is sufficient but also happy to add more assertions

integrated_channels/sap_success_factors/migrations/0009_sapsuccessfactorsenterprisecustomerconfiguration_deleted_at.py
alex-sheehan-edx: yeah, I'm forgetting why exactly but I do remember talking about that before... I'm pretty sure there's a reason? 😬

kiram15 (approved):
Looks good!

PR #1667, opened 2022-11-04
fix: Enterprise Coupons require comm worker to be a universal ent operator

grmartin:
* fix: Updating `seed_enterprise_devstack_data` Django Admin command to include `ecommerce_worker` in its list of service workers.


iloveagent57 (approved):

alex-sheehan-edx (approved):
lgtm!

PR #1669, opened 2022-11-07
fix: catch update transmissions that failed and need to be resent

alfredoguillem:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/transmitters/content_metadata.py
+++ b/integrated_channels/integrated_channel/transmitters/content_metadata.py
@@ -205,6 +205,7 @@ class ContentMetadataTransmitter(Transmitter):
                     elif action_name == 'delete':
                         transmission.remote_deleted_at = action_happened_at
                     transmission.save()
-                    transmission.remove_marked_for()
+                    if transmission.api_response_status_code < 300:
+                        transmission.remove_marked_for()
                     results.append(transmission)
         return results
--- a/tests/test_integrated_channels/test_integrated_channel/test_exporters/test_content_metadata.py
+++ b/tests/test_integrated_channels/test_integrated_channel/test_exporters/test_content_metadata.py
@@ -353,3 +353,52 @@ class TestContentMetadataExporter(unittest.TestCase, EnterpriseMockMixin):
         # We should make a call for the enterprise catalog, but no call for the metadata because no update needed
         assert mock_ent_catalog_api.return_value.get_content_metadata.call_count == 0
         assert mock_ent_catalog_api.return_value.get_catalog_diff.call_count == 1
+
+    def test__check_matched_content_updated_at_incomplete_transmission(self):
+        """
+        Test the __check_matched_content_updated_at function when the transmission is incomplete.
+        """
+        past_transmission_updated_at_after_mock = factories.ContentMetadataItemTransmissionFactory(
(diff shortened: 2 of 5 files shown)


integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -205,6 +205,7 @@ def _transmit_action(self, content_metadata_item_map, client_method, action_name
                     elif action_name == 'delete':
                         transmission.remote_deleted_at = action_happened_at
                     transmission.save()
-                    transmission.remove_marked_for()
+                    if transmission.api_response_status_code < 400:
```
alex-sheehan-edx, line 208: I would do `< 300` since 300's are technically considered errors

alex-sheehan-edx (commented):
Can we get a test of `_check_matched_content_updated_at` please? 😬

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_content_metadata.py
```diff
@@ -353,3 +353,43 @@ def test_export_fetches_content_only_when_update_needed(self, mock_ent_catalog_a
         # We should make a call for the enterprise catalog, but no call for the metadata because no update needed
         assert mock_ent_catalog_api.return_value.get_content_metadata.call_count == 0
         assert mock_ent_catalog_api.return_value.get_catalog_diff.call_count == 1
+
+    def test__check_matched_content_updated_at_incomplete_transmission(self):
+        """
+        Test the __check_matched_content_updated_at function when the transmission is incomplete.
+        """
+        past_transmission = factories.ContentMetadataItemTransmissionFactory(
```
alex-sheehan-edx, line 361: Since this generated transmission audit item has a `remote_updated_at` value that comes **after** your `mock_matched_items` updated at values, it shouldn't be found by `_check_matched_content_updated_at`. I'd love to have an additional audit item (or a separate test that has an audit item!) that _does_ get found because of it's update time (and not because it has a `marked_for` value. That way we'd know the `_check_matched_content_updated_at` method:
1) finds items that have a `marked_value`
2) finds items that have update times that come before the `matched_items` update time
3) does not find items that do not meet either of these criteria

alex-sheehan-edx (commented):
Awesome start! I think you're pretty close, though I have a couple comments and improvements that can be made. Let me know if you have any questions, I'm happy to chat about anything I've commented on.

tests/test_integrated_channels/test_integrated_channel/test_exporters/test_content_metadata.py
alfredoguillem: added a 3rd transmission, following the 3 points above we would only find 2 out of the 3 transmissions:

`past_transmission_updated_at_before_mock` and `past_transmission_updated_at_marked_for`

nedbat:
FWIW: the "chore" commit type is meant for work that could be automated, like "make upgrade."  This looks like a clear "fix" commit.

alex-sheehan-edx (approved):
Amazing! 😄 Awesome code.

alex-sheehan-edx:
Just make sure to bump version and change log before you merge!

bcitro (approved):

PR #1672, opened 2022-11-14
feat: adding new vars

kiram15:
**Merge checklist:**

**Post merge:**


integrated_channels/utils.py
```diff
@@ -460,3 +460,102 @@ def get_enterprise_customer_from_enterprise_enrollment(enrollment_id):
...
+    # for each config, then check what is the most recent time stamp associated
+    # if we pass in 'error' as True, we are looking for only error audit records
+    for learner_audit in learner_audits:
+        is_error_audit = learner_audit.status.isdigit() or int(learner_audit.status) >= 400
+        if (not error or (error and is_error_audit)):
+            date_list = [learner_audit.remote_created_at, learner_audit.remote_updated_at,
+                         learner_audit.remote_deleted_at]
+            res = [i for i in date_list if i is not None]
+            if res:
+                max_sync_time = max(max_sync_time, max(res))
```
marlonkeating, line 528: nit: Probably don't need a new variable here
```suggestion
            date_list = [learner_audit.remote_created_at, learner_audit.remote_updated_at,
                         learner_audit.remote_deleted_at]
            date_list = [i for i in date_list if i is not None]
            if date_list:
                max_sync_time = max(max_sync_time, max(res))
```

PR #1673, opened 2022-11-15
feat: Add health check for canvas integrated channels

marlonkeating:
[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-5607)

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/api/v1/urls.py
+++ b/integrated_channels/api/v1/urls.py
@@ -4,7 +4,7 @@ URL definitions for v1 Integrated Channel API endpoints.
 
 from django.urls import include, path
 
-from .views import IntegratedChannelsBaseViewSet
+from .views import IntegratedChannelHealthCheckView, IntegratedChannelsBaseViewSet
 
 app_name = 'v1'
 urlpatterns = [
@@ -15,6 +15,7 @@ urlpatterns = [
     path('degreed/', include('integrated_channels.api.v1.degreed.urls')),
     path('degreed2/', include('integrated_channels.api.v1.degreed2.urls')),
     path('cornerstone/', include('integrated_channels.api.v1.cornerstone.urls')),
+    path('configs/health-check', IntegratedChannelHealthCheckView.as_view(), name='health_check'),
     path('configs/', IntegratedChannelsBaseViewSet.as_view({'get': 'list'}), name='configs'),
     path('logs/', include('integrated_channels.api.v1.logs.urls')),
 ]
--- a/integrated_channels/api/v1/views.py
+++ b/integrated_channels/api/v1/views.py
@@ -1,7 +1,9 @@
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
-from rest_framework import permissions, viewsets
+from rest_framework import generics, permissions, viewsets
 from rest_framework.authentication import SessionAuthentication
+from rest_framework.renderers import JSONRenderer
 from rest_framework.response import Response
+from rest_framework.status import HTTP_200_OK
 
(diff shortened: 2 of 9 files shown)


johnnagro:
I really like the idea of wiring this up to an http endpoint. I wonder if we should standardize around the more generic `/config` URLs from here https://github.com/openedx/edx-enterprise/blob/master/integrated_channels/api/v1/urls.py#L18

maybe '/config/<uuid>/health` or something?

I think we should discuss standardizing on '/config' for more things?

integrated_channels/api/v1/urls.py
```diff
@@ -2,9 +2,9 @@
 URL definitions for v1 Integrated Channel API endpoints.
 """
 
-from django.urls import include, path
+from django.urls import include, path, re_path
```
alex-sheehan-edx, line 5: I don't think re_path is needed anymore

integrated_channels/api/v1/views.py
```diff
@@ -78,3 +83,45 @@ def list(self, request, *arg, **kwargs):
         response = blackboard_serializer.data + canvas_serializer.data + cornerstone_serializer.data + \
             degreed_serializer.data + degreed2_serializer.data + moodle_serializer.data + sap_serializer.data
         return Response(response)
+
+
+class IntegratedChannelHealthCheckView(generics.ListAPIView):
+    """
+        **Use Cases**
```
alex-sheehan-edx, line 90: great docstring 😄

integrated_channels/api/v1/views.py
```diff
@@ -78,3 +83,45 @@ def list(self, request, *arg, **kwargs):
...
+
+    @method_decorator(require_at_least_one_query_parameter('channel_code'))
+    @method_decorator(require_at_least_one_query_parameter('uuid'))
+    def get(self, request, *args, **kwargs):
+        channel_code = self.request.query_params.get('channel_code')
+        uuid = self.request.query_params.get('uuid')
+
+        # Get enterprise configuration
+        config_type = EnterpriseCustomerPluginConfiguration.get_class_by_channel_code(channel_code)
+        config = config_type.objects.get(uuid=uuid)
```
alex-sheehan-edx, line 120: some error handling around channel and config do not exist would be nice here

integrated_channels/api/v1/views.py
```diff
@@ -78,3 +83,45 @@ def list(self, request, *arg, **kwargs):
...
+        client_type = get_enterprise_client_by_channel_code(channel_code)
+        client = client_type(config)
+
+        health_status = client.health_check()
+        is_healthy = health_status == IntegratedChannelHealthStatus.HEALTHY
+        payload = {'is_healthy': is_healthy, 'health_status': health_status.value}
+        return Response(payload, status=HTTP_200_OK)
```
alex-sheehan-edx, line 127: I think the linter will complain about no new lines?

integrated_channels/api/v1/views.py
alex-sheehan-edx: if we're going to release this as is then also probably a list of supported channels with the validation?

alex-sheehan-edx (commented):
A couple comments and we should probably have some tests to along with the new views. But otherwise I think your approach is great and this is looking good.

integrated_channels/api/v1/urls.py
marlonkeating: Good catch, thanks!

alex-sheehan-edx (approved):
Superb! ✅

PR #1680, opened 2022-11-22
feat: adding in sync time to base config

kiram15:
**Merge checklist:**

**Post merge:**


integrated_channels/integrated_channel/models.py
```diff
@@ -205,6 +206,76 @@ def clean(self):
                 }
             )
 
+    def get_learner_data_audit_model(self):
+        return LearnerDataTransmissionAudit.get_completion_class_by_channel_code(self.channel_code())
+
+    def get_last_learner(self, error, config_id):
```
johnnagro, line 212: this is an instance method so, you don't need to pass in config_id, its already a variable available as `self.id`

nit: i'd maybe pick a clearer name - like `get_last_learner_synced_at` (maybe renamed `error` to `include_only_errors` or something like that)

integrated_channels/integrated_channel/models.py
```diff
@@ -205,6 +206,76 @@ def clean(self):
...
+        """
+        Gets every learner audit associated with the configuration and finds the most recent record.
```
johnnagro, line 214: this docstring should say that it finds the most recent _timestamp_ (or `None`) because thats what it returns - the current docstring makes it seem like you'll return the learner audit record most recently modified

integrated_channels/integrated_channel/models.py
```diff
@@ -205,6 +206,76 @@ def clean(self):
...
+        if error:
+            learner_audits = audit_model.objects.filter(
+                enterprise_customer_uuid=self.enterprise_customer.uuid,
+                plugin_configuration_id=config_id,
+                status__gte=400,
+            )
+        else:
+            learner_audits = LearnerDataTransmissionAudit.objects.filter(
+                enterprise_customer_uuid=self.enterprise_customer.uuid,
+                plugin_configuration_id=config_id,
```
johnnagro, line 228: this can be `self.id` rather than a passed `config_id`

integrated_channels/integrated_channel/models.py
```diff
@@ -205,6 +206,76 @@ def clean(self):
...
+            )
+        if learner_audits:
+            return learner_audits.aggregate(max=Max('completed_timestamp')).get('max')
+        else:
+            return None
+
+    def get_last_content(self, error, config_id):
```
johnnagro, line 235: similarly to the learner version of this method, you dont need to pass a `config_id` here. use the instance `self.id`.

nit: i'd consider tweakin the method name and docstring to be more clear as to what youre returning. also consider making `error` like `include_only_errors` or something

integrated_channels/integrated_channel/models.py
```diff
@@ -205,6 +206,76 @@ def clean(self):
...
+            content_audits = ContentMetadataItemTransmission.objects.filter(
+                enterprise_customer=self.enterprise_customer,
+                plugin_configuration_id=config_id,
+                integrated_channel_code=self.channel_code(),
+            )
+        if content_audits:
+            max_created = (content_audits.aggregate(max_created=Max('remote_created_at'))).get('max_created')
+            max_updated = (content_audits.aggregate(max_updated=Max('remote_updated_at'))).get('max_updated')
+            max_deleted = (content_audits.aggregate(max_deleted=Max('remote_deleted_at'))).get('max_deleted')
+            return max(x for x in [max_created, max_updated, max_deleted] if x is not None)
```
johnnagro, line 258: you may need to check to see if they're all None and return None in that case because I believe `max()` will error if the array ends up being all `None`


```
>>> max(x for x in [None, None, None] if x is not None)
  File "<console>", line 1, in <module>
ValueError: max() arg is an empty sequence
```

integrated_channels/integrated_channel/models.py
```diff
@@ -205,6 +206,76 @@ def clean(self):
...
+        else:
+            return None
+
+    def get_last_sync(self, error, config_id):
```
johnnagro, line 262: is this used anywhere?

johnnagro (changes requested):
pretty good - i made a couple of suggestions i think we should consider before merging.

integrated_channels/integrated_channel/models.py
kiram15: No it's not used anywhere as of now, because we are not returning the learner stuff atm, I'm mostly leaving it in so we don't have to rewrite it later

integrated_channels/integrated_channel/models.py
```diff
@@ -205,6 +206,88 @@ def clean(self):
...
+                plugin_configuration_id=self.id,
+                status__gte=400,
+            )
+        else:
+            learner_audits = audit_model.objects.filter(
+                enterprise_customer_uuid=self.enterprise_customer.uuid,
+                plugin_configuration_id=self.id,
+            )
+
+        print("HELLO CUTIE")
```
johnnagro, line 231: some debug statements we can remove? 😆

johnnagro (approved):
⛵

PR #1690, opened 2022-12-14
fix: cleaning up integrated channels inheritance column types

johnnagro:
## Description

- in fixing the inheritance of the `LearnerDataTransmissionAudit` records, we found places where the `completed_timestamp` column type deviated
- i renamed the non-standard columns and added in a new standard column
- long term we dont need these non-standard columns, we can format the dates at serialization time, save this work for the future
- a data migration should not be required, the existing columns moved over (for the integrations)
- ENT-6584


kiram15 (commented):
This all looks good, but I think there might still be some examples of the old completed_timestamp format in other files. It's not the biggest deal, but also they're easy fixes (or ways that we should implement a factory). EX: 
https://github.com/search?q=repo%3Aopenedx%2Fedx-platform%20completed_timestamp&type=code
https://github.com/openedx/edx-enterprise/blob/52a2cdfe1a3beb57960d2f6c0dc5ccb98246139a/integrated_channels/blackboard/client.py#L266
https://github.com/openedx/edx-enterprise/blob/52a2cdfe1a3beb57960d2f6c0dc5ccb98246139a/tests/test_integrated_channels/test_sap_success_factors/test_client.py#L85

johnnagro:
> This all looks good, but I think there might still be some examples of the old completed_timestamp format in other files. It's not the biggest deal, but also they're easy fixes (or ways that we should implement a factory). EX: https://github.com/search?q=repo%3Aopenedx%2Fedx-platform%20completed_timestamp&type=code
> 
> https://github.com/openedx/edx-enterprise/blob/52a2cdfe1a3beb57960d2f6c0dc5ccb98246139a/integrated_channels/blackboard/client.py#L266
> 
> https://github.com/openedx/edx-enterprise/blob/52a2cdfe1a3beb57960d2f6c0dc5ccb98246139a/tests/test_integrated_channels/test_sap_success_factors/test_client.py#L85

The blackboard example is documenting the payload blackboard expects, that payload is setup here:

https://github.com/openedx/edx-enterprise/blob/master/integrated_channels/blackboard/models.py#L377

and it now uses the new column. The SAP example is a similar situation.

alex-sheehan-edx (approved):
looks good to me

PR #1693, opened 2022-12-20
fix: adding an index for performance, testing syntax for no-lock

johnnagro:
## Description

- an `index_together` is needed on the `LearnerDataTransmissionAudit` tables for performance
- we want to add the index in a non-blocking/non-locking way
- we want to use MySQL-specific syntax but allow tests to still use `sqlite3`


integrated_channels/blackboard/migrations/0013_alter_blackboardlearnerdatatransmissionaudit_index_together.py
```diff
@@ -0,0 +1,41 @@
+# Generated by Django 3.2.15 on 2022-12-20 15:05
+
+from django.db import connection, migrations
+
+
+class Migration(migrations.Migration):
+    atomic = False
```
johnnagro, line 7: don't wrap this SQL statement in a transaction, which will add locking.

integrated_channels/blackboard/migrations/0013_alter_blackboardlearnerdatatransmissionaudit_index_together.py
```diff
@@ -0,0 +1,41 @@
...
+
+    dependencies = [
+        ('blackboard', '0012_move_and_recrete_completed_timestamp'),
+    ]
+
+    db_engine = connection.settings_dict['ENGINE']
+    if 'sqlite3' in db_engine:
```
johnnagro, line 14: for `sqlite3` (our test env) run the normal migration

integrated_channels/blackboard/migrations/0013_alter_blackboardlearnerdatatransmissionaudit_index_together.py
```diff
@@ -0,0 +1,41 @@
...
+        operations = [
+            migrations.AlterIndexTogether(
+                name='blackboardlearnerdatatransmissionaudit',
+                index_together={('enterprise_customer_uuid', 'plugin_configuration_id')},
+            ),
+        ]
+    else:
+        operations = [
+            migrations.SeparateDatabaseAndState(
```
johnnagro, line 23: for our stage/prod env - apply the normal migration to the state, but on the DB run our custom sql.

integrated_channels/blackboard/migrations/0013_alter_blackboardlearnerdatatransmissionaudit_index_together.py
```diff
@@ -0,0 +1,41 @@
...
+                    migrations.AlterIndexTogether(
+                        name='blackboardlearnerdatatransmissionaudit',
+                        index_together={('enterprise_customer_uuid', 'plugin_configuration_id')},
+                    ),
+                ],
+                database_operations=[
+                    migrations.RunSQL(sql="""
+                        CREATE INDEX "blackboard_blackboardlearnerdatatransmissionaudit_enterprise_customer_uuid_plugin_configuration_id_85936b55_idx"
+                        ON "blackboard_blackboardlearnerdatatransmissionaudit" ("enterprise_customer_uuid", "plugin_configuration_id")
+                        ALGORITHM=INPLACE LOCK=NONE
```
johnnagro, line 34: the syntax `ALGORITHM=INPLACE LOCK=NONE` avoids a lock

iloveagent57: maybe add a link to some docs near this?

iloveagent57 (approved):
Looks good, don't forget version number etc.  And maybe a link to docs about the `ALGORITHM` and `LOCK` thing.

integrated_channels/blackboard/migrations/0013_alter_blackboardlearnerdatatransmissionaudit_index_together.py
johnnagro: will do, i am going to pop something into the wiki somewhere

johnnagro: https://dev.mysql.com/doc/refman/5.7/en/innodb-online-ddl-operations.html

PR #1697, opened 2023-01-03
feat: non-blocking indexing on LDTA tables

johnnagro:
## Description

- changes piloted in https://github.com/openedx/edx-enterprise/pull/1693
- needed for performance fixes for ENT-6584


integrated_channels/cornerstone/models.py
```diff
@@ -236,7 +236,9 @@ class CornerstoneLearnerDataTransmissionAudit(LearnerDataTransmissionAudit):
 
     class Meta:
         app_label = 'cornerstone'
+        # XXX non-standard
```
alex-sheehan-edx, line 239: is the non standard comment referring to the user and course id unique key?

alex-sheehan-edx: if so, is it worth adding a line about why we make sure to have one and only one pair between user and course

integrated_channels/moodle/migrations/0023_alter_moodlelearnerdatatransmissionaudit_index_together.py
```diff
@@ -0,0 +1,40 @@
...
+                        index_together={('enterprise_customer_uuid', 'plugin_configuration_id')},
+                    ),
+                ],
+                database_operations=[
+                    migrations.RunSQL(sql="""
+                        CREATE INDEX moodle_mldta_85936b55_idx
+                        ON moodle_moodlelearnerdatatransmissionaudit (enterprise_customer_uuid, plugin_configuration_id)
+                        ALGORITHM=INPLACE LOCK=NONE
+                    """, reverse_sql="""
+                        DROP INDEX moodle_dldta_85936b55_idx
```
alex-sheehan-edx, line 35: these two index names don't match between the forward and backward migrations, is that expected?

alex-sheehan-edx (commented):
A couple of questions! Otherwise looking good.

integrated_channels/cornerstone/models.py
johnnagro: I dont know why we do that, i only know that we have that constraint and its non-standard so, in theory, there is the potential we could do something that processes the abstract class and run into a constraint issue.

integrated_channels/moodle/migrations/0023_alter_moodlelearnerdatatransmissionaudit_index_together.py
johnnagro: good catch, thanks.

integrated_channels/cornerstone/models.py
alex-sheehan-edx: it's because we pre-create the audit at the time that the student comes over to our system from cornerstone and we need a way of finding that audit at the time of grade reporting IIRC

alex-sheehan-edx: Altho it's not immediately obvious to me why that means we need a unique key. Maybe just to safe guard against repeated access from the student. IE if the learner comes over via external link a second time+.

alex-sheehan-edx: anyhoo, not necessary to add to the comment just thought it'd be good context for other people looking at the code

alex-sheehan-edx (approved):
lgtm, don't forget version bump 😄

PR #1700, opened 2023-01-04
fix: cleaning up serializer field inheritance

johnnagro:
## Description

- fixes https://2u-internal.atlassian.net/browse/ENT-6635
- standard, common fields for serialization are defined in the parent serializer but were not consistently being incorporated in the individual implementations
- one side effect was that the `active` field was missing from most of the implementations, causing the UI to show most configurations as disabled
- cleaned up formatting


integrated_channels/api/serializers.py
```diff
@@ -23,15 +25,18 @@ class Meta:
         fields = (
             'id',
             'display_name',
+            'channel_code',
```
johnnagro, line 28: these additions were in most/all of the implementations, so moving it to the common list

integrated_channels/api/v1/blackboard/serializers.py
```diff
@@ -12,24 +12,18 @@
...
     class Meta:
         model = BlackboardEnterpriseCustomerConfiguration
-        fields = ('id', 'client_id', 'client_secret', 'blackboard_base_url',
-                  'refresh_token', 'transmission_chunk_size', 'uuid', 'enterprise_customer',
-                  'oauth_authorization_url', 'is_valid', 'channel_code', 'active',
-                  'last_sync_attempted_at', 'last_content_sync_attempted_at',
-                  'last_learner_sync_attempted_at', 'last_sync_errored_at',
-                  'last_content_sync_errored_at', 'last_learner_sync_errored_at')
-
+        extra_fields = (
```
johnnagro, line 18: field-per-line is easier to read, better diffs

integrated_channels/integrated_channel/models.py
```diff
@@ -282,6 +282,19 @@ def get_last_sync(self, include_only_errors):
         else:
             return None
 
+    @property
+    def is_valid(self):
```
johnnagro, line 286: this should be a commonly defined method that is overridden. opportunities to clean this up but per our discussion in slack, too complicated to clean up now.

marlonkeating (approved):
Good cleanup!

PR #1701, opened 2023-01-05
fix: memoize synced_at datimes on config record

johnnagro:
## Description

- https://2u-internal.atlassian.net/browse/ENT-6584
- still seeing performance issues on config endpoints, query is complicated to index for properly
- lets memoize rather than query for most recent sync attempts/errors


integrated_channels/api/serializers.py
```diff
@@ -38,39 +38,3 @@ class Meta:
             'last_learner_sync_errored_at',
             'transmission_chunk_size',
         )
-
-    def get_last_sync_attempted_at(self, obj):
```
johnnagro, line 42: using model attributes now, rather than query methods

tests/test_integrated_channels/test_integrated_channel/test_models.py
```diff
@@ -174,3 +174,71 @@ def test_failed_incomplete_create_transmissions_getter(self):
...
+        first_timestamp = datetime.datetime.fromtimestamp(1400000000)
+        self.config.update_learner_synced_at(first_timestamp, True)
+        assert self.config.last_sync_attempted_at == first_timestamp
+        assert self.config.last_content_sync_attempted_at is None
+        assert self.config.last_learner_sync_attempted_at == first_timestamp
+        assert self.config.last_sync_errored_at is None
+        assert self.config.last_content_sync_errored_at is None
+        assert self.config.last_learner_sync_errored_at is None
+
+        second_timestamp = datetime.datetime.fromtimestamp(1500000000)
```
kiram15, line 232: Maybe add one test where both learner and content are populated but with different time stamps? So we can check that the `last_sync_attempted_at` and the `last_sync_errored_at` work correctly at identifying the latest of the two

kiram15 (approved):
LGTM, just one testing consideration!

PR #1702, opened 2023-01-10
fix: handle file compression check and test

jajjibhai008:
**Description**
Add file compression option on Reporting Form of frontend-admin-panel. Initially, that option was only enabled by the Django admin, not from the user request that came from the admin panel. But now users will be able to Enable compression from UI.

I apply validations on enable_compression by making a static function of compression_validation in models and using it in the serializer to make it secure.

Here is the picture of adding a button on the UI.

image

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -27,6 +27,7 @@ from enterprise.models import (
     AdminNotification,
     AdminNotificationRead,
     EnterpriseCustomerIdentityProvider,
+    EnterpriseCustomerReportingConfiguration,
     EnterpriseCustomerUser,
     SystemWideEnterpriseUserRoleAssignment,
 )
@@ -831,6 +832,14 @@ class EnterpriseCustomerReportingConfigurationSerializer(serializers.ModelSerial
         return value
 
     def validate(self, data):  # pylint: disable=arguments-renamed
+        error = EnterpriseCustomerReportingConfiguration.validate_compression(
+            data.get('enable_compression'),
+            data.get('data_type'),
+            data.get('delivery_method')
+        )
+        if error:
+            raise serializers.ValidationError(error)
+
         delivery_method = data.get('delivery_method')
         if not delivery_method and self.instance:
             delivery_method = self.instance.delivery_method
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -2733,13 +2733,36 @@ class EnterpriseCustomerReportingConfiguration(TimeStampedModel):
         """
         return self.__str__()
 
+    @classmethod
+    def validate_compression(cls, enable_compression, data_type, delivery_method):
+        """
+        Check enable_compression flag is set as expected
+
+        Arguments:
+            enable_compression (bool): file copression flag
+            data_type (str): report type
+            delivery_method (str): delivery method for sending files
+
+        Returns:
+            (dict): Validation Error
+        """
+        if data_type in cls.ALLOWED_NON_COMPRESSION_DATA_TYPES and delivery_method == cls.DELIVERY_METHOD_SFTP:
+            return {}
+        elif not enable_compression:
+            allowed_data_types = ", ".join(cls.ALLOWED_NON_COMPRESSION_DATA_TYPES)
+            error_message = (
+                f'Compression can only be disabled for the following data types: {allowed_data_types} and '
+                f'delivery method: {cls.DELIVERY_METHOD_SFTP}'
+            )
+            return {'enable_compression': error_message}
+        return {}
+
     def clean(self):
         """
         Override of clean method to perform additional validation on frequency, day_of_month/day_of week
         and compression.
         """
         validation_errors = {}
-
         # Check that the frequency selections make sense.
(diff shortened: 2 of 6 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -779,6 +780,7 @@ class Meta:
         default=[],
         child=serializers.EmailField()
     )
+    enable_compression = serializers.BooleanField()
```
saleem-latif, line 783: Why do we need to add `enable_compression` here? wouldn't this be added automatically by DRF?

enterprise/models.py
```diff
@@ -2733,13 +2733,38 @@ def __repr__(self):
...
+        Args:
+            enable_compression (bool): file copression flag
+            data_type (string): report type
+            delivery_method (string): delivery method for sending files
+
+        Returns:
+           Validation Error
+        """
+
+        if not enable_compression and (
```
saleem-latif, line 2749: can we simplify the condition here? I think following will be more readable.
```
if data_type in cls.ALLOWED_NON_COMPRESSION_DATA_TYPES or delivery_method == cls.DELIVERY_METHOD_SFTP:
    # Compression not needed and the can be disabled by the user.
    return
elif not enable_compression:
    allowed_data_types = ", ".join(cls.ALLOWED_NON_COMPRESSION_DATA_TYPES)
    error_message = (
        f'Compression can only be disabled for the following data types: {allowed_data_types} and '
         f'delivery method: {cls.DELIVERY_METHOD_SFTP}'
    )
    return {'enable_compression': error_message}
```

enterprise/models.py
```diff
@@ -2733,13 +2733,38 @@ def __repr__(self):
         """
         return self.__str__()
 
+    @classmethod
+    def validate_compression(cls, enable_compression, data_type, delivery_method):
+        """Check enable_compression flag is set as expected
```
saleem-latif, line 2738: Please update doc string format according to https://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html#docstrings

enterprise/models.py
```diff
@@ -2733,13 +2733,38 @@ def __repr__(self):
...
+        return {}
+
     def clean(self):
         """
         Override of clean method to perform additional validation on frequency, day_of_month/day_of week
         and compression.
         """
         validation_errors = {}
-
+        print('enable compression ki value', self.enable_compression)
```
saleem-latif, line 2767: nit: Please remove print statement.

saleem-latif (changes requested):

enterprise/api/v1/serializers.py
jajjibhai008: yes, you are right. I will update this.

enterprise/models.py
jajjibhai008: This looks good to me but there needs some changes in logic.
```
if data_type in cls.ALLOWED_NON_COMPRESSION_DATA_TYPES and delivery_method == cls.DELIVERY_METHOD_SFTP:
    # Compression not needed and the can be disabled by the user.
    return {}
elif not enable_compression:
    allowed_data_types = ", ".join(cls.ALLOWED_NON_COMPRESSION_DATA_TYPES)
    error_message = (
        f'Compression can only be disabled for the following data types: {allowed_data_types} and '
         f'delivery method: {cls.DELIVERY_METHOD_SFTP}'
    )
    return {'enable_compression': error_message}
return {}
```

saleem-latif (approved):

tests/test_enterprise/api/test_serializers.py
```diff
@@ -337,3 +338,8 @@ def test_create(self):
         self.data['pgp_encryption_key'] = 'invalid-key'
         serializer = EnterpriseCustomerReportingConfigurationSerializer(data=self.data)
         assert not serializer.is_valid()
+
+        # Valid Compression check should be flagged.
+        self.data['enable_compression'] = False
+        serializer = EnterpriseCustomerReportingConfigurationSerializer(data=self.data)
+        assert not serializer.is_valid()
```
HammadAhmadWaqas, line 345: Check error message here, It should have expected error message.

enterprise/models.py
```diff
@@ -2733,13 +2733,36 @@ def __repr__(self):
         """
         return self.__str__()
 
+    @classmethod
+    def validate_compression(cls, enable_compression, data_type, delivery_method):
```
HammadAhmadWaqas, line 2737: Add a separate test for this method and pass different scenarios using `ddt`.

HammadAhmadWaqas (changes requested):

enterprise/models.py
jajjibhai008: Already a test case exists in test_models.py for this method.

tests/test_enterprise/api/test_serializers.py
jajjibhai008: Good idea to make it more efficient.

HammadAhmadWaqas (approved):

PR #1708, opened 2023-01-24
fix: applied validatoins on report delivery_method

jajjibhai008:
***Description***
when someone creates an EnterpriseCustomerReportingConfiguration (ECRC) using the delivery_method email and tries to update the delivery_method to sftp then sftp password field is not shown on the Django admin form. Due to which 500 error occurred. 
***Solution***
I apply validation on delivery_method is that when someone tries to update this field either from the Django admin or through the front end then a proper message will be thrown to the user that you can not update the delivery_method.

Here is the ticket [ENT-1689](https://2u-internal.atlassian.net/browse/ENT-1689)
…ort in forms

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -841,6 +841,15 @@ class EnterpriseCustomerReportingConfigurationSerializer(serializers.ModelSerial
             raise serializers.ValidationError(error)
 
         delivery_method = data.get('delivery_method')
+        create_report = data.get('uuid') is not None
+        if create_report:
+            delivery_method_error = self.instance.validate_delivery_method(
+                data.get('uuid'),
+                delivery_method
+            )
+            if delivery_method_error:
+                raise serializers.ValidationError(delivery_method_error)
+
         if not delivery_method and self.instance:
             delivery_method = self.instance.delivery_method
 
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -2757,6 +2757,23 @@ class EnterpriseCustomerReportingConfiguration(TimeStampedModel):
             return {'enable_compression': error_message}
         return {}
 
+    def validate_delivery_method(self, create_report, delivery_method):
+        """
+        Check delivery_method is changed or not while updating report
+
+        Arguments:
+            create_report (str): report uuid
+            delivery_method (str): selected delivery method
+
+        Returns:
+            (dict): Validation Error
+        """
+        if create_report:
+            if self.delivery_method != delivery_method:
+                error_message = _('Delivery method cannot be updated')
+                return {'delivery_method': error_message}
+        return {}
+
     def clean(self):
         """
         Override of clean method to perform additional validation on frequency, day_of_month/day_of week
--- a/tests/test_enterprise/api/test_views.py
+++ b/tests/test_enterprise/api/test_views.py
@@ -4394,6 +4394,75 @@ class TestEnterpriseReportingConfigAPIViews(APITest):
                 assert response_content[key] == value
 
     @mock.patch('enterprise.rules.crum.get_current_request')
+    def test_reporting_config_validate_delivery_method(self, request_or_stub_mock):
+        """
+        Tests that the PUT endpoint raise error if delivery method is changed while report updation.
+        """
+        user, enterprise_customer = self._create_user_and_enterprise_customer('test_user', 'test_password')
+        model_item = {
+            'active': True,
+            'enable_compression': True,
+            'delivery_method': 'email',
+            'day_of_month': 1,
+            'day_of_week': None,
+            'hour_of_day': 1,
+            'enterprise_customer': enterprise_customer,
+            'email': 'test@test.com\nfoo@test.com',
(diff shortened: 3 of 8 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -840,7 +840,17 @@ def validate(self, data):  # pylint: disable=arguments-renamed
         if error:
             raise serializers.ValidationError(error)
 
+        uuid = data.get('uuid')
         delivery_method = data.get('delivery_method')
+        if uuid:
+            delivery_method_validity = EnterpriseCustomerReportingConfiguration.validate_delivery_method(
```
saleem-latif, line 846: nit: rename to `delivery_method_error`

enterprise/models.py
```diff
@@ -2757,6 +2757,24 @@ def validate_compression(cls, enable_compression, data_type, delivery_method):
...
+            uuid (str): customer uuid
+            delivery_method (str): selected delivery method
+
+        Returns:
+            (dict): Validation Error
+        """
+        if uuid:
+            old_instance = EnterpriseCustomerReportingConfiguration.objects.get(uuid=uuid)
+            if old_instance.delivery_method != delivery_method:
+                error_message = _('you can not change the delivery method')
```
saleem-latif, line 2774: nit: please change to `Delivery method can not be changed.`

saleem-latif (approved):

enterprise/api/v1/serializers.py
```diff
@@ -840,7 +840,17 @@ def validate(self, data):  # pylint: disable=arguments-renamed
         if error:
             raise serializers.ValidationError(error)
 
+        uuid = data.get('uuid')
```
HammadAhmadWaqas, line 843: the best way to capture this logically would be something like this:
`create_report = data.get('uuid') is None`

enterprise/models.py
```diff
@@ -2757,6 +2757,24 @@ def validate_compression(cls, enable_compression, data_type, delivery_method):
             return {'enable_compression': error_message}
         return {}
 
+    def validate_delivery_method(self, uuid, delivery_method):
+        """
+        Check delivery_method is changed or not while updating report
+
+        Arguments:
+            uuid (str): customer uuid
```
HammadAhmadWaqas, line 2765: `create_report` is more logical variable name.

enterprise/models.py
```diff
@@ -2757,6 +2757,24 @@ def validate_compression(cls, enable_compression, data_type, delivery_method):
...
+            delivery_method (str): selected delivery method
+
+        Returns:
+            (dict): Validation Error
+        """
+        if uuid:
+            old_instance = EnterpriseCustomerReportingConfiguration.objects.get(uuid=uuid)
```
HammadAhmadWaqas, line 2772: You don't need to get instances from db.

enterprise/api/v1/serializers.py
```diff
@@ -840,7 +840,17 @@ def validate(self, data):  # pylint: disable=arguments-renamed
...
         delivery_method = data.get('delivery_method')
+        if uuid:
+            delivery_method_error = EnterpriseCustomerReportingConfiguration.validate_delivery_method(
```
HammadAhmadWaqas, line 846: validate_delivery_method is not a class method or static method, how can you call without its object?

tests/test_enterprise/api/test_views.py
```diff
@@ -4393,6 +4393,71 @@ def test_reporting_config_put_permissions(self, has_feature_role, expected_statu
...
+                server=settings.TEST_SERVER,
+                reverse_url=reverse(
+                    'enterprise-customer-reporting-detail',
+                    kwargs={'uuid': str(test_config.uuid)}
+                ),
+            ),
+            data=put_data,
+            format='json',
+        )
+        assert response.status_code == status.HTTP_400_BAD_REQUEST
```
HammadAhmadWaqas, line 4459: Can you please assert the error message as well? Otherwise, 400 errors could be due to any reason. it doesn't prove anything.

tests/test_models.py
```diff
@@ -2173,6 +2173,56 @@ def test_clean_missing_sftp_fields(self):
...
+
+        expected_errors = [
+            'Delivery method cannot be updated',
+        ]
+        try:
+            config.clean()
+        except ValidationError as validation_error:
+            assert sorted(validation_error.messages) == sorted(expected_errors)
+
+    def test_clean_validate_delivery_method_working(self):
```
HammadAhmadWaqas, line 2205: instead of writing `working`, it should be written `creating`.

tests/test_models.py
```diff
@@ -2173,6 +2173,56 @@ def test_clean_missing_sftp_fields(self):
...
+        """
+        Test ``EnterpriseCustomerReportingConfiguration`` custom clean function validating function success flow.
```
HammadAhmadWaqas, line 2207: What is success flow?

HammadAhmadWaqas (approved):

PR #1719, opened 2023-02-22
chore: migrating uuid and updating column constraints

alex-sheehan-edx:
**Merge checklist:**

**Post merge:**


johnnagro (approved):

enterprise/migrations/0168_auto_20230222_1621.py
```diff
@@ -0,0 +1,46 @@
...
+    for hist_learner_credit_enrollment in hist_learner_credit_enrollments.objects.all():
+        hist_learner_credit_enrollment.uuid = uuid.uuid4()
+        hist_learner_credit_enrollment.save()
+
+
+def create_hist_license_enrollment_uuid(apps, schema_editor):
+    hist_licensed_enrollments = apps.get_model('enterprise', 'historicallicensedenterprisecourseenrollment')
+    for hist_licensed_enrollment in hist_licensed_enrollments.objects.all():
+        hist_licensed_enrollment.uuid = uuid.uuid4()
+        hist_licensed_enrollment.save()
```
iloveagent57, line 25: I don't think this is what you want - these uuids would be different from the actual models' `uuid` values.
But also, you shouldn't have to directly interact with the historical models - that should come for free when you do `model.save()` in the non-historical models.

iloveagent57 (changes requested):

enterprise/migrations/0168_auto_20230222_1621.py
```diff
@@ -0,0 +1,30 @@
...
+
+def create_license_enrollment_uuid(apps, schema_editor):
+    licensed_enrollments = apps.get_model('enterprise', 'LicensedEnterpriseCourseEnrollment')
+    for licensed_enrollment in licensed_enrollments.objects.all():
+        licensed_enrollment.uuid = uuid.uuid4()
+        licensed_enrollment.save()
+
+
+def create_learner_credit_enrollment_uuid(apps, schema_editor):
+    learner_credit_enrollments = apps.get_model('enterprise', 'learnercreditenterprisecourseenrollment')
```
iloveagent57, line 15: curious: does a lower-cased model name actually work as the second `apps.get_model()` arg?

iloveagent57 (approved):

PR #1721, opened 2023-02-23
chore: cleanup squashed migration files

johnnagro:
- reverts https://github.com/openedx/edx-enterprise/pull/1696
- https://2u-internal.atlassian.net/browse/ENT-6821
- cc @rgraber


alex-sheehan-edx (approved):
can we really just remove these migrations with nothing added??

PR #1724, opened 2023-03-08
feat: Add user_id support to enroll_learners_in_courses endpoint

pwnage101:


alex-sheehan-edx (approved):
We might benefit from a unit test where both user ID and email are supplied. But otherwise this looks great to me

enterprise/api/v1/views.py
```diff
@@ -318,9 +326,14 @@ def enroll_learners_in_courses(self, request, pk):
         for course_run in course_runs_modes:
             course_runs_modes[course_run] = get_best_mode_from_course_key(course_run)
 
-        for index, info in enumerate(enrollments_info):
+        for info in enrollments_info:
+            if 'email' not in info:
+                # TODO: error handling if there's neither an 'email' nor a 'user_id'.
```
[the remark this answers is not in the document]
alex-sheehan-edx: https://github.com/openedx/edx-enterprise/blob/master/enterprise/api/v1/serializers.py#L1259-L1284 I think this should probably live here?

enterprise/api/v1/views.py
```diff
@@ -334,10 +351,11 @@ def enroll_learners_in_courses(self, request, pk):
...
                 email_errors.append(email)
 
-        # Remove the bad emails from enrollments_info and the emails set, don't attempt to enroll or link bad emails.
-        for errored_user in email_errors:
-            enrollments_info[:] = [info for info in enrollments_info if info['email'] != errored_user]
-            emails.remove(errored_user)
+        # Remove the bad emails and bad user_ids from enrollments_info; don't attempt to enroll or link them.
+        enrollments_info = [
+            info for info in enrollments_info
+            if 'email' in info and info['email'] not in email_errors and info.get('user_id') not in user_id_errors
```
alex-sheehan-edx, line 357: could you explain why we're checking for email and user id exists in the errors lists, in different ways? Is there a reason why we couldn't do `info.get('email')` or `'user_id' in info and info['user_id']` to keep it consistent?

enterprise/utils.py
```diff
@@ -1952,14 +1967,30 @@ def enroll_subsidy_users_in_courses(enterprise_customer, subsidy_users_info, dis
         'failures': [],
     }
     for subsidy_user_info in subsidy_users_info:
-        user_email = subsidy_user_info.get('email').strip().lower()
+        user_id = subsidy_user_info.get('user_id')
+        user_email = subsidy_user_info.get('email').strip().lower() if 'email' in subsidy_user_info else None
```
alex-sheehan-edx, line 1971: Does this need to be None? Alternatively you could do `user_email = subsidy_user_info.get('email', '').strip().lower()`

alex-sheehan-edx (approved):
Couple of questions otherwise looks pretty good

enterprise/api/v1/views.py
pwnage101: I deliberated both ways, in the end I thought it was more robust to be more opinionated about there always being an email.  This is not a strongly held opinion and can go either way---the way you described is probably more readable if you're asking about it

enterprise/utils.py
pwnage101: I would prefer to keep this None for consistency with user_id, but what I was able to do was slightly shorten the line by using `subsidy_user_info['email'].strip().lower()` which is still protected by the following condition.

PR #1725, opened 2023-03-08
fix: check against an updated settings.DEFAULT_FILE_STORAGE to determine if in stage/prod environment

adamstankiewicz:
# Description 

Bug ticket: https://2u-internal.atlassian.net/browse/ENT-6873

The enterprise product supports customers who use their custom logo for the application header. This is determined by `safe_logo_url` on the `EnterpriseCustomerBrandingConfiguration` model. This method determines whether to use an S3 URL for stage/prod or the concatenated `settings.LMS_ROOT_URL + settings.MEDIA_URL`.

It makes this choice based on the `settings.DEFAULT_FILE_STORAGE` setting, which recently updated last week from `storages.backends.s3boto.S3BotoStorage` to `storages.backends.s3boto3.S3Boto3Storage`, thus breaking the conditional to use the S3 URL in stage/prod.

This PR makes `safe_logo_url` support both the old and new values for `settings.DEFAULT_FILE_STORAGE`.

Long term, we should find another way to flag on whether we are in a local vs. stage/prod environment than relying on `settings.DEFAULT_FILE_STORAGE` directly.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -1592,6 +1592,17 @@ class EnterpriseCustomerBrandingConfiguration(TimeStampedModel):
         """
         return self.__str__()
 
+    def _check_file_storage_environment(self):
+        """
+        Returns whether `settings.DEFAULT_FILE_STORAGE` is set for
+        stage/prod or dev environment.
+        """
+        allowed_default_file_storages = [
+            'storages.backends.s3boto.S3BotoStorage',
+            'storages.backends.s3boto3.S3Boto3Storage',
+        ]
+        return settings.DEFAULT_FILE_STORAGE in allowed_default_file_storages
+
     @property
     def safe_logo_url(self):
         """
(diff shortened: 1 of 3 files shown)


enterprise/models.py
```diff
@@ -1602,7 +1602,7 @@ def safe_logo_url(self):
 
         # AWS S3 storage is used in stage/production environments but file system
         # storage is used in devstack environment
-        if settings.DEFAULT_FILE_STORAGE == 'storages.backends.s3boto.S3BotoStorage':
+        if settings.DEFAULT_FILE_STORAGE == 'storages.backends.s3boto3.S3Boto3Storage':
```
adamstankiewicz, line 1605: Long term, we should find another way to flag on whether we are in a local vs. stage/prod environment than relying on `settings.DEFAULT_FILE_STORAGE` directly.

brobro10000 (approved):

PR #1726, opened 2023-03-13
feat: get all catalog queries api endpoint

brobro10000:
Serializes EnterpriseCatalogQuery Model
Distinctly classifies model field, `content_filter` as JSON to parse data
Creates nondescript ViewSet with default Django readonly endpoints
Denotes permissions based on a Jwt Authenticated user with session
Creates general viewset route

Writes a basic test

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -1406,3 +1406,16 @@ class EnterpriseCustomerUnlinkUsersSerializer(serializers.Serializer):
         required=False,
         default=False,
     )
+
+
+class EnterpriseCatalogQuerySerializer(serializers.ModelSerializer):
+    """
+    Serializer for the ``EnterpriseCatalogQuery`` model.
+    """
+
+    class Meta:
+        model = models.EnterpriseCatalogQuery
+        fields = '__all__'
+
+    # Parses from a dictionary to JSON
+    content_filter = serializers.JSONField(required=False)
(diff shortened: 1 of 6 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -1406,3 +1406,16 @@ class EnterpriseCustomerUnlinkUsersSerializer(serializers.Serializer):
...
+    """
+    Serializer for the ``EnterpriseCatalogQuery`` model.
+    """
+
+    class Meta:
+        model = models.EnterpriseCatalogQuery
+        fields = '__all__'
+
+    # Parses from a dictionary to JSON
+    content_filter = serializers.JSONField(required=False)
```
brobro10000, line 1421: Distinction required due to default dictionary output without serializing in this format

tests/test_enterprise/api/test_views.py
```diff
@@ -2890,84 +2889,68 @@ def test_enterprise_customer_catalogs_response_formats(self):
...
         """
-        client = APIClient()
-        # User is not logged in.
-        response = client.get(
-            settings.TEST_SERVER + reverse(self.CATALOG_QUERY_ENDPOINT, kwargs={'catalog_query_id': 1})
-        )
-        assert response.status_code == 403
-        response = response.json()
-        assert response['detail'] == 'Authentication credentials were not provided.'
+        ENTERPRISE_CATALOG_QUERY_ENDPOINT = reverse('enterprise_catalog_query-detail', args=['bad-uuid'])
```
pwnage101, line 2950: shouldn't 'bad-uuid' result in a 400 error to indicate that uuid validation failed?

pwnage101 (commented):
The code LGTM!  If I'm understanding correctly, the end result is that this doesn't remove the retrieve endpoint, but it only adds a list endpoint.    This would be good to emphasize in your commit message and provide a long-form explanation in your commit body.

macdiesel (approved):

PR #1729, opened 2023-03-16
feat: upgraded to node v18, added .nvmrc and updated workflows

BilalQamar95:
**Merge checklist:**

**Post merge:**


.github/workflows/publish.yml
```diff
@@ -11,14 +11,16 @@ jobs:
...
+      - name: Setup Nodejs Env
+        run: echo "NODE_VER=`cat .nvmrc`" >> $GITHUB_ENV
       - name: setup python
         uses: actions/setup-python@v2
         with:
           python-version: 3.8
       - name: Setup Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: 12
```
adamstankiewicz, line 21: oof, blast from the past all the way to node 12 haha

adamstankiewicz (approved):

adamstankiewicz:
@BilalQamar95 @abdullahwaheed These changes look good to me. I also pushed up a `CHANGELOG` entry and version bump as a breaking release v4.0.0.

We unfortunately don't have capacity right now to merge / release / verify in edx-platform, but feel free to take this on if you'd like. Otherwise, we'll come back to this PR at a later date.

BilalQamar95:
> @BilalQamar95 @abdullahwaheed These changes look good to me. I also pushed up a `CHANGELOG` entry and version bump as a breaking release v4.0.0.
> 
> We unfortunately don't have capacity right now to merge / release / verify in edx-platform, but feel free to take this on if you'd like. Otherwise, we'll come back to this PR at a later date.

Hi @adamstankiewicz, based on chagneLog entry it seems you are expecting this to go as breaking change. Since there are no major version bumps in packages or any code changes made alongside the Node.js version update, I believe we can treat this update as a minor version update. Please let me know if I'm missing something, curious to know your rationale behind proposing a major release here.

adamstankiewicz:
@BilalQamar95 I treated these changes as a new major version because, in the past, I've seen dropping Node.js support as causing breaking changes for some consumers when not released as a new major version. That said, we don't typically define an `engines` property in our package.json files so we haven't really explicitly defined what versions of Node our repos/libraries support...

Based on the guidance in [this article](https://medium.com/the-node-js-collection/maintainers-should-consider-following-node-js-release-schedule-ab08ed4de71a), I believe what the author proposes for maintainers makes sense. We probably should be explicitly defining the supported Node.js version(s) in our repos/libraries with the `engines` property. I'd be curious to hear your thoughts, not just for this repository, but as a pattern more broadly.
 
* Define an `engines` field that declare the [oldest active LTS version of Node.js](https://nodejs.dev/en/about/releases/) (i.e., Node 18): 

```json
{
  "engines": {
    "node": "^18",
    "npm": "^9"
  }
}
```

* "When a Node.js version reaches end-of-life, even if your library is finished, **bump a major version** updating the engines field to the current oldest active LTS of Node.js."
* "Consider [throwing a helpful exception](https://github.com/googleapis/gax-nodejs/pull/748) if your library is used on an unsupported Node.js version."
* "Consider documenting your version support policy." See [example](https://github.com/googleapis/google-cloud-node#supported-nodejs-versions) from the aforementioned article.

Some related tweets:
* "Changing any supported dependency (peer or direct) major version is a breaking change and needs a new major version" ([source](https://twitter.com/GlenTiki/status/1245053049342738433))
* "Repeat with me: dropping support for old [@nodejs](https://twitter.com/nodejs) release is a breaking change and it should be released in a major version." ([source](https://twitter.com/matteocollina/status/1245019022271406080?s=20))

[question] Speaking of documentation, are you aware if we've documented our process for dealing with Node.js upgrades anywhere in the Open edX wiki, perhaps taking into account the above pattern with `engines`? If not, that might be some documentation worth having for future Node.js upgrades.

BilalQamar95:
> @BilalQamar95 I treated these changes as a new major version because, in the past, I've seen dropping Node.js support as causing breaking changes for some consumers when not released as a new major version. That said, we don't typically define an `engines` property in our package.json files so we haven't really explicitly defined what versions of Node our repos/libraries support...
> 
> Based on the guidance in [this article](https://medium.com/the-node-js-collection/maintainers-should-consider-following-node-js-release-schedule-ab08ed4de71a), I believe what the author proposes for maintainers makes sense. We probably should be explicitly defining the supported Node.js version(s) in our repos/libraries with the `engines` property. I'd be curious to hear your thoughts, not just for this repository, but as a pattern more broadly.
> 
> * Define an `engines` field that declare the [oldest active LTS version of Node.js](https://nodejs.dev/en/about/releases/) (i.e., Node 18):
> 
> ```json
> {
>   "engines": {
>     "node": "^18",
>     "npm": "^9"
>   }
> }
> ```
> 
> * "When a Node.js version reaches end-of-life, even if your library is finished, **bump a major version** updating the engines field to the current oldest active LTS of Node.js."
> * "Consider [throwing a helpful exception](https://github.com/googleapis/gax-nodejs/pull/748) if your library is used on an unsupported Node.js version."
> * "Consider documenting your version support policy." See [example](https://github.com/googleapis/google-cloud-node#supported-nodejs-versions) from the aforementioned article.
> 
> Some related tweets:
> 
> * "Changing any supported dependency (peer or direct) major version is a breaking change and needs a new major version" ([source](https://twitter.com/GlenTiki/status/1245053049342738433))
> * "Repeat with me: dropping support for old [@nodejs](https://twitter.com/nodejs) release is a breaking change and it should be released in a major version." ([source](https://twitter.com/matteocollina/status/1245019022271406080?s=20))
> 
> [question] Speaking of documentation, are you aware if we've documented our process for dealing with Node.js upgrades anywhere in the Open edX wiki, perhaps taking into account the above pattern with `engines`? If not, that might be some documentation worth having for future Node.js upgrades.

Yes that definitely makes sense, defining the supported Node.js version(s) in the engines field of the package.json file appears to be a good practice for maintaining repositories and libraries, especially in a multi-package, microfrontend environment like ours where it can serve to provide clear guidance to developers and users about required Node.js and npm versions for proper functionality and compatibility. Declaring the oldest active LTS version of Node.js in the engines field allows devs to use latest LTS version while ensuring compatibility with older versions, whereas updating the engines field to newest LTS version when an older LTS version reaches end-of-life could help ensure compatibility and security. 
There is merit to consider it as a pattern broadly, I was wondering if we should create a ticket for it for now and take this up separately across all repos/libraries. 

I will be updating the PR and we can release this as a major version, also we have documented the process for node v18 upgrade which you can find [here](https://openedx.atlassian.net/wiki/spaces/AC/pages/3715891201/Node+18+Upgrade).

adamstankiewicz:
> There is merit to consider it as a pattern broadly, I was wondering if we should create a ticket for it for now and take this up separately across all repos/libraries.

@BilalQamar95 I filed a new issue as an epic to define & track any work related to this updated pattern, should the team decide to adopt it more broadly: https://github.com/openedx/wg-frontend/issues/171 

Wasn't sure if FED-BOM prefers a single issue to track across all repos/libraries or if the team prefers to have a separate issue per repo/library; just created a single epic issue for now 😄 

If the team adopts the pattern more broadly, it might be worth prioritizing this pattern given its closely related to the ongoing Node 18 upgrade work. Just note that when adding the `engines` property, it should probably be treated as a breaking change for libraries, resulting in a new major version.

adamstankiewicz (approved):
LGTM. Thanks for adding in the `engines` field and releasing it as a new major version.

@BilalQamar95 Will you be planning on releasing it to PyPi after merge and upgrading to it in `edx-platform`? Or is that something enterprise-titans should plan on tackling after this PR merges?

BilalQamar95:
> LGTM. Thanks for adding in the `engines` field and releasing it as a new major version.
> 
> @BilalQamar95 Will you be planning on releasing it to PyPi after merge and upgrading to it in `edx-platform`? Or is that something enterprise-titans should plan on tackling after this PR merges?

@adamstankiewicz we don't have any plans or the required domain knowledge to release it to PyPi, we intend to rely on enterprise-titans to tackle this once PR is merged.

PR #1733, opened 2023-03-31
feat: new mark orphaned transmisisions command and logic to remove du…

alex-sheehan-edx:
…ring transmissions

**Merge checklist:**

**Post merge:**


integrated_channels/integrated_channel/models.py
```diff
@@ -241,6 +242,23 @@ def clean(self):
                 }
             )
 
+    def fetch_linked_content_audits(self):
```
alex-sheehan-edx, line 245: important to note, this lives on every customer config and will grab all content transmission audits that are connected to a catalog associated with the config.

integrated_channels/integrated_channel/tasks.py
```diff
@@ -89,6 +90,51 @@ def _log_batch_task_finish(task_name, channel_code, job_user_id,
...
+    Placeholder
+    """
+    start = time.time()
+
+    linked_metadata_audits = ContentMetadataItemTransmission.objects.none()
+    # Go over each integrated channel
+    for individual_channel in INTEGRATED_CHANNEL_CHOICES.values():
+        try:
+            # Iterate through each configuration for the channel
+            for config in individual_channel.objects.all():
```
alex-sheehan-edx, line 106: I think we want to keep this on all content and not just active? We might need to make sure that orphaned records fetched later on in this task doesn't try and grab content for deactivated content.

mattdrayer: right, it would just skip over the entire channel/metadata set if the config is inactive

integrated_channels/integrated_channel/tasks.py
```diff
@@ -89,6 +90,51 @@ def _log_batch_task_finish(task_name, channel_code, job_user_id,
...
+                # and add them to the list of linked content keys
+                linked_metadata_audits = linked_metadata_audits | config.fetch_linked_content_audits()
+        except Exception:  # pylint: disable=broad-except
+            LOGGER.exception(
+                # '[Integrated Channel] Batch mark_orphaned_content_metadata_audit failed with exception. '
+                # f'Configuration: {channel}',
+                # exc_info=True,
+            )
+
+    # Mark all content metadata audits not in the list of linked content keys as orphaned
```
mattdrayer, line 117: How is this different from the task/command?  Do we need the task/command if this is here?

mattdrayer (commented):
I think I'm following along well enough here -- everything seems logical.  I suppose the tests will tell the truth :)

integrated_channels/integrated_channel/tasks.py
alex-sheehan-edx: not sure I follow- this is the task that is invoked by the management command.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -380,6 +380,21 @@ def _check_matched_content_to_delete(self, enterprise_customer_catalog, items):
                     )
         return items_to_delete
 
+    def _get_customer_config_orphaned_content(self, max_set_count):
+        """
+        placeholder
```
mattdrayer, line 385: Don't forget to fill this in

integrated_channels/integrated_channel/management/commands/mark_orphaned_content_metadata_audits.py
```diff
@@ -0,0 +1,20 @@
...
+
+from django.core.management.base import BaseCommand
+
+from integrated_channels.integrated_channel.tasks import mark_orphaned_content_metadata_audit
+from integrated_channels.integrated_channel.management.commands import IntegratedChannelCommandMixin
+
+
+class Command(IntegratedChannelCommandMixin, BaseCommand):
+    """
+    placeholder
```
mattdrayer, line 13: Don't forget to fill in

integrated_channels/integrated_channel/management/commands/mark_orphaned_content_metadata_audits.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+
+    def handle(self, *args, **options):
+        """
+        placeholder
```
mattdrayer, line 18: Don't forget to fill in

integrated_channels/integrated_channel/models.py
```diff
@@ -241,6 +242,23 @@ def clean(self):
...
+        """
+        Placeholder
```
mattdrayer, line 247: Don't forget to fill in -- maybe consider including your dev note above

integrated_channels/integrated_channel/models.py
```diff
@@ -241,6 +242,23 @@ def clean(self):
...
+        """
+        enterprise_customer_catalogs = self.customer_catalogs_to_transmit or \
```
mattdrayer, line 249: This makes me think about a case where a catalog UUID was in the `customer_catalogs_to_transmit` set, along with another one, and then the former was removed.  From a remote control perspective, removing a catalog from this set is another orphan creation event.  The catalog wasn't actually deleted in this case, and it's still linked to the customer, but it's now absent from the collection of catalogs being synced.  Am I overthinking this?

integrated_channels/integrated_channel/models.py
```diff
@@ -814,3 +833,21 @@ def __repr__(self):
         Return uniquely identifying string representation.
         """
         return self.__str__()
+
+
+class OrphanedContentTransmissions(TimeStampedModel):
+    """
+    Placeholder
```
mattdrayer, line 840: Don't forget to fill in

mattdrayer (approved):
Mostly just some nits -- one question regarding a scenario involving specific catalogs tagged for transmission vs. all catalogs, otherwise LGTM.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -380,6 +380,21 @@ def _check_matched_content_to_delete(self, enterprise_customer_catalog, items):
...
+        """
+        OrphanedContentTransmissions = apps.get_model(
+            'integrated_channel',
+            'OrphanedContentTransmissions'
+        )
+        # Grab orphaned content metadata items for the customer, ordered by oldest to newest
```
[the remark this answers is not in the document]
alex-sheehan-edx: yeah the thought is, if we're doing work to fit this into the regularly scheduled jobs, I wanted to address the oldest issue.

integrated_channels/integrated_channel/models.py
alex-sheehan-edx: while that does pose a vulnerability, I believe we explicitly state in the customer's configuration that if any value is inputed into the `catalogs to transmit` box, then it will be prioritized over the catalogs assigned to the customer's profile.

alex-sheehan-edx: so I would label that more as user error than a system error

mattdrayer: so we'd just leave the no-longer-synced catalog metadata records orphaned at that point?  Say Catalog A has all courses with titles starting with "A", Catalog B has all courses with titles starting with "B", and Catalog C has all courses with titles starting with "C".  Catalog A and Catalog B are both specified as "catalogs to transmit", and Catalog C is not.  After the sync completes the total cache set @ the remote system is A+B.  Someone then removes Catalog B from "catalogs to transmit".  When this happens, the system only syncs Catalog A records going forward, and the expectation would be that Catalog B records are removed.  But I don't think this will happen in the current implementation...or would it?

marlonkeating (changes requested):

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -470,6 +526,24 @@ def export(self, **kwargs):
...
+        # If we're not at the max payload count, we can check for orphaned content and shove it in the delete payload
+        current_payload_count = len(items_to_create) + len(items_to_update) + len(items_to_delete)
+        if current_payload_count < max_payload_count:
+            space_left_in_payload = max_payload_count - current_payload_count
+            orphaned_content_to_delete = self._get_customer_config_orphaned_content(
+                max_set_count=space_left_in_payload,
+            )
+
+            for orphaned_item in orphaned_content_to_delete:
+                # Dry run mode, don't actually delete anything
```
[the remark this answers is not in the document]
alex-sheehan-edx: hmm it would have to be an env var- but yeah we could probably do something like that

bcitro (approved):
LGTM

marlonkeating (approved):

PR #1740, opened 2023-04-17
build: remove unnecessary codecov dependency and use GHA instead

matthugs:
**Merge checklist:**
- ❌ (won't do) Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files)


adamstankiewicz:
`codecov` was removed from requirements and the GitHub Action workflow for coverage was added instead through this PR: https://github.com/openedx/edx-enterprise/pull/1739

Closing this PR as a result.

PR #1742, opened 2023-04-26
feat: Q filtering to enterprise customer basic list

brobro10000:
Adds additional query filter 'name_or_uuid' that generalizes the value of  'name_or_uuid' to filter the 'name' and 'id' fields

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -186,9 +186,12 @@ class EnterpriseCustomerViewSet(EnterpriseReadWriteModelViewSet):
         Enterprise Customer's Basic data list without pagination
         """
         startswith = request.GET.get('startswith')
+        q = request.GET.get('q')
         queryset = self.get_queryset().order_by('name')
         if startswith:
             queryset = queryset.filter(name__istartswith=startswith)
+        if q:
+            queryset = queryset.filter(Q(name__icontains=q) | Q(uuid__icontains=q))
         serializer = self.get_serializer(queryset, many=True)
         return Response(serializer.data)
 
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -15,6 +15,7 @@ from rest_framework import filters, generics, permissions, status, viewsets
 from rest_framework.authentication import SessionAuthentication
 from rest_framework.decorators import action
 from rest_framework.exceptions import NotFound, ValidationError
+from rest_framework.mixins import CreateModelMixin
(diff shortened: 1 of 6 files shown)


pwnage101 (changes requested):
`q` is pretty ambiguous of a field name.  Is this a pattern that you've seen elsewhere?  Even if it does stand for "query" colloquially, it's not even a query since it's actually used as a name or a uuid.  IMO a better parameter name would be `name_or_uuid`.

I'm also looking for unit tests changed/added to cover this new filter, as wel as some documentation somewhere in the code (ideally the docstring) explaining what all the fields do.  Even the existing `startswith` parameter needs documentation, since it's unclear what model field it refers to.

enterprise/api/v1/views.py
```diff
@@ -193,11 +193,18 @@ def get_serializer_class(self):
...
+        - name_or_uuid: filter by name or uuid in a single query parameter
+        - startswith: filter by name starting with the given string
         """
         startswith = request.GET.get('startswith')
+        name_or_uuid = request.GET.get('name_or_uuid')
         queryset = self.get_queryset().order_by('name')
         if startswith:
             queryset = queryset.filter(name__istartswith=startswith)
+        if name_or_uuid:
+            queryset = queryset.filter(Q(name__icontains=name_or_uuid) | Q(uuid__icontains=name_or_uuid))
```
pwnage101, line 207: I forgot to ask: are you intending these to be substring matches or exact matches?  substring matching for UUIDs never really makes sense, so I feel like at least the uuid_icontains needs to be changed to uuid=uuid.UUID(name_or_uuid), but that would throw an exception, so this line would need to be re-architected:
```suggestion
        if name_or_uuid:
            try:
                requested_customer uuid = uuid.UUID(name_or_uuid)
                queryset = queryset.filter(uuid=requested_customer_uuid)
            except ValueError:
                queryset = queryset.filter(name__icontains=name_or_uuid)
```

pwnage101: Huddled with Hamzah to answer my question: This query param actually needs to be a substring search on BOTH name and uuid because it's used in the frontend to drive autocompletion.

So, I'm happy with the current implementation, but I'd recommend additional details in the docstring to explain what the field is used for by the caller, and that under the hood it is a substring search, not an exact matcher.

pwnage101 (approved):
approved, with proposed docstring update

PR #1744, opened 2023-05-03
feat: new management command to remove duplicate transmission audits

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-7052

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/tasks.py
+++ b/integrated_channels/integrated_channel/tasks.py
@@ -9,6 +9,7 @@ from celery import shared_task
 from celery.utils.log import get_task_logger
 from edx_django_utils.monitoring import set_code_owner_attribute
 
+from django.conf import settings
 from django.contrib import auth
 from django.core.cache import cache
 from django.utils import timezone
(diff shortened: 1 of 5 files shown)


integrated_channels/integrated_channel/tasks.py
```diff
@@ -90,6 +91,45 @@ def _log_batch_task_finish(task_name, channel_code, job_user_id,
...
+    for unique_transmission in unique_transmissions:
+        duplicates = ContentMetadataItemTransmission.objects.filter(
+            id__in=ContentMetadataItemTransmission.objects.filter(
+                content_id=unique_transmission[0],
+                plugin_configuration_id=unique_transmission[1],
+                integrated_channel_code=unique_transmission[2]
+            ).order_by('-modified').values_list('id', flat=True)[1:]
+        )
+        duplicates_found += len(duplicates)
+        if getattr(settings, "REMOVE_DUPLICATE_CONTENT_TRASMISSION_FLAG", False):
```
mattdrayer, line 117: typo in flag name

integrated_channels/integrated_channel/tasks.py
```diff
@@ -90,6 +91,45 @@ def _log_batch_task_finish(task_name, channel_code, job_user_id,
         ))
 
 
+@shared_task
+@set_code_owner_attribute
+def remove_duplicate_transmission_audits():
+    """
+    Task to mark content metadata audits as orphaned if they are not linked to any customer catalogs.
```
mattdrayer, line 98: copy pasta

integrated_channels/integrated_channel/tasks.py
mattdrayer, line 117: Do we want to do all of the above work if the flag is not set/true?  Is this intended to be a "dry run" flag?

mattdrayer (approved):
Couple small things, otherwise LGTM

integrated_channels/integrated_channel/tasks.py
alex-sheehan-edx: If the flag exists and is set to true, then we want it to delete. Otherwise log and dry run

mattdrayer: Okay, if it's a "dry run" control, I recommend naming the parameter as such -- otherwise people may think the flag is an on/off switch for the entire workflow which isn't entirely true.

alex-sheehan-edx: basically we want to see exactly how many records the code intends to delete before we issue the operation

mattdrayer:
✨

PR #1746, opened 2023-05-05
feat: Add idempotent catalog creation endpoint

brobro10000:
Adds idempotency logic to create EnterpriseCustomerCatalog to prevent duplicate catalog configurations being created.
Endpoint should return an existing EnterpriseCustomerCatalog that matches the configuration intended to be created. 

Tests also added.

Updating branch, adding the CHANGELOG and bumping the version remain in body of work. 

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -1227,9 +1227,54 @@ class EnterpriseCustomerBrandingConfigurationViewSet(EnterpriseReadWriteModelVie
 
 
 class EnterpriseCustomerCatalogWriteViewSet(EnterpriseWriteOnlyModelViewSet):
+    """
+    API write only views for the ``enterprise-customer-catalog`` API endpoint.
+    """
     queryset = models.EnterpriseCustomerCatalog.objects.all()
     permission_classes = (permissions.IsAdminUser,)
     serializer_class = serializers.EnterpriseCustomerCatalogWriteOnlySerializer
+
+    def create(self, request, *args, **kwargs):
+        """
+        Creates a new EnterpriseCustomerCatalog and returns the created object.
+
+        If an EnterpriseCustomerCatalog already exists for the given enterprise_customer and enterprise_catalog_query,
+        returns the existing object.
+
+        URL: /enterprise/api/v1/enterprise-customer-catalog/
+
+        Method: POST
+
+        Payload::
+
+          {
+            "title":  string - Title of the catalog,
+            "enterprise_customer": string - UUID of an existing enterprise customer,
+            "enterprise_catalog_query": string - id of an existing enterprise catalog query,
+          }
+
+        Returns 201 if a new EnterpriseCustomerCatalog was created, 200 if an existing EnterpriseCustomerCatalog was
+        """
+
+        enterprise_customer_uuid = request.data.get('enterprise_customer')
+        enterprise_catalog_query_id = request.data.get('enterprise_catalog_query')
+        enterprise_customer_catalog_list = models.EnterpriseCustomerCatalog.objects.filter(
+            enterprise_customer=enterprise_customer_uuid)
+        for catalog in enterprise_customer_catalog_list:
+            catalog_query = catalog.enterprise_catalog_query
+            if catalog_query is not None and catalog_query.id == int(enterprise_catalog_query_id):
+                seralized_customer_catalog = serializers.EnterpriseCustomerCatalogWriteOnlySerializer(
+                    catalog)
+                LOGGER.info(
+                    'EnterpriseCustomerCatalog already exists for enterprise_customer_uuid: %s '
+                    'and enterprise_catalog_query_id: %s, using existing catalog: %s',
+                    enterprise_customer_uuid, enterprise_catalog_query_id, catalog.uuid)
+                return Response(seralized_customer_catalog.data, status=status.HTTP_200_OK)
+        LOGGER.info(
+            'Creating new EnterpriseCustomerCatalog for enterprise_customer_uuid: %s '
+            'and enterprise_catalog_query_id: %s',
+            enterprise_customer_uuid, enterprise_catalog_query_id)
+        return super().create(request, *args, **kwargs)
 
 
 class EnterpriseCustomerCatalogViewSet(EnterpriseReadOnlyModelViewSet):
--- a/tests/test_enterprise/api/test_views.py
+++ b/tests/test_enterprise/api/test_views.py
@@ -1688,22 +1688,63 @@ class TestEnterpriseCustomerCatalogWriteViewSet(BaseTestEnterpriseAPIViews):
         Test that a catalog can be created
         """
         enterprise_customer = factories.EnterpriseCustomerFactory(uuid=FAKE_UUIDS[0])
+        enterprise_catalog_query = factories.EnterpriseCatalogQueryFactory()
 
         self.set_jwt_cookie(ENTERPRISE_ADMIN_ROLE, str(enterprise_customer.uuid))
         self.user.is_staff = is_staff
         self.user.save()
+
         response = self.client.post(ENTERPRISE_CUSTOMER_CATALOG_ENDPOINT, {
             "title": "Test Catalog",
             "enterprise_customer": str(enterprise_customer.uuid),
+            "enterprise_catalog_query": str(enterprise_catalog_query.id),
         }, format='json')
         response_output = self.load_json(response.content)
 
         assert response.status_code == expected_status_code
+
         if expected_status_code == 201:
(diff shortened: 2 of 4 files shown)


enterprise/api/v1/views.py
```diff
@@ -1227,10 +1227,32 @@ def update_branding(self, request, enterprise_uuid):
...
+        returns the existing object.
+        """
+        enterprise_customer_uuid = request.data.get('enterprise_customer')
+        enterprise_catalog_query_id = request.data.get('enterprise_catalog_query')
+        enterprise_customer_catalog_list = models.EnterpriseCustomerCatalog.objects.filter(
+            enterprise_customer=enterprise_customer_uuid)
+        for catalog in enterprise_customer_catalog_list:
+            catalog_query = catalog.enterprise_catalog_query
+            if catalog_query is not None and catalog_query.id is int(enterprise_catalog_query_id):
+                seralized_customer_catalog = serializers.EnterpriseCustomerCatalogWriteOnlySerializer(
```
macdiesel, line 1251: Since we are expecting this endpoint to be pretty low volume could you add some additional logging in this area to let us know if a new catalog was created or if we returned an existing one?

tests/test_enterprise/api/test_views.py
```diff
@@ -1679,26 +1679,43 @@ class TestEnterpriseCustomerCatalogWriteViewSet(BaseTestEnterpriseAPIViews):
...
+
         response = self.client.post(ENTERPRISE_CUSTOMER_CATALOG_ENDPOINT, {
             "title": "Test Catalog",
             "enterprise_customer": str(enterprise_customer.uuid),
+            "enterprise_catalog_query": str(enterprise_catalog_query.id),
         }, format='json')
         response_output = self.load_json(response.content)
 
-        assert response.status_code == expected_status_code
+        if isIdempotent:
```
macdiesel, line 1705: I like that you attempted to reuse the DDT stuff here but I'm not sure that this is a good use case. I find this test hard to read and it's not immediately clear what is being tested and when. I would suggest writing a new test the deals only with the indempotency bits and to leave this test alone. 

Also I think that testing for idempotency warrants it's own test.

enterprise/api/v1/views.py
```diff
@@ -1227,10 +1227,40 @@ def update_branding(self, request, enterprise_uuid):
...
 
 class EnterpriseCustomerCatalogWriteViewSet(EnterpriseWriteOnlyModelViewSet):
+    """
+    API write only views for the ``enterprise-customer-catalog`` API endpoint.
+    """
     queryset = models.EnterpriseCustomerCatalog.objects.all()
     permission_classes = (permissions.IsAdminUser,)
     serializer_class = serializers.EnterpriseCustomerCatalogWriteOnlySerializer
 
+    def create(self, request, *args, **kwargs):
```
pwnage101, line 1237: Are the request parameters declared anywhere for this view?  IMO, you should at least declare them in the docstring, if not via drf_spectacular or something more robust.

enterprise/api/v1/views.py
```diff
@@ -1227,10 +1227,40 @@ def update_branding(self, request, enterprise_uuid):
...
+        If an EnterpriseCustomerCatalog already exists for the given enterprise_customer and enterprise_catalog_query,
+        returns the existing object.
+        """
+        enterprise_customer_uuid = request.data.get('enterprise_customer')
+        enterprise_catalog_query_id = request.data.get('enterprise_catalog_query')
+        enterprise_customer_catalog_list = models.EnterpriseCustomerCatalog.objects.filter(
+            enterprise_customer=enterprise_customer_uuid)
+        for catalog in enterprise_customer_catalog_list:
+            catalog_query = catalog.enterprise_catalog_query
+            if catalog_query is not None and catalog_query.id is int(enterprise_catalog_query_id):
```
pwnage101, line 1250: nit: numeral comparisons are typically written using `==` rather than `is`

enterprise/api/v1/views.py
```diff
@@ -1227,10 +1227,40 @@ def update_branding(self, request, enterprise_uuid):
...
+                seralized_customer_catalog = serializers.EnterpriseCustomerCatalogWriteOnlySerializer(
+                    catalog)
+                LOGGER.info(
+                    'EnterpriseCustomerCatalog already exists for enterprise_customer_uuid: %s\
+                        and enterprise_catalog_query_id: %s, using existing catalog: %s',
```
pwnage101, line 1255: Nit: this string will have like 20 spaces between your line breaks, the better way to do this multiline string is:
```suggestion
                    'EnterpriseCustomerCatalog already exists for enterprise_customer_uuid: %s '
                    'and enterprise_catalog_query_id: %s, using existing catalog: %s',
```

tests/test_enterprise/api/test_views.py
```diff
@@ -1688,23 +1688,58 @@ def test_create_catalog(self, is_staff, expected_status_code):
         Test that a catalog can be created
         """
         enterprise_customer = factories.EnterpriseCustomerFactory(uuid=FAKE_UUIDS[0])
+        enterprise_catalog_query = factories.EnterpriseCatalogQueryFactory(uuid=FAKE_UUIDS[0])
```
pwnage101, line 1691: should the test catalog query really have the same UUID as the test customer?

tests/test_enterprise/api/test_views.py
```diff
@@ -1688,23 +1688,58 @@ def test_create_catalog(self, is_staff, expected_status_code):
...
         self.user.save()
+
         response = self.client.post(ENTERPRISE_CUSTOMER_CATALOG_ENDPOINT, {
             "title": "Test Catalog",
             "enterprise_customer": str(enterprise_customer.uuid),
+            "enterprise_catalog_query": str(enterprise_catalog_query.id),
         }, format='json')
         response_output = self.load_json(response.content)
 
-        assert response.status_code == expected_status_code
```
pwnage101, line 1701: why remove this line?

tests/test_enterprise/api/test_views.py
```diff
@@ -1688,23 +1688,58 @@ def test_create_catalog(self, is_staff, expected_status_code):
...
+            "title": "Test Catalog",
+            "enterprise_customer": str(enterprise_customer.uuid),
+            "enterprise_catalog_query": str(enterprise_catalog_query.id),
+        }, format='json')
+        response_output = self.load_json(response.content)
+
+        duplicate_response = self.client.post(ENTERPRISE_CUSTOMER_CATALOG_ENDPOINT, {
+            "title": "Test Catalog 2",
+            "enterprise_customer": str(enterprise_customer.uuid),
+            "enterprise_catalog_query": str(enterprise_catalog_query.id),
```
pwnage101, line 1732: This is supposed to be enterprise_catalog_query.id and not enterprise_catalog_query.**uuid**?

tests/test_enterprise/api/test_views.py
```diff
@@ -1688,23 +1688,58 @@ def test_create_catalog(self, is_staff, expected_status_code):
...
+        }, format='json')
+        duplicate_response_output = self.load_json(duplicate_response.content)
+
+        assert response.status_code == 201
+        assert response_output['title'] == 'Test Catalog'
+        assert response_output['enterprise_customer'] == str(enterprise_customer.uuid)
+        assert duplicate_response.status_code == 200
+        assert duplicate_response_output['title'] == 'Test Catalog'
+        assert duplicate_response_output['enterprise_customer'] == str(enterprise_customer.uuid)
```
pwnage101, line 1741: I think one final check that's missing here is something to actually confirm in the database that no duplicate EnterpriseCustomerCatalog record has been created for this customer.  e.g. `len()` of that queryset should be 1.

enterprise/api/v1/views.py
pwnage101: Since we don't have drf_spectacular/redoc set up yet for this repo, honestly just listing the URL format and request body params in the docstring would be great!

tests/test_enterprise/api/test_views.py
brobro10000: Ahh looking at how the code is generated I guess they are not unique on declarations. Will increment the array 👍🏽

tests/test_enterprise/api/test_views.py
brobro10000: Removed too much during cleanup 😅

tests/test_enterprise/api/test_views.py
brobro10000: Yep, the catalog creation is based on the query id as opposed to the uuid

pwnage101 (approved):

PR #1750, opened 2023-05-08
feat: include auth_org_id in enterprise customer api

johnnagro:
## Description

- in order to include `auth_org_id` in GEAG requests, we need to be able to retrieve it in contexts outside enterprise (ecomm, subsidy service, etc)
- include `auth_org_id` in serializer, test factories, etc
- https://2u-internal.atlassian.net/browse/ENT-7079


adamstankiewicz (approved):
LGTM, with a version bump and CHANGELOG entry.

iloveagent57 (approved):

PR #1754, opened 2023-05-15
feat: hooking enterprise enrollments up to platform signals to write unenrollment records

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-6927

**Merge checklist:**

**Post merge:**


enterprise/migrations/0172_auto_20230512_2232.py
```diff
@@ -0,0 +1,37 @@
+# Generated by Django 3.2.19 on 2023-05-12 22:32
+
+from django.db import migrations, models
+
+from enterprise.models import EnterpriseCourseEnrollment
+
+
+def backfill_unenrolled_enterprise_enrollments(apps, schema_editor):
+    enterprise_enrollments = EnterpriseCourseEnrollment.objects.all()
```
alex-sheehan-edx, line 9: [context] we have to import this table directly in order to get access to the `course_enrollment` property

CHANGELOG.rst
```diff
@@ -18,6 +18,10 @@ Unreleased
 * Switch from ``edx-sphinx-theme`` to ``sphinx-book-theme`` since the former is
   deprecated
 
+[3.62.8]
+--------
+feat: hooking enterprise enrollments up to platform signals to write unenrollment records
```
iloveagent57, line 23: docs nit: maybe worth mentioning here that we default this new field to null because it's really only applicable for enrollments written in some future, EMET-enabled reality.

enterprise/__init__.py
```diff
@@ -2,6 +2,6 @@
 Your project description goes here.
 """
 
-__version__ = "3.62.7"
+__version__ = "3.62.8"
```
iloveagent57, line 5: nit: probably worth a minor version bump with the model change.

iloveagent57 (approved):

johnnagro (approved):

PR #1757, opened 2023-05-21
Revert "feat: Added course_run_url for executive education courses"

irfanuddinahmad:
Reverts openedx/edx-enterprise#1756


CHANGELOG.rst
```diff
@@ -18,10 +18,6 @@ Unreleased
 * Switch from ``edx-sphinx-theme`` to ``sphinx-book-theme`` since the former is
   deprecated
 
-[3.64.0]
---------
-feat: Updated course_run_url for Executive Education courses
```
adamstankiewicz, line 23: [clarification] given this version is already published on [PyPi](https://pypi.org/project/edx-enterprise/) at 3.64.0, it feels a bit odd to remove the changelog entry as I believe this 3.64.0 version would still be published on PyPi? Should we instead treat it as a new released version, e.g. `3.64.1` or `3.65.0` to avoid a mismatch between published versions on PyPi and the changelog verisons?

adamstankiewicz (approved):

adamstankiewicz (approved):

zamanafzal (approved):

PR #1758, opened 2023-05-22
feat: new enterprise endpoint to surface recently unenrolled subsidized enrollments

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-7180

**Merge checklist:**

**Post merge:**


enterprise/api/v1/urls.py
```diff
@@ -57,6 +57,11 @@
         views.EnterpriseSubsidyFulfillmentViewSet.as_view({'post': 'cancel_enrollment'}),
         name='enterprise-subsidy-fulfillment-cancel-enrollment'
     ),
+    re_path(
+        r'enterprise-subsidy-fulfillment/recently_unenrolled/?$',
```
macdiesel, line 61: @iloveagent57 Thoughts on the restful-nes of this? I think that one of our goals is to remain restful in our api implementation? Should we be using an endpoint like `enterprise-course-enrollment` and sending in params to filter on `learner credit enrollments` and `order_by` pick the records we want? 

As I'm reading it this seems a bit overly specific and like an RPC?

iloveagent57: The way I would think about what Brian is proposing:
* Implement this as an `EnterpriseSubsidyFulfillment` list view.
* Allow that view to be filtered on the `unenrolled` boolean and filtered by the `modified` datetime.

This will give clients of the endpoint more flexibility in how they request data (i.e. let the client decide what "recent" means).

This brings up another point - perhaps `unenrolled` should really be a datetime field called `unenrolled_at`?

johnnagro (approved):

enterprise/api/v1/views.py
```diff
@@ -658,6 +665,64 @@ def get_serializer_class(self):
...
+
+        if not self.request.user.is_staff:
+            ent_customer_uuid = get_enterprise_customer_from_user_id(self.request.user.id)
+
+            enterprise_customer_filter = Q(
+                enterprise_course_enrollment__enterprise_customer_user__enterprise_customer__uuid=ent_customer_uuid
+            )
+            unenrolled_queryset = unenrolled_queryset.filter(enterprise_customer_filter)
+
+        return unenrolled_queryset
```
iloveagent57, line 696: It's not necessary for users with the operator role to be marked with `is_staff=true`.  You may just want to remove this filter from the queryset if the endpoint will only ever be called by operators.

iloveagent57 (approved):
Looks good, docstring nit.

PR #1759, opened 2023-05-24
fix: handle case where LearnerCreditEnterpriseCourseEnrollment already exists

pwnage101:
When calling enroll_learners_in_courses (a.k.a. the bulk enrollment endpoint), gracefully handle the case when a subsidized enrollment (for learner credit 2) is being re-created after being revoked previously.

Before this change, the existing LearnerCreditEnterpriseCourseEnrollment object would not be found (due to transaction_id changing), then the view would fatally fail to create a new one due to a unique constraint on the `enterprise_course_enrollment_id` field.  After this change, it finds the existing object and updates a couple fields on it to re-use it.

ENT-7204

### DIFF
--- a/enterprise/api/v1/views.py
+++ b/enterprise/api/v1/views.py
@@ -746,10 +746,10 @@ class EnterpriseSubsidyFulfillmentViewSet(EnterpriseWrapperApiViewSet):
             /enterprise/api/v1/subsidy-fulfillment/{fulfillment_source_uuid}/cancel-enrollment/
         """
         try:
-            enrollment = get_object_or_404(
+            subsidy_fulfillment = get_object_or_404(
                 self.get_subsidy_fulfillment_queryset(), uuid=fulfillment_source_uuid
             )
-            if enrollment.is_revoked:
+            if subsidy_fulfillment.is_revoked:
                 return Response(
                     status=HTTP_400_BAD_REQUEST,
                     data={'detail': 'Enrollment is already canceled.'}
@@ -761,19 +761,19 @@ class EnterpriseSubsidyFulfillmentViewSet(EnterpriseWrapperApiViewSet):
             )
 
         try:
-            username = enrollment.enterprise_course_enrollment.enterprise_customer_user.username
+            username = subsidy_fulfillment.enterprise_course_enrollment.enterprise_customer_user.username
             enrollment_api.update_enrollment(
                 username,
-                enrollment.enterprise_course_enrollment.course_id,
+                subsidy_fulfillment.enterprise_course_enrollment.course_id,
                 is_active=False,
             )
-            enrollment.revoke()
+            subsidy_fulfillment.revoke()
         except Exception as exc:  # pylint: disable=broad-except
             msg = (
-                f'Subsidized enrollment terminations error: unable to unenroll User {username}'
-                f'from Course {enrollment.course_id}  because: {str(exc)}'
+                f'Subsidized enrollment terminations error: unable to unenroll User {username} '
+                f'from Course {subsidy_fulfillment.enterprise_course_enrollment.course_id} because: {str(exc)}'
             )
-            LOGGER.error('{msg}: {exc}'.format(msg=msg, exc=exc))
+            LOGGER.error(msg)
             return Response(msg, status=HTTP_500_INTERNAL_SERVER_ERROR)
         return Response(status=HTTP_200_OK)
 
--- a/enterprise/models.py
+++ b/enterprise/models.py
@@ -2083,13 +2083,27 @@ class EnterpriseFulfillmentSource(TimeStampedModel):
         """
         Marks this object as revoked and marks the associated EnterpriseCourseEnrollment
         as "saved for later".  This object and the associated EnterpriseCourseEnrollment are both saved.
-        """
-        self.is_revoked = True
-        self.enterprise_course_enrollment.saved_for_later = True
-        self.enterprise_course_enrollment.unenrolled = True
-        self.enterprise_course_enrollment.save()
 
-        # TODO revoke entitlements as well?
+        TODO: revoke entitlements as well?
+        """
+        if self.enterprise_course_enrollment:
+            self.enterprise_course_enrollment.saved_for_later = True
+            self.enterprise_course_enrollment.unenrolled = True
+            self.enterprise_course_enrollment.save()
+
+        self.is_revoked = True
+        self.save()
+
+    def reactivate(self, **kwargs):
+        """
+        Idmpotently reactivates this enterprise fulfillment source.
+        """
+        if self.enterprise_course_enrollment:
+            self.enterprise_course_enrollment.saved_for_later = False
+            self.enterprise_course_enrollment.unenrolled = False
+            self.enterprise_course_enrollment.save()
+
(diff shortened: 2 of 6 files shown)


enterprise/utils.py
```diff
@@ -1822,10 +1822,17 @@ def customer_admin_enroll_user_with_status(
...
+            subsidy_enrollment_obj, created = subsidized_enterprise_course_enrollment_model().objects.get_or_create(
                 enterprise_course_enrollment=obj,
+                defaults={"transaction_id": transaction_id},
             )
+            if not created:
+                # It's possible that the subsidized enterprise course enrollment already existed, whose
+                # ``transaction_id`` would correspond to a reversed ledgered transaction, and ``is_revoked`` would
+                # probably be True.  All that's in the past now, so we should override those two fields.
+                subsidy_enrollment_obj.transaction_id = transaction_id
+                subsidy_enrollment_obj.is_revoked = False
```
alex-sheehan-edx, line 1834: should we set the fulfillment's enterprise enrollment to `unenrolled=False` ?

enterprise/utils.py
```diff
@@ -1822,10 +1822,17 @@ def customer_admin_enroll_user_with_status(
...
-                transaction_id=transaction_id,
+            subsidy_enrollment_obj, created = subsidized_enterprise_course_enrollment_model().objects.get_or_create(
                 enterprise_course_enrollment=obj,
+                defaults={"transaction_id": transaction_id},
             )
+            if not created:
+                # It's possible that the subsidized enterprise course enrollment already existed, whose
+                # ``transaction_id`` would correspond to a reversed ledgered transaction, and ``is_revoked`` would
+                # probably be True.  All that's in the past now, so we should override those two fields.
+                subsidy_enrollment_obj.transaction_id = transaction_id
```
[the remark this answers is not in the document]
alex-sheehan-edx: (this is more of a thought than a request)

alex-sheehan-edx (commented):
Would love a test for this, one tiny nit and an open question, but otherwise looks pretty good to me.

enterprise/utils.py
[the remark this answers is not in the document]
pwnage101: We do have django simple history enabled on this model.  Maybe that's enough, at least for the case of manual debugging and introspection?

enterprise/api/v1/views.py
```diff
@@ -771,7 +771,7 @@ def cancel_enrollment(self, request, fulfillment_source_uuid):
         except Exception as exc:  # pylint: disable=broad-except
             msg = (
                 f'Subsidized enrollment terminations error: unable to unenroll User {username}'
-                f'from Course {enrollment.course_id}  because: {str(exc)}'
+                f'from Course {enrollment.enterprise_course_enrollment.course_id}  because: {str(exc)}'
```
pwnage101, line 774: This change wasn't actually needed/related to this PR, but I did run into it.  The enrollment here has no course_id attribute.

enterprise/models.py
```diff
@@ -2083,13 +2083,27 @@ def revoke(self):
...
-        self.enterprise_course_enrollment.save()
+        self.save()
 
-        # TODO revoke entitlements as well?
+    def renew(self, **kwargs):
+        """
+        Idmpotently reactivates this enterprise fulfillment source.
+        """
+        if self.enterprise_course_enrollment:
+            self.enterprise_course_enrollment.saved_for_later = False
```
pwnage101, line 2102: Really not sure what saved_for_later means, but I threw it in.  Reviewers, please let me know if this is correct.

enterprise/settings/test.py
```diff
@@ -170,6 +170,7 @@ def root(*args):
         'service_user': SERVICE_USER_THROTTLE_RATE,
     },
     'DATETIME_FORMAT': '%Y-%m-%dT%H:%M:%SZ',
+    'TEST_REQUEST_DEFAULT_FORMAT': 'json',
```
pwnage101, line 173: Here's another bonus change that allows all of our client.post() calls in unit tests to be cleaner.

test_utils/factories.py
```diff
@@ -393,6 +395,7 @@ class Meta:
     transaction_id = factory.LazyAttribute(lambda x: FAKER.uuid4())
     enterprise_course_enrollment = factory.SubFactory(EnterpriseCourseEnrollmentFactory)
     is_revoked = False
+    fulfillment_type = FulfillmentTypes.LEARNER_CREDIT
```
pwnage101, line 398: Another set of unrelated changes here.  I just noticed that all the `LearnerCreditEnterpriseCourseEnrollment` objects in unit tests had the wrong fulfillment_type.

tests/test_enterprise/api/test_views.py
```diff
@@ -4436,6 +4427,84 @@ def test_bulk_enrollment_in_bulk_courses_nonexisting_user_id(
...
+    @mock.patch("enterprise.utils.lms_enroll_user_in_course")
+    def test_bulk_enrollment_enroll_after_cancel(
+        self,
+        mock_platform_enrollment,
+        mock_get_course_mode,
+        mock_enrollment_api,
+    ):
+        """
+        Test that even after a cancelled enterprise enrollment, an attempt to re-enroll the same learner in content
+        results in expected state and payload.
```
pwnage101, line 4442: High level architecture of this test: I use factories to create the enrollment initially, then use 2 APIs to subsequently cancel then re-enroll.

enterprise/models.py
iloveagent57: `saved_for_later` does basically what it says - when true, the enrollment is rendered in the "Saved for later" section of the learner portal dashboard.  I would probably just leave the value unchanged during renewal, but it's not a big deal either way.

enterprise/api/v1/views.py
iloveagent57: nit: `enrollment` seems like not the right variable name in here - isn't it really a fulfillment?

enterprise/models.py
```diff
@@ -2111,6 +2125,13 @@ class LearnerCreditEnterpriseCourseEnrollment(EnterpriseFulfillmentSource):
     .. no_pii:
     """
 
+    def renew(self, transaction_id=None, **kwargs):
+        """
+        Idmpotently reactivates this LearnerCreditEnterpriseCourseEnrollment.
+        """
+        self.transaction_id = transaction_id
```
iloveagent57, line 2132: It's probably worth throwing an exception if the transaction id hasn't changed from the original fulfillment to the renewed fulfillment.  Or at least log a warning, maybe?

iloveagent57 (commented):
This looks great, just one open suggestion/question of what to do if the transaction_id hasn't changed during renewal.

Also, consider using the verb `reenroll` (or "refulfill", "reactivate", whatevs) to help disambiguate from enterprise subscription renewals.

enterprise/api/v1/views.py
pwnage101: totally, i can change that too

iloveagent57 (approved):

enterprise/settings/test.py
pwnage101: backing out this change since it had the side effect of making PUT requests use JSON when they shouldn't.

PR #1764, opened 2023-06-01
feat: Added org_id for exec_ed landing page

irfanuddinahmad:
**Merge checklist:**

**Post merge:**


enterprise_learner_portal/api/v1/serializers.py
```diff
@@ -81,3 +84,16 @@ def _get_course_overview(self, course_run_id):
                 return overview
 
         return None
+
+    def _get_course_run_url(self, request, course_run_id):
+        """
+        Get the appropriate course url while incorporating Executive Education content.
+        """
+        course_run_url = get_course_run_url(request, course_run_id)
```
adamstankiewicz, line 92: @irfanuddinahmad Can you help me understand a bit better why we need conditional logic here specifically for Exec Ed when `get_course_run_url` itself calls `course_home_url`, which has support for the recently added `CourseHomeUrlCreationStarted` filter?

Does that filter not get applied when `course_home_url` is used via edx-enterprise and its use of `get_course_run_url` here? If it does get applied already, do we need any conditional logic here?

irfanuddinahmad: @adamstankiewicz We need the active enterprise for the user to get the [org_id](https://github.com/openedx/edx-enterprise/blob/882bf06279b6e5709c6765cf3f5c61e294b3e85b/enterprise/models.py#L195) ... that information is not available to the [filter pipeline](https://github.com/edx/federated-content-connector/blob/806416bc674de6d98ef054333c52d25b6cb5c6f5/federated_content_connector/filters/pipeline.py#L27)

adamstankiewicz: @irfanuddinahmad Ahh, yes, sorry... I think I meant to ask / was wondering more explicitly about whether there was a way around needing to rely on `exec_ed_base_url = getattr(settings, 'EXEC_ED_LANDING_PAGE', None)`. (Excuse the thinking aloud as I type below haha)

But, I think the problem you're likely running into is that this method does not have easy access to knowing whether the course is a GetSmarter Exec Ed course. You're using the resultant URL returned by `get_course_run_url` to match `EXEC_ED_LANDING_PAGE` setting as is. An alternative would be retrieving course data via `get_course_data` and have a conditional based on the resulting `course_type` and `product_source`.

But, in that case, the [filter](https://github.com/edx/federated-content-connector/blob/806416bc674de6d98ef054333c52d25b6cb5c6f5/federated_content_connector/filters/pipeline.py#L32) is already calling `get_course_data` and uses a similar conditional around the same `EXEC_ED_LANDING_PAGE` setting.

Given the setting `EXEC_ED_LANDING_PAGE` exists in edx-platform and is already accessible in both the filter and this edx-enterprise API, I think it makes the most sense how you have it ✅ 

IF we were adding a net-new `EXEC_ED_LANDING_PAGE` setting solely for edx-enterprise (not the same one used between both edx-enterprise and edx-platform), I'd push more for the `get_course_data` and needing a conditional.

Long story short, because `EXEC_ED_LANDING_PAGE` is not a net-new setting here (already exists and is accessible by edx-enterprise to support the filter anyways), I feel it should be fine to use `EXEC_ED_LANDING_PAGE` here to flag on whether the url is for GetSmarter or not.

zamanafzal (approved):

PR #1767, opened 2023-06-07
feat: unique constraint on transmission audits to prevent duplicates

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-7053

**Merge checklist:**

**Post merge:**


integrated_channels/integrated_channel/migrations/0028_alter_contentmetadataitemtransmission_unique_together.py
```diff
@@ -0,0 +1,42 @@
...
+                state_operations=[
+                    migrations.AlterUniqueTogether(
+                        name='contentmetadataitemtransmission',
+                        unique_together={('integrated_channel_code', 'plugin_configuration_id', 'content_id')},
+                    ),
+                ],
+                database_operations=[
+                    migrations.RunSQL(sql="""
+                        ALTER TABLE integrated_channel_contentmetadataitemtransmission ADD CONSTRAINT
+                        integrated_channel_contentmetadataitemtransmission_integrated_channel_code_plugin_configuration_id_content_id_4a1f9a6a_uniq
```
johnnagro, line 31: this index name is almost certainly too long for mysql57 - try installing this branch of edx-enterprise into devstack lms and run migrations, i bet you get an error

johnnagro (changes requested):

integrated_channels/integrated_channel/migrations/0028_alter_contentmetadataitemtransmission_unique_together.py
johnnagro: looks like 64 char limit https://dev.mysql.com/doc/refman/5.7/en/identifier-length.html

PR #1779, opened 2023-06-22
feat: adding managent command to clear error state

kiram15:
**Merge checklist:**
- x ] `make static` has been run to update webpack bundling if any static content was updated

**Post merge:**

### DIFF
--- a/enterprise/management/commands/update_config_last_errored_at.py
+++ b/enterprise/management/commands/update_config_last_errored_at.py
@@ -73,7 +73,7 @@ class Command(IntegratedChannelCommandMixin, BaseCommand):
         """
         try:
             has_learner_errors, has_content_errors = True, True
-            yesterday = datetime.now() - timedelta(days=1)
+            yesterday = datetime.utcnow() - timedelta(days=1)
             for channel_code, (ConfigModel, LearnerAuditModel) in MODELS.items():
                 configs = ConfigModel.objects.all()
                 for config in configs:
@@ -108,7 +108,7 @@ class Command(IntegratedChannelCommandMixin, BaseCommand):
                         LOGGER.info(
                             'Config with id {}, channel code {}, enterprise customer {}'
                             ' error information has been updated'.format(
-                            config.id, channel_code, config.enterprise_customer.uuid
+                                config.id, channel_code, config.enterprise_customer.uuid
                             )
                         )
                     config.save()
@@ -123,4 +123,3 @@ class Command(IntegratedChannelCommandMixin, BaseCommand):
         LOGGER.info('Begin nulling out outdated last_sync_errored_at in configs')
         self.update_config_last_errored_at()
         LOGGER.info('Finished nulling out outdated last_sync_errored_at in configs')
-
--- a/enterprise/management/commands/update_config_last_errored_at.py
+++ b/enterprise/management/commands/update_config_last_errored_at.py
@@ -1,0 +1,117 @@
+"""
+Backfill missing audit record foreign keys.
+"""
+import logging
+from datetime import datetime, timedelta
+
+from django.core.management.base import BaseCommand
+from django.utils.translation import gettext as _
+
+from integrated_channels.blackboard.models import (
+    BlackboardEnterpriseCustomerConfiguration,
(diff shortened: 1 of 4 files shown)


enterprise/management/commands/update_config_last_errored_at.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        For each audit record kind (learner and content), find all the records in batch, and lookup
+        if they've had recent sync errors in the last day. If not, clear out the last_content_sync_errored_at
+        value associated with the configuration.
+        """
+        try:
+            has_learner_errors, has_content_errors = True, True
+            yesterday = datetime.now() - timedelta(days=1)
+            for channel_code, (ConfigModel, LearnerAuditModel) in MODELS.items():
+                configs = ConfigModel.objects.all().filter()
```
alex-sheehan-edx, line 78: is the filter here necessary?

enterprise/management/commands/update_config_last_errored_at.py
```diff
@@ -0,0 +1,117 @@
...
+                        enterprise_customer=config.enterprise_customer,
+                        integrated_channel_code=channel_code,
+                        plugin_configuration_id=plugin_id,
+                        api_response_status_code__gt=299
+                    )
+                    if not errored_content_audits:
+                        config.last_content_sync_errored_at = None
+                        has_content_errors = False
+
+                    if not has_learner_errors and not has_content_errors:
```
alex-sheehan-edx, line 106: why is `last_sync_errored_at` dependent on both no learner and content errors whereas the `last_learner_sync_errored_at` is only dependent on learner errors?

alex-sheehan-edx (commented):
mostly good, just a few implementation q's!

tests/test_management.py
```diff
@@ -1930,3 +1931,77 @@ def test_normal_run(self):
...
+        assert csod_config.last_sync_errored_at is None
+        assert csod_config.last_content_sync_errored_at is None
+        assert csod_config.last_learner_sync_errored_at is None
+
+    def test_invalid_audits(self):
+        """
+        Verify that the management command runs when all records have the same subdomain
+        """
+        old_timestamp = datetime.now() - timedelta(days=5)
+        print('before ', old_timestamp)
```
alex-sheehan-edx, line 1985: cruft

tests/test_management.py
```diff
@@ -1930,3 +1931,77 @@ def test_normal_run(self):
...
+        old_timestamp = old_timestamp.replace(tzinfo=pytz.UTC)
+        print('after ', old_timestamp)
```
alex-sheehan-edx, line 1987: more cruft `:all-the-cruft:`

enterprise/management/commands/update_config_last_errored_at.py
```diff
@@ -0,0 +1,117 @@
...
+                        config.last_sync_errored_at = None
+                    config.save()
```
alex-sheehan-edx, line 108: Do we want to log some context here?

alex-sheehan-edx (commented):
sorry, couple more questions!

enterprise/management/commands/update_config_last_errored_at.py
kiram15: So there's 3: last, last_learner, and last_content. We use them all in different contexts, but as you said last sync is dependent on both content and learner to show the user whats the most recent error regardless of type.

enterprise/management/commands/update_config_last_errored_at.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def update_config_last_errored_at(self):
+        """
+        For each audit record kind (learner and content), find all the records in batch, and lookup
+        if they've had recent sync errors in the last day. If not, clear out the last_content_sync_errored_at
+        value associated with the configuration.
+        """
+        try:
+            has_learner_errors, has_content_errors = True, True
+            yesterday = datetime.now() - timedelta(days=1)
```
alex-sheehan-edx, line 76: should this be UTC?

alex-sheehan-edx (approved):
One question about timezones, and as long as you address the linter, this LGTM! 😄

PR #1780, opened 2023-06-23
feat: add button to update customer modified time

katrinan029:
## Description
Currently, to force a customer’s catalog to be retransmitted (updated) through the integrated channels, our team relies on the modified at times of the customer’s catalog object in the enterprise catalog service. While this is perfectly fine to use for our immediate team members, we’d like to have a more straight forward and intuitive approach. 

## Solution
A button to customers' integrated channel configuration admin page that will trigger an update on all content on the next transmission run by updating the the enterprise customer’s modified time.
Screenshot 2023-06-22 at 10 24 41 PM

On success, we will notify the user that the modified time was updated.
Screenshot 2023-06-23 at 11 41 38 AM

On error, we will notify the user that the modified time was not successfully updated.
Screenshot 2023-06-23 at 11 42 17 AM

[JIRA ENT-2713](https://2u-internal.atlassian.net/browse/ENT-6713)

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/blackboard/admin/__init__.py
+++ b/integrated_channels/blackboard/admin/__init__.py
@@ -91,20 +91,16 @@ class BlackboardEnterpriseCustomerConfigurationAdmin(DjangoObjectActions, admin.
             obj.enterprise_customer.save()
             messages.success(
                 request,
-                "The blackboard enterprise customer content metadata "
-                "“<BlackboardEnterpriseCustomerConfiguration for Enterprise {enterprise_name}>” "
-                "was updated successfully.".format(
-                    enterprise_name=obj.enterprise_customer.name
-                ),
+                f'''The blackboard enterprise customer content metadata
+                “<BlackboardEnterpriseCustomerConfiguration for Enterprise
+                {obj.enterprise_customer.name}>” was updated successfully.''',
             )
         except ValidationError:
             messages.error(
                 request,
-                "The blackboard enterprise customer content metadata "
-                "“<BlackboardEnterpriseCustomerConfiguration for Enterprise {enterprise_name}>” "
-                "was not updated successfully.".format(
-                    enterprise_name=obj.enterprise_customer.name
-                ),
+                f'''The blackboard enterprise customer content metadata
+                “<BlackboardEnterpriseCustomerConfiguration for Enterprise
+                {obj.enterprise_customer.name}>” was not updated successfully.''',
             )
         return HttpResponseRedirect(
             "/admin/blackboard/blackboardenterprisecustomerconfiguration"
--- a/integrated_channels/blackboard/admin/__init__.py
+++ b/integrated_channels/blackboard/admin/__init__.py
@@ -3,7 +3,10 @@ Admin integration for configuring Blackboard app to communicate with Blackboard 
 """
 from config_models.admin import ConfigurationModelAdmin
 
-from django.contrib import admin
+from django.contrib import admin, messages
+from django.core.exceptions import ValidationError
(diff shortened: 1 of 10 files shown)


integrated_channels/blackboard/admin/__init__.py
```diff
@@ -78,6 +82,39 @@ def customer_oauth_authorization_url(self, obj):
...
+                ),
+            )
+        return HttpResponseRedirect(
+            "/admin/blackboard/blackboardenterprisecustomerconfiguration"
+        )
+    update_modified_time.label = "Update Customer Modified Time"
+    update_modified_time.short_description = (
+        "Update modified time for this Enterprise Customer "
+    )
+    "to retransmit courses metadata"
```
alex-sheehan-edx, line 116: This probably has to go inside the parentheses?  😂

integrated_channels/canvas/admin/__init__.py
```diff
@@ -62,6 +66,39 @@ def customer_oauth_authorization_url(self, obj):
...
+                ),
+            )
+        return HttpResponseRedirect(
+            "/admin/canvas/canvasenterprisecustomerconfiguration"
+        )
+    update_modified_time.label = "Update Customer Modified Time"
+    update_modified_time.short_description = (
+        "Update modified time for this Enterprise Customer "
+    )
+    "to retransmit courses metadata"
```
alex-sheehan-edx, line 100: Same here 😄

integrated_channels/cornerstone/admin/__init__.py
```diff
@@ -69,6 +72,39 @@ def enterprise_customer_name(self, obj):
...
+                ),
+            )
+        return HttpResponseRedirect(
+            "/admin/cornerstone/cornerstoneenterprisecustomerconfiguration"
+        )
+    update_modified_time.label = "Update Customer Modified Time"
+    update_modified_time.short_description = (
+        "Update modified time for this Enterprise Customer "
+    )
+    "to retransmit courses metadata"
```
alex-sheehan-edx, line 106: as well as all the other channels 😁

alex-sheehan-edx (commented):
Looks fantastic! Slight nit on your button descriptions and also I would maybe suggest renaming the button to "force content metadata transmission" or something like that as that is closer to the intended use case of the button, rather than what it's doing under the hood.

integrated_channels/blackboard/admin/__init__.py
katrinan029: good catch! I used Black to auto formatter because the linters were complaining about long lines and I completely overlooked this. It's odd that this still passed all the lint and code style checks.

integrated_channels/canvas/admin/__init__.py
```diff
@@ -62,6 +66,38 @@ def customer_oauth_authorization_url(self, obj):
...
+        Updates the modified time of the customer record to retransmit courses metadata
+        and redirects to configuration view with success or error message.
+        """
+        try:
+            obj.enterprise_customer.save()
+            messages.success(
+                request,
+                "The canvas enterprise customer content metadata "
+                "“<CanvasEnterpriseCustomerConfiguration for Enterprise {enterprise_name}>” "
+                "was updated successfully.".format(
```
alex-sheehan-edx, line 80: (tiny nit) Generally we've moved from python's string `format` to `f-strings` when interpolating

alex-sheehan-edx (approved):
Talked offline about format vs fstrings, once that's fixed up this PR is a 👍 from me. Don't forget your CHANGELOG and enterprise init version!

PR #1783, opened 2023-06-27
feat: marking orphaned content audits when catalogs are deleted

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-7329

**Merge checklist:**

**Post merge:**


enterprise/signals.py
```diff
@@ -30,6 +38,15 @@
...
 _UNSAVED_FILEFIELD = 'unsaved_filefield'
+INTEGRATED_CHANNELS = [
+    BlackboardEnterpriseCustomerConfiguration,
+    CanvasEnterpriseCustomerConfiguration,
+    CornerstoneEnterpriseCustomerConfiguration,
+    DegreedEnterpriseCustomerConfiguration,
+    Degreed2EnterpriseCustomerConfiguration,
+    MoodleEnterpriseCustomerConfiguration,
+    SAPSuccessFactorsEnterpriseCustomerConfiguration,
+]
```
katrinan029, line 49: question: I see that this list doesn't include xapi and I'm assuming that's intentional. What is xapi and why would we exclude it?

katrinan029 (approved):

enterprise/signals.py
alex-sheehan-edx: xapi is sort of the forgotten stepchild of the integrated channels 😆 No one's worked on it for quite some time and It doesn't go through the normal transmission sync process as the other channels as it serves more as a way for people to request the integrated channels transmission data rather than receive it from our integration. I tend to think of xapi as a series of api end points rather than the system that pushes content out to integrated customers.

enterprise/signals.py
```diff
@@ -315,6 +332,15 @@ def delete_enterprise_catalog_data(sender, instance, **kwargs):     # pylint: di
             exc_info=exc
         )
 
+    customer = instance.enterprise_customer
+    for channel in INTEGRATED_CHANNELS:
+        if channel.objects.filter(enterprise_customer=customer, active=True).exists():
+            try:
+                mark_orphaned_content_metadata_audit.delay()
+                continue
```
iloveagent57, line 340: could we actually `break` here instead?  this audit job runs for metadata across all channels, right?

alex-sheehan-edx: yes it does

enterprise/signals.py
```diff
@@ -315,6 +332,15 @@ def delete_enterprise_catalog_data(sender, instance, **kwargs):     # pylint: di
...
         )
 
+    customer = instance.enterprise_customer
+    for channel in INTEGRATED_CHANNELS:
+        if channel.objects.filter(enterprise_customer=customer, active=True).exists():
+            try:
+                mark_orphaned_content_metadata_audit.delay()
+                break
+            except Exception as exc:  # pylint: disable=broad-except
+                logger.exception(f'Failed to mark orphaned content metadata audits. Task failed with exception: {exc}')
```
iloveagent57, line 342: nit: you won't actually know if the task failed in this context - that task is running asynchronously off in a worker process somewhere else.  The only kind of exception we'd catch here is about a problem submitting the task - maybe update the log message to reflect that?  But just straight invoking the task without this `try/except` would be fine, too.  We probably want this to fail loudly since it's not user-facing and is a very sparsely occurring event.

iloveagent57 (approved):

enterprise/signals.py
alex-sheehan-edx: I'm fine just taking the try/except out

PR #1786, opened 2023-06-28
chore: upgrade course_enrollment from audit to verified

emrosarioa:
## Description

If an enterprise learner has enrolled in the audit course mode, we would like to support an automatic upgrade to the paid (verified) enrollment track.
In [edx-platform](https://github.com/openedx/edx-platform/pull/32595), the `lms_enroll_user_in_course` function is updated to `lms_update_or_create_enrollment`to allow for an update to the course_mode.
This PR updates the name of the function and any related comments.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -46,9 +46,13 @@ from enterprise.constants import (
 from enterprise.logging import getEnterpriseLogger
 
 try:
-    from openedx.features.enterprise_support.enrollments.utils import lms_enroll_user_in_course
+    from openedx.features.enterprise_support.enrollments.utils import (
+        lms_enroll_user_in_course,
+        lms_update_or_create_enrollment,
+    )
 except ImportError:
     lms_enroll_user_in_course = None
+    lms_update_or_create_enrollment = None
 
 try:
     from openedx.core.djangoapps.course_groups.models import CourseUserGroup
@@ -1792,17 +1796,30 @@ def customer_admin_enroll_user_with_status(
     succeeded = False
     new_enrollment = False
     enterprise_fulfillment_source_uuid = None
+    emet_enable_auto_upgrade_enrollment_mode = getattr(
+        settings,
+        'ENABLE_ENTERPRISE_BACKEND_EMET_AUTO_UPGRADE_ENROLLMENT_MODE',
+        False,
+    )
     try:
-        # enrolls a user in a course per LMS flow, but this method does not create enterprise records yet so we need to
-        # create it immediately after calling lms_enroll_user_in_course. lms_enroll_user_in_course can return None if
-        # Enrollment already exists, does not fail in this case.
-        new_enrollment = lms_enroll_user_in_course(
-            user.username,
-            course_id,
-            course_mode,
-            enterprise_customer.uuid,
-            is_active=True,
-        )
+        # enrolls a user in a course per LMS flow, but this method doesn't create enterprise records
+        # yet so we need to create it immediately after calling lms_update_or_create_enrollment.
+        if emet_enable_auto_upgrade_enrollment_mode:
+            new_enrollment = lms_update_or_create_enrollment(
+                user.username,
+                course_id,
+                course_mode,
+                is_active=True,
+                enterprise_uuid=enterprise_customer.uuid,
+            )
+        else:
+            new_enrollment = lms_enroll_user_in_course(
+                user.username,
+                course_id,
+                course_mode,
+                enterprise_customer.uuid,
+                is_active=True,
+            )
         succeeded = True
         LOGGER.info("Successfully enrolled user %s in course %s", user.id, course_id)
     except (CourseEnrollmentError, CourseUserGroup.DoesNotExist) as error:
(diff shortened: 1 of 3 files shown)


pwnage101 (approved):

enterprise/utils.py
```diff
@@ -1792,17 +1796,31 @@ def customer_admin_enroll_user_with_status(
     succeeded = False
     new_enrollment = False
     enterprise_fulfillment_source_uuid = None
+    emet_enable_auto_upgrade_enrollment_mode = get_configuration_value(
+        'FEATURE_ENABLE_EMET_AUTO_UPGRADE_ENROLLMENT_MODE',
```
adamstankiewicz, line 1800: [clarification] Where does `get_configuration_value` pull its config settings from? I believe `FEATURE_ENABLE_EMET_AUTO_UPGRADE_ENROLLMENT_MODE` is only defined for the MFE configuration API under the `MFE_CONFIG_OVERRIDES` setting, which afaik is different than where `get_configuration_value` pulls from.

Out of curiosity, I'm wondering whether we _need_ to introduce a conditional feature flag for supporting audit -> verified enrollment upgrades? What is the harm/risk in not including a feature flag on the API side of things? 

The only reason the audit -> verified upgrade code path has a feature flag with `FEATURE_ENABLE_EMET_AUTO_UPGRADE_ENROLLMENT_MODE` is because the API didn't support it yet. But, if the API supports it (without a feature flag), in theory I believe we could then remove `FEATURE_ENABLE_EMET_AUTO_UPGRADE_ENROLLMENT_MODE` altogether on the frontend side of things. 

For example, we might implement these API changes without a feature flag, enable the feature flags on stage then production, and then once verified, remove `FEATURE_ENABLE_EMET_AUTO_UPGRADE_ENROLLMENT_MODE` altogether from frontend-app-learner-portal-enterprise.

emrosarioa: I think you are correct that `get_configuration_value` won't have access to `FEATURE_ENABLE_EMET_AUTO_UPGRADE_ENROLLMENT_MODE`, so will revert that change!
I remember that there will be some time between when the changes in `edx-enterprise` are available in the new release and when the changes are available in staging or prod in `edx-platform`. I wanted to be able to fall back onto the old `lms_enroll_user_in_course` function while there is a bit of time misalignment between the two deployments. Perhaps a feature flag on the API side would work best?

adamstankiewicz: Following up from Slack conversation about this feature flag, I didn't realize the coupling between edx-platform and edx-enterprise at the time. I see why introducing a temporary feature flag here makes sense :) Feel free to ignore this comment thread now.

enterprise/utils.py
```diff
@@ -1792,17 +1796,32 @@ def customer_admin_enroll_user_with_status(
...
+        # enrolls a user in a course per LMS flow, but this method doesn't create enterprise records
+        # yet so we need to create it immediately after calling lms_update_or_create_enrollment.
+        # lms_update_or_create_enrollment can return None if Enrollment already exists,
+        # does not fail in this case.
+        if emet_enable_auto_upgrade_enrollment_mode:
+            new_enrollment = lms_update_or_create_enrollment(
+                user.username,
+                course_id,
+                course_mode,
+                True,
```
adamstankiewicz, line 1814: nit: I wonder if it may be slightly more readable to include the kwarg property key when passing `True` so that the code is more self-descriptive/documented (e.g., `is_active=True` below).

tests/test_enterprise/api/test_views.py
```diff
@@ -4499,7 +4501,8 @@ def test_bulk_enrollment_in_bulk_courses_nonexisting_user_id(
     @mock.patch(
         'enterprise.api.v1.views.enterprise_customer.get_best_mode_from_course_key'
     )
-    @mock.patch("enterprise.utils.lms_enroll_user_in_course")
+    @mock.patch('enterprise.utils.lms_update_or_create_enrollment')
+    @override_settings(ENABLE_ENTERPRISE_BACKEND_EMET_AUTO_UPGRADE_ENROLLMENT_MODE=True)
```
adamstankiewicz, line 4505: [suggestion] If there's a chance of having the feature flag disabled on production still, it might be worth having test coverage when `ENABLE_ENTERPRISE_BACKEND_EMET_AUTO_UPGRADE_ENROLLMENT_MODE ` is both `True` and `False` to assert both code paths still work separately. Otherwise, I believe we're losing coverage on the `else` path now?

tests/test_enterprise/test_utils.py
```diff
@@ -97,14 +98,15 @@ def test_get_platform_logo_url(self, logo_url, expected_logo_url, mock_get_logo_
         mock_get_logo_url.return_value = logo_url
         self.assertEqual(get_platform_logo_url(), expected_logo_url)
 
-    @mock.patch('enterprise.utils.lms_enroll_user_in_course')
+    @override_settings(ENABLE_ENTERPRISE_BACKEND_EMET_AUTO_UPGRADE_ENROLLMENT_MODE=True)
+    @mock.patch('enterprise.utils.lms_update_or_create_enrollment')
```
adamstankiewicz, line 102: [suggestion] Similar to the previous comment, I wonder if we should have test cases covering both the feature flag being enabled and disabled to assert on both code baths while both still exist.

PR #1748, opened 2023-05-07
feat: handle enrollments for "Invite Only" courses

tecoholic:
### Description

This PR introduces some improvements to the 'Manage Learners' page of Enterprise Customer admin that will allow enrolling learners into "Invite Only" courses.

Refer #1745 for full details about the issue.

### Testing instructions

* Follow the testing instructions in #1745 and reproduce the issue
* Switch to the PR branch
* Try enrolling learners into the course marked as "Invite Only", with the *Force Enrollment* checkbox checked, this time it should work without throwing an error.
* Test with both a learner who already has a registered account and an email which is not registered in the system to ensure, "enrollment on registration" also works for "Invite Only" courses.

image

**Merge checklist:**
- [ ] Any new requirements are in the right place (do **not** manually modify the `requirements/*.txt` files) - **No new requirements added**
    - `base.in` if needed in production but edx-platform doesn't install it
    - `test-master.in` if edx-platform pins it, with a matching version
    - `make upgrade && make requirements` have been run to regenerate requirements
- [ ] `make static` has been run to update webpack bundling if any static content was updated - **Running make static didn't introduce any changes**
- [ ] `./manage.py makemigrations` has been run - **No DB migrations introduced**
    - Checkout the [Database Migration](https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations) Confluence page for helpful tips on creating migrations.
    - *Note*: This **must** be run if you modified any models.
      - It may or may not make a migration depending on exactly what you modified, but it should still be run.
    - This should be run from either a venv with all the lms/edx-enterprise requirements installed or if you checked out edx-enterprise into the src directory used by lms, you can run this command through an lms shell.
        - It would be `./manage.py lms makemigrations` in the shell.
- [ ] [Version](https://github.com/openedx/edx-enterprise/blob/master/enterprise/__init__.py) bumped
- [ ] [Changelog](https://github.com/openedx/edx-enterprise/blob/master/CHANGELOG.rst) record added
- [ ] Translations updated (see docs/internationalization.rst but also this isn't blocking for merge atm)

**Post merge:**
- [ ] Tag pushed and a new [version](https://github.com/openedx/edx-enterprise/releases) released
    - *Note*: Assets will be added automatically. You just need to provide a tag (should match your version number) and title and description.
[...]

PR #1790, opened 2023-07-05
chore: version bump

e0d:
Work to understand and resolve the doc errors on #1748


tecoholic:
@e0d Thanks a lot for moving this over and handling it.

PR #1792, opened 2023-07-06
feat: discovery work into implementing integrated skills content

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-7324

**Merge checklist:**

**Post merge:**


alex-sheehan-edx:
My primary concerns/questions to reviewers are: 
1) where to implement channel specific support for skills
meaning I'm thinking about where do we cut off unsupported channels in the skill exporting/transmitting process? (re https://github.com/openedx/edx-enterprise/pull/1792/files#diff-bd1856bd7ffab338eb216ef940de368327f49152f8673cc007eac15557220c78R592).
2) Is [my](https://github.com/openedx/edx-enterprise/pull/1792/files#diff-8e527cbda17b71cd26cccce8c4c077fbd4a731dc04ac83a1601a7efc57b4d451R623)  `associated_catalog_keys = models.ManyToManyField` approach to the skills transmission audit table efficient enough to support the [look ups in the exporter](https://github.com/openedx/edx-enterprise/pull/1792/files#diff-bd1856bd7ffab338eb216ef940de368327f49152f8673cc007eac15557220c78R595-R631)  ?

integrated_channels/integrated_channel/models.py
```diff
@@ -393,6 +393,13 @@ def transmit_content_metadata(self, user):
         transmitter = self.get_content_metadata_transmitter()
         transmitter.transmit(*exporter.export())
 
+        # Instead of ^
+
+        # create, update, delete, skills_metadata, relations_metadata = exporter.export()
+        # transmitter.transmit(create, update, delete)
+        # transmitter.transmit_skills(skills_metadata)
+        # transmitter.transmit_skill_associations(relations_metadata)
```
alex-sheehan-edx, line 401: we could also deal with channel support/order of operations here

PR #1796, opened 2023-07-07
fix: exporter sanitizing content to delete 

alex-sheehan-edx:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/constants.py
+++ b/enterprise/constants.py
@@ -204,6 +204,10 @@ HTTP_STATUS_STRINGS = {
     503: 'The server is temporarily unavailable.',
 }
 
+IC_CREATE_ACTION = 'create'
+IC_UPDATE_ACTION = 'update'
+IC_DELETE_ACTION = 'delete'
+
 
 class FulfillmentTypes:
     LICENSE = 'license'
--- a/integrated_channels/integrated_channel/exporters/content_metadata.py
+++ b/integrated_channels/integrated_channel/exporters/content_metadata.py
@@ -14,7 +14,14 @@ from django.conf import settings
 from django.db.models import Q
 
 from enterprise.api_client.enterprise_catalog import EnterpriseCatalogApiClient
-from enterprise.constants import EXEC_ED_CONTENT_DESCRIPTION_TAG, EXEC_ED_COURSE_TYPE, TRANSMISSION_MARK_CREATE
+from enterprise.constants import (
+    EXEC_ED_CONTENT_DESCRIPTION_TAG,
+    EXEC_ED_COURSE_TYPE,
+    IC_CREATE_ACTION,
+    IC_DELETE_ACTION,
+    IC_UPDATE_ACTION,
+    TRANSMISSION_MARK_CREATE,
+)
 from enterprise.utils import get_content_metadata_item_id
 from integrated_channels.integrated_channel.exporters import Exporter
 from integrated_channels.utils import generate_formatted_log, truncate_item_dicts
@@ -445,7 +452,19 @@ class ContentMetadataExporter(Exporter):
         ordered_and_chunked_orphaned_content = orphaned_content.order_by('created')[:max_set_count]
         return ordered_and_chunked_orphaned_content
 
-    # pylint: disable=too-many-statements
+    def _sanitize_and_set_item_metadata(self, item, metadata, action):
+        """
+        Helper method to sanitize and set the metadata of an audit record according to
+        the provided action being performed on the item.
+        """
+        metadata_transformed_for_exec_ed = self._transform_exec_ed_content(metadata)
+        transformed_item = self._transform_item(metadata_transformed_for_exec_ed, action=action)
+
+        item.channel_metadata = transformed_item
+        item.content_title = transformed_item.get('title')
+        item.content_last_changed = transformed_item.get('content_last_modified')
+        item.save()
+
     def export(self, **kwargs):
         """
         Export transformed content metadata if there has been an update to the consumer's catalogs
@@ -454,7 +473,7 @@ class ContentMetadataExporter(Exporter):
             self.enterprise_customer.enterprise_customer_catalogs.all()
 
         # a maximum number of changes/payloads to export at once
-        # default to something huge to simplifly logic, the max system int size
+        # default to something huge to simplify logic, the max system int size
         max_payload_count = kwargs.get('max_payload_count', sys.maxsize)
 
         self._log_info(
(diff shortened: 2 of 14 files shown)


enterprise/constants.py
```diff
@@ -204,6 +204,9 @@ def json_serialized_course_modes():
     503: 'The server is temporarily unavailable.',
 }
 
+IC_CREATE_ACTION = 'create'
+IC_UPDATE_ACTION = 'update'
```
alex-sheehan-edx, line 208: at this point I don't think these are used but figured I'd finish the set

integrated_channels/integrated_channel/models.py
```diff
@@ -251,12 +251,16 @@ def fetch_orphaned_content_audits(self):
...
+        non_existent_catalogs_filter = Q(enterprise_customer_catalog_uuid__in=customer_catalog_uuids)
+        null_catalogs_filter = Q(enterprise_customer_catalog_uuid__isnull=True)
+
         return ContentMetadataItemTransmission.objects.filter(
             integrated_channel_code=self.channel_code(),
             enterprise_customer=self.enterprise_customer,
             remote_deleted_at__isnull=True,
             remote_created_at__isnull=False,
-        ).exclude(enterprise_customer_catalog_uuid__in=customer_catalog_uuids)
+        ).filter(~non_existent_catalogs_filter | null_catalogs_filter)
```
alex-sheehan-edx, line 263: we figured out that this was working as intended, but I do think this is more readable.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -533,8 +538,26 @@ def export(self, **kwargs):
                         existing_record.channel_metadata = transformed_item
                         update_payload[key] = existing_record
 
-            for key, item in items_to_delete.items():
-                delete_payload[key] = item
+            if items_to_delete:
```
alex-sheehan-edx, line 541: If we want to sanitize the item to be deleted we have to now request the metadata for it.

katrinan029 (approved):
LGTM! Thank you for walking through these changes with me.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -492,48 +511,44 @@ def export(self, **kwargs):
...
+            for key, item in items_to_update.items():
+                metadata_transformed_for_exec_ed = self._transform_exec_ed_content(
+                    key_to_content_metadata_mapping[key]
+                )
+                transformed_item = self._transform_item(metadata_transformed_for_exec_ed, action=IC_UPDATE_ACTION)
+
+                item.content_title = transformed_item.get('title')
+                item.enterprise_customer_catalog_uuid = enterprise_customer_catalog.uuid
+                item.save()
+                # Intentionally setting the channel_metadata and content_last_changed
```
johnnagro, line 541: is there a way where we don't need to do this? i feel like we shouldn't encode the sync status into the presence or absence of the `content_last_changed` field - now that we have a more robust sync status on the audit records, can we just always set this field?

integrated_channels/sap_success_factors/exporters/content_metadata.py
```diff
@@ -52,12 +53,12 @@ def transform_provider_id(self, content_metadata_item):  # pylint: disable=unuse
...
         return self.enterprise_configuration.provider_id
 
-    def transform_status(self, content_metadata_item):
+    def transform_for_action_status(self, content_metadata_item, action):
         """
         Return the status of the content item.
         """
         # lets not overwrite something we've already tried to set INACTIVE
-        if content_metadata_item.get('status') == 'INACTIVE':
+        if content_metadata_item.get('status') == 'INACTIVE' or action == IC_DELETE_ACTION:
```
johnnagro, line 61: i believe you only need to branch on `action == IC_DELETE_ACTION` - i'd remove the `content_metadata_item.get('status') == 'INACTIVE'` part. this was added to deal with the old way of doing things where we were setting the status in the transmit and we didnt want the transformer to stomp on that change.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -14,7 +14,14 @@
 from django.db.models import Q
 
 from enterprise.api_client.enterprise_catalog import EnterpriseCatalogApiClient
-from enterprise.constants import EXEC_ED_CONTENT_DESCRIPTION_TAG, EXEC_ED_COURSE_TYPE, TRANSMISSION_MARK_CREATE
+from enterprise.constants import (
+    EXEC_ED_CONTENT_DESCRIPTION_TAG,
+    EXEC_ED_COURSE_TYPE,
+    IC_DELETE_ACTION,
+    IC_UPDATE_ACTION,
+    IC_CREATE_ACTION,
```
katrinan029, line 22: I think this need to be reordered to fix the lint error.

```
IC_CREATE_ACTION,
IC_DELETE_ACTION,
IC_UPDATE_ACTION,
```

johnnagro (changes requested):
the approach looks good. i would recommend a few tweaks. i'd love to know if we can factor out that need to hold back on the content_last_modified timestamp.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -596,25 +611,21 @@ def _transform_item(self, content_metadata_item):
...
-                    'transform_{edx_data_schema_key}'.format(
-                        edx_data_schema_key=edx_data_schema_key
-                    ),
-                    None
-                )
+                getattr(self, f'transform_{edx_data_schema_key}', None)
             )
+
+            transformer_for_action = (
+                getattr(self, f'transform_for_action_{content_metadata_type}_{edx_data_schema_key}', None)
```
katrinan029, line 620: quick question: In what instance will this be used? Where is this attribute defined?

alex-sheehan-edx: The intention's to go into the channel specific content metadata exporter and use the defined mappings to tailor and sanitize the code (see https://github.com/openedx/edx-enterprise/blob/master/integrated_channels/sap_success_factors/exporters/content_metadata.py). There we have two different types of mapping methods, `transform_{attribute}` and `transform_{content_type}_{attribute}` because some external platforms care about and want to distinguish between metadata depending on what type of content it is, ie SAP wants different titles for course runs than for course and programs. In this case we're supporting the old implementation while also allowing for the caller to specify more detail/actions.  `transform_{content_type}_{attribute}` isn't currently being used by any channel yet but I wanted to add the functionality as we know channels do this content specific mapping already and we may want to support it in an upgrade in the future.

integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -586,7 +601,7 @@ def _transform_exec_ed_content(self, content):
                 content['full_description'] = EXEC_ED_CONTENT_DESCRIPTION_TAG + description
         return content
 
-    def _transform_item(self, content_metadata_item):
+    def _transform_item(self, content_metadata_item, action=None):
```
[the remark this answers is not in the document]
alex-sheehan-edx: ahhh yes we did talk about that didn't we

PR #1798, opened 2023-07-12
chore: fix get_default_site function of Enterprise Customer admin page

yuanyuan-git-tech:
https://2u-internal.atlassian.net/browse/ENT-7322
Screenshot 2023-07-12 at 3 27 04 PM
Screenshot 2023-07-12 at 3 27 14 PM
Screenshot 2023-07-13 at 11 34 24 AM
Screenshot 2023-07-12 at 3 27 33 PM
**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -157,6 +158,22 @@ def get_default_customer_type():
...
 
 
+def get_default_site():
+    """
+    Get default site id to use when creating a new EnterpriseCustomer model.
+    The default value depending on what environment the person is in.
+    In production, it should be 'courses.edx.org'.
+    In stage it should be 'courses.stage.edx.org'.
+    """
+    ENVIRONMENT = os.getenv('DJANGO_ENVIRONMENT', 'development')
```
johnnagro, line 168: rather than hard-coding the options here, we should add a configuration setting and/or reuse an existing one such as [LMS_BASE](https://github.com/openedx/edx-platform/blob/cb3a0e91e9b3e39ad5817c5e04abc8c1c9ce0f73/cms/envs/common.py#L770)

That way is more flexible - it can run in any environment (devstack, hosted devstack, sandboxes, etc) where the domains might vary widely, as well as any opensource community install.

johnnagro (changes requested):

enterprise/migrations/0177_auto_20230712_1925.py
```diff
@@ -0,0 +1,33 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('enterprise', '0176_auto_20230712_1751'),
+    ]
+
+    operations = [
```
alex-sheehan-edx, line 12: is there a reason why these changes are in a second migration file?

enterprise/models.py
```diff
@@ -157,6 +157,18 @@ def get_default_customer_type():
     return enterprise_customer_type.id
 
 
+def get_default_site():
```
alex-sheehan-edx, line 160: 👍

enterprise/models.py
```diff
@@ -223,41 +238,44 @@ class Meta:
     )
 
     enable_data_sharing_consent = models.BooleanField(
+        "Activate data sharing consent prompt",
```
alex-sheehan-edx, line 241: it wasn't clear to me that this param is intended as the `verbose_name`, could we add the named param to this and the other fields below that include verbose names?

alex-sheehan-edx (commented):
Looks good! I just have a couple quick comments/questions before we merge. Also don't forget to add your change log and version bumb!

alex-sheehan-edx (approved):
Make sure you add a change log and bump your version before you merge!

johnnagro (approved):

PR #1799, opened 2023-07-12
docs: Replace paver update_db.

dianakhuang:
We are planning on removing paver, and
we have created an alternative Make command for
running migrations. This updates the documentation.

https://github.com/openedx/devstack/issues/1085

I haven't bumped the version or added a changelog entry because this is a minor docs change.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -146,7 +146,7 @@ as an example) and run migrations:
 
 .. code-block:: bash
 
-    $ paver update_db
+    $ make migrate
     # Or use a more down-to-the-root command (replace aws with your version of config)
     $ ./manage.py lms migrate --settings=devstack
 
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -15,7 +15,7 @@ perform database migrations.
 
 .. code-block:: bash
 
-    $ paver update_db
+    $ make migrate
 
     # Or use a more down-to-the-root command (replace aws with your version of config)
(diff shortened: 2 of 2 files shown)


docs/development.rst
```diff
@@ -146,7 +146,7 @@ as an example) and run migrations:
 
 .. code-block:: bash
 
-    $ paver update_db
+    $ make migrate
     # Or use a more down-to-the-root command (replace aws with your version of config)
```
robrap, line 150: Not part of your PR, but what is meant by `aws` here?

robrap (approved):

docs/development.rst
robrap: Also, I just remembered that the old call had `paver install_prereqs` baked in. Does that need to be added here?

macdiesel (approved):

docs/development.rst
dianakhuang: We have `make requirements`, but I don't think this documentation requires the requirements to be installed. (`aws` is the old name of the `production` settings, I believe.)

robrap: 1. This comment doesn't make sense in this file, because we're using `devstack`, and it probably doesn't need to be replaced?
2. In the other file, it probably makes sense to replace `aws` in the comment and the example command to `production`.

Since this merged already, I guess we'll see if one of us do a follow-up PR. :)

PR #1803, opened 2023-07-18
fix: removing items to delete dependency on the catalog service

alex-sheehan-edx:
https://splunk.edx.org/en-US/app/search/search?q=search%20_sanitize_and_set_item_metadata%20NYIF%2BBOPC.4x&earliest=-24h%40h&latest=now&display.page.search.mode=smart&dispatch.sample_ratio=1&sid=1689635604.3264070

follow up/iteration on https://2u-internal.atlassian.net/browse/ENT-7406

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/exporters/content_metadata.py
+++ b/integrated_channels/integrated_channel/exporters/content_metadata.py
@@ -18,7 +18,6 @@ from enterprise.constants import (
     EXEC_ED_CONTENT_DESCRIPTION_TAG,
     EXEC_ED_COURSE_TYPE,
     IC_CREATE_ACTION,
-    IC_DELETE_ACTION,
     IC_UPDATE_ACTION,
     TRANSMISSION_MARK_CREATE,
 )
@@ -511,8 +510,7 @@ class ContentMetadataExporter(Exporter):
             self._log_info(f'diff items_to_update: {items_to_update}')
             self._log_info(f'diff items_to_delete: {items_to_delete}')
 
-            content_keys_filter = list(items_to_create.keys()) + list(items_to_update.keys()) + \
-                list(items_to_delete.keys())
+            content_keys_filter = list(items_to_create.keys()) + list(items_to_update.keys())
             if content_keys_filter:
                 content_metadata_items = self.enterprise_catalog_api.get_content_metadata(
                     self.enterprise_customer,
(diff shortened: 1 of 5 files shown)


integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -537,11 +535,9 @@ def export(self, **kwargs):
 
                 update_payload[key] = item
             for key, item in items_to_delete.items():
-                self._sanitize_and_set_item_metadata(item, key_to_content_metadata_mapping[key], IC_DELETE_ACTION)
-                # Sanity check
-                item.enterprise_customer_catalog_uuid = enterprise_customer_catalog.uuid
+                metadata = self._apply_delete_transformation(item.channel_metadata)
```
alex-sheehan-edx, line 538: Because we can't request metadata of content that isn't in the customer's catalog, we have to make do with the audit's channel metadata and leave it up to the individual channel as to what needs sanitizing

integrated_channels/sap_success_factors/exporters/content_metadata.py
```diff
@@ -47,6 +47,21 @@ class SapSuccessFactorsContentMetadataExporter(ContentMetadataExporter):
...
+        Specific transformations required for "deleting" a course on a SAP external service.
+        """
+        # Applying the metadata payload update to "delete" the course on SAP instances
+        metadata['status'] = 'INACTIVE'
+
+        # Sanity check as we've seen issues with schedule structure
+        metadata_schedule = metadata.get('schedule')
+        if metadata_schedule:
+            schedule = metadata_schedule[0]
+            if not schedule.get('startDate') or not schedule.get('endDate'):
```
katrinan029, line 61: question: Do we care about the `active` property whether it's `true` or `false`?

katrinan029 (approved):

integrated_channels/sap_success_factors/exporters/content_metadata.py
alex-sheehan-edx: I don't believe so, the only comment I saw from @bcitro's original work was on start and end date being empty. But also maybe good to double check with him?

bcitro (approved):
LGTM

integrated_channels/sap_success_factors/exporters/content_metadata.py
bcitro: start and end are the only ones I've seen errors for, so I think that's sufficient to check here

PR #1812, opened 2023-07-20
feat: incorporate additional attributes to degreed v2

katrinan029:
## Description
The Degreed v2 API incorporate additional fields that can be added to the catalog transmission to enhance the current integration.

## Solution
Add the following fields:

- Cost-Units
- cost-unit-type
- difficulty
- video-url

Additionally, we incorporated the `show_course_price` config as a core attribute to all the integrated channels and set the default to `false`.

[JIRA Ticket](https://2u-internal.atlassian.net/browse/ENT-5808)

**Merge checklist:**

**Post merge:**


integrated_channels/degreed2/exporters/content_metadata.py
```diff
@@ -153,3 +162,46 @@ def transform_obsolete(self, content_metadata_item):  # pylint: disable=unused-a
...
+            advertised_course_run = get_advertised_course_run(content_metadata_item)
+            if advertised_course_run and 'first_enrollable_paid_seat_price' in advertised_course_run:
+                price = advertised_course_run.get('first_enrollable_paid_seat_price') or 0
+            else:
+                for course_run in content_metadata_item.get('course_runs', []):
+                    if 'first_enrollable_paid_seat_price' in course_run and is_course_run_active(course_run):
+                        price = course_run.get('first_enrollable_paid_seat_price') or 0
+                        break
+
+        return price
```
katrinan029, line 182: question: if Degreed is expecting the following payload

```
payload = { "data": { "attributes": {
            "external-id": "string",
            "provider-code": "string",
            "title": "string",
            "summary": "string",
            "publish-date": "2023-07-19T17:28:24.871Z",
            "url": "string",
            "obsolete": True,
            "image-url": "string",
            "language": "string",
            "duration": 0,
            "duration-type": "string",
            "cost-units": 0,
            "cost-unit-type": "string",
            "format": "string",
            "difficulty": "string",
            "video-url": "string",
            "visibility-groups": [
                {
                    "id": "string",
                    "type": "groups"
                }
            ]
        } } }
```

is this the right data structure to return?

kiram15 (approved):
LGTM, and some good testing data!

PR #1827, opened 2023-07-26
feat: add grade value to degreed v2

katrinan029:
## Description
The Degreed v2 API incorporated the option to add the value a learner's grade. 
https://developer.degreed.com/reference/post_api-v2-completions

## Solution
Add the `percentile` field to support passing the grade value.

[JIRA](https://2u-internal.atlassian.net/browse/ENT-5809)

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/degreed2/client.py
+++ b/integrated_channels/degreed2/client.py
@@ -114,7 +114,6 @@ class Degreed2APIClient(IntegratedChannelApiClient):
         Returns: status_code, response_text
         """
         json_payload = json.loads(payload)
-        json_payload['data']['attributes']['percentile'] = json_payload['data']['attributes']['percentile'] * 100
         LOGGER.info(self.make_log_msg(
             json_payload.get('data').get('attributes').get('content-id'),
             f'Attempting find course via url: {self.get_completions_url()}'),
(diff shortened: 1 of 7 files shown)


integrated_channels/degreed2/client.py
```diff
@@ -114,6 +114,7 @@ def create_course_completion(self, user_id, payload):
         Returns: status_code, response_text
         """
         json_payload = json.loads(payload)
+        json_payload['data']['attributes']['percentile'] = json_payload['data']['attributes']['percentile'] * 100
```
alex-sheehan-edx, line 117: 🤔 Wondering if this can go in the exporter

alex-sheehan-edx (approved):
If you move the data transformation in your client bit to the degreed2 exporter then you can also write a test to make sure grade percent comes out of the degreed2 `get_learner_data_records` method!  Once you get that, this looks great to me

integrated_channels/degreed2/client.py
katrinan029: good suggestion - I think it's a little cleaner moving this into the exporter as well!

integrated_channels/degreed2/exporters/learner_data.py
```diff
@@ -32,6 +32,7 @@ def get_learner_data_records(
 
         If no remote ID can be found, return None.
         """
+        percent_grade = kwargs.get('grade_percent', None)
```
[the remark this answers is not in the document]
katrinan029: Done!

PR #1831, opened 2023-07-28
feat: keycloak devstack service

johnnagro:
## Description

- a [keycloak](https://www.keycloak.org/) devstack entry for testing
- data is persisted to a local volume
- `make dev.up.keycloak` & `make dev.stop.keycloak`
- http://localhost:8080 for local testing
- `ngrok http 8080` for stage/prod testing


marlonkeating (approved):
The Keycloak instance starts up, I can log in and create entities, and the data persists after shutting down.  Looks good!

image

PR #1848, opened 2023-08-17
feat: adding event bus functionality

kiram15:
Adding event bus producer functionality to unenrollment event.
https://2u-internal.atlassian.net/browse/ENT-7560

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/signals.py
+++ b/enterprise/signals.py
@@ -4,6 +4,8 @@ Django signal handlers.
 
 from logging import getLogger
 
+from openedx_events.event_bus import get_producer
+
 from django.db.models.signals import post_delete, post_save, pre_save
 from django.dispatch import receiver
 
@@ -349,16 +351,23 @@ def enterprise_unenrollment_receiver(sender, **kwargs):     # pylint: disable=un
     """
     enrollment = kwargs.get('enrollment')
     enterprise_enrollment = models.EnterpriseCourseEnrollment.objects.filter(
-        course_id=enrollment.course.course_key,
-        enterprise_customer_user__user_id=enrollment.user.id,
+        course_id=enrollment.course_id,
+        enterprise_customer_user__user_id=enrollment.enterprise_customer_user.user_id,
     ).first()
     if enterprise_enrollment:
         logger.info(
-            f"Marking EnterpriseCourseEnrollment as unenrolled for user {enrollment.user} and "
-            f"course {enrollment.course.course_key}"
+            f"Marking EnterpriseCourseEnrollment as unenrolled for user "
+            f"{enrollment.enterprise_customer_user.user_id} and course {enrollment.course_id}"
         )
         enterprise_enrollment.unenrolled_at = localized_utcnow()
         enterprise_enrollment.save()
+        get_producer().send(
+            signal=COURSE_UNENROLLMENT_COMPLETED,
+            topic='course-unenrollment-completed',
+            event_key_field='enrollment.course.course_key',
+            event_data={'enrollment': kwargs.get('enrollment')},
+            event_metadata=kwargs.get('metadata')
+        )
 
 
 def create_enterprise_enrollment_receiver(sender, instance, **kwargs):     # pylint: disable=unused-argument
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -29,6 +29,7 @@ edx-tincan-py35
 edx-rbac
 jsondiff
 jsonfield
+openedx-events>=8.3.0
 path.py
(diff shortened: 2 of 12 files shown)


enterprise/signals.py
```diff
@@ -349,16 +351,23 @@ def enterprise_unenrollment_receiver(sender, **kwargs):     # pylint: disable=un
...
-            f"Marking EnterpriseCourseEnrollment as unenrolled for user {enrollment.user} and "
-            f"course {enrollment.course.course_key}"
+            f"Marking EnterpriseCourseEnrollment as unenrolled for user "
+            f"{enrollment.enterprise_customer_user.user_id} and course {enrollment.course_id}"
         )
         enterprise_enrollment.unenrolled_at = localized_utcnow()
         enterprise_enrollment.save()
+        get_producer().send(
+            signal=COURSE_UNENROLLMENT_COMPLETED,
+            topic='course-unenrollment-completed',
```
kiram15, line 366: Are we going to have more events in this topic? Should we keep the name more general?

enterprise/signals.py
```diff
@@ -349,16 +351,23 @@ def enterprise_unenrollment_receiver(sender, **kwargs):     # pylint: disable=un
...
     if enterprise_enrollment:
         logger.info(
-            f"Marking EnterpriseCourseEnrollment as unenrolled for user {enrollment.user} and "
-            f"course {enrollment.course.course_key}"
+            f"Marking EnterpriseCourseEnrollment as unenrolled for user "
+            f"{enrollment.enterprise_customer_user.user_id} and course {enrollment.course_id}"
         )
         enterprise_enrollment.unenrolled_at = localized_utcnow()
         enterprise_enrollment.save()
+        get_producer().send(
```
kiram15, line 364: Does this need to be gated behind a feature_flag?

enterprise/signals.py
```diff
@@ -349,16 +351,23 @@ def enterprise_unenrollment_receiver(sender, **kwargs):     # pylint: disable=un
...
+            signal=COURSE_UNENROLLMENT_COMPLETED,
```
johnnagro, line 365: this is a generic unenrollment signal but you only fire it for enterprise enrollments - i'd probably produce an event for all of these signals, not just the enterprise ones. i think this event is emitted somewhere upstream and that is where we should put that signal on the bus

enterprise/signals.py
johnnagro, line 366: did you look into using a topic like 'course-enrollment-lifecycle'?

johnnagro (changes requested):

enterprise/signals.py
justinhynes: I don't think it's a hard requirement but a best practice. It makes sense to have a feature flag in edx-platform or Credentials as there is likely going to be some time before the event bus is widely adopted by the Open edX Community. Relevant link to docs: https://openedx.atlassian.net/wiki/spaces/AC/pages/3508699151/How+to+start+using+the+Event+Bus#Producing

If there are any consumers of `edx-enterprise` outside of edx.org/2U, I would say it's a necessity. At the least, a temporary one is good in case we need to disable the functionality without a revert/code release.

enterprise/signals.py
justinhynes, line 365: Should we check if `COUSE_ENROLLMENT_COMPLETED` is not None before using it? It looks like it can _potentially_ be None depending on if there was an `ImportError`.

enterprise/signals.py
justinhynes: FWIW, my understanding is the topics that are named with "{blah]-lifecycle" are when a topic handles multiple distinct events in a single topic. 

The `{environment}-learning-certificate-lifecycle` events receive the "certificate awarded" and "certificate revoked" events. 

@kiram15 -- One thing I will note, another best practice is that the name of the topic should contain the architecture subdomain the event belongs to. In this case, it is the `learning` subdomain. Whatever this topic name ends up being, it should include this subdomain (e.g. `learning-{topic_name}`). 

Relevant link:

enterprise/signals.py
kiram15: The reason I didn't put the check on it was later on in the code, we have a check that doesn't connect the receiver unless it is not None. I can put another check in if this is not sufficient though, but this seems to do it already. 

`if COURSE_UNENROLLMENT_COMPLETED is not None:
    COURSE_UNENROLLMENT_COMPLETED.connect(enterprise_unenrollment_receiver)`

PR #1851, opened 2023-08-22
feat: Added the ability to get AI generated learner engagement and learner progress summary.

saleem-latif:
__Jira Ticket:__ [ENT-7362](https://2u-internal.atlassian.net/browse/ENT-7362)

__Description:__
This PR introduces a new endpoint to get AI generated summary of learner progress and learner engagement summary that will be displayed on enterprise admin portal.

**Merge checklist:**

**Post merge:**


enterprise/api/v1/views/analytics_summary.py
```diff
@@ -0,0 +1,58 @@
...
+
+        # Validate payload data
+        serializer = AnalyticsSummarySerializer(data=request.data)
+        if not serializer.is_valid():
+            return Response(
+                data={
+                    'errors': serializer.errors,
+                },
+                status=status.HTTP_400_BAD_REQUEST,
+            )
```
muhammad-ammar, line 49: Instead of `if`, What about just
```python
serializer.is_valid(raise_exception=True)
```

saleem-latif: @muhammad-ammar If I use `raise_exception=True` then returned response does not contain `errors` key at the top level and response is

```
{
    'learner_progress': ['<error-message>'],
    'learner_engagement': ['<error-message>'],
}
```
this structure is same as the actual response

```
{
    'learner_progress': '<summary>',
    'learner_engagement':  '<summary>',
}
```

This could cause confusion, that is why I am using the `if` condition to change the structure of the response.

irfanuddinahmad (approved):

jajjibhai008 (approved):

sameenfatima78 (approved):

enterprise/api/v1/views/analytics_summary.py
```diff
@@ -0,0 +1,58 @@
...
+
+from enterprise.api.utils import (
+    generate_prompt_for_learner_engagement_summary,
+    generate_prompt_for_learner_progress_summary,
+)
+from enterprise.api.v1.serializers import AnalyticsSummarySerializer
+from enterprise.models import ChatGPTResponse, EnterpriseCustomer
+
+
+class AnalyticsSummaryView(generics.GenericAPIView):
```
muhammad-ammar, line 21: If we `generics.CreateAPIView` then do we still need to set ` http_method_names = ['post']`?

muhammad-ammar (approved):
LGTM

enterprise/api/v1/views/analytics_summary.py
saleem-latif: Yes, we would still need this.

PR #1861, opened 2023-09-01
fix: Remove unnecessary default_app_config definitions.

feanil:
Django now detects this configuration automatically. You can remove
default_app_config.  This removes 11 warnings from edx-platform startup
and deals with something that's going to be fully ignored in django
4.1

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,15 @@ Change Log
 
 Unreleased
 ----------
+
+[4.1.7]
+fix: Remove unnecessary default_app_config definitions.
+
+Django now detects this configuration automatically. You can remove
+default_app_config.  This removes 11 warnings from edx-platform startup
+and deals with something that's going to be fully ignored in django
+4.1
+
 [4.1.6]
 -------
 fix: putting api_credentials bool in api response to access in admin portal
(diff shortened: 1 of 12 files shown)


CHANGELOG.rst
```diff
@@ -15,6 +15,15 @@ Change Log
 
 Unreleased
 ----------
+
+[4.1.7]
```
kdmccormick, line 19: Format the version number as a heading

```suggestion

[4.1.7]
-------
```

kdmccormick, line 19: You'll also need to bump the package version to 4.1.7

kdmccormick (changes requested):
Nice warning squashes. I haven't tested it, but I trust that you feel confident in the changes.

Two requests, but looks good otherwise.

Also, just a non-blocking nit: you might consider this a `chore:` rather than a `fix:` since there is no user-facing change.

CHANGELOG.rst
feanil: @kdmccormick I did bump the package number in enterprise/__init__.py, are there other places I'm not aware of?

iloveagent57:
Hey @feanil we actually just addressed this via https://github.com/openedx/edx-enterprise/pull/1829
Sorry for the duplicate work.

feanil:
Sweet, no worries.  I'll close this!

PR #1868, opened 2023-09-08
feat: enterprise sso orchestration record table api implementation

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-7576

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views/enterprise_customer_sso_configuration.py
+++ b/enterprise/api/v1/views/enterprise_customer_sso_configuration.py
@@ -6,18 +6,64 @@ from edx_rbac.decorators import permission_required
 from rest_framework import permissions, viewsets
 from rest_framework.decorators import action
 from rest_framework.response import Response
-from rest_framework.status import HTTP_200_OK, HTTP_404_NOT_FOUND
+from rest_framework.status import (
+    HTTP_200_OK,
+    HTTP_201_CREATED,
+    HTTP_400_BAD_REQUEST,
+    HTTP_403_FORBIDDEN,
+    HTTP_404_NOT_FOUND,
+)
 
 from django.contrib import auth
+from django.core.exceptions import FieldDoesNotExist, ValidationError
+from django.db import transaction
 
 from enterprise import models
+from enterprise.api.utils import get_enterprise_customer_from_user_id
 from enterprise.api.v1 import serializers
 from enterprise.logging import getEnterpriseLogger
+from enterprise.models import EnterpriseCustomer, EnterpriseCustomerSsoConfiguration, EnterpriseCustomerUser
 from enterprise.utils import localized_utcnow
 
 User = auth.get_user_model()
 
 LOGGER = getEnterpriseLogger(__name__)
+
+BAD_CUSTOMER_ERROR = 'Must provide valid enterprise customer'
+CONFIG_UPDATE_ERROR = 'Error updating SSO configuration record'
+CONFIG_CREATE_ERROR = 'Error creating SSO configuration record'
+
+
+class EnterpriseCustomerInactiveException(Exception):
+    """
+    Exception raised when an enterprise customer is inactive.
+    """
+
+
+def check_user_part_of_customer(user, enterprise_customer):
+    """
+    Checks if a user is in an enterprise customer.
+    """
+    ent_customer = enterprise_customer.enterprise_customer_users.filter(user_id=user.id)
+    if ent_customer and not ent_customer.first().active:
+        raise EnterpriseCustomerInactiveException
+    return ent_customer.exists() or user.is_staff
+
+
+def get_customer_from_request(request):
+    """
+    Gets the enterprise customer from the request or the user.
+    """
+    if customer_uuid := request.query_params.get('enterprise_customer'):
+        return customer_uuid
+    return get_enterprise_customer_from_user_id(request.user.id)
+
+
+def fetch_configuration_record(kwargs):
+    """
+    Fetches the configuration record for the given uuid.
+    """
+    return EnterpriseCustomerSsoConfiguration.all_objects.filter(pk=kwargs.get('configuration_uuid'))
 
 
 class EnterpriseCustomerSsoConfigurationViewSet(viewsets.ModelViewSet):
@@ -30,7 +76,8 @@ class EnterpriseCustomerSsoConfigurationViewSet(viewsets.ModelViewSet):
     serializer_class = serializers.EnterpriseCustomerSsoConfiguration
 
     # ``can_manage_enterprise_orchestration_configs`` maps to the edx operator system wide role. Meaning only operators
-    # can complete orchestration record configuration process.
+    # can complete orchestration record configuration process. The only intended caller for this endpoint is the SSO
+    # orchestration api.
     @permission_required(
         'enterprise.can_manage_enterprise_orchestration_configs',
     )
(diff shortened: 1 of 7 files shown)


enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -59,3 +78,115 @@ def oauth_orchestration_complete(self, request, configuration_uuid, *args, **kwa
         sso_configuration_record.active = True
         sso_configuration_record.save()
         return Response(status=HTTP_200_OK)
+
+    @permission_required(
+        'enterprise.can_access_admin_dashboard',
+        fn=lambda request, *args, **kwargs: get_customer_from_request(request)
```
marlonkeating, line 84: Do we need lambda here?
```suggestion
        fn=get_customer_from_request
```

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -59,3 +78,115 @@ def oauth_orchestration_complete(self, request, configuration_uuid, *args, **kwa
...
+    )
+    def create(self, request, *args, **kwargs):
+        # Force the enterprise customer to be the one associated with the user
+        request_data = request.data.dict().copy()
+        requesting_user_customer = request_data.get('enterprise_customer')
+        if requesting_user_customer:
+            try:
+                enterprise_customer = EnterpriseCustomer.objects.get(uuid=requesting_user_customer)
+            except EnterpriseCustomer.DoesNotExist:
+                return Response({'error': 'Must provide valid enterprise customer'}, status=HTTP_400_BAD_REQUEST)
```
marlonkeating, line 137: Same response text is used in multiple places here.

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -59,3 +78,115 @@ def oauth_orchestration_complete(self, request, configuration_uuid, *args, **kwa
...
+            if not check_user_part_of_customer(request.user, enterprise_customer):
+                return Response({'error': 'Must provide valid enterprise customer'}, status=HTTP_400_BAD_REQUEST)
+            request_data['enterprise_customer'] = enterprise_customer
+        else:
+            return Response({'error': 'Must provide valid enterprise customer'}, status=HTTP_400_BAD_REQUEST)
+        try:
+            new_record = EnterpriseCustomerSsoConfiguration.objects.create(**request_data)
+        except TypeError as e:
+            LOGGER.error(f'Error creating SSO configuration record: {e}')
```
marlonkeating, line 146: Same error text used twice

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -59,3 +78,115 @@ def oauth_orchestration_complete(self, request, configuration_uuid, *args, **kwa
...
+    @permission_required(
+        'enterprise.can_access_admin_dashboard',
+        fn=lambda request, *args, **kwargs: get_customer_from_request(request)
+    )
+    def retrieve(self, request, *args, **kwargs):
+        configuration_uuid = kwargs.get('configuration_uuid')
+        # Look for the configuration record
+        sso_configuration_record = self.queryset.filter(pk=configuration_uuid).first()
+        if not sso_configuration_record:
+            return Response(status=HTTP_404_NOT_FOUND)
```
marlonkeating, line 91: This code block of getting the sso config from the request is used in at least a few places, might be worthwhile to put in helper method.

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -59,3 +78,115 @@ def oauth_orchestration_complete(self, request, configuration_uuid, *args, **kwa
...
+    @permission_required(
+        'enterprise.can_access_admin_dashboard',
+        fn=lambda request, *args, **kwargs: get_customer_from_request(request)
+    )
+    def destroy(self, request, *args, **kwargs):
+        configuration_uuid = kwargs.get('configuration_uuid')
+        sso_configuration_record = self.queryset.filter(pk=configuration_uuid)
+        if not sso_configuration_record:
+            return Response(status=HTTP_404_NOT_FOUND)
+        sso_configuration_record.update(is_removed=True)
```
marlonkeating, line 191: Do we need to check if the user is authorized to delete?

tests/test_enterprise/api/test_views.py
```diff
@@ -7245,3 +7250,406 @@ def test_sso_configuration_oauth_orchestration_complete(self):
...
+        config_pk = uuid.uuid4()
+        enterprise_sso_orchestration_config = EnterpriseCustomerSsoConfigurationFactory(
+            uuid=config_pk,
+            enterprise_customer=self.enterprise_customer,
+        )
+        url = settings.TEST_SERVER + reverse(
+            self.SSO_CONFIGURATION_EXISTING_RECORD_ENDPOINTS,
+            kwargs={'configuration_uuid': config_pk}
+        )
+        response = self.client.get(url)
```
marlonkeating, line 7270: Code block is used in a number of tests

marlonkeating (changes requested):

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
alex-sheehan-edx: I believe it does? I don't believe `permission_required` is smart enough to grab the customer uuid from the request object on its own.

alex-sheehan-edx: just passing in that custom made get_customer_from_request isn't enough, so unless you can think of a way

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
alex-sheehan-edx: this was done as a minor security issue in exposing customer IDs, I wanted to keep the message the same if the user requests a customer that doesn't exist and a customer that the user doesn't belong to

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
alex-sheehan-edx: one is a logging message the other is the response returned to the user

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
marlonkeating: That part makes sense, but we can at least put the common error message string in a central location.

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
marlonkeating: In places above(like update) we use that same permission wrapper, but in addition check to make sure the person is part of the customer that owns the configuration being updated.  I would expect the same check if we're deleting a configuration.

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
marlonkeating: Hmm, can we at least remove the *args, **kwargs arguments in the lambda that aren't used?

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
marlonkeating: Sure but we can keep the error string in a single place.

tests/test_enterprise/api/test_views.py
alex-sheehan-edx: tried to refactor and reuse code!

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
alex-sheehan-edx: FWICT it's expected to take in the same params as the wrapped method, and the endpoints subclass django mixins that expect args and kwargs as params. So I can't really find a way to do it without having to ignore a linter rule somewhere

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -6,19 +6,50 @@
...
+    return enterprise_customer.enterprise_customer_users.filter(user_id=user.id).exists() or user.is_staff
+
+
+def get_customer_from_request(request):
+    """
+    Gets the enterprise customer from the request or the user.
+    """
+    if customer_uuid := request.query_params.get('enterprise_customer'):
+        return customer_uuid
+    return get_enterprise_customer_from_user_id(request.user.id)
```
johnnagro, line 44: what happens if a user belongs to more than 1 enterprise customer?

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -6,19 +6,50 @@
...
+BAD_CUSTOMER_ERROR = 'Must provide valid enterprise customer'
+CONFIG_UPDATE_ERROR = 'Error updating SSO configuration record'
+CONFIG_CREATE_ERROR = 'Error creating SSO configuration record'
+
+
+def check_user_part_of_customer(user, enterprise_customer):
+    """
+    Checks if a user is in an enterprise customer.
+    """
+    return enterprise_customer.enterprise_customer_users.filter(user_id=user.id).exists() or user.is_staff
```
johnnagro, line 35: should we be checking to see if the enterprise customer is active?

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
johnnagro, line 44: what happens if a user belongs to more than 1 enterprise customer?

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
johnnagro, line 35: should we be checking to see if the enterprise customer is active? throwing an "unauthorized" might be an appropriate response in the case that a customer account has been deactivated.

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -59,3 +91,114 @@ def oauth_orchestration_complete(self, request, configuration_uuid, *args, **kwa
...
+        # Force the enterprise customer to be the one associated with the user
+        request_data = request.data.dict().copy()
+        requesting_user_customer = request_data.get('enterprise_customer')
+        if requesting_user_customer:
+            try:
+                enterprise_customer = EnterpriseCustomer.objects.get(uuid=requesting_user_customer)
+            except EnterpriseCustomer.DoesNotExist:
+                return Response({'error': BAD_CUSTOMER_ERROR}, status=HTTP_400_BAD_REQUEST)
+            if not check_user_part_of_customer(request.user, enterprise_customer):
+                return Response({'error': BAD_CUSTOMER_ERROR}, status=HTTP_400_BAD_REQUEST)
```
johnnagro, line 150: if the user doesnt have access to a particular enterprise customer account, the status code should probably a 403 - the request is valid (so not a bad request) its just that we wont fulfill it

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -59,3 +91,114 @@ def oauth_orchestration_complete(self, request, configuration_uuid, *args, **kwa
...
+    )
+    def create(self, request, *args, **kwargs):
+        # Force the enterprise customer to be the one associated with the user
+        request_data = request.data.dict().copy()
+        requesting_user_customer = request_data.get('enterprise_customer')
+        if requesting_user_customer:
+            try:
+                enterprise_customer = EnterpriseCustomer.objects.get(uuid=requesting_user_customer)
+            except EnterpriseCustomer.DoesNotExist:
+                return Response({'error': BAD_CUSTOMER_ERROR}, status=HTTP_400_BAD_REQUEST)
```
johnnagro, line 148: i'd probably go for a 403 or 404 here rather than a plain 400 - the request is valid, we just cannot find the resource (in this case we may not want them to be able to disambiguate between not found and not allowed)

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -59,3 +91,114 @@ def oauth_orchestration_complete(self, request, configuration_uuid, *args, **kwa
...
+    )
+    def update(self, request, *args, **kwargs):
+        sso_configuration_record = fetch_configuration_record(kwargs)
+        if not sso_configuration_record:
+            return Response(status=HTTP_404_NOT_FOUND)
+        # Make sure the requesting user is part of the enterprise customer associated with the configuration record
+        # or are a staff member
+        config_customer = sso_configuration_record.first().enterprise_customer
+        if not check_user_part_of_customer(request.user, config_customer):
+            return Response(status=HTTP_404_NOT_FOUND)
```
johnnagro, line 173: should be a 403

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -59,3 +91,114 @@ def oauth_orchestration_complete(self, request, configuration_uuid, *args, **kwa
...
+
+        # If the request includes a customer uuid, ensure the new customer is valid
+        if new_customer := request.data.dict().get('enterprise_customer'):
+            try:
+                enterprise_customer = EnterpriseCustomer.objects.get(uuid=new_customer)
+            except EnterpriseCustomer.DoesNotExist:
+                return Response({'error': BAD_CUSTOMER_ERROR}, status=HTTP_400_BAD_REQUEST)
```
johnnagro, line 180: 403

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -59,3 +91,114 @@ def oauth_orchestration_complete(self, request, configuration_uuid, *args, **kwa
...
+            if not check_user_part_of_customer(request.user, enterprise_customer):
+                return Response({'error': BAD_CUSTOMER_ERROR}, status=HTTP_400_BAD_REQUEST)
```
johnnagro, line 182: 403

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -59,3 +91,114 @@ def oauth_orchestration_complete(self, request, configuration_uuid, *args, **kwa
...
+    @permission_required(
+        'enterprise.can_access_admin_dashboard',
+        fn=lambda request, *args, **kwargs: get_customer_from_request(request)
+    )
+    def destroy(self, request, *args, **kwargs):
+        sso_configuration_record = fetch_configuration_record(kwargs)
+        if not sso_configuration_record:
+            return Response(status=HTTP_404_NOT_FOUND)
+        if not check_user_part_of_customer(request.user, sso_configuration_record.first().enterprise_customer):
+            return Response(status=HTTP_404_NOT_FOUND)
```
johnnagro, line 202: 403

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -59,3 +91,114 @@ def oauth_orchestration_complete(self, request, configuration_uuid, *args, **kwa
...
+        return Response(serializer.data, status=HTTP_200_OK)
+
+    @permission_required(
+        'enterprise.can_access_admin_dashboard',
+        fn=lambda request, *args, **kwargs: get_customer_from_request(request)
+    )
+    def destroy(self, request, *args, **kwargs):
+        sso_configuration_record = fetch_configuration_record(kwargs)
+        if not sso_configuration_record:
+            return Response(status=HTTP_404_NOT_FOUND)
```
johnnagro, line 200: 403

johnnagro (approved):
i have some recommendations on status codes which might make the API more easy to understand

marlonkeating (approved):

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
alex-sheehan-edx: that's why I included the customer param, the requester can always simple include the enterprise customer that they want to be associated with the request. If none is supplied then all we can do is go off of the first linked one.

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
alex-sheehan-edx: I was debating on this, part of me thought it would be better to not reveal customer data

enterprise/api/v1/views/enterprise_customer_sso_configuration.py
alex-sheehan-edx: uhhh 404 is probably best here?

PR #1871, opened 2023-09-13
feat: enterprise sso orchestrator api client implementation

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-7571

**Merge checklist:**

**Post merge:**


enterprise/api_client/sso_orchestrator.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def __init__(self):
+        self.base_url = get_sso_orchestrator_api_base_url()
+        self.session = None
+
+    def _get_orchestrator_callback_url(self, config_pk):
+        """
+        get the callback url for the SSO Orchestrator API
+        """
+        lms_base_url = get_configuration_value('LMS_ROOT_URL', settings.LMS_ROOT_URL)
```
johnnagro, line 41: is this the correct way to call `get_configuration_value`? it doesn't seem correct to me to supply the setting as the default value for when the setting does not exist? perhaps i misunderstand

enterprise/api_client/sso_orchestrator.py
```diff
@@ -0,0 +1,140 @@
...
+from enterprise.errors import ClientError
+from enterprise.utils import (
+    get_configuration_value,
+    get_sso_orchestrator_api_base_url,
+    get_sso_orchestrator_configure_path,
+    get_sso_orchestrator_basic_auth_username,
+    get_sso_orchestrator_basic_auth_password,
+)
+
+class EnterpriseSSOOrchestratorApiClient:
```
johnnagro, line 20: we may still want to extend from something inside edx-rest-api-client (perhaps a version which supports BasicAuth rather than OAuth) so that we benefit from all the nice default addons like custom user-agents which are app aware, request id forwarding, etc.

enterprise/api_client/sso_orchestrator.py
```diff
@@ -0,0 +1,140 @@
...
+                "Failed to create SSO Orchestrator auth headers: username required.",
+                HTTPStatus.INTERNAL_SERVER_ERROR.value
+            )
+        orchestrator_password = get_sso_orchestrator_basic_auth_password()
+        if not orchestrator_password:
+            raise ClientError(
+                "Failed to create SSO Orchestrator auth headers: password required.",
+                HTTPStatus.INTERNAL_SERVER_ERROR.value
+            )
+        return f"Basic {base64.b64encode(f'{orchestrator_username}:{orchestrator_password}'.encode('utf-8')).decode()}"
```
johnnagro, line 76: doesn't `requests` handle this for us? https://docs.python-requests.org/en/latest/user/authentication/#basic-authentication

enterprise/api_client/sso_orchestrator.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def _create_session(self):
+        """
+        create a requests session object
+        """
+        if not self.session:
+            self.session = requests.Session()
+            self.session.headers.update({
```
johnnagro, line 84: https://docs.python-requests.org/en/latest/user/authentication/#basic-authentication

enterprise/api_client/sso_orchestrator.py
```diff
@@ -0,0 +1,140 @@
...
+                "Authorization": self._create_auth_header(),
+            })
+
+    # No real need for this method at the moment as the orchestrator only supports post
```
johnnagro, line 88: old comments should be removed

enterprise/api_client/sso_orchestrator.py
```diff
@@ -0,0 +1,140 @@
...
+    #         )
+    #     return response.json()
+
+    def _post(self, url, data=None):
+        """
+        make a GET request to the SSO Orchestrator API
+        """
+        self._create_session()
+        response = self.session.post(url, data=data)
+        if response.status_code != HTTPStatus.OK.value:
```
johnnagro, line 109: nit: you may want to have a wider range of 'success' such as `status_code < 400`

enterprise/api_client/sso_orchestrator.py
```diff
@@ -0,0 +1,140 @@
...
+        config_data,
+        config_pk,
+        enterprise_data,
+        is_sap=False, 
+        updating_existing_record=False,
+    ):
+        """
+        configure an SSO orchestration record
+        """
+        # TODO: validate config_data
```
johnnagro, line 127: what do you mean by validate in this context?

enterprise/errors.py
```diff
@@ -31,3 +31,13 @@ class UnlinkUserFromEnterpriseError(Exception):
     """
     An error occurred while unlinking a user from an enterprise.
     """
+
+class ClientError(Exception):
```
johnnagro, line 35: i'd consider renaming this and/or namespacing this inside the SSO client/work work so its clearly the "Enterprise SSO" ClientError and not any of the other ones.

you may also want this to extend a `requests` Exception? duno

johnnagro (changes requested):

enterprise/api_client/sso_orchestrator.py
alex-sheehan-edx: it's how we've done it in other areas

enterprise/api_client/sso_orchestrator.py
alex-sheehan-edx: I didn't see anything about basic auth in the rest api client but I can double check

enterprise/api_client/sso_orchestrator.py
alex-sheehan-edx: it's very likely haha

enterprise/api_client/sso_orchestrator.py
alex-sheehan-edx: uhm that was an old comment, I think validation should come from the model method, not the api client.

enterprise/api_client/sso_orchestrator.py
alex-sheehan-edx: I think it's the other way 'round and we're filtering for failed requests so shouldn't it be `status_code >= 300`?

enterprise/api_client/sso_orchestrator.py
alex-sheehan-edx: https://github.com/openedx/edx-enterprise/blob/426413c29c07a14a2b5a971255c63d1e58f4796d/enterprise/models.py#L625-L661

johnnagro (approved):

enterprise/api_client/sso_orchestrator.py
```diff
@@ -0,0 +1,132 @@
...
+        if not orchestrator_username:
+            raise SsoOrchestratorClientError(
+                "Failed to create SSO Orchestrator auth headers: username required.",
+            )
+        orchestrator_password = get_sso_orchestrator_basic_auth_password()
+        if not orchestrator_password:
+            raise SsoOrchestratorClientError(
+                "Failed to create SSO Orchestrator auth headers: password required.",
+            )
+        return HTTPBasicAuth(orchestrator_username, SsoOrchestratorClientError)
```
praveenram2u, line 84: typo here? should this say username, password?

enterprise/api_client/sso_orchestrator.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        callback_url = self._get_orchestrator_callback_url(str(config_pk))
+        if updating_existing_record:
+            callback_url = f"{callback_url}?updating_existing_record=true"
+        config_request_data_key = 'samlConfiguration'
+        if is_sap:
+            config_request_data_key = 'sapsfConfiguration'
+        config_data['uuid'] = str(config_data['uuid'])
+        request_data = {
+            config_request_data_key: config_data,
```
praveenram2u, line 127: We may need both SAML config and SAP SF config for SAP? I thought we still had a metadata URL / XML and such.

enterprise/api_client/sso_orchestrator.py
```diff
@@ -0,0 +1,134 @@
...
+        # probably want config value validated for this
+        return urljoin(self.base_url, get_sso_orchestrator_configure_path())
+
+    def _create_auth_header(self):
+        """
+        create the basic auth header for requests to the SSO Orchestrator API
+        """
+        if orchestrator_username := get_sso_orchestrator_basic_auth_username():
+            if orchestrator_password := get_sso_orchestrator_basic_auth_password():
+                return HTTPBasicAuth(orchestrator_password, orchestrator_username)
```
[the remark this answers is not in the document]
alex-sheehan-edx: 🤦

PR #1877, opened 2023-09-19
feat: add features dict to EnterpriseCustomerViewSet

adamstankiewicz:
Ticket: https://2u-internal.atlassian.net/browse/ENT-7713

Adds support for exposing Waffle-based feature flags onto the default paginated response in the `EnterpriseCustomerViewSet` such that micro-frontends already calling API endpoint(s) from this viewset have access to the Waffle-based feature flags without introducing new API integrations.

See ADR ([preview](https://github.com/openedx/edx-enterprise/blob/d00f0d8fd9095f725ef9793b7ac20b669f70181f/docs/decisions/0012-enterprise-feature-flags-waffle.rst)) for more details around this approach.

Example API response with `enterprise_features` object included:

```js
{
    "next": null,
    "previous": null,
    "count": 1,
    "num_pages": 1,
    "current_page": 1,
    "start": 0,
    "results": [
        {
            "uuid": "a067aacd-667d-4bd6-81e2-e332ef5ca1cd",
            "name": "Test Enterprise",
            "slug": "test-enterprise",
            "active": true,
            // ...
        }
    ],
    "enterprise_features": {
        "top_down_assignment_real_time_lcm": true
    }
}
```

**Merge checklist:**

**Post merge:**


tests/test_enterprise/api/test_filters.py
```diff
@@ -223,7 +223,19 @@ def test_filter(self, is_staff, is_linked_to_enterprise, has_access):
...
                 assert enterprise_customer_response[key] == value
         else:
-            assert response == {'count': 0, 'next': None, 'previous': None, 'results': []}
+            mock_empty_200_success_response = {
+                'next': None,
+                'previous': None,
+                'count': 0,
+                'num_pages': 1,
+                'current_page': 1,
+                'start': 0,
```
pwnage101, line 232: I don't understand why num_pages, current_page, and start were all net additions.  Didn't you just extend DefaultPagination?  Not a blocker.

enterprise/api/pagination.py
```diff
@@ -5,8 +5,41 @@
...
+            self: PaginationWithFeatureFlags instance.
+            data (dict): Results for current page.
+
+        Returns:
+            (Response): DRF response object containing ``features`` dict.
+        """
+        paginated_response = super().get_paginated_response(data)
+        paginated_response.data.update({
+            'features': {
+                'top_down_assignment_real_time_lcm': top_down_assignment_real_time_lcm(),
```
pwnage101, line 38: Recommendation: add a function to toggles.py to serialize this dictionary so that adding a new feature flag is limited to just updating one file.

pwnage101 (approved):

enterprise/api/pagination.py
adamstankiewicz: Great suggestion! I will update this PR after afternoon meetings.

iloveagent57 (approved):

tests/test_enterprise/api/test_filters.py
iloveagent57: The default pagination was probably the vanilla DRF page-number paginator set via settings file.  Adam's new paginator extends the edX-DRF default paginator: https://github.com/openedx/edx-drf-extensions/blob/master/edx_rest_framework_extensions/paginators.py

PR #1883, opened 2023-09-25
feat: added dry-run mode for content_metadata transmission

zamanafzal:
Ticket: https://2u-internal.atlassian.net/browse/ENT-5959

Adds support for dry run mode to work with transmission of content metadata.

Dry Run mode could be enabled using this [flag](https://github.com/openedx/edx-enterprise/blob/733e7c90c5992ec1437537115b10c60dfc167329/integrated_channels/integrated_channel/models.py#L137). 

**Merge checklist:**

**Post merge:**


integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -153,6 +153,20 @@ def _transmit_action(self, content_metadata_item_map, client_method, action_name
         for chunk in islice(chunk_items, transmission_limit):
             json_payloads = [item.channel_metadata for item in list(chunk.values())]
             serialized_chunk = self._serialize_items(json_payloads)
+            if self.enterprise_configuration.dry_run_mode_enabled:
+                encoded_serialized_payload4 = encode_binary_data_for_logging(serialized_chunk)
```
johnnagro, line 157: i wrote the encoded data logging facility and i am undecided if it is actually useful.

you may need to/want to log each item, not a whole chunk (which is often ~500 courses). this is because splunk has a line-length limit and will truncate long log lines. if you chose to use the encoded data logging you likely need to log each item in the chunk.

integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -153,6 +153,20 @@ def _transmit_action(self, content_metadata_item_map, client_method, action_name
...
+                enterprise_customer_uuid = self.enterprise_configuration.enterprise_customer.uuid
+                LOGGER.info(generate_formatted_log(
+                    self.enterprise_configuration.channel_code(),
+                    enterprise_customer_uuid,
+                    None,
+                    None,
```
johnnagro, line 163: another reason why you may want to log for each item here is so that you can set the `course_or_course_run_key` variable in the `generate_log_line` method - otherwise it might be difficult to trace the lines for specific content items.

johnnagro (approved):
depending on how you intent to consume/parse these dry-run logs you may want to make the logging more granular

zamanafzal:
Thank you @johnnagro for your valuable feedback.

integrated_channels/integrated_channel/transmitters/content_metadata.py
zamanafzal: Hi @johnnagro, should I encode the data while logging? Is there any specific use case we have due to which we are encoding this data while logging?

alex-sheehan-edx: Not sure about encoding, but I second @johnnagro's point on iterating over and logging individual items in the chunk. We've run into situations where critical pieces of logs/identifiers were lost because the log was truncated.

alex-sheehan-edx (commented):
A side effect of this implementation is that content transmission audits will retain the `marked_for` field value set by the exporter (as it's the job of the transmitter to remove the value). Theoretically this is fine as the value simply denotes the exporter identifying the content needs updating in some way, but we should double check that we want the dry run to  effect records in the database (perhaps it's even a benefit?).

zamanafzal:
@johnnagro @alex-sheehan-edx I've added course_run_key variable in the generate_log_line method.

sameenfatima78 (approved):

PR #1886, opened 2023-09-26
fix: fixed how we determine active value for SAP

sameenfatima78:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/utils.py
+++ b/enterprise/utils.py
@@ -1196,6 +1196,19 @@ def get_course_run_duration_info(course_run):
     return duration_info
 
 
+def get_advertised_or_closest_course_run(content_metadata_item):
+    """
+    Returns advertised course run of a course. If the advertised run does not exist, it looks for the closest run.
+    """
+    course_run = get_advertised_course_run(content_metadata_item)
+    # get the closest run if advertised run doesn't exist
+    if not course_run:
+        course_runs = content_metadata_item.get('course_runs')
+        if course_runs:
+            course_run = get_closest_course_run(course_runs)
+    return course_run
+
+
 def get_duration_of_course_or_courserun(content_metadata_item):
     """
     Returns duration start, end dates given a piece of content_metadata item
@@ -1212,12 +1225,10 @@ def get_duration_of_course_or_courserun(content_metadata_item):
         start = content_metadata_item.get('start')
         end = content_metadata_item.get('end')
     elif content_metadata_item.get('content_type') == 'course':
-        course_runs = content_metadata_item.get('course_runs')
-        if course_runs:
-            course_run = get_closest_course_run(course_runs)
-            if course_run:
-                start = course_run.get('start')
-                end = course_run.get('end')
+        course_run = get_advertised_or_closest_course_run(content_metadata_item)
+        if course_run:
+            start = course_run.get('start')
+            end = course_run.get('end')
     if not start:
         return 0, None, None
     start_date = parse_datetime_handle_invalid(start)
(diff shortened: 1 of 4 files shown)


enterprise/utils.py
```diff
@@ -1196,6 +1196,25 @@ def get_course_run_duration_info(course_run):
...
+    """
+    Returns the advertised_course_run from a course if it exists. If not, returns the closest run.
+    If we receive a course run, returns None.
+
+    Argument:
+        course: A course with course runs.
+    Returns:
+        advertised or closest course run in a course.
+    """
+    course_run = get_advertised_course_run(content_metadata_item)
```
alex-sheehan-edx, line 1209: According to https://github.com/openedx/enterprise-catalog/blob/master/docs/decisions/0007-normalized-metadata-course-types.rst I would maybe use the new `normalized_metadata` field instead of advertised course run

integrated_channels/sap_success_factors/exporters/content_metadata.py
```diff
@@ -168,6 +169,10 @@ def transform_schedule(self, content_metadata_item):
         """
         Return the schedule of the content item.
         """
+        if content_metadata_item.get('content_type') == 'course':
+            course_run = get_advertised_or_closest_course_run(content_metadata_item)
```
alex-sheehan-edx, line 173: here as well for `normalized_metadata`

alex-sheehan-edx (commented):
One quick comment, overall looks good to me

tests/test_integrated_channels/test_sap_success_factors/test_exporters/test_content_metadata.py
```diff
@@ -303,11 +303,13 @@ def test_transform_courserun_title_excludes_start(self):
                         "key": "course-v1:CatalystX+IL-BSL.S1x+2T2017",
                         "start": "2017-05-16T16:00:00Z",
                         "end": "2017-06-30T23:59:00Z",
+                        "content_type": "courserun"
```
bcitro, line 306: From looking at some example data in production, I don't think this would be in the actual metadata for a course ContentMetadata item.

tests/test_integrated_channels/test_sap_success_factors/test_exporters/test_content_metadata.py
```diff
@@ -303,11 +303,13 @@ def test_transform_courserun_title_excludes_start(self):
...
                     },
                     {
                         "key": "course-v1:CatalystX+IL-BSL.S1x+2T2018",
                         "start": "2018-05-16T16:00:00Z",
                         "end": "2018-06-30T23:59:00Z",
+                        "content_type": "courserun"
```
bcitro, line 312: same here

integrated_channels/sap_success_factors/exporters/content_metadata.py
```diff
@@ -168,6 +169,10 @@ def transform_schedule(self, content_metadata_item):
...
+            if course_run:
+                content_metadata_item = course_run
         duration, start, end = get_duration_of_course_or_courserun(content_metadata_item)
```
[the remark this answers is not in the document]
sameenfatima78: Yep, just had a look at a sample content metadata item in prod and it makes sense.

What if we pass the `content_metadata_item` as is to this method and then use this new utility that I just created `get_advertised_or_closest_course_run` instead of `get_closest_course_run` [here](https://github.com/openedx/edx-enterprise/blob/master/enterprise/utils.py#L1217)?

sameenfatima78:
@bcitro This is ready for review again

integrated_channels/sap_success_factors/exporters/content_metadata.py
sameenfatima78: Makese sense but in this case, we also need to check whether the course run is available for enrollment. For that, we need extra information that is not available in `normalized_metadata` but is available in the course run body.

PR #1887, opened 2023-09-27
[Django 4.2]: Migrations from Django 4.2

irtazaakram:
Related PR https://github.com/openedx/edx-enterprise/pull/1882

```
--
-- Alter field enterprise_course_enrollment on learnercreditenterprisecourseenrollment
--
-- (no-op)
--
-- Alter field enterprise_course_entitlement on learnercreditenterprisecourseenrollment
--
-- (no-op)
--
-- Alter field enterprise_course_enrollment on licensedenterprisecourseenrollment
--
-- (no-op)
--
-- Alter field enterprise_course_entitlement on licensedenterprisecourseenrollment
--
-- (no-op)
```


awais786 (approved):

saleem-latif (approved):

enterprise/__init__.py
```diff
@@ -2,4 +2,4 @@
 Your project description goes here.
 """
 
-__version__ = "4.3.3"
+__version__ = "4.3.2"
```
saleem-latif, line 5: @irtazaakram I think you accidently reduced the version number instead of increasing it. Also, add and entry in [CHANGELOG](https://github.com/openedx/edx-enterprise/blob/master/CHANGELOG.rst) file.

saleem-latif (changes requested):

enterprise/__init__.py
saleem-latif: This one has already been used and is currently deployed on PyPI https://github.com/openedx/edx-enterprise/releases/tag/4.3.2

irtazaakram: 🤦 Fixed it and updated changelog

saleem-latif (approved):

PR #1889, opened 2023-10-02
Encryption fields added to moodle config

MueezKhan246:
Description:
Added encryption fields to the model of moodle config.

JIRA:
https://2u-internal.atlassian.net/jira/software/c/projects/ENT/issues/ENT-5613


integrated_channels/moodle/migrations/0028_auto_20230928_1530.py
```diff
@@ -0,0 +1,58 @@
...
+        migrations.AddField(
+            model_name='moodleenterprisecustomerconfiguration',
+            name='decrypted_token',
+            field=fernet_fields.fields.EncryptedCharField(blank=True, help_text="The encrypted API user's token used to obtain new tokens.", max_length=255, null=True, verbose_name='Encrypted Webservice Token'),
+        ),
+        migrations.AddField(
+            model_name='moodleenterprisecustomerconfiguration',
+            name='decrypted_username',
+            field=fernet_fields.fields.EncryptedCharField(blank=True, help_text="The encrypted API user's username used to obtain new tokens.", max_length=255, null=True, verbose_name='Encrypted Webservice Username'),
+        ),
```
sameenfatima78, line 56: Unsure why these field names are repeating here... can you please create a fresh migration with your latest changes?

sameenfatima78 (commented):
Left some questions. otherwise looks good!

integrated_channels/moodle/migrations/0028_auto_20230928_1530.py
```diff
@@ -0,0 +1,58 @@
...
+def populate_decrypted_fields(apps, schema_editor):
+    """
+    Populates the encryption fields with the data previously stored in database.
+    """
+    MoodleEnterpriseCustomerConfiguration = apps.get_model('moodle', 'MoodleEnterpriseCustomerConfiguration')
+
+    for moodle_enterprise_configuration in MoodleEnterpriseCustomerConfiguration.objects.all():
+        moodle_enterprise_configuration.decrypted_username = moodle_enterprise_configuration.username
+        moodle_enterprise_configuration.decrypted_password = moodle_enterprise_configuration.password
+        moodle_enterprise_configuration.decrypted_token = moodle_enterprise_configuration.token
```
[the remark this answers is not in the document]
MueezKhan246: the scope of this PR is only to copy original values in the new column for now?

integrated_channels/moodle/migrations/0028_auto_20230928_1530.py
MueezKhan246: "field names are repeating" Sorry, I am a bit unsure about what was meant by this.

sameenfatima78: The fields are being added twice if you take a look at the migration file

MueezKhan246: the historical data table and current data table?

MueezKhan246:
@johnnagro kindly please review this, whenever you get the time to do so.

CHANGELOG.rst
```diff
@@ -15,6 +15,11 @@ Change Log
 
 Unreleased
 ----------
+[4.6.9]
+-------
+refactor: removing "null=true" constraint from encrypted fields
```
sameenfatima78, line 20: [question] why did we feel the need to disable null values for this field?

integrated_channels/moodle/migrations/0028_auto_20230928_1530.py
```diff
@@ -0,0 +1,59 @@
...
+        moodle_enterprise_configuration.save()
```
sameenfatima78, line 18: [suggestion] I think this part should be covered in a separate migration. As per the docs:

> Add the new encrypted field with a different name and initialize its values as null, otherwise decryption will be attempted before anything has been encrypted.

We should initially add new field and save nulls, then the second migration can take care of copying values so decryption only takes place after encryption.

CHANGELOG.rst
sameenfatima78: As per docs, we should be saving null values first https://django-fernet-fields.readthedocs.io/en/latest/#migrations

integrated_channels/moodle/models.py
```diff
@@ -64,6 +66,38 @@ class MoodleEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguratio
...
+            )
+        return self.decrypted_username
+
+    @encrypted_username.setter
+    def encrypted_username(self, value):
+        """
+        Set the encrypted username.
+        """
+        self.decrypted_username = value
+
```
sameenfatima78, line 100: [curious] Are these properties and setters even needed? We don't need to push encrypted data anywhere in the app. Adding this property would have made sense if we were required to refer to this data in a serializer to display it somewhere in the application. Curious to know your thoughts.

integrated_channels/moodle/models.py
```diff
@@ -64,6 +66,38 @@ class MoodleEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguratio
...
     )
 
+    decrypted_username = EncryptedCharField(
+        max_length=255,
+        verbose_name="Encrypted Webservice Username",
+        blank=True,
+        help_text=_(
+            "The encrypted API user's username used to obtain new tokens."),
+        default="",
+    )
```
sameenfatima78, line 76: [nit] Once we address the above feedback, please make sure to add back null=True and remove default value.

integrated_channels/moodle/models.py
```diff
@@ -73,6 +107,38 @@ class MoodleEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguratio
...
     )
 
+    decrypted_password = EncryptedCharField(
+        max_length=255,
+        verbose_name="Encrypted Webservice Password",
+        blank=True,
+        help_text=_(
+            "The encrypted API user's password used to obtain new tokens."),
+        default="",
+    )
```
sameenfatima78, line 117: [nit] Once we address the above feedback, please make sure to add back null=True and remove default value.

integrated_channels/moodle/models.py
```diff
@@ -73,6 +107,38 @@ class MoodleEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguratio
...
+            )
+        return self.decrypted_password
+
+    @encrypted_password.setter
+    def encrypted_password(self, value):
+        """
+        Set the encrypted password.
+        """
+        self.decrypted_password = value
+
```
sameenfatima78, line 141: Same feedback as above.

integrated_channels/moodle/models.py
```diff
@@ -82,6 +148,38 @@ class MoodleEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguratio
...
     )
 
+    decrypted_token = EncryptedCharField(
+        max_length=255,
+        verbose_name="Encrypted Webservice Token",
+        blank=True,
+        help_text=_(
+            "The encrypted API user's token used to obtain new tokens."),
+        default="",
+    )
```
sameenfatima78, line 158: [nit] Once we address the above feedback, please make sure to add back null=True and remove default value.

integrated_channels/moodle/models.py
```diff
@@ -82,6 +148,38 @@ class MoodleEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguratio
...
+            )
+        return self.decrypted_token
+
+    @encrypted_token.setter
+    def encrypted_token(self, value):
+        """
+        Set the encrypted token.
+        """
+        self.decrypted_token = value
+
```
sameenfatima78, line 182: Same feedback as above.

integrated_channels/utils.py
```diff
@@ -501,3 +501,13 @@ def get_enterprise_client_by_channel_code(channel_code):
...
     return _enterprise_client_model_by_channel_code[channel_code]
+
+
+def dummy_reverse(_apps, _schema_editor):
+    """
+    Reverse a data migration but do nothing.
+
+    :param _apps:
+    :param _schema_editor:
+    :return:
```
sameenfatima78, line 512: [suggestion] A better way to achieve this would be to use `RunPython.noop` as a reverse function. https://stackoverflow.com/questions/66479161/what-does-runpython-noop-do

sameenfatima78 (changes requested):
Left some comments.

integrated_channels/moodle/models.py
zamanafzal: Why did we use this filed name decrypted_username except for encrypted_username?

integrated_channels/moodle/models.py
zamanafzal: Why did we use this filed name decrypted_password except for encrypted_password?

sameenfatima78: I think it's for extra caution so whoever accesses these values in app knows this data will be available in decrypted form. It's only saved in encrypted form in DB at rest.

integrated_channels/utils.py
MueezKhan246: @sameenfatima78 suggested method can be better.

integrated_channels/moodle/models.py
MueezKhan246: @sameenfatima78 @zamanafzal "decrypted" is used here to remain consistent with other usage of fernet fields in code, this decrypted fields return unencrypted values when referenced in code.

integrated_channels/moodle/models.py
MueezKhan246: @sameenfatima78 as originally the plan was to also make changes in the code to handle encrypted fields these changes were made and kept in the same pr so changes to model are pushed here.

integrated_channels/moodle/migrations/0028_auto_20230928_1530.py
MueezKhan246: @sameenfatima78 it can be a better strategy

CHANGELOG.rst
MueezKhan246: @sameenfatima78 because the unencrypted fields are using default null=False

integrated_channels/moodle/models.py
```diff
@@ -82,6 +101,15 @@ class MoodleEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguratio
...
     )
 
+    decrypted_token = EncryptedCharField(
+        max_length=255,
+        verbose_name="Encrypted Webservice Token",
+        blank=True,
+        help_text=_(
+            "The encrypted API user's token used to obtain new tokens."),
+        null=True,
+    )
```
sameenfatima78, line 111: Just to clear out the confusion with encrypted/decrypted data, can we add this line in help_text on all newly added fields? "It will be encrypted when stored in the database."

sameenfatima78 (commented):
Looks good! Just one small suggestion.

sameenfatima78 (commented):
Left some suggestions but looks good overall!

sameenfatima78 (approved):
LGTM

hamzawaleed01:
@MueezKhan246 , @sameenfatima78 

I've got some question just for my understanding
1. Do we need to add stuff like [this](https://github.com/openedx/edx-enterprise/blob/2a3ea7c5791959cfcb51b3c399e480481f486998/enterprise/models.py#L2979) and [this](https://github.com/openedx/edx-enterprise/blob/2a3ea7c5791959cfcb51b3c399e480481f486998/enterprise/models.py#L2994) in upcoming PRs 🤔  
2. Once the data is migrated, @MueezKhan246 will create another PR to delete the unencrypted columns?
3. Once the 2nd point is done ☝️, Mueez will have to make sure that in those API endpoints that are fetching Moodle config, we're sending unencrypted data right? 🤔

PR #1896, opened 2023-10-03
chore: sso orchestrator configs should start inactive and be activated upon successful configuration

alex-sheehan-edx:
The ideal behavior is that:
- we always tell the orchestrator to create the integrations as "active"
- we show the config as inactive until the point the enterprise is able to use their IDP
  - this point should happen when the oauth complete endpoint is hit by the orchestrator
- we don't need to update the validated at every time the integration is used
- 
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views/enterprise_customer_sso_configuration.py
+++ b/enterprise/api/v1/views/enterprise_customer_sso_configuration.py
@@ -146,6 +146,10 @@ class EnterpriseCustomerSsoConfigurationViewSet(viewsets.ModelViewSet):
                 ' not been marked as submitted.'
             )
 
+        # Mark the configuration record as active IFF this the record has never been configured.
+        if not sso_configuration_record.configured_at:
+            sso_configuration_record.active = True
+
         sso_configuration_record.configured_at = localized_utcnow()
         # Completing the orchestration process for the first time means the configuration record is now configured and
         # can be considered active. However, subsequent configurations to update the record should not be reactivated,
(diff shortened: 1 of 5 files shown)


enterprise/api/v1/views/enterprise_customer_sso_configuration.py
```diff
@@ -146,6 +146,10 @@ def oauth_orchestration_complete(self, request, configuration_uuid, *args, **kwa
                 ' not been marked as submitted.'
             )
 
+        # Mark the configuration record as active IFF this the record has never been configured.
+        if sso_configuration_record.configured_at:
```
marlonkeating, line 150: Logic to match comment
```suggestion
        # Mark the configuration record as active IFF this the record has never been configured.
        if not sso_configuration_record.configured_at:
```

enterprise/models.py
```diff
@@ -4055,7 +4055,14 @@ def submit_for_configuration(self, updating_existing_record=False):
             is_sap = True
         else:
             for field in self.base_saml_config_fields:
-                config_data[utils.camelCase(field)] = getattr(self, field)
+                
+                if field == "active":
+                    if not updating_existing_record:
+                        config_data['enable'] = True
```
marlonkeating, line 4061: Are we moving forward under the assumption we won't be saving draft configurations?

marlonkeating (changes requested):

enterprise/models.py
alex-sheehan-edx: MVP will not implement drafts, however this updated logic doesn't necessarily block us from doing so. This is in our method to send a config to the orchestrator for configuration. If the self service tool is sending a POST for a draft to our API, we’d have to make sure the api view doesn’t submit for configuration, but that's part of the drafts implementation.

marlonkeating (approved):

PR #1923, opened 2023-10-23
[ENT-7716] feat: Truncate API Response before writing to DB

justEhmadSaeed:
JIRA Ticket: [ENT-7716](https://2u-internal.atlassian.net/browse/ENT-7716)

Description:
To prevent errors of maxing out packet sizes for the field of APIResponseRecord.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/integrated_channel/transmitters/content_metadata.py
+++ b/integrated_channels/integrated_channel/transmitters/content_metadata.py
@@ -12,7 +12,7 @@ import requests
 from django.apps import apps
 from django.conf import settings
 
-from enterprise.utils import localized_utcnow
+from enterprise.utils import localized_utcnow, truncate_string
 from integrated_channels.exceptions import ClientError
 from integrated_channels.integrated_channel.client import IntegratedChannelApiClient
 from integrated_channels.integrated_channel.transmitters import Transmitter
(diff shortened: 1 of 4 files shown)


integrated_channels/integrated_channel/transmitters/content_metadata.py
```diff
@@ -228,6 +228,13 @@ def _transmit_action(self, content_metadata_item_map, client_method, action_name
                     transmission.api_response_status_code = response_status_code
                     was_successful = response_status_code < 300
                     api_content_response = self._filter_api_response(response_body, content_id)
+                    (api_content_response, was_truncated) = truncate_string(api_content_response)
+                    if was_truncated:
+                        self._log_info(
+                            f'integrated_channel_content_transmission_id={transmission.id}, '
+                            f'api response truncated',
+                            course_or_course_run_key=content_id
+                        )
```
sameenfatima78, line 237: I think we can remove this log, won't help us capture anything that we aren't already capturing through other logs.

sameenfatima78 (approved):
One small comment otherwise looks good!

integrated_channels/integrated_channel/transmitters/content_metadata.py
justEhmadSaeed: it was suggested by @zamanafzal to observe the frequency of response exceeding the defined max limit so that we can update the max limit in the future accordingly.

sameenfatima78: Makes sense.

PR #1940, opened 2023-11-08
Enable incomplete course learner transmission

MueezKhan246:
Description:
Enable incomplete course learner transmission through flag in configuration.

JIRA:
https://2u-internal.atlassian.net/jira/software/c/projects/ENT/issues/ENT-7837


tests/test_integrated_channels/test_integrated_channel/test_transmitters/test_learner_data.py
```diff
@@ -194,3 +194,51 @@ def test_learner_data_transmission_dry_run_mode(self, already_transmitted_mock,
...
+        serialized_payload = learner_data_transmission_audit_mock.serialize(enterprise_configuration=self.enterprise_config)
+        encoded_serialized_payload = encode_data_for_logging(serialized_payload)
+        self.learner_transmitter.transmit(
+            LearnerExporterMock,
+            remote_user_id='user_id'
+        )
+        # with enable_incomplete_progress_transmission = True we should be able to call this method
+        assert self.learner_transmitter.client.create_course_completion.called
+
+        # Set feature flag to false
```
sameenfatima78, line 230: nit: should call it a boolean flag instead of feature flag

integrated_channels/integrated_channel/transmitters/learner_data.py
```diff
@@ -309,7 +309,10 @@ def transmit(self, payload, **kwargs):  # pylint: disable=arguments-differ
                 enterprise_enrollment_id
             )
 
-            if not learner_data.course_completed:
+            if not learner_data.course_completed and (
+                not hasattr(self.enterprise_configuration, 'enable_incomplete_progress_transmission')
+                or not self.enterprise_configuration.enable_incomplete_progress_transmission
+            ):
                 # The user has not completed the course, so we shouldn't send a completion status call
```
sameenfatima78, line 316: we need to update this comment to include enable transmission flag info

sameenfatima78 (approved):
LGTM

integrated_channels/integrated_channel/transmitters/learner_data.py
```diff
@@ -309,7 +309,10 @@ def transmit(self, payload, **kwargs):  # pylint: disable=arguments-differ
                 enterprise_enrollment_id
             )
 
-            if not learner_data.course_completed:
+            if not learner_data.course_completed and (
+                not hasattr(self.enterprise_configuration, 'enable_incomplete_progress_transmission')
```
zamanafzal, line 313: why we are checking `enable_incomplete_progress_transmission` two times?

zamanafzal: will we send the data to integrated_channel when the learner has passed the course?

MueezKhan246: > why we are checking `enable_incomplete_progress_transmission` two times?

@zamanafzal in test it is being tested twice so both the cases where if it is either set to true or false can be observed.

> will we send the data to integrated_channel when the learner has passed the course?
@zamanafzal well the approach was when this condition "not learner_data.course_completed" will be false the newly added checks won't affect current flow 
and so yes we should be able to send the data to integrated_channel when the learner has passed the course.

cc. @sameenfatima78

zamanafzal: I guess it doesn't make sense to add it two times.
we can use it like this
`if not learner_data.course_completed and not getattr(self.enterprise_configuration, 'enable_incomplete_progress_transmission', False):`

MueezKhan246: @zamanafzal yes, it will be better to refactor it like you suggested.

PR #1941, opened 2023-11-09
feat: Add an api-docs page that lives at `[LMS_ROOT_URL]/enterprise/api-docs/`

iloveagent57:
We somehow haven't had api-docs available for edx-enterprise for a while.  This adds them such that they exist at http://localhost:18000/enterprise/api-docs/

**Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,12 @@ Change Log
 
 Unreleased
 ----------
+Nothing unreleased.
+
+[4.7.0]
+--------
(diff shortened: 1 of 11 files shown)


alex-sheehan-edx (approved):
I think you have to try/catch openedx imports? 
ie 
```
  File "/home/runner/work/edx-enterprise/edx-enterprise/enterprise/urls.py", line 7, in <module>
    from edx_api_doc_tools import make_api_info, make_docs_urls
```
 other than that and the linter, lgtm

PR #1951, opened 2023-11-28
feat: add marked_authorized flag to SSO config

marlonkeating:
[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-7990)

## TODO
- [x] Error `"Field 'marked_authorized' doesn't have a default value"`
- [ ] Bump version

**Merge checklist:**

**Post merge:**


alex-sheehan-edx (commented):
You're going to want to backfill the data for existing records since this is null and blank false. Otherwise looks good to me :)

enterprise/models.py
```diff
@@ -3955,6 +3955,15 @@ class Meta:
         )
     )
 
+    marked_authorized = models.BooleanField(
+        blank=False,
+        null=False,
+        default=False,
+        help_text=_(
+            "If the edX Service Provider metadata was authorized from the Identity Provider side."
```
alex-sheehan-edx, line 3963: maybe make a comment that it's the admin indicating the service provider metadata was uploaded- it's come back to bite us in the past when the admin mindlessly checks the box and doesn't actually do the thing we're asking and we don't want to confuse any devs/ecs that the actions been done... more so that they TOLD us it was done lol

enterprise/migrations/0197_auto_20231130_2239.py
```diff
@@ -0,0 +1,23 @@
...
+
+    dependencies = [
+        ('enterprise', '0196_backfill_sso_marked_authorized'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='enterprisecustomerssoconfiguration',
+            name='marked_authorized',
+            field=models.BooleanField(default=False, help_text='Whether admin has indicated the service provider metadata was uploaded.'),
```
alex-sheehan-edx, line 16: huh I would have thought you needed to specify the blank/null values in the migration... I guess that's an ORM level thing and not a DB level thing? 🤔 I should really know this 😂

alex-sheehan-edx: uhh scratch that.. you do it need it... what is this migration doing?

alex-sheehan-edx: I see this is setting the default value- I think this can go on the original migration file? 0195

alex-sheehan-edx (approved):
tested locally, looks great

alex-sheehan-edx:
of course don't forget version bump/change log etc

PR #1966, opened 2023-12-09
Removing unencrypted user credentials data

MueezKhan246:
Description:
removed un encrypted user credentials data and testing feature flag.

JIRA:
https://2u-internal.atlassian.net/jira/software/c/projects/ENT/issues/ENT-5613


MueezKhan246:
@sameenfatima78 
This PR is to be deployed after [Replacing Encrypted Fields With Non Encrypted](https://github.com/openedx/edx-enterprise/pull/1897) gets tested on prod.

integrated_channels/moodle/migrations/0031_auto_20231204_0737.py
```diff
@@ -0,0 +1,23 @@
...
+            model_name='historicalmoodleenterprisecustomerconfiguration',
+            name='use_encrypted_user_data',
+            field=models.BooleanField(default=False, help_text='When set to True, the configured customer will use encrypted columns data to make client requests, this a boolean flag for testing purpose'),
+        ),
+        migrations.AddField(
+            model_name='moodleenterprisecustomerconfiguration',
+            name='use_encrypted_user_data',
+            field=models.BooleanField(default=False, help_text='When set to True, the configured customer will use encrypted columns data to make client requests, this a boolean flag for testing purpose'),
+        ),
+    ]
```
sameenfatima78, line 23: [Question] These fields are added here but are being removed in the second migration?

integrated_channels/moodle/migrations/0032_auto_20231208_2345.py
```diff
@@ -0,0 +1,45 @@
+# Generated by Django 3.2.23 on 2023-12-08 23:45
```
sameenfatima78, line 1: We can skip this migration for now. Let's first check whether the transmissions work for rest of the customers and then we can add the migration later on.

integrated_channels/moodle/models.py
```diff
@@ -57,15 +57,6 @@ class MoodleEnterpriseCustomerConfiguration(EnterpriseCustomerPluginConfiguratio
...
 
-    username = models.CharField(
-        max_length=255,
-        verbose_name="Webservice Username",
-        blank=True,
-        help_text=_(
-            "The API user's username used to obtain new tokens."
-        )
-    )
-
```
sameenfatima78, line 68: As discussed, we should keep the 3 fields in the model for now

integrated_channels/moodle/client.py
```diff
@@ -143,12 +142,7 @@ def __init__(self, enterprise_configuration):
         """
         super().__init__(enterprise_configuration)
         self.config = apps.get_app_config('moodle')
-        token = (
-            enterprise_configuration.decrypted_token
-            if getattr(settings, 'FEATURES', {}).get('USE_ENCRYPTED_USER_DATA', False)
-            else enterprise_configuration.token
-        )
-        self.token = token or self._get_access_token()
+        self.token = enterprise_configuration.decrypted_token or self._get_access_token()
```
sameenfatima78, line 145: Instead of removing this feature flag, let's enable it for prod in `edx-internal` first. Once testing is done on prod, we can then remove this flag and reference the decrypted value directly. 

**Benefits**: If something goes wrong, we can disable the feature flag back in edx-internal that offers speedy deployment within 6-7 minutes. Wherease edx-enterprise takes around 3-4 hours to get deployed.

sameenfatima78 (changes requested):

integrated_channels/moodle/migrations/0031_auto_20231204_0737.py
MueezKhan246: @sameenfatima78 this was added when I was using the feature flag by keeping it inside the config model.
should be removed.

integrated_channels/moodle/migrations/0032_auto_20231208_2345.py
MueezKhan246: @sameenfatima78 updated.

integrated_channels/moodle/models.py
MueezKhan246: @sameenfatima78 updated.

integrated_channels/moodle/client.py
MueezKhan246: that can be better approach for testing on prod.
@sameenfatima78 updated.

hamzawaleed01 (approved):

hamzawaleed01:
LGTM! let's wait for a review from @zamanafzal and @sameenfatima78

integrated_channels/api/v1/moodle/serializers.py
```diff
@@ -12,8 +14,12 @@ class Meta:
             'moodle_base_url',
             'service_short_name',
             'category_id',
-            'username',
-            'password',
-            'token',
+            'encrypted_username',
```
zamanafzal, line 17: Do we currently have configs using decrypted data?

MueezKhan246: @zamanafzal the goal was to only provide encrypted values other than where its decrypted value is required on our end i.e, while making requests using API client

zamanafzal (approved):

integrated_channels/api/v1/moodle/serializers.py
```diff
@@ -1,6 +1,8 @@
 """
```
sameenfatima78, line 1: As discussed in the meeting, let's first test the serializer changes first by hitting the endpoint and verifying results. We should also check where else is this serializer being used.

integrated_channels/moodle/models.py
```diff
@@ -230,7 +230,10 @@ def is_valid(self):
         incorrect_items = {'incorrect': []}
         if not self.moodle_base_url:
             missing_items.get('missing').append('moodle_base_url')
-        if not self.token and not (self.username and self.password):
+        if (not self.token and not (self.username and self.password)) or (
+            not self.decrypted_token
+            and not (self.decrypted_username and self.decrypted_password)
+        ):
```
sameenfatima78, line 236: shouldn't we directly just check the decrypted values here now that we are completely replacing the fields?

integrated_channels/api/v1/moodle/serializers.py
MueezKhan246: @sameenfatima78 for now unencrypted values should not be removed from the serializer?

integrated_channels/moodle/models.py
MueezKhan246: @sameenfatima78 as both unencrypted and encrypted values exist for now 
therefore i was making the check using both.

integrated_channels/api/v1/moodle/serializers.py
sameenfatima78: First we should check where exactly is this data being used, then we can merge the PR as is with encrypted columns.

integrated_channels/moodle/models.py
sameenfatima78: We should merge the PR with new columns now. The old can stay in DB but this pr should handle entirely replacing the fields with encrypted columns everywhere.

sameenfatima78 (approved):
Looks good once we have added the version upgrade files.

PR #1969, opened 2023-12-18
[ENT-7950] - Post skills metadata for Degreed2 

hamzawaleed01:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api_client/enterprise_catalog.py
+++ b/enterprise/api_client/enterprise_catalog.py
@@ -318,7 +318,7 @@ class EnterpriseCatalogApiClient(UserAPIClient):
     @UserAPIClient.refresh_token
     def get_content_metadata_content_identifier(self, enterprise_uuid, content_id):  # pylint: disable=inconsistent-return-statements
         """
-        Return all content metadata contained in the catalogs associated with the the
+        Return all content metadata contained in the catalogs associated with the
         given EnterpriseCustomer and content_id.
         """
         try:
@@ -341,6 +341,7 @@ class EnterpriseCatalogApiClient(UserAPIClient):
                 "Exception raised in EnterpriseCatalogApiClient::get_content_metadata_content_identifier: [%s]",
                 str(exc),
             )
+            return {}
 
 
 class NoAuthEnterpriseCatalogClient(NoAuthAPIClient):
--- a/enterprise/api_client/enterprise_catalog.py
+++ b/enterprise/api_client/enterprise_catalog.py
@@ -28,9 +28,9 @@ class EnterpriseCatalogApiClient(UserAPIClient):
     REFRESH_CATALOG_ENDPOINT = ENTERPRISE_CATALOG_ENDPOINT + '/{}/refresh_metadata'
     CATALOG_DIFF_ENDPOINT = ENTERPRISE_CATALOG_ENDPOINT + '/{}/generate_diff'
     ENTERPRISE_CUSTOMER_ENDPOINT = 'enterprise-customer'
-    CONTENT_METADATA_IDENTIFIER_ENDPOINT = (
-        ENTERPRISE_CUSTOMER_ENDPOINT + "/{}/content-metadata/" + "{}"
-    )
+    CONTENT_METADATA_IDENTIFIER_ENDPOINT = ENTERPRISE_CUSTOMER_ENDPOINT + \
+        "/{}/content-metadata/" + "{}"
+    
     APPEND_SLASH = True
     GET_CONTENT_METADATA_PAGE_SIZE = getattr(settings, 'ENTERPRISE_CATALOG_GET_CONTENT_METADATA_PAGE_SIZE', 50)
 
--- a/enterprise/api_client/enterprise_catalog.py
+++ b/enterprise/api_client/enterprise_catalog.py
@@ -8,6 +8,7 @@ from logging import getLogger
 from urllib.parse import urljoin
 
 from requests.exceptions import ConnectionError, RequestException, Timeout  # pylint: disable=redefined-builtin
+from rest_framework.exceptions import NotFound
 
 from django.conf import settings
 
@@ -30,7 +31,6 @@ class EnterpriseCatalogApiClient(UserAPIClient):
     ENTERPRISE_CUSTOMER_ENDPOINT = 'enterprise-customer'
     CONTENT_METADATA_IDENTIFIER_ENDPOINT = ENTERPRISE_CUSTOMER_ENDPOINT + \
         "/{}/content-metadata/" + "{}"
-    
     APPEND_SLASH = True
     GET_CONTENT_METADATA_PAGE_SIZE = getattr(settings, 'ENTERPRISE_CATALOG_GET_CONTENT_METADATA_PAGE_SIZE', 50)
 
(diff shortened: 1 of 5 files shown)


enterprise/api_client/enterprise_catalog.py
```diff
@@ -28,6 +28,9 @@ class EnterpriseCatalogApiClient(UserAPIClient):
     REFRESH_CATALOG_ENDPOINT = ENTERPRISE_CATALOG_ENDPOINT + '/{}/refresh_metadata'
     CATALOG_DIFF_ENDPOINT = ENTERPRISE_CATALOG_ENDPOINT + '/{}/generate_diff'
     ENTERPRISE_CUSTOMER_ENDPOINT = 'enterprise-customer'
+    CONTENT_METADATA_IDENTIFIER_ENDPOINT = (
+        ENTERPRISE_CUSTOMER_ENDPOINT + "/{}/content-metadata/" + "{}"
+    )
```
sameenfatima78, line 33: [Question] Why is the string enclosed in a tuple here?

enterprise/api_client/enterprise_catalog.py
```diff
@@ -312,6 +315,18 @@ def enterprise_contains_content_items(self, enterprise_uuid, content_ids):
...
+        """
+        Return all content metadata contained in the catalogs associated with the the
+        given EnterpriseCustomer and content_id.
+        """
+        api_url = self.get_api_url(
+            f"{self.CONTENT_METADATA_IDENTIFIER_ENDPOINT.format(enterprise_uuid, content_id)}")
+        response = self.client.get(api_url)
+        response.raise_for_status()
+        return response.json()
+
```
sameenfatima78, line 329: Shouldn't we handle any exceptions here? like connection failure with enterprise-catalogs service. I'd suggest we take a look at this endpoint in enterprise-catalogs to see what kind of error does it raise.

integrated_channels/degreed2/client.py
```diff
@@ -275,6 +276,27 @@ def create_content_metadata(self, serialized_data):
...
+            )
+        )
+        # 2. Transmit to degreed
+        skills = metadata['skill_names']
+        if skills:
+            course_id = a_course.get("external-id")
+            self.assign_course_skills(course_id, skills)
         return status_code, response_body
 
     def update_content_metadata(self, serialized_data):
```
sameenfatima78, line 302: the skills should also be assigned when course is updated.

enterprise/api_client/enterprise_catalog.py
hamzawaleed01: Formatter did this 😮‍💨

enterprise/api_client/enterprise_catalog.py
```diff
@@ -312,6 +315,33 @@ def enterprise_contains_content_items(self, enterprise_uuid, content_ids):
         response.raise_for_status()
         return response.json()['contains_content_items']
 
+    @UserAPIClient.refresh_token
+    def get_content_metadata_content_identifier(self, enterprise_uuid, content_id):  # pylint: disable=inconsistent-return-statements
+        """
+        Return all content metadata contained in the catalogs associated with the the
```
sameenfatima78, line 321: nit: remove the second "the"

enterprise/api_client/enterprise_catalog.py
```diff
@@ -312,6 +315,33 @@ def enterprise_contains_content_items(self, enterprise_uuid, content_ids):
...
+                enterprise_uuid,
+                content_id,
+                str(exc),
+            )
+            return {}
+        except (RequestException, ConnectionError, Timeout) as exc:
+            LOGGER.exception(
+                "Exception raised in EnterpriseCatalogApiClient::get_content_metadata_content_identifier: [%s]",
+                str(exc),
+            )
```
sameenfatima78, line 343: shouldn't we return something here?

integrated_channels/degreed2/client.py
```diff
@@ -275,6 +277,36 @@ def create_content_metadata(self, serialized_data):
                 f'Degreed2APIClient create_content_metadata failed with status {status_code}: {response_body}',
                 status_code=status_code
             )
+        # once course is created successfully, we need to do 2 more steps
+        # 1. Fetch skills from enterprise-catalog
+        client = EnterpriseCatalogApiClient()
```
sameenfatima78, line 282: wouldn't it be better to keep move client in the Degreed2 initialization? Otherwise it will create a new client for every course create and update call.

integrated_channels/degreed2/client.py
```diff
@@ -303,6 +335,36 @@ def update_content_metadata(self, serialized_data):
             patch_url,
             course_id
         )
+        # once course is updated successfully, we need to do 2 more steps
+        # 1. Fetch skills from enterprise-catalog
+        client = EnterpriseCatalogApiClient()
```
sameenfatima78, line 340: wouldn't it be better to keep this client in the Degreed2 initialization? Otherwise it will create a new client for every course create and update call.

integrated_channels/degreed2/client.py
```diff
@@ -275,6 +277,36 @@ def create_content_metadata(self, serialized_data):
...
+        )
+        LOGGER.info(
+            generate_formatted_log(
+                self.enterprise_configuration.channel_code(),
+                self.enterprise_configuration.enterprise_customer.uuid,
+                None,
+                None,
+                f'[Degreed2Client] metadata: {metadata}'
+            )
+        )
```
sameenfatima78, line 295: This log may not be providing us any useful info that isn't already being logged.

integrated_channels/degreed2/client.py
```diff
@@ -275,6 +277,36 @@ def create_content_metadata(self, serialized_data):
...
+            LOGGER.info(
+                generate_formatted_log(
+                    self.enterprise_configuration.channel_code(),
+                    self.enterprise_configuration.enterprise_customer.uuid,
+                    None,
+                    None,
+                    f'[Degreed2Client] transmitted skills: {metadata["skill_names"]},'
+                    f'for course: {external_id}'
+                )
+            )
```
sameenfatima78, line 309: We already are logging something similar in the assign_course_skill method so this log can be removed as well.

integrated_channels/degreed2/client.py
```diff
@@ -275,6 +277,36 @@ def create_content_metadata(self, serialized_data):
...
+            generate_formatted_log(
+                self.enterprise_configuration.channel_code(),
+                self.enterprise_configuration.enterprise_customer.uuid,
+                None,
+                None,
+                f'[Degreed2Client] metadata: {metadata}'
+            )
+        )
+        # 2. Transmit to degreed
+        skills = metadata['skill_names']
```
sameenfatima78, line 297: nit: It's always safe to use `metadata.get('skill_names', [])` just as a caution. `metadata['skill_names']` will throw a `KeyError` in case we return empty dict.

integrated_channels/degreed2/client.py
```diff
@@ -275,6 +277,36 @@ def create_content_metadata(self, serialized_data):
...
+        if skills:
+            self.assign_course_skills(external_id, skills)
```
sameenfatima78, line 299: This method `assign_course_skills` may raise `ClientError`, it's safer to enclose this call inside try/except block and handle clienterror.

integrated_channels/degreed2/client.py
```diff
@@ -303,6 +335,36 @@ def update_content_metadata(self, serialized_data):
...
+            LOGGER.info(
+                generate_formatted_log(
+                    self.enterprise_configuration.channel_code(),
+                    self.enterprise_configuration.enterprise_customer.uuid,
+                    None,
+                    None,
+                    f'[Degreed2Client] transmitted skills: {metadata["skill_names"]},'
+                    f'for course: {course_id}'
+                )
+            )
```
sameenfatima78, line 367: Following the DRY principe, this whole block can be put inside a helper method and be called inside create and update course method.

integrated_channels/degreed2/client.py
hamzawaleed01: yeah you're right! I just added it temporarily to see the exact response in production once to test things. if everything works as expected then I'll remove this log.

sameenfatima78 (approved):

PR #1979, opened 2024-01-04
feat: add "Setup auth org ID" action for Enterprise Customers

pwnage101:
ENT-8169

screenshots
===

Button NOT present when the `ENTERPRISE_SSO_ORCHESTRATOR_CONFIGURE_EDX_OAUTH_PATH` setting is NOT defined:
Screenshot 2024-01-04 at 16 46 42

Button present when the `ENTERPRISE_SSO_ORCHESTRATOR_CONFIGURE_EDX_OAUTH_PATH` setting is defined:
Screenshot 2024-01-05 at 13 44 19

Clicking the button leads you to a confirmation page with help text:
Screenshot 2024-01-05 at 13 45 05

Clicking "Setup auth_org_id" results in a success message displayed:
Screenshot 2024-01-05 at 13 45 11

Finally, clicking "Cancel" navigates back to the "Change Enterprise Customer" screen with updated value for the "Auth org id" field.

**Merge checklist:**

**Post merge:**


.python-version
```diff
@@ -1 +0,0 @@
-3.8.12/envs/venv
```
pwnage101, line 1: I assume this file was accidentally committed, since it theoretically conflicts with everybody else's local `.python-version`, and none of our other repos have this file committed.

enterprise/templates/enterprise/admin/setup_auth_org_id.html
```diff
@@ -0,0 +1,52 @@
...
+  {% endif %}
+  &rsaquo; {% if has_change_permission %}
+    <a href="{% url opts|admin_urlname:'change' enterprise_customer.uuid %}">
+      {{ enterprise_customer|truncatewords:"18" }}
+    </a>
+  {% else %}
+    {{ enterprise_customer|capfirst }}
+  {% endif %}
+  &rsaquo;
+  {% trans "Setup auth org ID" %}
```
pwnage101, line 21: I don't know where to put this translation, do I need to run a make command or something?

enterprise/admin/views.py
```diff
@@ -174,7 +176,8 @@ def get_form_view(self, request, customer_uuid, additional_context=None):
         render the form with appropriate context.
         """
         context = self._build_context(request, customer_uuid)
-        context.update(additional_context)
+        if additional_context:
+            context.update(additional_context)
```
pwnage101, line 180: This fixes a bug when the caller of `get_form_view()` doesn't pass `additional_context`.

pwnage101:
* Updated text to just match what Joe proposed, with minor alterations.  
* Updated the text "Auth org id" to be consistent across all user-facing text, and matches the pre-existing field display name.

enterprise/api_client/sso_orchestrator.py
```diff
@@ -67,6 +68,14 @@ def _get_orchestrator_configure_url(self):
         # probably want config value validated for this
         return urljoin(self.base_url, get_sso_orchestrator_configure_path())
 
+    def _get_orchestrator_configure_edx_oauth_url(self):
+        """
+        get the configure-edx-oautht url for the SSO Orchestrator API
```
adamstankiewicz, line 73: nit: is `oautht` a typo?

enterprise/templates/enterprise/admin/setup_auth_org_id.html
```diff
@@ -0,0 +1,54 @@
...
+</div>
+{% endblock %}
+
+{% block content %}
+<div id="content-main">
+  <div class="forms-panel">
+    <h1>{% trans "Setup Auth org id" %}</h1>
+    <p>
+      This action is required for customers who will have learners in executive education courses. Setting up the
+      Auth org id will enable the enterprise's learners to take Exec ed or OCM courses using the same set of login
```
adamstankiewicz, line 31: nit: `Exec Ed` (capital `Ed`)?

enterprise/templates/enterprise/admin/setup_auth_org_id.html
```diff
@@ -0,0 +1,54 @@
...
+      credentials. Clicking the button below will facilitate the necessary steps with our identity vendor, Auth0, and
```
adamstankiewicz, line 32: nit: perhaps "**external** identity vendor"?

enterprise/templates/enterprise/admin/setup_auth_org_id.html
```diff
@@ -0,0 +1,54 @@
...
+      will overwrite any value that may already be in the "Auth org id" field.
+    </p>
+    <form action="" method="post" id="setup-auth-org-id-form">
+      {% csrf_token %}
+      <div style="display:flex";>
+          <div style="vertical-align:top">
+              <input type="hidden" name="post" value="submit" />
+              <input type="hidden" name="action" value="submit" />
+              <input type="submit" value="{% trans "Setup Auth org id" %}" />
```
adamstankiewicz, line 41: [consider] May be helpful to make this button more prominent than the "Cancel" button to the right. Given this button sort of acts like a save button for the `EnterpriseCustomer`, I might recommend applying the `var(--default-button-bg)` CSS variable as the button background (same background color as Django's default "Save"):

image

```suggestion
              <input type="submit" value="{% trans "Setup Auth org id" %}" style="background: var(--default-button-bg);" />
```

enterprise/templates/enterprise/admin/setup_auth_org_id.html
```diff
@@ -0,0 +1,54 @@
...
+          </div>
+          <div style="margin-left:0.5rem">
+              <input type="button" value="Cancel" onclick="history.go(-1)"/>
```
adamstankiewicz, line 44: [consider] May be helpful to make this button less prominent than the "Setup Auth org id" button to the left. Might recommend applying the `var(--close-button-bg)` CSS variable as the button background (same background color as Django's other cancel-type buttons):

image

```suggestion
              <input type="button" value="Cancel" onclick="history.go(-1)" style="background: var(--close-button-bg);" />
```

adamstankiewicz, line 44: nit: should this string also include `trans` to mark it for translation similar to the other button text?

enterprise/templates/enterprise/admin/setup_auth_org_id.html
iloveagent57: You can run `make extract_translations` I think? https://github.com/openedx/edx-platform/blob/2209549e68776f5db3c77a571bba115918fcf3f7/Makefile#L40-L53
That said, you're making changes to an internal tool, so I wouldn't let this be a blocker.

iloveagent57 (approved):

PR #1983, opened 2024-01-09
ENT-8018 - feat: create integrated channel API request log table

hamzawaleed01:
**Merge checklist:**

**Post merge:**


integrated_channels/cornerstone/migrations/0031_cornerstoneapirequestlogs.py
```diff
@@ -0,0 +1,24 @@
...
+        migrations.CreateModel(
+            name='CornerstoneAPIRequestLogs',
+            fields=[
+                ('integratedchannelapirequestlogs_ptr', models.OneToOneField(auto_created=True, on_delete=django.db.models.deletion.CASCADE, parent_link=True, primary_key=True, serialize=False, to='integrated_channel.integratedchannelapirequestlogs')),
+                ('user_agent', models.CharField(max_length=255)),
+                ('user_ip', models.GenericIPAddressField()),
+            ],
+            bases=('integrated_channel.integratedchannelapirequestlogs',),
+        ),
+    ]
```
sameenfatima78, line 24: Question: Curious why we created a separate model for cornerstone?

integrated_channels/integrated_channel/models.py
```diff
@@ -879,3 +879,48 @@ class Meta:
...
+
+
+class IntegratedChannelAPIRequestLogs(TimeStampedModel):
+    """
+     A model to track basic information about every API call we make from the integrated channels.
+    """
+    enterprise_customer = models.ForeignKey(
+        EnterpriseCustomer, on_delete=models.CASCADE)
+    enterprise_customer_configuration_uuid = models.UUIDField(
+        blank=False, null=False)
```
sameenfatima78, line 891: isn't this a plain text id instead of uuid?

sameenfatima78 (commented):
Left some questions, plus we need to add tests for the two models for the dunder methods

integrated_channels/cornerstone/migrations/0031_cornerstoneapirequestlogs.py
hamzawaleed01: For CSOD we've two extra fields. See line#19-20

integrated_channels/integrated_channel/models.py
hamzawaleed01: oh yes,, lemme fix it! 🥇

sameenfatima78 (approved):

integrated_channels/cornerstone/models.py
```diff
@@ -318,3 +319,37 @@ class CornerstoneCourseKey(models.Model):
...
     class Meta:
         app_label = 'cornerstone'
+
+
+class CornerstoneAPIRequestLogs(IntegratedChannelAPIRequestLogs):
+    """
+     A model to track basic information about every API call we make from the integrated channels.
+    """
+    user_agent = models.CharField(max_length=255)
+    user_ip = models.GenericIPAddressField(blank=False, null=False)
```
zamanafzal, line 329: is it confirmed that this field will always have data?

integrated_channels/cornerstone/models.py
```diff
@@ -318,3 +319,37 @@ class CornerstoneCourseKey(models.Model):
...
+    def __str__(self):
+        """
+        Return a human-readable string representation of the object.
+        """
+        return (
+            f'<CornerstoneAPIRequestLogs {self.id}'
+            f' for enterprise customer {self.enterprise_customer}, '
+            f', enterprise_customer_configuration_id: {self.enterprise_customer_configuration_id}>'
+            f', endpoint: {self.endpoint}'
+            f', payload: {self.payload}'
```
zamanafzal, line 343: I think we should remove the payload from here as this will be a huge string.

integrated_channels/cornerstone/models.py
hamzawaleed01: Good catch! I think no, CSOD right now is _pull_ not _push_ so an HTTP call typically includes standard information like user agent and IP. But I think it should be optional since there are many endpoints in CSOD client which are POST and won't have IP

zamanafzal (approved):

PR #1988, opened 2024-01-16
Replaced unencrypted columns of user data credentials in moodle config

MueezKhan246:
Description:
Replaced unencrypted columns of user data credentials in moodle config.
As discussed the previously unencrypted columns can be deprecated and the their name can be assigned to the new encrypted columns.

JIRA:
https://2u-internal.atlassian.net/jira/software/c/projects/ENT/issues/ENT-8032


justEhmadSaeed (approved):
lgtm

integrated_channels/moodle/client.py
```diff
@@ -196,8 +184,8 @@ def _get_access_token(self):
                 'Content-Type': 'application/x-www-form-urlencoded',
             },
             data={
-                "username": decrypted_username if use_encrypted_user_data else username,
-                "password": decrypted_password if use_encrypted_user_data else password,
+                "username": self.enterprise_configuration.username,
```
zamanafzal, line 187: we are changing it now because `username` is now encrypted for every moddle config?

zamanafzal (approved):

integrated_channels/moodle/migrations/0031_auto_20240116_1158.py
```diff
@@ -0,0 +1,98 @@
...
+            model_name='moodleenterprisecustomerconfiguration',
+            name='decrypted_password',
+        ),
+        migrations.RemoveField(
+            model_name='moodleenterprisecustomerconfiguration',
+            name='decrypted_token',
+        ),
+        migrations.RemoveField(
+            model_name='moodleenterprisecustomerconfiguration',
+            name='decrypted_username',
```
sameenfatima78, line 36: Have we tried to run this migration locally on some test data to ensure no data is lost during the process? I believe removing the field may cause data loss. We should look online if Django provides a solution for this. Apparently there is a model state and then there is a db state. We'll have to rename the field at both levels. https://b0uh.github.io/django-model-change-field-name-with-no-db-impact.html

sameenfatima78: https://stackoverflow.com/questions/3498140/migrating-django-model-field-name-change-without-losing-data

sameenfatima78 (changes requested):

integrated_channels/moodle/migrations/0031_auto_20240116_1158.py
MueezKhan246: @sameenfatima78 yes the auto-generated migration was prone to data loss,
I have restructured the migrations to prevent data loss, the approach i am using seemed simplest to me
I have used two separate migrations so that first the unencrypted columns become "deprecated" and only after that the name of the the older unencrypted columns can get assigned encrypted columns

sameenfatima78 (approved):
Good to go! 🚢

PR #1989, opened 2024-01-17
feat: mgmt command to explore removing exec-ed cat flag

johnnagro:
## Description

A mgmt command to enumerate the enterprise customer catalogs and test a query modification idea. This tests the new filter in `search/all` and compares the counts to the existing catalog in enterprise catalog service (which honors the flags). The goal is for these counts to be the same after my proposed query modification.

The goal is to run this mgmt command in PROD and review the results. If the data checks out, we can make a similar mgmt command to actually alter these catalogs. Then we can remove the flag in this system and in the catalog service.

- https://2u-internal.atlassian.net/browse/ENT-7893
- https://github.com/openedx/course-discovery/pull/4236

### DIFF
--- a/enterprise/management/commands/compare_discovery_and_enterprise_catalogs.py
+++ b/enterprise/management/commands/compare_discovery_and_enterprise_catalogs.py
@@ -1,0 +1,76 @@
+"""
+Django management command to explore Exec Ed inclusion flag migration
+"""
+
+import copy
+import json
+import logging
+
+from django.core.management import BaseCommand
+
+from enterprise.api_client.discovery import CourseCatalogApiServiceClient
+from enterprise.api_client.enterprise_catalog import EnterpriseCatalogApiClient
+from enterprise.models import EnterpriseCatalogQuery, EnterpriseCustomerCatalog
+from integrated_channels.utils import batch_by_pk
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Enumerate the catalog filters and log information about how we might migrate them.
+    """
+
+    def handle(self, *args, **options):
+        enterprise_catalog_client = EnterpriseCatalogApiClient()
+        discovery_client = CourseCatalogApiServiceClient()
+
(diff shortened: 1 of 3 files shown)


enterprise/management/commands/compare_discovery_and_enterprise_catalogs.py
```diff
@@ -0,0 +1,63 @@
...
+        for catalog_query_batch in batch_by_pk(EnterpriseCatalogQuery):
+            for catalog_query in catalog_query_batch:
+                logger.info(f'{catalog_query.id} {catalog_query.include_exec_ed_2u_courses}')
+
+                if catalog_query.include_exec_ed_2u_courses:
+                    logger.info(f'{catalog_query.id} already includes exec ed')
+                    continue
+
+                if catalog_query.content_filter.get('course_type'):
+                    logger.info(f'{catalog_query.id} already references course_type somehow')
```
alex-sheehan-edx, line 37: lol

enterprise/management/commands/compare_discovery_and_enterprise_catalogs.py
```diff
@@ -0,0 +1,63 @@
...
+                new_content_filter['course_type__exclude'] = 'executive-education-2u'
+                new_content_filter_json = json.dumps(new_content_filter)
+                logger.info(f'{catalog_query.id} new filter: {new_content_filter_json}')
+
+        for cusrtomer_catalog_batch in batch_by_pk(EnterpriseCustomerCatalog):
+            for customer_catalog in cusrtomer_catalog_batch:
+                logger.info(f'{customer_catalog.uuid}')
+
+                if customer_catalog.content_filter.get('course_type'):
+                    logger.info(f'{customer_catalog.uuid} already references course_type somehow')
```
alex-sheehan-edx, line 50: just for clarity sake, should you differentiate this from the above log? To avoid confusing debugging sessions

alex-sheehan-edx (approved):
looks good, tiniest of logging nits

PR #1994, opened 2024-01-22
feat: update cornerstone client to store API calls in DB

hamzawaleed01:
…in db

**Merge checklist:**

**Post merge:**


tests/test_integrated_channels/test_cornerstone/test_client.py
```diff
@@ -0,0 +1,55 @@
+"""
```
sameenfatima78, line 1: Nice of you for adding the missing tests 💯

integrated_channels/utils.py
```diff
@@ -501,3 +508,38 @@ def get_enterprise_client_by_channel_code(channel_code):
...
+    enterprise_customer,
+    enterprise_customer_configuration_id,
+    endpoint,
+    payload,
+    time_taken,
+    status_code,
+    response_body,
+):
+    """
+    Creates new record in CornerstoneAPIRequestLogs table.
```
sameenfatima78, line 523: nit: CornerstoneAPIRequestLogs -> IntegratedChannelAPIRequestLogs

integrated_channels/utils.py
```diff
@@ -501,3 +508,38 @@ def get_enterprise_client_by_channel_code(channel_code):
...
+        LOGGER.error(
+            f"store_api_call raised error while storing API call: {e}"
+            f"enterprise_customer={enterprise_customer}"
+            f"enterprise_customer_configuration_id={enterprise_customer_configuration_id},"
+            f"endpoint={endpoint}"
+            f"payload={payload}"
+            f"time_taken={time_taken}"
+            f"status_code={status_code}"
+            f"response_body={response_body}"
+        )
```
sameenfatima78, line 545: I think object initialization functionality can be included within class, just like we do in the constructor (init). I believe this method can be added into the model itself as a class method. https://stackoverflow.com/questions/10586787/when-should-i-use-classmethod-and-when-def-methodself

integrated_channels/integrated_channel/models.py
```diff
@@ -881,41 +881,58 @@ class Meta:
     resolved = models.BooleanField(default=False)
 
 
-class IntegratedChannelAPIRequestLogs(TimeStampedModel):
+class BaseIntegratedChannelAPIRequestLogs(TimeStampedModel):
```
sameenfatima78, line 884: [Question] Curious to know the intention behind creating this abstract class.

integrated_channels/integrated_channel/admin/__init__.py
```diff
@@ -85,3 +89,13 @@ class ApiResponseRecordAdmin(admin.ModelAdmin):
...
+
+
+@admin.register(IntegratedChannelAPIRequestLogs)
+class CornerstoneAPIRequestLogAdmin(admin.ModelAdmin):
+    """
+    Django admin model for IntegratedChannelAPIRequestLogs.
+    """
+
+    class Meta:
+        model = IntegratedChannelAPIRequestLogs
```
sameenfatima78, line 101: Shouldn't this class be removed now that we only have one table for all channels?

integrated_channels/integrated_channel/migrations/0031_alter_integratedchannelapirequestlogs_options.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 3.2.22 on 2024-01-25 09:25
```
sameenfatima78, line 1: We can create one single migration for these two migrations: 

- integrated_channels/integrated_channel/migrations/0031_alter_integratedchannelapirequestlogs_options.py
- integrated_channels/integrated_channel/migrations/0032_auto_20240125_0936.py

integrated_channels/integrated_channel/models.py
```diff
@@ -881,41 +881,58 @@ class Meta:
...
+        EnterpriseCustomer, on_delete=models.CASCADE
+    )
     enterprise_customer_configuration_id = models.IntegerField(
-        blank=False, null=False, help_text='ID from the EnterpriseCustomerConfiguration model')
-    endpoint = models.TextField(blank=False, null=False)
+        blank=False,
+        null=False,
+        help_text="ID from the EnterpriseCustomerConfiguration model",
+    )
+    endpoint = models.TextField(
```
sameenfatima78, line 897: Will this field be saving urls? If yes, we can use Django's URLField. It's just a CharField with url validation. https://docs.djangoproject.com/en/5.0/ref/models/fields/#urlfield

sameenfatima78 (changes requested):

integrated_channels/integrated_channel/models.py
hamzawaleed01: So In my previous PR where I extended/inherited Cornerstone model from IntergratedChannelApiRequestLogs model - when I looked at the table, it basically created two tables(first one with all columns and second one with ForeignKey relation with the first one). 
But I wanted two different tables so had to create a base abstract class; which we can remove now if we want but just left it as it is in case if needed in future.

hamzawaleed01: Removed!

integrated_channels/integrated_channel/admin/__init__.py
hamzawaleed01: My bad! 😮‍💨 removed

integrated_channels/integrated_channel/migrations/0031_alter_integratedchannelapirequestlogs_options.py
hamzawaleed01: Done!

integrated_channels/cornerstone/client.py
```diff
@@ -113,6 +117,16 @@ def create_course_completion(self, user_id, payload):
                 'Content-Type': 'application/json'
             }
         )
+        duration_seconds = time.time() - start_time
+        self.IntegratedChannelAPIRequestLogs.store_api_call(
```
sameenfatima78, line 121: [Question] Any good reason to store this as an attribute within class?

sameenfatima78 (approved):
Just one question but looks good otherwise

integrated_channels/cornerstone/client.py
hamzawaleed01: nope, I can move it outside the class - which I suggested to Mueez as well 😄

hamzawaleed01: I tried moving it outside the class but got `AppRegistryNotReady("Models aren't loaded yet.")`. Now placing `get_model` inside `create_course_completion`

PR #1998, opened 2024-01-23
feat: mgmt command to migrate catalog exclusions

johnnagro:
## Description
- mgmt command to enumerate the catalog filters and add course_type exclusions where required
- https://github.com/openedx/edx-enterprise/pull/1989
- https://2u-internal.atlassian.net/browse/ENT-7893

### DIFF
--- a/enterprise/management/commands/add_exec_ed_exclusion_to_catalogs.py
+++ b/enterprise/management/commands/add_exec_ed_exclusion_to_catalogs.py
@@ -1,0 +1,90 @@
+"""
+Django management command to add Exec Ed exclusion flag to catalogs
+"""
+
+import logging
+
+from django.core.management import BaseCommand
+
+from enterprise.models import EnterpriseCatalogQuery, EnterpriseCustomerCatalog
+from integrated_channels.utils import batch_by_pk
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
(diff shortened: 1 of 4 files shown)


enterprise/management/commands/add_exec_ed_exclusion_to_catalogs.py
```diff
@@ -0,0 +1,90 @@
...
+                    continue
+
+                if catalog_query.content_filter.get('course_type'):
+                    logger.info(
+                        'add_exec_ed_exclusion_to_catalogs '
+                        f'query {catalog_query.id} already references course_type somehow'
+                    )
+                    continue
+
+                if catalog_query.content_filter.get('aggregation_key'):
```
alex-sheehan-edx, line 46: why is this an issue?

alex-sheehan-edx: meaning, for the sake of tagging query filters for exec ed, why do we care about aggregation key? Because it's selective on a lower lever than course type?

alex-sheehan-edx (approved):
one question but otherwise lgtm

enterprise/management/commands/add_exec_ed_exclusion_to_catalogs.py
johnnagro: the rationale is that if they're selecting specific courses they already know which are which

PR #2003, opened 2024-01-28
fix: add missing comma to catalog query fields list.

iloveagent57:
**Post merge:**


alex-sheehan-edx (approved):
:shipit:

PR #2023, opened 2024-02-16
Added channel_name for api call logs records

MueezKhan246:
Description:
Added channel_name for making it easier to delete api call logs records.

JIRA:
https://2u-internal.atlassian.net/jira/software/c/projects/ENT/issues/ENT-8020


justEhmadSaeed (approved):
lgtm

sameenfatima78 (commented):
we shouldn't hardcode the name. try using constants instead

MueezKhan246:
> we shouldn't hardcode the name. try using constants instead
@sameen i was searching if the names were defined in a dicts or enums but could not find it.
just replaced it with `self.enterprise_configuration.channel_code()` shared by @hamzawaleed01.

hamzawaleed01 (approved):

sameenfatima78 (commented):
looks all good!

PR #2040, opened 2024-03-01
feat: adding management command to remove expired pending group memberships

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-8524

In this PR:
- Updated EnterpriseCustomerUser readonly serializer to pull group UUIDs instead of membership records
- Updated EnterpriseCustomerUser readonly serializer to account for groups that apply to all contexts
- Updated EnterpriseGroup `/learners/` endpoint to account for groups that apply to all contexts
- New ``applies_to_all_contexts`` flag on EnterpriseGroup records
- New management command ``remove_expired_pending_group_memberships`` that will remove pending membership records if they are older than 90 days
- Updated factories
- Associated unit tests
- Fix to the enterprise UserFactory class to avoid collisions in test data

**Merge checklist:**

**Post merge:**


enterprise/management/commands/remove_expired_pending_group_memberships.py
```diff
@@ -0,0 +1,46 @@
...
+        $ ./manage.py remove_expired_pending_group_memberships
+    """
+    help = 'Removes pending group memberships if they are older than 90 days.'
+
+    def add_arguments(self, parser):
+        parser.add_argument("-e", "--enterprise_custom")
+
+    def handle(self, *args, **options):
+        queryset = EnterpriseGroupMembership.objects.all()
+        if enterprise_arg := options.get("enterprise_customer"):
```
marlonkeating, line 32: Are we documenting the optional `enterprise_customer` argument somewhere?

enterprise/management/commands/remove_expired_pending_group_memberships.py
```diff
@@ -0,0 +1,46 @@
...
+    Management command for ensuring any pending group membership, ie memberships associated with a pending enterprise
+    user, are removed after 90 days.
+
+    Example usage:
+        $ ./manage.py remove_expired_pending_group_memberships
+    """
+    help = 'Removes pending group memberships if they are older than 90 days.'
+
+    def add_arguments(self, parser):
+        parser.add_argument("-e", "--enterprise_custom")
```
marlonkeating, line 28: typo?
```suggestion
        parser.add_argument("-e", "--enterprise_customer")
```

alex-sheehan-edx: 😂 YIKES good catch haha

enterprise/management/commands/remove_expired_pending_group_memberships.py
alex-sheehan-edx: uhhh Let me add it to the doc string, it's not necessary for the overnight job that's to run however I liked the added flexibility for other users down the line

alex-sheehan-edx: updated the doc string- I'll write a quick unit test for it just to be certain

marlonkeating (approved):

PR #2043, opened 2024-03-05
ENT-8540 - feat: disable failed content transmissions for 24hrs

hamzawaleed01:
**Merge checklist:**

**Post merge:**


integrated_channels/integrated_channel/exporters/content_metadata.py
```diff
@@ -201,6 +203,8 @@ def _check_matched_content_updated_at(
                     remote_deleted_at__isnull=True,
                     remote_created_at__isnull=False,
                 )
+                content_query.add(Q(remote_errored_at__lt=self.LAST_24_HRS) | Q(
+                    remote_errored_at__isnull=True), Q.AND)
```
sameenfatima78, line 207: shouldn't the query be `remote_errored_at__gt=self.LAST_24_HRS` or am I missing something here? 🤔

integrated_channels/integrated_channel/models.py
```diff
@@ -671,6 +673,11 @@ class Meta:
         blank=True,
         null=True
     )
+    remote_errored_at = models.DateTimeField(
+        help_text='Date when the content transmission was failed in the remote API.',
+        blank=True,
+        null=True
+    )
```
sameenfatima78, line 680: how are we ensuring that this field will always be populated? and what happens if we get a null. Plus aren't we require to backfill the value for this field for previously failed transmissions?

integrated_channels/integrated_channel/exporters/content_metadata.py
hamzawaleed01: Its basically excluding records which errored in last 24 hrs. That's why `errored_at < last 24hrs` 🤔 

Or other way to read this is `errored_at < one_day_ago`

integrated_channels/integrated_channel/models.py
hamzawaleed01: I also thought first about backfilling but later I added this condition in the query
https://github.com/openedx/edx-enterprise/pull/2043/files/c6dedfbf9de37eccfdd2a391e77c303feb4dd531#diff-bd1856bd7ffab338eb216ef940de368327f49152f8673cc007eac15557220c78R207

Now basically we're selecting those records which errored 24hrs ago OR whose errored_at is `Null`

sameenfatima78: Hmm, so this will work. But shouldn't we start populating it for future enrollments?

hamzawaleed01: It will populate for future enrollments - whenever an error occurs
https://github.com/openedx/edx-enterprise/pull/2043/files#diff-817b72970a2242a621a435d66ad70f35a223261ef3574f97a0ef5abcc1916530R254

zamanafzal (approved):

sameenfatima78 (approved):

PR #2044, opened 2024-03-06
feat: admin pages for enterprise groups and enterprise group memberships

alex-sheehan-edx:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/admin/__init__.py
+++ b/enterprise/admin/__init__.py
@@ -456,7 +456,7 @@ class EnterpriseCustomerUserAdmin(admin.ModelAdmin):
     )
 
     list_display = ('username', 'user_email', 'get_enterprise_customer')
-    search_fields = ('user_id',)
+    search_fields = ('user_id', 'user_email',)
 
     @admin.display(
         description='Enterprise Customer'
@@ -581,6 +581,11 @@ class PendingEnterpriseCustomerUserAdmin(admin.ModelAdmin):
         'user_email',
         'enterprise_customer',
         'created'
+    )
+
+    search_fields = (
+        'user_email',
+        'id'
     )
 
     readonly_fields = (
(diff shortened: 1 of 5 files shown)


enterprise/admin/__init__.py
```diff
@@ -1155,3 +1160,60 @@ def mark_configured(self, request, obj):
...
+        Return the non-deleted members of a group
+        """
+        return obj.members.filter(is_removed=False)
+
+    @admin.display(description="Number of members in group")
+    def count(self, obj):
+        """
+        Return the number of members in a group
+        """
+        return len(obj.members.filter(is_removed=False))
```
marlonkeating, line 1192: nit
```suggestion
        return len(self.members(obj))
```

enterprise/models.py
```diff
@@ -4287,3 +4287,24 @@ class Meta:
         # ie no issue if multiple fields have: group = A and pending_enterprise_customer_user = NULL
         unique_together = (("group", "enterprise_customer_user"), ("group", "pending_enterprise_customer_user"))
         ordering = ['-modified']
+
+    def clean(self, *args, **kwargs):
+        """
+        Ensure that records added via Django Admin have matching customer records between learner and group.
+        """
+        user = self.enterprise_customer_user or self.pending_enterprise_customer_user
```
marlonkeating, line 4295: nit: Duplicate line [here](https://github.com/openedx/edx-enterprise/pull/2044/files#diff-d950991245f878f58adc157895d6431af4385d522ab994aa25726e9abc2b434dR4309), maybe make into a helper property?

marlonkeating (approved):

enterprise/models.py
alex-sheehan-edx: yeah I think we can pull this out, good catch

PR #2049, opened 2024-03-15
feat: send group membership invite, remove, and reminder emails

katrinan029:
**Merge checklist:**

**Post merge:**


enterprise/api_client/braze.py
```diff
@@ -9,26 +9,87 @@
 
 logger = logging.getLogger(__name__)
 
+ENTERPRISE_BRAZE_ALIAS_LABEL = 'Enterprise'  # Do Not change this, this is consistent with other uses across edX repos.
 
-class BrazeAPIClient:
+class BrazeAPIClient(BrazeClient):
```
alex-sheehan-edx, line 14: Love the updates to the client!

enterprise/management/commands/send_groups_reminder_email.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        Command's entry point.
+        """
+        dry_run = options['dry_run']
+        days_before_purge_date = options['days_before_purge_date']
+
+        # get all group memberships
+        for group_membership in EnterpriseGroupMembership.objects.all():
+            # loop through pending customer list in each group membership
+            for pending_enterprise_customer_user in group_membership.pending_enterprise_customer_users:
```
alex-sheehan-edx, line 58: a group membership doesn't have a list of pending users, it either has a single pending user or a realized user. You can actually do a filter for only memberships that apply- something like
```
EnterpriseGroupMembership.objects.filter(enterprise_customer_user__isnull=True)
```
That way you're only grabbing memberships that have not been realized. The full line would look something like 
```
for pending_user in EnterpriseGroupMembership.objects.filter(enterprise_customer_user__isnull=True).values_list('pending_enterprise_customer_user', flat=True):
```

alex-sheehan-edx: FWIW I _**also**_ think you could do the filtering for your created_at times here as well! You can create a query `Q` object that ors the operations of checking the `created_at` value on the table.

alex-sheehan-edx: LMK if you wanna chat about it!

alex-sheehan-edx (commented):
Generally, the right direction. I think we need to think about which membership records we're pulling from the DB and we can simplify down some of the resulting logic given we have access to the pending learner record

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -118,9 +119,15 @@ def get_learners(self, *args, **kwargs):
...
         group_uuid = kwargs.get('group_uuid')
+        # so our endpoint would look something like this
+        # GET api/v1/enterprise-group/<group_uuid>/learners/?filter=pending_learner_id
+        filter_for_pending_memberships = kwargs.get('pending_learner_id')
+
         try:
             group_object = self.get_queryset().get(uuid=group_uuid)
             members = group_object.get_all_learners()
+            if filter_for_pending_memberships is not None:
+                members = members.filter(learner_uuid=None)
```
alex-sheehan-edx, line 130: For performance reasons, I would probably pass a bool into `get_all_learners` and update the DB query

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -237,6 +249,9 @@ def assign_learners(self, request, group_uuid):
                 'new_learners': total_new_users_processed,
                 'existing_learners': total_existing_users_processed,
             }
+            for membership in memberships:
+                    if budget_expiration is not None and catalog_uuid is not None:
+                        send_group_membership_invitation_notification.delay(group.enterprise_customer, membership, budget_expiration, catalog_uuid)
```
alex-sheehan-edx, line 254: why are we sending the invitation notification twice in assign learners?

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -273,6 +288,8 @@ def remove_learners(self, request, group_uuid):
                 )
                 records_deleted += len(records_to_delete)
                 records_to_delete.delete()
+                for record in records_to_delete:
```
alex-sheehan-edx, line 291: Do these records still exist in memory after they are deleted?

enterprise/api/v1/serializers.py
```diff
@@ -590,6 +590,8 @@ class EnterpriseGroupSerializer(serializers.ModelSerializer):
     class Meta:
         model = models.EnterpriseGroup
         fields = ('enterprise_customer', 'name', 'uuid')
+    
+    enterprise_customer = EnterpriseCustomerSerializer()
```
alex-sheehan-edx, line 594: I think we should probably have a test to ensure this is giving us the values we want

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -200,6 +208,10 @@ def assign_learners(self, request, group_uuid):
                     )
                     for ecu in ecus
                 ]
+                # i think we still need to send invitation emails to 'realized' customers as well 
+                for ent_customer_user in ent_customer_users:
+                    if budget_expiration is not None and catalog_uuid is not None:
```
alex-sheehan-edx, line 213: I would put this outside of the for loop because it applies to all learners in this case and we'd only have to check once

alex-sheehan-edx (commented):
Looks great so far- one comment on filtering for only pending learners in group and unit tests but otherwise on the right track 👍

enterprise/api/v1/views/enterprise_group.py
katrinan029: good catch. I think we should keep this loop here since it applies to all memberships and remove the `send_group_membership_invitation_notification` function up above where we are looping over `for ent_customer_user in ent_customer_users` since that one is only sending to existing learners.

enterprise/api/v1/views/enterprise_group.py
katrinan029: removing these lines since we are sending the emails below to all memberships

enterprise/api/v1/views/enterprise_group.py
katrinan029: moving this line up!

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -158,6 +162,8 @@ def assign_learners(self, request, group_uuid):
         except models.EnterpriseGroup.DoesNotExist as exc:
             raise Http404 from exc
         if requested_emails := request.POST.dict().get('learner_emails'):
+            budget_expiration = request.POST.dict().get('budget_expiration')
+            catalog_uuid = request.POST.dict().get('catalog_uuid')
```
iloveagent57, line 166: fyi, you can use `request.data.get('whatever payload key)` here: https://www.django-rest-framework.org/api-guide/requests/#data

enterprise/tasks.py
```diff
@@ -182,3 +207,133 @@ def send_sso_configured_email(
...
+
+
+@shared_task
+@set_code_owner_attribute
+def send_group_membership_invitation_notification(
+    enterprise_customer_uuid,
+    group_membership_uuid,
+    budget_expiration,
+    catalog_uuid
+):
```
iloveagent57, line 219: One way you could make this task much more efficient is to have it process a list of of membership uuids, rather than a single membership per task invocation.  That will help cut down a lot on the number of calls to the catalog service, and more importantly, on the number of calls to Braze.  The Braze client methods to create aliases and send messages already accept lists of recipients as input, so it should be pretty easy to adjust the logic you already have to work on a list of members.

enterprise/tasks.py
```diff
@@ -182,3 +207,133 @@ def send_sso_configured_email(
...
+    Arguments:
+        * enterprise_customer_uuid (string)
+        * group_membership_uuid (string)
+        * budget_expiration (datetime)
+        * catalog_uuid (string)
+    """
+    enterprise_customer = get_enterprise_customer(enterprise_customer_uuid)
+    group_membership = enterprise_group_membership_model().objects.filter(
+        uuid=group_membership_uuid
+    ).values().last()
```
iloveagent57, line 232: you can access the related `EnterpriseCustomerUser` record directly from the group membership record, b/c the membership record has a FK to ECU.
Also, if you use `select_related('enterprise_customer_user')`, the related model will be fetched with still only a single query against the DB.

iloveagent57: ...same idea for the Pending ECU record.

enterprise/tasks.py
```diff
@@ -182,3 +207,133 @@ def send_sso_configured_email(
...
+    Send braze email notification when learner is removed from a group.
+
+    Arguments:
+        * enterprise_customer_uuid (string)
+        * group_membership_uuid (string)
+    """
+    enterprise_customer = get_enterprise_customer(enterprise_customer_uuid)
+    group_membership = enterprise_group_membership_model().objects.filter(
+        uuid=group_membership_uuid
+    ).values().last()
```
iloveagent57, line 294: ...and same idea here about using related records via the `ForiegnKey` relationship.

enterprise/tasks.py
```diff
@@ -182,3 +207,133 @@ def send_sso_configured_email(
...
         LOGGER.exception(message)
         raise exc
+
+
+@shared_task
+@set_code_owner_attribute
+def send_group_membership_invitation_notification(
+    enterprise_customer_uuid,
+    group_membership_uuid,
+    budget_expiration,
```
iloveagent57, line 217: I would genericize the name of this parameter a little more. The edx-enterprise code has no notion or definition of budgets or policies.  Maybe something like `act_by_date`?

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -239,6 +244,14 @@ def assign_learners(self, request, group_uuid):
                 'new_learners': total_new_users_processed,
                 'existing_learners': total_existing_users_processed,
             }
+            if budget_expiration is not None and catalog_uuid is not None:
+                for membership in memberships:
```
iloveagent57, line 248: you could use `utils.batch()` here to break these memberships up into lists to send off to your async task.

tests/test_enterprise/api/test_views.py
```diff
@@ -7624,6 +7680,7 @@ def test_successful_assign_learners_to_group(self):
                 enterprise_customer_user__isnull=True
             )
         ) == 10
+        assert mock_send_group_membership_invitation_notification.call_count == 20
```
iloveagent57, line 7683: nit: it'd be good to make some assertion about what arguments this task was called with.

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -112,16 +113,19 @@ def get_learners(self, *args, **kwargs):
...
+                        'enterprise_customer': EnterpriseCustomerSerializer
                     },
                 ],
             }
 
         """
 
         group_uuid = kwargs.get('group_uuid')
+        # GET api/v1/enterprise-group/<group_uuid>/learners?filter_for_pecus=true
+        filter_for_pecus = self.request.query_params.get('filter_for_pecus', None)
```
iloveagent57, line 125: You can change this method signature to have a `request` param, so instead of `def get_learners(self, *args, **kwargs)`, make it `def get_learners(self, request, *args, **kwargs)`.  This you can do `request.query_params.get(...)` on this line.

iloveagent57, line 125: Also, it'd be good to add something to the docstring for this method about `filter_for_pecus` so that it's knowable from our API docs.

iloveagent57: Naming nit for this query param: maybe `pending_users_only` to avoid any ambiguity around the word "filter"?

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -157,7 +161,9 @@ def assign_learners(self, request, group_uuid):
             customer = group.enterprise_customer
         except models.EnterpriseGroup.DoesNotExist as exc:
             raise Http404 from exc
-        if requested_emails := request.POST.dict().get('learner_emails'):
+        if requested_emails := request.data.get('learner_emails'):
+            act_by_date = request.data.get('act_by_date')
+            catalog_uuid = request.data.get('catalog_uuid')
```
iloveagent57, line 166: We should do some input validation on these optional query params, you could use a DRF serializer to achieve this. For instance, you'd want `learner_emails` to be a list of strings that all matched some email regex, which DRF serializers make easy by using `EmailField()`.

iloveagent57: ...and also, `learner_emails` is required for _anything_ to get done by this view.

tests/test_enterprise/api/test_views.py
```diff
@@ -7624,6 +7681,15 @@ def test_successful_assign_learners_to_group(self):
                 enterprise_customer_user__isnull=True
             )
         ) == 10
+        assert mock_send_group_membership_invitation_notification.call_count == 1
+        group_uuids = list(EnterpriseGroupMembership.objects.filter(group=self.group_2).values_list('uuid', flat=True))
+        mock_send_group_membership_invitation_notification.assert_has_calls([
+            mock.call(self.enterprise_customer.uuid,
+                      list(reversed(group_uuids)),
```
alex-sheehan-edx, line 7688: small nit but could we clean up the formatting on this?

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -157,18 +172,34 @@ def assign_learners(self, request, group_uuid):
             customer = group.enterprise_customer
         except models.EnterpriseGroup.DoesNotExist as exc:
             raise Http404 from exc
-        if requested_emails := request.POST.dict().get('learner_emails'):
+
+        if requested_emails := request.data.get('learner_emails'):
```
alex-sheehan-edx, line 176: we need learner emails here, we probably want to return an error 400 if they aren't a part of the request data

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -157,18 +172,34 @@ def assign_learners(self, request, group_uuid):
...
+            request_data = {}
+            if act_by_date := request.data.get('act_by_date'):
```
alex-sheehan-edx, line 178: do we need a conditional here? Could we just set the `request_data` to the result of `request.data.get('act_by_date')`

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -109,19 +116,27 @@ def get_learners(self, *args, **kwargs):
                 'previous': null,
                 'results': [
                     {
-                        'learner_uuid': 'enterprise_customer_user_id',
-                        'pending_learner_id': 'pending_enterprise_customer_user_id',
-                        'enterprise_group_membership_uuid': 'enterprise_group_membership_uuid',
+                        'learner_id': integer or None,
+                        'pending_learner_id': integer or None,
+                        'enterprise_group_membership_uuid': UUID,
+                        'enterprise_customer': EnterpriseCustomerSerializer,
```
alex-sheehan-edx, line 122: this would be an object of enterprise customer details

enterprise/tasks.py
```diff
@@ -182,3 +208,130 @@ def send_sso_configured_email(
         )
         LOGGER.exception(message)
         raise exc
+
+
+@shared_task
+@set_code_owner_attribute
+def send_group_membership_invitation_notification(
+    enterprise_customer_uuid,
```
alex-sheehan-edx, line 216: instead of the uuid, can we just accept the customer record? Since we'll have that in the view

enterprise/api/v1/serializers.py
```diff
@@ -599,6 +599,7 @@ class EnterpriseGroupMembershipSerializer(serializers.ModelSerializer):
     learner_id = serializers.IntegerField(source='enterprise_customer_user.id', allow_null=True)
     pending_learner_id = serializers.IntegerField(source='pending_enterprise_customer_user.id', allow_null=True)
     enterprise_group_membership_uuid = serializers.UUIDField(source='uuid', allow_null=True, read_only=True)
+    enterprise_customer = EnterpriseCustomerSerializer(source='group.enterprise_customer', read_only=True)
```
alex-sheehan-edx, line 602: Discussed offline with @katrinan029 and I think we don't need the customer details in the membership serializer. Too much duplicated data and the information is easily fetched by the customer endpoint

enterprise/tasks.py
```diff
@@ -182,3 +208,130 @@ def send_sso_configured_email(
...
+            "Groups learner invitation email could not be sent "
+            f"to {recipients} for enterprise {enterprise_customer_name}."
+        )
+        LOGGER.exception(message)
+        raise exc
+
+
+@shared_task
+@set_code_owner_attribute
+def send_group_membership_removal_notification(enterprise_customer_uuid, membership_uuids):
```
alex-sheehan-edx, line 284: same for here and using the customer instead of the customer uuid as a param

alex-sheehan-edx (commented):
Some thoughts!

enterprise/tasks.py
iloveagent57: Celery tasks generally don't deal well with accepting model instances as arguments - the general pattern is to pass in object identifiers, and then fetch those within the task.  This is because task name, options, and input has to be serialized (with either json or pickle) before its sent to the broker and then to the worker.

kiram15 (approved):
Looks great!

enterprise/tasks.py
alex-sheehan-edx: ignore this, I agree with @iloveagent57!

enterprise/api/v1/views/enterprise_group.py
katrinan029: removing this from serializer

enterprise/api/v1/views/enterprise_group.py
katrinan029: the 400 error is returned in the last line of this method
`return Response(data="Error: missing request data: `learner_emails`.", status=400)`

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -157,18 +172,34 @@ def assign_learners(self, request, group_uuid):
...
             raise Http404 from exc
-        if requested_emails := request.POST.dict().get('learner_emails'):
+
+        if requested_emails := request.data.get('learner_emails'):
+            request_data = {}
+            if act_by_date := request.data.get('act_by_date'):
+                request_data['act_by_date'] = act_by_date
+            if catalog_uuid := request.data.get('catalog_uuid'):
+                request_data['catalog_uuid'] = catalog_uuid
+            request_data['learner_emails'] = requested_emails.split(',')
```
iloveagent57, line 182: I have the same question as Kira - why can't we pass `request.data` directly into this serializer for validation?  This is a POST endpoint, so the caller can/should provide some JSON data as the request body, which means they can provide a list instead of a comma-delimited string.

katrinan029: Thanks for this callout. I updated this to `request.data.getlist` for the emails since `request.data.get` only returns the last item in the list per the specs.

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -157,18 +171,34 @@ def assign_learners(self, request, group_uuid):
             customer = group.enterprise_customer
         except models.EnterpriseGroup.DoesNotExist as exc:
             raise Http404 from exc
-        if requested_emails := request.POST.dict().get('learner_emails'):
+        import pdb
+        pdb.set_trace()
+        if requested_emails := request.POST.getlist('learner_emails'):
```
katrinan029, line 176: using `request.POST.getlist` here because default return value is `None` if the key doesn't exist. If we use `request.data.getlist`, we get an error `AttributeError: 'dict' object has no attribute 'getlist'`

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -157,18 +171,32 @@ def assign_learners(self, request, group_uuid):
             customer = group.enterprise_customer
         except models.EnterpriseGroup.DoesNotExist as exc:
             raise Http404 from exc
-        if requested_emails := request.POST.dict().get('learner_emails'):
+        if requested_emails := request.POST.getlist('learner_emails'):
+            request_data = {}
+            request_data['act_by_date'] = request.data.get('act_by_date')
+            request_data['catalog_uuid'] = request.data.get('catalog_uuid')
+            request_data['learner_emails'] = requested_emails
```
iloveagent57, line 178: You shouldn't need to do any of this - you should be able to pass `request.data` directly into the serializer like so:
```python
param_serializer = serializers.EnterpriseGroupAssignLearnersRequestDataSerializer(request.data)
param_serializer.is_valid(raise_exception=True) # will raise a validation error and cause a 400 response automatically
act_by_date = param_serializer.data['act_by_date']
# ...and so on
```
This is a pattern we've been using with success on the enterprise-access service, e.g. https://github.com/openedx/enterprise-access/blob/main/enterprise_access/apps/api/v1/views/subsidy_access_policy.py#L808

enterprise/tasks.py
```diff
@@ -182,3 +207,130 @@ def send_sso_configured_email(
...
+        'catalog_content_count'
+    ] = enterprise_catalog_client.get_catalog_content_count(catalog_uuid)
+
+    braze_trigger_properties['act_by_date'] = act_by_date.strftime('%B %d, %Y')
+    pecu_emails = []
+    ecus = []
+    for membership_uuid in membership_uuids:
+        group_membership = enterprise_group_membership_model().objects.filter(
+            uuid=membership_uuid
+        )
```
iloveagent57, line 247: This is going to perform one query per memberhsip, but you could actually do this in a single query like so:
```python
membership_records = enterprise_group_membership_model().objects.filter(uuid__in=membership_uuids)
for group_membership in membership_records:
    if group_membership.pending_enterprise_customer_user is not None:
        pecu_emails.append(group_membership.pending_enterprise_customer_user.user_email)
    else:
        ecus.append({
            'user_email': group_membership.enterprise_customer_user.user_email,
            'user_id': group_membership.enterprise_customer_user.user_id
        })
```

enterprise/tasks.py
```diff
@@ -182,3 +207,130 @@ def send_sso_configured_email(
...
+    contact_email = enterprise_customer.contact_email
+    enterprise_customer_name = enterprise_customer.name
+    braze_trigger_properties['contact_admin_link'] = braze_client_instance.generate_mailto_link(contact_email)
+    braze_trigger_properties['enterprise_customer_name'] = enterprise_customer_name
+    pecu_emails = []
+    ecus = []
+    for membership_uuid in membership_uuids:
+        group_membership = enterprise_group_membership_model().objects.filter(
+            uuid=membership_uuid
+        )
```
iloveagent57, line 303: same pattern as above would work here, too.

iloveagent57 (commented):
This is looking good, couple small suggestions for simplicity and performance improvements.

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -176,91 +194,101 @@ def assign_learners(self, request, group_uuid):
...
-                        enterprise_customer=customer, user_id=user.id, active=True
-                    ) for user in existing_users
-                }
-                models.EnterpriseCustomerUser.objects.bulk_create(
-                    ecu_by_email.values(),
-                    ignore_conflicts=True,
-                )
+        param_serializer = serializers.EnterpriseGroupRequestDataSerializer(data=request.data)
+        param_serializer.is_valid(raise_exception=True)
+        act_by_date = param_serializer.data['act_by_date']
```
kiram15, line 199: [nit] Might do something like this for these 
```act_by_date, catalog_uuid, learner_emails = param_serializer.data['act_by_date'], param_serializer.data['catalog_uuid'], param_serializer.data['learner_emails']```
```total_records_processed, total_existing_users_processed, total_new_users_processed = 0, 0, 0```

File: enterprise/api/v1/views/enterprise_group.py [hunk 30]
```diff
@@ -176,91 +194,101 @@ def assign_learners(self, request, group_uuid):
...
+        total_records_processed = 0
+        total_existing_users_processed = 0
+        total_new_users_processed = 0
+        for user_email_batch in utils.batch(learner_emails[: 1000], batch_size=200):
+            user_emails_to_create = []
+            memberships_to_create = []
+            # ecus: enterprise customer users
+            ecus = []
+            # Gather all existing User objects associated with the email batch
+            existing_users = User.objects.filter(email__in=user_email_batch)
```
kiram15, line 211: This seems like a red flag to me based on all the problems Alex has with filtering and querying the User table and it taking so long. I would look into this PR and see what he's doing with the SQL string https://github.com/openedx/edx-enterprise/pull/2057/files#diff-d950991245f878f58adc157895d6431af4385d522ab994aa25726e9abc2b434dR4274

kiram15: Though he was doing `email__icontains` and not `email__in` which would be faster but maybe it should be something more like the [_get_filtered_ecu_ids](https://github.com/openedx/edx-enterprise/blob/aa09eea59ce9405332208b2e8ba2888038fddea0/enterprise/models.py#L4274)

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -279,31 +307,44 @@ def remove_learners(self, request, group_uuid):
             - ``records_deleted``:
                 Number of membership records removed
         """
+        # import pdb
```
kiram15, line 310: crust

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -279,31 +307,44 @@ def remove_learners(self, request, group_uuid):
...
+            )
+            records_deleted += len(records_to_delete)
+            records_to_delete_uuids = [record.uuid for record in records_to_delete]
+            for records_to_delete_uuids_batch in utils.batch(records_to_delete_uuids, batch_size=200):
+                send_group_membership_removal_notification.delay(
+                    customer.uuid,
+                    records_to_delete_uuids_batch,
+                    catalog_uuid)
+            records_to_delete.delete()
+            # Woohoo! Records removed! Now to update the soft deleted records
```
kiram15, line 339: 🥳

tests/test_enterprise/test_tasks.py
```diff
@@ -180,3 +194,117 @@ def test_send_enterprise_email_notification(self, mock_send_notification, mock_e
...
+            trigger_properties={
+                'contact_admin_link': mock_admin_mailto,
+                'enterprise_customer_name': self.enterprise_customer.name,
+                'catalog_content_count': mock_catalog_content_count,
+            },
+        )]
+        mock_braze_api_client().send_campaign_message.assert_has_calls(calls)
+
+    @mock.patch('enterprise.tasks.BrazeAPIClient', return_value=mock.MagicMock())
+    def test_sso_configuration_oauth_orchestration_email(self, mock_braze_client):
```
kiram15, line 292: Should we have a test for if something goes wrong?

katrinan029: I moved this test over to the more appropriate testing file but it is unrelated to this work. the sso changes should be in a separate pr for the fix we discussed.

enterprise/api/v1/views/enterprise_group.py
katrinan029: i think for readability I will keep the variable assignments in separate lines.

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -176,91 +194,101 @@ def assign_learners(self, request, group_uuid):
...
-            # Create all our memberships, bulk_create will batch for us.
-            memberships = models.EnterpriseGroupMembership.objects.bulk_create(
-                memberships_to_create, ignore_conflicts=True
+                for ecu in ecus
+            ]
+            total_existing_users_processed += len(ent_customer_users)
+            memberships_to_create.extend(ent_customer_users)
+
+        # Go over (in batches) all emails that don't have User objects
+        for emails_to_create_batch in utils.batch(user_emails_to_create, batch_size=200):
```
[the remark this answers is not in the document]
katrinan029: this is following the batch size consistent with the previous work.

enterprise/models.py
```diff
@@ -4369,6 +4376,8 @@ def get_all_learners(self, user_query=None, sort_by=None, desc_order=False, fetc
                 'recent_action': lambda t: t.recent_action,
             }
             members = sorted(members, key=lambda_keys.get(sort_by), reverse=desc_order)
+        if pending_users_only:
```
alex-sheehan-edx, line 4379: I don't think this can be tact onto the end of the other work being done - say you have an implicitly defined group and pass the `pending_users_only` bool, you would theoretically receive 0 records. Instead I think `pending_users_only` should be passed into both `_get_explicit_group_members` and `_get_implicit_group_members` above and handled by the methods. For the implicit fetch, all you have to do is skip the steps [here](https://github.com/openedx/edx-enterprise/pull/2049/files#diff-d950991245f878f58adc157895d6431af4385d522ab994aa25726e9abc2b434dL4307-L4317) that involve the customer_users if the `pending_users_only` value is True

alex-sheehan-edx (commented):
Some thoughts - let me know if you want to discuss

enterprise/models.py
alex-sheehan-edx: (and instead have `customer_users` be an empty array (`[]`)

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -176,91 +194,101 @@ def assign_learners(self, request, group_uuid):
...
+            memberships_to_create, ignore_conflicts=True
+        )
+        total_records_processed += len(memberships)
+        data = {
+            'records_processed': total_records_processed,
+            'new_learners': total_new_users_processed,
+            'existing_learners': total_existing_users_processed,
+        }
+        membership_uuids = [membership.uuid for membership in memberships]
+        if act_by_date is not None and catalog_uuid is not None:
```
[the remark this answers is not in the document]
katrinan029: agreed with the syntax suggestion! I think we want both values here because the email template requires an expiration date and course count.

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -279,31 +307,44 @@ def remove_learners(self, request, group_uuid):
...
+            existing_users = User.objects.filter(email__in=user_email_batch).values_list("id", flat=True)
+            group_q = Q(group=group)
+            ecu_in_q = Q(enterprise_customer_user__user_id__in=existing_users)
+            pecu_in_q = Q(pending_enterprise_customer_user__user_email__in=user_email_batch)
+            records_to_delete = models.EnterpriseGroupMembership.objects.filter(
+                group_q & (ecu_in_q | pecu_in_q),
+            )
+            records_deleted += len(records_to_delete)
+            records_to_delete_uuids = [record.uuid for record in records_to_delete]
+            for records_to_delete_uuids_batch in utils.batch(records_to_delete_uuids, batch_size=200):
```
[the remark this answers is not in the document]
katrinan029: yup! moving this logic down

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -275,35 +306,50 @@ def remove_learners(self, request, group_uuid):
...
+            records_to_delete_uuids = [record.uuid for record in records_to_delete]
+            records_to_delete.delete()
+            for records_to_delete_uuids_batch in utils.batch(records_to_delete_uuids, batch_size=200):
+                send_group_membership_removal_notification.delay(
+                    customer.uuid,
+                    records_to_delete_uuids_batch,
+                    catalog_uuid)
+            # Woohoo! Records removed! Now to update the soft deleted records
+            deleted_records = models.EnterpriseGroupMembership.all_objects.filter(
+                group_q & (ecu_in_q | pecu_in_q),
```
alex-sheehan-edx, line 346: [Request] Since you have `records_to_delete_uuids` above, could we swap our this query for 
```
uuid__in=records_to_delete_uuids
```

alex-sheehan-edx (approved):
Just a few small nits/changes but otherwise looks good to me!

PR #2057, opened 2024-03-28
feat: allowing for sorting and filtering of the enterprise group learner endpoints

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-8503
**Merge checklist:**

**Post merge:**


tests/test_enterprise/api/test_views.py
```diff
@@ -7400,6 +7401,101 @@ def test_list_learner_statuses(self):
...
+        # Test sorting by the three sortable values
+        for sort_by_value in ['recentAction', 'memberStatus', 'memberDetails']:
+            url = settings.TEST_SERVER + reverse(
+                'enterprise-group-learners',
+                kwargs={'group_uuid': self.group_1.uuid},
+            ) + f"?q={sort_by_value}"
+            response = self.client.get(url)
+            results = response.json().get('results')
+            returned_sorted_values = [
+                value.get(re.sub(r'(?<!^)(?=[A-Z])', '_', sort_by_value).lower()) for value in results
```
marlonkeating, line 7476: There's already a helper method in this repo: https://github.com/openedx/edx-enterprise/blob/0e78b47c2d873c932e46826e031c357a5d0f9e76/enterprise/management/commands/manufacture_data.py#L56

enterprise/models.py
```diff
@@ -4259,11 +4260,24 @@ class Meta:
...
+            sort_by (optional): specify how the list of returned members should be ordered. Supported sorting values
+                                are `memberDetails`, `memberStatus`, and `recentAction`. Ordering can be reversed by
+                                supplying a `-` at the beginning of the sorting value ie `-memberStatus`.
+        """
+        if q:
+            # pylint: disable=unsupported-binary-operation
+            # unfortunately, user is a property not a field on the ecu model. In order to filter ecu's by name or email
+            # means we have to filter users by name/email and then apply fetched user ID's to the ecu table
+            user_ids = User.objects.filter(
+                Q(email__contains=q) | Q(first_name__contains=q) | Q(last_name__contains=q)
```
marlonkeating, line 4279: Should we be using `__icontains` for case-insensitive searching here?

marlonkeating (changes requested):

enterprise/models.py
```diff
@@ -4344,6 +4377,37 @@ def membership_user(self):
...
+            return "accepted"
+        return "pending"
+
+    @property
+    def recent_action(self):
+        """
+        Return the timestamp of the most recent action relating to the membership
+        """
+        if self.is_removed:
+            return self.modified
```
kiram15, line 4406: Can something not be modified after it's removed?

tests/test_enterprise/api/test_views.py
```diff
@@ -7400,6 +7401,100 @@ def test_list_learner_statuses(self):
         statuses = [result.get('member_status') for result in response.json().get('results')]
         assert statuses.sort() == ['accepted', 'pending'].sort()
 
+    def test_list_learners_filtered(self):
```
kiram15, line 7404: nit: not blocking but might want to include a test with the page being passed in?

kiram15 (approved):
Looks good! Just one question and one nit

enterprise/models.py
alex-sheehan-edx: potentially manually, but not programmatically AFAIK

tests/test_enterprise/api/test_views.py
alex-sheehan-edx: I believe @katrinan029 already has a pagination test! https://github.com/openedx/edx-enterprise/blob/master/tests/test_enterprise/api/test_views.py#L7437-L7440

tests/test_enterprise/api/test_views.py
```diff
@@ -7400,6 +7401,103 @@ def test_list_learner_statuses(self):
...
+            if sort_by_value == 'memberDetails':
+                assert returned_sorted_values == sorted(returned_sorted_values, key=lambda t: t.get('user_email'))
+            else:
+                assert returned_sorted_values == sorted(returned_sorted_values)
+
+        # Test sorting in reverse order
+        url = settings.TEST_SERVER + reverse(
+            'enterprise-group-learners',
+            kwargs={'group_uuid': self.group_1.uuid},
+        ) + "?q=-memberDetails"
```
marlonkeating, line 7489:
```suggestion
        ) + "?sort_by=-memberDetails"
```

tests/test_enterprise/api/test_views.py
```diff
@@ -7400,6 +7401,103 @@ def test_list_learner_statuses(self):
...
+        """
+        # Test sorting by the three sortable values
+        for sort_by_value in ['recentAction', 'memberStatus', 'memberDetails']:
+            url = settings.TEST_SERVER + reverse(
+                'enterprise-group-learners',
+                kwargs={'group_uuid': self.group_1.uuid},
+            ) + f"?q={sort_by_value}"
+            response = self.client.get(url)
+            results = response.json().get('results')
+            returned_sorted_values = [value.get(convert_to_pascal(sort_by_value)) for value in results]
```
marlonkeating, line 7479: Does the serializer export in snake case or pascal case?

marlonkeating (changes requested):

tests/test_enterprise/api/test_views.py
alex-sheehan-edx: good catch!

tests/test_enterprise/api/test_views.py
alex-sheehan-edx:
```
    class Meta:
        model = models.EnterpriseGroupMembership
        fields = (
            'learner_id',
            'pending_learner_id',
            'enterprise_group_membership_uuid',
            'member_details',
            'recent_action',
            'member_status',
        )
```

Backend behavior is to surface data with snake case. I guess the follow up question is whether the endpoint should accept values in pascal or snake case.... My original thinking was that the frontend would want to deal in pascal and to support it but I'm ok enforcing the rule on consumers ¯\_(ツ)_/¯

alex-sheehan-edx: disregard my previous comment, I see what you mean! The test wasn't defensive enough against my craziness.

marlonkeating (approved):

iloveagent57 (changes requested):
We've talked offline about this - there's a couple of problematic things in here that shouldn't be merged, namely passing in an arbitrary query and the poor performance of that User query.  Requesting changes to block merging.

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -117,11 +124,12 @@ def get_learners(self, *args, **kwargs):
             }
 
         """
-
+        q = self.request.query_params.get('q')
```
iloveagent57, line 127: We should really do some validation of these query params before sending them off to the ORM. You could make a serializer class that you pass `self.request.query_params` into and call `is_valid()` on.

enterprise/models.py
```diff
@@ -4259,11 +4260,34 @@ class Meta:
...
+            var_q = f"%{q}%"
+            sql_string = """
+                select entcu.id from enterprise_enterprisecustomeruser entcu
+                join auth_user au on entcu.user_id = au.id
+                where entcu.enterprise_customer_id = %s and au.email like %s;
+            """
+            # Raw sql is picky about uuid format
+            customer_id = str(self.enterprise_customer.pk).replace("-", "")
+            ecus = EnterpriseCustomerUser.objects.raw(sql_string, (customer_id, var_q))
+            ecus_ids_list = [ecu.id for ecu in ecus]
```
iloveagent57, line 4290: This method is getting long and has lots of conditionals that make it hard to follow.  You could break this piece off into a helper like `def _get_filtered_ecu_ids(self, user_query=None)`.

enterprise/models.py
```diff
@@ -4259,11 +4260,34 @@ class Meta:
...
         if self.applies_to_all_contexts:
             members = []
```
iloveagent57, line 4292: Also, this whole chunk could be something like `def _get_implicit_group_members(self, user_query=None, ecu_ids=None)`.

enterprise/models.py
```diff
@@ -4285,9 +4313,24 @@ def get_all_learners(self):
...
                     group=self,
                 ))
-            return members
         else:
-            return self.members.filter(is_removed=False)
+            members = self.members.filter(is_removed=False)
+            if q:
+                ecu_filter = Q(enterprise_customer_user__id__in=ecus_ids_list)
+                pecu_filter = Q(pending_enterprise_customer_user__user_email__icontains=q)
+                members = members.filter(ecu_filter | pecu_filter)
```
iloveagent57, line 4321: You should try to check how many SQL statements this generates. I think you could add a select related to your `self.members.filter(...)` to get it down to only a single query.

enterprise/models.py
```diff
@@ -4344,6 +4387,37 @@ def membership_user(self):
...
+    @property
+    def member_status(self):
+        """
+        Return the status of the membership
+        """
+        if self.is_removed:
+            return "removed"
+        if self.enterprise_customer_user:
+            return "accepted"
+        return "pending"
```
iloveagent57, line 4408: These should be constants.  Also, this is a good indication that maybe you want to add a field called `status` to your membership model.

iloveagent57 (changes requested):

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -117,11 +124,29 @@ def get_learners(self, *args, **kwargs):
...
+            is_reversed = sort_by[0] == '-'
+            sort_by = utils.convert_to_snake(sort_by.lstrip('-'))
+            query_params['sort_by'] = sort_by
+
+        param_serializers = serializers.EnterpriseGroupLearnersRequestQuerySerializer(
+            data=query_params
+        )
+        param_serializers.is_valid()
+
+        if not param_serializers.is_valid():
```
iloveagent57, line 139: you don't need this call twice

enterprise/models.py
```diff
@@ -4259,35 +4264,107 @@ class Meta:
...
+            ))
+        return members
+
+    def _get_explicit_group_members(self, user_query=None, fetch_removed=False):
+        """
+        Fetch explicitly defined members of a group, indicated by an existing membership record
+        """
+        members = self.members.filter(is_removed=fetch_removed).select_related(
+            'enterprise_customer_user', 'pending_enterprise_customer_user'
+        )
```
iloveagent57, line 4332: Just calling this out: if `fetch_removed` is True, these members will be _only_ the members that have been removed from the group.

enterprise/models.py
```diff
@@ -4259,35 +4264,107 @@ class Meta:
...
+        if self.applies_to_all_contexts:
+            members = self._get_implicit_group_members(user_query)
         else:
-            return self.members.filter(is_removed=False)
+            members = self._get_explicit_group_members(user_query, fetch_removed)
+        if sort_by:
+            desc_order = False
+            if sort_by[0] == '-':
+                desc_order = True
+                sort_by = sort_by[1:]
```
iloveagent57, line 4360: You're already doing a bunch of computation around this `-` at the view layer, why don't you add another parameter to this function called `reversed` and have `sort_by` be only the name of the key to sort by, without the `-`?

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -117,11 +124,29 @@ def get_learners(self, *args, **kwargs):
             }
 
         """
+        query_params = self.request.query_params.copy()
+        is_reversed = False
+        if sort_by := query_params.get('sort_by'):
+            is_reversed = sort_by[0] == '-'
+            sort_by = utils.convert_to_snake(sort_by.lstrip('-'))
+            query_params['sort_by'] = sort_by
```
iloveagent57, line 132: wouldn't things be much simpler if you accepted an optional, boolean `reversed` query parameter?

tests/test_enterprise/api/test_views.py
```diff
@@ -7397,9 +7398,132 @@ def test_list_learner_statuses(self):
...
+        long_query_string = "foo".join('bar' for _ in range(320))
+        url = settings.TEST_SERVER + reverse(
+            'enterprise-group-learners',
+            kwargs={'group_uuid': self.group_1.uuid},
+        ) + f"?user_query={long_query_string}"
+        response = self.client.get(url)
+        assert response.status_code == 400
+        assert response.data.get('user_query')
+
+    def test_list_learners_filtered(self):
```
iloveagent57, line 7426: It'd be good to have an additional test in here to protect against any future changes to de-parameterize the raw sql.  Like pass in some Little Bobby Tables type of user query.
https://xkcd.com/327/

enterprise/models.py
alex-sheehan-edx: hmmm not a bad idea

tests/test_enterprise/api/test_views.py
alex-sheehan-edx: done!

iloveagent57 (approved):

PR #2081, opened 2024-04-23
fix: fix the PATCH method of ``EnterpriseCourseEnrollmentView``.

iloveagent57:
Passes in a dictionary used for computation of course run resume URLs in the `patch()` method of the `EnterpriseCourseEnrollmentView`.

ENT-8608

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,11 @@ Change Log
 
 Unreleased
 ----------
+* Nothing
+
+[4.16.2]
+---------
+* fix: fix the PATCH method of ``EnterpriseCourseEnrollmentView``.
 
 [4.16.1]
 ---------
@@ -39,7 +44,6 @@ Unreleased
 [4.15.9]
 --------
 * fix: return a 404 response for inactive CSOD customers while fetching courses
-
 
 [4.15.8]
 --------
--- a/tests/test_enterprise_learner_portal/api/test_views.py
+++ b/tests/test_enterprise_learner_portal/api/test_views.py
@@ -238,15 +238,45 @@ class TestEnterpriseCourseEnrollmentView(TestCase):
                 )
             )
 
-    @mock.patch('enterprise_learner_portal.api.v1.views.EnterpriseCourseEnrollmentSerializer')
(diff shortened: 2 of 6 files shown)


CHANGELOG.rst
```diff
@@ -15,6 +15,11 @@ Change Log
 
 Unreleased
 ----------
+
+[4.16.0]
+---------
+* fix: fix the PATCH method of ``EnterpriseCourseEnrollmentView``.
```
adamstankiewicz, line 21: super nit: if the version bump is to 4.16, should the commit prefix be `feat`? Or release it as a patch fix (4.15.13).

tests/test_enterprise_learner_portal/api/test_views.py
```diff
@@ -238,15 +238,45 @@ def test_view_requires_openedx_installation(self):
...
+        mock_get_overviews.return_value = [{
+            'id': self.course_run_id,
+            'start': 'a datetime object',
+            'end': 'a datetime object',
+            'display_name_with_default': 'a default name',
+            'pacing': 'instructor',
+            'display_org_with_default': 'my university',
+        }]
+        mock_get_resume_urls.return_value = {
+            self.course_run_id: '/courses/course-v1:MITx+6.86x+2T2024'
```
adamstankiewicz, line 269: nit: The API returns the `resume_course_run_url` as an absolute URL, versus the relative URL depicted here. Might be good to make the mocked URL absolute to have parity with the actual URL structure?

adamstankiewicz (approved):
LGTM, with 2 small nits!

tests/test_enterprise_learner_portal/api/test_views.py
iloveagent57: The serializer will actually build an absolute url out of whatever [get_resume_urls_for_course_enrollments()](https://github.com/openedx/edx-platform/blob/7cf97ec0de812659f69f565eff24f7fa342d1d49/lms/djangoapps/learner_home/views.py#L261C5-L261C43) returns. And since the above function uses `reverse()` to get the course resume url, a relative path is what we'll get before the serializer builds an absolute URL - so the mock is already set up in a way that matches what we'd see in a real request.

PR #2089, opened 2024-05-01
fix: update group invite and removal notification tasks

katrinan029:
https://2u-internal.atlassian.net/browse/ENT-8857

**Merge checklist:**

**Post merge:**


enterprise/tasks.py
```diff
@@ -255,16 +255,18 @@ def send_group_membership_invitation_notification(
...
-        recipients.append(braze_client_instance.create_recipient_no_external_id(pecu_email))
-    braze_client_instance.create_braze_alias(
-        [pecu_emails],
-        ENTERPRISE_BRAZE_ALIAS_LABEL,
-    )
-    for ecu in ecus:
-        recipients.append(braze_client_instance.create_recipient(
-            user_email=ecu['user_email'],
-            lms_user_id=ecu['user_id']))
+    if len(pecu_emails) > 0:
```
alex-sheehan-edx, line 258: NIT not necessary but you can do `if pecu_emails:` and python should equate an empty array to False

enterprise/tasks.py
```diff
@@ -255,16 +255,18 @@ def send_group_membership_invitation_notification(
...
+        for pecu_email in pecu_emails:
+            recipients.append(braze_client_instance.create_recipient_no_external_id(pecu_email))
+        braze_client_instance.create_braze_alias(
+            pecu_emails,
+            ENTERPRISE_BRAZE_ALIAS_LABEL,
+        )
+    if len(ecus) > 0:
```
alex-sheehan-edx, line 265: I don't think this check is necessary- if `ecus` is empty, it will just not iterate through the loop

marlonkeating (changes requested):
The implementation looks good as far as I can see, but could we add unit test(s) that would have caught the original issue?

alex-sheehan-edx (commented):
Small nits that apply to both sections. LMK if you wanna chat about writing tests

marlonkeating (approved):

alex-sheehan-edx (approved):

PR #2090, opened 2024-05-02
feat: updates tasks usage of create_recipient to create_recipients

brobro10000:
**Merge checklist:**

**Post merge:**


enterprise/tasks.py
```diff
@@ -244,27 +245,29 @@ def send_group_membership_invitation_notification(
...
-        recipients.append(braze_client_instance.create_recipient(
-            user_email=ecu['user_email'],
-            lms_user_id=ecu['user_id']))
+    for user_id_by_email_chunk in batch_dict(user_id_by_email, MAX_NUM_IDENTIFY_USERS_ALIASES):
+        recipients_by_email = braze_client_instance.create_recipients(
+            alias_label=ENTERPRISE_BRAZE_ALIAS_LABEL,
+            user_id_by_email=user_id_by_email_chunk
+        )
+        for _, recipient in recipients_by_email.items():
+            recipients.append(recipient)
```
iloveagent57, line 270: code-golf non-blocker: `recipients.extend(recipients_by_email.values())` will do the same thing.

enterprise/tasks.py
```diff
@@ -320,11 +322,13 @@ def send_group_membership_removal_notification(enterprise_customer_uuid, members
...
-            user_email=ecu['user_email'],
-            lms_user_id=ecu['user_id']
-        ))
+    for user_id_by_email_chunk in batch_dict(user_id_by_email, MAX_NUM_IDENTIFY_USERS_ALIASES):
+        recipients_by_email = braze_client_instance.create_recipients(
+            alias_label=ENTERPRISE_BRAZE_ALIAS_LABEL,
+            user_id_by_email=user_id_by_email_chunk
+        )
+        for _, recipient in recipients_by_email.items():
+            recipients.append(recipient)
```
iloveagent57, line 331: hmm, this is identical to the chunk above, maybe you can make a helper method and use it like so:
```python
def _recipients_for_identified_users(user_id_by_email):
    recipients = []
    for user_id_by_email_chunk in batch_dict(user_id_by_email, MAX_NUM_IDENTIFY_USERS_ALIASES):
        recipients_by_email = braze_client_instance.create_recipients(
            alias_label=ENTERPRISE_BRAZE_ALIAS_LABEL,
            user_id_by_email=user_id_by_email_chunk
        )
        recipients.extend(recipients_by_email.values())
```
...and then you can call it from within the tasks by doing
```python
recipients.extend(_recipients_for_identified_users(user_id_by_email))
```

enterprise/utils.py
```diff
@@ -2298,6 +2299,22 @@ def batch(iterable, batch_size=1):
...
+
+    Arguments:
+        dict (dict): A dictionary to chunk
+        chunk_size (int): the size of each chunk. Defaults to 1.
+    Returns:
+        generator: iterates through each chunk of a dictionary
+    """
+    it = iter(dict_data.items())
+    for _ in range(0, len(dict_data), chunk_size):
+        yield dict(islice(it, chunk_size))
```
iloveagent57, line 2315: ah, very nice! you may want to add a couple of unit tests just for this new utility function.

iloveagent57:
Looks great, couple of small suggestions! I tagged in @katrinan029 for review, too.

enterprise/tasks.py
```diff
@@ -209,6 +209,58 @@ def send_sso_configured_email(
...
+    """
+    Helper function for create_recipients that takes a dictionary of user_email keys and
+    user_id values, batches them in groups of the maximum number of users alias based
+    on the braze documentation, and destructures the individual recipient values from
+    recipients_by_email and returns a list of recipients.
+
+    Arguments:
+        * user_id_by_email (dict): A dictionary of user_email key and user_id values
+        * maximum_aliases_per_batch (int):  An integer denoting the max allowable aliases to identify
+                                            per create_recipient call to braze.
```
katrinan029, line 226: nit: `create_recipients`

katrinan029 (approved):
LGTM!

PR #2106, opened 2024-05-15
feat: Introduce disable_expiry field on the enterprise customer model

brobro10000:
Introduces `disable_expiry_messaging_for_learner_credit` on the `EnterpriseCustomer` model with default == `false`.
This field was introduced to have users opt-into disabling expiration messaging on the enterprise learner and enterprise admin portals. It would disable modals and alerts related to learner credit expiration.
Serializes the new field.
Makes the field editable on the django admin enterprise customer screen under `Recommended default settings for all enterprise customers`
Screenshot 2024-05-15 at 3 53 43 PM

Pending Tests

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -312,6 +312,13 @@ class Meta:
         on_delete=models.CASCADE
     )
 
+    disable_expiry_messaging_for_learner_credit = models.BooleanField(
+        verbose_name="Disable expiration messaging for learner credit",
+        default=False,
+        help_text=_("Automatically disabled. If unchecked, learners and admins will not receive email, "
```
adamstankiewicz, line 318: [clarification, sanity check]

>  If unchecked, learners and admins will not receive email

Should this say, "If **checked**, learners and admins will not receive email"? Only when the new config field is truthy/_enabled_ is the expiration messaging disabled? 🤔 

Similar concern for the "automatically disabled" bit. Does this mean the `disable_expiry_messaging_for_learner_credit` is disabled, with a double negative?

enterprise/models.py
```diff
@@ -312,6 +312,13 @@ class Meta:
...
+                    "and modal notifications on expiring learner credit plans.")
```
adamstankiewicz, line 319: nit: maybe remove "modal" since it's both modals and alert banners? By not including the exact UI element (e.g., modal) here, it means the expiration experience can freely change its UI elements (e.g., if the expiration experience eventually doesn't have a modal one day).

enterprise/models.py
brobro10000: First point. You're correct if checked they should not receive the following 👍🏽  good catch.

Second point, I can see how the double negative could seem confusing. Just wondering if removing this line would serve the same purpose so it becomes `If checked, learners and admins will not receive email, "
                    "and notifications on expiring learner credit plans.`

enterprise/admin/__init__.py
```diff
@@ -207,8 +207,8 @@ class EnterpriseCustomerAdmin(DjangoObjectActions, SimpleHistoryAdmin):
                        'enable_generation_of_api_credentials')
         }),
         ('Recommended default settings for all enterprise customers', {
-            'fields': ('site', 'customer_type', 'enable_learner_portal',
-                       'enable_integrated_customer_learner_portal_search',
+            'fields': ('site', 'customer_type', 'disable_expiry_messaging_for_learner_credit',
```
adamstankiewicz, line 210: nit/suggestion: given this field likely is not one that will be frequently used for most customers, I wonder if we should de-prioritize it within this grouping. For example, `enable_learner_portal`, etc. are more critical customer config settings that I feel should be displayed first in this section for usability/discoverability of the more common fields.

I might suggest perhaps grouping the new field alongside the other learner-credit related settings (see screenshot), or towards the last item in this section.:

image

adamstankiewicz (approved):
LGTM, with a couple minor nits/suggestions regarding the help text and ordering of the field in Django Admin. Also, this will need a version bump and CHANGELOG entry upon merge :)

enterprise/admin/__init__.py
brobro10000: I went ahead and moved it right below the enablement of learner credit.
Screenshot 2024-05-16 at 12 23 43 PM

PR #2109, opened 2024-05-17
feat: allowing for group members to be specified when fetched

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-8974

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -1781,7 +1781,7 @@ class EnterpriseGroupLearnersRequestQuerySerializer(serializers.Serializer):
     show_removed = serializers.BooleanField(required=False, default=False)
     is_reversed = serializers.BooleanField(required=False, default=False)
     page = serializers.IntegerField(required=False)
-    lms_users = serializers.ListField(
-        child=serializers.IntegerField(required=True),
+    learners = serializers.ListField(
+        child=serializers.EmailField(required=True),
         required=False,
     )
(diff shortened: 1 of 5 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -1735,7 +1735,7 @@ class EnterpriseGroupLearnersRequestQuerySerializer(serializers.Serializer):
     show_removed = serializers.BooleanField(required=False, default=False)
     is_reversed = serializers.BooleanField(required=False, default=False)
     page = serializers.IntegerField(required=False)
-    lms_users = serializers.ListField(
-        child=serializers.IntegerField(required=True),
+    learners = serializers.ListField(
+        child=serializers.EmailField(required=True),
```
katrinan029, line 1739: [suggestion] can we update the `get_learners` docstring in `enterprise/api/v1/views/enterprise_group.py` to include this required query param?

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -162,6 +162,13 @@ def get_learners(self, request, *args, **kwargs):
                                                     pending_users_only=pending_users_only,
                                                     fetch_removed=show_removed)
 
+            if learners := param_serializers.validated_data.get('learners'):
```
katrinan029, line 165: [question] since the `learners` field is required in the serializer, do we still need this conditional block?

enterprise/api/v1/serializers.py
katrinan029, line 1739: [nit] I think the `required` field could be set to `False`?

katrinan029 (changes requested):
Looks good! Just had a couple of clarifying questions/nit

enterprise/api/v1/views/enterprise_group.py
alex-sheehan-edx: it's not required, emails are required if you provide the `learners` param.

enterprise/api/v1/serializers.py
alex-sheehan-edx: The email fields are required for the `learners` query param, the `learners` query param is not required for a request, which should be reflected by our tests

katrinan029 (approved):

PR #2115, opened 2024-05-29
[ENT-8931] - Schema improvements in integrate-channels

hamzawaleed01:
**Merge checklist:**

**Post merge:**


sameeramin (approved):
LGTM 👍🏼

integrated_channels/integrated_channel/transmitters/learner_data.py
```diff
@@ -134,6 +134,7 @@ def single_learner_assessment_grade_transmit(self, exporter, **kwargs):
 
             learner_data.status = str(code)
             learner_data.error_message = body if code >= 400 else ''
+            learner_data.is_transmitted = code < 400
```
sameenfatima78, line 137: [Question] Have we checked how will this statement behave in case code is None or empty string?

tests/test_management.py
```diff
@@ -2161,3 +2161,26 @@ def test_remove_stale_integrated_channel_api_logs(self):
...
+        Test that the command updates is_transmitted for successful transmissions.
+        """
+        factories.CornerstoneLearnerDataTransmissionAuditFactory(
+            subdomain='edx',
+            error_message='',
+            status=200,
+            user_id=self.user.id
+        )
+        call_command('mark_learner_transmissions_transmitted_true')
+        assert 1 == CornerstoneLearnerDataTransmissionAudit.objects.filter(is_transmitted=True).count()
```
sameenfatima78, line 2186: Can we write a few more tests for other channels?

integrated_channels/integrated_channel/management/commands/mark_learner_transmissions_transmitted_true.py
```diff
@@ -0,0 +1,46 @@
+
+"""
+Unlink inactive enterprise learners of SAP Success Factors from related EnterpriseCustomer(s).
```
sameenfatima78, line 3: 😂

hamzawaleed01: 🫣

integrated_channels/integrated_channel/transmitters/learner_data.py
hamzawaleed01: It's taken from the line above it.... which works fine for us 🤷

tests/test_management.py
hamzawaleed01: Yep, added some more tests... and we've factories only for SAP and Cornerstone at the moment.

zamanafzal (approved):
LGTM

zamanafzal:
@hamzawaleed01 Please don't forget to update version and changelog.

PR #2126, opened 2024-06-05
Making blackboard unencrypted client credentials nullable 8010

MueezKhan246:
Description:
Making unencrypted client credentials nullable, so they can be removed.
[REFERENCE](https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations#EverythingAboutDatabaseMigrations-Howtodropacolumn)

JIRA:
https://2u-internal.atlassian.net/jira/software/c/projects/ENT/issues/ENT-8010


sameeramin (approved):
LGTM 👍🏼

hamzawaleed01 (approved):

integrated_channels/blackboard/models.py
```diff
@@ -146,6 +147,7 @@ def encrypted_client_id(self, value):
     client_secret = models.CharField(
         max_length=255,
         blank=True,
+        null=True,
         default='',
```
zamanafzal, line 151: Any reason we are defining default when `null=True` and `blank=True`.

sameenfatima78 (approved):

integrated_channels/blackboard/models.py
MueezKhan246: [REFERENCE](https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations#EverythingAboutDatabaseMigrations-Howtodropacolumn)

@zamanafzal because of the feedback i got here https://github.com/openedx/edx-platform/pull/34912

zamanafzal: default='' in this context seems redundant since you already have blank=True and null=True. I guess it can be removed now

PR #2131, opened 2024-06-06
Revert "Making blackboard unencrypted client credentials nullable 8010"

MueezKhan246:
Reverts openedx/edx-enterprise#2126


sameenfatima78 (approved):

sameenfatima78 (commented):
Can we add a new tag and create a new release for these reverted changes

MueezKhan246:
> Can we add a new tag and create a new release for these reverted changes

@sameenfatima78 that would be better i was going to make single release for both reversions.

PR #2135, opened 2024-06-07
fix: change in name property function 

kiram15:
Small change to change where we are pulling name from for enterprise group memberships, now we are pulling from the UserProfile object

https://2u-internal.atlassian.net/browse/ENT-9045

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -1162,7 +1162,9 @@ def name(self):
         """
         Return linked user's name.
         """
-        if self.user is not None:
+        if self.user.profile is not None:
+            return f"{self.user.profile.name}"
```
marlonkeating, line 1166: Do we have a sense of the performance implications of adding another foreign key reference here, particularly when we know we'll be populating some big lists with this data?

kiram15: My knowledge is that foreign key references really only have significant performance impact if you are doing things like inserting and deleting, and have far less if you are just accessing them.

marlonkeating: Not necessarily.  We can see that accessing `self.user` requires a call to the database [here](https://github.com/openedx/edx-enterprise/pull/2135/files#diff-d950991245f878f58adc157895d6431af4385d522ab994aa25726e9abc2b434dR1138).  If `profile` is a similar foreign key lookup, that means we're doubling the number of database hits to get each name of what could be a very long list of ecu's.

iloveagent57: You both raise good points! Referencing `self.user.profile` will cause two DB reads, one to fetch the user record, and then another to read from the `UserProfile` table.  One way to eliminate this extra read is by changing the fetch of `user` to use a `select_related('profile')`, so that we read the row of a user and its related profile in a single SQL statement (using a join).
```python
    @property
    def user(self):
        """
        Return User associated with this instance.
        Return :class:`django.contrib.auth.models.User` instance associated with this
        :class:`EnterpriseCustomerUser` instance via email.
        """
        try:
            return User.objects.get(pk=self.user_id).select_related('profile')
        except User.DoesNotExist:
            return None
```

While I'm looking at this code, another possible area for improvement is around the use of `property`. Right now, referencing `self.user` will call the `self.user()` function every time it's referenced. We could maybe switch these `property` usages to instead be Django `cached_property`, so that the values are cached for the lifetime of the object. Don't worry about that in this PR, but it's maybe a good chance for a quick/easy win.

enterprise/models.py
```diff
@@ -1142,9 +1145,14 @@ def user(self):
         """
         try:
             return User.objects.get(pk=self.user_id)
+            # return User.objects.select_related('profile').get(pk=self.user_id)
```
marlonkeating, line 1148: Deciding against this?

enterprise/models.py
```diff
@@ -1132,6 +1132,9 @@ def save(self, *args, **kwargs):
 
         return super().save(*args, **kwargs)
 
+    def get(self, *args, **kwargs):
+        return super().select_related('profile').get(*args, **kwargs)
```
iloveagent57, line 1136: Does this actually work? Because `profile` is related to an `auth.User`, not an `EnterpriseCustomerUser`.

enterprise/models.py
iloveagent57: This is closer to what we actually want, but making it work in unittests is kinda tricky, since we don't have a `UserProfile` to import into edx-enterprise (whereas with `User`, it's just a core Django model).  There's probably some way to use django model factories to do it, though.

iloveagent57 (approved):
Looks good, we can monitor for performance.

CHANGELOG.rst
```diff
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * nothing unreleased
 
+[4.20.7]
+--------
+* fix: Removed redundant check for group membership
```
marlonkeating, line 22: Does this describe the change here?

marlonkeating (approved):

PR #2145, opened 2024-06-12
fix: cancel-fulfillment endpoint is now idempotent

iloveagent57:
Additionally, this fixes a bug whereby users with the `enterprise.can_manage_enterprise_fulfillment` or
`enterprise.can_access_admin_dashboard` permissions in at least one enterprise could successfully authenticate against these endpoints for *any* enterprise.
ENT-9088

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/admin/__init__.py
+++ b/enterprise/admin/__init__.py
@@ -1217,3 +1217,36 @@ class EnterpriseGroupMembershipAdmin(admin.ModelAdmin):
         'enterprise_customer_user',
         'pending_enterprise_customer_user',
     )
+
+
+@admin.register(models.LearnerCreditEnterpriseCourseEnrollment)
+class LearnerCreditEnterpriseCourseEnrollmentAdmin(admin.ModelAdmin):
+    """
+    Django admin model for LearnerCreditEnterpriseCourseEnrollmentAdmin.
+    """
+    list_display = (
+        'uuid',
+        'fulfillment_type',
+        'enterprise_course_enrollment',
+        'is_revoked',
+        'modified',
+    )
+
+    readonly_fields = (
+        'uuid',
+        'enterprise_course_enrollment',
+    )
+
+    list_filter = ('is_revoked',)
+
+    search_fields = (
+        'uuid',
+        'enterprise_course_enrollment__id'
+        'enterprise_course_enrollment__user_id',
+    )
+
+    ordering = ('-modified',)
+
+    class Meta:
+        fields = '__all__'
+        model = models.LearnerCreditEnterpriseCourseEnrollment
--- a/enterprise/api/v1/views/enterprise_subsidy_fulfillment.py
+++ b/enterprise/api/v1/views/enterprise_subsidy_fulfillment.py
@@ -3,19 +3,20 @@ Views for the Enterprise Subsidy Fulfillment API.
 """
 
 from edx_rbac.decorators import permission_required
+from edx_rbac.mixins import PermissionRequiredMixin
 from rest_framework import status
 from rest_framework.decorators import action
 from rest_framework.exceptions import ValidationError
 from rest_framework.response import Response
-from rest_framework.status import HTTP_200_OK, HTTP_400_BAD_REQUEST, HTTP_404_NOT_FOUND, HTTP_500_INTERNAL_SERVER_ERROR
+from rest_framework.status import HTTP_200_OK, HTTP_204_NO_CONTENT, HTTP_404_NOT_FOUND, HTTP_500_INTERNAL_SERVER_ERROR
 
-from django.db.models import Q
+from django.http import Http404
 from django.shortcuts import get_object_or_404
 from django.utils.dateparse import parse_datetime
+from django.utils.functional import cached_property
 from django.utils.translation import gettext as _
 
 from enterprise import models
-from enterprise.api.utils import get_enterprise_customer_from_user_id
 from enterprise.api.v1 import serializers
 from enterprise.api.v1.views.base_views import EnterpriseWrapperApiViewSet
 from enterprise.logging import getEnterpriseLogger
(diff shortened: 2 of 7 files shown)


enterprise/admin/__init__.py
```diff
@@ -1214,3 +1214,36 @@ class EnterpriseGroupMembershipAdmin(admin.ModelAdmin):
         'enterprise_customer_user',
         'pending_enterprise_customer_user',
     )
+
+
+@admin.register(models.LearnerCreditEnterpriseCourseEnrollment)
+class LearnerCreditEnterpriseCourseEnrollmentAdmin(admin.ModelAdmin):
+    """
+    Django admin model for LearnerCreditEnterpriseCourseEnrollmentAdmin.
+    """
```
iloveagent57, line 1223: There was previously no way to observe these records (or toggle the revoked state) via the django admin site.

enterprise/api/v1/views/enterprise_subsidy_fulfillment.py
```diff
@@ -94,47 +95,62 @@ class EnterpriseSubsidyFulfillmentViewSet(EnterpriseWrapperApiViewSet):
...
+            return None
 
-        # Filters to match fulfillment enrollments' and entitlements' enterprise customer uuid to the requesting
-        # user's enterprise customer uuid.
-        subsidy_fulfillment_filter = Q(
-            enterprise_course_enrollment__enterprise_customer_user__enterprise_customer__uuid=enterprise_customer_uuid
-        )
-        subsidy_fulfillment_filter |= Q(
-            enterprise_course_entitlement__enterprise_customer_user__enterprise_customer__uuid=enterprise_customer_uuid
-        )
```
iloveagent57, line 118: This was the bad part that broken authorization - most `FulfillmentSource` records don't have a related enrollment, and this whole clause evaluated to `true`.

enterprise/api/v1/views/enterprise_subsidy_fulfillment.py
```diff
@@ -202,25 +215,19 @@ def unenrolled(self, request, *args, **kwargs):
...
         """
-        queryset = self.get_unenrolled_fulfillment_queryset()
+        queryset = self._get_unenrolled_fulfillment_queryset()
         serializer_class = self.get_unenrolled_fulfillment_serializer_class()
         serializer = serializer_class(queryset, many=True)
         return Response(serializer.data)
 
-    @permission_required(
-        'enterprise.can_access_admin_dashboard',
-        fn=lambda request, fulfillment_source_uuid: get_enterprise_customer_from_user_id(request.user.id)
```
iloveagent57, line 212: This was another part of authorization jankiness, it technically didn't "break" anything because of the old queryset filtering above, but what this permission check actually does is ask "does the user have this permission for the enterprise customer they are linked to", but it doesn't take into account _at all_ the customer associated with _the resource being requested_ (a Fulfillment Source record).

enterprise/api/v1/views/enterprise_subsidy_fulfillment.py
```diff
@@ -229,22 +236,16 @@ def retrieve(self, request, fulfillment_source_uuid, *args, **kwargs):
...
         """
         Cancel a single subsidized enrollment. Assumes fulfillment source has a valid enterprise enrollment.
-            /enterprise/api/v1/subsidy-fulfillment/{fulfillment_source_uuid}/cancel-enrollment/
+        /enterprise/api/v1/enterprise-subsidy-fulfillment/{fulfillment_source_uuid}/cancel-fulfillment/
         """
         try:
-            subsidy_fulfillment = get_object_or_404(
-                self.get_subsidy_fulfillment_queryset(), uuid=fulfillment_source_uuid
-            )
+            subsidy_fulfillment = self.requested_fulfillment_source
```
iloveagent57, line 245: the idempotency bits are in here, changing 400 -> 200, 200 -> 204

test_utils/__init__.py
```diff
@@ -274,12 +274,19 @@ class APITest(APITestCase):
     Base class for API Tests.
     """
 
+    @classmethod
+    def setUpClass(cls):
+        """
+        Perform operations common to all test classes.
+        """
+        super().setUpClass()
+        cls.user = cls.create_user(username=TEST_USERNAME, email=TEST_EMAIL, password=TEST_PASSWORD)
```
iloveagent57, line 283: Primary change here is making `create_user` side-effect free.

enterprise/api/v1/views/enterprise_subsidy_fulfillment.py
```diff
@@ -229,22 +236,16 @@ def retrieve(self, request, fulfillment_source_uuid, *args, **kwargs):
...
-    @permission_required(
-        'enterprise.can_enroll_learners',
-        fn=lambda request, fulfillment_source_uuid: get_enterprise_customer_from_user_id(request.user.id)
-    )
-    def cancel_enrollment(self, request, fulfillment_source_uuid):
+    def cancel_enrollment(self, request, fulfillment_source_uuid):  # pylint: disable=unused-argument
         """
         Cancel a single subsidized enrollment. Assumes fulfillment source has a valid enterprise enrollment.
-            /enterprise/api/v1/subsidy-fulfillment/{fulfillment_source_uuid}/cancel-enrollment/
+        /enterprise/api/v1/enterprise-subsidy-fulfillment/{fulfillment_source_uuid}/cancel-fulfillment/
```
adamstankiewicz, line 242: nit: is it worth also updating the docstring here?

image

adamstankiewicz (approved):

enterprise/api/v1/views/enterprise_subsidy_fulfillment.py
pwnage101: Why would it be common that a `LearnerCreditEnterpriseCourseEnrollment` would not have a related enrollment? Don't we typically create these records at the same time we create the enrollment too?

enterprise/api/v1/views/enterprise_subsidy_fulfillment.py
```diff
@@ -94,47 +95,62 @@ class EnterpriseSubsidyFulfillmentViewSet(EnterpriseWrapperApiViewSet):
...
-        # Get licensed enrollments under the supplied fulfillment source uuid and repeat the same process.
-        licensed_enrollments = models.LicensedEnterpriseCourseEnrollment.objects.filter(
-            uuid=fulfillment_source_uuid
-        )
-        if not self.request.user.is_staff:
-            licensed_enrollments = licensed_enrollments.filter(subsidy_fulfillment_filter)
+        # Get learner credit fulfillments under the supplied fulfillment source uuid.
+        learner_credit_enrollment = models.LearnerCreditEnterpriseCourseEnrollment.objects.filter(
+            uuid=fulfillment_source_uuid,
+        ).first()
```
pwnage101, line 142: Is this a candidate for .get() instead of .filter().first()?  The `uuid` field has a unique constraint.

enterprise/api/v1/views/enterprise_subsidy_fulfillment.py
pwnage101: I feel like typically pylint would flag this as an unused function parameter...

tests/test_enterprise/api/test_views.py
```diff
@@ -4165,7 +4183,14 @@ def test_cancel_fulfillment_belonging_to_different_enterprise(self):
...
+
+        other_enterprise_customer = factories.EnterpriseCustomerFactory(
+            name="other_test_enterprise",
+        )
+        other_enterprise_user = factories.EnterpriseCustomerUserFactory(
+            enterprise_customer=other_enterprise_customer,
+        )
         other_enrollment = factories.EnterpriseCourseEnrollmentFactory(
             enterprise_customer_user=other_enterprise_user
         )
```
pwnage101, line 4196: Don't these all have sub-factories? I feel like `other_licensed_course_enrollment = factories.LicensedEnterpriseCourseEnrollmentFactory()` alone may have been adequate.

pwnage101 (approved):
Bravo on transitioning this to proper edx-rbac, my safe space in the edx codebase.

enterprise/api/v1/views/enterprise_subsidy_fulfillment.py
iloveagent57: whoops, typo in my comment (but not the code)! most records don't have a related *entitlement*. All of them should have related enrollments.

PR #2147, opened 2024-06-13
refactor: made `career_engagement_network_message` nullable

sameeramin:
**Merge checklist:**

**Post merge:**


hamzawaleed01 (approved):

enterprise/signals.py
```diff
@@ -424,3 +424,14 @@ def generate_default_orchestration_record_display_name(sender, instance, **kwarg
 
 if COURSE_UNENROLLMENT_COMPLETED is not None:
     COURSE_UNENROLLMENT_COMPLETED.connect(enterprise_unenrollment_receiver)
+
+
+@receiver(pre_save, sender=models.EnterpriseCustomer)
```
zamanafzal, line 429: Any reason we are using pre_save and not post save?

enterprise/signals.py
```diff
@@ -424,3 +424,14 @@ def generate_default_orchestration_record_display_name(sender, instance, **kwarg
...
+def copy_data_from_career_engagement_network_message(sender, instance, **kwargs):  # pylint: disable=unused-argument
+    """
+    Copy data from `career_engagement_network_message` and `enable_career_engagement_network_on_learner_portal` to
+    `learner_portal_sidebar_content` and `enable_learner_portal_sidebar_message` respectively.
+    """
+    enable_cen = instance.enable_career_engagement_network_on_learner_portal
```
zamanafzal, line 435: Isn't this will start giving errors in the next build when we will remove `enable_career_engagement_network_on_learner_portal` field?

sameeramin: I don't think so because we will be removing all the references to the old fields including this signal in next release.

enterprise/signals.py
zamanafzal: The sugesstion was about changing `@receiver(pre_save, sender=models.EnterpriseCustomer)` to `@receiver(post_save, sender=models.EnterpriseCustomer)`. we can cover both of them in one signal I guess.

sameeramin: @zamanafzal One benefit that I noticed for the `pre_save` signal is that we don't have to explicitly call the `save()` (which will lead to max recursion error in this case) and the `update()` method. It will take care of it by itself.

zamanafzal (approved):

PR #2152, opened 2024-06-14
feat: atomically re-activate ECE when re-activating SCE

pwnage101:
Also, update ECE.unenrolled everywhere we update ECE.unenrolled_at. This field seems to have been completely overlooked. I'm not sure if we use the `unenrolled` field anywhere, but it certainly can be confusing to anyone looking at the database.

ENT-8991


iloveagent57 (commented):
Reasonable!

iloveagent57 (approved):
Also great!

PR #2159, opened 2024-06-25
feat: add failed statuses for groups

katrinan029:
Currently, the frontend lacks the data to show failed statuses from both Braze and internal API errors. This work introduces two new failed statuses, `email_error` and `internal_api_error`, and incorporates an additional field `errored_at` into the `EnterpriseGroupMembership` model. 

Additionally, a new method has been implemented to allow for patch requests to the endpoint `api/v1/enterprise-group/<group_uuid>/learners/`. This method allows updating group membership records for email errors that occur in enterprise-access. That related PR can be found here: https://github.com/openedx/enterprise-access/pull/501

https://2u-internal.atlassian.net/browse/ENT-9040

# Testing the endpoint locally
1. checkout both branches `knguyen2/ent-9040` and `https://github.com/openedx/enterprise-access/pull/501`
2. create an enterprise group at http://localhost:18000/admin/enterprise/enterprisegroup/ 
3. create a group membership http://localhost:18000/admin/enterprise/enterprisegroupmembership/
4. from postman, send a patch request with body (raw) to `http://localhost:18000/enterprise/api/v1/enterprise-group/<group_uuid>/learners/`:
`
{
    "learner": "edx@exampl.com",
    "status": "internal_api_error"
}
`
5. confirm that the status was updated at the endpoing `http://localhost:18000/enterprise/api/v1/enterprise-group/<group_uuid>/learners/`

**Merge checklist:**

**Post merge:**


enterprise/api/v1/views/enterprise_group.py
```diff
@@ -46,22 +46,32 @@ def remove_group_membership_records(group, catalog_uuid=None, user_emails=None):
...
+        deleted_records.update(
+            status=constants.GROUP_MEMBERSHIP_REMOVED_STATUS,
+            removed_at=localized_utcnow()
+        )
+        return len(deleted_records)
+    except Exception as exc:
+        failed_deleted_records = models.EnterpriseGroupMembership.all_objects.filter(
+            uuid__in=records_to_delete_uuids,
+        )
+        failed_deleted_records.update(
```
kiram15, line 70: So the way we have it now, if one errors, the whole batch all gets set to the same errored time?

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -133,6 +143,35 @@ def create(self, request, *args, **kwargs):
...
+
+        Request Arguments:
+        - ``group_uuid`` (URL location, required): The uuid of the group which learners should be updated
+
+        """
+        group = self.get_queryset().get(uuid=group_uuid)
+        data = self.request.data
+        learner = data.get("learner")
+        try:
+            pecu_in_q = Q(pending_enterprise_customer_user__user_email=learner)
```
kiram15, line 161: Why are we constraining this endpoint to only include pending learners?

enterprise/models.py
```diff
@@ -4490,6 +4496,8 @@ def recent_action(self):
         """
         Return the timestamp of the most recent action relating to the membership
         """
+        if self.errored_at:
+            return self.errored_at
         if self.is_removed:
```
kiram15, line 4501: If they are removed, don't we want to show that action before the errored at one? It seems like the more relevant one

tests/test_enterprise/test_tasks.py
```diff
@@ -267,6 +268,52 @@ def test_send_group_membership_invitation_notification(self, mock_braze_api_clie
...
+
+        with raises(BrazeClientError):
+            send_group_membership_invitation_notification(
+                self.enterprise_customer.uuid,
+                membership_uuids,
+                act_by_date,
+                catalog_uuid)
+            pending_membership.refresh_from_db()
+            assert pending_membership.status == 'email_error'
+            assert pending_membership.errored_at == localized_utcnow()
```
kiram15, line 315: Maybe also check recent action?

enterprise/api/v1/views/enterprise_group.py
katrinan029: yep. I think that this is the most efficient way because if we decide to loop through each record to delete and also send the braze email, we would be making multiple calls to braze. let me know if you have thoughts on a better approach.

enterprise/api/v1/views/enterprise_group.py
katrinan029: I updated this method more clear in our intent `update_pending_learner_status`. this endpoint only serves the reminder email updates from `enterprise-access`, which we are only sending to pending learners since existing enterprise customers automatically have the `accepted` status

enterprise/models.py
katrinan029: I think that it's possible the error could occur while we attempt to remove a learner so in that case there wouldn't be a removed status and we would want to display the error time.

tests/test_enterprise/test_tasks.py
katrinan029: yes, good call!

enterprise/api/v1/views/enterprise_group.py
```diff
@@ -46,22 +46,33 @@ def remove_group_membership_records(group, catalog_uuid=None, user_emails=None):
...
-    )
-    return len(deleted_records)
+        deleted_records.update(
+            status=constants.GROUP_MEMBERSHIP_REMOVED_STATUS,
+            removed_at=localized_utcnow()
+        )
+        return len(deleted_records)
+    except Exception as exc:
+        failed_deleted_records = models.EnterpriseGroupMembership.all_objects.filter(
+            uuid__in=records_to_delete_uuids,
```
kiram15, line 68: Should we maybe leave in a comment naming that this will error out all records even if only one failed as a opportunity for improvement in the future?

kiram15 (approved):
Looks good!

PR #2163, opened 2024-07-03
refactor: replace openai usage with xpert api

muhammad-ammar:
**Description:** Replace openai usage with xpert api.

**JIRA:** https://2u-internal.atlassian.net/browse/ENT-8979

---

**Merge checklist:**

**Post merge:**


enterprise/api_client/xpert_ai.py
```diff
@@ -0,0 +1,30 @@
...
+    Generate response using xpert api.
+    """
+    completion_endpoint = getattr(settings, 'CHAT_COMPLETION_API', None)
+    completion_endpoint_key = getattr(settings, 'CHAT_COMPLETION_API_KEY', None)
+
+    headers = {'Content-Type': 'application/json', 'x-api-key': completion_endpoint_key}
+
+    body = {'message_list': [{'role': role, 'content': prompt},]}
+
+    response = requests.post(
```
jajjibhai008, line 23: I think we should use try catch to handle exceptions or any kind of error from the xpert endpoint.

enterprise/api_client/xpert_ai.py
```diff
@@ -0,0 +1,30 @@
...
+        completion_endpoint,
+        headers=headers,
+        data=json.dumps(body),
+        timeout=(5, 20)  # connect and read timeout
```
jajjibhai008, line 27: nit: we can put these timouts variables in settings file for better readabilioty and reuseability.

enterprise/api_client/xpert_ai.py
```diff
@@ -0,0 +1,30 @@
...
+import json
+
+import requests
+
+from django.conf import settings
+
+
+def chat_completion(prompt, role):
+    """
+    Generate response using xpert api.
```
jajjibhai008, line 14: [Suggestion] You can describe arguments in docstring for better understanding and readability.
Like:
```
Arguments:
        prompt (str): {Description}
        role (str): {Description}
```

mahamakifdar19 (approved):

PR #2167, opened 2024-07-09
fix: Updates messaging on disable expiry for learner credit

brobro10000:
Updates the language of the helptext for the `disable_expiry_messaging_for_learner_credit` field on the enterprise customer model to accurately convey the behavior of the toggle.

**Merge checklist:**

**Post merge:**


CHANGELOG.rst
```diff
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * nothing unreleased
 
+[4.21.4]
+---------
+* feat: Updates messaging on disable expiry for learner credit
```
adamstankiewicz, line 22: nit: if this a `feat`, then the minor version should be bumped, not the patch version (i.e., `4.22.0` vs. `4.21.4`). Given this is updating help text, seems like it'd be more in line with `fix` (i.e., not really introducing net-new functionality).

iloveagent57 (approved):

enterprise/models.py
```diff
@@ -316,8 +316,9 @@ class Meta:
         verbose_name="Disable expiration messaging for learner credit",
         default=False,
         help_text=_("If checked, learners and admins will not receive notifications leading up to the expiration "
-                    "date for learner credit plans. Notifications signaling the expiration (and loss of usability) "
-                    "itself will still appear.")
+                    "date for learner credit plans. This includes removing the expiration date from the subsidy "
+                    "box for upcoming expiratons. The subsidy summary box will not display for expired plans. "
+                    "Other, notifications signaling the expiration (and loss of usability) itself will still appear.")
```
iloveagent57, line 321: Nit: stray comma after "Other".  Whether you want to regen migrations to fix this is up to you :p

brobro10000: Good catch. Regened 👍🏽

enterprise/migrations/0217_alter_enterprisecustomer_disable_expiry_messaging_for_learner_credit_and_more.py
```diff
@@ -0,0 +1,23 @@
...
+
+    dependencies = [
+        ('enterprise', '0216_enterprisegroupmembership_errored_at_and_more'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='enterprisecustomer',
+            name='disable_expiry_messaging_for_learner_credit',
+            field=models.BooleanField(default=False, help_text='If checked, learners and admins will not receive notifications leading up to the expiration date for learner credit plans. This includes removing the expiration date from the subsidy box for upcoming expiratons. The subsidy summary box will not display for expired plans. Other notifications signaling the expiration (and loss of usability) itself will still appear.', verbose_name='Disable expiration messaging for learner credit'),
```
adamstankiewicz, line 16: nit: "expiratons" -> "expirations"

adamstankiewicz (approved):

PR #2177, opened 2024-07-23
Feat: Adding Enterprise-Customer-Support endpoint

pmakwana93:
Adding Enterprise-Customer-Support endpoint to consolidate data from various API endpoints
JIRA: https://2u-internal.atlassian.net/browse/ENT-9183

This endpoint will be used in the Customer Support Tool to display information in the users table we need data for admins, pending admins, learners and pending learners. This is a replacement endpoint in lieu of using 4 existing endpoints:
- admins: enterprise/api/v1/enterprise-customer/
- learner: enterprise/api/v1/enterprise-learner/
- pending learner: enterprise/api/v1/pending-enterprise-learner/
- pending admin: enterprise/api/v1/pending-enterprise-admin/

A valid enterprise customer UUID needs to be passed in as endpoint input as per:
`enterprise/api/v1/enterprise-customer-support/5368db2a-5faf-48a7-99d2-f2e9b5e0fc6f`

NOTE: Filtering and Sorting functionality will be added in a separate PR and ticket

Testing:
- Call the endpoint with a valid enterprise UUID
- added unit tests

1. Testing Admin Enterprise Customer User:
Enterprise Admin

2. Testing Non-Admin Enterprise Customer User:
Enterprise Non Admin

3. Testing Pending Admi Enterprise Customer User:
Pending Admin

4. Testing Pending Non-Admin Enterprise Customer User:
Pending Non Admin

5. Testing Admin + Pending Admin Enterprise Customer User:
Enteprise + Pending

5. Testing Paginated Response (response count is set to 8 objects per page)
Paginated Response

**Merge checklist:**

**Post merge:**


enterprise/api/v1/serializers.py
```diff
@@ -1828,3 +1828,111 @@ class EnterpriseGroupLearnersRequestQuerySerializer(serializers.Serializer):
...
+        """
+        Get enterprise customer user name
+        """
+        return obj.enterprise_customer.name
+
+    def get_user_email(self, obj):
+        """
+        Get enterprise customer user email
+        """
+        return obj.enterprise_customer.contact_email
```
katrinan029, line 1872: (issue): the user email should be coming from the user and not the enterprise customer.

enterprise/api/v1/serializers.py
```diff
@@ -1828,3 +1828,111 @@ class EnterpriseGroupLearnersRequestQuerySerializer(serializers.Serializer):
...
+        """
+        Get enterprise customer user id
+        """
+        return obj.user_id
+
+    def get_user_name(self, obj):
+        """
+        Get enterprise customer user name
+        """
+        return obj.enterprise_customer.name
```
katrinan029, line 1866: (issue): the user name should be derived from the user and not enterprise customer

enterprise/api/v1/views/enterprise_user.py
```diff
@@ -0,0 +1,66 @@
...
+        enterprise_uuid = kwargs.get('enterprise_uuid', None)
+
+        try:
+            enterprise_customer_queryset = models.EnterpriseCustomerUser.objects.filter(
+                enterprise_customer__uuid=enterprise_uuid
+            )
+
+            if enterprise_customer_queryset.exists():
+                serializer = serializers.EnterpriseUserSerializer(enterprise_customer_queryset, many=True)
+                return response.Response(serializer.data)
```
katrinan029, line 41: (issue): we should return a list of both pending and existing enterprise users. I think how the code is structured currently, it's returning one or the other.

I think we can try something like

```
class EnterpriseUserViewSet(EnterpriseReadOnlyModelViewSet):
    """
    API views for the ``enterprise-user`` API endpoint.
    """
    queryset = models.EnterpriseCustomerUser.objects.all()
    filter_backends = (DjangoFilterBackend,)
    permission_classes = (permissions.IsAuthenticated,)

    USER_ID_FILTER = 'enterprise_customer_users__user_id'

    def retrieve(self, request, *args, **kwargs):
        """
        - Filter down the queryset of groups available to the requesting uuid.
        """
        enterprise_uuid = kwargs.get('enterprise_uuid', None)
        users = []
        try:
            enterprise_customer_queryset = models.EnterpriseCustomerUser.objects.filter(
                enterprise_customer__uuid=enterprise_uuid,
            )
            users.extend(enterprise_customer_queryset)
            pending_enterprise_customer_queryset = models.PendingEnterpriseCustomerUser.objects.filter(
                enterprise_customer__uuid=enterprise_uuid
            ).order_by('user_email')
            users.extend(pending_enterprise_customer_queryset)
        except ValidationError:
            # did not find UUID match in either EnterpriseCustomerUser or  PendingEnterpriseCustomerUser
            return response.Response(
                {'detail': 'Could not find enterprise uuid {}'.format(enterprise_uuid)},
                status=status.HTTP_404_NOT_FOUND
            )
        serializer = serializers.EnterpriseUserSerializer(
            users,
            many=True
        )
        serializer_data = sorted(
            serializer.data, key=lambda k: k['is_admin'], reverse=True)
        return response.Response(serializer_data)
```

enterprise/api/v1/urls.py
```diff
@@ -197,6 +198,13 @@
         enterprise_group.EnterpriseGroupViewSet.as_view({'post': 'remove_learners'}),
         name='enterprise-group-remove-learners'
     ),
+    re_path(
```
katrinan029, line 201: (nit): let's rename this to something like `r'^enterprise-customer/(?P<enterprise_uuid>[A-Za-z0-9-]+)/users/?$'`

enterprise/api/v1/views/enterprise_user.py
```diff
@@ -0,0 +1,66 @@
...
+
+from enterprise import models
+from enterprise.api.v1 import serializers
+from enterprise.api.v1.views.base_views import EnterpriseReadOnlyModelViewSet
+from enterprise.logging import getEnterpriseLogger
+
+LOGGER = getEnterpriseLogger(__name__)
+
+
+class EnterpriseUserViewSet(EnterpriseReadOnlyModelViewSet):
```
katrinan029, line 18: (issue): can we also support pagination here that returns 8 entries at a time?

enterprise/api/v1/views/enterprise_user.py
katrinan029: (issue): apologies for not mentioning this in the ticket but could we also support searching for an individual user name or email?

katrinan029 (changes requested):
Looking good so far! I left some feedback and my apologies for leaving out some of the requirements in the ticket.

enterprise/api/v1/urls.py
pmakwana93: `enterprise-customer` is an already existing endpoint, so it might be difficult to bundle this functionality under that endpoint. Does the rename to `enterprise-customer-user` work instead?
Reference: https://github.com/openedx/edx-enterprise/blob/master/enterprise/api/v1/urls.py#L41

enterprise/api/v1/views/enterprise_user.py
pmakwana93: > (issue): we should return a list of both pending and existing enterprise users. I think how the code is structured currently, it's returning one or the other.
> 

Apologies, but my understanding from our initial discussion was that a user could be an enterprise customer user or a pending customer user - but not both. Is that no longer the case? Both of these datasets come from two different models thus I had split them up accordingly

But happy to make the changes if necessary

enterprise/api/v1/urls.py
pmakwana93: Discussed offline - we will go with `enterprise-customer-support`

katrinan029 (approved):
This looks great! Tested locally and the data is returned as expected! Reminder to bump up the version in `CHANGELOG.rst` and `__init__.py` before merging your code.

PR #2179, opened 2024-07-24
fix: don't allow saving duplicate EnterpriseCatalogQuery content_filter

marlonkeating:
## Problem

[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-8661)

The EnterpriseCatalogQuery model is synced to a corresponding [CatalogQuery model](https://github.com/openedx/enterprise-catalog/blob/0c832240bd67a1748406d126fd0305eeb63733a2/enterprise_catalog/apps/catalog/models.py#L54) in enterprise-catalog.  Currently, if we save a EnterpriseCatalogQuery with a content_filter value that is a duplicate of another, and try to update the corresponding CatalogQuery, the update will fail and the two entities will be in an inconsistent state.

## Solution

When saving an EnterpriseCatalogQuery model with a changed content_filter, we will now check to see if the new content_filter already exists in enterprise-catalog, and raise a ValidationError if it does, showing an error on the admin page.

image

**Merge checklist:**

**Post merge:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * nothing unreleased
 
+[4.23.1]
+---------
+* fix: don't allow saving duplicate EnterpriseCatalogQuery content_filter
+
 [4.22.6]
 ----------
 * feat: add created time to reponse
--- a/docs/decisions/0014-enterprise-catalog-sync-catalogs.rst
+++ b/docs/decisions/0014-enterprise-catalog-sync-catalogs.rst
@@ -1,0 +1,43 @@
+Catalog syncing behavior with enterprise-catalog service
+------------------------------------------------------------------------------
+
+Status
+======
+
+Accepted
+
+Context
+=======
+We would like to have a way to reuse a saved set of courses configured by a query and use them in multiple catalogs across customers.
+
+
+Decisions
+=========
+
+- **Sync catalog/customer data with enterprise-catalog**
+As the catalog data is saved in both edx-enterprise and enterprise-catalog, we need to make sure that catalog data is consistent across both.
+In the context of edx-enterprise, changes to EnterpriseCatalogQuery and EnterpriseCustomerCatalog will be propagated to their counterparts
+in the enterprise-catalog service.
+
+
+Consequences
+============
+
+- **EnterpriseCatalogQuery Sync Process**
+The enterprise-catalog service enforces a uniqueness constraint for the `content_filter` field, and so edx-enterprise must enforce it as well
+or the sync operation will fail and the entities will be out of sync with each other.
+
+In the Django Admin console, whenever a new EnterpriseCatalogQuery is created, or an existing one edited, we first query enterprise-catalog to see if its 
+`content_filter` is already in use (by calculating the `content_filter` hash via a call to enterprise-catalog, and then attempting to retrieve
+an existing catalog query from enterprise-catalog by that hash).  If the new `content_filter` is unique, we will save the changes and propagate
+them to enterprise-catalog. 
+
+If the `content_filter` is *not* unique, we display an error on the EnterpriseCatalogQuery edit page and don't commit the change. 
+
+
+- **EnterpriseCustomerCatalog Sync Process**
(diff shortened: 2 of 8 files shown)


enterprise/models.py
```diff
@@ -2422,6 +2422,20 @@ def __str__(self):
         """
         return "<EnterpriseCatalogQuery '{title}' >".format(title=self.title)
 
+    def clean(self):
```
pwnage101, line 2425: nit: missing docstring explaining the purpose of this check

pwnage101 (approved):

enterprise/api_client/enterprise_catalog.py
```diff
@@ -110,6 +113,60 @@ def get_enterprise_catalog(self, catalog_uuid, should_raise_exception=True):
...
+        try:
+            response = self.client.get(api_url, data=catalog_query)
+            response.raise_for_status()
+            return response.json()
+        except (RequestException, ConnectionError, Timeout) as exc:
+            LOGGER.exception(
+                'Failed to get catalog query hash for "[%s]" due to: [%s]',
+                catalog_query, str(exc)
+            )
+            if should_raise_exception:
```
iloveagent57, line 166: Will this end up raising an exception if the call returns a 404 and `should_raise_exception==True`?

enterprise/models.py
```diff
@@ -2422,6 +2422,20 @@ def __str__(self):
...
+        previous_values = EnterpriseCatalogQuery.objects.filter(id=self.id).first()
+        if previous_values:
+            old_filter = previous_values.content_filter
+            new_filter = self.content_filter
+            if not old_filter == new_filter:
+                catalog_client = EnterpriseCatalogApiClient()
+                old_hash = catalog_client.get_catalog_query_hash(old_filter)
+                new_hash = catalog_client.get_catalog_query_hash(new_filter)
+                if not old_hash == new_hash:
+                    hash_catalog_response = catalog_client.get_enterprise_catalog_by_hash(new_hash)
```
iloveagent57, line 2435: maybe we want `should_raise_exception=False` in this call.

CHANGELOG.rst
```diff
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * nothing unreleased
 
+[4.22.3]
```
iloveagent57, line 20: nit: I think it's worth bumping to `4.23` for this fix.

iloveagent57 (commented):
This is great, just a minor question/suggestion on exception-handling.

docs/decisions/0014-enterprise-catalog-sync-catalogs.rst
```diff
@@ -0,0 +1,43 @@
...
+------------------------------------------------------------------------------
+
+Status
+======
+
+Accepted
+
+Context
+=======
+ECS team would like to have a way to reuse a saved set of courses configured by a query and use them in multiple catalogs across customers.
```
iloveagent57, line 11: tiny nit: it's more than just a specific team that wants this. The enterprise sub-system as a whole benefits from this re-usability of queries.

docs/decisions/0014-enterprise-catalog-sync-catalogs.rst
```diff
@@ -0,0 +1,43 @@
...
+
+
+Consequences
+============
+
+- **EnterpriseCatalogQuery Sync Process**
+The enterprise-catalog service enforces a uniqueness constraint for the `content_filter` field, and so edx-enterprise must enforce it as well
+or the sync operation will fail and the entities will be out of sync with each other.
+
+In the admin console, whenever a new EnterpriseCatalogQuery is created, or an existing one edited, we first query enterprise-catalog to see if its
```
iloveagent57, line 30: nit: I would say specifically "Django Admin" just to help disambiguate from the admin portal.

iloveagent57 (approved):
ADR looks great! Two small nits.

enterprise/api_client/enterprise_catalog.py
marlonkeating: It will raise an exception regardless of `should_raise_exception`, keeping with the behavior of other enterprise-catalog client methods.

PR #2191, opened 2024-08-01
test: pii check stage actually checks PII instead of not at all

pwnage101:


pwnage101:
Docs test is failing because the RST in a file that I did not edit is invalid: https://github.com/openedx/edx-enterprise/blob/9bbfae93ba01827a0de9698a07ca72b003e59f69/docs/decisions/0014-enterprise-catalog-sync-catalogs.rst?plain=1#L17

Not fixing in this PR.

PR #2201, opened 2024-08-09
Added logs for data dog inspection for permission check in handler function

MueezKhan246:
Description:
Added logs for data dog inspection for permission check in handler function.

JIRA:
https://2u-internal.atlassian.net/jira/software/c/projects/ENT/issues/ENT-9248

### DIFF
--- a/enterprise/api/v1/decorators.py
+++ b/enterprise/api/v1/decorators.py
@@ -5,6 +5,10 @@ Decorators for Enterprise API views.
 from functools import wraps
 
 from rest_framework.exceptions import PermissionDenied, ValidationError
+from enterprise.logging import getEnterpriseLogger
+
+LOGGER = getEnterpriseLogger(__name__)
+
 
 
 def require_at_least_one_query_parameter(*query_parameter_names):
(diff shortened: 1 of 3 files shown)


enterprise/api/v1/decorators.py
```diff
@@ -57,14 +61,30 @@ def _wrapped_view(request, *args, **kwargs):
             user = request.user
             pk = fn(request, **kwargs) if fn else kwargs.get('pk')
 
+            LOGGER.info(
```
zamanafzal, line 64: How about adding a unique keyword at the start so that all the logs related to permissions can be separated immediately? e.g 
LOGGER.info(f"[User_Permissions_Check]User {user.username} has permission: {has_permission}")

zamanafzal (approved):

enterprise/api/v1/decorators.py
MueezKhan246: @zamanafzal keyword added.

sameeramin (approved):
LGTM 👍🏼

sameenfatima78 (commented):
Looks good!

PR #2206, opened 2024-08-19
Feat: Add option to show soft deleted group memberships in django admin

pmakwana93:
Adding option to show soft deleted group memberships in django admin
JIRA: https://2u-internal.atlassian.net/browse/ENT-9116

**Testing**
Setup:
EnterpriseCustomerUser `Davidson Jones` is part of soft deleted group membership by marking their record as "is_removed" = 1
EnterpriseCustomerUser `Jackson LLC` is not part of soft deleted group membership and their record has "is_removed" = 0

1. Keeping default behavior to only show non-soft deleted results:
all users

2. Filtering by `is_removed` = No shows non-deleted results (Jackson LLC):
not deleted

3. Filtering by `is_removed` = Yes shows only soft-deleted result (Davidson Jones):
soft deleted

4. Searching for soft deleted group membership displays warning
is_removed

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/admin/__init__.py
+++ b/enterprise/admin/__init__.py
@@ -1273,7 +1273,11 @@ class EnterpriseGroupMembershipAdmin(admin.ModelAdmin):
         """
         Return a QuerySet of all model instances.
         """
-        qs = self.model.available_objects.get_queryset()
+        show_soft_deletes = request.GET.get('is_removed__exact', False)
+        if show_soft_deletes:
+            qs = self.model.all_objects.get_queryset()
+        else:
+            qs = self.model.available_objects.get_queryset()
 
         ordering = self.get_ordering(request)
         if ordering:
--- a/enterprise/admin/__init__.py
+++ b/enterprise/admin/__init__.py
@@ -1255,6 +1255,7 @@ class EnterpriseGroupMembershipAdmin(admin.ModelAdmin):
     """
     model = models.EnterpriseGroupMembership
     list_display = ('group', 'membership_user',)
+    list_filter = ('is_removed',)
     search_fields = (
         'uuid',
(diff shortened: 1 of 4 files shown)


enterprise/admin/__init__.py
```diff
@@ -1268,6 +1269,22 @@ class EnterpriseGroupMembershipAdmin(admin.ModelAdmin):
         'pending_enterprise_customer_user',
     )
 
+    def get_queryset(self, request):
```
katrinan029, line 1272: [suggestion] could we add in a test for this?

enterprise/admin/__init__.py
```diff
@@ -1255,6 +1255,7 @@ class EnterpriseGroupMembershipAdmin(admin.ModelAdmin):
     """
     model = models.EnterpriseGroupMembership
     list_display = ('group', 'membership_user',)
+    list_filter = ('is_removed',)
```
katrinan029, line 1258: [nit/suggestion] since we're adding in the filter, I think we could also add the same field in the `list_display` to make it clear to the user which list they are viewing.

`list_display = ('group', 'membership_user', 'is_removed')`

katrinan029 (changes requested):

enterprise/admin/__init__.py
pmakwana93: Updated as per suggestion
Screenshot 2024-08-20 at 11 26 03 AM

enterprise/admin/__init__.py
pmakwana93: Updated as per suggestion

katrinan029 (approved):
thanks for adding in the test! looks great

PR #2208, opened 2024-08-22
Added encrypted credentials column in sap configuration

MueezKhan246:
Description:
Added encryption fields to the model of SAP config.

JIRA:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * nothing unreleased
 
+[4.23.10]
+----------
+* feat: added encrypted columns for user credentials for SAP config
+
 [4.23.9]
 ----------
 * feat: Add option to show soft deleted group memberships in django admin
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,4 +2,4 @@
 Your project description goes here.
 """
 
-__version__ = "4.23.9"
+__version__ = "4.23.10"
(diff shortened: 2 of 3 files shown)


sameeramin (commented):
Looks good but according to this [docs](https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations) you need to register a signal as well to copy data from old fields to the new ones

sameeramin (approved):
LGTM 👍🏼

enterprise/signals.py
```diff
@@ -464,3 +464,14 @@ def generate_default_orchestration_record_display_name(sender, instance, **kwarg
...
+
+@receiver(pre_save, sender=SAPSuccessFactorsEnterpriseCustomerConfiguration)
+def update_decrypted_credentials(sender, instance, **kwargs):     # pylint: disable=unused-argument
+    """
+    Ensure that the decrypted credentials have same values as unencrypted credentials.
+    """
+    if instance.key != instance.decrypted_key:
+        instance.encrypted_key = instance.key
+    if instance.secret != instance.decrypted_secret:
+        instance.encrypted_secret = instance.secret
```
sameenfatima78, line 477: as per discussion in call, please update to refer model attributes here

sameenfatima78 (approved):
Looks good just added on suggestion

enterprise/signals.py
MueezKhan246: @sameenfatima78 updated.

MueezKhan246:
@sameenfatima78 i removed the fields from serializer as i was unable to confirm that the endpoint was public through DataDog. i will be adding those in follow up PR.

PR #2215, opened 2024-08-26
Data migration for encrypted credentials column in sap configuration

MueezKhan246:
Description:
added data migration to populate encrypted columns and replaced references from unencrypted to encrypted columns.
this PR must only be merged when
1st [Removing decrypted client secret in sap configuration from code #2228](https://github.com/openedx/edx-enterprise/pull/2228)
then
2nd [Migration for removing decrypted client secret in sap configuration #2230](https://github.com/openedx/edx-enterprise/pull/2230)
and then
3rd [Added encrypted client secret in sap configuration #2231](https://github.com/openedx/edx-enterprise/pull/2231)
gets deployed

JIRA:


tests/test_integrated_channels/test_api/test_sap_success_factors/test_views.py
```diff
@@ -70,10 +72,6 @@ def test_get(self, mock_current_request):
                          self.sap_config.sapsf_company_id)
         self.assertEqual(int(data.get('sapsf_user_id')),
                          self.sap_config.sapsf_user_id)
-        self.assertEqual(data.get('key'),
-                         self.sap_config.key)
-        self.assertEqual(data.get('secret'),
-                         self.sap_config.secret)
```
sameenfatima78, line 76: can we assert this on the encrypted columns?

sameenfatima78 (changes requested):
A couple of issues in the PR that requires modifications before merging. 

- First of all, that `decrypted_secret` field should have been stored as `EncryptedCharField` in the previous PR. 
- I believe the order in which we are applying migrations is going to create serious data inconsistency issues as we are populating the new fields with data first then making a change to the decrypted_secret to change its field type in the third migration. Please be aware that changing the field type in the model does not automatically update the existing records in the database. Only new data saved will be encrypted before storing in DB but the old data will remain stored in plain text.
- Would advice creating a separate release to convert the field type first then in the second release create a migration to populate the data in the new column in encrypted form and replace the old columns with new ones. Finally, in the third release, drop the old columns following the guidelines mentioned in the database migrations doc.  


@sameenfatima78 decrypted_secret field type should have been set to `EncryptedCharField` in the first release but instead it remained `models.CharField` as there was not much information about altering the definition of field in the doc i tried altering it here, but you are right the migration order is off data migration should be attempted after field is set to `EncryptedCharField`

tests/test_integrated_channels/test_api/test_sap_success_factors/test_views.py
MueezKhan246: @sameenfatima78 each time encryption is attempted is made in the getter a new value encrypted value is returned, which is not possible to calculate and use in assertion

tests/test_integrated_channels/test_api/test_sap_success_factors/test_views.py
```diff
@@ -91,20 +89,43 @@ def test_update(self, mock_current_request):
...
+            'sapsf_user_id': 893489,
+            'user_type': 'user',
+            'encrypted_secret': '1000',
+        }
+        self.client.put(url, payload)
+        self.sap_config.refresh_from_db()
+        self.assertEqual(self.sap_config.decrypted_secret, '1000')
+        populate_decrypted_fields_sap_success_factors(apps)
+        self.sap_config.refresh_from_db()
+        self.assertEqual(self.sap_config.encrypted_secret, client_secret)
```
sameeramin, line 127: I don't think we will need this test anymore once this migration is done. Since this is a one-off change, I would suggest using a flag to skip coverage something like [this](https://github.com/openedx/edx-enterprise/blob/0e99db8b8fa39b3f38b9a4d23a8c2fa2c23dc0c3/integrated_channels/canvas/migrations/0037_canvasenterprisecustomerconfiguration_copy_id_and_secret_and_more.py#L6).

sameeramin (approved):

tests/test_integrated_channels/test_api/test_sap_success_factors/test_views.py
MueezKhan246: @sameeramin i think if migration file is going to remain in code, we can keep the test case.

sameenfatima78 (approved):

zamanafzal (approved):

PR #2219, opened 2024-08-29
Altering client id and secret in sap configuration

MueezKhan246:
Description:
altered decrypted_secret field type to be encrypted and made unencrypted credentials nullable so refernces can be removed from test cases.
this PR can be merged as it has no dependency.

JIRA:


sameenfatima78 (approved):

integrated_channels/sap_success_factors/models.py
```diff
@@ -120,10 +121,11 @@ class SAPSuccessFactorsEnterpriseCustomerConfiguration(EnterpriseCustomerPluginC
         blank=True,
         default='',
         verbose_name="Client Secret",
-        help_text=_("OAuth client secret.")
+        help_text=_("OAuth client secret."),
+        null=True
```
zamanafzal, line 125: Any reason to make it `null=True` now?

MueezKhan246: @zamanafzal in the follow up PR we need to remove all references for these fields so while creating records of SAP config in the test cases if these are not set to be nullable removing references will cause errors.

zamanafzal (approved):

PR #2259, opened 2024-10-07
feat: add default enrollment models

iloveagent57:
Adds two models, `DefaultEnterpriseEnrollmentIntention` and `DefaultEnterpriseEnrollmentRealization`. The former defines, for a customer, a course/run that associated users should be automatically enrolled into. The latter represents the relationship between that intention record and a realized EnterpriseCourseEnrollment; persisting realized enrollments in this way will help to make read operations related to default enrollments much more efficient.
ENT-9577

**Merge checklist:**

**Post merge:**


docs/decisions/0015-default-enrollments.rst
brobro10000, line 1: [curious] 2 observations.

If the `DefaultEnterpriseEnrollmentIntention` `content_key` enrollable course differs from `DefaultEnterpriseEnrollmentRealization`, `EnterpriseCourseEnrollment` enrolled course, are there any consequences from the mismatch?

Were there any alternatives considered?

enterprise/models.py
```diff
@@ -2461,6 +2461,103 @@ class LicensedEnterpriseCourseEnrollment(EnterpriseFulfillmentSource):
...
+        ),
+    )
+    content_key = models.CharField(
+        max_length=255,
+        blank=False,
+        null=False,
+        help_text=_(
+            "A course or course run that related users should be automatically enrolled into."
+        ),
+    )
```
brobro10000, line 2506: [questionx2] is there some benefit to including `parent_content_key` as a field? Should we bring the pattern of `is_assigned_course_run` from access here as well, `is_default_course_run`?

brobro10000 (approved):
Few non-blocking questions. Great job!

docs/decisions/0015-default-enrollments.rst
```diff
@@ -0,0 +1,54 @@
...
+   or a specific course run for automatic enrollment. In the case of specifying a course,
+   the default enrollment flow should cause the "realization" of default enrollments for learners
+   into the currently-advertised, enrollable run for a course.
+2. Default Enrollments should be loosely coupled to subsidy type.
+3. Unenrollment: If a learner chooses to unenroll from a default course, they should not be automatically re-enrolled.
+4. Graceful handling of license revocation: Upon license revocation, we currently downgrade the learner’s
+   enrollment mode to ``audit``. This fact should be visible from any new APIs exposed
+   in the domain of default enrollments.
+5. Non-enrollable courses: If a course becomes unenrollable, default enrollments should fail gracefully,
+   and in a way that's obvious to the learner.
```
adamstankiewicz, line 28: [thought] To clarify, the intent is not to attempt enrollment/redemption for known, non-enrollable courses, as deemed by the proposed REST API surfacing the default enrollment intentions based on the learner's enrollment state, e.g.:

```
{
  "default_enterprise_enrollment_intentions": {
    "needs_enrollment": {
      "enrollable": [{ ... }],
      "not_enrollable": [{ ... ]}
    },
    "already_enrolled": [{ ... }]
  }
}
```

I'm not sure whether there is necessarily a product/UX requirement here that, if we fail to realize (enroll/redeem) a default enrollment intention, we'd need to show any error messaging to the learner. The learner had no notion that they were supposed to automatically get a default enrollment, so it might be confusing to surface error messaging to that effect. I was considering it might "fail gracefully" by just not getting a realized `EnterpriseCourseEnrollment`, but ensuring we have observability/alerting where relevant.

Probably something to clarify with Product?

docs/decisions/0015-default-enrollments.rst
```diff
@@ -0,0 +1,54 @@
...
+
+Decision
+========
+We will implement two new models, ``DefaultEnterpriseEnrollmentIntention`` to represent the course/runs that
+learners should be default-enrolled into for a given enterprise, and ``DefaultEnterpriseEnrollmentRealization``
```
adamstankiewicz, line 33: nit: I wonder if using `auto-enrolled`, specifically post-logistration, here might make sense?

docs/decisions/0015-default-enrollments.rst
```diff
@@ -0,0 +1,54 @@
...
+which represents the mapping between the intention and actual enrollment record(s) for the learner/customer.
```
adamstankiewicz, line 34: nit/consider: "... between the intention and actual, **realized** enrollment record(s) for the learner/customer."

docs/decisions/0015-default-enrollments.rst
```diff
@@ -0,0 +1,54 @@
...
+
+Qualities
+---------
+1. Flexibility: The ``DefaultEnterpriseEnrollmentIntention`` model will allow specification of either a course
+   or course run.
+2. Business logic: The API for this domain (future ADR) will implement the business logic around choosing
+   the appropriate course run, for answering which if any catalogs are applicable to the course,
+   and the enrollability of the course.
```
adamstankiewicz, line 42: [clarification] Does "enrollability" here include the state of the learner's existing `EnterpriseCourseEnrollment`(s) or only the content metadata level of "enrollable"?

docs/decisions/0015-default-enrollments.rst
```diff
@@ -0,0 +1,54 @@
...
+3. Non-Tightly Coupled to subsidy type: Nothing in the domain of default enrollments will persist data
+   related to a subsidy (although a license or transaction identifier will ultimately become associated with
+   an ``EnterpriseCourseEnrollment`` record during realization).
+
+Consequences
+============
+1. It's a flexible design.
+2. It relies on a network call to enterprise-catalog to fetch content metadata (we can use caching to
+   make this efficient).
```
adamstankiewicz, line 51: Currently, as spec'ed, I believe relies on two network calls:
1. Retrieve content metadata for the provided `content_key`.
2. Retrieve the list of applicable catalogs to the `content_key`.

Do you think it'd be worth mentioning the 2 calls here?

docs/decisions/0015-default-enrollments.rst
```diff
@@ -0,0 +1,54 @@
...
+3. We're introducing more complexity in terms of how subsidized enterprise enrollments
+   can come into existence.
+4. The realization model makes the provenance of default enrollments explicit and easy to examine.
```
adamstankiewicz, line 54: :tada:

enterprise/models.py
```diff
@@ -2461,6 +2461,103 @@ class LicensedEnterpriseCourseEnrollment(EnterpriseFulfillmentSource):
...
+        EnterpriseCustomer,
+        blank=False,
+        null=False,
+        related_name="default_course_enrollments",
+        on_delete=models.deletion.CASCADE,
+        help_text=_(
+            "The customer for which this default enrollment will be realized.",
+        )
+    )
+    content_type = models.CharField(
```
adamstankiewicz, line 2490: [curious] What's your thinking around how this field would be utilized? From a functionality perspective, I'm not we'd necessary need it per se for the computed properties to function BUT having this field would definitely help better understand/analyze/query these models to understand how many courses vs. course runs are configured.

Do you think this is intended to be manually set at time of configuring a `DefaultEnterpriseEnrollmentIntention`, or would it be programmatically set after saving based on deducing whether the provided `content_key` is a course vs. course run (perhaps based on the content metadata API request made to enterprise-catalog which I believe returns the `content_key` vs. `parent_content_key` relationship?

enterprise/models.py
```diff
@@ -2461,6 +2461,103 @@ class LicensedEnterpriseCourseEnrollment(EnterpriseFulfillmentSource):
...
+        max_length=127,
```
adamstankiewicz, line 2491: nit/curious: Why `127` here?

docs/decisions/0015-default-enrollments.rst
```diff
@@ -0,0 +1,54 @@
...
+Core requirements
+-----------------
+1. Internal staff should be able to configure one or more default enrollments with either a course
+   or a specific course run for automatic enrollment. In the case of specifying a course,
+   the default enrollment flow should cause the "realization" of default enrollments for learners
+   into the currently-advertised, enrollable run for a course.
+2. Default Enrollments should be loosely coupled to subsidy type.
+3. Unenrollment: If a learner chooses to unenroll from a default course, they should not be automatically re-enrolled.
+4. Graceful handling of license revocation: Upon license revocation, we currently downgrade the learner’s
+   enrollment mode to ``audit``. This fact should be visible from any new APIs exposed
```
adamstankiewicz, line 25: Good call to make the course mode known in the proposed `default-enterprise-enrollment-intentions/learner-status/` GET API.

docs/decisions/0015-default-enrollments.rst
iloveagent57: This is nice, so I copied it into a new section of the ADR.

enterprise/models.py
iloveagent57: (2^7 - 1). i.e the same reason we use `max_length` of 255 instead of 256. I'm not sure how much this matters, if it matters at all. There used to be some thing about using one fewer chars than a power of two to make page fetches more efficient (I think).

enterprise/models.py
iloveagent57: Either one - we could populate it automatically based on a simple regular expression.

docs/decisions/0015-default-enrollments.rst
```diff
@@ -0,0 +1,54 @@
...
+to manage default enrollments via both learner credit and subscription subsidy types.
+
+Core requirements
+-----------------
+1. Internal staff should be able to configure one or more default enrollments with either a course
+   or a specific course run for automatic enrollment. In the case of specifying a course,
+   the default enrollment flow should cause the "realization" of default enrollments for learners
+   into the currently-advertised, enrollable run for a course.
+2. Default Enrollments should be loosely coupled to subsidy type.
+3. Unenrollment: If a learner chooses to unenroll from a default course, they should not be automatically re-enrolled.
```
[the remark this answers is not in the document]
iloveagent57: I think there should be a distinct ADR (loosely) describing the proposed API contract.

enterprise/models.py
adamstankiewicz: Interesting, TIL haha.

adamstankiewicz (approved):
LGTM, one addtl nit about the name of `PendingEnterpriseCourseEnrollments` vs. `PendingEnrollments` in the ADR.

PR #2264, opened 2024-10-17
feat: Create django admin for default enrollments

brobro10000:
Adds a TieredCache call to content metadata's, `get_content_metadata_content_identifier` to determine the `content_type` of created and updated content. If the content type does not exist, the validation in `clean` prevents creation of the inline object. 

A side effect of including this check is the fleshing out of the `current_course_run` and `current_course_run_key` properties on the `DefaultEnterpriseEnrollmentIntention` model.

Found on the enterprise customer django admin screen, `EnterpriseCustomerDefaultEnterpriseEnrollmentIntentionInline`
Screenshot 2024-10-21 at 1 50 52 PM

Cleaned up `EnterpriseCustomerDefaultEnterpriseEnrollment` admin page
Screenshot 2024-10-21 at 1 09 54 PM

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -2511,12 +2513,41 @@ class DefaultEnterpriseEnrollmentIntention(TimeStampedModel, SoftDeletableModel)
     )
     history = HistoricalRecords()
 
+    def get_cached_content_metadata(self):
```
adamstankiewicz, line 2516: [inform] Based on our discussion, the underlying `EnterpriseCatalogAPIClient` method to be called (i.e., `get_content_metadata_content_identifier`) has a try/except with reasonable returns of empty dict. I think we can nix this helper method and directly call `get_and_cache_catalog_content_metadata` where needed.

enterprise/models.py
```diff
@@ -2511,12 +2513,41 @@ class DefaultEnterpriseEnrollmentIntention(TimeStampedModel, SoftDeletableModel)
...
         Metadata describing the current course run for this default enrollment intention.
         """
-        return {}
+        content_metadata_items = self.get_cached_content_metadata()
+        if not content_metadata_items:
+            return {}
+
+        content_metadata_item = content_metadata_items[0]
+        if self.content_type == 'course':
+            return get_advertised_course_run(content_metadata_item)
```
adamstankiewicz, line 2548: [inform] `get_advertised_course_run` returns `None` if there is no `advertised_course_run_uuid` (e.g., `null`). We may want to ensure consistency in the return value (e.g., `if None: return {}`).

enterprise/models.py
```diff
@@ -2511,12 +2513,31 @@ class DefaultEnterpriseEnrollmentIntention(TimeStampedModel, SoftDeletableModel)
     )
     history = HistoricalRecords()
 
+    def get_cached_content_metadata(self):
```
adamstankiewicz, line 2516: [consideration] Would it be helpful if this was a `@property`? E.g., `def content_metadata_for_content_key(self)`. I believe the benefit would be accessing the metadata via `self.content_metadata_for_content_key` vs. needing to call `self.get_cached_content_metadata()` each time.

The revised naming is just a suggestion :)

enterprise/models.py
```diff
@@ -2511,12 +2513,31 @@ class DefaultEnterpriseEnrollmentIntention(TimeStampedModel, SoftDeletableModel)
...
+        Retrieves the cached content metadata for the instance content_key
+        Determines catalog from the catalogs associated to the enterprise customer
+        """
+        return get_and_cache_catalog_content_metadata(
+                enterprise_customer_uuid=self.enterprise_customer.uuid,
+                content_key=self.content_key,
+            )
+
+
     @cached_property
```
adamstankiewicz, line 2527: [consideration, sanity check] Upon further reflection, I'm not actually sure we'd want to have this property use the `@cached_property` decorator vs. the normal `@property` decorator.

Rationale: I believe `@cached_property` will persist as long as the model instance does, unless manually invalidated or the model instance is updated (e.g., explicitly re-saved). However, the `current_course_run` ideally should be able to change over time based on the configured `content_key` (i.e., when it's a top-level course, relying on the advertised run) without needing to manually re-save instances or manually invalidate the cached property value.

I think this is mitigated by the fact that we'd be utilizing cached (`TieredCache`) content metadata here regardless, but will pull in fresh data once the TTL elapses.

enterprise/models.py
```diff
@@ -2511,12 +2513,32 @@ class DefaultEnterpriseEnrollmentIntention(TimeStampedModel, SoftDeletableModel)
...
         """
-        return {}
+        content_metadata = self.content_metadata_for_content_key
+        if not content_metadata:
+            return {}
+        if self.content_type == 'course':
+            if course_run := get_advertised_course_run(content_metadata):
+                return course_run
+            return {}
+        course_runs = content_metadata.get('course_runs', {})
```
adamstankiewicz, line 2540:
```suggestion
        course_runs = content_metadata.get('course_runs', [])
```

enterprise/models.py
```diff
@@ -2511,12 +2513,41 @@ class DefaultEnterpriseEnrollmentIntention(TimeStampedModel, SoftDeletableModel)
...
-        return {}
+        content_metadata_items = self.get_cached_content_metadata()
+        if not content_metadata_items:
+            return {}
+
+        content_metadata_item = content_metadata_items[0]
+        if self.content_type == 'course':
+            return get_advertised_course_run(content_metadata_item)
+        course_runs = content_metadata_item.get('course_runs', {})
+        return course_runs.get(self.content_key, {})
```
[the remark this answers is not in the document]
brobro10000: `next((course_run for course_run in course_runs if course_run['key'].lower() == self.content_key.lower()), {})` :)

enterprise/admin/__init__.py
```diff
@@ -1332,8 +1358,7 @@ class DefaultEnterpriseEnrollmentIntentionAdmin(admin.ModelAdmin):
 
     readonly_fields = (
         'current_course_run_key',
-        'current_course_run_enrollable',
-        'current_course_run_enroll_by_date',
```
brobro10000, line 1336: It was decided for the moment, since the properties on the model are yet to be fleshed out, these fields should be removed.

iloveagent57 (approved):

adamstankiewicz (approved):

PR #2269, opened 2024-10-22
fix: follow-up / updates related to ENT-9579

adamstankiewicz:
[context] Contribution to related PR https://github.com/openedx/edx-enterprise/pull/2264

* Updates Django Admin fields for models.DefaultEnterpriseEnrollmentIntention
* Ensures DefaultEnterpriseEnrollmentIntentionAdmin uses SoftDeletableModel's all_objects model manager to be able to view/restore soft-deleted instances.
* Adds help text in Django Admin for SoftDeletableModel's is_removed boolean field.
* Alters content_type model field to allow blank/null.
* Adds unique constraint for enterprise_customer and content_key to prevent duplicates.
* Remove `current_*` prefixes on computed properties since a configured content_key might not be the *current* course run.
* Updates models.DefaultEnterpriseEnrollmentIntention `clean` method to check for soft-deleted duplicates to raise ValidationError.
* Ensures `self.content_type` is already updated on save, even if already set (e.g., if content_key changes to a different content type).
* Updates related tests.

**Merge checklist:**

**Post merge:**


enterprise/admin/__init__.py
```diff
@@ -1370,5 +1392,22 @@ class DefaultEnterpriseEnrollmentIntentionAdmin(admin.ModelAdmin):
...
 
     class Meta:
-        fields = '__all__'
         model = models.DefaultEnterpriseEnrollmentIntention
+
+    def get_queryset(self, request):  # pylint: disable=unused-argument
+        """
+        Return a QuerySet of all model instances.
+        """
+        return self.model.all_objects.get_queryset()
```
adamstankiewicz, line 1401: Ensures soft-deleted instances are still discoverable in Django Admin.

enterprise/migrations/0225_alter_defaultenterpriseenrollmentintention_content_type_and_more.py
```diff
@@ -0,0 +1,27 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name='defaultenterpriseenrollmentintention',
+            name='content_type',
+            field=models.CharField(blank=True, choices=[('course', 'Course'), ('course_run', 'Course Run')], help_text='The type of content (e.g. a course vs. a course run).', max_length=127, null=True),
+        ),
+        migrations.AlterField(
+            model_name='historicaldefaultenterpriseenrollmentintention',
+            name='content_type',
+            field=models.CharField(blank=True, choices=[('course', 'Course'), ('course_run', 'Course Run')], help_text='The type of content (e.g. a course vs. a course run).', max_length=127, null=True),
```
adamstankiewicz, line 21: Allows `blank=True` and `null=True`

enterprise/models.py
```diff
@@ -2513,78 +2515,149 @@ class DefaultEnterpriseEnrollmentIntention(TimeStampedModel, SoftDeletableModel)
...
+        """
+        Determines the content_type for a given content_key by validating the return value
+        from `content_metadata_for_content_key`. First determines if the configured content_key
+        matches the returned key, then checks if it matches any of the returned course runs.
+        
+        Returns either COURSE, COURSE_RUN, or None (if neither can be determined).
+        """
+        content_metadata = self.content_metadata_for_content_key
+        if not content_metadata:
+            return None
```
brobro10000, line 2601: [nit] Could be simplified to a 1 liner similar to above
``` py
if not (content_metadata := self.content_metadata_for_content_key):
```

enterprise/models.py
```diff
@@ -2513,78 +2515,149 @@ class DefaultEnterpriseEnrollmentIntention(TimeStampedModel, SoftDeletableModel)
     )
     history = HistoricalRecords()
 
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=['enterprise_customer', 'content_key'],
+                name='unique_default_enrollment_intention',
+            )
```
iloveagent57, line 2523: Good idea

enterprise/models.py
```diff
@@ -2513,78 +2515,149 @@ class DefaultEnterpriseEnrollmentIntention(TimeStampedModel, SoftDeletableModel)
...
+                args=[existing_record.uuid],
+            )
+            message = _(
+                f'A default enrollment intention with this enterprise customer and '
+                f'content key already exists, but is soft-deleted. Please restore '
+                f'it <a href="{existing_record_admin_url}">here</a>.',
+            )
+            raise ValidationError({
+                'content_key': mark_safe(message)
+            })
```
iloveagent57, line 2642: Good idea, too.

iloveagent57 (approved):
nice

enterprise/models.py
```diff
@@ -2513,78 +2515,149 @@ class DefaultEnterpriseEnrollmentIntention(TimeStampedModel, SoftDeletableModel)
...
+
+        if not self.course_run:
+            raise ValidationError({
+                'content_key': _('The content key did not resolve to a valid course run.')
+            })
```
brobro10000, line 2647: [nit] As part of our async conversation, this check is also an inferred check for the `content_type` since both calls will fail (`self.course_run` and `self.determine_content_type()`)

[optional] I would add a comment about how this check is also an inferred check on the `content_type`

brobro10000 (approved):
Great cleanup with super minor nits 👍🏽

PR #2270, opened 2024-10-23
feat: creating new enterprise group members endpoint

kiram15:
Adding in an endpoint for the data table that is created in this ticket. We will be fetching info about the user (email, joined_org, name) from the enterprise_customer_user, and then fetching the total number of enrollments associated with that enterprise user. 

[Jira](https://2u-internal.atlassian.net/browse/ENT-9670)

Members are sorted on name
Screenshot 2024-11-19 at 10 47 17 PM

Testing: 
The link is http://localhost:18000/enterprise/api/v1/enterprise-customer-members/{enterprise-uuid}
- Check that each row has an enterprise_customer_user dict with info like first/last name and an enrollment count
- Add `?user_query=test` to url to make sure it is only filtering for either emails or names that contain the user query

**Merge checklist:**

**Post merge:**


enterprise/api/v1/views/enterprise_customer_members.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class EnterpriseCustomerMembersViewSet(EnterpriseReadOnlyModelViewSet):
+    """
+    API views for the ``enterprise-customer-members`` API endpoint.
+    """
+    queryset = models.EnterpriseCustomerUser.objects.all()
+    serializer_class = serializers.EnterpriseMembersSerializer
+
+    # filter_backends = (DjangoFilterBackend, filters.OrderingFilter)
```
katrinan029, line 62: let's remove this comment.

katrinan029 (changes requested):
Looks good so far! I think we're just missing the sorting piece since we want the members table to be alphabetized per the design specs. https://www.figma.com/design/eMWRd2jVa4Jqc6y7qrUfaa/Groups-Phase-2?node-id=1193-6018&t=PancQEnrxtpLLI1J-1

tests/test_enterprise/api/test_views.py
```diff
@@ -9976,6 +9976,52 @@ def test_list_users_filtered(self):
...
+        data = response.json().get('results')
+
+        # list should be sorted alphabetically by name
+        print(data)
+        self.assertEqual(data[0]['enterprise_customer_user']['first_name'], user_3.first_name)
+        self.assertEqual(data[1]['enterprise_customer_user']['first_name'], user_2.first_name)
+        self.assertEqual(data[2]['enterprise_customer_user']['first_name'], user_1.first_name)
+
+
 class TestDefaultEnterpriseEnrollmentIntentionViewSet(BaseTestEnterpriseAPIViews):
```
adamstankiewicz, line 10025: I think the test is failing because the decorators `@ddt.ddt` and `@mark.django_db` were removed when `TestEnterpriseCustomerMembers` was added. When adding them back in, note this viewset's tests show 17 tests on `master` vs. 16 tests on this PR. If ddt was properly working for the test class, the `test_default_enrollment_intentions_learner_status_already_enrolled_active_audit` would actually be two tests:

* `test_default_enrollment_intentions_learner_status_already_enrolled_active_audit_1`
* `test_default_enrollment_intentions_learner_status_already_enrolled_active_audit_2`

katrinan029 (approved):
looks good! let's add the e2e test for the view in `edx_platform`

PR #2276, opened 2024-10-30
feat: create DefaultEnterpriseEnrollmentRealization during bulk enrollment

adamstankiewicz:
## Description

**Goal:** When using the bulk enrollment API to auto-enroll learners in default enterprise enrollment intentions, related `DefaultEnterpriseEnrollmentRealization` records should be created.

[ENT-9684](https://2u-internal.atlassian.net/browse/ENT-9684)

### CHANGELOG

* Extended the `enroll_learners_in_courses` POST API within `EnterpriseCustomerViewSet` to (optionally) accept a new property `is_default_auto_enrollment: <boolean>` on each bulk enrollment attempt. When `is_default_auto_enrollment` is `True` and the enrollment has succeeded:
  * Perform a lookup of `DefaultEnterpriseEnrollmentIntention` records that exist for the enterprise customer.
  * Find a matching `DefaultEnterpriseEnrollmentIntention` to the associated `EnterpriseCourseEnrollment` based on matching course run key for the intention and course id in the enrollment.
  * If a matching `DefaultEnterpriseEnrollmentIntention` is found, gets/creates `DefaultEnterpriseEnrollmentRealization` record for the `intended_enrollment` and `realized_enrollment`.

* Updated `DefaultEnterpriseEnrollmentRealization` model to alter the `realized_enrollment` field to be a `OneToOneField` instead of a `ForeignKey`. This change makes it have more parity with the related models like `LicensedEnterpriseCourseEnrollment` and `LearnerCreditEnterpriseCourseEnrollment`.

* Added property `default_enterprise_enrollment_realization` to `EnterpriseCourseEnrollment` model, similar to the existing `license` and `learner_credit_fulfillment` properties.

* Resolves deprecation warning with `SoftDeleteableModel` related to the `DefaultEnterpriseEnrollmentIntention` model.

> DeprecationWarning: ModelBase.objects model manager will include soft-deleted objects in an upcoming release; please use ModelBase.available_objects to continue excluding soft-deleted objects. See https://django-model-utils.readthedocs.io/en/stable/models.html#softdeletablemodel for more information.

* Renames a field in the `metadata` dictionary returned by the `DefaultEnterpriseEnrollmentIntentionViewSet` (`learner_status` action).
  * I.e., `total_default_enterprise_course_enrollments` -> `total_default_enterprise_enrollment_intentions`

**Merge checklist:**

**Post merge:**


enterprise/api/v1/views/default_enterprise_enrollments.py
```diff
@@ -129,8 +129,10 @@ def learner_status(self, request):
             )
 
         # Retrieve configured default enrollment intentions for the enterprise customer
-        default_enrollment_intentions_for_customer = models.DefaultEnterpriseEnrollmentIntention.objects.filter(
-            enterprise_customer=enterprise_customer_uuid,
+        default_enrollment_intentions_for_customer = (
+            models.DefaultEnterpriseEnrollmentIntention.available_objects.filter(
```
iloveagent57, line 133: Breadcrumb for future readers on `SoftDeletableModel`, TIL!
> Note that relying on the default objects manager to filter out not-deleted instances is deprecated. objects will include deleted objects in a future release. Until then, the recommended course of action is to use the manager all_objects when you want to include all instances.
https://django-model-utils.readthedocs.io/en/latest/models.html#softdeletablemodel

enterprise/utils.py
```diff
@@ -1920,6 +1941,50 @@ def customer_admin_enroll_user_with_status(
...
+
+    if created:
+        LOGGER.info(
+            "Created default enterprise enrollment realization for default enrollment "
+            "intention %s and enterprise course enrollment %s",
+            default_enterprise_enrollment_intention.uuid,
+            enterprise_course_enrollment.id
+        )
+
+    return default_enterprise_enrollment_intention, default_enterprise_enrollment_realization
```
iloveagent57, line 1985: chefs kiss

tests/test_enterprise/api/test_views.py
```diff
@@ -5792,6 +5835,128 @@ def test_enroll_learners_in_courses_partial_failure(self, mock_get_course_mode,
...
+                        'email': 'abc@test.com',
+                        'course_run_key': fake_catalog_api.FAKE_COURSE_RUN.get('key'),
+                        'mode': VERIFIED_COURSE_MODE,
+                        'license_uuid': '2aa06d86-e2bf-40b8-81e7-b0449fa5541f',
+                        'is_default_auto_enrollment': True,
+                    },
+                ]
+            },
+            'fulfillment_source': LicensedEnterpriseCourseEnrollment,
+            'expected_enrollment_realization_count': 0,
```
iloveagent57, line 5855: Shouldn't this test create a new realization?

iloveagent57 (commented):
Looks great, one question about that positive test case.

enterprise/api/v1/views/default_enterprise_enrollments.py
adamstankiewicz: Yeah, I think the future change to always need to rely on `available_objects` was inspired by this [issue thread](https://github.com/jazzband/django-model-utils/issues/364).

tests/test_enterprise/api/test_views.py
adamstankiewicz: D'oh, you're correct! This wasn't working properly as I had not added `is_default_auto_enrollment` to `EnrollmentsInfoSerializer`, so the field wasn't including during processing, always falling back to `is_default_auto_enrollment=False`.

Updated/fixed via https://github.com/openedx/edx-enterprise/pull/2276/commits/24886e1522fcf43fd70642b2dfd989b2a77ca710

iloveagent57 (approved):

PR #2279, opened 2024-11-01
feat: add new endpoint to unlink the logged in user

jajjibhai008:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/views/enterprise_customer.py
+++ b/enterprise/api/v1/views/enterprise_customer.py
@@ -17,6 +17,7 @@ from rest_framework.status import (
     HTTP_202_ACCEPTED,
     HTTP_400_BAD_REQUEST,
     HTTP_409_CONFLICT,
+    HTTP_422_UNPROCESSABLE_ENTITY,
 )
 
 from django.contrib import auth
@@ -460,7 +461,6 @@ class EnterpriseCustomerViewSet(EnterpriseReadWriteModelViewSet):
         """
         Unlinks users with the given emails from the enterprise.
         """
-
         serializer = serializers.EnterpriseCustomerUnlinkUsersSerializer(
             data=request.data
         )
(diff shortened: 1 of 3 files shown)


enterprise/api/v1/views/enterprise_customer.py
```diff
@@ -487,3 +487,32 @@ def unlink_users(self, request, pk=None):  # pylint: disable=unused-argument
...
+
+        try:
+            models.EnterpriseCustomerUser.objects.unlink_user(
+                enterprise_customer=enterprise_customer, user_email=user_email, is_relinkable=is_relinkable
+            )
+        except (models.EnterpriseCustomerUser.DoesNotExist, models.PendingEnterpriseCustomerUser.DoesNotExist):
+            msg = "[UNLINK_SELF] User with email {} does not exist in enterprise {}.".format(
+                user_email, enterprise_customer
+            )
+            LOGGER.warning(msg)
```
iloveagent57, line 513: It might be good to respond with an error status here, like `return Response(status=422_UNPROCESSABLE_ENTITY)` and maybe include the exception string, too.

enterprise/api/v1/views/enterprise_customer.py
```diff
@@ -487,3 +487,32 @@ def unlink_users(self, request, pk=None):  # pylint: disable=unused-argument
...
+
+        serializer.is_valid(raise_exception=True)
+        enterprise_customer = self.get_object()
+        user_email = serializer.data.get('user_email')
+        is_relinkable = serializer.data.get('is_relinkable', True)
+
+        try:
+            models.EnterpriseCustomerUser.objects.unlink_user(
+                enterprise_customer=enterprise_customer, user_email=user_email, is_relinkable=is_relinkable
+            )
```
iloveagent57, line 508: We need to have this view only be able to unlink the user making the request.  So I don't think we need `user_email` in the request payload. Instead, we can use `request.user.email` as the value to pass into `unlink_user()`.

iloveagent57: I see you're basically enforcing that in the `IsUseremailInRequest` permission class. I think you can get rid of that permission class and just go with 
```
models.EnterpriseCustomerUser.objects.unlink_user(
    enterprise_customer=enterprise_customer, user_email=request.user.email, is_relinkable=is_relinkable
)
```
here in the view.

iloveagent57 (commented):
Good idea! Just two small suggestions.

enterprise/api/v1/serializers.py
```diff
@@ -1653,6 +1653,21 @@ class EnterpriseCustomerUnlinkUsersSerializer(serializers.Serializer):
     )
 
 
+class EnterpriseCustomerUnlinkSelfUserSerializer(serializers.Serializer):
```
iloveagent57, line 1656: We can just drop this serializer now.

enterprise/api/v1/views/enterprise_customer.py
```diff
@@ -487,3 +487,27 @@ def unlink_users(self, request, pk=None):  # pylint: disable=unused-argument
...
+    def unlink_self(self, request, pk=None):  # pylint: disable=unused-argument
+        """
+        Unlink request user from the enterprise.
+        """
+        user_email = request.user.email
+        enterprise_customer = self.get_object()
+
+        try:
+            models.EnterpriseCustomerUser.objects.unlink_user(
+                enterprise_customer=enterprise_customer, user_email=user_email, is_relinkable=True
```
irfanuddinahmad, line 501: do we want to never relink?

irfanuddinahmad (approved):

iloveagent57 (approved):

PR #2286, opened 2024-11-13
feat: update EnterpriseGroupMembershipSerializer to include enrollment count

katrinan029:
# Description
Adds enrollment count to the EnterpriseGroupMembershipSerializer to provide data to [data table](https://www.figma.com/design/eMWRd2jVa4Jqc6y7qrUfaa/Groups-Phase-2?node-id=917-55910&t=EcyGhyLrvBat6SkF-1) in the groups detail page data 

https://2u-internal.atlassian.net/browse/ENT-9525

# Test
- Visit url `http://localhost:18000/enterprise/api/v1/enterprise-group/<group_uuid>/learners` and verify that the enrollments field exists. 
- The group uuid can be found here http://localhost:18000/admin/enterprise/enterprisegroup/. Create a group if one doesn't already exist and add members to that group here http://localhost:18000/admin/enterprise/enterprisegroupmembership/ 

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -4692,13 +4692,17 @@ def _get_filtered_ecu_ids(self, user_query):
...
-            select entcu.id from enterprise_enterprisecustomeruser entcu
-            join auth_user au on entcu.user_id = au.id
-            where entcu.enterprise_customer_id = %s and au.email like %s;
+            with users as (
+                select ecu.id, au.email, coalesce(aup.name, concat(au.first_name, ' ', au.last_name)) as full_name
+                from enterprise_enterprisecustomeruser ecu
+                inner join auth_user au on ecu.user_id = au.id
+                left join auth_userprofile aup on au.id = aup.user_id
+                where ecu.enterprise_customer_id = %s
+            ) select * from users where email like %s or full_name like %s;
```
bcitro, line 4701: in the previous query we were only selecting the `id` column from `enterprise_enterprisecustomeruser`, should we still be doing that? (instead of `select *`)

enterprise/models.py
```diff
@@ -4692,13 +4692,17 @@ def _get_filtered_ecu_ids(self, user_query):
         # https://docs.djangoproject.com/en/5.0/topics/security/
         var_q = f"%{user_query}%"
         sql_string = """
-            select entcu.id from enterprise_enterprisecustomeruser entcu
-            join auth_user au on entcu.user_id = au.id
-            where entcu.enterprise_customer_id = %s and au.email like %s;
+            with users as (
+                select ecu.id, au.email, coalesce(aup.name, concat(au.first_name, ' ', au.last_name)) as full_name
```
kiram15, line 4696: Also I was testing on stage with a combination of set profile.name and user.first_name and last_name and I noticed a problem with how this query is working. It fails the test case of having to fall back to first/last from auth_user and never does, but I fixed it with this NULLIF function.                     
```coalesce(NULLIF(aup.name, ''), concat(au.first_name, ' ', au.last_name)) as full_name```

File: tests/test_enterprise/api/test_views.py [hunk 3]
```diff
@@ -8372,64 +8373,6 @@ def test_list_learners_bad_sort_by(self):
...
-            enterprise_customer_user=None,
-        )
-        existing_membership = EnterpriseGroupMembershipFactory(
-            group=group,
-            pending_enterprise_customer_user=None,
-            enterprise_customer_user__enterprise_customer=self.enterprise_customer,
-        )
-        existing_user = existing_membership.enterprise_customer_user.user
-        # Changing email to something that we know will be unique for collision purposes
-        existing_user.email = "ayylmao@example.com"
```
kiram15, line 8408: cali boy alex started ayylmao it needs to live on

tests/test_enterprise/api/test_views.py
```diff
@@ -8663,22 +8608,6 @@ def test_successful_list_with_filters(self):
...
-        Test that we properly sanitize member user query filters
-        https://xkcd.com/327/
-        """
-        # url: 'http://testserver/enterprise/api/v1/enterprise_group/<group uuid>/learners/'
-        url = settings.TEST_SERVER + reverse(
-            'enterprise-group-learners',
-            kwargs={'group_uuid': self.group_1.uuid},
-        )
-        # The problematic child
-        filter_query_param = "?user_query=Robert`); DROP TABLE enterprise_enterprisecustomeruser;--"
```
kiram15, line 8677: also this bobby tables joke has to be in platform too

katrinan029: yep! moving both tables to platform

kiram15 (approved):
LGTM!

PR #2287, opened 2024-11-14
fix: ensure correct `admin_users` in EnterpriseCustomerUserViewSet response

adamstankiewicz:
https://2u-internal.atlassian.net/browse/ENT-9741

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -721,24 +721,11 @@ class EnterpriseCustomerUserReadOnlySerializer(serializers.ModelSerializer):
         )
 
     user = UserSerializer()
+    enterprise_customer = serializers.SerializerMethodField()
     data_sharing_consent_records = serializers.SerializerMethodField()
     groups = serializers.SerializerMethodField()
     role_assignments = serializers.SerializerMethodField()
     enterprise_group = serializers.SerializerMethodField()
-
-    def to_representation(self, instance):
-        """
-        Override to pass the instance of `enterprise_customer` to the nested serializer.
-        """
-        representation = super().to_representation(instance)
-        enterprise_customer_instance = instance.enterprise_customer
-        enterprise_customer_serializer = EnterpriseCustomerSerializer(
-            instance=enterprise_customer_instance,
-            context=self.context
-        )
-        representation['enterprise_customer'] = enterprise_customer_serializer.data
-
-        return representation
 
     def _get_role_assignments_by_ecu_id(self, enterprise_customer_users):
         """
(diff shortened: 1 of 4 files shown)


enterprise/api/v1/serializers.py
```diff
@@ -721,12 +721,25 @@ class Meta:
         )
 
     user = UserSerializer()
-    enterprise_customer = EnterpriseCustomerSerializer()
     data_sharing_consent_records = serializers.SerializerMethodField()
     groups = serializers.SerializerMethodField()
     role_assignments = serializers.SerializerMethodField()
     enterprise_group = serializers.SerializerMethodField()
 
+    def to_representation(self, instance):
```
iloveagent57, line 729: could you do this via a `SerializerMethodField()` instead?  It'd be a little more explicit, and it would help with auto-discovery of fields in tools like drf-yasg or drf-spectacular.

iloveagent57 (approved):
Looks good, one suggestion.  Was the problem that the context wasn't being passed along?

enterprise/api/v1/serializers.py
adamstankiewicz: I did try this originally, but seemed to run into an issue where it would actually end up breaking the `enterprise-customer` list/detail APIs (insofar as the tests go), regarding the `enterprise_notification_banner` suddenly starting to return `None` instead of the current empty state dict.

I will try again, though, as I'm blanking if I was properly passing the `enterprise_customer` explicitly through an `instance` kwarg or not.

The `to_representation` was a pivot away from `SerializerMethodField`. I agree it would be preferred; will update soon.

adamstankiewicz:
> Was the problem that the context wasn't being passed along?

More that the `instance` is `None`. The `self.context` was seemingly needed for `get_enterprise_notification_banner`:

```python
def get_enterprise_notification_banner(self, obj):
        """
        Return the notification text if exist OR None
        """

        try:
            # this serializer is also called from tpa_pipeline and request is not available in the first place there.
            request = self.context['request']
            user_id = request.user.id
            LOGGER.error('[Admin Notification API] Get enterprise notification banner request object not found,'
                         ' Enterprise Customer :{} Exception: {}'.format(obj.slug, exc))
            return None
```

enterprise/api/v1/serializers.py
adamstankiewicz: Updated to use `SerializerMethodField`. When I originally tried it, I wasn't passing `context=self.context`, but ensuring its passed works fine!

pwnage101 (approved):

PR #2298, opened 2024-12-04
fix: use single constant for course mode priority; only fetch catalog list with resolved course run

adamstankiewicz:
### Description

1. During stage QA, it was observed that a course run with only an `honor` seat was not enrollable via the LMS bulk enrollment API when realizing default enterprise enrollment intentions. It was determined that this was due to relying on a constant `BEST_MODE_ORDER` that didn't include audit/honor seat modes; as such, when trying to enroll in this `honor`-only course run, it was falling back to attempt to enroll in a non-existent `audit` seat, resulting in a 409 on the LMS bulk enrollment API. This PR removes `BEST_MODE_ORDER` to singularly rely on the existing `COURSE_MODE_SORT_ORDER` instead, which includes `honor`.
2. When calling the `/enterprise/api/v1/default-enterprise-enrollment-intentions/learner-status/` LMS API locally with a default enrollment intention configured for course with a non-existent advertised course run, the `DefaultEnterpriseEnrollmentIntention` model continues to attempt to call `contains_content_items` to retrieve a list of applicable catalog UUIDs associated with a resolved course run. However, if a course run cannot be resolved, we shouldn't be making the call to enterprise-catalog, and simply fallback to an empty list in the model's `applicable_enterprise_catalog_uuids` property.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/constants.py
+++ b/enterprise/constants.py
@@ -59,6 +59,7 @@ class CourseModes:
     VERIFIED = 'verified'
     UNPAID_EXECUTIVE_EDUCATION = 'unpaid-executive-education'
 
+
 # Course mode sorting based on slug
 COURSE_MODE_SORT_ORDER = [
     CourseModes.VERIFIED,
--- a/enterprise/constants.py
+++ b/enterprise/constants.py
@@ -59,14 +59,6 @@ class CourseModes:
     VERIFIED = 'verified'
     UNPAID_EXECUTIVE_EDUCATION = 'unpaid-executive-education'
 
-
-BEST_MODE_ORDER = [
-    CourseModes.VERIFIED,
-    CourseModes.PROFESSIONAL,
-    CourseModes.NO_ID_PROFESSIONAL,
-    CourseModes.UNPAID_EXECUTIVE_EDUCATION,
-]
-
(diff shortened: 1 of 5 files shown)


enterprise/constants.py
```diff
@@ -60,13 +60,6 @@ class CourseModes:
     UNPAID_EXECUTIVE_EDUCATION = 'unpaid-executive-education'
 
 
-BEST_MODE_ORDER = [
-    CourseModes.VERIFIED,
-    CourseModes.PROFESSIONAL,
-    CourseModes.NO_ID_PROFESSIONAL,
-    CourseModes.UNPAID_EXECUTIVE_EDUCATION,
-]
```
adamstankiewicz, line 68: [inform/context] This removed constant has the same ordering as `COURSE_MODE_SORT_ORDER`, save for `CourseModes.UNPAID_EXECUTIVE_EDUCATION`. However, I believe this is OK as course runs with a seat mode of `CourseModes.UNPAID_EXECUTIVE_EDUCATION` do not have any seats with mode of `CourseModes.AUDIT` or `CourseModes.HONOR`.

PR #2301, opened 2024-12-10
fix: unenrolled defaults are now not enrollable

iloveagent57:
The default enrollment ``learner_status`` view now considers intended courses from which the requested user has unenrolled as no-longer realizable.
ENT-9839

**Merge checklist:**

**Post merge:**


tests/test_enterprise/api/test_default_enrollment_views.py
```diff
@@ -0,0 +1,726 @@
...
+
+    @ddt.data(
+        {'run_is_enrollable': False, 'unenrollment_exists': False},
+        {'run_is_enrollable': True, 'unenrollment_exists': True},
+    )
+    @mock.patch('enterprise.content_metadata.api.EnterpriseCatalogApiClient')
+    @mock.patch('enterprise.models.utils.get_best_mode_from_course_key')
+    @mock.patch.object(EnterpriseCourseEnrollment, 'course_enrollment', new_callable=mock.PropertyMock)
+    @ddt.unpack
+    def test_default_enrollment_intentions_learner_status_content_not_enrollable(
```
iloveagent57, line 285: This is the only changed test, everything else has just been refactored into this new file.

enterprise/api/v1/views/default_enterprise_enrollments.py
```diff
@@ -211,6 +211,12 @@ def _partition_default_enrollment_intentions_by_enrollment_status(
                 already_enrolled.append((intention, non_audit_enrollment))
                 continue
 
+            if non_audit_enrollment and non_audit_enrollment.unenrolled:
```
adamstankiewicz, line 214: [sanity check] If the existing enrollment is audit (vs. the `non_audit_enrollment` here), the learners unenrolls, and there's non-audit modes that exist, would we also want to prevent automatic re-enrollment, too?

For example, would the conditions below the `NOTE` comment also need to take into account `audit_enrollment.unenrolled`?

tests/test_enterprise/api/test_default_enrollment_views.py
adamstankiewicz: Good call to start breaking up the giant `test_views` file :)

tests/test_enterprise/api/test_default_enrollment_views.py
```diff
@@ -0,0 +1,726 @@
...
+from rest_framework.reverse import reverse
+
+from django.conf import settings
+
+from enterprise.constants import ENTERPRISE_LEARNER_ROLE
+from enterprise.models import EnterpriseCourseEnrollment
+from test_utils import FAKE_UUIDS, TEST_PASSWORD, factories, fake_catalog_api
+
+from .constants import AUDIT_COURSE_MODE, VERIFIED_COURSE_MODE
+from .test_views import BaseTestEnterpriseAPIViews, create_mock_default_enterprise_enrollment_intention
```
adamstankiewicz, line 21: nit: should `create_mock_default_enterprise_enrollment_intention` be moved into this file, if it's only related to default enrollment views?

adamstankiewicz (commented):
A couple small nits, with a broader sanity check question about the logic considering `unenrolled` for audit enrollments.

tests/test_enterprise/api/test_default_enrollment_views.py
iloveagent57: I tried to do that, but the test for default realizations still lives in `test_views.py` under the bulk enrollment test class, and it needs this function, so I figured it makes more sense to keep it in the "base" test file and import it in the more specific test file.

adamstankiewicz: Ah, got it. Missed that it was still used for default realizations!

enterprise/api/v1/views/default_enterprise_enrollments.py
iloveagent57: I thought about that but decided against it, I can leave a comment in the code. My reasoning is that default enrollments should be "exactly once" per (user, enterprise), if possible. If they enrolled in audit and unenrolled, it means they likely never had a default intention realized for them. This isn't a strongly-held position though.

adamstankiewicz: > If they enrolled in audit and unenrolled, it means they likely never had a default intention realized for them.

Ah yeah, this probably represents the source of most audit enrollments. The exception would be for courses that are audit-only, which likely rarely happens in production; it does happen somewhat frequently on stage, though we shouldn't optimize for bad stage data 😉 

Let's leave it as is for now, with a comment!

adamstankiewicz (approved):

PR #2314, opened 2025-01-14
fix: enrollment intention saves should not be blocked on catalog inclusion

pwnage101:
Before this commit, normal replication delays from discovery->catalog can easily result in enrollment intentions being un-saveable when simultaneously modifying catalog definitions to include the content. Conversely, normal replication delays can also allow saving an enrollment intention that would eventually become invalid due to content being removed from the catalog, creating a false sense of correctness.

After this commit, we no longer assert during intention save() that the customer's catalogs actually contain the content. Instead, we assert that the content exists at all in the enterprise-catalog database by querying the customer- and catalog-agnostic
`/api/v1/content-metadata/?content_identifiers=<content_key>` endpoint.

Subsequent commits/PRs may introduce dynamic warning messaging (non-blocking) on django admin pages for the customer and intention when the intention's content key is not contained in the customer's catalogs.

ENT-9840

### DIFF
--- a/docs/decisions/0015-default-enrollments.rst
+++ b/docs/decisions/0015-default-enrollments.rst
@@ -65,7 +65,7 @@ will always primarily be associated with the course run associated with the ``De
 * If the content_key is a specific course run, we'll always try to enroll in the explicitly
   configured course run key (not the advertised course run).
 
-This content_key will be passed to the ``EnterpriseCatalogApiClient().get_content_metadata_content_identifier``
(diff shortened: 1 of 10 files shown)


pwnage101:
Unfortunately, the API that I switched to (`/api/v1/content-metadata/?content_identifiers=<>`) doesn't automatically return a top-level course when given a course run key.  This was not caught within existing unit tests because the assumption is baked into the mocks.  I'll need to make a few more tweaks and put it back up for review.

iloveagent57 (approved):

adamstankiewicz (approved):

PR #2321, opened 2025-01-25
feat: new mgmt command to validate default enrollment intentions

pwnage101:
Introduced validate_default_enrollment_intentions.py to check that all DefaultEnterpriseEnrollmentIntention objects have a valid content_key which actually belongs to at least one of the related customer's catalogs.

ENT-9941

### DIFF
--- a/enterprise/management/commands/validate_default_enrollment_intentions.py
+++ b/enterprise/management/commands/validate_default_enrollment_intentions.py
@@ -1,0 +1,121 @@
+"""
+Django management command to validate that DefaultEnterpriseEnrollmentIntention
+objects have enrollable content.
+"""
+import logging
+from datetime import timedelta
+
+from django.core.management import BaseCommand, CommandError
+from django.db.models import Max
+from django.db.models.functions import Greatest
+from django.utils import timezone
+
+from enterprise.content_metadata.api import get_and_cache_customer_content_metadata
+from enterprise.models import DefaultEnterpriseEnrollmentIntention
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Enumerate the catalog filters and log information about how we might migrate them.
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.delay_minutes = None
+        super().__init__(*args, **kwargs)
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--delay-minutes',
+            dest='delay_minutes',
+            required=False,
+            type=int,
+            default=30,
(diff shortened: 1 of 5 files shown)


enterprise/management/commands/validate_default_enrollment_intentions.py
```diff
@@ -0,0 +1,141 @@
...
+        super().__init__(*args, **kwargs)
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--delay-minutes',
+            dest='delay_minutes',
+            required=False,
+            type=int,
+            default=30,
+            help="How long after a customer's catalog has been updated are we allowed to evaluate the customer."
```
pwnage101, line 35: 30 minutes is a wild guess, I'll probably take a look at some real life task durations and come up with something more accurate.

enterprise/management/commands/validate_default_enrollment_intentions.py
```diff
@@ -0,0 +1,140 @@
...
+        self.delay_minutes = options.get("delay_minutes")
+
+        customers = EnterpriseCustomer.objects.annotate(
+            catalogs_modified_latest=Greatest(
+                Max("enterprise_customer_catalogs__modified"),
+                Max("enterprise_customer_catalogs__enterprise_catalog_query__modified"),
+            ),
+        ).prefetch_related(
+            "default_enrollment_intentions",
+        )
```
iloveagent57, line 103: Will this fetch all enterprise customer records? Can we do it in "reverse" by first identifying only the customers that have any default enrollment intentions?

iloveagent57 (commented):
Looks great, only suggestion is to do it without fetching all EnterpriseCustomers.

enterprise/management/commands/validate_default_enrollment_intentions.py
```diff
@@ -0,0 +1,146 @@
...
+        result['invalid_intentions'] = [intention for intention, valid in results.items() if not valid]
+        return result
+
+    def handle(self, *args, **options):
+        self.delay_minutes = options.get("delay_minutes")
+
+        customers_with_intentions = DefaultEnterpriseEnrollmentIntention.objects.values_list(
+            "enterprise_customer__uuid",
+        )
+        customers_with_intentions = [values[0] for values in customers_with_intentions]
```
iloveagent57, line 99: FYI you can pass in `flat=True` to values list if you're only selecting a single column, and that will give you back singles values instead of 1-tuples.

iloveagent57 (approved):
Looks great, don't forget version bump!

PR #2324, opened 2025-01-31
fix: gracefully handle case where intention has no related catalogs

pwnage101:
ENT-9941


iloveagent57 (approved):
LGTM, remember the version bump.

PR #2339, opened 2025-02-19
Refactor  add foreign key to ecu part 2

jesperhodge:
## Merge and deploy part 1 https://github.com/openedx/edx-enterprise/pull/2338 first!!

Then change the target branch for this PR to master.

## Description

See https://github.com/openedx/edx-enterprise/pull/2338

Part 2 of adding a foreign key to ECU table.
This adds an index to the new field.

**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -1066,7 +1066,7 @@ class EnterpriseCustomerUser(TimeStampedModel):
     user_fk = models.IntegerField(
         null=True,
         blank=False,
-        db_index=False,
+        db_index=True,
```
iloveagent57, line 1069: with this change, we might also need to specify in the `Meta` class for this model the name of the index to use on this field: https://docs.djangoproject.com/en/5.1/ref/models/options/#django.db.models.Options.indexes
it would be something like
```python
    class Meta:
        indexes = [
            models.Index(fields=["user_fk"], name="idx_enterprise_user_fk"),
        ]
```

jesperhodge:
Test coverage not relevant since we don't test migrations

enterprise/migrations/0230_alter_enterprisecustomeruser_user_fk_and_more.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('enterprise', '0229_enterprisecustomeruser_user_fk_and_more'),
+    ]
+
+    db_engine = connection.settings_dict['ENGINE']
+    if 'sqlite3' in db_engine:
+        operations = []
```
iloveagent57, line 13: Will we want this index to exist in sqlite3?

iloveagent57: Actually, I guess we don't really care, since it would just be for unit tests.

enterprise/migrations/0229_enterprisecustomeruser_user_fk_and_more.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2.17 on 2025-02-25 22:38
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
iloveagent57, line 6: This migration already exists in the `master` branch now, right? Does this branch need a rebase?

iloveagent57 (approved):

PR #2341, opened 2025-02-21
Refactor  add foreign key to ecu part 3

jesperhodge:
Part 3 of #2338 .

Should be merged after other two parts are merged and deployed.

To Do:
Edit save() method or something to make sure any record added after merge will add the correct value to `user_fk`.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/management/commands/backfill_ecu_table_user_foreign_key.py
+++ b/enterprise/management/commands/backfill_ecu_table_user_foreign_key.py
@@ -16,6 +16,50 @@ log = logging.getLogger(__name__)
 User = auth.get_user_model()
 
 
+def safe_bulk_update(entries, properties, max_retries):
+    """Performs bulk_update with retry logic."""
+    for attempt in range(1, max_retries + 1):
+        try:
+            with transaction.atomic():
+                EnterpriseCustomerUser.objects.bulk_update(
+                    entries, properties
+                )
+            return len(entries)
+        except DatabaseError as e:
+            wait_time = 2 ** attempt
+            log.warning(f"Attempt {attempt}/{max_retries} failed: {e}. Retrying in {wait_time}s.")
+            sleep(wait_time)
+    raise Exception(f"Bulk update failed after {max_retries} retries.")
+
+
+def _fetch_and_update_in_batches(queryset, batch_limit, batch_sleep, max_retries):
+    """
+    Fetches and updates records in batches.
+    Only loads and updates a subset of records at a time to avoid memory and performance issues.
+    """
+    total_processed = 0
+    batch = []
+    for ecu in queryset.iterator(chunk_size=batch_limit):
+        ecu.user_fk = ecu.user_id
+        batch.append(ecu)
+
+        if len(batch) >= batch_limit:
+            safe_bulk_update(batch, ['user_fk'], max_retries)
+
+            total_processed += len(batch)
+            log.info(f"Processed {total_processed} records.")
+
+            batch = []  # Reset batch
+            sleep(batch_sleep)  # Avoid DB overload
+
+    # Process any remaining rows (last batch)
+    if batch:
+        count = safe_bulk_update(batch, ['user_fk'], max_retries)
+        total_processed += count
+        log.info(f"Final batch processed. Total {total_processed} records updated.")
+    return True
+
+
 class Command(BaseCommand):
     """
     Management command to copy `user_id` values to the foreign key `user_fk` field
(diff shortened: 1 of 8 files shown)


tests/test_enterprise/management/test_backfill_ecu_table_user_foreign_key.py
```diff
@@ -0,0 +1,129 @@
...
+    def test_copies_user_id_to_user_fk(self):
+        ecu = EnterpriseCustomerUser.objects.first()
+        ecu.user_fk = None
+
+        # use bulk_update to prevent save() method from setting user_fk to user_id
+        EnterpriseCustomerUser.objects.all().bulk_update([ecu], ['user_fk'])
+        ecu.refresh_from_db()
+        assert ecu.user_fk is None
+        call_command(self.command)
+        assert EnterpriseCustomerUser.objects.first().user_fk is EnterpriseCustomerUser.objects.first().user_id
```
iloveagent57, line 73: nit: another way to do this is to explicitly refresh the `ecu` model instance from the DB, and then assert that the user FK == the user id:
```python
ecu.refresh_from_db()
assert ecu.user_fk == ecu.user_id
```

tests/test_enterprise/management/test_backfill_ecu_table_user_foreign_key.py
```diff
@@ -0,0 +1,129 @@
...
+        are fetched in chunks.
+        This test ensures that the QuerySet is not evaluated before being passed to `_fetch_and_update_in_batches`.
+        """
+        call_command(self.command, batch_limit=3)
+
+        _, kwargs = mock_fetch_and_update.call_args
+        queryset = kwargs.get("queryset")
+
+        # Check that QuerySet was NOT evaluated before `_fetch_and_update_in_batches`
+        assert queryset._result_cache is None, "QuerySet was evaluated too early!"  # pylint: disable=protected-access
```
iloveagent57, line 129: The reason this test works is that the unit tests use a sqlite3 database, and not mysql.

enterprise/management/commands/backfill_ecu_table_user_foreign_key.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def _fetch_and_update_in_batches(queryset, batch_limit, batch_sleep, max_retries):
+    """
+    Fetches and updates records in batches.
+    Only loads and updates a subset of records at a time to avoid memory and performance issues.
+    """
+    total_processed = 0
+    batch = []
+    for ecu in queryset.iterator(chunk_size=batch_limit):
```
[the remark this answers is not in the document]
jesperhodge: done

tests/test_enterprise/management/test_backfill_ecu_table_user_foreign_key.py
jesperhodge: done

tests/test_enterprise/management/test_backfill_ecu_table_user_foreign_key.py
jesperhodge: Then I'm removing the test. Otherwise it will only produce false positives.

enterprise/management/commands/backfill_ecu_table_user_foreign_key.py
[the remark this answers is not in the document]
jesperhodge: I refactored it to enterprise because I want to avoid duplication, in case someone fixes or changes something.

enterprise/management/commands/backfill_ecu_table_user_foreign_key.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+
+    def backfill_ecu_table_user_foreign_key(self, options):
+        """
+        Backfills user_fk from user_id in batches with timeout, retries, and progress reporting.
+        """
+        batch_limit = options.get('batch_limit', 250)
+        max_retries = options.get('max_retries', 5)
+        batch_sleep = options.get('batch_sleep', 2)
+        queryset = EnterpriseCustomerUser.objects.filter(user_fk__isnull=True)
```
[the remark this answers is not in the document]
jesperhodge: done

enterprise/management/commands/backfill_ecu_table_user_foreign_key.py
```diff
@@ -0,0 +1,126 @@
...
+def _fetch_and_update_in_batches(queryset, batch_limit, batch_sleep, max_retries):
+    """
+    Fetches and updates records in batches.
+    Only loads and updates a subset of records at a time to avoid memory and performance issues.
+    Note: you cannot use django's queryset.iterator() method
+    as MySQL does not support it and will still load everything into memory.
+    """
+    batch_counter = 1
+    total_processed = 0
+    for batch in batch_by_pk(EnterpriseCustomerUser, extra_filter=queryset.query.where, batch_size=batch_limit):
```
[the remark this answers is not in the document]
jesperhodge: done

integrated_channels/utils.py
```diff
@@ -430,7 +430,8 @@ def batch_by_pk(ModelClass, extra_filter=Q(), batch_size=10000):
         for item in items_batch:
             ...
     """
-    qs = ModelClass.objects.filter(extra_filter).order_by('pk')[:batch_size]
+    model_manager = model_manager or ModelClass.objects
+    qs = model_manager.filter(extra_filter).order_by('pk')[:batch_size]
```
iloveagent57, line 434: We need to update line 441 below to be `qs = model_manager.filter(....)`.

enterprise/management/commands/backfill_ecu_table_user_foreign_key.py
```diff
@@ -0,0 +1,132 @@
...
+log = logging.getLogger(__name__)
+User = auth.get_user_model()
+
+
+def safe_bulk_update(entries, properties, max_retries):
+    """Performs bulk_update with retry logic."""
+    for attempt in range(1, max_retries + 1):
+        try:
+            with transaction.atomic():
+                EnterpriseCustomerUser.objects.bulk_update(
```
iloveagent57, line 26: This also needs to be `all_objects` (I just noticed this now from testing out with SQL logging enabled).

tests/test_enterprise/management/test_backfill_ecu_table_user_foreign_key.py
```diff
@@ -0,0 +1,114 @@
...
+from test_utils import factories
+
+EXCEPTION = "DUMMY_TRACE_BACK"
+
+User = auth.get_user_model()
+
+
+@mark.django_db
+@ddt.ddt
+class CreateEnterpriseCourseEnrollmentCommandTests(TestCase):
```
iloveagent57, line 27: It would be good to have a test case with an ECU instance that has `linked=False` so we can ensure that the command really is hitting all the rows of the table.

iloveagent57 (approved):
Looks great, nice work! Don't forget to bump the changelog/version when you squash.

PR #2364, opened 2025-03-21
fix: Don't raise a warning when Braze isn't installed

kdmccormick:
Most Open edX instances will not have Braze installed. Currently those instances will all see these warnings every time any edx-platform code is loaded, even for management commands, pylint, and mypy:

    [enterprise.api_client.braze] - Cannot import BrazeClient, calls to Braze will not work
    [enterprise.tasks] - Cannot import BrazeClientError

We're removing those to help reduce noise for the majority of operators.
____________________________

@iloveagent57 , I imagine you guys want to be able to tell when braze doesn't load properly. The problem with logging at the top level of the module is that everybody sees it, even if they don't use enterprise or braze at all. Is there some enterprise-specific or braze-specific utility function you could log the warning from instead? That way, operators will only see it if they're trying to us a feature that might need braze.

### DIFF
--- a/enterprise/api_client/braze.py
+++ b/enterprise/api_client/braze.py
@@ -10,7 +10,6 @@ logger = logging.getLogger(__name__)
 try:
     from braze.client import BrazeClient
 except ImportError:
-    logger.warning('Cannot import BrazeClient, calls to Braze will not work')
     BrazeClient = None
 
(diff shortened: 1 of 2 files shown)


iloveagent57:
We can introduce a `get_braze_client()` function that does the import and logs a warning if not importable, similarly to what edx-platform does.

kdmccormick:
SGTM 👍🏻

PR #2367, opened 2025-03-28
fix: assign_learners use case insensitive lookup with received email addresses

marlonkeating:
[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-10246)

Currently we allow learner emails to be entered via csv file, and do not restrict them to the email casing they registered with.  As such, we need to support looking up existing `User` records in a case-insensitive manner.

## Testing Instructions
- Deploy this change locally
- Using Admin Portal, invite a learner to a group via csv, where their email in the csv does *not* match the casing their email was originally registered with
- Verify the learner is invited to the group successfully, and no duplicate `PendingEnterpriseCustomerUser` records are created in the process

**Merge checklist:**

**Post merge:**

### DIFF
--- a/tests/test_enterprise/api/test_views.py
+++ b/tests/test_enterprise/api/test_views.py
@@ -8809,6 +8809,31 @@ class TestEnterpriseGroupViewSet(APITest):
         assert pending_membership.status == GROUP_MEMBERSHIP_PENDING_STATUS
         assert membership.status == GROUP_MEMBERSHIP_ACCEPTED_STATUS
 
+    # TODO: Figure out how to make sqlite mirror MySQL case-insensitivity for IN queries
+    # def test_assign_learners_case_insensitive_emails(self):
+    #     """
+    #     Test assigning learners to a group with emails whose case differs from their registered user emails
+    #     """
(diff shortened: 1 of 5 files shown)


tests/test_enterprise/api/test_views.py
```diff
@@ -8809,6 +8809,31 @@ def test_assign_learners_revives_previously_removed_members(self):
         assert pending_membership.status == GROUP_MEMBERSHIP_PENDING_STATUS
         assert membership.status == GROUP_MEMBERSHIP_ACCEPTED_STATUS
 
+    # TODO: Figure out how to make sqlite mirror MySQL case-insensitivity for IN queries
```
marlonkeating, line 8812: The following line does a case insensitive lookup under MySQL, but does not under the database engine our unit tests use, causing this unit test to fail.  
https://github.com/openedx/edx-enterprise/blob/db8ac49972b490a8dd0d88c24d9576cba4a54379/enterprise/api/v1/views/enterprise_group.py#L313
While alternative lookup schemes exist, the ones I've found that are database-independent would incur a performance hit of some kind.

PR #2369, opened 2025-04-03
WIP: Abstract retrieval of Sap SuccessFactor information into one resource

zwidekalanga:
**Merge checklist:**

**Post merge:**

### DIFF
--- a/integrated_channels/api/v1/sap_success_factors/serializers.py
+++ b/integrated_channels/api/v1/sap_success_factors/serializers.py
@@ -26,3 +26,11 @@ class SAPSuccessFactorsConfigSerializer(EnterpriseCustomerPluginConfigSerializer
 
     encrypted_key = serializers.CharField(required=False, allow_blank=False, read_only=False)
     encrypted_secret = serializers.CharField(required=False, allow_blank=False, read_only=False)
+
+
+class SAPUserInfoRequestSerializer(serializers.Serializer):
+    """
+    Serializer for SAP user info request parameters.
+    """
+    org_id = serializers.CharField(required=True)
+    loggedinuserid = serializers.CharField(required=True)
--- a/integrated_channels/api/v1/sap_success_factors/views.py
+++ b/integrated_channels/api/v1/sap_success_factors/views.py
@@ -2,11 +2,14 @@
 Viewsets for integrated_channels/v1/sap_success_factors/
 """
 from rest_framework import exceptions, permissions, status, viewsets
+from rest_framework.decorators import api_view, permission_classes
+from rest_framework.response import Response
 
 from integrated_channels.api.v1.mixins import PermissionRequiredForIntegratedChannelMixin
+from integrated_channels.exceptions import ClientError
 from integrated_channels.sap_success_factors.models import SAPSuccessFactorsEnterpriseCustomerConfiguration
 
-from .serializers import SAPSuccessFactorsConfigSerializer
+from .serializers import SAPSuccessFactorsConfigSerializer, SAPUserInfoRequestSerializer
 
 
 class SAPSuccessFactorsConfigurationViewSet(PermissionRequiredForIntegratedChannelMixin, viewsets.ModelViewSet):
@@ -15,3 +18,59 @@ class SAPSuccessFactorsConfigurationViewSet(PermissionRequiredForIntegratedChann
     permission_required = 'enterprise.can_access_admin_dashboard'
 
     configuration_model = SAPSuccessFactorsEnterpriseCustomerConfiguration
+
+
+@api_view(['GET'])
+@permission_classes([permissions.IsAuthenticated])
+def retrieve_additional_userinfo(request):
+    """
+    Retrieve additional user information from SAP SuccessFactors.
+    
+    GET params:
+        org_id: Organization ID in SAP
+        loggedinuserid: ID of the logged-in user
+    
+    Returns:
+        User information retrieved from SAP SuccessFactors.
+    """
+    serializer = SAPUserInfoRequestSerializer(data=request.query_params)
+    if not serializer.is_valid():
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+    
+    org_id = serializer.validated_data['org_id']
+    logged_in_user_id = serializer.validated_data['loggedinuserid']
+    
+    # Find the SAP configuration for this organization
+    try:
+        # Use the company ID (org_id) to find the right configuration
+        enterprise_config = SAPSuccessFactorsEnterpriseCustomerConfiguration.objects.get(
+            sapsf_company_id=org_id,
+            active=True
+        )
+    except SAPSuccessFactorsEnterpriseCustomerConfiguration.DoesNotExist:
+        return Response(
+            {"error": f"No active SAP configuration found for organization ID: {org_id}"},
+            status=status.HTTP_404_NOT_FOUND
+        )
+    
+    # Create the client and retrieve user information
+    from integrated_channels.sap_success_factors.client import SAPSuccessFactorsAPIClient
+    client = SAPSuccessFactorsAPIClient(enterprise_config)
+    
+    try:
+        user_data = client.get_user_details(logged_in_user_id)
+        
+        # Return specifically formatted data to match Auth0's expected structure
+        response_data = {
+            "email": user_data.get('d', {}).get('email'),
+            "given_name": user_data.get('d', {}).get('firstName'),
+            "surname": user_data.get('d', {}).get('lastName'),
+            "full_data": user_data.get('d', {})
+        }
+        
+        return Response(response_data)
+    except ClientError as e:
+        return Response(
+            {"error": str(e)},
+            status=status.HTTP_500_INTERNAL_SERVER_ERROR
+        )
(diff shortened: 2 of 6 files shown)


integrated_channels/api/v1/sap_success_factors/views.py
```diff
@@ -15,3 +18,59 @@ class SAPSuccessFactorsConfigurationViewSet(PermissionRequiredForIntegratedChann
     permission_required = 'enterprise.can_access_admin_dashboard'
 
     configuration_model = SAPSuccessFactorsEnterpriseCustomerConfiguration
+
+
+@api_view(['GET'])
+@permission_classes([permissions.IsAuthenticated])
```
sameenfatima78, line 24: Since it's a sensitive and critical logging flow, I'd suggest adding a more specific permission so we can restrict it's access to other users. Only staff users and Auth0 should be able to call this endpoint. We might need a custom permission class here. Thoughts, @zamanafzal ?

integrated_channels/api/v1/sap_success_factors/views.py
```diff
@@ -15,3 +18,59 @@ class SAPSuccessFactorsConfigurationViewSet(PermissionRequiredForIntegratedChann
...
+def retrieve_additional_userinfo(request):
+    """
+    Retrieve additional user information from SAP SuccessFactors.
+    
+    GET params:
+        org_id: Organization ID in SAP
```
sameenfatima78, line 30: This value actually represents the organization ID in Auth0. The intended flow is to first fetch the [EnterpriseCustomer](https://github.com/openedx/edx-enterprise/blob/47f37480285e8c962d90556c9f71ad877d1c510b/enterprise/models.py#L174) using this org_id as "auth_org_id", and then retrieve the corresponding SAP SuccessFactors configuration for that customer. From there, we can extract the company_id:
https://github.com/openedx/edx-enterprise/blob/47f37480285e8c962d90556c9f71ad877d1c510b/integrated_channels/sap_success_factors/models.py#L90

integrated_channels/api/v1/sap_success_factors/views.py
```diff
@@ -15,3 +18,59 @@ class SAPSuccessFactorsConfigurationViewSet(PermissionRequiredForIntegratedChann
...
+    org_id = serializer.validated_data['org_id']
+    logged_in_user_id = serializer.validated_data['loggedinuserid']
+    
+    # Find the SAP configuration for this organization
+    try:
+        # Use the company ID (org_id) to find the right configuration
+        enterprise_config = SAPSuccessFactorsEnterpriseCustomerConfiguration.objects.get(
+            sapsf_company_id=org_id,
+            active=True
+        )
```
sameenfatima78, line 49: This is the block that we'll need to modify. Please look at my comment above.

integrated_channels/api/v1/sap_success_factors/views.py
```diff
@@ -15,3 +18,59 @@ class SAPSuccessFactorsConfigurationViewSet(PermissionRequiredForIntegratedChann
...
+    except SAPSuccessFactorsEnterpriseCustomerConfiguration.DoesNotExist:
+        return Response(
+            {"error": f"No active SAP configuration found for organization ID: {org_id}"},
+            status=status.HTTP_404_NOT_FOUND
+        )
+    
+    # Create the client and retrieve user information
+    from integrated_channels.sap_success_factors.client import SAPSuccessFactorsAPIClient
```
sameenfatima78, line 57: can we include this import at the top?

integrated_channels/api/v1/sap_success_factors/views.py
```diff
@@ -15,3 +18,59 @@ class SAPSuccessFactorsConfigurationViewSet(PermissionRequiredForIntegratedChann
...
+    try:
+        user_data = client.get_user_details(logged_in_user_id)
+        
+        # Return specifically formatted data to match Auth0's expected structure
+        response_data = {
+            "email": user_data.get('d', {}).get('email'),
+            "given_name": user_data.get('d', {}).get('firstName'),
+            "surname": user_data.get('d', {}).get('lastName'),
+            "full_data": user_data.get('d', {})
+        }
```
sameenfatima78, line 69: This can be written as this to make it more readable and avoid duplicate code. 
```python
user_info = user_data.get("d", {})
response_data = {
    "email": user_info.get("email"),
    "given_name": user_info.get("firstName"),
    "surname": user_info.get("lastName"),
    "full_data": user_info
}

File: integrated_channels/sap_success_factors/models.py [hunk 6]
```diff
@@ -408,3 +411,109 @@ def _payload_data(self):
...
+    )
+    
+    class Meta:
+        app_label = 'sap_success_factors'
+        verbose_name = _('SAP User Info Retrieval Audit')
+        verbose_name_plural = _('SAP User Info Retrieval Audits')
+        indexes = [
+            models.Index(fields=['enterprise_customer', 'created']),
+            models.Index(fields=['sapsf_user_id', 'created']),
+        ]
```
sameenfatima78, line 513: It's best to include explicit index names.
 
```python
from django.db.models import Index

indexes = [
    Index(fields=['enterprise_customer', 'created'], name='idx_customer_created'),
    Index(fields=['sapsf_user_id', 'created'], name='idx_sapsf_created'),
]

File: integrated_channels/api/v1/sap_success_factors/serializers.py [hunk 7]
```diff
@@ -26,3 +26,11 @@ class Meta:
...
     encrypted_key = serializers.CharField(required=False, allow_blank=False, read_only=False)
     encrypted_secret = serializers.CharField(required=False, allow_blank=False, read_only=False)
+
+
+class SAPUserInfoRequestSerializer(serializers.Serializer):
+    """
+    Serializer for SAP user info request parameters.
+    """
+    org_id = serializers.CharField(required=True)
+    loggedinuserid = serializers.CharField(required=True)
```
sameenfatima78, line 36: nit: Can we align the variable naming conventions here? The first uses underscores, while the second does not.

integrated_channels/sap_success_factors/client.py
```diff
@@ -373,6 +373,228 @@ def _call_post_with_session(self, url, payload):
             )
         return response.status_code, response.text
 
+    def get_user_details(self, logged_in_user_id):
```
sameenfatima78, line 376: This method is too long and can be broken down into smaller private methods. Maybe we can create a different method for each step?

integrated_channels/sap_success_factors/client.py
```diff
@@ -373,6 +373,228 @@ def _call_post_with_session(self, url, payload):
...
+        Makes three API calls to SAP:
+        1. Get SAML Assertion using IDP endpoint
+        2. Exchange for Access Token
+        3. Call OData User endpoint
+        
+        Args:
+            logged_in_user_id (str): The SAP user ID of the logged-in user
+            
+        Returns:
+            dict: User data from SAP including email, name, etc.
```
sameenfatima78, line 389: Can we include an example response here? This will help in testing/debugging.

integrated_channels/sap_success_factors/client.py
```diff
@@ -373,6 +373,228 @@ def _call_post_with_session(self, url, payload):
...
+        
+        transaction_log = SAPSuccessFactorsUserInfoRetrievalAudit.objects.create(
+            enterprise_customer=self.enterprise_configuration.enterprise_customer,
+            sapsf_user_id=logged_in_user_id
+        )
+        
+        try:
+            # STEP 1: Get SAML Assertion
+            idp_url = urljoin(self.enterprise_configuration.sapsf_base_url, "oauth/idp")
+            token_url = urljoin(self.enterprise_configuration.sapsf_base_url, "oauth/token")
```
sameenfatima78, line 403: These two can be added as model properties.

```python 
@property
def sapsf_idp_url(self):
    return urljoin(self.enterprise_configuration.sapsf_base_url, "oauth/idp")

File: integrated_channels/sap_success_factors/client.py [hunk 11]
```diff
@@ -373,6 +373,228 @@ def _call_post_with_session(self, url, payload):
...
+        
+        Args:
+            logged_in_user_id (str): The SAP user ID of the logged-in user
+            
+        Returns:
+            dict: User data from SAP including email, name, etc.
+        """
+        # Create a transaction log entry
+        from integrated_channels.sap_success_factors.models import SAPSuccessFactorsUserInfoRetrievalAudit
+        from django.utils import timezone
```
sameenfatima78, line 393: Can we include these imports at the top?

sameenfatima78 (changes requested):
Approach looks good. Just a couple of small suggestions for improvements.

sameenfatima78:
Discussed via Slack:

I think it does make sense to migrate the logic to enterprise, that's what we initially discussed too, if I remember correctly. The only challenge right now is that there's a heavy dependency on Integrated Channels because the credentials/configs live there and the logic is SAP specific.
As a way to break that dependency, what if we:
- Input all the required information directly into the self-serve form (see attached).
- The form already captures a few fields needed for SAP API calls. We could add the missing fields here and enforce validation to mark them required.
- This way, we can fully rely on the SSO config table (that the self-serve form populates) instead of retrieving info from the SAP configuration model.

Regarding the audit model:
- Is it possible to handle this via enhanced logging instead?
- Or if we still want to persist the data in a table, we can generalize the model by renaming it and its attributes.
We could also dump SAP/SAML-specific fields (like tokens, request time, etc.) into a JSON field to keep it flexible for other IDPs in the future.

PR #2375, opened 2025-04-15
fix: enterprise-customer-support sorting/pagination using view

marlonkeating:
[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-9655)

This change fixes the sorting/pagination of the `enterprise-customer-support` api.  Instead of merging the `EnterpriseCustomerUser` and `PendingEnterpriseCustomerUser` entities outside of the django ORM, we now utilize a database view that does the merging, which allows operations like sorting and filtering to take place within the context of the database.

## Testing Instructions
- Pull in this change into local devstack environment
- Run database migration
- Navigate to http://localhost:18000/enterprise/api/v1/enterprise-customer-support/{enterprise_customer_uuid}
- Verify results are sorted by admin status first, then alphabetical ordering

**Merge checklist:**

**Post merge:**


enterprise/migrations/0233_create_vw_enterprise_customer_ecus_pecus.py
```diff
@@ -0,0 +1,56 @@
...
+        # together to facilitate querying both simultaneously
+        migrations.RunSQL(
+            sql="""
+                CREATE VIEW view_enterprise_customer_support_users AS
+                    SELECT ecu.enterprise_customer_id,
+                    au.email AS user_email,
+                    FALSE AS is_pending,
+                    ecu.user_id,
+                    au.username,
+                    au.first_name,
```
kiram15, line 19: I don't think we should be using the first_name and last_name from auth_user, but instead the name field from auth_userprofile with first and last name as a backup like [here](https://github.com/openedx/edx-enterprise/blob/e5c210fe559f094bc37e10ec3bf9df663033c6a2/enterprise/models.py#L4741)

tests/test_enterprise/api/test_views.py
```diff
@@ -9944,6 +9945,76 @@ def test_list_users_filtered(self):
...
+        for ordering_scenario, expected_ordering in [
+            ('administrator', [ecu_learner_eml, pecu_learner_eml, ecu_admin_eml, pecu_admin_eml]),
+            ('-administrator', [ecu_admin_eml, pecu_admin_eml, ecu_learner_eml, pecu_learner_eml]),
+            ('learner', [ecu_learner_eml, pecu_learner_eml, ecu_admin_eml, pecu_admin_eml]),
+            ('-learner', [ecu_admin_eml, pecu_admin_eml, ecu_learner_eml, pecu_learner_eml]),
+            ('details', [ecu_admin_eml, ecu_learner_eml, pecu_admin_eml, pecu_learner_eml]),
+            ('-details', [pecu_learner_eml, pecu_admin_eml, ecu_learner_eml, ecu_admin_eml]),
+            ('', [ecu_admin_eml, pecu_admin_eml, ecu_learner_eml, pecu_learner_eml])
+        ]:
+            print(f'sorting: {ordering_scenario}')
```
kiram15, line 10008: nit: remove print

enterprise/migrations/0233_create_vw_enterprise_customer_ecus_pecus.py
marlonkeating: Good call out!  For this work I'm aiming for data parity with the old model since it's used by support tools, and we don't use `first_name` and `last_name` for our logic, so I've added this as[ follow-up work.](https://2u-internal.atlassian.net/browse/ENT-10350)

enterprise/migrations/0233_create_vw_enterprise_customer_ecus_pecus.py
```diff
@@ -0,0 +1,56 @@
...
+                    au.is_staff,
+                    au.is_active,
+                    au.date_joined,
+                    EXISTS(
+                        SELECT * FROM enterprise_systemwideenterpriseuserroleassignment AS era
+                        INNER JOIN enterprise_systemwideenterpriserole AS ra ON era.role_id = ra.id
+                        WHERE era.enterprise_customer_id = ecu.enterprise_customer_id
+                        AND era.user_id = au.id
+                        AND ra.name = 'enterprise_admin'
+                    ) AS is_admin
```
iloveagent57, line 30: I think you could move this to the primary join block, and then compute `is_admin` with a CASE statement - I think it will be more efficient (maybe):
```sql
...
CASE WHEN ra.name = 'enterprise_admin' THEN true ELSE false END AS is_admin
FROM enterprise_enterprisecustomeruser AS ecu
INNER JOIN auth_user AS au ON au.id = ecu.user_fk
INNER JOIN enterprise_systemwideenterpriseuserroleassignment AS era ON era.enterprise_customer_id = ecu.enterprise_customer_id AND era.user_id = au.id
INNER JOIN enterprise_systemwideenterpriserole AS ra ON era.role_id = ra.id
```

enterprise/migrations/0233_create_vw_enterprise_customer_ecus_pecus.py
```diff
@@ -0,0 +1,56 @@
...
+                    0 AS user_id,
+                    '' AS username,
+                    '' AS first_name,
+                    '' AS last_name,
+                    FALSE AS is_staff,
+                    FALSE AS is_active,
+                    NULL AS date_joined,
+                    EXISTS(
+                        SELECT * FROM enterprise_pendingenterprisecustomeradminuser AS pecau
+                        WHERE pecau.user_email = pecu.user_email
```
iloveagent57, line 48: it might be more efficient to left join this table in, rather than use EXISTS with a subquery.

iloveagent57 (commented):
This looks great! Have you been able to test it locally over a decently-sized set of data?

enterprise/migrations/0233_create_vw_enterprise_customer_ecus_pecus.py
marlonkeating: Good suggestion!  It was slightly more efficient, so I added it in.

enterprise/migrations/0233_create_vw_enterprise_customer_ecus_pecus.py
marlonkeating: It was drastically more efficient, as it turned out.  Querying our largest production customer with the current implementation took around 7 seconds, with this it got down to ~1 second!

Now the bad news:  if an ecu has more than one role assigned to them, duplicate results will be returned for each role they have.

In the end, I went with a common table expression solution that @bcitro suggested, which brings the query down to ~5 seconds.  If you can think of a way to make the join work without producing duplicates, do let me know!

iloveagent57 (approved):

PR #2381, opened 2025-04-25
feat: adds EnterpriseCustomerAdmin model

katrinan029:
Creates an EnterpriseCustomer Admin model to support the admin onboarding feature.
https://2u-internal.atlassian.net/browse/ENT-10137 

**Merge checklist:**

**Post merge:**
- [x] Tag pushed and a new [version](https://github.com/openedx/edx-enterprise/releases) released

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,10 @@ Unreleased
 --------
 * feat: adds EnterpriseCustomerAdmin model
 
+[5.12.10]
+---------
+* fix: fixes datetime attribute error
+
 [5.12.9]
 --------
 * fix: sync ecu table user_fk changes with django
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -523,7 +523,7 @@ class EnterpriseCourseEnrollmentAdminViewSerializer(serializers.ModelSerializer)
         passing certificate.
 
         Arguments:
-            course_details : the details for the exececutive education course run
+            course_details : the details for the executive education course run
             certificate_info: A dict containing the following key:
                 ``is_passing``: whether the  user has a passing certificate in the course run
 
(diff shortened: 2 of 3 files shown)


enterprise/models.py
```diff
@@ -4941,3 +4941,48 @@ class OnboardingFlow(TimeStampedModel):
...
+        related_name='admin_records',
+        on_delete=models.deletion.CASCADE,
+        help_text=_("The enterprise customer user who is an admin.")
+    )
+    last_login = models.DateTimeField(
+        null=True,
+        blank=True,
+        help_text=_("The last time the admin logged into the admin portal.")
+    )
+    completed_tour_flows = JSONField(
```
kiram15, line 4967: We might want to change this to a ManyToMany field so we can directly connect it to the OnboardingFlow model, like `completed_tour_flows = models.ManyToManyField(OnboardingFlow, null=True, blank=True)`

kiram15 (commented):
Just one change!

tests/test_models.py
```diff
@@ -3033,3 +3033,56 @@ def test_sso_config_default_display_name(self):
...
+        self.admin.save()
+        assert self.admin.onboarding_tour_dismissed is True
+
+    def test_onboarding_tour_completed(self):
+        """
+        Test onboarding_tour_completed field.
+        """
+        self.admin.onboarding_tour_completed = True
+        self.admin.save()
+        assert self.admin.onboarding_tour_completed is True
```
jesperhodge, line 3088: I don't know what tests are usually necessary for models, but it might make sense to also test creation as well as the relationships?

Just make sure that `self.admin.enterprise_customer_user` is a user, and that that user's `admin_records` (or if you implement my suggestion above, `admin_record`) point back to this model.

Similarly for onboarding flows.

jesperhodge (changes requested):
Amazing work, Katrina. I just left a comment about one of the relationships, but everything works really well - I ran the migrations and looked at the database tables and everything looks great.

jesperhodge (approved):
Looks great!

PR #2394, opened 2025-05-15
feat: Add SAP SuccessFactors user details API

zwidekalanga:
**Merge checklist:**

**Post merge:**


enterprise/utils.py
```diff
@@ -2614,3 +2614,45 @@ def filter_in_case_insensitive(fieldname, values):
...
+
+    Returns:
+        dict: The user details from the identity provider, or None if not available
+    """
+    # Get the identity provider configuration
+    identity_provider = None
+    if enterprise_customer.has_single_idp:
+        identity_provider = (
+            enterprise_customer.default_provider_idp or
+            enterprise_customer.enterprise_customer_identity_providers.first()
```
zamanafzal, line 2635: Any reason we are not always giving the priority to `default_idp`.

enterprise/utils.py
```diff
@@ -2614,3 +2614,45 @@ def filter_in_case_insensitive(fieldname, values):
...
+        ).identity_provider
+
+    if not identity_provider:
+        LOGGER.info(
+            "No identity provider found for enterprise customer %s", 
+            enterprise_customer.uuid
+        )
+        return None
+    
+    idp_config = identity_provider.get_config()
```
zamanafzal, line 2645: Could you please clarify how identity_provider ends up calling the get_config method? I'm trying to follow the flow.

enterprise/utils.py
```diff
@@ -2614,3 +2614,45 @@ def filter_in_case_insensitive(fieldname, values):
...
+    
+    username_attr_key = idp_config.conf.get('attr_username', 'urn:oid:0.9.2342.19200300.100.1.1')
+    
+    if not username_attr_key:
```
zamanafzal, line 2649: Wondering if this `not username_attr_key `condition ever be `True` given the default in .get()?

enterprise/api/v1/serializers.py
```diff
@@ -2356,3 +2356,11 @@ def to_representation(self, instance):
         if 'completed_tour_flows' in representation:
             representation['completed_tour_flows'] = [str(uuid) for uuid in representation['completed_tour_flows']]
         return representation
+
+
+class EnterpriseSapSfUserInfoRequestSerializer(serializers.Serializer):
```
sameenfatima78, line 2361: replace `EnterpriseSapSfUserInfoRequestSerializer` -> `EnterpriseSSOUserInfoRequestSerializer`

enterprise/utils.py
```diff
@@ -2614,3 +2614,45 @@ def filter_in_case_insensitive(fieldname, values):
...
+            enterprise_customer.default_provider_idp or
+            enterprise_customer.enterprise_customer_identity_providers.first()
+        ).identity_provider
+
+    if not identity_provider:
+        LOGGER.info(
+            "No identity provider found for enterprise customer %s", 
+            enterprise_customer.uuid
+        )
+        return None
```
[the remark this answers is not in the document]
zwidekalanga: @sameenfatima78, I'm not entirely sure how to resolve this issue in the SAP SuccessFactors (SAPSF) scenario, since we’re not currently passing a specific IDP configuration when making the API call. At the moment, we assume the user is using the SAPSF IDP by default. However, since it's possible for an enterprise to have multiple IDPs configured, this assumption could lead to conflicts or unexpected behavior.

To handle this, I propose that we default the enterprise IDP to the one we create if the configure payload includes an SSO configuration that contains SAPSF-related information. Additionally, we should support an optional parameter to allow overriding the default and explicitly specifying the desired IDP configuration.

For this endpoint specifically for SAPSF-related integrations with Auth0, we would continue to default to the SAPSF IDP configuration for enterprises identified as SAPSF users. For non-SAPSF enterprises, they can use the optional parameter to override the default behavior.

Let me know if you think it makes sense to include the override functionality as part of this initial implementation—I'm happy to add it if so.

enterprise/utils.py
zwidekalanga: @zamanafzal I think we might have to changes this entirely based on the feedback from @sameenfatima78 👇, but you can also maybe give me feedback on what you think about the below comment.

enterprise/utils.py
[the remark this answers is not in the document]
zwidekalanga: Had a discussion with @sameenfatima78 and @zamanafzal, and we've decided to keep the current implementation. We just need to inform the Sales team that only one IDP can be linked to an enterprise. If multiple IDPs are configured, the SAPSF one should be set as the default for SAPSF enterprises to ensure authentication via Auth0.

enterprise/utils.py
```diff
@@ -2614,3 +2614,34 @@ def filter_in_case_insensitive(fieldname, values):
...
+    Args:
+        enterprise_customer (EnterpriseCustomer): The enterprise customer to get the identity provider from
+        user_id (str): The user ID to get details for
+
+    Returns:
+        dict: The user details from the identity provider, or None if not available
+    """
+    # Get the identity provider configuration
+    identity_provider = None
+    if enterprise_customer.has_single_idp:
```
sameenfatima78, line 2632: nit: Is this condition necessary?

enterprise/utils.py
```diff
@@ -2614,3 +2614,34 @@ def filter_in_case_insensitive(fieldname, values):
...
+            enterprise_customer.enterprise_customer_identity_providers.first()
+        ).identity_provider
+    if not identity_provider:
+        LOGGER.info(
+            "No identity provider found for enterprise customer %s", 
+            enterprise_customer.uuid
+        )
+        return None
+    idp_config = identity_provider.get_config()
+    username_attr_key = idp_config.conf.get('attr_username', 'urn:oid:0.9.2342.19200300.100.1.1')
```
sameenfatima78, line 2644: nit: Can we use a named constant to hold value for the username attribute? We can define that in the CONSTANTS file and import from there. Also would be nice to also include a comment explaining where is this attribute mapping coming from. 
```python
DEFAULT_USERNAME_ATTR = 'urn:oid:0.9.2342.19200300.100.1.1'

import DEFAULT_USERNAME_ATTR from ..constants
username_attr_key = idp_config.conf.get('attr_username', DEFAULT_USERNAME_ATTR)

**Review by sameenfatima78 — COMMENTED** [review]
Can we include some unit tests to cover the new changes

**Review by zamanafzal — APPROVED** [review]
LGTM. 
+1 to Sameen's suggestion of adding unit tests.

File: tests/test_enterprise/test_utils.py [hunk 8]
```diff
@@ -22,12 +24,19 @@
...
     parse_lms_api_datetime,
     serialize_notification_content,
     truncate_string,
 )
+
+# Fix circular import: ensure we import from project root test_utils, not this file
+project_root = os.path.join(os.path.dirname(__file__), '..', '..')
+if project_root not in sys.path:
+    sys.path.insert(0, project_root)
+
```
sameenfatima78, line 39: To fix circular import errors, we mostly directly import it within the test/method.

sameenfatima78 (approved):
one small suggestion but looks good. Please ensure to update version and draft a new release with that new version tag.

PR #2408, opened 2025-06-05
feat: creates EnterpriseCustomerAdmin records for EnterpriseCustomerUsers with admin roles

katrinan029:
https://2u-internal.atlassian.net/browse/ENT-10216

**Testing Instructions**
1. make sure you have a enterprise_admin roles for users and enterprises here http://localhost:18000/admin/enterprise/systemwideenterpriseuserroleassignment/
2. go into your `lms-shell`
3. run the dry run command `./manage.py lms create_enterprise_customer_admins --dry-run` and verify that you see logs `Would create EnterpriseCustomerAdmin for user {enterprise_user.user_email} and enterprise {enterprise_user.enterprise_customer.name}`
4. run the command `./manage.py lms create_enterprise_customer_admins` and verify that you see logs. I.e. `Created EnterpriseCustomerAdmin for user enterprise_admin_the-whinery-spirits-company@example.com and enterprise The Whinery Spirits Company
Created EnterpriseCustomerAdmin for user enterprise_admin_test-enterprise@example.com and enterprise Test Enterprise`

**Merge checklist:**

**Post merge:**
- [x] Tag pushed and a new [version](https://github.com/openedx/edx-enterprise/releases) released
- [x] After versioned build finishes in [GitHub Actions](https://github.com/openedx/edx-enterprise/actions), verify version has been pushed to [PyPI](https://pypi.org/project/edx-enterprise/)


enterprise/management/commands/create_enterprise_customer_admins.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        if not EnterpriseCustomerAdmin.objects.filter(enterprise_customer_user=enterprise_user).exists():
+            if dry_run:
+                logger.info(
+                    f'Would create EnterpriseCustomerAdmin for user {enterprise_user.user_email} '
+                    f'and enterprise {enterprise_user.enterprise_customer.name}'
+                )
+            else:
+                try:
+                    with transaction.atomic():
```
jesperhodge, line 73: See my other comment - this will be more optimal if you have one query for the whole batch of admins to create, and then wrap that into transaction.atomic if desired.

enterprise/management/commands/create_enterprise_customer_admins.py
```diff
@@ -0,0 +1,85 @@
...
+        batch_size = options['batch_size']
+
+        if dry_run:
+            logger.info('DRY RUN - No changes will be made')
+
+        admin_role_assignments = SystemWideEnterpriseUserRoleAssignment.objects.filter(
+            role__name='enterprise_admin'
+        ).select_related('user', 'enterprise_customer')
+
+        for role_assignments_batch in batch(admin_role_assignments, batch_size=batch_size):
```
jesperhodge, line 51: Awesome that you used `batch` here! That completely avoid problems with loading too many records into memory.

tests/test_enterprise/management/test_create_enterprise_customer_admins.py
```diff
@@ -0,0 +1,80 @@
...
+from enterprise import roles_api
+from test_utils.factories import (
+    EnterpriseCustomerFactory,
+    EnterpriseCustomerUserFactory,
+    UserFactory,
+)
+
+
+@mark.django_db
+@ddt.ddt
```
jesperhodge, line 27: nit: not sure ddt is needed here

tests/test_enterprise/management/test_create_enterprise_customer_admins.py
```diff
@@ -0,0 +1,80 @@
...
+class CreateEnterpriseCustomerAdminsCommandTests(TestCase):
```
jesperhodge, line 28: I like the tests you wrote, clean and concise while catching what's important.

There could maybe be some sort of test that checks that if an error occurs on one record it's swallowed and logged instead of throwing an error and crashing the whole job, but I think it's optional - since it's mostly a one-time task. Up to you whether you want to implement that.

I don't see any worrisome edge cases otherwise, I think the tests are sufficient.

enterprise/management/commands/create_enterprise_customer_admins.py
```diff
@@ -0,0 +1,109 @@
...
+                else:
+                    try:
+                        with transaction.atomic():
+                            admin_records = [
+                                EnterpriseCustomerAdmin(enterprise_customer_user=eu)
+                                for eu in enterprise_users_to_create
+                            ]
+                            EnterpriseCustomerAdmin.objects.bulk_create(admin_records)
+
+                            for enterprise_user in enterprise_users_to_create:
```
jesperhodge, line 101: I ran your command, seems great for the most part.

Nice to know: if you write this code into your edx-platform-> LMS private.py file, you will see all the SQL queries when you run your management command:

```
### UNCOMMENT TO ENABLE DEBUG LOGGING, e.g. DATABASE QUERIES AND SUCH ###

from .production import *  # This imports all settings including LOGGING

DEBUG = True
LOGGING['loggers']['django.db.backends'] = {
    'level': 'DEBUG',
    'handlers': ['console', 'local'],
    # We have a root 'django' logger enabled
    # that we don't want to propagage too, so that
    # this doesn't print multiple times.
    'propagate': True,
}
LOGGING['handlers']['console'] = {
    'level': 'DEBUG',
    'class': 'logging.StreamHandler',
}
```

So I used this and had a look at the SQL generated.
For the most part after your changes you have very optimal SQL like this:
`(0.001) INSERT INTO `enterprise_enterprisecustomeradmin` (`created`, `modified`, `uuid`, `enterprise_customer_user_id`, `last_login`, `onboarding_tour_dismissed`, `onboarding_tour_completed`) VALUES ('2025-06-10 18:56:00.304411', '2025-06-10 18:56:00.304411', '18c154374eff411595fb665dadd0402a', 1, NULL, 0, 0), ('2025-06-10 18:56:00.304481', '2025-06-10 18:56:00.304481', 'a6faa3e4571e461c87299bc6aa3867d2', 3, NULL, 0, 0); args=('2025-06-10 18:56:00.304411', '2025-06-10 18:56:00.304411', '18c154374eff411595fb665dadd0402a', 1, None, False, False, '2025-06-10 18:56:00.304481', '2025-06-10 18:56:00.304481', 'a6faa3e4571e461c87299bc6aa3867d2', 3, None, False, False); alias=default
`

You can see there that it's now one query inserting 2 records at once - that will be however many admins your task creates.

Only one little thing I observed was the following:
There's a bunch of SELECT statements that are generated per user. So it looks like the query still reads one user at a time, in a loop, after creating them in bulk.
I think this line here is probably the culprit:
`for enterprise_user in enterprise_users_to_create:`
That creates a loop and for every entry in the loop it sends a read request to the database.

I think it's probably no problem because I assume we are not creating that many admin records, but better be safe.

The problem does not occur unless you access the attribute. Basically when you just log the stringified version of enterprise_user, everything is fine, but when you do `.user_email` and `.name`, it will create a new query for each admin.
The simplest way to fix this is to just do something like `f'Created EnterpriseCustomerAdmin for user {enterprise_user}`, that is probably sufficient for logging and fixes it (outputs something like "<EnterpriseCustomerUser 3>: Test Enterprise - 25". If you really want these details, you probably want to do a bulk prefetch with prefetch_related: enterprise_customer for these users.

jesperhodge (changes requested):

enterprise/management/commands/create_enterprise_customer_admins.py
```diff
@@ -0,0 +1,108 @@
...
+                    (role_assignment.user.id, role_assignment.enterprise_customer.uuid)
+                )
+                if enterprise_user and enterprise_user.id not in existing_admin_enterprise_user_ids:
+                    enterprise_users_to_create.append(enterprise_user)
+
+            if enterprise_users_to_create:
+                if dry_run:
+                    for enterprise_user in enterprise_users_to_create:
+                        logger.info(
+                            f'Would create EnterpriseCustomerAdmin for user {enterprise_user.user_email} '
```
jesperhodge, line 89:
```suggestion
                            f'Would create EnterpriseCustomerAdmin for user {enterprise_user} '
```

enterprise/management/commands/create_enterprise_customer_admins.py
```diff
@@ -0,0 +1,108 @@
...
+                            f'and enterprise {enterprise_user.enterprise_customer.name}'
```
jesperhodge, line 90:
```suggestion
```

tests/test_enterprise/management/test_create_enterprise_customer_admins.py
```diff
@@ -0,0 +1,69 @@
...
+        """
+        Test that dry run shows what would be created but doesn't create records.
+        """
+        with LogCapture(level=logging.INFO) as log_capture:
+            call_command(self.command, '--dry-run')
+
+            assert EnterpriseCustomerAdmin.objects.count() == 0
+
+            expected_msg = (
+                f'Would create EnterpriseCustomerAdmin for user {self.user.email} '
```
jesperhodge, line 52:
```suggestion
                f'Would create EnterpriseCustomerAdmin for user {self.enterprise_customer_user} '
```

tests/test_enterprise/management/test_create_enterprise_customer_admins.py
```diff
@@ -0,0 +1,69 @@
...
+                f'and enterprise {self.enterprise_customer.name}'
```
jesperhodge, line 53:
```suggestion
```

jesperhodge (changes requested):

jesperhodge (approved):

PR #2442, opened 2025-08-18
feat: fetch SAP userid by remote_id_field_name

sameenfatima78:
For SAP customers leveraging Auth0 for SSO , the UID value returned in the SAML response represents the Auth0 User ID, not the SAP User ID. However, SAP integrations require the SAP User ID to properly report learner activity and completions. To support this, these changes introduce logic to fetch the SAP User ID from UserSocialAuth records, using the relevant remote user ID attribute (the attribute that Auth0 returns as the SAP User ID).
This attribute name can be provided as a query parameter to the UserMapping API, which will then retrieve the correct ID.

Related work: https://github.com/edx/edx-platform/pull/33 
 
**Note**: These changes are made only to support existing SAP + Auth0 customers.
We cannot alter current learner UIDs because such changes would directly impact authentication and SSO flows. For new SAP customers using Auth0, the recommended approach is to configure the User ID attribute in SAMLProviderConfig to directly pass the SAP User ID attribute returned by Auth0.

**JIRA**:
- https://2u-internal.atlassian.net/browse/ENT-10804
- https://2u-internal.atlassian.net/browse/ENT-11158

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * nothing unreleased
 
+[6.5.7] - 2025-11-28
+---------------------
+* feat: fetch SAP user id by remote_id_field_name
+
 [6.5.6] - 2025-11-26
 ---------------------
 * chore: upgrade python requirements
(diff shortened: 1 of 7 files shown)


sameeramin (approved):
Looks good to me 👍🏼

PR #2451, opened 2025-08-28
fix: add migration for django 5.2 support in edx-platform

UsamaSadiq:
## Description
- Under `Django 5.2` env, a new migration is being created which will be needed in edx-platform for Django 5.2 support so ran the `makemigrations` to generate the new migration under Django52 env.
- Bumped the package version so it can be used in platform


awais786 (approved):

UsamaSadiq:
Raw SQL for the migration is following so it shouldn't cause any issue in edx-platform 
```
BEGIN;
--
-- Alter field realized_enrollments on defaultenterpriseenrollmentintention
--
-- (no-op)
COMMIT;
```

PR #2452, opened 2025-08-30
fix: upgrade djangorestframework and skip automatic validators

pwnage101:
Background:
---

About two weeks ago edx-platform upgraded DRF from 3.14->3.16.1 and enterprise team started noticing the `POST /api/v1/pending-enterprise-learner` endpoint stopped being idempotent. It would return 400 errors instead of 204 No Content on duplicate requests.

Root Cause:
---

DRF 3.15 introduced a feature to automatically add uniqueness validators to a model serializer if any are found on the underlying model. While this feature is nice, it broke assumptions in the logic for many of our serializers.

Solution:
---
Disable the new feature for affected model serializers by overriding the `Meta.validators` attribute.

ENT-10865

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -1032,6 +1032,11 @@ class PendingEnterpriseCustomerUserSerializer(serializers.ModelSerializer):
         fields = (
             'enterprise_customer', 'user_email'
         )
+        # [ENT-10865] Explicitly set the validators attribute to prevent ModelSerializer
+        # from automatically generating uniqueness validators based on unique constraints
+        # on the underlying model. While this new feature is nice, it breaks assumptions
+        # in existing logic within this serializer.
+        validators = []
 
     def to_representation(self, instance):
         '''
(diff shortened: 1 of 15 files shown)


requirements/test.txt
```diff
@@ -160,84 +161,83 @@ diff-cover==9.6.0
...
+    # via -r requirements/test-master.txt
 django-waffle==5.0.0
     # via
-    #   -r /home/runner/work/edx-enterprise/edx-enterprise/requirements/test-master.txt
+    #   -r requirements/test-master.txt
     #   edx-django-utils
     #   edx-drf-extensions
     #   edx-toggles
-djangorestframework==3.14.0
+djangorestframework==3.16.1
```
pwnage101, line 204: Tests fail after upgrading this, but before my other changes in this PR.

tests/test_enterprise/api/test_filters.py
```diff
@@ -54,7 +54,7 @@ def test_filter_for_list(self, is_staff, expected_data_length):
         assert len(data['results']) == expected_data_length
 
     @ddt.data(
-        (False, False, {'detail': 'Not found.'}),
+        (False, False, {'detail': 'No EnterpriseCustomer matches the given query.'}),
```
pwnage101, line 57: All of these changes are the result of the new DRF library being more explicit about 404 errors. There should be no impact because consumers shouldn't really use this message string to drive business logic.

enterprise/api/v1/serializers.py
```diff
@@ -1032,6 +1032,11 @@ class Meta:
         fields = (
             'enterprise_customer', 'user_email'
         )
+        # [ENT-10865] Explicitly set the validators attribute to prevent ModelSerializer
+        # from automatically generating uniqueness validators based on unique constraints
+        # on the underlying model. While this new feature is nice, it breaks assumptions
+        # in existing logic within this serializer.
+        validators = []
```
pwnage101, line 1039: This is the actual fix to the 400 errors in prod.

brobro10000 (approved):
Requires a version bump and change log. This may not be able to go out right away due to EdX-platform pipeline paused during weekends.

PR #2460, opened 2025-09-15
feat: Add JSON logging capability to integrated_channels

sameeramin:
This PR implements enhanced structured logging utilities for Integrated Channels to improve Datadog monitoring and alerting, addressing log level inconsistencies and poor structured data parsing identified in ENT-8024.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/tests/test_integrated_channels/test_canvas/test_utils.py
+++ b/tests/test_integrated_channels/test_canvas/test_utils.py
@@ -176,3 +176,40 @@ class TestCanvasUtils(unittest.TestCase):
                 mock_session,
                 'nonMatchingedXCourseId',
             )
+
+    @mock.patch('integrated_channels.canvas.utils.CanvasUtil.find_course_in_account')
+    @mock.patch('integrated_channels.canvas.utils.CanvasUtil.find_root_canvas_account')
+    @mock.patch('integrated_channels.canvas.utils.LOGGER')
+    def test_find_course_under_root_account_with_logging(
+        self, mock_logger, mock_find_root_canvas_account, mock_find_course_in_account
+    ):
+        """Test that logging occurs when course is found under root account."""
+        # Mock the course object
+        mock_course = {'id': 123, 'integration_id': 'test-course-id', 'name': 'Test Course'}
+
+        # Mock the root account
+        mock_root_account = {'id': 1, 'parent_account_id': None}
+
+        # Setup the mock calls:
+        # First call returns None (not found in enterprise account)
+        # Second call returns the course (found in root account)
+        mock_find_course_in_account.side_effect = [None, mock_course]
+        mock_find_root_canvas_account.return_value = mock_root_account
+
+        mock_session, _ = refresh_session_if_expired(self.get_oauth_access_token)
+
+        # Call the method
+        result = CanvasUtil.find_course_by_course_id(self.enterprise_config, mock_session, 'test-course-id')
+
+        # Verify the result
+        assert result == mock_course
+
+        # Verify the logging call was made with correct parameters
+        mock_logger.info.assert_called_once_with(
+            'Found course under root Canvas account',
+            extra={
+                'channel_name': 'canvas',
+                'enterprise_customer_uuid': self.enterprise_config.enterprise_customer.uuid,
+                'course_or_course_run_key': 'test-course-id',
+            },
+        )
--- a/tests/test_integrated_channels/test_cornerstone/test_client.py
+++ b/tests/test_integrated_channels/test_cornerstone/test_client.py
@@ -60,3 +60,20 @@ class TestCornerstoneApiClient(unittest.TestCase):
         assert IntegratedChannelAPIRequestLogs.objects.count() == 1
         assert len(responses.calls) == 1
         assert output == (200, '"{}"')
+
+    @unittest.mock.patch('integrated_channels.cornerstone.client.LOGGER')
+    def test_cleanup_duplicate_assignment_records_logging(self, mock_logger):
+        """
+        Test that cleanup_duplicate_assignment_records logs the not implemented message correctly.
+        """
+        cornerstone_api_client = CornerstoneAPIClient(self.csod_config)
+
+        # Call the method with dummy courses
+        cornerstone_api_client.cleanup_duplicate_assignment_records(['course1', 'course2'])
+
+        # Verify the logging call was made with correct parameters
+        mock_logger.error.assert_called_once()
+        args, kwargs = mock_logger.error.call_args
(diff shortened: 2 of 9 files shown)


integrated_channels/logger.py
```diff
@@ -0,0 +1,128 @@
...
+            log_entry = {
+                'timestamp': datetime.fromtimestamp(record.created).strftime('%Y-%m-%d %H:%M:%S,%f')[:-3],
+                'level': record.levelname,
+                'logger': record.name,
+                'message': record.getMessage(),
+                'module': record.module,
+                'function': record.funcName,
+                'line': record.lineno
+            }
+
```
zamanafzal, line 31: I think we will need correlation ids  to track related operations.

```
import uuid
import threading
# Thread-local storage for correlation IDs
_local = threading.local()

def set_correlation_id(correlation_id=None):
    """Set correlation ID for current thread"""
    _local.correlation_id = correlation_id or str(uuid.uuid4())

def get_correlation_id():
    """Get correlation ID for current thread"""
    return getattr(_local, 'correlation_id', None)

correlation_id = get_correlation_id()
if correlation_id:
    log_entry['correlation_id'] = correlation_id
```

integrated_channels/logger.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+
+    def __init__(self):
+        self.use_json = USE_JSON_LOGGING
+        super().__init__()
+
+    def format(self, record):
+        if self.use_json:
+            log_entry = {
+                'timestamp': datetime.fromtimestamp(record.created).strftime('%Y-%m-%d %H:%M:%S,%f')[:-3],
```
zamanafzal, line 23: How about making the timestamp ISO 8601 compliant?

integrated_channels/logger.py
```diff
@@ -0,0 +1,128 @@
...
+                'level': record.levelname,
+                'logger': record.name,
+                'message': record.getMessage(),
+                'module': record.module,
+                'function': record.funcName,
+                'line': record.lineno
+            }
+
+            log_entry['service'] = 'integrated_channels'
+
```
zamanafzal, line 33: IS it possible to add perfomance metrics here aswell?

```
# In format method, add:
if hasattr(record, 'duration_ms'):
    log_entry['performance.duration_ms'] = record.duration_ms
if hasattr(record, 'status_code'):
    log_entry['http.status_code'] = record.status_code
if hasattr(record, 'payload_size'):
    log_entry['payload.size_bytes'] = record.payload_size
```

integrated_channels/logger.py
```diff
@@ -0,0 +1,128 @@
...
+    else:
+        formatted_message = generate_formatted_log(
+            channel_name=context.get("channel_name", None),
+            enterprise_customer_uuid=context.get("enterprise_customer_uuid", None),
+            lms_user_id=context.get("lms_user_id", None),
+            course_or_course_run_key=context.get("course_or_course_run_key", None),
+            message=message,
+            plugin_configuration_id=context.get("plugin_configuration_id", None)
+        )
+        context_logger(formatted_message)
```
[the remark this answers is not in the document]
sameeramin: Yeah, you're right, thanks for pointing this out

integrated_channels/logger.py
sameeramin: We can, but we need to make it backward compatible; that's why I used the same format as before

integrated_channels/logger.py
```diff
@@ -0,0 +1,128 @@
...
+    logger_name = name or 'integrated_channels'
+    logger = logging.getLogger(logger_name)
+
+    # Avoid adding handlers multiple times
+    if logger.handlers:
+        return logger
+
+    # Create console handler
+    console_handler = logging.StreamHandler(sys.stdout)
+    console_handler.setLevel(logging.INFO)
```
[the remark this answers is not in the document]
sameeramin: Sure, I'll make sure to update that.

integrated_channels/logger.py
sameeramin: Yeah, you’re right! We will be implementing `correlation_id`, `span_id`, and `trace_id` in next iteration

integrated_channels/logger.py
```diff
@@ -0,0 +1,128 @@
...
+    def __init__(self):
+        self.use_json = USE_JSON_LOGGING
+        super().__init__()
+
+    def format(self, record):
+        if self.use_json:
+            log_entry = {
+                'timestamp': datetime.fromtimestamp(record.created).strftime('%Y-%m-%d %H:%M:%S,%f')[:-3],
+                'level': record.levelname,
+                'logger': record.name,
```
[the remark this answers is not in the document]
sameeramin: `severity`, `level`, `status`, etc, are reserved attributes so that they will map to the Datadog `status` attribute.

integrated_channels/logger.py
```diff
@@ -0,0 +1,128 @@
...
+                'line': record.lineno
+            }
+
+            log_entry['service'] = 'integrated_channels'
+
+            if record.exc_info:
+                log_entry['error.kind'] = record.exc_info[0].__name__
+                log_entry['error.message'] = str(record.exc_info[1])
+                log_entry['error.stack'] = self.formatException(record.exc_info)
+
```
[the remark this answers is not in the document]
sameeramin: These attributes have already been added to the dict, but yeah, I think it's a good idea to move them under `error` attributes

integrated_channels/logger.py
sameeramin: Added `status_code` and will also add other fields in phase 2, along with categorizations

sameenfatima78 (approved):

sameeramin:
Hey @pwnage101, Thank you for reviewing the PR I absolutely agree with the approach you proposed. Let me refactor my code.

pwnage101 (approved):
LGTM!

pwnage101:
I minimally tested this approach and it seems to work, but I highly recommend you do your own testing to validate the approach for more scenarios:

```diff
-_message_with_keyvalue %{regex("(?<message>.*)"):message_keys:keyvalue}
+_message_with_keyvalue (%{regex(".*(?= \\{.*\\})"):message} %{data:json_keys:json}|%{regex("(?<message>.*)"):message_keys:keyvalue})
```

This matches existing messages optionally with keyvalue pairs, but if the message ends with JSON, then it will switch to that pattern which uses the `json` filter.

Please perform testing on the actual Grox parser edit page (but do not hit save until you get peer review on the new regex): https://app.datadoghq.com/logs/pipelines/pipeline/of_y4sdqStWBHmooL2W3UQ/processors/edit/Wz-1KN_ISsyavH79U5dmgg

PR #2491, opened 2025-12-16
ENT-11235: Updating schema to add invited_date and joined_date fields to customer admin.

sjasti-sonata-svg:
**Merge checklist:**

**Post merge:**


enterprise/models.py
```diff
@@ -1092,6 +1092,27 @@ class EnterpriseCustomerUser(TimeStampedModel):
...
+        help_text="Timestamp when the user was invited."
+    )
+
+    joined_date = models.DateTimeField(
+        null=True,
+        blank=True,
+        help_text="Timestamp when the user accepted the invite / joined."
+    )
+
+    invited_by = models.ForeignKey(
```
kiram15: I think you need to plan out more about what fields you need to create. The auth table that you're using here is a huge and cumbersome table that always risks introducing problems. If you want to do what you're proposing and add an invited_by field, instead of connecting to the auth_user table, I would instead repurpose the EnterpriseCustomerAdmin table [here](https://github.com/openedx/edx-enterprise/blob/47f997cfeb0deb1ea15163935e147457327f5058/enterprise/models.py#L4975). Right now, it is being used only to track a user's progress through the admin-portal onboarding tour, but could be used for this purpose, because it connects to a EnterpriseCustomerUser, which is connected to the Auth table. 

**But** I don't think this we should add this field at all, since I see no need to know which admin they were invited by in any of the Figma mocks, and I don't see any reason we should store this unnecessary information that isn't being reflected anywhere on the frontend unless we have a good reason.

kiram15 (dismissed):
I'm confused as to why we are changing fields in the EnterpriseCustomerUser table. When you are going to pull all the admins for the new admin DataTable, we should not be going through a list of the EnterpriseCustomerUsers, but rather another table of all the admins. As I said in the comment, I think the best approach would be adding the `invited_date` and `joined_date` fields to EnterpriseCustomerAdmin table, and then plan to pull in the (relevant, non-onboarding tour) information for the DataTable y'all will be creating. 

I also think that whenever we're adding new fields, its good to include in the PR how they will be getting populated. [Here is how](https://github.com/openedx/edx-enterprise/blob/47f997cfeb0deb1ea15163935e147457327f5058/enterprise/api/__init__.py#L27) EnterpriseCustomerAdmins are currently created (in a post_save django signal where an "enterprise_admin" system-wide role is assigned to the user). I'm assuming the plan is to have the invited_date be when the [PendingEnterpriseCustomerAdminUser](https://github.com/openedx/edx-enterprise/blob/47f997cfeb0deb1ea15163935e147457327f5058/enterprise/models.py#L3954C7-L3954C41) record is created, and then a joined_date be populated if/when the EnterpriseCustomerAdmin record is created, but in order for me to review these fields, I need to know more information about these dates.

enterprise/api/v1/serializers.py
```diff
@@ -2047,7 +2047,7 @@ class Meta:
     is_admin = serializers.SerializerMethodField()
 
     def is_enterprise_customer_user(self, obj):
-        return hasattr(obj, 'user_id') and obj.user_id > 0
+        return bool(obj.user_id)
```
kiram15: Why are we changing this? And I don't see it getting used anywhere?

pwnage101:
Before merging, please squash your commits to only one commit and use a more descriptive title, such as "feat: add date fields to track enterprise user invitation lifecycle".  Please also change the title of this PR to match that more descriptive title.

msaha-sonata (changes requested):

enterprise/models.py
```diff
@@ -5006,6 +5011,17 @@ class EnterpriseCustomerAdmin(TimeStampedModel):
         default=False,
         help_text=_("Whether the admin has completed the onboarding tour.")
     )
+    invited_date = models.DateTimeField(
+        null=True,
+        blank=True,
+        help_text="Timestamp when the admin was invited."
+    )
+
+    joined_date = models.DateTimeField(
```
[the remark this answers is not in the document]
sjasti-sonata-svg: Hi @msaha-sonata ,
Added the reason.

tests/test_enterprise/test_signals.py
```diff
@@ -415,6 +417,48 @@ def test_delete_pending_enterprise_admin_user(self):
         ).delete()
         self._assert_pending_ecus_exist(should_exist=False)
 
+@mark.django_db
+class TestEnterpriseCustomerAdminSignals(unittest.TestCase):
```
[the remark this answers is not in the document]
sjasti-sonata-svg: Hi @msaha-sonata ,
Added the test cases.

enterprise/models.py
```diff
@@ -3971,6 +3970,13 @@ class PendingEnterpriseCustomerAdminUser(TimeStampedModel):
     user_email = models.EmailField(null=False, blank=False)
     history = HistoricalRecords()
 
+    """Track the date and time when a pending enterprise admin invite was created."""
+    invited_date = models.DateTimeField(
+        null=True,
+        blank=True,
+        help_text="Timestamp when the admin invite was created."
+    )
```
pwnage101: This field seems redundant; there's already a `created` field inherited from TimeStampedModel. Furthermore, this PR lacks code changes for populating this field.

enterprise/models.py
```diff
@@ -5006,6 +5012,18 @@ class EnterpriseCustomerAdmin(TimeStampedModel):
         default=False,
         help_text=_("Whether the admin has completed the onboarding tour.")
     )
+    """Track when the admin was invited."""
```
pwnage101: This is the incorrect syntax for code comments.  Use `#` prefix:


```suggestion
    # Track when the admin was invited.
```

enterprise/models.py
```diff
@@ -5006,6 +5012,18 @@ class EnterpriseCustomerAdmin(TimeStampedModel):
...
+    invited_date = models.DateTimeField(
+        null=True,
+        blank=True,
+        help_text="Timestamp when the admin was invited."
+    )
+    """Track when the admin accepted the invite and joined."""
+    joined_date = models.DateTimeField(
```
pwnage101: This model also inherits TimeStampedModel, so joined_date is redundant with `created`.

enterprise/management/commands/create_enterprise_customer_admins.py
```diff
@@ -110,7 +110,8 @@ def handle(self, *args, **options):
                             admin_records = [
                                 EnterpriseCustomerAdmin(
                                     enterprise_customer_user=eu,
-                                    last_login=timezone.now()
+                                    last_login=timezone.now(),
+                                    joined_date=timezone.now()
```
pwnage101: Nit: leave a trailing comma.

CHANGELOG.rst
pwnage101: Your commit should never include changelog entries from other commits.  You still need to rebase your commit onto the upstream origin.

tests/test_enterprise/test_signals.py
```diff
@@ -415,6 +417,67 @@ def test_delete_pending_enterprise_admin_user(self):
...
```
pwnage101: This feature no longer uses signals, so it should not be considered a signals test.  You should relocate these test cases.  Probably [tests/test_models.py](https://github.com/openedx/edx-enterprise/blob/080a8565dc1598086435ced92dc96ad3a7ca261b/tests/test_models.py)

enterprise/signals.py
```diff
@@ -8,11 +8,16 @@
...
+from django.utils.timezone import now
 
 from enterprise import models, roles_api
 from enterprise.api import activate_admin_permissions
 from enterprise.api_client.enterprise_catalog import EnterpriseCatalogApiClient
 from enterprise.decorators import disable_for_loaddata
+from enterprise.models import (
+    EnterpriseCustomerAdmin,
+    PendingEnterpriseCustomerAdminUser,
+)
```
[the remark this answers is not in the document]
sjasti-sonata-svg: Done.

tests/test_enterprise/test_signals.py
sjasti-sonata-svg: DONE.

PR #2543, opened 2026-03-04
feat: add GradeEventContextEnricher pipeline step for grade analytics

pwnage101:
Introduces enterprise/filters/grades.py with GradeEventContextEnricher, a PipelineStep for the org.openedx.learning.grade.context.requested.v1 filter that enriches grade analytics event context with the learner's enterprise UUID. Adds openedx-filters as a dependency and unit tests covering both enrichment and no-op branches.

ENT-11563

You can see it enriching the payload with the enterprise uuid correctly when fired directly 
Screenshot 2026-05-18 at 8 29 18 PM

And in the COURSE_GRADE_PASSED_FIRST_TIME event sender also has the enterprise uuid information 
Screenshot 2026-05-18 at 8 31 56 PM

Also tested with a user that doesn't have any enterprise associations, just prints enterprise uuid empty 
Screenshot 2026-05-18 at 8 33 45 PM

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

---

Blocked by:
* https://github.com/openedx/openedx-filters/pull/333

Blocks:
* https://github.com/openedx/openedx-platform/pull/38097

### DIFF
--- a/enterprise/filters/grades.py
+++ b/enterprise/filters/grades.py
@@ -2,11 +2,14 @@
 Pipeline step for enriching grade analytics event context.
 """
 import copy
+import logging
 from typing import Any
 
 from openedx_filters.filters import PipelineStep
 
 from enterprise.models import EnterpriseCourseEnrollment
+
+log = logging.getLogger(__name__)
 
 
 class GradeEventContextEnricher(PipelineStep):
@@ -39,6 +42,11 @@ class GradeEventContextEnricher(PipelineStep):
                     "course_id": <unchanged>,
                 }
         """
+        log.info(
+            "GradeEventContextEnricher running: user_id=%s, course_id=%s",
+            str(user_id),
+            str(course_id),
+        )
         uuids = EnterpriseCourseEnrollment.get_enterprise_uuids_with_user_and_course(user_id, course_id)
         if uuids:
             context = copy.deepcopy(context)  # create a copy to avoid altering the passed-in value.
--- a/enterprise/filters/grades.py
+++ b/enterprise/filters/grades.py
@@ -1,0 +1,47 @@
+"""
+Pipeline step for enriching grade analytics event context.
+"""
+import copy
+from typing import Any
+
+from openedx_filters.filters import PipelineStep
+
+from enterprise.models import EnterpriseCourseEnrollment
+
+
+class GradeEventContextEnricher(PipelineStep):
+    """
+    Enriches a grade analytics event context dict with the learner's enterprise UUID.
+
+    This step is intended to be registered as a pipeline step for the
+    ``org.openedx.learning.grade.context.requested.v1`` filter.
+
+    If the user is enrolled in the given course through an enterprise, the enterprise
+    UUID is added to the context under the key ``"enterprise_uuid"``. If the user has
+    no enterprise course enrollment, the context is returned unchanged.
+    """
+
+    def run_filter(self, context: dict, user_id: int, course_id: str) -> dict[str, Any]:  # pylint: disable=arguments-differ
+        """
+        Add enterprise UUID to the event context if the user has an enterprise enrollment.
+
+        Arguments:
+            context (dict): the event tracking context dict.
+            user_id (int): the ID of the user whose grade event is being emitted.
+            course_id (str): the course key for the grade event.
+
+        Returns:
+            dict: updated pipeline data with the enriched ``context`` dict::
+
+                {
+                    "context": <enriched context>,
+                    "user_id": <unchanged>,
+                    "course_id": <unchanged>,
+                }
+        """
+        uuids = EnterpriseCourseEnrollment.get_enterprise_uuids_with_user_and_course(user_id, course_id)
+        if uuids:
+            context = copy.deepcopy(context)  # create a copy to avoid altering the passed-in value.
+            # Warning: Selecting the first element is not likely deterministic!
+            context["enterprise_uuid"] = str(uuids[0])
+        return {"context": context, "user_id": user_id, "course_id": course_id}
--- a/tests/filters/test_grades.py
+++ b/tests/filters/test_grades.py
@@ -1,0 +1,73 @@
+"""
+Tests for enterprise.filters.grades pipeline step.
+"""
+import uuid
+from unittest.mock import patch
+
+from django.test import TestCase
+
+from enterprise.filters.grades import GradeEventContextEnricher
+
+
+class TestGradeEventContextEnricher(TestCase):
+    """
+    Tests for GradeEventContextEnricher pipeline step.
+    """
+
+    def _make_step(self):
+        return GradeEventContextEnricher(
+            filter_type="org.openedx.learning.grade.context.requested.v1",
+            running_pipeline=[],
+        )
+
+    @patch("enterprise.models.EnterpriseCourseEnrollment.get_enterprise_uuids_with_user_and_course")
+    def test_enriches_context_when_enterprise_enrollment_found(self, mock_get_uuids):
+        """
+        When an enterprise course enrollment exists, enterprise_uuid is added to context.
+        """
+        enterprise_uuid = uuid.uuid4()
+        mock_get_uuids.return_value = [enterprise_uuid]
+
+        step = self._make_step()
+        context = {"org": "TestOrg", "course_id": "course-v1:org+course+run"}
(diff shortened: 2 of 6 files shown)


enterprise/filters/grades.py
```diff
@@ -0,0 +1,39 @@
...
+
+    This step is intended to be registered as a pipeline step for the
+    ``org.openedx.learning.grade.context.requested.v1`` filter.
+
+    If the user is enrolled in the given course through an enterprise, the enterprise
+    UUID is added to the context under the key ``"enterprise_uuid"``. If the user has
+    no enterprise course enrollment, the context is returned unchanged.
+    """
+
+    def run_filter(self, context, user_id, course_id):  # pylint: disable=arguments-differ
```
pwnage101: Add python type hints
```suggestion
    def run_filter(self, context: dict, user_id: int, course_id: str) -> dict[str, Any]:  # pylint: disable=arguments-differ
```

enterprise/filters/grades.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+        Add enterprise UUID to the event context if the user has an enterprise enrollment.
+
+        Arguments:
+            context (dict): the event tracking context dict.
+            user_id (int): the ID of the user whose grade event is being emitted.
+            course_id (str or CourseKey): the course key for the grade event.
```
pwnage101: Per the filter definition, it's impossible for course_id to be a CourseKey: https://github.com/openedx/openedx-filters/blob/d4b09fdb850436bbcea638066c124a47b5025084/openedx_filters/learning/filters.py#L1584

Fixed docstring:
```suggestion
            course_id (str): the course key for the grade event.
```

enterprise/filters/grades.py
```diff
@@ -0,0 +1,39 @@
...
+
+        Returns:
+            dict: updated pipeline data with the enriched ``context`` dict.
```
pwnage101: Be more specific about the shape of the output dict.
```suggestion
            dict: updated pipeline data with the enriched ``context`` dict.
                {
                    "context": <enriched context>,
                    "user_id": <unchanged>,
                    "course_id": <unchanged>,
                }
```

enterprise/filters/grades.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+        uuids = EnterpriseCourseEnrollment.get_enterprise_uuids_with_user_and_course(
+            str(user_id),
+            str(course_id),
+        )
+        if uuids:
+            return {"context": {**context, "enterprise_uuid": str(uuids[0])}}
+        return {"context": context}
```
pwnage101: Missing passthrough values and should probably just utilize updated_context
```suggestion
        return {"context": context, "user_id": user_id, "course_id": course_id}
```

enterprise/filters/grades.py
```diff
@@ -0,0 +1,39 @@
...
+
+        Returns:
+            dict: updated pipeline data with the enriched ``context`` dict.
+        """
+        uuids = EnterpriseCourseEnrollment.get_enterprise_uuids_with_user_and_course(
+            str(user_id),
+            str(course_id),
+        )
+        if uuids:
+            return {"context": {**context, "enterprise_uuid": str(uuids[0])}}
```
pwnage101: 1. This lacks passthrough values, 2. probably shouldn't just be an early-return since that's less maintainable, and 3. leverages complicated destructuring instead of something more pythonic and something which allows us to retain one exit point of the function.   Solve all 3 problems at once:
```suggestion
            context = deepcopy(context)  # create a copy to avoid altering the passed-in value.
            # Warning: Selecting the first element is not likely deterministic!
            context["enterprise_uuid"] = str(uuids[0])  
```

requirements/test.txt
```diff
@@ -392,6 +392,8 @@ openedx-atlas==0.7.0
     # via -r requirements/test-master.txt
 openedx-events==10.5.0
     # via -r requirements/test-master.txt
+openedx-filters==2.1.0
```
pwnage101: ❌ 3.3.0 now supports python 3.11.  We should take this opportunity to deploy with the latest version.
```suggestion
openedx-filters==3.3.0
```

Ditto for the rest of the requirements changes in this PR.

enterprise/filters/grades.py
```diff
@@ -0,0 +1,39 @@
...
+            user_id (int): the ID of the user whose grade event is being emitted.
+            course_id (str or CourseKey): the course key for the grade event.
+
+        Returns:
+            dict: updated pipeline data with the enriched ``context`` dict.
+        """
+        uuids = EnterpriseCourseEnrollment.get_enterprise_uuids_with_user_and_course(
+            str(user_id),
+            str(course_id),
+        )
```
pwnage101: 1. `user_id` can stay an `int`.  I looked at the function, it's okay.
2. course_id can be assumed to be a `str`, that's promised by the type hint.

```suggestion
        uuids = EnterpriseCourseEnrollment.get_enterprise_uuids_with_user_and_course(user_id, course_id)
```

tests/filters/test_grades.py
```diff
@@ -0,0 +1,74 @@
...
+    def _make_step(self):
+        return GradeEventContextEnricher(
+            "org.openedx.learning.grade.context.requested.v1",
+            [],
+        )
+
+    @patch(
+        "enterprise.filters.grades.EnterpriseCourseEnrollment"
+        ".get_enterprise_uuids_with_user_and_course"
+    )
```
pwnage101: 1. Overly strict line length limits.
2. Unnecessary patching of module at import (`enterprise.filters.grades`) instead of source (`enterprise.models`).
```suggestion
    @patch("enterprise.models.EnterpriseCourseEnrollment.get_enterprise_uuids_with_user_and_course")
```
Ditto for other tests.

tests/filters/test_grades.py
```diff
@@ -0,0 +1,74 @@
...
+class TestGradeEventContextEnricher(TestCase):
+    """
+    Tests for GradeEventContextEnricher pipeline step.
+    """
+
+    def _make_step(self):
+        return GradeEventContextEnricher(
+            "org.openedx.learning.grade.context.requested.v1",
+            [],
+        )
```
pwnage101, line 21: Use argument names to improve readability:
```suggestion
        return GradeEventContextEnricher(
            filter_type="org.openedx.learning.grade.context.requested.v1",
            running_pipeline=[],
        )
```

pwnage101 (commented):
Aside from these changes, please also remember to bump the version of edx-enterprise.

Apologies if this wasn't clear, but these draft claude-generated PRs were meant to offer guidance and direction, not to be ready to go as-is.

enterprise/filters/grades.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        Add enterprise UUID to the event context if the user has an enterprise enrollment.
+
+        Arguments:
+            context (dict): the event tracking context dict.
+            user_id (int): the ID of the user whose grade event is being emitted.
+            course_id (str): the course key for the grade event.
+
+        Returns:
+            dict: updated pipeline data with the enriched ``context`` dict::
```
kiram15, line 37: This weird :: and line break is dictated by ReadTheDocs

pwnage101 (commented):
FYI, this PR and all future ones introducing a new filter will need to install the filter via enterprise/settings/common.py

enterprise/filters/grades.py
```diff
@@ -0,0 +1,47 @@
...
+
+                {
+                    "context": <enriched context>,
+                    "user_id": <unchanged>,
+                    "course_id": <unchanged>,
+                }
+        """
+        uuids = EnterpriseCourseEnrollment.get_enterprise_uuids_with_user_and_course(user_id, course_id)
```
pwnage101, line 50: Please start this pipeline step with a log line to mirror all the other enterprise pipeline steps and improve stage/prod testing experience. Examples:
- https://github.com/openedx/edx-enterprise/blob/master/enterprise/filters/accounts.py#L54-L58
- https://github.com/openedx/edx-enterprise/blob/master/enterprise/filters/dashboard.py#L37-L42

Be sure to not log PII.  In this case, probably just the user_id and course_id are sufficient (exclude the context to keep log size down).

pwnage101 (commented):
✅  LGTM assuming you add the log line.

kiram15 (approved):

PR #2548, opened 2026-03-05
feat: add several courseware access related pipeline steps

pwnage101:
This commit replaces most enterprise-specific logic serving the purpose of conditionally blocking courseware access, often redirecting to a DSC page.

Enterprise steps added:
- EnterpriseStartDateAccessFailureStep supplies an enterprise-specific start-date error for enterprise learners.
- ActiveEnterpriseCheckStep denies access when the learner's active EnterpriseCustomer differs from the enrollment's customer.

Consent steps added:
- DataSharingConsentRedirectStep supplies a redirect URL to collect data sharing consent when it is required.
- DataSharingConsentCourseAccessStep denies access when consent is required.

ENT-11544

---

Blocked by:
* https://github.com/openedx/openedx-filters/pull/336

Blocks:
* https://github.com/edx/edx-platform/pull/338
* https://github.com/edx/edx-platform/pull/339
* https://github.com/edx/edx-platform/pull/340
* https://github.com/openedx/openedx-platform/pull/38102

### DIFF
--- a/consent/helpers.py
+++ b/consent/helpers.py
@@ -2,11 +2,162 @@
 Helper functions for the Consent application.
 """
 
+import logging
+from urllib.parse import urlencode
+
+from edx_django_utils.cache import TieredCache
+
 from django.apps import apps
+from django.conf import settings
+from django.contrib.sites.models import Site
+from django.urls import reverse
 
 from consent.models import ProxyDataSharingConsent
 from enterprise.api_client.discovery import get_course_catalog_api_service_client
-from enterprise.utils import get_enterprise_customer
+from enterprise.utils import get_active_enterprise_customer_user, get_enterprise_customer
+
+# ENT-11576: CONSENT_FAILED_PARAMETER, ConsentApiClient, enterprise_customer_uuid_for_request,
+# and get_data_consent_share_cache_key will be migrated from the platform's enterprise_support
+# module into edx-enterprise, eliminating these cross-boundary imports.
+try:
+    from openedx.features.enterprise_support.api import (
+        CONSENT_FAILED_PARAMETER,
+        ConsentApiClient,
+        enterprise_customer_uuid_for_request,
+    )
+    from openedx.features.enterprise_support.utils import get_data_consent_share_cache_key
+except ImportError:
+    CONSENT_FAILED_PARAMETER = 'consent_failed'
+    ConsentApiClient = None
+    enterprise_customer_uuid_for_request = None
+    get_data_consent_share_cache_key = None
+
+LOGGER = logging.getLogger(__name__)
+
+
+def consent_needed_for_course(request, user, course_id, enrollment_exists=False):
+    """
+    Determine whether ``user`` must grant data-sharing consent before accessing ``course_id``.
+    """
+    # Consent is never required if the enterprise feature is disabled.
+    if not getattr(settings, 'ENABLE_ENTERPRISE_INTEGRATION', False):
+        return False
+
+    LOGGER.info(
+        "[ENTERPRISE DSC] Determining if user [%s] must consent to data sharing for course [%s]",
+        user.username, course_id,
+    )
+
+    active_enterprise_learner_info = get_active_enterprise_customer_user(user)
+    if not active_enterprise_learner_info:
+        LOGGER.info(
+            "[ENTERPRISE DSC] Consent from user [%s] is not needed for course [%s]. "
+            "The user is not linked to an enterprise.",
+            user.username, course_id,
+        )
+        return False
+
+    active_enterprise_customer = active_enterprise_learner_info.enterprise_customer
+
+    consent_cache_key = get_data_consent_share_cache_key(
+        user.id, course_id, str(active_enterprise_customer.uuid),
+    )
+    cached = TieredCache.get_cached_response(consent_cache_key)
+    if cached.is_found and cached.value == 0:
+        LOGGER.info(
+            "[ENTERPRISE DSC] Consent from user [%s] is not needed for course [%s]. "
+            "The DSC cache was checked and the value was 0.",
+            user.username, course_id,
+        )
+        return False
+
+    if not active_enterprise_customer.enable_data_sharing_consent:
+        LOGGER.info(
+            "[ENTERPRISE DSC] DSC is disabled for enterprise customer [%s]. "
+            "Consent from user [%s] is not needed for course [%s]",
+            active_enterprise_customer.slug, user.username, course_id,
+        )
+        TieredCache.set_all_tiers(consent_cache_key, 0, settings.DATA_CONSENT_SHARE_CACHE_TIMEOUT)
+        return False
+
+    current_enterprise_uuid = enterprise_customer_uuid_for_request(request)
+    if str(current_enterprise_uuid) != str(active_enterprise_customer.uuid):
+        LOGGER.info(
+            '[ENTERPRISE DSC] Enterprise mismatch. USER: [%s], RequestEnterprise: [%s], '
+            'LearnerEnterprise: [%s]',
+            user.username, current_enterprise_uuid, active_enterprise_customer.uuid,
+        )
+        TieredCache.set_all_tiers(consent_cache_key, 0, settings.DATA_CONSENT_SHARE_CACHE_TIMEOUT)
+        return False
+
+    enterprise_domain = Site.objects.get(domain=active_enterprise_customer.site.domain)
+    if enterprise_domain != request.site:
+        LOGGER.info(
+            '[ENTERPRISE DSC] Site mismatch. USER: [%s], RequestSite: [%s], '
+            'LearnerEnterpriseDomain: [%s]',
+            user.username, request.site, enterprise_domain,
+        )
+        TieredCache.set_all_tiers(consent_cache_key, 0, settings.DATA_CONSENT_SHARE_CACHE_TIMEOUT)
+        return False
+
+    client = ConsentApiClient(user=request.user)
+    consent_required = client.consent_required(
+        username=user.username,
+        course_id=course_id,
+        enterprise_customer_uuid=current_enterprise_uuid,
+        enrollment_exists=enrollment_exists,
+    )
+    if not consent_required:
+        LOGGER.info(
+            "[ENTERPRISE DSC] Consent from user [%s] is not needed for course [%s]. "
+            "The user's current enterprise does not require data sharing consent.",
+            user.username, course_id,
+        )
+        TieredCache.set_all_tiers(consent_cache_key, 0, settings.DATA_CONSENT_SHARE_CACHE_TIMEOUT)
+        return False
+
+    LOGGER.info(
+        "[ENTERPRISE DSC] Consent from user [%s] is needed for course [%s]. "
+        "The user's current enterprise requires data sharing consent, and it has not been given.",
+        user.username, course_id,
+    )
+    return True
+
+
+def get_enterprise_consent_url(request, course_id, user=None, return_to=None, enrollment_exists=False, source='lms'):
+    """
+    Build a URL to redirect the user to the data-sharing consent page for a specific course.
+
+    Arguments:
+        request: Django request object.
+        course_id: Course key/identifier string.
+        user: user to check for consent. If None, uses ``request.user``.
+        return_to: url name for the page to return to after consent is granted; defaults to
+            ``request.path``.
+        enrollment_exists: forwarded to ``consent_needed_for_course``.
+        source: opaque string identifying the caller, recorded on the consent URL.
+    """
+    user = user or request.user
+    LOGGER.info(
+        'Getting enterprise consent url for user [%s] and course [%s].',
+        user.username,
+        course_id,
+    )
+    if not consent_needed_for_course(request, user, course_id, enrollment_exists=enrollment_exists):
+        return None
+    return_path = request.path if return_to is None else reverse(return_to, args=(course_id,))
+    url_params = {
+        'enterprise_customer_uuid': enterprise_customer_uuid_for_request(request),
+        'course_id': course_id,
+        'source': source,
+        'next': request.build_absolute_uri(return_path),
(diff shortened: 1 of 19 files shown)


enterprise/settings/common.py
```diff
@@ -20,6 +20,20 @@
         "fail_silently": False,
         "pipeline": ["enterprise.filters.grades.GradeEventContextEnricher"],
     },
+    "org.openedx.learning.course.start_date.validation_failed.v1": {
+        "fail_silently": True,
```
kiram15: I thought we agreed that fail_silently should be False, why the change?

enterprise/filters/courseware.py
```diff
@@ -0,0 +1,138 @@
...
+        if enterprise_customer and enterprise_customer.get('enable_learner_portal'):
+            has_enterprise_enrollment = EnterpriseCourseEnrollment.objects.filter(
+                course_id=course_key,
+                enterprise_customer_user__user_id=user.id,
+                enterprise_customer_user__enterprise_customer__uuid=enterprise_customer['uuid'],
+            ).exists()
+            if has_enterprise_enrollment:
+                if start_date == DEFAULT_START_DATE:
+                    developer_message = (
+                        "Course has not started, and the learner is enrolled via an enterprise subsidy."
```
kiram15: Do we have to worry about translations for the static messages? Or should these be extracted into a constants file?

tests/filters/test_courseware.py
```diff
@@ -0,0 +1,251 @@
...
+        request.user.is_authenticated = True
+        mock_get_request.return_value = request
+        step = self._make_step()
+        course_key = MagicMock()
+        start_date = datetime(2026, 9, 1)
+        with pytest.raises(CourseStartDateValidationFailed.OverrideStartDateError) as exc_info:
+            step.run_filter(
+                course_key=course_key,
+                start_date=start_date,
+                display_error_to_user=True,
```
kiram15: Either add the param `display_error_to_user` to the filter definition, or remove it from these tests

consent/helpers.py
```diff
@@ -2,11 +2,169 @@
...
+        course_id: Course key/identifier string.
+        user: user to check for consent. If None, uses ``request.user``.
+        return_to: url name for the page to return to after consent is granted; defaults to
+            ``request.path``.
+        enrollment_exists: forwarded to ``consent_needed_for_course``.
+        source: opaque string identifying the caller, recorded on the consent URL.
+    """
+    user = user or request.user
+    LOGGER.info(
+        'Getting enterprise consent url for user [%s] and course [%s].',
```
kiram15, line 142: The other logs have [ENTERPRISE DSC] as a prefix, should add in here to standardize

enterprise/filters/courseware.py
pwnage101: Supporting translation is out of scope, and I'd go so far as to say that the backend shouldn't be responsible for translations at all.  That's why we rely on the error code, a universal langauge.  Why this logic doesn't just rely on the error code alone, and shove messaging into the frontend is unfortunate, but that's just how it was designed.

consent/helpers.py
pwnage101: If I was just writing this from scratch, then I'd agree, but:

1. This is just migrating existing logic, and I'm slightly more partial to just keeping things the way they are.
2. My preference is probably to just remove ALL prefixes instead of add them everywhere because the logger already includes the module path, so adding it to the message part is redundant.

enterprise/utils.py
```diff
@@ -839,6 +850,69 @@ def get_enterprise_customer_user(user_id, enterprise_uuid):
...
+    and redirected to the enterprise learner portal when access is denied.
+
+    ``enterprise_customer_from_session_or_learner_data`` is used in lieu of direct model access
+    because it caches the customer data on the request session, making it safe to call on
+    frequently-hit views without incurring repeated database queries.
+    """
+    # enterprise_customer is either None (learner not linked to any customer) or a serialized
+    # EnterpriseCustomer representing the learner's active linked customer.
+    enterprise_customer = enterprise_customer_from_session_or_learner_data(request)
+    if not (enterprise_customer and enterprise_customer.get('enable_learner_portal')):
```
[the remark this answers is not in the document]
pwnage101: enterprise_customer_from_session_or_learner_data is actually not optional.  The entire enterprise plugin is ONLY compatible with edx/edx-platform or openedx/openedx-platform, which is guaranteed to provide enterprise_customer_from_session_or_learner_data until we remove it.

enterprise/filters/courseware.py
```diff
@@ -0,0 +1,112 @@
...
+
+        try:
+            active_ecu = EnterpriseCustomerUser.objects.get(user_id=user.id, active=True)
+            if enterprise_enrollments.filter(enterprise_customer_user=active_ecu).exists():
+                return {"user": user, "course_key": course_key}
+            active_enterprise_name = active_ecu.enterprise_customer.name
+        except (EnterpriseCustomerUser.DoesNotExist, EnterpriseCustomerUser.MultipleObjectsReturned):
+            # Ideally this should not happen — there should be exactly 1 active enterprise customer.
+            log.error("Multiple or No Active Enterprise found for the user %s.", user.id)
+            active_enterprise_name = "Incorrect"
```
marlonkeating, line 95: Can we have separate catch blocks for `DoesNotExist` and `MultipleObjectsReturned`?

consent/helpers.py
```diff
@@ -2,11 +2,166 @@
...
+    Determine whether ``user`` must grant data-sharing consent before accessing ``course_id``.
+    """
+    if not getattr(settings, 'ENABLE_ENTERPRISE_INTEGRATION', False):
+        return False
+    if (
+        ConsentApiClient is None
+        or enterprise_customer_uuid_for_request is None
+        or get_data_consent_share_cache_key is None
+    ):
+        return False
```
marlonkeating: nit: could combine these checks 
```suggestion
    if (
        not getattr(settings, 'ENABLE_ENTERPRISE_INTEGRATION', False)
        or ConsentApiClient is None
        or enterprise_customer_uuid_for_request is None
        or get_data_consent_share_cache_key is None
    ):
        return False
```

consent/helpers.py
```diff
@@ -2,11 +2,166 @@
...
+    if str(current_enterprise_uuid) != str(active_enterprise_customer.uuid):
+        LOGGER.info(
+            '[ENTERPRISE DSC] Enterprise mismatch. USER: [%s], RequestEnterprise: [%s], '
+            'LearnerEnterprise: [%s]',
+            user.username, current_enterprise_uuid, active_enterprise_customer.uuid,
+        )
+        TieredCache.set_all_tiers(consent_cache_key, 0, settings.DATA_CONSENT_SHARE_CACHE_TIMEOUT)
+        return False
+
+    enterprise_domain = Site.objects.get(domain=active_enterprise_customer.site.domain)
```
marlonkeating, line 93: Should we put a guard against no domain being found for the customer?

enterprise/filters/courseware.py
```diff
@@ -0,0 +1,112 @@
...
+        start_date: datetime,
+    ) -> dict:
+        request = get_current_request()
+        if request is None:
+            return {"course_key": course_key, "start_date": start_date}
+        user = request.user
+        if not user.is_authenticated:
+            return {"course_key": course_key, "start_date": start_date}
+        if not enterprise_learner_enrolled(request, course_key):
+            return {"course_key": course_key, "start_date": start_date}
```
marlonkeating: nit: could consolidate these two if's

enterprise/filters/courseware.py
pwnage101: For now I'm going to leave it as-is since this is a copy+paste from the original and I don't want to risk changing behavior.

consent/helpers.py
pwnage101: Addressed by removing the unnecessary None checks.

consent/helpers.py
pwnage101: Similarly to [my other comment](https://github.com/openedx/edx-enterprise/pull/2548#discussion_r3375459127), this is a copy+paste of original logic which I don't want to change too much.

enterprise/filters/courseware.py
```diff
@@ -0,0 +1,110 @@
...
+from opaque_keys.edx.keys import CourseKey
+from openedx_filters.filters import PipelineStep
+from openedx_filters.learning.filters import CourseStartDateValidationFailed, CoursewareAccessChecksRequested
+
+# Courses whose start date equals DEFAULT_START_DATE are considered "unscheduled" in the platform.
+# Messages for these courses omit the specific date and use a generic "has not started" wording.
+try:
+    from xmodule.course_metadata_utils import DEFAULT_START_DATE
+except ImportError:
+    DEFAULT_START_DATE = datetime(2040, 1, 1, tzinfo=ZoneInfo("UTC"))
```
iloveagent57, line 19: why not `datetime.max`?

enterprise/utils.py
```diff
@@ -839,6 +850,76 @@ def get_enterprise_customer_user(user_id, enterprise_uuid):
...
+    courseware gating (e.g. data-sharing consent enforcement or start-date error overrides)
+    and redirected to the enterprise learner portal when access is denied.
+
+    ``enterprise_customer_from_session_or_learner_data`` is used in lieu of direct model access
+    because it caches the customer data on the request session, making it safe to call on
+    frequently-hit views without incurring repeated database queries.
+    """
+    # enterprise_customer is either None (learner not linked to any customer) or a serialized
+    # EnterpriseCustomer representing the learner's active linked customer.
+    enterprise_customer = enterprise_customer_from_session_or_learner_data(request)
```
iloveagent57, line 891: `enterprise_customer_from_session_or_learner_data` could also be none here, right? Maybe that should default to a no-op lambda or something?

iloveagent57 (approved):
two small nits, looks good otherwise!

enterprise/filters/courseware.py
pwnage101: I'm absolutely certain there would have been a better way of choosing this value, but I'm not about to rock the boat.  This value has been a hard-coded date in the core platform logic for more than 10 years.  Here's some context from the last time it got bumped: https://github.com/openedx/openedx-platform/pull/38277#issuecomment-4179615691

enterprise/utils.py
pwnage101: I've decided that I wouldn't pollute the business logic with `imported_platform_thing is None` checks for two reasons:
1. The plan is to migrate most of these imported functions into edx-enterprise soon, so I'd need to follow up later and remove all the temporary checks.
2. edx-enterprise is not compatible with anything other than {edx,openedx}-platform.  Installing edx-enterprise into a different project should cause crashing, and that's expected behavior.

pwnage101: Ah, I glossed over your proposal to make it a no-op lambda.  That still means it wouldn't crash if installed outside of the platform, but crashing is kind of the behavior I want.

consent/helpers.py
```diff
@@ -2,11 +2,162 @@
...
+LOGGER = logging.getLogger(__name__)
+
+
+def consent_needed_for_course(request, user, course_id, enrollment_exists=False):
+    """
+    Determine whether ``user`` must grant data-sharing consent before accessing ``course_id``.
+    """
+    # Consent is never required if the enterprise feature is disabled.
+    if not getattr(settings, 'ENABLE_ENTERPRISE_INTEGRATION', False):
+        return False
```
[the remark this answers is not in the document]
pwnage101: This is intentional.  edx-enterprise does not support being installed outside openedx-platform or edx-platform.

enterprise/filters/courseware.py
```diff
@@ -0,0 +1,110 @@
...
+            active_enterprise_name = "Incorrect"
+
+        enrollment_enterprise_name = (
+            enterprise_enrollments.first().enterprise_customer_user.enterprise_customer.name
+        )
+        user_message = (
+            f"You are enrolled in this course with '{enrollment_enterprise_name}'. However, you are "
+            f"currently logged in as a '{active_enterprise_name}' user. Please log in with "
+            f"'{enrollment_enterprise_name}' to access this course."
+        )
```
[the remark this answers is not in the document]
pwnage101: valid concern, but again I'm not interested in changing behavior of legacy code.

enterprise/utils.py
```diff
@@ -839,6 +850,76 @@ def get_enterprise_customer_user(user_id, enterprise_uuid):
...
+    if not getattr(settings, 'ENABLE_ENTERPRISE_INTEGRATION', False) or not user.is_authenticated:
+        return None
+    EnterpriseCustomerUser = apps.get_model('enterprise', 'EnterpriseCustomerUser')
+    try:
+        return EnterpriseCustomerUser.objects.get(user_id=user.id, active=True)
+    except EnterpriseCustomerUser.DoesNotExist:
+        LOGGER.info(
+            "Active EnterpriseCustomerUser for user [%s] does not exist", user.username,
+        )
+        return None
```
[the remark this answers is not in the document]
pwnage101: There's other guardrails in place aside from database-level constraints which limit only one active customer for a user.  While this code could be more defensive, I'm also not interested in change the behavior of legacy code since that would expand the scope of this work.

consent/helpers.py
```diff
@@ -2,11 +2,162 @@
...
+
+    enterprise_domain = Site.objects.get(domain=active_enterprise_customer.site.domain)
+    if enterprise_domain != request.site:
+        LOGGER.info(
+            '[ENTERPRISE DSC] Site mismatch. USER: [%s], RequestSite: [%s], '
+            'LearnerEnterpriseDomain: [%s]',
+            user.username, request.site, enterprise_domain,
+        )
+        TieredCache.set_all_tiers(consent_cache_key, 0, settings.DATA_CONSENT_SHARE_CACHE_TIMEOUT)
+        return False
```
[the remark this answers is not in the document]
pwnage101: Legacy issue from migrated code.  The current task is to migrate the code, not fix all the bugs in it.

PR #2551, opened 2026-03-05
feat: add pipeline steps for logistration filters

pwnage101:
Implement steps for five logistration filters replacing logic in the `user_authn` platform module, and register them in `ENTERPRISE_FILTERS_CONFIG`:

- LogistrationContextEnricher (for LogistrationContextRequested)
- LoginFormEnterpriseOverrides (for LoginFormTPAOverridesRequested)
- RegistrationFormEnterpriseOverrides (for RegistrationFormTPAOverridesRequested)
- LogistrationCookieSetter (for LogistrationResponseRendered)
- PostLoginEnterpriseRedirect (for PostLoginRedirectURLRequested)

[ENT-11568](https://2u-internal.atlassian.net/browse/ENT-11568)

---

Integration Testing in 2U Devstack
---

Setup:

1. Launch the required containers:
   - `make dev.up.lms+enterprise-catalog+frontend-app-account+frontend-app-authn`
2. Add this line to `/etc/hosts`:
   - `127.0.0.1 edx.devstack.keycloak`
3. Create the `user_authn.enable_enterprise_redirect_to_authn` waffle flag and enable it globally.

Control Test:

4. Set up your devstack to run the following branches:
   - lms: release-ulmo
   - edx-enterprise: master
5. from edx-enterprise, run two commands:
   - `make dev.provision.keycloak`
   - `./scripts/provision-integration-test-ENT-11568.sh`
6. Follow the script output to conduct entire integration test suite.

Integration Test:

7. Reconfigure your devstack to switch to the new code:
   - lms: pwnage101/ENT-11568-edx
     - `pip install -e /edx/src/edx-enterprise/`
     - `pip install -e /edx/src/openedx-filters/`
   - edx-enterprise: pwnage101/ENT-11568
   - openedx-filters: pwnage101/ENT-11568
8. Repeat steps 5-6 and confirm everything works exactly the same.

---

Related:
* https://github.com/openedx/edx-enterprise/pull/2662
* https://github.com/openedx/openedx-filters/pull/337
* https://github.com/openedx/edx-enterprise/pull/2551
* https://github.com/edx/edx-platform/pull/407
* https://github.com/openedx/openedx-platform/pull/38105
* https://github.com/edx/devstack/pull/227

### DIFF
--- a/docs/saml_testing.rst
+++ b/docs/saml_testing.rst
@@ -12,10 +12,16 @@ Prerequisites
 -------------
 
 * A running `devstack <https://github.com/openedx/devstack>`_ environment with
-  the LMS container up.
+  the following containers up and running:
+
+  * ``make dev.up.lms+enterprise-catalog+frontend-app-account+frontend-app-authn``
+
 * The edx-enterprise branch you want to test installed as an editable package
-  inside the LMS container (``/edx/src/edx-enterprise``).
-* Docker Compose (the ``docker compose`` CLI plugin).
+  inside the LMS container (``pip install -e /edx/src/edx-enterprise``).
+
+* The following line added to ``/etc/hosts`` on your host machine:
+
+  * ``127.0.0.1 edx.devstack.keycloak``
 
 Starting Keycloak
 -----------------
@@ -26,10 +32,8 @@ From the **edx-enterprise** repository root:
 
    $ make dev.up.keycloak
 
-This starts a Keycloak 26.x container (``edx.devstack.keycloak``) on the
-devstack Docker network, exposed at ``http://localhost:8080``.  Keycloak data is
-persisted in a Docker volume (``keycloak_data``) so the container can be stopped
-and restarted without losing state.
+This starts a Keycloak container (``edx.devstack.keycloak``) on the
+devstack Docker network, exposed at ``http://localhost:8080``.
 
 Provisioning
 ------------
@@ -40,45 +44,20 @@ Provisioning configures **both** Keycloak and the LMS in a single step:
 
    $ make dev.provision.keycloak
 
-Under the hood this runs two commands:
+Under the hood this accomplishes the following:
 
-1. ``keycloak-config-cli`` imports every realm definition in
-   ``keycloak-realms/`` into Keycloak.  Each file is one tenant realm (currently
-   ``gryffindor`` and ``slytherin``), each with a SAML client and two test users.
-2. ``provision-tpa.py`` runs inside the LMS container and, for each tenant,
-   creates the matching ``SAMLProviderConfig``, the ``EnterpriseCustomer`` link,
-   branding (logo + colors), and a login-flow LMS learner account.  A single
-   shared ``SAMLConfiguration`` (the LMS service-provider config) is created once.
+* [Keycloak] Provisions two "realms" within Keycloak, ``gryffindor`` and ``slytherin``, each representing an IdP.
 
-A "tenant" is one Keycloak realm plus one enterprise customer.  The realm name,
-the SAML slug, the ``provider_id`` (``saml-<name>``), and the enterprise slug are
-all the same arbitrary token (e.g. ``gryffindor``), so one memorable name
-identifies everything about the tenant.  Adding a tenant means dropping a new
-``keycloak-realms/<name>.json`` and adding a matching entry to the ``TENANTS``
-list in ``provision-tpa.py``.
+* [Keycloak] Provisions several Keycloak users within those realms.
 
-Shared configuration (the Keycloak URL, the LMS entity ID, the ACS URL, and the
-attribute OIDs) plus each tenant's SSO usernames live in
-``keycloak-devstack.env``.  The usernames are the single source of truth: the
-realm JSON substitutes them via ``$(env:...)`` and ``provision-tpa.py`` reads the
-same variables, so a username is defined in exactly one place.
+* [LMS] provisions a SAMLConfiguration record to globally enable SAML auth.
 
-The examples below use the ``gryffindor`` tenant; ``slytherin`` behaves
-identically -- substitute its name to test tenant isolation.
+* [LMS] provisions matching LMS records for each of the Keycloak realms:
 
-Host setup
-----------
-
-The LMS redirects to Keycloak using the Docker hostname
-``edx.devstack.keycloak``.  Your browser needs to resolve that name to
-localhost.
-
-Add this line to ``/etc/hosts`` on the machine where your browser runs (your
-laptop, **not** a remote codespace):
-
-.. code-block:: text
-
-   127.0.0.1 edx.devstack.keycloak
+  * EnterpriseCustomer
+  * SAMLProviderConfig + EnterpriseCustomerIdentityProvider (to link the enterprise customer with the Keycloak IdP)
+  * EnterpriseCustomerBrandingConfiguration (to give the login/registration pages a distinctive look)
+  * User + EnterpriseCustomerUser (to provide LMS-side users corresponding to the IdP side users).
 
 Testing the SAML login flow
 ----------------------------
@@ -92,10 +71,10 @@ Testing the SAML login flow
 
 3. Log in with the test credentials:
 
-   =========  =============================
+   =========  ======================
    Username   ``gryffindor_learner``
    Password   ``testpass``
-   =========  =============================
+   =========  ======================
 
 4. Validate that you were **not** prompted to log into the existing LMS user.
    The ``enterprise_associate_by_email`` pipeline step should discover that the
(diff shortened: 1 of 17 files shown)


enterprise/filters/logistration.py
```diff
@@ -0,0 +1,290 @@
...
+    third-party-auth adjustments. It also injects the enterprise slug login URL and the
+    enterprise-enabled flag consumed by the logistration page's JS.
+    """
+
+    def run_filter(self, context: dict[str, Any]) -> dict[str, Any]:  # pylint: disable=arguments-differ
+        """
+        Enrich context with enterprise customer data.
+        """
+        request = get_current_request()
+        enterprise_customer = enterprise_customer_for_request(request)
```
kiram15, line 80: Should we have a check here for if `enterprise_customer_for_request` is None?

enterprise/filters/logistration.py
```diff
@@ -0,0 +1,290 @@
...
+    mirroring the original platform behavior: it sets the ``experiments_is_enterprise``
+    cookie from ``context['enable_enterprise_sidebar']`` and deletes the enterprise
+    customer cookie so that subsequent requests show the default login page.
+    """
+
+    def run_filter(self, response: Any, context: dict[str, Any]) -> dict[str, Any]:  # pylint: disable=arguments-differ
+        """
+        Apply enterprise cookie handling to the logistration response.
+        """
+        handle_enterprise_cookies_for_logistration(get_current_request(), response, context)
```
kiram15, line 140: Should we have a check for if `handle_enterprise_cookies_for_logistration` is None?

enterprise/filters/logistration.py
```diff
@@ -0,0 +1,290 @@
...
+                #
+                # NOTE: [2026-07-22] This new logic to derive skip_registration_optional_checkboxes
+                # is technically different than the legacy platform logic. Historically,
+                # current_provider was ignored and a *possibly different* provider config was looked
+                # up by directly querying the database. Most or all of the time, that looked-up
+                # provider config was the same as current_provider, but it was technically possible
+                # for it to be different (e.g., the looked-up provider config happened to be a
+                # disabled version). The new logic below is arguably more correct (and far simpler)
+                # because we make no effort to re-lookup the provider config. We just use the
+                # config already in-use by the current TPA pipeline.
```
kiram15: I think this deserves it's own explicit test cases if it does change behavior, even if its more correct now.

enterprise/filters/logistration.py
pwnage101: All platform imports are intentionally assumed to exist.  Platform is a hard requirement for edx-enterprise to be installed into.  Furthermore, all enterprise_support imports will soon migrate into this repo and I don't want to go back later to a bunch of None checks and rip them out.

enterprise/filters/logistration.py
pwnage101: All platform imports are intentionally assumed to exist.  Platform is a hard requirement for edx-enterprise to be installed into.  Furthermore, all enterprise_support imports will soon migrate into this repo and I don't want to go back later to a bunch of None checks and rip them out.

scripts/provision-integration-test-ENT-11568.sh
```diff
@@ -1,17 +1,11 @@
 #!/usr/bin/env bash
```
iloveagent57, line 1: I don't mind leaving references to the ticket number in docs, but this filename is a bit weird. Maybe `logistration-pipeline-integration-test.sh`?

iloveagent57 (approved):

scripts/provision-integration-test-ENT-11568.sh
pwnage101: My goal is to polish and document as much official logic as possible in official locations:
* enterprise/devstack_api.py
* enterprise/roles_api.py
* provision-tpa.py

But stuff in scripts/ are one-time use and not intended for long-term support.  The other script, `provision-integration-test-ENT-11544.sh` for a previous ticket does not even work anymore because so much of devstack_api.py has changed, but there's no problem because it's not a supported feature and not "on our books" as a maintenance liability.

The ticket number in the script name reinforces the temporariness and unsupported nature of the script.

docs/saml_testing.rst
marlonkeating, line 1: nit: Might we want to mention here or elsewhere that we should add an entry to the `.ssh/config` file for keycloak?
For example, I needed to add `LocalForward 8080 localhost:8080` to mine.
That said, if this is purely a devstack requirement that isn't applicable on other environments like Codespaces, then this probably isn't the best place for that documentation.

enterprise/filters/logistration.py
pwnage101: Reasonable.  Added a test: `test_optional_checkboxes_flag_read_only_from_current_provider`

enterprise/filters/logistration.py
```diff
@@ -0,0 +1,269 @@
...
+    with multiple enterprises otherwise return the next url.
+    """
+
+    def run_filter(  # pylint: disable=arguments-differ
+        self, redirect_url: str, user: Any, next_url: str,
+    ) -> dict[str, Any]:
+        """
+        Return enterprise selection page URL if user is associated with multiple enterprises.
+        """
+        response = get_enterprise_learner_data_from_api(user)
```
marlonkeating: nit: more specific variable name
```suggestion
        learner_data = get_enterprise_learner_data_from_api(user)
```

marlonkeating (approved):

docs/saml_testing.rst
pwnage101: Ah, yes I did not test with Hosted Devstack, only Codespaces which uses automatic forwarding.  Come to think of it, port 8080 is probably not the best port due to its high potential for conflicts.

Since the .ssh/config file contents are only documented in Confluence as part of Hosted Devstack docs, I will not pollute this repo with any mention of it.

enterprise/filters/logistration.py
```diff
@@ -0,0 +1,299 @@
...
+                current_provider
+                and current_provider.skip_registration_form
+                and enterprise_customer_for_request(get_current_request())
+        ):
+            log.info(
+                "RegistrationFormEnterpriseOverrides running: provider=%s",
+                getattr(current_provider, 'provider_id', None),
+            )
+            field_overrides = current_provider.get_register_form_data(
+                running_pipeline.get('kwargs')
```
kramakrushna: Could we guard running_pipeline, running_pipeline.get('kwargs'), and details here?  Right now, this assumes the TPA pipeline always provides kwargs -> details, but if a provider or pipeline shape is missing/changed, running_pipeline.get('kwargs') or details.get('email') will raise. A small guard would make this safer in edge cases.”
**Why this matters:** this is a real runtime crash candidate on login flows with unexpected pipeline data.

enterprise/filters/logistration.py
```diff
@@ -0,0 +1,299 @@
...
+        if response and len(response) > 1:
+            log.info(
+                "PostLoginEnterpriseRedirect running: user_id=%s linked to %s enterprises",
+                user.id,
+                len(response),
+            )
+            # Check to see if the destination has an enterprise in it. In this case if user is associated
+            # with that enterprise, activate that enterprise and bypass the selection page.
+            if re.match(_enterprise_enrollment_url_regex(), urllib.parse.unquote(redirect_url)):
+                enterprise_in_url = re.search(UUID4_REGEX, redirect_url).group(0)
```
kramakrushna: This branch assumes re.search(UUID4_REGEX, redirect_url) always succeeds once the re.match(...) check passes, but that’s brittle.  If the URL is encoded slightly differently or the regex changes, re.search(...). group(0) will raise.
Could we extract the UUID from the matched URL more defensively, or handle the no-match case explicitly?

enterprise/filters/logistration.py
```diff
@@ -0,0 +1,299 @@
...
+                # up by directly querying the database. Most or all of the time, that looked-up
+                # provider config was the same as current_provider, but it was technically possible
+                # for it to be different (e.g., the looked-up provider config happened to be a
+                # disabled version). The new logic below is arguably more correct (and far simpler)
+                # because we make no effort to re-lookup the provider config. We just use the
+                # config already in-use by the current TPA pipeline.
+                if (
+                        field_name == 'marketing_emails_opt_in'
+                        and getattr(current_provider, 'skip_registration_optional_checkboxes', False)
+                ):
```
kramakrushna, line 244: RegistrationFormEnterpriseOverrides intentionally changes how skip_registration_optional_checkboxes is resolved compared with the legacy platform logic. That seems reasonable, but it should be called out more explicitly since it changes behavior in an edge case.

kramakrushna (commented):
Overall this looks like a solid extraction of the enterprise logistration behavior into openedx-filters, and the new tests/docs are helpful. I have three main concerns before merge:

1. LoginFormEnterpriseOverrides assumes running_pipeline -> kwargs -> details always exists; if that shape changes or is missing, this can crash the login flow.

2. PostLoginEnterpriseRedirect relies on re.search(UUID4_REGEX, redirect_url).group(0) after a regex match; this is brittle and could raise if the URL format changes or is encoded differently.
3. RegistrationFormEnterpriseOverrides intentionally changes how skip_registration_optional_checkboxes is resolved compared with the legacy platform logic. That seems reasonable, but it should be called out more explicitly since it changes behavior in an edge case.

kramakrushna (approved):

enterprise/filters/logistration.py
pwnage101: I'll implement more robust extraction using named groups, along these lines:

```diff
-    return fr'/enterprise/{UUID4_REGEX}/course/{settings.COURSE_KEY_REGEX}/enroll'
+    return fr'/enterprise/(?P<enterprise_uuid>{UUID4_REGEX})/course/{settings.COURSE_KEY_REGEX}/enroll'
...
-            if re.match(_enterprise_enrollment_url_regex(), urllib.parse.unquote(redirect_url)):
-                enterprise_in_url = re.search(UUID4_REGEX, redirect_url).group(0)
+            url_match = re.match(_enterprise_enrollment_url_regex(), urllib.parse.unquote(redirect_url))
+            if url_match:
+                enterprise_in_url = url_match.group('enterprise_uuid')
```

enterprise/filters/logistration.py
pwnage101: Instead of a guard, I'll approach this using contract enforcement via static type checking in edx-platform CI so that we can catch any issues upstream before they cause errors downstream in the pipeline steps.

enterprise/filters/logistration.py
pwnage101: There's already a very detailed and explicit code comment above this line, explaining the divergence and rational.

enterprise/settings/test.py
```diff
@@ -184,6 +184,7 @@ def root(*args):
 # These are standard regexes for pulling out info like course_ids, usage_ids, etc.
 COURSE_KEY_PATTERN = r'(?P<course_key_string>[^/+]+(/|\+)[^/+]+(/|\+)[^/?]+)'
 COURSE_ID_PATTERN = COURSE_KEY_PATTERN.replace('course_key_string', 'course_id')
+COURSE_KEY_REGEX = COURSE_KEY_PATTERN.replace('P<course_key_string>', ':')
```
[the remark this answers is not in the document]
pwnage101: you are mistaken.

PR #2552, opened 2026-03-05
feat: add DashboardContextEnricher pipeline step for student dashboard filter

pwnage101:
Jira: [ENT-11569](https://2u-internal.atlassian.net/browse/ENT-11569)

Summary: 
A DashboardContextEnricher pipeline step is defined in enterprise/filters/dashboard.py and injects all enterprise dashboard context keys (enterprise_message, consent_required_courses, is_enterprise_user, and enterprise portal keys). Unit tests cover the pipeline step with enterprise and non-enterprise users.


enterprise/filters/dashboard.py
```diff
@@ -0,0 +1,59 @@
...
+        request = context.get('request')
+        user = context.get('user')
+        course_enrollments = context.get('course_enrollment_pairs', [])
+
+        if user is None:
+            return {'context': context, 'template_name': template_name}
+
+        try:
+            enterprise_message = get_dashboard_consent_notification(request, user, course_enrollments)
+        except Exception:  # pylint: disable=broad-except
```
pwnage101: AI added these exception handlers, but I'm not really a fan of them.  If we didn't catch exceptions before, let's not just randomly start doing it, especially if we don't even know the specific exception and are forced to disable `broad-except` which is a bad code smell.  Furthermore, swallowing the exception is probably extra bad for visibility of regressions during local/stage testing.

enterprise/filters/dashboard.py
```diff
@@ -0,0 +1,59 @@
...
+
+class DashboardContextEnricher(PipelineStep):
+    """
+    Enrich the student dashboard context with enterprise-specific data.
+
+    Injects: enterprise_message, consent_required_courses, is_enterprise_user, and
+    enterprise learner portal context keys.
+    """
+
+    def run_filter(self, context, template_name):  # pylint: disable=arguments-differ
```
pwnage101: Nit: I feel like the function signature should have type hints, at least the inputs for now:
```suggestion
    def run_filter(self, context: dict[str, Any], template_name: str):  # pylint: disable=arguments-differ
```

Just copy the types from the filter itself: https://github.com/openedx/openedx-filters/blob/d517006b0f61050f027de90459e50e2745825992/openedx_filters/learning/filters.py#L829C34-L829C48

CHANGELOG.rst
```diff
@@ -17,9 +17,12 @@ Unreleased
 ----------
 * nothing unreleased
 
-[7.0.4] - 2026-04-15
+[7.0.5] - 2026-04-17
```
iloveagent57: there's gonna be a merge conflict on this, as I recently merged in a 7.0.5 release

iloveagent57 (approved):

PR #2553, opened 2026-03-05
feat: add EnterpriseEnrollmentViewProcessor pipeline step for CourseEnrollmentStarted filter

pwnage101:
## Description
An EnterpriseEnrollmentViewProcessor pipeline step is defined in enterprise/filters/enrollment.py for the new openedx-filter CourseEnrollmentViewStarted ([here](https://github.com/openedx/openedx-filters/pull/363)). 

The pipeline step checks whether the enrollment user is linked to an enterprise customer, and if so, calls the enterprise and consent API clients to post the enterprise course enrollment and provide consent.

## Jira
[ENT-11570](https://2u-internal.atlassian.net/browse/ENT-11570)

## Testing instructions
Requires all three branches installed together (follow instructions [here](https://2u-internal.atlassian.net/wiki/spaces/SOL/pages/3710353504/Enterprise+Plugin+Ticket+Runbook#Devstack-setup-for-testing))

Use this branch, openedx-filters branch https://github.com/openedx/openedx-filters/pull/363, and edx-platform branch [kiram15/ENT-11570 ](https://github.com/edx/edx-platform/pull/306/changes)

Needs ENABLE_ENTERPRISE_INTEGRATION=True, an API key, an EnterpriseCustomer, and a learner linked to it via EnterpriseCustomerUser.

In the LMS shell, POST to the enrollment endpoint as the worker, patching the two API clients to assert they fire:

```
from unittest import mock
from rest_framework.test import APIClient

EC_UUID = "<enterprise-uuid>"
CID = "<course-id>"
worker = APIClient(); worker.credentials(HTTP_X_EDX_API_KEY=settings.EDX_API_KEY)

with mock.patch("enterprise.filters.enrollment.EnterpriseApiServiceClient") as MEnt, \
     mock.patch("enterprise.filters.enrollment.ConsentApiServiceClient") as MCon:
    resp = worker.post("/api/enrollment/v1/enrollment",
        {"user": "<username>", "course_details": {"course_id": CID},
         "mode": "audit", "linked_enterprise_customer": EC_UUID}, format="json")
    print(resp.status_code)
    print(MEnt.return_value.post_enterprise_course_enrollment.call_args)
    print(MCon.return_value.provide_consent.call_args)
```

    
Expected: 200; the LMS log shows EnterpriseEnrollmentViewProcessor running: …; post_enterprise_course_enrollment is called with (username, course_id); provide_consent is called with username, course_id, and enterprise_customer_uuid.

I also personally tested these other scenarios 
- no linked_enterprise: 200, neither client called
- no API-key permission: 200, neither client called
- enterprise post fails: 400, consent not called, no enrollment created
- consent fails: 400, enterprise called once, no enrollment created

---

Related:
* https://github.com/openedx/edx-enterprise/pull/2553
* https://github.com/openedx/openedx-filters/pull/363
* https://github.com/edx/edx-platform/pull/306/changes

### DIFF
--- a/enterprise/api_client/xpert_ai.py
+++ b/enterprise/api_client/xpert_ai.py
@@ -28,7 +28,7 @@ def chat_completion(prompt, role):
     }
 
     body = {
-        'messages': [{'role': role, 'content': prompt},],
+        'messages': [{'role': role, 'content': prompt}],
         'client_id': settings.ENTERPRISE_ANALYSIS_CLIENT_ID,
         'system_message': settings.ENTERPRISE_ANALYSIS_SYSTEM_PROMPT
     }
--- a/enterprise/filters/enrollment.py
+++ b/enterprise/filters/enrollment.py
@@ -5,6 +5,7 @@ import logging
 from typing import Any
 
 from crum import get_current_request
+from opaque_keys.edx.keys import CourseKey
 from openedx_filters.filters import PipelineStep
 from openedx_filters.learning.filters import CourseEnrollmentViewStarted
 from requests.exceptions import HTTPError
@@ -39,30 +40,36 @@ class EnterpriseEnrollmentViewProcessor(PipelineStep):
     def run_filter(  # pylint: disable=arguments-differ
             self,
             user: AbstractBaseUser,
-            course_key: Any,
-            has_api_key_permissions: bool
+            course_key: CourseKey,
+            requester_is_backend_service: bool
     ) -> dict[str, Any]:
         """
         Post enterprise enrollment and consent if the user is an enterprise customer user.
         """
         log.info(
-            "EnterpriseEnrollmentViewProcessor running: user_id=%s, course_key=%s, api_permissions=%s",
+            "EnterpriseEnrollmentViewProcessor running: user_id=%s, course_key=%s, is_backend_service=%s",
             user.id,
             str(course_key),
-            has_api_key_permissions,
+            requester_is_backend_service,
         )
 
         request = get_current_request()
         linked_enterprise = None
         if request and hasattr(request, 'data'):
-            linked_enterprise = request.data.get("linked_enterprise_customer")
+            linked_enterprise = request.data.get('linked_enterprise_customer')
 
-        if linked_enterprise is None or not has_api_key_permissions:
-            return {
-                'user': user,
-                'course_key': course_key,
-                'has_api_key_permissions': has_api_key_permissions
-            }
+        result = {
+            'user': user,
+            'course_key': course_key,
+            'requester_is_backend_service': requester_is_backend_service,
+        }
+
+        if not linked_enterprise or not requester_is_backend_service:
+            return result
+
+        if EnterpriseApiServiceClient is None or ConsentApiServiceClient is None or EnterpriseApiException is None:
+            log.warning('enterprise_support.api is unavailable; skipping enterprise enrollment side effects')
+            return result
 
         username = user.username
         course_id = str(course_key)
@@ -96,5 +103,5 @@ class EnterpriseEnrollmentViewProcessor(PipelineStep):
         return {
             'user': user,
             'course_key': course_key,
-            'has_api_key_permissions': has_api_key_permissions
+            'requester_is_backend_service': requester_is_backend_service
         }
--- a/enterprise/filters/enrollment.py
+++ b/enterprise/filters/enrollment.py
@@ -1,0 +1,100 @@
+"""
+Pipeline steps for the course enrollment filter.
+"""
+import logging
+from typing import Any
+
+from crum import get_current_request
+from openedx_filters.filters import PipelineStep
+from openedx_filters.learning.filters import CourseEnrollmentViewStarted
+from requests.exceptions import HTTPError
+
+from django.contrib.auth.base_user import AbstractBaseUser
+
+log = logging.getLogger(__name__)
+
+try:
+    from openedx.features.enterprise_support.api import (
+        ConsentApiServiceClient,
+        EnterpriseApiException,
+        EnterpriseApiServiceClient,
+    )
+except ImportError:
+    ConsentApiServiceClient = None
+    EnterpriseApiException = None
+    EnterpriseApiServiceClient = None
+
+
+class EnterpriseEnrollmentViewProcessor(PipelineStep):
+    """
+    Enrollment pipeline step: notify enterprise API and record consent.
+
+    When an enterprise customer user enrolls in a course, this step calls the enterprise and consent
+    API clients to post the enrollment and provide consent on behalf of the enterprise customer.
+
+    This step is intended to be registered as a pipeline step for the
+    ``org.openedx.learning.course.enrollment.view.started.v1`` filter.
+    """
+
+    def run_filter(  # pylint: disable=arguments-differ
+            self,
+            user: AbstractBaseUser,
+            course_key: Any,
+            has_api_key_permissions: bool
+    ) -> dict[str, Any]:
+        """
+        Post enterprise enrollment and consent if the user is an enterprise customer user.
+        """
+        log.info(
+            "EnterpriseEnrollmentViewProcessor running: user_id=%s, course_key=%s, api_permissions=%s",
+            user.id,
+            str(course_key),
+            has_api_key_permissions,
+        )
+
+        request = get_current_request()
+        linked_enterprise = None
+        if request and hasattr(request, 'data'):
+            linked_enterprise = request.data.get("linked_enterprise_customer")
+
+        if linked_enterprise is None or not has_api_key_permissions:
+            return {
+                'user': user,
+                'course_key': course_key,
+                'has_api_key_permissions': has_api_key_permissions
+            }
+
+        username = user.username
+        course_id = str(course_key)
+
+        try:
+            EnterpriseApiServiceClient().post_enterprise_course_enrollment(username, course_id)
+        except EnterpriseApiException as exc:
+            log.exception(
+                "Failed to post enterprise course enrollment for user %s in course %s.",
+                username, course_id,
+            )
+            raise CourseEnrollmentViewStarted.PreventEnrollment(
+                f"Failed to post enterprise course enrollment for {username} in {course_id}."
+            ) from exc
+
+        try:
+            ConsentApiServiceClient().provide_consent(
+                username=username,
+                course_id=course_id,
+                enterprise_customer_uuid=str(linked_enterprise),
+            )
+        except HTTPError as exc:
+            log.exception(
+                "Failed to record enterprise consent for user %s in course %s.",
+                username, course_id,
+            )
+            raise CourseEnrollmentViewStarted.PreventEnrollment(
+                f"Failed to record enterprise consent for {username} in {course_id}."
+            ) from exc
+
+        return {
+            'user': user,
+            'course_key': course_key,
+            'has_api_key_permissions': has_api_key_permissions
+        }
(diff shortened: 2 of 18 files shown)


enterprise/filters/enrollment.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def run_filter(self, user: AbstractBaseUser, course_key: Any, mode: str) -> dict[str, Any]:  # pylint: disable=arguments-differ
+        """
+        Post enterprise enrollment and consent if the user is an enterprise customer user.
+        """
+        log.info(
+            "EnterpriseEnrollmentPostProcessor running: user_id=%s, course_key=%s",
+            user.id,
+            str(course_key),
+        )
```
pwnage101, line 54: Nit: Can't hurt to toss in the `mode`.

enterprise/filters/enrollment.py
```diff
@@ -0,0 +1,81 @@
...
+        try:
+            from openedx.features.enterprise_support.api import (  # pylint: disable=import-outside-toplevel
+                ConsentApiServiceClient,
+                EnterpriseApiServiceClient,
+            )
+        except ImportError:
+            return {'user': user, 'course_key': course_key, 'mode': mode}
```
pwnage101: ❌ Use defensive imports at the top level instead of deferred imports within the filter.  Example:
https://github.com/openedx/edx-enterprise/blob/748462bc085fdb9fc956cf722b5902fa643daaff/enterprise/filters/accounts.py#L11-L15

Also update the unit tests to patch these platform imports.  See my other comment for specific instructions.

tests/filters/test_enrollment.py
```diff
@@ -0,0 +1,202 @@
...
+from types import ModuleType
+from unittest.mock import MagicMock, patch
+
+from django.test import TestCase
+
+from enterprise.filters.enrollment import EnterpriseEnrollmentPostProcessor
+from test_utils.factories import UserFactory
+
+
+def _make_mock_api_module():
```
pwnage101: ❌ Both `_make_mock_api_module` and `_make_openedx_modules` are unconventional and a bit hard to grok.  Please stick with the already established pattern of using patch decorators at the top of each test like in https://github.com/openedx/edx-enterprise/blob/master/tests/filters/test_accounts.py .  In your case, there would just be two patch decorators, e.g.:

```python
    @patch('enterprise.filters.enrollment.ConsentApiServiceClient')
    @patch('enterprise.filters.enrollment.EnterpriseApiServiceClient')
    def test_returns_unchanged_args_for_non_enterprise_user(
        self,
        mock_enterprise_client,
        mock_consent_client
    ):
        [...]
```

That said, please don't patch enterprise modules like in that linked file, just patch platform imports such as in my examples above.  For test_accounts.py the AI attempted to actually patch locally available modules (bad), resulting in  this really gibberish mock: https://github.com/openedx/edx-enterprise/blob/master/tests/filters/test_accounts.py#L51C1-L57C21

```python
        mock_ecu = MagicMock()
        (
            mock_ecu_objects.filter.return_value
            .order_by.return_value
            .select_related.return_value
            .first.return_value
        ) = mock_ecu
```

pwnage101 (commented):
So, I'm just going to submit this review but before you implement fixes I'd like to meet with you synchronously because there's a lot to unpack and we should discuss an alternate path.

It looks like the `CourseEnrollmentStarted` filter you were planning to use (possibly adopted from my draft) is installed at the [model layer](https://github.com/edx/edx-platform/blob/release-ulmo/common/djangoapps/student/models/course_enrollment.py#L727), **not** the [view layer](https://github.com/edx/edx-platform/blob/eff8e6a1321532f2e14d6e4f9a30b3d748b599c4/openedx/core/djangoapps/enrollments/views.py#L777-L796) containing the enterprise logic we are trying to replace.

Keep in mind enterprise enrollment creation was [added nearly 10 years ago](https://github.com/openedx/openedx-platform/pull/14542) to the view layer, and meanwhile the model layer enroll() method is called from management commands, other views, entitlements, program enrollments, auto-auth, tests, etc. (in all over 80 call sites), so THERE BE DRAGONS if we change this.

enterprise/api_client/xpert_ai.py
```diff
@@ -28,7 +28,7 @@ def chat_completion(prompt, role):
     }
 
     body = {
-        'messages': [{'role': role, 'content': prompt},],
+        'messages': [{'role': role, 'content': prompt}, ],
```
kiram15: Had to fix unrelated file that was failing on linter

enterprise/filters/enrollment.py
kiram15: Repo has a warning when going over 5 params passed in, figured we didn't need it

enterprise/api_client/xpert_ai.py
pwnage101: The way to fix this is actually to just remove the comma, like this:
```suggestion
        'messages': [{'role': role, 'content': prompt}],
```

enterprise/filters/enrollment.py
pwnage101: I'm guessing you removed mode from the filter signature, so this is no longer possible anyway. Resolving.

enterprise/filters/enrollment.py
```diff
@@ -0,0 +1,95 @@
...
+                course_id=course_id,
+                enterprise_customer_uuid=str(linked_enterprise),
+            )
+        except Exception as exc:
+            log.exception(
+                "Failed to provide enterprise consent for user %s in course %s.",
+                username, course_id,
+            )
+            raise CourseEnrollmentViewStarted.PreventEnrollment(
+                f"Failed to provide enterprise consent for {username} in {course_id}."
```
pwnage101: Nit: the consent has already been "provided".  I think a better verb would be "record" or "post":
```suggestion
                f"Failed to record enterprise consent for {username} in {course_id}."
```

tests/test_enterprise/test_settings.py
```diff
@@ -212,3 +212,29 @@ def test_additions_dict_isolated_from_subsequent_mutation(self):
         existing[FILTER_A]['pipeline'].append(STEP_Y)
 
         assert additions[FILTER_A]['pipeline'] == [STEP_X]
+
+
+class TestEnterpriseFiltersConfig(unittest.TestCase):
```
pwnage101, line 217: I don't mind this test being added, but curious why you decided to add it?  Seems unrelated.

enterprise/filters/enrollment.py
```diff
@@ -0,0 +1,95 @@
...
+                'course_key': course_key,
+                'has_api_key_permissions': has_api_key_permissions
+            }
+
+        username = user.username
+        course_id = str(course_key)
+
+        try:
+            EnterpriseApiServiceClient().post_enterprise_course_enrollment(username, course_id)
+        except Exception as exc:
```
pwnage101: ❌ Catching broad `Exception` could be behavior/logic change, and is also side-stepping the original design of the function.  The old platform logic caught only `EnterpriseApiException`, and `post_enterprise_course_enrollment` is specifically designed to raise `EnterpriseApiException`.

Fix:
```suggestion
        except EnterpriseApiException as exc:
```

enterprise/filters/enrollment.py
```diff
@@ -0,0 +1,95 @@
...
+                f"Failed to post enterprise course enrollment for {username} in {course_id}."
+            ) from exc
+
+        try:
+            ConsentApiServiceClient().provide_consent(
+                username=username,
+                course_id=course_id,
+                enterprise_customer_uuid=str(linked_enterprise),
+            )
+        except Exception as exc:
```
pwnage101: Similarly, there's no need for a broad except here, as provide_consent is only designed to raise HTTPError.  The original platform logic did not catch that error, but since we're already improving upon the original, lets do one better:


```suggestion
        except requests.exceptions.HTTPError as exc:
```

tests/test_enterprise/test_settings.py
kiram15: Not directly related to the work, more related to the general filter initiative as an smoke test

enterprise/filters/enrollment.py
```diff
@@ -0,0 +1,100 @@
...
+    API clients to post the enrollment and provide consent on behalf of the enterprise customer.
+
+    This step is intended to be registered as a pipeline step for the
+    ``org.openedx.learning.course.enrollment.view.started.v1`` filter.
+    """
+
+    def run_filter(  # pylint: disable=arguments-differ
+            self,
+            user: AbstractBaseUser,
+            course_key: Any,
```
pwnage101: I _think_ the type of this was `CourseKey`

tests/filters/test_enrollment.py
```diff
@@ -0,0 +1,226 @@
...
+        mock_request.data = {}
+        mock_get_current_request.return_value = mock_request
+
+        step = self._make_step()
+        result = step.run_filter(
+            user=user,
+            course_key=course_key,
+            has_api_key_permissions=True,
+        )
+        self.assertEqual(
```
pwnage101: Nit: use modern pytest-style assertions.

tests/filters/test_enrollment.py
```diff
@@ -0,0 +1,226 @@
...
+        mock_consent_client,
+        mock_get_current_request,
+    ):
+        """
+        When the user is not linked to an enterprise customer, return the
+        arguments unchanged without calling any API clients.
+        """
+        user = UserFactory.create(username="regular-user")
+        course_key = MagicMock()
+        course_key.__str__.return_value = "course-v1:org+course+run"
```
pwnage101: when updating the type hint to CourseKey, this can be simplified to just
```suggestion
        course_key = CourseKey.from_string("course-v1:org+course+run")
```

tests/filters/test_enrollment.py
```diff
@@ -0,0 +1,260 @@
...
+        mock_enterprise_client,
+        mock_consent_client,
+        mock_get_current_request,
+    ):
+        """
+        When the user is not linked to an enterprise customer, return the
+        arguments unchanged without calling any API clients.
+        """
+        user = UserFactory.create(username="regular-user")
+        course_key = MagicMock()
```
pwnage101: Dead line of code.

tests/filters/test_enrollment.py
```diff
@@ -0,0 +1,260 @@
...
+        When the user request is sent without proper api key permissions, return the
+        arguments unchanged without calling any API clients.
+        """
+        user = UserFactory.create(username="regular-user")
+        enterprise_uuid = uuid.uuid4()
+        mock_request = MagicMock()
+        mock_request.data = {"linked_enterprise_customer": str(enterprise_uuid)}
+        mock_get_current_request.return_value = mock_request
+
+        course_key = MagicMock()
```
pwnage101: Dead line of code.

tests/filters/test_enrollment.py
```diff
@@ -0,0 +1,260 @@
...
+        When the consent API client raises, the step raises PreventEnrollment
+        after the enterprise enrollment post has already happened.
+        """
+        user = UserFactory.create(username="enterprise-learner")
+        enterprise_uuid = uuid.uuid4()
+        mock_request = MagicMock()
+        mock_request.data = {"linked_enterprise_customer": str(enterprise_uuid)}
+        mock_get_current_request.return_value = mock_request
+
+        course_key = MagicMock()
```
pwnage101: Dead line of code.

pwnage101 (commented):
LGTM, just remove the dead code and merge.

kiram15 (approved):

PR #2555, opened 2026-03-05
feat: add pluggable override for course home progress username obfuscation

pwnage101:
ENT-11572

---

Related:
* https://github.com/openedx/edx-enterprise/pull/2555
* https://github.com/openedx/openedx-platform/pull/38108


enterprise/overrides/course_home_progress.py
```diff
@@ -0,0 +1,31 @@
...
+    Arguments:
+        prev_fn: the previous (default) implementation — returns None, not called.
+        request: the current HTTP request.
+        student: the Django User object for the student being viewed.
+
+    Returns:
+        str or None: the generic enterprise name, or None if the learner is not an
+        enterprise SSO user with a configured generic name.
+    """
+    # Deferred import — will be replaced with internal path in epic 17.
```
pwnage101: "epic 17" is claude cruft, change reference to actual ticket:
```suggestion
    # Deferred import — will be replaced with internal path in ENT-11576.
```

enterprise/overrides/course_home_progress.py
```diff
@@ -0,0 +1,31 @@
+"""
+Pluggable override for the course home progress view username obfuscation.
+"""
+import logging
+
+log = logging.getLogger(__name__)
+
+
+def enterprise_obfuscated_username(prev_fn, request, student):  # pylint: disable=unused-argument
```
pwnage101: Nit: disable unused-argument on the specific argument not used
```suggestion
def enterprise_obfuscated_username(
    prev_fn,  # pylint: disable=unused-argument
    request,
    student, # pylint: disable=unused-argument
):
```

pwnage101: ❌ most important requested change: Please install the override in this PR.  That likely means adding this line to `enterprise/settings/common.py`:

```diff
 def plugin_settings(settings):
+    # blah blah blah documentation
+    OVERRIDE_TRANSFORM = 'enterprise.overrides.course_home_progress.enterprise_obfuscated_username'
```

enterprise/overrides/course_home_progress.py
pwnage101: ❌ Also, I think Claude randomly decided to use a **deferred** import (inside the function) instead of a **defensive** import (top-level try/except) based on vibes.  Our established norm is actually to use top-level defensive imports to simplify unit tests.

Look at how complicated the tests have to be just to patch `openedx.features.enterprise_support.utils`.  That also makes unit tests sensitive to upstream changes to module names/paths. It would be way simpler if the tests could just patch `enterprise.overrides.get_enterprise_learner_generic_name`, and then when we migrate enterprise_support later (in ENT-11576) we won't need to update the unit tests, strictly speaking.

tests/test_enterprise/test_overrides_course_home_progress.py
```diff
@@ -0,0 +1,91 @@
...
+    def test_enterprise_learner_returns_generic_name(self):
+        """
+        When get_enterprise_learner_generic_name returns a non-empty string,
+        should return that string.
+        """
+        request = MagicMock()
+        student = MagicMock()
+        expected_name = 'Enterprise Learner'
+
+        with patch(f'{OVERRIDE_MODULE}.get_enterprise_learner_generic_name', return_value=expected_name) as mock_fn:
```
pwnage101: Nit: When patching, I actually prefer we don't use any indirection when referencing the module path.  IMO, Unit tests benefit from being a bit more explicit.  Furthermore, we have nothing planned which would cause this path to break.  Finally, i generally prefer we override using function decorators, not `with` contexts---this pattern is more stylistically consistent with the rest of enterprise backend test files.

pwnage101 (commented):
LGTM! just one minor nit.

marlonkeating (approved):

PR #2561, opened 2026-03-06
chore: drop Python 3.11 support

feanil:
## Summary

- Drop Python 3.11 support: remove from CI test matrix, tox envlist, and package classifiers
- Regenerate pinned requirements using Python 3.12
- Bump version to 7.0.0 — dropping Python support is a breaking change, so this is a major version bump

## Context

Python 3.11 is being dropped across the Open edX ecosystem as part of the move
to standardize on Python 3.12. See the tracking issue for the full list of repos:
https://github.com/openedx/public-engineering/issues/499

## Test plan

- [ ] CI passes with Python 3.12 only

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.12']
-        toxenv: [quality, docs, py312-django52, pii_check]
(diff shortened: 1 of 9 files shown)


pwnage101:
Almost there, looks to be just missing a CHANGELOG.rst entry

pwnage101 (approved):
Approved to unblock

PR #2594, opened 2026-04-10
feat: enable academies by default for all enterprise customers with search (ENT-11220)

sjasti-sonata-svg:
## Context

Currently, Academies is only available to two customers (2U and Siemens). Moving forward, Academies should be a default for all Teams and Enterprise Subscription customers that have search enabled.

### 1. Model default changed (`enterprise/models.py`)
Changed `enable_academies` default from `False` to `True` so that all **new** enterprise customers will have academies enabled automatically.

### 2. Data migration (`enterprise/migrations/0244_enable_academies_for_existing_customers.py`)
Enables academies for all **existing** enterprise customers that have `enable_integrated_customer_learner_portal_search=True`. This ensures existing customers with search enabled also get academies turned on, not just new ones.

## Ticket
https://2u-internal.atlassian.net/browse/ENT-11220

## Test plan
- [ ] Verify new enterprise customers are created with `enable_academies=True`
- [ ] Verify existing customers with `enable_integrated_customer_learner_portal_search=True` get `enable_academies=True` after migration
- [ ] Verify existing customers without search enabled are not affected
- [ ] Verify academies section appears on learner portal for customers with search enabled
- [ ] Verify academies section does not appear when no academies exist in the catalog

image

image

Postman Testing
image

image


rgopalrao-sonata-png (approved):
Migration changes are fine

iloveagent57 (approved):

PR #2602, opened 2026-04-24
fix: update social media links for email compatibility

brobro10000:
Updates the ADMIN_ENROLL email template via a Django data migration to fix social media icon layout issues across email clients.

Details

Adds a new migration that updates the `html_template` field on the `EnrollmentNotificationEmailTemplate` model (template_type=ADMIN_ENROLL)
Refactors the social media section to use a single `<table>` with one` <tr>` and multiple `<td>` cells
Removes usage of display: inline-block on `<table>` elements, which is inconsistently supported in email clients
Preserves all existing links and assets while only modifying layout structure

Why
The previous implementation used multiple inline-block tables, which can be ignored or overridden by clients like Gmail and Outlook, causing icons to stack vertically.
Using a single table row is a widely supported, “bulletproof” email pattern that ensures consistent horizontal rendering.

Impact

Affects the `ADMIN_ENROLL` template stored in the database via migration
Applies to all newly sent enrollment emails after migration is run
No changes to template variables or business logic

Reference

See --> https://www.caniemail.com/search/?s=inline

Screenshot - Note, original implementation also displayed correct layout, but due to email limitation with `inline-block`, was rendering as verticle
Screenshot 2026-04-24 at 11 15 37 AM

**Merge checklist:**

**Post merge:**


enterprise/migrations/0246_enrollment_email_update_02.py
brobro10000, line 1: See https://github.com/openedx/edx-enterprise/blob/master/enterprise/migrations/0137_enrollment_email_update.py for original migration implementation.

enterprise/migrations/0246_enrollment_email_update_02.py
```diff
@@ -0,0 +1,186 @@
...
+                      <td align="center" style="padding:0 7px;Margin:0;font-size:0;line-height:0;">
+                        <a href="https://api.whatsapp.com/send?text=edX" target="_blank">
+                          <img src="https://fzvpwi.stripocdn.email/content/guids/CABINET_4d3c6887b8ac137f656a3dd54bb5f6c8/images/18781614127069989.png" alt="WhatsApp" width="35" height="35" border="0" style="display:block;border:0;outline:none;text-decoration:none;">
+                        </a>
+                      </td>
+                    </tr>
+                  </tbody>
+                </table>
+              </td>
+            </tr>
```
brobro10000, line 138: Bulk of changes located here here compared to https://github.com/openedx/edx-enterprise/blob/master/enterprise/migrations/0137_enrollment_email_update.py

iloveagent57 (approved):
gotta be a better way to do this :)

PR #2617, opened 2026-05-12
feat: own OPEN_EDX_FILTERS_CONFIG injection

pwnage101:
ENT-11830

---

* Blocks: https://github.com/openedx/openedx-platform/pull/38634
* Blocks: https://github.com/edx/edx-platform/pull/281

### DIFF
--- a/enterprise/settings/common.py
+++ b/enterprise/settings/common.py
@@ -1,6 +1,46 @@
 """
 Common plugin settings for the enterprise app.
 """
+from typing import Any
+
+# Shape of an OPEN_EDX_FILTERS_CONFIG map: ``{filter_type: {'fail_silently': bool, 'pipeline': list[str]}}``.
+FiltersConfig = dict[str, dict[str, Any]]
+
+
+ENTERPRISE_FILTERS_CONFIG: FiltersConfig = {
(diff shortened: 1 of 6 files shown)


enterprise/settings/common.py
```diff
@@ -1,6 +1,46 @@
...
 """
+from typing import Any
+
+# Shape of an OPEN_EDX_FILTERS_CONFIG map: ``{filter_type: {'fail_silently': bool, 'pipeline': list[str]}}``.
+FiltersConfig = dict[str, dict[str, Any]]
+
+
+ENTERPRISE_FILTERS_CONFIG: FiltersConfig = {
+    "org.openedx.learning.account.settings.read_only_fields.requested.v1": {
+        "fail_silently": True,
```
kiram15: Shouldn't this be false?

pwnage101: I could change it! But for context this was just copying what's already merged upstream: https://github.com/openedx/openedx-platform/blob/master/lms/envs/common.py#L3291

pwnage101:
Added 1 commit to change the read_only_fields pipeline to NOT fail silently, which keeps that change isolated from the commit to migrate the settings dictionary.

kiram15 (approved):

PR #2639, opened 2026-06-12
fix: downgrade notification banner log level

bcitro:
Downgrades the log level from ERROR to DEBUG for notification banner "request object not found" exceptions to reduce logging noise. These exceptions are expected when the serializer is called from a code path that does not include `request`.

**Merge checklist:**

**Post merge:**

### DIFF
--- a/enterprise/api/v1/serializers.py
+++ b/enterprise/api/v1/serializers.py
@@ -321,12 +321,16 @@ class EnterpriseCustomerSerializer(serializers.ModelSerializer):
         """
 
         try:
-            request = self.context['request']
+            request = self.context.get('request')
+            if not request:
+                # this serializer is called from tpa_pipeline and other code paths
+                # where 'request' is not available
+                LOGGER.debug('[Admin Notification API] Get enterprise notification banner request object not found,'
+                             ' Enterprise Customer :{}'.format(obj.slug))
+                return None
             user_id = request.user.id
         except Exception as exc:  # pylint: disable=broad-except
-            # this serializer is called from tpa_pipeline and other code paths where 'request'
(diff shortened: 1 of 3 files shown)


brobro10000 (approved):

enterprise/api/v1/serializers.py
```diff
@@ -321,11 +321,12 @@ def get_enterprise_notification_banner(self, obj):
...
 
         try:
-            # this serializer is also called from tpa_pipeline and request is not available in the first place there.
             request = self.context['request']
             user_id = request.user.id
         except Exception as exc:  # pylint: disable=broad-except
-            LOGGER.error('[Admin Notification API] Get enterprise notification banner request object not found,'
+            # this serializer is called from tpa_pipeline and other code paths where 'request'
+            # is not available, so some exceptions are expected.
+            LOGGER.debug('[Admin Notification API] Get enterprise notification banner request object not found,'
```
iloveagent57: what about something like this?
```python
request = self.context.get('request')
if not request:
    LOGGER.debug(...)
    return None
user_id = request.user.id
# and continue with remaining logic
```

iloveagent57: this way if something unexpectedly raises (something not pertaining to the presence of `request`), it'll be loud and obvious. or if you wanted to be more defensive, wrap that whole thing above in a broad try/except:
```python
try:
    request = self.context.get('request')
    if not request:
        LOGGER.debug(...)
        return None
    user_id = request.user.id
except Exception as exc:
    LOGGER.error() # we're guaranteed to not be dealing with a keyerror here, so its ok to error
```

PR #2645, opened 2026-06-18
refactor: reorganize helper functions into model-dependent and model-independent to break an import cycle

pwnage101:
Begin to split up `utils.py` functions into two buckets:

* `utils.py`: Model-independent functions go here.  Typical use case would be date parsing, settings fetching, etc.
* `core_api.py`: Model-dependent functions go here.  Typical use case would be fetching a learner's active enterprise, or checking if a learner is enrolled in a course.

ENT-11544

### DIFF
--- a/enterprise/core_api.py
+++ b/enterprise/core_api.py
@@ -1,0 +1,91 @@
+"""
+Core Python API for the enterprise app.
+
+These are core enterprise utility functions that rely on access to enterprise models or data.
+
+Important: ``models.py`` files should NEVER import this module, since that would create an import cycle.
+"""
+from opaque_keys.edx.keys import CourseKey
+
+from django.conf import settings
+from django.contrib.auth.models import AbstractUser
+from django.http import HttpRequest
+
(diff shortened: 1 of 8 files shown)


enterprise/utils.py
```diff
@@ -1,5 +1,9 @@
 """
 Utility functions for enterprise app.
+
+Going forward, functions in this file should be lightweight and NOT depend on
```
kiram15, line 4: I would add that functions that must rely on enterprise models should live in core_api.py

enterprise/core_api.py
```diff
@@ -0,0 +1,91 @@
+"""
+Core Python API for the enterprise app.
+
+Use this module to gain regulated access to enterprise data.
```
kiram15: This is kind of a confusing sentence, I would change it to something like "Use this module for utility functions that rely on access to enterprise models or data".

kiram15 (approved):
Looks good, just a couple nits about comments

PR #2662, opened 2026-07-24
feat: add multiple SSO tenants during devstack provisioning

pwnage101:
The goal is ultimately to support integration testing logistration flows which depend on many different enterprise SSO configurations, including multi-tenant configurations.

Changes directly in service of that goal:

- Update provision-tpa.py to provision multiple SSO tenants (gryffindor, slytherin).
- Add devstack_api.py helpers for SAML providers, enterprise branding, etc.
- Add ENT-11568 integration-test script.

Cleanup work along the way:

- Eliminated legacy code that provisioned several redundant worker and admin users (devstack already provisions them).
- Add `assign_system_wide_enterprise_role` management command to finally fix this: https://github.com/edx/devstack/blob/4e42eeda/provision-enterprise-access.sh#L25
- [Performance] Add --no-create-users flag to seed_enterprise_devstack_data command

ENT-11568

---

Smoke Test
---

0. Add `127.0.0.1 edx.devstack.keycloak` to /etc/hosts.
1. Start relevant devstack services:
   ```
   make dev.up.lms+frontend-app-authn+frontend-app-learner-portal-enterprise
   ```
2. From the edx-enterprise directory, provision keycloak and run this script:
   ```
   make dev.provision.keycloak
   ./scripts/provision-integration-test-ENT-11568.sh
   ```
3. Disable the AuthN MFE in `py_configuration_files/lms.py`:
   - set `ENABLE_AUTHN_MICROFRONTEND = False`
   - `make lms-stop && make lms-up`
4. Open a fresh incognito tab and go to this URL: http://localhost:18000/login?tpa_hint=saml-gryffindor
   - Expect to see a Gryffindor logo.
   - email: `gryffindor_learner@example.com`
   - password: `edx`
5. Open a fresh incognito tab and go to this URL: http://localhost:18000/auth/login/tpa-saml/?idp=gryffindor
   - Expect to get redirected to a keycloak login page titled "GRYFFINDOR".
   - username: `gryffindor_learner`
   - password: `testpass`
6. Repeat steps 3-5 with the AuthN MFE enabled.


enterprise/management/commands/assign_system_wide_enterprise_role.py
```diff
@@ -0,0 +1,102 @@
+"""
+Management command for assigning a system-wide enterprise role to an existing user.
+"""
```
pwnage101, line 3: See also https://github.com/edx/devstack/pull/227

provision-tpa.py
```diff
@@ -75,122 +107,130 @@
...
+# Bypass that guard for this local-only pull only; scoping the patch to this call
+# keeps the security check active everywhere else. The validator is imported by
+# name into the tasks module, so patch it there.
+from unittest import mock
+
+print('\n--- Step 4: Fetch SAML metadata (saml --pull) ---')
 try:
-    call_command('saml', pull=True)
+    with mock.patch('common.djangoapps.third_party_auth.tasks.validate_saml_metadata_url'):
+        call_command('saml', pull=True)
```
pwnage101, line 209: This patch is needed to bypass [recently-added validation in openedx-platform](https://github.com/openedx/openedx-platform/commit/70a56246dd9c9df57c596e64bdd8a11b1d9da054) which forces all SAML providers to use HTTPS.  This is a non-starter for devstack and using keycloak locally.

enterprise/management/commands/seed_enterprise_devstack_data.py
```diff
@@ -55,67 +65,56 @@ def handle(self, *args, **options):
...
-            get_or_create_enterprise_user(
-                username='enterprise_worker',
-                role=ENTERPRISE_OPERATOR_ROLE,
-                applies_to_all_contexts=True,
-            ),
-            get_or_create_enterprise_user(
-                username='ecommerce_worker',
-                role=ENTERPRISE_OPERATOR_ROLE,
-                applies_to_all_contexts=True,
-            ),
```
pwnage101, line 102: For one reason or another, these were all useless cruft.  Devstack itself is responsible for provisioning worker users, and it does a better job at that.

enterprise/management/commands/assign_system_wide_enterprise_role.py
```diff
@@ -0,0 +1,102 @@
...
+
+    help = 'Assign a system-wide enterprise role to an existing user.'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--username',
+            required=True,
+            help='Username of the existing user to assign the role to.',
+        )
+        parser.add_argument(
```
kiram15, line 44: Do we validate this string anywhere based on the roles? Similar to this line https://github.com/openedx/edx-enterprise/blob/4b78a3673b4cba5f0e2695630202b05a71da6139/enterprise/devstack_api.py#L203

CHANGELOG.rst
```diff
@@ -17,6 +17,12 @@ Unreleased
 ----------
 * nothing unreleased
 
+[8.7.2] - 2026-07-24
+---------------------
+* feat: add multiple SSO tenants during devstack provisioning
```
kiram15, line 22: Might also add in a comment about removing feature roles

enterprise/management/commands/assign_system_wide_enterprise_role.py
```diff
@@ -0,0 +1,102 @@
+"""
+Management command for assigning a system-wide enterprise role to an existing user.
```
kiram15, line 2: Is this only intended for devstack use or is it a general ops command? If only for devstack, we should be explicit here or have some kind of settings check

provision-tpa.py
```diff
@@ -4,67 +4,99 @@
 Run inside the LMS container via:
     manage.py lms shell < provision-tpa.py
 
-All configuration is read from environment variables (see keycloak-devstack.env).
+SHARED configuration -- the one Keycloak service and the single LMS service
+provider -- is read from environment variables (see keycloak-devstack.env).
+
+PER-TENANT configuration lives in the TENANTS list below.  A "tenant" is one
```
kiram15, line 10: Such a tiny nit but why are there so many spaces in between the sentences?

kiram15 (approved):
Looks good, but wasn't able to get the redirect to the keycloak login locally in testing

enterprise/devstack_api.py
```diff
@@ -154,18 +266,21 @@ def get_or_create_user(username: str, is_staff: bool = False) -> AbstractBaseUse
...
     except IntegrityError:
         user = User.objects.get(username=username)
         LOGGER.info('Using existing user: %s', username)
 
+    profile_name = f'{first_name} {last_name}'.strip() or 'Test Enterprise User'
     UserProfile.objects.update_or_create(
         user=user,
-        defaults={'name': 'Test Enterprise User'},
+        defaults={'name': profile_name},
     )
```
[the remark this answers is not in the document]
pwnage101: Assume the LMS-side import never fails.

enterprise/devstack_api.py
```diff
@@ -134,18 +137,127 @@ def get_or_create_enterprise_catalog(enterprise_customer: EnterpriseCustomer) ->
...
+        attr_email: SAML attribute mapped to the user's email.
+        attr_first_name: SAML attribute mapped to the user's first name.
+        attr_last_name: SAML attribute mapped to the user's last name.
+
+    Returns:
+        The existing or newly created EnterpriseCustomerIdentityProvider link.
+    """
+    if site is None:
+        site = Site.objects.get_current()
+    provider_config = SAMLProviderConfig(
```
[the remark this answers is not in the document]
pwnage101: Assume the LMS-side import never fails.

provision-tpa.py
pwnage101: Oh it's a habit that my mom taught me because as recently as the 90s that's how all typed text was written.  It's standard for using typewriters.  Furthermore, my IDE is `vim` which includes two spaces after a period by default for over 3 decades and I just haven't changed that setting yet.

But you're right, a single space after a period is the modern convention.

enterprise/management/commands/assign_system_wide_enterprise_role.py
pwnage101: Good idea, i'll codify VALID_ROLES somewhere in constants.py

CHANGELOG.rst
pwnage101: We're not removing feature roles, just direct feature role assignments.  This is effectively a no-op because the feature role assignments were redundant with system-wide role assignments.

enterprise/devstack_api.py
```diff
@@ -336,43 +515,16 @@ def enroll_learner_in_course(
...
+    Django groups are a legacy technique to authorize access to certain older
+    enterprise API endpoints, and is distinct from the newer edx-rbac
+    system-wide enterprise roles.  Until all the legacy APIs consumed by
+    frontend-app-admin-portal are modernized to leverage edx-rbac authz, this
+    step to provision the group memberships are still required.
+    """
     if role == ENTERPRISE_LEARNER_ROLE:
         return
     Group.objects.get(name=ENTERPRISE_DATA_API_ACCESS_GROUP).user_set.add(user)
     Group.objects.get(name=ENTERPRISE_ENROLLMENT_API_ACCESS_GROUP).user_set.add(user)
```
kiram15, line 499: Must use `ensure_enterprise_groups` before this point or it fails in testing on a new devstack with `django.contrib.auth.models.Group.DoesNotExist: Group matching query does not exist.`

enterprise/management/commands/assign_system_wide_enterprise_role.py
pwnage101: I think this is appropriate to keep simple and devstack-agnostic.  It does what's on the tin and has an isolated purpose.  It could even be generally useful in staging.

PR #2685, opened 2026-08-27
fix: add Django admin helper text for demo data field

rgopalrao-sonata-png:
## Summary
Jira ticket:https://2u-internal.atlassian.net/browse/ENT-11523
- Add helper text to `EnterpriseCustomer.enable_demo_data_for_analytics_and_lpr` in the Django admin, clarifying that demo/dummy analytics and Learner Progress Report data only appears when the enterprise user is also registered as an admin on the analytics demo portal, and to contact support or use Django admin to set this up.
- Add migration `0252` applying the same `help_text` change to `EnterpriseCustomer` and `HistoricalEnterpriseCustomer`.
- Bump version to 8.9.4 and update the changelog.

## Test plan
- [x] Verified locally in a running devstack LMS that the `EnterpriseCustomer` admin change page renders the updated help text
- [x] `pycodestyle`/`isort` pass on changed files
image


sjasti-sonata-svg (approved):

enterprise/migrations/0252_alter_enterprisecustomer_enable_demo_data_for_analytics_and_lpr_and_more.py
```diff
@@ -0,0 +1,37 @@
+from django.db import migrations, models
```
brobro10000, line 3: Please generate the migration via the `makemigrations` command from the lms shell.

brobro10000 (changes requested):
Please ensure migrations are correctly generated from the application and not AI generated.

brobro10000 (approved):

PR #2556, opened 2026-03-05
feat: add CalculateEnterpriseDiscountedPrice pipeline step for course mode checkout

pwnage101:
https://2u-internal.atlassian.net/browse/ENT-11573

openedx-filters [#338](https://github.com/openedx/openedx-filters/pull/338) — CourseModePriceRequested filter definition
edx-enterprise [#2556](https://github.com/openedx/edx-enterprise/pull/2556) — CalculateEnterpriseDiscountedPrice pipeline step
edx-platform [#365](https://github.com/edx/edx-platform/pull/365) — view wired to invoke the filter

Prerequisites: local devstack running with all three branches installed ([Runbook](https://2u-internal.atlassian.net/wiki/spaces/SOL/pages/3710353504/Enterprise+Plugin+Ticket+Runbook)); an enterprise customer, an enterprise learner, and a course with a verified mode that has a non-empty SKU and a future expiration date.

### E2E Testing Instructions: 
Note: replace any enterprise customer uuids, users, or course keys with your own applicable local data. 

In lms/envs/devstack.py, confirm/add, then restart after to capture changes:
```
AUTOMATIC_AUTH_FOR_TESTING = True
RESTRICT_AUTOMATIC_AUTH = False
```

**Test 1: Check if pipeline correctly hooked up, and then check against control user with no enterprise link**
Write the test file in lms-shell
```
cat > /tmp/test1.py << 'EOF'
from django.test import Client

client = Client()
response = client.get(
    '/auto_auth',
    {'username': 'e2e_control_user', 'password': 'testpass123'},
    follow=True,
)
print("auto_auth status:", response.status_code)

course_id = "course-v1:edX+DemoX+Demo_Course"
response = client.get(f"/course_modes/choose/{course_id}/", follow=True)
print("choose status:", response.status_code)
content = response.content.decode()
print("contains '149':", "149" in content)
EOF
```
Then execute the file in the bash script `exec(open('/tmp/test1.py').read())`

_Expected: choose status: 200, contains '149': True._

**Test 2: Check against enterprise-linked user with SKU present on course**
```
cat > /tmp/test2.py << 'EOF'
from django.test import Client

client = Client()
client.get('/auto_auth', {'username': 'e2e_enterprise_user', 'password': 'testpass123'}, follow=True)

enterprise_uuid = "8400978a-7eb5-43a8-8a02-aa6063e842e2"
course_id = "course-v1:edX+DemoX+Demo_Course"

[...]

PR #2688, opened 2026-09-03
feat: add support contact tag pipeline step

brobro10000:
ENT-11574

Adds `SupportContactEnterpriseTagStep`, registered in `ENTERPRISE_FILTERS_CONFIG`
(`enterprise/settings/common.py`) against the new `SupportContactContextRequested`
openedx-filter. Appends the `enterprise_learner` tag to support tickets submitted by
requesters linked to a customer account — the same check and tag value as the original
inline `contact_us.py` logic, just moved behind a filter.

Follows the pattern established by the other edx-enterprise filter pipeline steps
(`discounts.py`, `courseware.py`, `enrollment.py`): top-level imports guarded by
`try/except ImportError`, no broad exception swallowing, `fail_silently: False` (matching
every current `ENTERPRISE_FILTERS_CONFIG` entry — errors propagate rather than being
swallowed), and tests built on real Django model factories rather than module-mocking.

Registration lives in `enterprise/settings/common.py`'s `plugin_settings()`, not in
`openedx-platform`/`edx-platform` settings — per the `ENT-11830` ownership handoff (already
merged in all 3 repos before this branch existed).

## Related PRs

* openedx/openedx-filters — filter: https://github.com/openedx/openedx-filters/pull/390
* openedx/openedx-platform — call-site swap, merges last: https://github.com/openedx/openedx-platform/pull/39076
* edx/edx-platform — sibling PR: https://github.com/edx/edx-platform/pull/455

## Testing

New: `tests/filters/test_support.py` — 4 tests for `SupportContactEnterpriseTagStep`, using a
real `UserFactory` instance and `RequestFactory`. The existing generic
`TestEnterpriseFiltersConfig` smoke test in `tests/test_enterprise/test_settings.py` also
covers our new `ENTERPRISE_FILTERS_CONFIG` entry automatically (no changes needed there).

Locally verified: `pytest tests/filters/ tests/test_enterprise/test_settings.py` — 80 passed.
`isort --check-only` clean.

## Changelog / version

Bumped `__version__` to `8.10.0` and added a `CHANGELOG.rst` entry, matching current practice
(e.g. #2556).

Devstack testing (all 3 relevant branches checked out together) is required before merging,
per the enterprise plugin ticket runbook — see handoff prompt in the openedx-platform PR.

### DIFF
--- a/enterprise/filters/support.py
+++ b/enterprise/filters/support.py
@@ -1,0 +1,32 @@
+"""
+Pipeline steps for the support views filters.
+"""
+from crum import get_current_request
+from openedx_filters.filters import PipelineStep
+
+# This import will be replaced with an internal path in ENT-11576 when
+# enterprise_support is migrated into edx-enterprise.
+try:
+    from openedx.features.enterprise_support.api import enterprise_customer_for_request
+except ImportError:
+    enterprise_customer_for_request = None
+
+
+class SupportContactEnterpriseTagStep(PipelineStep):
+    """
+    Append a support-ticket tag for linked customer-account requests.
+
+    This step is intended to be registered as a pipeline step for the
+    ``org.openedx.learning.support.contact.context.requested.v1`` filter.
+    """
+
+    def run_filter(self, tags, user):  # pylint: disable=arguments-differ
+        """
+        Append 'enterprise_learner' to tags if the requester is linked to a customer account.
+        """
+        request = get_current_request()
+        customer = enterprise_customer_for_request(request)
+        if customer and 'enterprise_learner' not in tags:
+            tags = [*tags, 'enterprise_learner']
+
+        return {'tags': tags, 'user': user}
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Unreleased
 ----------
 * nothing unreleased
 
+[8.10.0] - 2026-09-08
+----------------------
+* feat: add SupportContactEnterpriseTagStep pipeline step (ENT-11574)
+
 [8.9.4] - 2026-08-27
 ---------------------
 * fix: add Django admin helper text for ``enable_demo_data_for_analytics_and_lpr`` explaining that
--- a/enterprise/__init__.py
+++ b/enterprise/__init__.py
@@ -2,4 +2,4 @@
 Your project description goes here.
 """
 
-__version__ = "8.9.4"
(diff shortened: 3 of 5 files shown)


enterprise/filters/support.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def run_filter(self, tags, request, user):  # pylint: disable=arguments-differ
+        """
+        Append 'enterprise_learner' to tags if the requester is linked to a customer account.
+        """
+        if enterprise_customer_for_request is None:
+            log.warning('enterprise_support.api is unavailable: skipping support contact tag injection')
+            return {'tags': tags, 'request': request, 'user': user}
+
+        customer = enterprise_customer_for_request(request)
```
pwnage101, line 28:
```suggestion
        request = crum.get_current_request()
        customer = enterprise_customer_for_request(request)
```

brobro10000: Applied — used `from crum import get_current_request` + bare `get_current_request()` instead of the `crum.`-qualified form in the suggestion, to match the existing convention in courseware.py/dashboard.py/enrollment.py/logistration.py/course_modes.py (all use `from crum import get_current_request`). Also removed `request` from the pipeline step's `run_filter` signature and the returned dict, per the matching change on the openedx-filters side. Pushed.

enterprise/filters/support.py
```diff
@@ -0,0 +1,40 @@
...
+    ``org.openedx.learning.support.contact.context.requested.v1`` filter.
+    """
+
+    def run_filter(self, tags, user):  # pylint: disable=arguments-differ
+        """
+        Append 'enterprise_learner' to tags if the requester is linked to a customer account.
+        """
+        if enterprise_customer_for_request is None:
+            log.warning('enterprise_support.api is unavailable: skipping support contact tag injection')
+            return {'tags': tags, 'user': user}
```
pwnage101: Remove this block.  We're going to remove the defensive import in ENT-11576 soon anyway, so lets just not add this so that we don't need to remove it later.

pwnage101 (approved):
just some nits.

enterprise/filters/support.py
brobro10000: Removed the block, and dropped the now-unused `log`/`logging` import that only existed for that warning. Directly calling the guarded-import symbol now, matching discounts.py/courseware.py. Pushed.

PR #2693, opened 2026-09-14
fix: change SupportEnterpriseEnrollmentDataInjector to accept/return full enrollments list

brobro10000:
## Description

Changes `SupportEnterpriseEnrollmentDataInjector.run_filter` to accept/return the full
`enrollments` list and attach `enterprise_course_enrollments` to each matching enrollment
dict **in place**, instead of returning a separate dict keyed by `course_id` that the
platform view had to manually zip back onto `enrollments` via a for-loop.

This is the companion PR to
[openedx-filters#396](https://github.com/openedx/openedx-filters/pull/396), which makes the
matching change to `SupportEnrollmentDataRequested`'s signature.

## Why

Code review flagged this design on edx-platform#462:

- [Passing an empty placeholder dict to the filter is a red flag](https://github.com/edx/edx-platform/pull/462#discussion_r3972331858) —
  a non-enterprise pipeline-step implementer needs the real `enrollments` data to augment,
  not just a hard-coded key to bolt on.
- [The zip-back-together for-loop should be deleted](https://github.com/edx/edx-platform/pull/462#discussion_r3972354578)
  by changing the filter's fundamental design to make in-place changes to `enrollments`. The
  litmus test: the resulting platform file shouldn't contain the word "enterprise" at all.

`user` remains an explicit filter argument (unlike the earlier `SupportContactContextRequested`
fix): this is a support tool where staff look up *other* users' enrollments, so
`crum.get_current_request().user` would incorrectly resolve to the staff member, not the
enrollee.

## Sequencing note

This depends on the openedx-filters companion PR merging/releasing first.
`lms/djangoapps/support/views/enrollments.py` (edx-platform#462, openedx-platform#39083) is
**not** updated by this PR — its call site + tests will be updated once both this PR and the
openedx-filters PR are merged and released, to avoid re-breaking their currently-green CI
against a signature mismatch.

## Testing

- `pytest tests/filters/test_support.py` — 7 passed locally.
- Self-check per CLAUDE.md: 4 files changed (2 effective, excluding the 1:1 test pair),
  ~54 changed lines — under the 400 LoC / 10 file thresholds, no split needed.


enterprise/filters/support.py
```diff
@@ -47,16 +47,16 @@ class SupportEnterpriseEnrollmentDataInjector(PipelineStep):
     """
     Inject enterprise course enrollment data into the support enrollment view.
 
-    Builds a dict of enterprise course enrollments (with data-sharing consent records)
-    keyed by course_id.
+    Attaches enterprise course enrollment records (with data-sharing consent records) to each
+    matching enrollment dict in place, keyed by course_id.
```
pwnage101: "keyed by course_id" is misleading because the added key is actually "enterprise_course_enrollments".  It's not wrong depending how you interpret the sentence, but it's just irrelevant and can easily be confused with the new key being added.  Here's a clearer version:
```suggestion
    Attaches enterprise course enrollment records (with data-sharing consent records) to each
    matching enrollment dict via a new `enterprise_course_enrollments` key.
```

enterprise/filters/support.py
```diff
@@ -47,16 +47,16 @@ class SupportEnterpriseEnrollmentDataInjector(PipelineStep):
...
 
     This step is intended to be registered as a pipeline step for the
     ``org.openedx.learning.support.enrollment.data.requested.v1`` filter.
     """
 
-    def run_filter(self, enrollment_data, user):  # pylint: disable=arguments-differ
+    def run_filter(self, enrollments, user):  # pylint: disable=arguments-differ
```
pwnage101: Nit: All other pipeline steps to date are type hinted, making this the sole exception which lacks types.
```suggestion
    def run_filter(self, enrollments_data: list[dict], user: Any) -> dict:  # pylint: disable=arguments-differ
```

pwnage101 (commented):
Looking good so far!  Just consider updating the `enrollments` variable name and address the nit.

enterprise/filters/support.py
brobro10000: Fixed — updated the docstring to "...to each matching enrollment dict via a new `enterprise_course_enrollments` key", dropping the misleading "keyed by course_id" phrasing.

enterprise/filters/support.py
brobro10000: Fixed — added type hints: `def run_filter(self, enrollments_data: list[dict], user: Any) -> dict:`, matching the other pipeline steps.

enterprise/filters/support.py
```diff
@@ -66,12 +68,18 @@ def run_filter(self, enrollment_data, user):  # pylint: disable=arguments-differ
...
             enterprise_customer_id = ecr.enterprise_customer_user.enterprise_customer_id
             key = f'{course_id}-{enterprise_customer_id}'
             serialized['data_sharing_consent'] = consent_by_key.get(key)
-            enrollment_data.setdefault(course_id, []).append(serialized)
+            enterprise_data_by_course_id.setdefault(course_id, []).append(serialized)
+
+        for enrollment in enrollments_data:
+            enrollment['enterprise_course_enrollments'] = enterprise_data_by_course_id.get(
+                enrollment['course_id'], []
+            )
```
pwnage101, line 83: Nit: i'm not going to block on this but conventionally in edX code when the arguments break into a new line we split them up:
```suggestion
            enrollment['enterprise_course_enrollments'] = enterprise_data_by_course_id.get(
                enrollment['course_id'],
                [],
            )
```

pwnage101 (approved):

PR #2697, opened 2026-09-22
feat: implement the branding learner name and portal link overrides

pwnage101:
Implements two branding override functions:

- enterprise_learner_generic_name() returns the generic name configured by a customer that replaces sensitive SSO usernames, delegating otherwise.
- enterprise_learner_portal_link() returns the customer's learner portal URL, logo and name, delegating when the viewer has no portal-enabled customer.

ENT-12353


enterprise/overrides/branding.py
```diff
@@ -0,0 +1,112 @@
...
+            returned, whenever no enterprise generic name applies -- the platform's
+            callers render this result directly, with no fallback of their own.
+        user: the Django User object whose name is being displayed. Not necessarily the
+            user making the request: the progress page displays the student being viewed,
+            and the user dropdowns display the real user behind a masquerade.
+
+    Returns:
+        str: the enterprise generic name, or whatever the previous implementation returns.
+    """
+    request = get_current_request()
```
rgopalrao-sonata-png, line 62: The except ImportError fallback makes sense for this repo's tests, but it sets the helpers to None, and the overrides only check request is None. If the enterprise_support.utils import ever fails inside the LMS (for example if ENT-11576 moves it), a real request gets past that check and calls None(request), which raises TypeError: 'NoneType' object is not callable. So the header errors instead of falling back to the platform defaults.

Could we also return early when the helper is missing?


if request is None or get_enterprise_learner_generic_name is None:
    return prev_fn(user=user)

enterprise/overrides/branding.py
```diff
@@ -0,0 +1,112 @@
...
+
+    Arguments:
+        prev_fn: the previous (default) implementation. Must be called, and its result
+            returned, whenever the viewer has no portal-enabled customer.
+
+    Returns:
+        dict or None: the portal's absolute URL, logo and name, or whatever the previous
+        implementation returns.
+    """
+    request = get_current_request()
```
rgopalrao-sonata-png, line 99: can we add it here 
if request is None or get_enterprise_learner_portal is None:
    return prev_fn()

enterprise/overrides/branding.py
pwnage101: Soon I will migrate the entire enterprise_support module into this repository, and when I do that I don't want to edit every function to remove defensive function calls.  Writing it this way allows me to minimize the amount of code changes that will be required for that future work.

enterprise/overrides/branding.py
pwnage101: See https://github.com/openedx/edx-enterprise/pull/2697#discussion_r4097344388

